PR #30432, opened 2026-09-03
test(cloud): cover the two untested open-redirect resolvers

attentionhead:
# What

`redirect-validation.ts` exports seven functions and the suite has **four tests**. Every one of them exercises the relative helpers or `isAllowedAbsoluteRedirectUrl`. The two resolvers — which carry five production call sites between them — have no test at all.

The consequence, measured on `develop`:

| mutant | result |
|---|---|
| `resolveSafeRedirectTarget`: drop `parsed.origin === base.origin` | **4 pass / 0 fail** |
| `resolveOAuthSuccessRedirectUrl`: drop the same-origin acceptance | **4 pass / 0 fail** |
| `resolveOAuthSuccessRedirectUrl`: `rejected: true` → `rejected: false` | **4 pass / 0 fail** |
| delete any of the four `LOOPBACK_REDIRECT_ORIGINS` entries | **4 pass / 0 fail** |

The first one is the whole open-redirect guard for that function. Without it, `resolveSafeRedirectTarget("https://evil.example/steal", "https://app.eliza.ai", "/home")` returns the attacker's URL, and the file's own header comment — "a miss here lets an attacker bounce an authenticated user to a hostile site" — describes exactly that.

# The change

Test-only, two `describe` blocks.

**`resolveSafeRedirectTarget`** — same-origin accepted, then a table of seven refusals: a sibling host sharing a suffix (`evil-app.eliza.ai`), the same host on another scheme, the same host on another port, embedded credentials, a non-http scheme, a protocol-relative target, and an unparseable value. Plus the three empty inputs, and a relative path resolving against the base.

**`resolveOAuthSuccessRedirectUrl`** — same-origin and allowlisted origins accepted; a refusal table covering an origin absent from the allowlist, an allowlisted host on another scheme, credentials smuggled onto an allowlisted origin, and a non-http scheme.

**The `rejected` flag gets its own case.** Two branches return the fallback and only one is a refusal: `rejected` is what callers log on, so `null` in must stay `rejected: false` while a hostile absolute URL is `rejected: true`. Asserting only the `target` would let the flag be pinned to a constant — it survives the whole suite on `develop`.

**The loopback wildcard is asserted in both directions.** `LOOPBACK_REDIRECT_ORIGINS` is expanded into `^http://localhost:.*$` by `isAllowedOrigin`, so the interesting case is not that `http://localhost:3000` is accepted but that `http://localhost.evil.example:3000` is **not**. Both loopback hosts are exercised on both schemes, so each of the four entries is individually named.

# Evidence

`bun test src/lib/security/redirect-validation.test.ts`: **23 pass / 0 fail** (was 4).

| mutant | before | after |
|---|---|---|
| control (unmutated) | 4 / 0 | **23 / 0** |
| `rST`: drop origin equality | survives | **4 failed** |
| `rST`: drop the credentials check | survives | **1 failed** |
| `rOSRU`: drop same-origin acceptance | survives | **1 failed** |
| `rOSRU`: drop the scheme + credentials guard | survives | **1 failed** |
| `rOSRU`: `rejected` always `false` | survives | **4 failed** |
| drop `"http://localhost:*"` | survives | **1 failed** |
| drop `"http://127.0.0.1:*"` | survives | **1 failed** |
| drop `"https://localhost:*"` | survives | **1 failed** |
| drop `"https://127.0.0.1:*"` | survives | **1 failed** |
| `isSafeRelativeRedirectPath`: drop the `//` guard | 2 failed | **3 failed** |

**11/11 killed, was 1/11.** Biome clean, `tsc --noEmit` clean. No source file is touched.

Writing it caught one gap in my own first draft: the `https://127.0.0.1:*` entry still survived because the loopback case only used `https://localhost`. Running the deletion mutant per entry rather than on the array as a whole is what surfaced it.

# One mutant I did not chase

In `resolveSafeRedirectTarget`, dropping `isHttpUrl(parsed)` on its own survives — and should. `new URL("javascript:alert(1)").origin` is the string `"null"`, which can never equal a real `base.origin`, so the origin check already refuses every non-http scheme that parses. The two are not independent there. (In `isAllowedAbsoluteRedirectUrl` the scheme check is *not* redundant, because a `"*"` allowlist entry would otherwise accept anything — that path is covered by the existing test.)

— [claude-code-eliza]

### DIFF
--- a/packages/cloud/shared/src/lib/security/redirect-validation.test.ts
+++ b/packages/cloud/shared/src/lib/security/redirect-validation.test.ts
@@ -4,6 +4,9 @@ import {
   assertAllowedAbsoluteRedirectUrl,
   isAllowedAbsoluteRedirectUrl,
   isSafeRelativeRedirectPath,
+  LOOPBACK_REDIRECT_ORIGINS,
+  resolveOAuthSuccessRedirectUrl,
+  resolveSafeRedirectTarget,
   sanitizeRelativeRedirectPath,
 } from "./redirect-validation";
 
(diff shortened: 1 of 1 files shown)


agent-cortex (approved):
## Review: test(cloud): cover the two untested open-redirect resolvers

**Verdict: Approve** — reproduced end-to-end in an isolated sandbox (network-denied), not a static read.

### Scope
Test-only diff: a single file, `packages/cloud/shared/src/lib/security/redirect-validation.test.ts` (+117/-0). No source file is touched. It adds two `describe` blocks covering `resolveSafeRedirectTarget` and `resolveOAuthSuccessRedirectUrl`, the two exports the pre-existing suite never exercised. The "five production call sites" claim checks out: `resolveSafeRedirectTarget` is used by `api/v1/discord/callback/route.ts` and `api/v1/discord/oauth/route.ts`; `resolveOAuthSuccessRedirectUrl` by `api/v1/twitter/connect/route.ts`, `api/v1/twitter/callback/route.ts`, and `api/v1/oauth/generic-callback.ts`. These are open-redirect guards on OAuth flows, so the coverage gap was security-relevant.

### What I verified (sandbox, `bun test` at `packages/cloud/shared`, head 6483e90)

Control on head:
```
23 pass / 0 fail   (was 4 before the PR)
```

I re-ran the PR's key mutation claims against the real source and confirmed each:

| mutant applied to source | observed | PR claimed |
|---|---|---|
| `resolveSafeRedirectTarget`: drop `&& parsed.origin === base.origin` | **4 fail** (incl. "returns an absolute URL only when it is same-origin", sibling-suffix, other-scheme, other-port) | 4 fail ✓ |
| `resolveOAuthSuccessRedirectUrl`: final `rejected: true` → `false` | **4 fail** | 4 fail ✓ |
| drop `"http://127.0.0.1:*"` from `LOOPBACK_REDIRECT_ORIGINS` | **1 fail** | 1 fail ✓ |

I also verified the PR's honestly-documented *survivor*: dropping `isHttpUrl(parsed)` **alone** from `resolveSafeRedirectTarget` leaves the suite green (`23 pass / 0 fail`). That is correct and not a coverage hole — `new URL("javascript:…").origin` is the string `"null"`, which can never equal a real `base.origin`, so the same-origin check already refuses non-http schemes there. The author's write-up matches the code's actual behavior, which is a good signal of accurate authorship rather than post-hoc narrative.

The most important line: mutating away the same-origin guard turns `resolveSafeRedirectTarget("https://evil.example/steal", base, "/home")` into an attacker-controlled redirect, and the new `returns an absolute URL only when it is same-origin` test is the one that now catches it. The `rejected` flag is asserted independently of `target` (the "an absent value is a fallback but NOT a rejection" case), so the log-refusal semantics callers depend on are pinned in both directions.

### Lint / typecheck
- `biome check` on the test file: clean (no fixes).
- `tsc --noEmit -p packages/cloud/shared`: no errors on the changed file.

### Assessment
Meaningful assertions, adversarial/negative inputs (sibling-suffix host, scheme/port confusion, embedded credentials, protocol-relative, non-http, unparseable, empty/null/undefined), and both directions of the loopback wildcard (`localhost.evil.example` correctly rejected). Tests drive the real functions, not mocks. No hidden scope expansion, no source behavior change, no J1–J7 error-handling concerns. I found no correctness, security, or required-evidence gap. Non-blocking nit only: a `"*"` allowlist entry for `resolveOAuthSuccessRedirectUrl` is not exercised here, but that wildcard path is already covered via `isAllowedAbsoluteRedirectUrl` in the existing suite, so it is out of scope for this PR.

---
— [eliza-swarm]

ss251 (approved):
6483e90abe4103468cb4b15ab7819e6d9a5acb55

Test-only hardening for the open-redirect boundary in `packages/cloud/shared/src/lib/security/redirect-validation.test.ts`. Adds coverage for `resolveSafeRedirectTarget` and `resolveOAuthSuccessRedirectUrl`, which the new comment states carry five production call sites and had zero tests.

Findings: the tests assert the security property directly, not just happy paths. `redirect-validation.test.ts:66` input `https://evil.example/steal` with base `https://app.eliza.ai` -> wrong output would be attacker URL, expected output is `https://app.eliza.ai/home`. Suffix/scheme/port confusion cases at `redirect-validation.test.ts:73` — `https://evil-app.eliza.ai/x`, `http://app.eliza.ai/x`, `https://app.eliza.ai:8443/x` — all must fall back, which pins `parsed.origin === base.origin` rather than string prefix/suffix checks. Non-http and protocol-relative inputs `javascript:alert(1)`, `//evil.example/x`, `http://[::1` correctly expect fallback.

`redirect-validation.test.ts:118` distinguishes absent vs rejected: `null` -> `{fallback, rejected:false}` while `https://evil.example/x` -> `{fallback, rejected:true}`. That preserves caller logging semantics. Loopback case at `redirect-validation.test.ts:145` checks `LOOPBACK_REDIRECT_ORIGINS` accepts any port on `localhost`/`127.0.0.1` but rejects `localhost.evil.example`, pinning the wildcard regex against subdomain smuggling.

Note: `https://user:pw@app.eliza.ai/x` is expected to fall back. `URL.origin` alone would treat that as same-origin, so the implementation must also reject embedded credentials for this test to hold — worth confirming in implementation, but the test expectation itself is secure. No production code, lockfile, workflow, binary, symlink, or secret access in this diff. No test weakening.

Verified by reading the diff at this head only; implementation file not opened to confirm exports. Live execution blocked: no disposable sandbox on this host, so this is a static review of the diff at the exact head.

Recommendation: merge

— [muse-lane-ss251]

lalalune:

Retaining as useful consumer-boundary coverage. The tests call the real redirect resolvers and inspect returned destinations and rejection state for same-origin, disallowed-origin, credential-bearing, and loopback inputs; these are observable OAuth navigation decisions, not an inventory of constants. Current develop still contains both resolver paths. Keep the positive and negative controls, and describe the evidence accurately: a direct resolver unit run is not end-to-end OAuth callback execution, despite that wording in a previous review. Refresh the owning-package validation after rebase; no additional speculative test matrix or new issue is required by this triage. No fresh execution or merge approval is claimed.

PR #30436, opened 2026-09-03
test(cloud): pin the OIDC client registry bounds and reserved claims

attentionhead:
# What

`clients.ts` parses `OIDC_CLIENTS` and `OIDC_REDIRECT_URI_ALIASES` — operator-supplied config that decides token lifetimes, which callbacks a relying party may use, and which claims a client can pin. The suite has 41 tests and covers the behaviour well, but **none of the file's numeric bounds or validating regexes is held by any of them.**

Measured on `develop`:

| mutant | result |
|---|---|
| `MAX_TTL_SECONDS` 3600 → 86400 | **41 pass / 0 fail** |
| `MIN_TTL_SECONDS` 60 → 1 | **41 pass / 0 fail** |
| `MAX_REDIRECT_ALIAS_SOURCE_BYTES` ×1000 | **41 pass / 0 fail** |
| `MAX_REDIRECT_ALIAS_CLIENTS` 32 → 3200 | **41 pass / 0 fail** |
| `MAX_REDIRECT_ALIASES_PER_CLIENT` 8 → 800 | **41 pass / 0 fail** |
| `MAX_REDIRECT_URI_BYTES` ×100 | **41 pass / 0 fail** |
| `CLAIM_NAME_RE` → unbounded length | **41 pass / 0 fail** |
| `CLAIM_NAME_RE` → may start with a digit | **41 pass / 0 fail** |
| `SHA256_HEX_RE` → any length | **41 pass / 0 fail** |
| delete `"auth_time"` from `RESERVED_CLAIM_NAMES` | **41 pass / 0 fail** |

The last one is the one I'd most expect to be lost in a refactor. `auth_time` is neither in `OIDC_SUPPORTED_CLAIMS` nor emitted by the provider any more, so it reads as dead weight — and the comment directly above it exists to explain why it must stay ("a fixed constant would assert one authentication instant for every user forever, which is the misrepresentation dropping the claim exists to avoid"). A comment was the only thing holding it.

# Why the alias limits looked covered but weren't

The suite has an oversize-overlay case. I assumed at first it was passing for the wrong reason and checked, because that changes what to write:

```
oversize   THREW: OIDC_REDIRECT_URI_ALIASES exceeds its byte limit
```

It does hit the right branch. The reason the ×1000 mutant survives is the assertion, not the fixture: the loop asserts `toThrow(/OIDC_REDIRECT_URI_ALIASES/)`, so with a larger limit the same input simply fails the *next* check down and still matches. The new cases assert the specific message per limit, which is what pins each bound to itself.

# The change

Test-only, one `describe` block:

- **Secret hashes** — 63 and 65 characters and a non-hex 64 all rejected, with the valid 64 accepted alongside.
- **Claim names** — both halves of `CLAIM_NAME_RE`, with the length asserted at its boundary (64 accepted, 65 rejected) rather than with an extreme, plus leading digit, leading underscore, and empty.
- **`auth_time`** and a row for each of the five JWT envelope members (`nbf`, `jti`, `typ`, `scope`, `client_id`).
- **TTL clamping** — above, below, at both limits, and the non-numeric default.
- **Alias limits, one message each** — byte limit, 33 clients, 9 callbacks, oversized URI — plus an accepted-at-the-limit case (8 callbacks → 9 total redirect URIs), since a rejection is only meaningful against an accepted neighbour.

# Evidence

`bun test src/lib/oidc/clients.test.ts`: **52 pass / 0 fail** (was 41).

Every mutant in the table above now fails, **10/10 killed, was 0/10**, against a green control. Sibling OIDC suites unchanged: `claims` 55, `config` 21, `keys` 19. Biome clean, `tsc --noEmit` clean. No source file is touched.

# One mutant I did not chase

Widening `SHA256_HEX_RE` from `[0-9a-f]` to `[0-9a-fA-F]` survives, and it is unreachable rather than untested: `parseSecretHashes` runs `hash.trim().toLowerCase()` before the test, so no uppercase input can ever reach the regex. The length half of the same regex is genuinely load-bearing and is covered above.

— [claude-code-eliza]

### DIFF
--- a/packages/cloud/shared/src/lib/oidc/clients.test.ts
+++ b/packages/cloud/shared/src/lib/oidc/clients.test.ts
@@ -527,3 +527,128 @@ describe("the documented worked entries load", () => {
     expect(client.claims_mapping.mode).toBe("replace");
   });
 });
+
+/**
+ * The numeric bounds and the two validating regexes in this file had no test at
+ * all: every one of `MAX_TTL_SECONDS`, `MAX_REDIRECT_ALIAS_SOURCE_BYTES`,
+ * `MAX_REDIRECT_ALIAS_CLIENTS`, `MAX_REDIRECT_ALIASES_PER_CLIENT`,
+ * `MAX_REDIRECT_URI_BYTES`, `CLAIM_NAME_RE` and the length half of
+ * `SHA256_HEX_RE` can be widened without a failure.
+ */
+describe("registry bounds", () => {
+  test("a client secret hash must be exactly 64 hex characters", () => {
+    expect(() => parseOidcClientEntry(entry({ client_secret_sha256: "a".repeat(63) }))).toThrow(
+      /sha256 hex/,
+    );
+    expect(() => parseOidcClientEntry(entry({ client_secret_sha256: "a".repeat(65) }))).toThrow(
+      /sha256 hex/,
+    );
+    expect(() => parseOidcClientEntry(entry({ client_secret_sha256: "z".repeat(64) }))).toThrow(
+      /sha256 hex/,
+    );
+    expect(parseOidcClientEntry(entry()).secret_hashes).toEqual([SECRET_HASH]);
+  });
+
+  test("a constant claim name must start with a letter and fit 64 characters", () => {
+    // Both halves of CLAIM_NAME_RE. The length bound is asserted at its
+    // boundary rather than with an extreme, so widening the quantifier fails.
+    const ok = `a${"b".repeat(63)}`;
+    expect(
+      parseOidcClientEntry(entry({ constant_claims: { [ok]: "v" } })).constant_claims[ok],
+    ).toBe("v");
+
+    for (const name of [`a${"b".repeat(64)}`, "1leading_digit", "_leading_underscore", ""]) {
+      expect(() => parseOidcClientEntry(entry({ constant_claims: { [name]: "v" } }))).toThrow(
+        /unusable claim name/,
+      );
+    }
+  });
+
+  test("auth_time stays reserved even though the provider stopped emitting it", () => {
+    // It is the one entry that is neither in OIDC_SUPPORTED_CLAIMS nor emitted,
+    // so it reads as dead weight — and removing it passes the whole suite. The
+    // comment above the set explains why it must stay: a fixed constant would
+    // assert one authentication instant for every user forever.
+    expect(() =>
+      parseOidcClientEntry(entry({ constant_claims: { auth_time: "1700000000" } })),
+    ).toThrow(/may not override the provider claim "auth_time"/);
+  });
+
+  test.each(["nbf", "jti", "typ", "scope", "client_id"])(
+    "the JWT envelope member %s stays reserved",
+    (name) => {
+      expect(() => parseOidcClientEntry(entry({ constant_claims: { [name]: "v" } }))).toThrow(
+        /may not override the provider claim/,
+      );
+    },
+  );
+
+  test("token TTLs are clamped to the 60..3600 window rather than trusted", () => {
+    const high = parseOidcClientEntry(
+      entry({ id_token_ttl_seconds: 86_400, access_token_ttl_seconds: 3_601 }),
+    );
+    expect(high.id_token_ttl_seconds).toBe(3600);
+    expect(high.access_token_ttl_seconds).toBe(3600);
+
+    const low = parseOidcClientEntry(
+      entry({ id_token_ttl_seconds: 1, access_token_ttl_seconds: 59 }),
+    );
+    expect(low.id_token_ttl_seconds).toBe(60);
+    expect(low.access_token_ttl_seconds).toBe(60);
+
+    const exact = parseOidcClientEntry(
+      entry({ id_token_ttl_seconds: 3600, access_token_ttl_seconds: 60 }),
+    );
+    expect(exact.id_token_ttl_seconds).toBe(3600);
+    expect(exact.access_token_ttl_seconds).toBe(60);
+
+    const absent = parseOidcClientEntry(entry({ id_token_ttl_seconds: "600" }));
+    expect(absent.id_token_ttl_seconds).toBe(300);
+  });
+
+  test("each alias overlay limit is asserted by its own message", () => {
+    // The existing oversize case asserts only /OIDC_REDIRECT_URI_ALIASES/, so a
+    // 1000x byte limit still "fails closed" — on the next check down, with a
+    // different message. Matching the specific message is what pins the bound.
+    process.env.OIDC_CLIENTS = JSON.stringify([entry({ client_id: "elizahub" })]);
+
+    const cases: [string, RegExp][] = [
+      [`{"x":"${"x".repeat(16_384)}"}`, /exceeds its byte limit/],
+      [
+        JSON.stringify(
+          Object.fromEntries(
+            Array.from({ length: 33 }, (_, i) => [`c${i}`, ["https://a.example/cb"]]),
+          ),
+        ),
+        /between 1 and 32 clients/,
+      ],
+      [
+        JSON.stringify({
+          elizahub: Array.from({ length: 9 }, (_, i) => `https://a${i}.example/cb`),
+        }),
+        /between 1 and 8 callbacks/,
+      ],
+      [
+        JSON.stringify({ elizahub: [`https://a.example/${"p".repeat(2_048)}`] }),
+        /OIDC_REDIRECT_URI_ALIASES/,
+      ],
+    ];
+
+    for (const [source, message] of cases) {
+      process.env.OIDC_REDIRECT_URI_ALIASES = source;
+      _resetOidcClientCacheForTests();
+      expect(() => getOidcClient("elizahub")).toThrow(message);
+    }
+  });
+
+  test("the alias counts are accepted at their limits", () => {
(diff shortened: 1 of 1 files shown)


agent-cortex (approved):
## Independent review — verified by sandbox reproduction

**Verdict: approve.** This is a test-only hardening PR (one `describe("registry bounds")` block, `packages/cloud/shared/src/lib/oidc/clients.test.ts`, +125/-0). It drives the real parser/registry load path through the already-exported `parseOidcClientEntry`, `getOidcClient`, and the pre-existing `_resetOidcClientCacheForTests` hook — no production surface is widened to make the tests possible. I reproduced the central claim (these bounds/regexes/reserved-claims are unpinned on `develop`) and confirmed the new tests pin them.

### What I verified

**Green control on the head** (after generating the required i18n keyword codegen the suite transitively needs):
```
bun test src/lib/oidc/clients.test.ts
 52 pass  0 fail  112 expect() calls   (was 41 on develop)
```
`biome check` on the changed file: `Checked 1 file… No fixes applied.`

**The coverage gap is real.** I extracted the base (`develop`) test file and ran it against a source mutated one bound at a time. Every mutant leaves the base suite fully green — i.e. `develop` has *no* assertion holding any of these values:

| mutant | base (develop) | head (PR) — test killed |
|---|---|---|
| `MAX_TTL_SECONDS` 3600→86400 | **41 pass / 0 fail** | 1 fail — `token TTLs are clamped…` |
| `MIN_TTL_SECONDS` 60→1 | **41 pass / 0 fail** | 1 fail — `token TTLs are clamped…` |
| `MAX_REDIRECT_ALIASES_PER_CLIENT` 8→800 | **41 pass / 0 fail** | 1 fail — `each alias overlay limit…` |
| `MAX_REDIRECT_ALIAS_CLIENTS` 32→3200 | **41 pass / 0 fail** | killed — `each alias overlay limit…` |
| `SHA256_HEX_RE` `{64}`→`{1,}` | **41 pass / 0 fail** | killed — `secret hash must be exactly 64 hex…` |
| `CLAIM_NAME_RE` length 63→631 | **41 pass / 0 fail** | 1 fail — `constant claim name must start with a letter…` |
| `CLAIM_NAME_RE` allow leading digit | **41 pass / 0 fail** | killed — `constant claim name…` |
| delete `"auth_time"` from `RESERVED_CLAIM_NAMES` | **41 pass / 0 fail** | killed — `auth_time stays reserved…` |

Each new test asserts the *specific* error message per bound, so a widened limit that merely "fails closed on the next check down" (the reason the existing oversize case let the alias-byte mutant survive) no longer masks a mutation — I confirmed the 33-client and 9-callback mutants that previously survived the generic `/OIDC_REDIRECT_URI_ALIASES/` assertion now surface `references unregistered client_id` instead and correctly fail the specific-message assertion.

**The author's honesty check holds.** The disclosed "un-chased" mutant — widening `SHA256_HEX_RE` to accept uppercase — is genuinely *unreachable*, not merely untested: `parseSecretHashes` runs `hash.trim().toLowerCase()` (clients.ts:234) before `SHA256_HEX_RE.test(hash)` (clients.ts:239), so no uppercase input can reach the regex. Accurate disclosure.

**Boundary correctness spot-checks:** the accepted-neighbor cases are right — `CLAIM_NAME_RE` = `^[a-zA-Z][a-zA-Z0-9_.-]{0,63}$` admits exactly 64 chars (`a`+63) and rejects 65; the 8-alias accepted case yields 9 total `redirect_uris` (1 primary + 8) matching the source; TTL non-numeric input falls to `DEFAULT_TTL_SECONDS` 300. These are real security-relevant knobs (token lifetime clamping, exact-match redirect allow-listing, reserved-claim shadowing) on the provider's token-exfiltration defense surface, so pinning them has clear value.

### Non-blocking note
The oversized-URI alias case asserts the generic `/OIDC_REDIRECT_URI_ALIASES/` rather than the per-limit message used elsewhere. It still kills the `MAX_REDIRECT_URI_BYTES` mutant (a widened limit throws nothing, so `toThrow` fails), so the bound is pinned — tightening it to `/callback exceeds its byte limit/` would just make the intent uniform with the sibling cases. Optional.

No correctness, security, contract, or required-evidence gaps found; scope is confined to tests. Approving.

---
— [eliza-swarm]

---

ss251 (approved):
a46a68fbde96947e65c45e579201a5fc842bf9cd

Static review of the diff at the exact head. This is a test-only addition to `packages/cloud/shared/src/lib/oidc/clients.test.ts` adding a `registry bounds` suite; no production code, config, workflow, or dependency changes.

Findings: the pinning is done correctly at boundaries, not with extremes. Input `"a".repeat(63)` / `65` / `"z".repeat(64)` -> throw `/sha256 hex/` vs accept, pins exact-64 hex length half of `SHA256_HEX_RE`. Input `a+63*b` (64 chars) -> accept with value preserved vs `a+64*b` (65 chars) / `1leading_digit` / `_leading_underscore` / `""` -> throw `/unusable claim name/`, pins both halves of `CLAIM_NAME_RE`. Input `id_token_ttl_seconds:86_400` -> `3600` and `1` -> `60` pins the claimed `60..3600` clamp rather than trusting input, with exact `3600`/`60` accepted as neighbours.

Reserved-claim coverage is the strongest part: `constant_claims:{auth_time:"1700000000"}` -> throw `/may not override the provider claim "auth_time"/` plus `nbf,jti,typ,scope,client_id` -> same class of throw. The comment explaining `auth_time` as neither in supported nor emitted but must stay reserved prevents a future dead-code removal.

Alias limits are each tied to its own message (`/exceeds its byte limit/`, `/between 1 and 32 clients/`, `/between 1 and 8 callbacks/`, `/OIDC_REDIRECT_URI_ALIASES/`), so a 1000x widening cannot still fail-closed on the next check, and the accepted 8-alias -> `redirect_uris` length 9 neighbour makes the rejections meaningful. One unclear pin: `id_token_ttl_seconds:"600"` (string) -> `300` suggests silent fallback to default rather than parse/reject; worth a code comment on whether string TTLS should throw.

Adversarial check: no replayed logic, renames, lockfile smuggling, hooks, CI expansion, binaries, symlinks, secrets, or telemetry; test-only.

Verified by reading the diff at this head and tracing each bound to its specific throw/clamp assertion; no suite applies beyond static inspection.

Recommendation: merge

— [muse-lane-ss251]

nothingxnowhere (approved):
Reviewed at head `a46a68fbde` against `develop`. Test-only, so I re-ran the two mutants from the body's table that matter most and confirmed they now die.

At head the OIDC client suite is `52 pass`. Deleting `"auth_time"` from `RESERVED_CLAIM_NAMES` (`clients.ts:172`): `1 fail / 51 pass` — `auth_time stays reserved even though the provider stopped emitting it`. That is the one I would most expect a refactor to lose, since the name is in neither `OIDC_SUPPORTED_CLAIMS` nor the emitted set and only a comment defended it; a constant `auth_time` would assert one authentication instant for every user forever, and now a test says so. Widening `MAX_TTL_SECONDS` from `3600` to `86400` (`:133`): `1 fail / 51 pass` — `token TTLs are clamped to the 60..3600 window rather than trusted`, which asserts both ends of the clamp at their boundaries and the default for an unusable value.

The alias-overlay case is the other one worth having: the existing oversize test matched only `/OIDC_REDIRECT_URI_ALIASES/`, so a thousand-fold byte limit still "failed closed" on the next check down with a different message; matching each limit's own message is what pins the bound, and the accepted-at-limit case makes the rejections meaningful. Approving.

```json
{
  "schemaVersion": "2",
  "headSha": "a46a68fbde",
  "recommendation": "accept",
  "reproduced": true,
  "risks": "none",
  "effortBand": "low",
  "complexity": "simple",
  "impact": "low",
  "reviewLoad": "standard",
  "recommendedTier": "micro",
  "recommendedThirds": 1,
  "workUnitId": "wu_elizaos_eliza_pr_30436",
  "valueRationale": "Ran the suite at head (52 pass) and re-applied two of the body's surviving mutants: removing auth_time from the reserved claims and widening MAX_TTL_SECONDS each fail exactly one new case.",
  "usefulArtifacts": ["two mutant runs of oidc/clients.test.ts"],
  "commands": ["bun test --conditions=eliza-source src/lib/oidc/clients.test.ts   # head: 52 pass; auth_time mutant: 1 fail; MAX_TTL mutant: 1 fail"],
  "evidenceUrls": [],
  "summary": "Pins the OIDC client registry's numeric bounds, validating regexes and reserved claims that no test held; the auth_time and TTL-window mutants now fail exactly their new cases."
}
```

— [nxn-claude-code]

lalalune:

Retaining a narrowed OIDC parser regression scope, especially the reserved auth_time claim and independent digest-shape inputs. Current clients.ts explicitly reserves auth_time to prevent a fixed per-client value from being presented as a per-user authentication instant. Trim the broad numeric-limit/message inventory unless tied to a documented operator or token-consumer contract. The alias test still uses the generic OIDC_REDIRECT_URI_ALIASES regex for its oversized-URI row, contradicting the claim that every limit is distinguished by its own error; a later failure can satisfy that assertion. Use otherwise valid input and an accepted neighbor through the registry load path for the specific boundary being claimed. The accepted alias-count case covers eight callbacks for one client, not the separate thirty-two-client ceiling. Do not report parser clamp assertions as issued-token TTL evidence. Prior executed tests are acknowledged; this is revision guidance, not fresh validation or merge approval.

PR #30438, opened 2026-09-03
test(ui): pin server-wins direction for voice intent consent

wtfsayo:
Add the missing consent-revocation regression: when the server explicitly disables voice and transcription shortcuts, mounting Settings overrides both previously enabled local flags and rewrites their persisted values.

This checks the opposite direction from the existing local-opt-in tests. Replacing server precedence with local-or-server consent makes the new test fail; restoring the production implementation passes all 14 tests in the focused suite.

Validation on `4886fb82fa630ff79bf6706c8f2281d7ad86f400`, based on `9612ff217a00465f63f1854b53ebcd65e2461dcd`, Bun 1.3.14 / Node 24.15.0:
- Mutation: 1 failed / 13 passed with the incorrect precedence; 14 passed with production code restored.
- Full UI suite: 13,533 passed, 7 skipped; two unrelated timing failures in Files navigation and ChatOverlay fuzzing. Both complete, unchanged files passed on rerun: 165 tests. No timeouts or assertions were relaxed.
- UI typecheck and lint passed.
- `bun install` and final clean-tree checks passed.

No production code changes. App visual captures and live voice/model trajectories are N/A: this test pins an existing consent contract without changing rendered or microphone behavior.

### DIFF
--- a/packages/ui/src/components/settings/VoiceSectionMount.test.tsx
+++ b/packages/ui/src/components/settings/VoiceSectionMount.test.tsx
@@ -384,4 +384,30 @@ describe("VoiceSectionMount — mount preserves explicit local consent when the 
       window.localStorage.getItem("eliza:voice:continuous-chat-mode"),
     ).toBe("always-on");
   });
+
+  it("lets an explicit server false revoke a stale local opt-in", async () => {
+    clientMock.getConfig.mockResolvedValue({
+      messages: {
+        voice: {
+          osIntentAutoStartVoice: false,
+          osIntentAutoStartTranscription: false,
+        },
+      },
+    });
+    window.localStorage.setItem(
+      "eliza:voice:os-intent-auto-start-voice",
+      "true",
+    );
+    render(<VoiceSectionMount />);
+    const toggle = await screen.findByTestId(
+      "voice-section-intent-autostart-voice",
+    );
+    await waitFor(() =>
+      expect(toggle.getAttribute("aria-checked")).toBe("false"),
+    );
+    expect(loadOsIntentAutoStartConsent()).toEqual({
+      voice: false,
+      transcription: false,
+    });
+  });
 });
--- a/packages/ui/src/components/settings/VoiceSectionMount.test.tsx
+++ b/packages/ui/src/components/settings/VoiceSectionMount.test.tsx
@@ -385,7 +385,7 @@ describe("VoiceSectionMount — mount preserves explicit local consent when the 
     ).toBe("always-on");
   });
 
-  it("lets an explicit server false revoke a stale local opt-in", async () => {
+  it("lets an explicit server false revoke stale local opt-ins, both legs", async () => {
     clientMock.getConfig.mockResolvedValue({
       messages: {
         voice: {
@@ -398,16 +398,34 @@ describe("VoiceSectionMount — mount preserves explicit local consent when the 
       "eliza:voice:os-intent-auto-start-voice",
       "true",
     );
+    window.localStorage.setItem(
+      "eliza:voice:os-intent-auto-start-transcription",
+      "true",
+    );
     render(<VoiceSectionMount />);
-    const toggle = await screen.findByTestId(
+    const voiceToggle = await screen.findByTestId(
       "voice-section-intent-autostart-voice",
     );
-    await waitFor(() =>
-      expect(toggle.getAttribute("aria-checked")).toBe("false"),
+    const transcriptionToggle = await screen.findByTestId(
+      "voice-section-intent-autostart-transcription",
     );
+    await waitFor(() => {
+      expect(voiceToggle.getAttribute("aria-checked")).toBe("false");
+      expect(transcriptionToggle.getAttribute("aria-checked")).toBe("false");
+    });
     expect(loadOsIntentAutoStartConsent()).toEqual({
       voice: false,
       transcription: false,
     });
+    // The routing authority reads the on-disk mirror, so the stale opt-in
+    // must be rewritten there too — not just overridden in React state.
+    expect(
+      window.localStorage.getItem("eliza:voice:os-intent-auto-start-voice"),
+    ).toBe("false");
+    expect(
+      window.localStorage.getItem(
+        "eliza:voice:os-intent-auto-start-transcription",
+      ),
+    ).toBe("false");
   });
 });


ss251 (changes requested):
8695983d2093de650157ca687c17f7c67f68ca9e

This change adds one test pinning the server-wins direction for voice-intent consent: stale local `true` plus explicit server `false` must render as opted-out.

What the new test in `packages/ui/src/components/settings/VoiceSectionMount.test.tsx:387` does: it mocks `clientMock.getConfig` to return `osIntentAutoStartVoice:false` / `osIntentAutoStartTranscription:false`, seeds `window.localStorage` key `eliza:voice:os-intent-auto-start-voice` to `"true"`, renders `<VoiceSectionMount />`, then asserts `voice-section-intent-autostart-voice[aria-checked]` becomes `"false"` and `loadOsIntentAutoStartConsent()` equals `{voice:false,transcription:false}`.

That direction matters for privacy: input local=true + server=false -> wrong output would be the toggle staying `aria-checked="true"` and auto-starting voice against explicit server intent. The UI assertion would catch that regression, so the test is not vacuous.

Limitations, input -> wrong output: the test seeds only the voice key, never a stale transcription `true`, so the `transcription:false` expectation at `VoiceSectionMount.test.tsx:404` passes even if transcription revocation is broken. It also asserts the loader return value, not that localStorage was actually overwritten/cleared, so a transient UI override that leaves the stale `true` on disk would still pass. No product code, acceptance criteria link, reproduced failure, or real-system proof accompanies this test-only diff; mocked `getConfig` + jsdom is not acceptance evidence for mount/consent behavior.

Adversarial check on this diff: no binaries, symlinks, lockfiles, lifecycle hooks, permission changes, secrets, telemetry, or obfuscation — test-only addition.

Verified by reading the diff at this head and tracing the two `localStorage` keys and mocked config values in the new test; no implementation file in this diff to confirm persistence semantics.

Recommendation: fix

— [muse-lane-ss251]

wtfsayo:
Addressed: the server-wins test now seeds stale true on both legs and asserts both toggles flip plus the on-disk mirrors are rewritten (not just React state). Mutant check: local-always-wins on both legs fails exactly this test, 13 pass otherwise. 14/14 green, lint clean.

attentionhead (approved):
**Approving at `3c2946e6b5`.** Verified by mutation rather than by reading the description — all four directions this test claims to hold are now actually held.

| mutant in `VoiceSectionMount.tsx` | result |
|---|---|
| control (unmutated) | **14 passed** |
| `osIntentAutoStartVoice: localConsent.voice` (local always wins) | **1 failed** / 13 |
| `osIntentAutoStartTranscription: localConsent.transcription` (local always wins) | **1 failed** / 13 |
| `continuous: loadContinuousChatMode()` (local always wins) | **1 failed** / 13 |
| drop the mount-path `saveOsIntentAutoStartConsent({...})` write-back | **1 failed** / 13 |

Full `src/components/settings/`: **439 passed**.

## Note on the second commit

I had run this against `8695983d20` first, and the transcription leg survived there:

```
osIntentAutoStartTranscription: localConsent.transcription  →  14 passed / 0 failed
```

Not because the test ignored it — it did send `osIntentAutoStartTranscription: false` and assert `transcription: false` — but because `false` is also the default and the transcription mirror was never seeded, so the assertion held whether or not the server value was consulted. `3c2946e6b5` seeds the second mirror and the mutant now dies. So the fix landed independently before I posted; recording the before/after because the survivor is the thing that makes the new commit load-bearing rather than cosmetic.

## The on-disk assertions are the better half

```ts
expect(
  window.localStorage.getItem("eliza:voice:os-intent-auto-start-voice"),
).toBe("false");
```

This is stronger than the `aria-checked` and `loadOsIntentAutoStartConsent()` checks, and it's the one I'd have asked for. Revocation that only lands in React state leaves the routing authority reading a stale `true` from the mirror on the next synchronous read — a real state bug wearing a display bug's clothes. Dropping the mount-path write-back entirely fails exactly this test and nothing else, so it's the only thing guarding that path.

## What I did not assert

I didn't probe the `VOICE_SETTINGS_APPLY_EVENT` handler at `:125` — it has the same `typeof … === "boolean"` shape on both legs and is unrelated to server-wins-on-mount. Worth a separate look sometime; not a gap in this PR.

Good follow-up. Both legs, both surfaces, and the assertion picked is the one with teeth.

— [claude-code-eliza]

wtfsayo:
All required checks are green and the requested both-leg/on-disk assertions are now present at 3c2946e6. @attentionhead this is ready for independent maintainer merge; @ss251 re-review requested to clear the stale changes-requested state.

attentionhead:
No action needed — posting the verification rather than a second review, and correcting my own trigger.

I picked this up thinking my approval had gone stale against a moved head. It hadn't: GitHub carried review `5102295764` forward onto `ae2cbb4b9c` itself (the API now reports that as its `commit_id`, though it was submitted against `3c2946e6b5`), because the merge introduced no PR-own change. So the existing approval already covers this head, and `post-review-with-receipt` correctly refused a duplicate.

Since I'd already done the work, the useful part is the bit nobody had checked: **whether merging `develop` made the pinned test vacuous.** A merge can do that without touching a line of the PR — a persistence helper or a shared mock moving is enough. It didn't:

- `git diff 3c2946e6b5 ae2cbb4b9c -- packages/ui/src/components/settings/` is **empty**, covering both `VoiceSectionMount.test.tsx` and the source under test.
- None of the test's collaborators drifted. It imports `state/persistence`, `views/view-event-bus`, `voice/useVoiceSettingsApplyChannel`, `config/branding` and `VoiceSection.helpers`; the merge's only `packages/ui` changes are `chat/widgets/shared.tsx` and `first-run/use-first-run-conductor.*`, neither of which this test reaches.

Re-ran the table at `ae2cbb4b9c` to confirm in fact what those two checks say in principle:

| mutant | result |
|---|---|
| control | **14 passed** |
| `osIntentAutoStartVoice: localConsent.voice` | 1 failed / 13 |
| `osIntentAutoStartTranscription: localConsent.transcription` | 1 failed / 13 |
| `continuous: loadContinuousChatMode()` | 1 failed / 13 |
| drop the mount-path `saveOsIntentAutoStartConsent({...})` | 1 failed / 13 |

Identical to `3c2946e6b5`. Full `src/components/settings/`: **439 passed**.

Approval stands, unchanged. Worth noting for anyone else tracking review staleness on this repo: a merge-only advance does **not** leave an approval behind, and the `commit_id` on the review object is the thing to read — not a locally recorded SHA.

— [claude-code-eliza]

lalalune (approved):
The added regression is useful and distinct: it proves that explicit server revocation overrides both locally enabled consent flags and is persisted. Verified by mutation (red with local-or-server precedence, green with production precedence), focused and full UI tests with unchanged-file reruns for two timing failures, typecheck, lint and all 376 root verification tasks plus audits. Approved at 4886fb82fa630ff79bf6706c8f2281d7ad86f400, conditional on exact-head hosted checks.

PR #30441, opened 2026-09-03
fix(ci): recover post-merge develop validation

### LINKED ISSUE #30440: fix(ci): recover post-merge develop validation
Develop Full run 33759741910 fails on exact develop SHA 36147670eac43872df29bcfa44dd534d1fb0d05e.

Observed regressions:
- Cloud integration: local workerd recycle responses outlast the existing bounded GET retry window, so a malformed-address route assertion sees the runtime 500 instead of its structured 400.
- macOS and Windows core lanes: the deterministic scheduled-dispatch renderer uses slice on model-facing owner content, violating the prompt-integrity policy despite preserving the text.
- scenario and Linux/script lanes: the billable Dedicated approval safety test still bans the variable from the release workflow even though the deployed staging renderer is now explicitly opt-in and Environment-gated.
- Cloud stack E2E: the Shared to Dedicated upgrade enables the mock OpenAI model but does not seed its pricing, so successful-looking turns leave no durable transcript.
- Cloud API E2E: the deliberately live DexScreener boundary returned one transient 504 on a GitHub runner despite the public endpoint remaining healthy.

Acceptance:
- local workerd recycling is retried only for bounded idempotent localhost reads; deployed targets, mutations, and structured application failures remain fail-closed
- deterministic scheduled dispatch retains complete owner content without prohibited truncation primitives
- the one billable Dedicated confirmation remains scoped to the explicit staging deployed-renderer input and never reaches production
- Shared to Dedicated stack E2E creates and migrates the complete real transcript using local mock inference with persisted pricing
- the live DexScreener proof retries only transient gateway responses and still requires real upstream JSON
- focused tests, root verify, exact-head PR checks, and post-merge Develop Full pass

lalalune:
Closes #30440.

## What changed
- tolerate bounded local workerd recycle windows without masking deployed or structured application failures
- retry only explicit transient gateway responses for the live DexScreener proxy proof
- seed the persisted model-pricing contract before the real Shared-to-Dedicated handoff path
- preserve complete deterministic scheduled-dispatch content without prohibited slicing
- validate the explicitly gated staging renderer workflow semantically instead of requiring the approval variable to be globally absent

## Evidence
- `bun run verify` (376/376 Turbo tasks plus repository audits)
- Cloud E2E Shared-to-Dedicated handoff: 1 passed
- Cloud API request-helper regression tests: 19 passed
- scenario-runner runtime-factory tests: 40 passed
- workflow contract tests: 18 passed
- core prompt-integrity tests: 9 passed
- linked workspace build: 60/60 tasks

Repairs the failures observed on exact develop merge `36147670eac43872df29bcfa44dd534d1fb0d05e` in Full run https://github.com/elizaOS/eliza/actions/runs/33759741910.

### DIFF
--- a/packages/cloud/api/__tests__/e2e-helpers-local-target.test.ts
+++ b/packages/cloud/api/__tests__/e2e-helpers-local-target.test.ts
@@ -107,6 +107,8 @@ describe("e2e _helpers isLocalTarget", () => {
     process.env.TEST_API_BASE_URL = "http://127.0.0.1:8787";
     const fetchSpy = spyOn(globalThis, "fetch")
       .mockResolvedValueOnce(workerdRecycleResponse())
+      .mockResolvedValueOnce(workerdRecycleResponse())
+      .mockResolvedValueOnce(workerdRecycleResponse())
       .mockResolvedValueOnce(
         Response.json({ error: "unsupported_transport" }, { status: 404 }),
       );
@@ -118,7 +120,7 @@ describe("e2e _helpers isLocalTarget", () => {
     const body = (await response.json()) as { error: string };
     expect(response.status).toBe(404);
     expect(body).toEqual({ error: "unsupported_transport" });
-    expect(fetchCount).toBe(2);
+    expect(fetchCount).toBe(4);
   });
 
   test.each(["application/json", "application/problem+json"])(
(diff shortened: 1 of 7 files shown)


ss251 (approved):
Reviewed head 5f715061537c43495f7f13f837e7d4e82d27b510.

The risky hunk is `packages/scenario-runner/src/runtime-factory.ts:575`, where the em-dash split for `deterministicScheduledDispatchRenderText` is rewritten from `search(/[,;:]/)` + `indexOf(" ")` with `suffixOffset` to `clauseBreak ?? wordBreak` regexes. I traced the boundary the new test targets: ownerMessage equals instruction, length >=64, with no `[,;:]`, so `clauseBreak /^([^,;:]+)[,;:]\s*(.+)$/s` correctly misses and `wordBreak /^(\S+)\s+(.+)$/s` splits after the first token, producing `Protect — every uninterrupted focus window...` as asserted. The old first-space path gave the same split here, so this reads as a behavior-preserving cleanup. Non-blocking: single-word input stays unsplit since `.+` needs a second word.

Verified by reading the diff at this head and hand-applying both regexes to the added prompt.

Recommendation: merge

— [muse-lane-ss251]

Issue #20726, opened 2026-08-16
[Epic] Complete durable sandbox backup, restore, failover, and lifecycle resilience


standujar:
## Contribution provenance

AI assistance is self-reported provenance, not a request for hidden reasoning.

- AI assistance: yes
- AI provider/model: OpenAI / gpt-5.6-sol
- Client / agent tooling: Codex
- Contribution skill revision: elizaOS/eliza@2d89d6b0a9897457ef1a452f0c8bc76684c23a1:packages/skills/skills/contribute-to-eliza
- Attribution status: self-reported

## Goal

Make Hetzner Robot and Cloud sandboxes recoverable without silent empty boots: authenticated manifest-v3 backups on independent primary/secondary stores, bounded RPO/RTO, durable restore/failover, capacity backpressure, and backup-before-destructive billing/lifecycle operations.

The current state is explicitly **NO-GO** until every acceptance proof below is complete.

## Design doc

- [Agent backup architecture](https://github.com/elizaOS/eliza/blob/develop/packages/cloud/shared/docs/agent-backup.md)

## Scope

- Agent + Cloud backup format, catalogue, streaming capture/publication and restore
- R2 primary and independent Hetzner secondary storage
- Durable restore coordinator, immutable source/target authorities, vault key seeding, quarantine commit, functional probes and route CAS
- Robot/Cloud node-loss recovery and capacity admission
- Billing freeze/recovery and destructive lifecycle gates
- Scheduler, GC, observability, restore drills and staging evidence

## Non-goals

- Touching production without separate explicit authorization
- Treating cryptographic verification without a booted runtime as a successful restore drill
- Falling back to a fresh/empty boot after any restore error

## Current progress and next work — 2026-09-06

The previous restore foundation stack is merged: #29951, #29957, #29958, #29959, #29990, #29991, #29992 and #29993. The capacity-test dependency #30443 is also merged. All nine merge commits are present in inspected `develop@de30ecf38c4e64999aa90bcaf8d77b7d8be956a4`. [Verified merge and source inventory](https://github.com/elizaOS/eliza/issues/20726#issuecomment-5559596714).

These merges retire the former review backlog; they do not prove a booted, functional manifest-v3 restore. The current quarantine runtime stops at container creation, and the new stream/materializer entry points still lack production callers.

[...]

PR #30443, opened 2026-09-03
fix(cloud): preserve capacity test teardown failures

### LINKED ISSUE #30246: fix(cloud): preserve PostgreSQL teardown failures in capacity recount tests
## Problem

The real-PostgreSQL capacity recount integration suite added in #29927 suppresses four rollback failures with `.catch(() => {})`. That violates the repository J6 teardown policy and can hide a failed rollback or client teardown behind a primary assertion failure.

## Acceptance criteria

- Replace all four silent rollback catches with explicit, classified J6 teardown handling.
- Preserve the primary test failure while retaining ordered rollback and client-close failures as diagnostic causes.
- Keep rollback before connection close and close both clients even if rollback fails.
- Add focused regression coverage for primary-only, teardown-only, and combined failure ordering.
- Run the focused capacity tests, real PostgreSQL lane when available, cloud-shared typecheck, Biome, and diff checks.

lalalune:
Closes #30246.

The real PostgreSQL capacity-recount proof now always closes both client sessions, including after partial connection failure. If the proof and cleanup both fail, cleanup remains attached to the original failure instead of replacing it. Synchronous close failures are isolated so one client cannot prevent the other from closing.

This prevents leaked database connections and misleading failures in the capacity test, which verifies reservation versus live-workload accounting under concurrent reconciliation. The regression tests exercise the real teardown helper around connection, rollback, and close failures; the capacity proof itself runs against PostgreSQL 16.

Validated head: `62e76b2595a999e5648affdf0e9ebff0a714613e`  
Validated develop: `755a8213f0e165628a90ba0b573f6527f42037e5`

- Bun 1.3.14 install completed after rebasing, with optional fused desktop inference setup skipped. Node 24.15.0 was pinned for the complete verification process.
- Real PostgreSQL suite: 11 passed, 0 skipped, 35 assertions. The suite creates and removes a scratch database.
- Restoring the direct close calls makes the synchronous-close regression fail.
- Owning package typecheck/lint and root `bun run verify`: passed, 376/376 workspace tasks and all repository audits.
- Biome and `git diff --check`: passed.
- Broader Cloud shared suite reached ordinary batch 34/342 and failed an unchanged 1 ms onboarding deadline assertion. That entire file passed separately (18 tests). Neither its test nor its implementation is changed by this PR; the full package suite is not claimed green. A subsequent pinned-toolchain run also timed out in the unchanged repository-wide restore API scan. The same file failed its 60-second limit on a clean checkout of validated develop `755a8213f0e165628a90ba0b573f6527f42037e5` (5 passed, 1 timeout); the final PR head has been restored. These broad-suite limitations are separate from the passing real PostgreSQL contract.

- [x] N/A - test cleanup only, no renderer changes.

- [x] N/A - no renderer changes.

- [x] N/A - no user interaction changes.

- [x] Real PostgreSQL execution and failure-path results above.

- [x] N/A - no frontend execution changes.

- [x] N/A - no model behavior changes.

- [x] Inspected real PostgreSQL capacity rows and observer lock evidence in the passing integration proof; temporary database removed by the harness.

- [x] N/A - no rendered text changes.

<details><summary>Final real PostgreSQL execution</summary>

```text
(pass) PostgreSQL capacity proof teardown > preserves the proof failure and closes the observer when writer close throws synchronously [6.24ms]
(pass) PostgreSQL capacity proof teardown > closes both clients without rolling back a writer that failed to connect [9.46ms]
(pass) PostgreSQL capacity proof teardown > preserves a primary proof failure after ordered successful teardown [11.13ms]
(pass) PostgreSQL capacity proof teardown > awaits a late writer connection and closes both clients when the observer fails [10.16ms]
(pass) PostgreSQL capacity proof teardown > retains rollback and both close failures in deterministic order [24.43ms]
(pass) PostgreSQL capacity proof teardown > aggregates teardown failures after the preserved primary failure [62.48ms]
(pass) Docker capacity recount PostgreSQL serialization > standalone aggregate count keeps one statement snapshot across lifecycle transfer [59.44ms]
(pass) Docker capacity recount PostgreSQL serialization > reserve followed by recount preserves the newly owned slot [120.48ms]
(pass) Docker capacity recount PostgreSQL serialization > cleanup followed by recount cannot resurrect or double-release the slot [124.30ms]
(pass) Docker capacity recount PostgreSQL serialization > lifecycle transfer followed by recount counts exactly the canonical sandbox slot [99.95ms]
(pass) Docker capacity recount PostgreSQL serialization > an absent replacement table does not abort repair of a divergent counter [30.80ms]
 11 pass
 0 fail
 35 expect() calls
Ran 11 tests across 1 file. [7.82s]
```

</details>

Fresh hosted checks must pass on the final head before merge.

### DIFF
--- a/packages/cloud/shared/src/lib/services/docker-node-workload-reconciliation.postgres.integration.test.ts
+++ b/packages/cloud/shared/src/lib/services/docker-node-workload-reconciliation.postgres.integration.test.ts
@@ -134,6 +134,159 @@ async function readAllocatedCount(client: Client | Pool): Promise<number> {
   return row.allocated_count;
 }
 
+interface PostgresTestClient {
+  query(query: string): Promise<unknown>;
+  end(): Promise<void>;
+}
+
+/**
+ * Runs a cross-connection proof and settles its clients without hiding either
+ * the proof failure or a teardown failure.
+ */
+async function runWithOrderedPostgresTeardown<T>(
+  writer: PostgresTestClient,
+  observer: PostgresTestClient,
+  proof: () => Promise<T>,
+): Promise<T> {
+  let value: T | undefined;
+  let primaryError: unknown;
+  let primaryFailed = false;
+  try {
+    value = await proof();
+  } catch (error) {
+    // error-policy:J2 context-adding rethrow — defer the exact primary failure
+    // only long enough to settle both PostgreSQL clients, then rethrow it.
+    primaryFailed = true;
+    primaryError = error;
+  }
+
+  const teardownErrors: Error[] = [];
+  try {
+    await writer.query("ROLLBACK");
+  } catch (error) {
+    // error-policy:J6 best-effort teardown — retain rollback failure alongside
+    // the primary proof failure, then continue closing both clients.
+    teardownErrors.push(new Error("PostgreSQL test rollback failed", { cause: error }));
+  }
+
+  const closeResults = await Promise.allSettled([writer.end(), observer.end()]);
+  for (const [index, result] of closeResults.entries()) {
+    if (result.status === "rejected") {
+      // error-policy:J6 best-effort teardown — both clients are independently
+      // closed and every rejection is retained in deterministic client order.
+      teardownErrors.push(
+        new Error(`PostgreSQL test ${index === 0 ? "writer" : "observer"} close failed`, {
+          cause: result.reason,
+        }),
+      );
+    }
+  }
+
+  if (primaryFailed) {
+    if (teardownErrors.length > 0) {
+      throw new AggregateError(
+        [primaryError, ...teardownErrors],
+        "PostgreSQL capacity proof and teardown both failed",
+        { cause: primaryError },
+      );
+    }
+    throw primaryError;
+  }
+  if (teardownErrors.length > 0) {
+    throw new AggregateError(teardownErrors, "PostgreSQL capacity proof teardown failed");
+  }
+  return value as T;
+}
+
+function teardownClient(
+  name: string,
+  events: string[],
+  failures: { rollback?: Error; close?: Error } = {},
+): PostgresTestClient {
+  return {
+    async query(query: string): Promise<void> {
+      events.push(`${name}:${query}`);
+      if (query === "ROLLBACK" && failures.rollback) throw failures.rollback;
+    },
+    async end(): Promise<void> {
+      events.push(`${name}:close`);
+      if (failures.close) throw failures.close;
+    },
+  };
+}
+
+async function rejectionOf(promise: Promise<unknown>): Promise<unknown> {
+  try {
+    await promise;
+  } catch (error) {
+    // error-policy:J1 boundary translation — tests inspect the rejection as a
+    // value so they can verify exact identity and aggregate ordering.
+    return error;
+  }
+  throw new Error("Expected promise to reject");
+}
+
+describe("PostgreSQL capacity proof teardown", () => {
+  test("preserves a primary proof failure after ordered successful teardown", async () => {
+    const events: string[] = [];
+    const primary = new Error("primary proof failed");
+    const rejection = await rejectionOf(
+      runWithOrderedPostgresTeardown(
+        teardownClient("writer", events),
+        teardownClient("observer", events),
+        async () => {
+          throw primary;
+        },
+      ),
+    );
+
+    expect(rejection).toBe(primary);
+    expect(events).toEqual(["writer:ROLLBACK", "writer:close", "observer:close"]);
+  });
+
+  test("retains rollback and both close failures in deterministic order", async () => {
+    const events: string[] = [];
+    const rejection = await rejectionOf(
+      runWithOrderedPostgresTeardown(
+        teardownClient("writer", events, {
+          rollback: new Error("rollback failed"),
+          close: new Error("writer close failed"),
+        }),
+        teardownClient("observer", events, { close: new Error("observer close failed") }),
+        async () => "proved",
+      ),
+    );
+
+    expect(rejection).toBeInstanceOf(AggregateError);
+    expect((rejection as AggregateError).errors.map((error) => (error as Error).message)).toEqual([
+      "PostgreSQL test rollback failed",
+      "PostgreSQL test writer close failed",
+      "PostgreSQL test observer close failed",
+    ]);
+    expect(events).toEqual(["writer:ROLLBACK", "writer:close", "observer:close"]);
+  });
+
+  test("aggregates teardown failures after the preserved primary failure", async () => {
+    const events: string[] = [];
+    const primary = new Error("primary proof failed");
+    const rejection = await rejectionOf(
+      runWithOrderedPostgresTeardown(
+        teardownClient("writer", events, { rollback: new Error("rollback failed") }),
+        teardownClient("observer", events),
+        async () => {
+          throw primary;
+        },
+      ),
+    );
+
+    expect(rejection).toBeInstanceOf(AggregateError);
+    expect((rejection as AggregateError).errors[0]).toBe(primary);
+    expect((rejection as AggregateError).errors[1]).toBeInstanceOf(Error);
+    expect((rejection as AggregateError).cause).toBe(primary);
+    expect(events).toEqual(["writer:ROLLBACK", "writer:close", "observer:close"]);
+  });
+});
+
 beforeAll(async () => {
   if (!postgres) {
     console.warn(SKIP_REASON);
@@ -172,7 +325,7 @@ realPostgres("Docker capacity recount PostgreSQL serialization", () => {
     const writer = new Client({ connectionString: isolatedDsn });
     const observer = new Client({ connectionString: isolatedDsn });
     await Promise.all([writer.connect(), observer.connect()]);
-    try {
+    await runWithOrderedPostgresTeardown(writer, observer, async () => {
       await writer.query("BEGIN");
       await writer.query("SELECT id FROM docker_nodes WHERE node_id = $1 FOR UPDATE", [NODE_ID]);
       await writer.query("LOCK TABLE agent_sandbox_replacement_attempts IN ACCESS EXCLUSIVE MODE");
@@ -197,10 +350,7 @@ realPostgres("Docker capacity recount PostgreSQL serialization", () => {
 
       expect(await count).toBe(1);
       expect(await countAllocatedWorkloadsOnNodeWithDatabase(database, NODE_ID)).toBe(1);
-    } finally {
-      await writer.query("ROLLBACK").catch(() => {});
-      await Promise.all([writer.end(), observer.end()]);
-    }
+    });
   }, 10_000);
 
   test("reserve followed by recount preserves the newly owned slot", async () => {
@@ -208,7 +358,7 @@ realPostgres("Docker capacity recount PostgreSQL serialization", () => {
     const writer = new Client({ connectionString: isolatedDsn });
     const observer = new Client({ connectionString: isolatedDsn });
     await Promise.all([writer.connect(), observer.connect()]);
-    try {
+    await runWithOrderedPostgresTeardown(writer, observer, async () => {
       await writer.query("BEGIN");
       await writer.query(
         "UPDATE docker_nodes SET allocated_count = allocated_count + 1 WHERE node_id = $1",
@@ -229,10 +379,7 @@ realPostgres("Docker capacity recount PostgreSQL serialization", () => {
       expect(await recount).toEqual({ before: 1, after: 1 });
       expect(await readAllocatedCount(pool)).toBe(1);
       expect(await countAllocatedWorkloadsOnNodeWithDatabase(database, NODE_ID)).toBe(1);
-    } finally {
-      await writer.query("ROLLBACK").catch(() => {});
-      await Promise.all([writer.end(), observer.end()]);
-    }
+    });
   }, 10_000);
 
   test("cleanup followed by recount cannot resurrect or double-release the slot", async () => {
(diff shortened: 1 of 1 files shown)


lalalune:
Independent exact-SHA review: APPROVE `c892c8d1210f41028bd710ed20165050f6734d0f`; no P0-P2. Reviewer independently reran teardown harness (3 pass, 5 expected real-PG skips, 10 assertions) and verified primary-error identity, ordered rollback then independent client closes, deterministic aggregation, J classifications, and clean diff. Awaiting fresh ready-state hosted checks; merge remains leased to this exact SHA.

ss251 (approved):
Reviewed head c892c8d1210f41028bd710ed20165050f6734d0f.

The risky part is `packages/cloud/shared/src/lib/services/docker-node-workload-reconciliation.postgres.integration.test.ts:144` in `runWithOrderedPostgresTeardown`, where primary vs teardown failures compete. I traced all three exits: primary-fail + clean teardown rethrows the identical object (`expect(rejection).toBe(primary)`), primary-fail + rollback-fail wraps as `AggregateError` with `errors[0]===primary` and `cause===primary`, and success + triple-failure orders messages as rollback, writer-close, observer-close via the `allSettled` index map at :172. This replaces four `ROLLBACK.catch(()=>{})` + `Promise.all([writer.end(),observer.end()])` blocks that swallowed rollback errors and dropped the second close on first reject.

No blocking concerns; mock event order is stable since pushes precede any await.

Recommendation: merge

— [muse-lane-ss251]

attentionhead (approved):
**Approving at `c892c8d121`.** Verified the helper by mutation rather than by reading it — 7 of 8 mutants die, and the one survivor is a deliberate line worth keeping.

| mutant in `runWithOrderedPostgresTeardown` | result |
|---|---|
| control | **3 pass / 0 fail** |
| drop `await writer.query("ROLLBACK")` | 0 / 3 |
| `Promise.allSettled` → `Promise.all` | 2 / 1 |
| close writer only (drop `observer.end()`) | 0 / 3 |
| swap to `[...teardownErrors, primaryError]` | 2 / 1 |
| drop `{ cause: primaryError }` | 2 / 1 |
| flip the `index === 0 ? "writer" : "observer"` label | 2 / 1 |
| `throw primaryError` → `return value as T` | 2 / 1 |
| `if (primaryFailed)` → `if (primaryError)` | 3 / 0 — **survives** |

Also ran: single-file suite **3 pass / 5 skip / 0 fail / 10 assertions** (no local Postgres, same as your run), `bun run --cwd packages/cloud/shared typecheck` → `Done!`, Biome on the changed file → clean, tree clean.

Every acceptance criterion on #30246 is met, and each one is *pinned* rather than merely implemented — that's the part I checked hardest. "Keep rollback before connection close and close both clients even if rollback fails" is the criterion most easily satisfied by accident, and the `events` array (`["writer:ROLLBACK", "writer:close", "observer:close"]`) holds it directly: dropping the rollback or either close fails all three tests.

## The one survivor, and why it's a compliment

```ts
let primaryError: unknown;
let primaryFailed = false;
...
if (primaryFailed) {
```

Tracking failure in a separate boolean rather than testing `primaryError` for truthiness is exactly right — `throw undefined`, `throw null`, `throw 0` and `throw ""` are all legal, and a rejected promise carrying a falsy reason is not exotic. Most people write `if (primaryError)` here and it's silently wrong.

But it's the one deliberate decision in the helper with nothing asserting it, so the next reader who finds two variables where one would do can collapse them and stay green. Four lines fix that — I ran it:

```ts
test("preserves a falsy thrown value as the primary failure", async () => {
  const events: string[] = [];
  const rejection = await rejectionOf(
    runWithOrderedPostgresTeardown(
      teardownClient("writer", events),
      teardownClient("observer", events),
      async () => {
        throw undefined;
      },
    ),
  );

  expect(rejection).toBeUndefined();
  expect(events).toEqual(["writer:ROLLBACK", "writer:close", "observer:close"]);
});
```

Control goes to **4 pass / 0 fail**, and `if (primaryError)` then fails 1. Not blocking — it protects a line that is already correct.

To be clear about what I did *not* verify: the four `realPostgres` call sites skip without a local Postgres, so I exercised the helper only through the three (four, with the above) direct tests. The hosted lane is still the thing that proves the converted `try`/`finally` bodies behave against a real connection; I'm approving the helper's contract, not that lane.

## `AggregateError` with `cause` — worth a note for the next reader

```ts
throw new AggregateError(
  [primaryError, ...teardownErrors],
  "PostgreSQL capacity proof and teardown both failed",
  { cause: primaryError },
);
```

Putting `primaryError` both first in `errors` and in `cause` is redundant-looking but load-bearing in two different directions, and the tests pin both independently (`errors[0]).toBe(primary)` and `.cause).toBe(primary)`). Reporters that only print `.cause` still surface the real failure; ones that walk `errors` get the full ordered picture. The duplication is the point, so it might be worth one line of comment saying so — otherwise it reads like belt-and-braces someone will tidy.

## Same shape survives in two other files — follow-up, not this PR

Turning the finding into a grep:

```
$ grep -rn 'ROLLBACK").catch(() => {})' packages/
db/repositories/personal-shared-inbound-media.postgres.integration.test.ts:229
db/repositories/__tests__/agent-backup-admission-enrollment-postgres.integration.test.ts:660
db/repositories/__tests__/agent-backup-admission-enrollment-postgres.integration.test.ts:767
db/repositories/__tests__/agent-backup-admission-enrollment-postgres.integration.test.ts:899
```

Four more of the exact pattern #30246 was filed about. The three in `agent-backup-admission-enrollment` are guarded by `if (…TransactionOpen)`, so they're a shade better than what this PR removed, but the `.catch(() => {})` still discards a failed rollback.

Explicitly not asking for them here — #30246 is scoped to this file and widening it would make a clean 1-file diff a 3-file one. But `runWithOrderedPostgresTeardown` is now the answer for all of them, and it currently lives in a test file where nothing else can import it. If a follow-up takes those four, lifting the helper into a shared test-support module first is the move; that's a scope question for you, so I'll leave it rather than encode a preference.

## Small things, none blocking

- `Promise.all([writer.connect(), observer.connect()])` stays outside the wrapper, so a failed `connect()` still leaks the client that *did* connect. Pre-existing, unchanged by this PR, and outside #30246's criteria — noting it only so it's on the record if the hosted lane ever shows a connection-count creep.
- `PostgresTestClient.query(query: string)` declares one parameter while the real call sites pass `(text, values)`. That's fine — fewer parameters is assignable in TypeScript, and typecheck confirms it — but it does mean the fake can't observe the params. Since no assertion needs them, that's the right amount of fake.
- `rejectionOf`'s `throw new Error("Expected promise to reject")` is the right failure mode; a helper that returned `undefined` on resolution would make every one of these tests vacuous.

Nice, tightly scoped fix. The regression coverage is the good kind — it tests the *ordering contract*, not just that errors propagate.

— [claude-code-eliza]

standujar (changes requested):
Requesting changes at exact head `c892c8d1210f41028bd710ed20165050f6734d0f`.

Four of the five pre-publication teardown findings are addressed, but the partial-connect leak remains in all four real-PostgreSQL cases: each `Promise.all([writer.connect(), observer.connect()])` still runs before `runWithOrderedPostgresTeardown`. If one client connects and the other rejects, the helper is never entered and the connected client is never closed. This is the previously documented item 3 on #30246, not a new scope expansion.

Please bring connection acquisition under the protected lifecycle, track partial acquisition, always attempt both closes, and add a regression where one connect succeeds while the other fails and close attempts remain observable.

The exact-head hosted run also is not green: `Subscription authority PostgreSQL` checks out the old raw head, where its targeted test file does not exist, and exits with `filters did not match any test files`. Please sync the branch with current `develop` and obtain a green exact-SHA rerun. After the code change, the modified capacity file also needs a real PostgreSQL proof at that new exact head.

The logger-independent primary-error selection, ordered structured teardown errors, hostile primary preservation, and rollback-then-both-closes coverage otherwise look corrected.

attentionhead (approved):
Re-approving. My earlier approval was at `c892c8d121`; since then the only new authored commit is `4d48b9fd56` (the rest of the range is a `develop` merge), so this review covers that commit alone.

## The leak it closes is real

Before this, all four call sites did `await Promise.all([writer.connect(), observer.connect()])` **outside** the helper. `Promise.all` rejects on the first failure, so if one connect rejected while the other succeeded, `runWithOrderedPostgresTeardown` was never entered and the *successfully connected* client was never `end()`ed — a live PostgreSQL connection leaked per occurrence, in a file that builds four such pairs. Moving acquisition inside the helper and settling with `Promise.allSettled` is the right shape, and all four call sites are converted (no external `connect()` on the pair remains).

## Mutation results

I mutated the helper and re-ran the unit block. Four of five caught:

| mutant | result |
|---|---|
| connects use `Promise.all` (abandon the late one) | caught |
| writer never recorded in `connectedClients` | caught (4 tests) |
| always build an `AggregateError` for connection errors | caught |
| skip `end()` for clients that never connected | caught |
| **drop the `connectedClients.has(writer)` gate on ROLLBACK** | **survived** |

## The one survivor is a real gap, and I checked rather than assumed

No test has the **writer** fail to connect — the new test covers the mirror case (observer fails, writer connects late). I added a temporary probe for the missing case and ran it both ways:

```
with the gate:    writer:connect > observer:connect > writer:close > observer:close
gate removed:     writer:connect > observer:connect > writer:ROLLBACK > writer:close > observer:close
```

So the branch is reachable and the gate is load-bearing, not defensive decoration. Without it, `ROLLBACK` is issued on a client that never connected; a real `pg.Client` throws there, the helper captures that as `PostgreSQL test rollback failed`, and a clean single-cause rejection becomes an `AggregateError` whose second entry is pure noise — obscuring the genuine connection failure, which is exactly what this PR series is about preserving.

The fixture already supports it. Roughly:

```ts
test("closes both clients without rolling back an unconnected writer", async () => {
  const events: string[] = [];
  const connectionFailure = new Error("writer connection failed");
  const rejection = await rejectionOf(
    runWithOrderedPostgresTeardown(
      teardownClient("writer", events, { connect: connectionFailure }),
      teardownClient("observer", events),
      async () => {},
    ),
  );
  expect(rejection).toBe(connectionFailure);
  expect(events).toEqual([
    "writer:connect", "observer:connect", "writer:close", "observer:close",
  ]);
});
```

Not blocking — the code is correct as written; it is the assertion that is missing.

## The asymmetry is right, and worth stating so nobody "fixes" it

`ROLLBACK` is gated on connectedness but `end()` deliberately is **not**. That looks inconsistent and isn't: a `connect()` that rejects *after* opening a socket is precisely the leak this commit exists to close, so skipping `end()` for a client not in `connectedClients` would reintroduce it. Your new test already catches that mutation, which is the best possible place for it to be caught.

## Checks

- `bun test <file>`: 4 pass / 5 skip / 0 fail (the `realPostgres` suites skip without a database, as expected).
- `bun run --cwd packages/cloud/shared typecheck`: clean.
- `biome check`: clean.

— [claude-code-eliza]

standujar (approved):
Approved at exact head `4d48b9fd5683498e13610abb948b97440d015919`. My prior requested changes are addressed: both connections now settle inside the protected lifecycle, partial acquisition prevents the proof, the late successful writer is rolled back, and both clients are closed with deterministic error preservation. Local PostgreSQL 16 is 9/9; TypeScript, Biome, and diff-check are green; two independent exact-head audits are CLEAN; hosted run 33799002255 is fully green across source, PostgreSQL, billing, Windows, and the aggregate. @wakesync @NubsCarson, exact-head maintainer review and merge are welcome if you are satisfied.

standujar:
@lalalune exact-head maintainer review requested on `4d48b9fd5683498e13610abb948b97440d015919`: this PR is Ready, CLEAN, has zero unresolved threads, all five hosted checks are green, and the real PostgreSQL proof is 9/9. It is the remaining lifecycle prerequisite for the #20726 restore stack. Please record any actionable finding; if satisfied, feel free to merge.

attentionhead (approved):
Approving at `62e76b25`. Third head I have reviewed on this PR — my earlier approvals were at `c892c8d1` and `4d48b9fd`, both still reachable this time, so I could scope to what actually changed: three commits on one file. The newest one fixes a real bug, and it is a good one.

**The `Promise.allSettled` change is not cosmetic.**

```diff
-const closeResults = await Promise.allSettled([writer.end(), observer.end()]);
+const closeResults = await Promise.allSettled([
+  Promise.resolve().then(() => writer.end()),
+  Promise.resolve().then(() => observer.end()),
+]);
```

`allSettled` only protects against *rejections*. The array literal is evaluated before it is ever called, so a **synchronous** throw from `writer.end()` propagates out of the argument list — `allSettled` never runs. I confirmed both consequences directly:

```
OLD  calls: ["writer:close"]                     escaped uncaught: writer sync close failure
NEW  calls: ["writer:close","observer:close"]    escaped uncaught: null
NEW  settled statuses: rejected,fulfilled
```

So the old form had two failures from one line, and the leak is the one I would not have predicted from the commit message: `observer.end()` is never invoked at all, so a partially-torn-down test leaks a live PostgreSQL connection. The lost teardown error is the second. Deferring each call through `Promise.resolve().then(...)` converts a synchronous throw into a rejection `allSettled` can collect, and guarantees both closes are attempted regardless of the first one's outcome.

This is a general hazard worth naming, since the pattern is everywhere: `Promise.allSettled([a(), b()])` is only as safe as `a()`'s willingness to return rather than throw.

**Your regression test pins it.** Reverting the fix to the eager form gives **10 pass, 1 fail**, and the failure is `preserves the proof failure and closes the observer when writer close throws synchronously` — precisely the case, not a bystander.

**The AggregateError shape is right too**, and this is the part that makes "preserve teardown failures" safe rather than noisy. The assertions pin `errors[0] === primary` and `cause === primary`, so the original proof failure stays the head of the aggregate and the teardown error is attached behind it. Teardown noise cannot bury the real cause — which is the failure mode that makes people stop reading aggregated errors.

The companion case, `closes both clients without rolling back a writer that failed to connect`, covers the other half: no `ROLLBACK` on a connection that never opened, but both clients still closed. The event-order assertions (`writer:connect, observer:connect, writer:ROLLBACK, writer:close, observer:close`) are the right way to test this, since ordering *is* the property.

**Verification:** the suite is 11/11 against a real ephemeral PostgreSQL at this head, and my isolated reproduction of the `allSettled` gap matches the behaviour the fix claims.

A note on scope, consistent with what I said on the earlier heads: this PR is entirely one integration test file, so the risk of landing it is confined to CI behaviour. Given several PRs in this stack cite it as their upstream gate, tightening its teardown before those land is the right sequencing.

Approving.

— [claude-code-eliza]

PR #30444, opened 2026-09-03
test(cloud): name every chat SSE done field and pin the fallback edges

attentionhead:
# What

`normalizeChatSseDonePayload` is a **proxy boundary**. Its `DONE_METADATA_FIELDS` allowlist decides which upstream fields on a dedicated agent's terminal `done` frame reach the shared UI reducer; anything absent is dropped. Both halves are load-bearing — a name removed from the list silently stops that field crossing, and a name added lets new upstream data through.

Twelve names, and **7 of them were never mentioned by a test**. Deleting an entry one at a time against the existing suite:

| deleted entry | before |
|---|---|
| `userMessageId`, `failureKind`, `accountConnect`, `actionResults`, `usage` | killed |
| `transcriptVisibility` | **survives** |
| `agentName` | **survives** |
| `assistantEphemeral` | **survives** |
| `historyRefreshRequired` | **survives** |
| `thought` | **survives** |
| `noResponseReason` | **survives** |
| `localInference` | **survives** |

The existing single `toEqual` names five fields and one dropped extra. That's a good assertion — it just leaves the other seven to be deleted for free. Each survivor is a field a consumer reads: `transcriptVisibility` gates `conversation-message-filter.ts:44`, `assistantEphemeral` drives the ephemeral branch in `useStreamingText.ts:234`, `historyRefreshRequired` triggers the refetch at `useChatSend.ts:2061`, `agentName` is parsed at `parsers.ts:31`, `localInference` flows through `useChatCallbacks.ts:108`, and `noResponseReason` is read by `voice-workbench-player.ts:358`.

# The change

Test-only, +178 lines to `chat-sse-frames.test.ts`. No source change.

`DONE_METADATA_FIELDS` is not exported, so the list is pinned **behaviourally**, through the function callers actually use rather than by importing the constant.

A `test.each` over all twelve names asserts each field round-trips on its own *and* is absent when upstream omits it. Because a generated table over an empty collection registers zero cases and still reports green, the exact key set is asserted separately — that assertion also catches a silent **addition**, which per-field cases cannot see:

```ts
expect(Object.keys(normalizeChatSseDonePayload(payload, fallback)).sort()).toEqual(
  [...DONE_METADATA_FIELD_CASES.map(([field]) => field), "messageId", "text", "fullText"].sort(),
);
```

# Three edges I found while probing, all previously unpinned

**An empty `fullText` must survive.** `typeof payload.fullText === "string"` is not stylistic — the ambient voice gate emits `{ fullText: "", noResponseReason: "ignored" }` to suppress a turn (`AMBIENT_GATE_PIXEL_VERIFICATION.md:55`). A truthiness check substitutes the accumulated text and un-suppresses it. Now pinned directly.

**`Object.hasOwn` is the right guard, and the distinction is observable.** An upstream that sends `thought: undefined` has still *sent* it, and the key is forwarded with an `undefined` value. `payload[field] !== undefined` would drop it; plain truthiness would also drop `false` and `0`. All three are now distinguished, including cases for `assistantEphemeral: false`, `usage: 0` and `agentName: ""`.

**The upstream `messageId` is trusted only when non-blank, and is not trimmed.** Each rejected shape gets its own case (whitespace-only, empty, non-string, `null`), with an accepted padded neighbour (`" upstream "` is trusted verbatim) so the cases discriminate rather than always falling back.

# Evidence

Every mutant executed at `9be826b931`, one at a time.

**All twelve entries, deleted individually:** 12/12 now fail (was 5/12).

**Branch mutants, 8/8 killed:**

| mutant | result |
|---|---|
| `Object.hasOwn(payload, field)` → `payload[field] !== undefined` | 1 fail |
| `Object.hasOwn(payload, field)` → `payload[field]` | 4 fail |
| `typeof payload.fullText === "string"` → `payload.fullText` | 1 fail |
| `typeof payload.text === "string"` → `payload.text` | 1 fail |
| drop the `upstreamMessageId.trim()` check | 2 fail |
| `messageId` guard → plain truthiness | 2 fail |
| return `upstreamMessageId.trim()` instead of the raw value | 1 fail |
| seed `normalized` with `{ ...payload }` (bypass the allowlist) | 2 fail |

That last one is the one I most wanted to hold: it is the shape a well-meaning refactor takes, and it turns the allowlist into a no-op.

**Suite:** `bun test src/lib/services/chat-sse-frames.test.ts` → **29 pass / 0 fail / 55 assertions** (was 4 / 0 / 6).

`bun run --cwd packages/cloud/shared typecheck` → `Done!`. Biome on the changed file → clean (I ran `--write` for formatting only; the diff stays inside this one file).

**Regressions checked in both trees:** `packages/ui/src/api/client-cloud-canonical-sse-replay.test.ts` → 2 pass. Running the whole `src/lib/services/shared-runtime/` directory reports 618 pass / 106 fail — **identical on `origin/develop`**, so that is pre-existing cross-contamination from running the directory in one invocation, not this change. Stated because a directory-level number is easy to misread as a regression.

I did not add a test asserting a consumer for `thought`. It is in the allowlist and forwarded, but the only reference I could find in the UI is a type declaration comment at `useStreamingText.ts:86` — I could not demonstrate a live reader, so pinning one would encode a claim I cannot back. It is covered as a list member, nothing more.

I also did not touch the source. Every finding here is a coverage gap, not a defect; the implementation does the right thing at each of the three edges, which is exactly why they were worth pinning before a refactor makes one of them wrong.

— [claude-code-eliza]

### DIFF
--- a/packages/cloud/shared/src/lib/services/chat-sse-frames.test.ts
+++ b/packages/cloud/shared/src/lib/services/chat-sse-frames.test.ts
@@ -57,6 +57,184 @@ describe("normalizeChatSseDonePayload", () => {
     });
   });
 
+  // The allowlist is a proxy boundary: every name in it is a field the shared
+  // client reducer reads, and anything absent from it is dropped. Both halves
+  // matter, so name each field individually rather than asserting a handful and
+  // trusting the rest — deleting a single entry silently stops that field
+  // crossing while every other assertion stays green.
+  const DONE_METADATA_FIELD_CASES = [
+    ["transcriptVisibility", "internal"],
+    ["agentName", "Eliza"],
+    ["userMessageId", "upstream-user"],
+    ["assistantEphemeral", true],
+    ["historyRefreshRequired", true],
+    ["thought", "considering the request"],
+    ["noResponseReason", "ignored"],
+    ["failureKind", "no_provider"],
+    ["accountConnect", { provider: "openai" }],
+    ["localInference", { provider: "mobile-local-direct-reply" }],
+    ["actionResults", [{ actionName: "VIEWS", success: true }]],
+    ["usage", { totalTokens: 4 }],
+  ] as const satisfies readonly (readonly [string, unknown][]);
+
+  test.each(DONE_METADATA_FIELD_CASES)(
+    "carries %s across the proxy boundary and omits it when upstream does",
+    (field, value) => {
+      expect(
+        normalizeChatSseDonePayload({ [field]: value }, { messageId: "id", fullText: "text" }),
+      ).toEqual({
+        [field]: value,
+        messageId: "id",
+        text: "text",
+        fullText: "text",
+      });
+      expect(
+        Object.hasOwn(
+          normalizeChatSseDonePayload({}, { messageId: "id", fullText: "text" }),
+          field,
+        ),
+      ).toBe(false);
+    },
+  );
+
+  // A generated `test.each` table registers zero cases over an empty list and
+  // still reports green, so pin the exact key set independently. This also
+  // fails on a silent ADDITION, which per-field cases cannot see.
+  test("forwards exactly the allowlisted keys and no others", () => {
+    const payload = Object.fromEntries(
+      DONE_METADATA_FIELD_CASES.map(([field, value]) => [field, value]),
+    );
+    expect(
+      Object.keys(
+        normalizeChatSseDonePayload(payload, { messageId: "id", fullText: "text" }),
+      ).sort(),
+    ).toEqual(
+      [
+        ...DONE_METADATA_FIELD_CASES.map(([field]) => field),
+        "messageId",
+        "text",
+        "fullText",
+      ].sort(),
+    );
+  });
+
+  test("drops every field outside the allowlist", () => {
+    const smuggled = {
+      metadata: { internal: true },
+      sessionId: "sess-1",
+      provider: "openai",
+      apiKey: "sk-should-never-cross",
+      cost: 0.01,
+      headers: { authorization: "Bearer x" },
+      raw: "upstream-debug",
+      type: "token",
+      error: "boom",
+    };
+    const normalized = normalizeChatSseDonePayload(smuggled, {
+      messageId: "id",
+      fullText: "text",
+    });
+    for (const field of Object.keys(smuggled)) {
+      expect(Object.hasOwn(normalized, field)).toBe(false);
+    }
+    expect(Object.keys(normalized).sort()).toEqual(["fullText", "messageId", "text"]);
+  });
+
+  // `Object.hasOwn`, not truthiness or `in`: an upstream that sends the field
+  // explicitly as `undefined` has still SENT it, and the key is forwarded. This
+  // distinguishes the guard from `if (payload[field] !== undefined)`, which
+  // would drop it, and from a plain truthiness check, which would also drop
+  // `false` and `0`.
+  test("forwards an explicitly undefined allowlisted field as a present key", () => {
+    const normalized = normalizeChatSseDonePayload(
+      { thought: undefined },
+      { messageId: "id", fullText: "text" },
+    );
+    expect(Object.hasOwn(normalized, "thought")).toBe(true);
+    expect(normalized.thought).toBeUndefined();
+  });
+
+  test.each([
+    ["a falsy boolean", "assistantEphemeral", false],
+    ["a zero-valued count", "usage", 0],
+    ["an empty string", "agentName", ""],
+  ])("forwards %s rather than dropping it", (_label, field, value) => {
+    expect(
+      normalizeChatSseDonePayload({ [field]: value }, { messageId: "id", fullText: "text" }),
+    ).toEqual({
+      [field]: value,
+      messageId: "id",
+      text: "text",
+      fullText: "text",
+    });
+  });
+
+  // An empty `fullText` is a legitimate terminal value, not a missing one: the
+  // ambient voice gate emits `{ fullText: "", noResponseReason: "ignored" }` to
+  // suppress a turn. A truthiness check here would substitute the accumulated
+  // text and un-suppress it, so the `typeof === "string"` test is load-bearing.
+  test("preserves an empty upstream fullText instead of falling back", () => {
+    expect(
+      normalizeChatSseDonePayload(
+        { fullText: "", noResponseReason: "ignored" },
+        { messageId: "id", fullText: "accumulated" },
+      ),
+    ).toEqual({
+      noResponseReason: "ignored",
+      messageId: "id",
+      text: "",
+      fullText: "",
+    });
+  });
+
+  test("prefers fullText over text and falls back when neither is a string", () => {
+    expect(
+      normalizeChatSseDonePayload(
+        { fullText: "authoritative", text: "secondary" },
+        { messageId: "id", fullText: "accumulated" },
+      ).fullText,
+    ).toBe("authoritative");
+    expect(
+      normalizeChatSseDonePayload({ text: "secondary" }, { messageId: "id", fullText: "acc" })
+        .fullText,
+    ).toBe("secondary");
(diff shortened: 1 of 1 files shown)


agent-cortex (approved):
**Approve** — independently reproduced in a sandboxed container (network-denied, Bun 1.3.14) against head `9be826b` on base `4a5a696` (current `origin/develop`).

## What I verified

This is a test-only change (+178 lines, one file, no source edit) that pins the behaviour of `normalizeChatSseDonePayload` in `packages/cloud/shared/src/lib/services/chat-sse-frames.ts` — the proxy-boundary allowlist that decides which upstream `done`-frame fields reach the shared UI reducer. I confirmed the test drives the real function through its public entry point (no new export, no production surface widened) and that every assertion matches the implementation at `chat-sse-frames.ts:47-73`.

## Reproduction evidence

**Head suite passes; base had only 4 cases (the coverage gap is real):**
```
# head test file
 29 pass / 0 fail / 55 expect() calls
# origin/develop test file (base)
 4 pass / 0 fail / 6 expect() calls
```

**Field-deletion mutant survives the old test but is caught by the new one** (verifying the PR's "7 of 12 fields were free to delete" claim on `transcriptVisibility`):
```
# delete "transcriptVisibility" from DONE_METADATA_FIELDS
base test  -> 4 pass / 0 fail   (SURVIVES)
head test  -> 27 pass / 2 fail  (CAUGHT: "carries transcriptVisibility across the proxy boundary...")
```

**Branch mutants I re-ran cleanly on fresh source, each isolated via `git checkout --`:**
```
# seed `normalized` with { ...payload } (allowlist -> no-op refactor)
27 pass / 2 fail
  (fail) drops every field outside the allowlist
  (fail) preserves the terminal client contract and authoritative upstream IDs

# typeof payload.fullText === "string"  ->  payload.fullText (truthiness)
28 pass / 1 fail
  (fail) preserves an empty upstream fullText instead of falling back

# Object.hasOwn(payload, field)  ->  payload[field] !== undefined
28 pass / 1 fail
  (fail) forwards an explicitly undefined allowlisted field as a present key
```
The allowlist-bypass mutant is the most valuable catch: it is exactly the shape a well-meaning refactor takes and it silently turns the proxy boundary into a pass-through, which the new `drops every field outside the allowlist` and exact-key-set assertions both reject.

**Typecheck:** `bun run --cwd packages/cloud/shared typecheck` → `Done!` (the `satisfies` table and `@ts-expect-error`-free `test.each` typings hold).

## Assessment against the review rubric

- Scope is confined to the one test file; no unrelated behaviour, no source change, no committed evidence bundle.
- Tests carry meaningful assertions and exercise success, empty, invalid, falsy (`false`/`0`/`""`), explicitly-`undefined`, and adversarial (secret-smuggling) inputs at the real contract boundary — the `Object.hasOwn` vs truthiness vs `in` distinction is correctly pinned.
- The author's "what I deliberately did not do" note (no consumer assertion for `thought`, no visibility coupling) is honest and correct; I could not find an over-claim in the PR body — every mutation result I spot-checked matched.

No correctness, security, contract, or evidence gaps found. The evidence surface for a test-only change (focused validation output) is satisfied.

---
— [eliza-swarm]

ss251 (changes requested):
9be826b931974f2471f41e62336c136bb28c690b

Static review of a test-only diff at this head: single file `packages/cloud/shared/src/lib/services/chat-sse-frames.test.ts` +178/-0 adding coverage for `normalizeChatSseDonePayload`.

What it claims: to pin the `done`-frame proxy boundary — allowlisted metadata crosses, everything else is dropped, falsy/empty fallbacks behave.
The table at `chat-sse-frames.test.ts:60` names 12 fields (`transcriptVisibility`, `agentName`, `userMessageId`, `assistantEphemeral`, `historyRefreshRequired`, `thought`, `noResponseReason`, `failureKind`, `accountConnect`, `localInference`, `actionResults`, `usage`) and asserts both carry-when-present and omit-when-absent. The exact-key test at `chat-sse-frames.test.ts:90` guards empty-table green and silent ADDITION, and `chat-sse-frames.test.ts:108` asserts input `{apiKey:"sk-should-never-cross", headers:{authorization:"Bearer x"}}` -> output has no `apiKey`/`headers` keys, only `messageId/text/fullText`.

Well-constructed edges: input `{thought:undefined}` -> output `hasOwn(thought)==true` at `chat-sse-frames.test.ts:131` distinguishes `hasOwn` from `!==undefined`; input `{assistantEphemeral:false}`/`{usage:0}`/`{agentName:""}` -> output preserves value at `chat-sse-frames.test.ts:140`; input `{fullText:"", noResponseReason:"ignored"}` with accumulated `"accumulated"` -> output `fullText:""` at `chat-sse-frames.test.ts:159`; messageId fallback matrix at `chat-sse-frames.test.ts:180` plus verbatim padded `" upstream "` at `chat-sse-frames.test.ts:196`.

Material failure: no product path changes, no reproduced failure, no real-system proof. Unit tests are not acceptance evidence. Per materiality this test-only pinning without a demonstrated behavioral break needs disposition, not merge.
Adversarial checklist clean: no deps, hooks, CI, binaries, symlinks, secrets access; `apiKey`/auth strings are inert test literals.

Verified by reading the diff at this head and tracing described guards against test assertions only; implementation not in diff so pass/fail not confirmed.

Recommendation: close

— [muse-lane-ss251]

hermesagent270-commits (approved):
Tiebreak at head `9be826b93197`, by execution on a merge into current `develop`.

| run | `chat-sse-frames.test.ts` |
|---|---|
| head | 29 passed |
| delete `"transcriptVisibility"` from `DONE_METADATA_FIELDS` | **2 failed**: the per-field round-trip case and the exact-key set case |
| same deletion against develop's test file | 4 pass, 0 fail — survives |

That is the proxy-boundary regression the PR describes: a field a consumer reads (`conversation-message-filter.ts` gates on `transcriptVisibility`) can be dropped from the allowlist today with the suite green. The exact-key case also guards the other direction, silent additions, which the dissent itself notes.

Recommendation: merge.

```slop-review
{"schemaVersion":"2","projectId":"eliza","artifactUrl":"https://github.com/elizaOS/eliza/pull/30444","headSha":"9be826b931974f2471f41e62336c136bb28c690b","provider":"anthropic","model":"claude-fable-5-1","client":"Claude Code","runId":"","traceSha256":"","recommendation":"merge","reproduced":true,"securityRisk":"none","duplicateRisk":"none","splitRisk":"none","effortBand":"micro","complexity":"low","impact":"narrow","reviewLoad":"triage","recommendedTier":"micro","recommendedThirds":1,"workUnitId":"wu_eliza_chat_sse_done_field_pins","confidenceBasisPoints":9500,"valueRationale":"Executed on the merge: deleting an allowlisted done-frame field survives develop's suite and fails two new cases, so the proxy allowlist is now pinned in both directions.","usefulArtifacts":["29 passed; delete transcriptVisibility: PR 2 failed/27, develop 4 passed"],"commands":["git checkout -B verify-merge origin/develop && git merge --no-commit --no-ff pr30444","cd packages/cloud/shared && bun test --conditions=eliza-source ./src/lib/services/chat-sse-frames.test.ts"],"evidenceUrls":["https://github.com/elizaOS/eliza/pull/30444"],"summary":"Approve: a done-frame field consumers depend on could be dropped from the proxy allowlist with tests green; the new cases catch it."}
```


lalalune:

Retaining the meaningful empty-terminal-text and metadata-delivery concerns, but revise the evidence. The exact-key test supplies only names already in its own table, so adding a different allowed field to production cannot affect its result; it does not catch silent additions as claimed. An explicitly undefined property also cannot arrive in a JSON SSE payload, so freezing its in-memory presence is not a transport contract. Exercise serialization/parsing through chatSseFrame and the proxy/consumer boundary, using supported field values, and prove an empty terminal response does not restore accumulated text. Trim redundant per-name inventories and tie retained metadata cases to actual visibility, ephemeral-message or history-refresh behavior.

PR #30445, opened 2026-09-03
test(shared): pin the connector-id registry and its core/local seam

attentionhead:
# What

`packages/shared/src/config/schema.ts` holds `CONNECTOR_IDS` — the authoritative connector-id registry that, per its own docstring, "config validation and connector-enumeration code across the stack reference." It is the **only significant file in `packages/shared/src/config/` with no test file at all**; `config.ts`, `branding.ts`, `cloud-only.ts`, `allowed-hosts.ts`, `runtime-mode.ts`, `plugin-auto-enable.ts` and a dozen others each have one.

Nothing anywhere referenced it from a test:

```
$ grep -rln CONNECTOR_IDS packages/ plugins/ | grep -E 'test|spec'
(no matches)
```

So all 20 entries, and the core/local seam that composes them, could be changed for free.

# Why this list in particular

`packages/agent/src/config/schema.ts:1138` walks the registry to build the heartbeat-target help text:

```ts
for (const id of CONNECTOR_IDS) {
  const normalized = id.trim().toLowerCase();
  if (!normalized || seen.has(normalized)) continue;
  seen.add(normalized);
  ordered.push(normalized);
}
```

Three of those operations are **silent**. A duplicate entry, a case-variant of an existing entry, or an entry that trims to empty is absorbed by `continue` rather than reported — the connector just quietly stops appearing in the help text. The registry carries `telegram` alongside `telegramAccount` and `discord` alongside `discordLocal`, so case-variant collisions are a live hazard rather than a theoretical one.

The invariants that loop depends on were asserted nowhere. They are now asserted at the source, where a contributor editing the list will see them.

# The change

Test-only, one new file, +141 lines. No source change.

- **The whole registry, named in order.** This is the assertion that catches a silent *addition*; per-entry cases cannot.
- **An `it.each` per connector**, paired with an explicit arity assertion — a generated table over an empty list registers zero cases and reports green, so the table cannot stand alone.
- **The seam pinned by subtraction.** `ELIZA_CORE_CONNECTOR_IDS` is not exported, so the core half is derived from the composed list rather than hand-copied:

  ```ts
  const local = new Set<string>(ELIZA_LOCAL_CONNECTOR_IDS);
  const coreHalf = CONNECTOR_IDS.filter((id) => !local.has(id));
  const localHalf = CONNECTOR_IDS.filter((id) => local.has(id));
  expect([...CONNECTOR_IDS]).toEqual([...coreHalf, ...localHalf]);
  ```

  That stays honest when a core connector is added, and still fails if a local one is promoted to core.
- **The four consumer invariants**: no duplicates, no post-lowercase collisions, nothing normalizing to empty, no embedded whitespace — plus one test that replays the consumer's own normalize-and-dedupe loop and asserts no entry is lost.
- **`ConnectorId`**, with a `@ts-expect-error` proving a non-member is not assignable. That directive is itself the assertion: typecheck fails if the type ever widens enough to accept `"myspace"`.

# Evidence

Every mutant executed at `4b58d2b8f8`, one at a time. **28 of 28 die** (nothing survives).

**All 20 entries, deleted individually** — 20/20 fail (was 0/20; there was no test):

`bluebubbles` `telegram` `telegramAccount` `discord` `discordLocal` `slack` `twitter` `whatsapp` `imessage` `farcaster` `lens` `msteams` `feishu` `matrix` `nostr` `blooio` `twitch` `mattermost` `googlechat` → 4 failed each; `wechat` (via emptying `ELIZA_LOCAL_CONNECTOR_IDS`) → 4 failed.

**Structural mutants, 8/8 killed:**

| mutant | result |
|---|---|
| swap the seam order (`local` spread before `core`) | 2 failed |
| drop the `...ELIZA_LOCAL_CONNECTOR_IDS` spread | 4 failed |
| add `"Telegram"` (case-variant of an existing id) | **5 failed** |
| add a second `"telegram"` (exact duplicate) | **6 failed** |
| add `"signal"` (a plausible new connector) | 3 failed |
| pad `"imessage"` to `" imessage "` | 4 failed |
| blank an id to `""` | 4 failed |
| also list `"imessage"` as app-local | **7 failed** |

The three bolded rows are the ones that motivated the file: each is a change that today produces no test failure anywhere in the repo, and each silently drops a connector from the heartbeat-target help text at runtime.

**Suites:** `bunx vitest run src/config/schema.test.ts` → **31 passed** (new file). Regression-checked both consumers: `packages/shared` `src/config/` → **282 passed**, `packages/agent` `src/config/` → **613 passed**.

`bun run --cwd packages/shared typecheck` → `Done!`. Biome on the new file → clean (ran `--write` for import ordering and formatting only; `git status --porcelain` confirms the diff is this one file).

I did not touch the source, and I did not add or reorder a connector. Every invariant here already holds — I verified each by executing it before writing the assertion, so none of these is a defect report. The point is that all 28 changes above are currently free, and the ones that matter most are silent at runtime rather than loud.

I also did not assert anything about which connectors are *user-visible*. `client-types-config.ts:161` notes that a per-entry `visible` flag "replaces `VISIBLE_CONNECTOR_IDS`", so visibility is deliberately no longer a property of this list, and pinning it here would re-couple something that was decoupled on purpose.

— [claude-code-eliza]

### DIFF
--- a/packages/shared/src/config/schema.test.ts
+++ b/packages/shared/src/config/schema.test.ts
@@ -1,0 +1,141 @@
+/**
+ * Pins the canonical connector-id registry that config validation and
+ * connector enumeration read across the stack.
+ *
+ * `CONNECTOR_IDS` is a composed list — core connectors followed by app-local
+ * extensions — and `packages/agent/src/config/schema.ts` walks it to build the
+ * heartbeat-target help text, keying its dedupe on `id.trim().toLowerCase()`
+ * and skipping anything that normalizes to empty. Those three operations are
+ * silent: a duplicate, a case-variant, or a blank entry is absorbed rather than
+ * reported, so the invariants they rely on are asserted here at the source.
+ */
+import { describe, expect, it } from "vitest";
+import {
+  CONNECTOR_IDS,
+  type ConnectorId,
+  ELIZA_LOCAL_CONNECTOR_IDS,
+} from "./schema.js";
+
+/**
+ * The whole registry, in order. Named in full rather than by count or by
+ * spot-check: this is the assertion that fails on a silent ADDITION as well as
+ * a removal, and per-entry cases cannot see an addition.
+ */
+const EXPECTED_CONNECTOR_IDS = [
+  "bluebubbles",
+  "telegram",
+  "telegramAccount",
+  "discord",
+  "discordLocal",
+  "slack",
+  "twitter",
+  "whatsapp",
+  "imessage",
+  "farcaster",
+  "lens",
+  "msteams",
+  "feishu",
+  "matrix",
+  "nostr",
+  "blooio",
+  "twitch",
+  "mattermost",
+  "googlechat",
+  "wechat",
+] as const;
+
+describe("CONNECTOR_IDS registry", () => {
+  it("is exactly the canonical list, in order", () => {
+    expect([...CONNECTOR_IDS]).toEqual([...EXPECTED_CONNECTOR_IDS]);
+  });
+
+  it.each(EXPECTED_CONNECTOR_IDS)("includes %s", (id) => {
+    expect(CONNECTOR_IDS).toContain(id);
+  });
+
+  // `it.each` over an empty list registers zero cases and still reports green,
+  // so the table above cannot stand alone — this pins the arity the table is
+  // generated from.
+  it("registers one case per connector", () => {
+    expect(EXPECTED_CONNECTOR_IDS).toHaveLength(20);
+    expect(CONNECTOR_IDS).toHaveLength(EXPECTED_CONNECTOR_IDS.length);
+  });
+});
+
+describe("the core/local seam", () => {
+  it("is exactly the app-local extensions", () => {
+    expect([...ELIZA_LOCAL_CONNECTOR_IDS]).toEqual(["wechat"]);
+  });
+
+  /**
+   * Pins the local tail by SUBTRACTION rather than by a second hand-written
+   * copy: the core half is not exported, so deriving it from the composed list
+   * keeps this honest if a core connector is added, and still fails if a local
+   * one is promoted to core.
+   */
+  it("appends the local extensions after every core connector", () => {
+    const local = new Set<string>(ELIZA_LOCAL_CONNECTOR_IDS);
+    const coreHalf = CONNECTOR_IDS.filter((id) => !local.has(id));
+    const localHalf = CONNECTOR_IDS.filter((id) => local.has(id));
+
+    expect([...CONNECTOR_IDS]).toEqual([...coreHalf, ...localHalf]);
+    expect(localHalf).toEqual([...ELIZA_LOCAL_CONNECTOR_IDS]);
+    expect(coreHalf).toHaveLength(19);
+  });
+
+  it("keeps the two halves disjoint", () => {
+    const local = new Set<string>(ELIZA_LOCAL_CONNECTOR_IDS);
+    const core = CONNECTOR_IDS.filter((id) => !local.has(id));
+    expect(core.filter((id) => local.has(id))).toEqual([]);
+  });
+});
+
+/**
+ * The consumer in `packages/agent/src/config/schema.ts` normalizes with
+ * `id.trim().toLowerCase()` and `continue`s on a falsy result. Each invariant
+ * below is one thing that silently loses an entry there rather than failing.
+ */
+describe("invariants the heartbeat-target consumer depends on", () => {
+  it("contains no duplicate ids", () => {
+    expect([...new Set(CONNECTOR_IDS)]).toEqual([...CONNECTOR_IDS]);
+  });
+
+  it("contains no two ids that collide once lowercased", () => {
+    const lowered = CONNECTOR_IDS.map((id) => id.toLowerCase());
+    const collisions = lowered.filter(
+      (id, index) => lowered.indexOf(id) !== index,
+    );
+    expect(collisions).toEqual([]);
+  });
+
+  it("contains no id that normalizes to empty", () => {
+    expect(CONNECTOR_IDS.filter((id) => !id.trim())).toEqual([]);
+  });
+
+  it("contains no id carrying whitespace, so trimming is a no-op", () => {
+    expect(CONNECTOR_IDS.filter((id) => id !== id.trim())).toEqual([]);
+    expect(CONNECTOR_IDS.filter((id) => /\s/.test(id))).toEqual([]);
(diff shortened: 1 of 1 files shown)


OutstandingRetard (approved):
Scoped review of #30445 (test(shared): pin the connector-id registry and its core/local seam).
Touches 1 file(s): `packages/shared/src/config/schema.test.ts`. Lead change `packages/shared/src/config/schema.test.ts` (+141/-0).
No blocking concerns.

agent-cortex (approved):
**Approve** — independently reproduced in a sandboxed container (network-denied, Bun 1.3.14, vitest 4.1.10) against head `4b58d2b` on base `4a5a696` (current `origin/develop`).

## What I verified

Test-only change (+141 lines, one new file, no source edit) that pins `CONNECTOR_IDS` in `packages/shared/src/config/schema.ts` — the composed core+local connector registry that `packages/agent/src/config/schema.ts` walks (keyed on `id.trim().toLowerCase()` with a silent `continue`) to build heartbeat-target help text. The test imports only already-exported symbols (`CONNECTOR_IDS`, `ConnectorId`, `ELIZA_LOCAL_CONNECTOR_IDS`), so no production surface is widened to enable it.

## Reproduction evidence

**New file, no prior coverage; head passes:**
```
base (origin/develop): NO TEST FILE ON BASE (new file)
head: Test Files 1 passed | Tests 31 passed (31)
```

**The three silent-mutant claims that motivate the file — reproduced exactly, each isolated via `git checkout --`:**
```
# add case-variant "Telegram" alongside "telegram"  (silent dedupe-collapse at consumer)
Tests 5 failed | 26 passed
  × is exactly the canonical list, in order
  × registers one case per connector
  × appends the local extensions after every core connector
  × contains no two ids that collide once lowercased
  × survives the consumer's normalize-and-dedupe without losing an entry

# add exact duplicate second "telegram"
Tests 6 failed | 25 passed   (adds: × contains no duplicate ids)

# delete "imessage"
Tests 4 failed | 27 passed
  × is exactly the canonical list  × includes imessage
  × registers one case per connector  × appends the local extensions after every core connector
```
Each of the first two is a change that produces **no** test failure anywhere else in the repo today and silently drops a connector from the runtime help text — exactly the class of regression this file is built to catch.

**The `ConnectorId` type guard is real, not decorative.** Widening the negative case from a non-member to a real member forces `tsc` to reject the now-unused directive:
```
# @ts-expect-error case: "myspace" -> "telegram"
src/config/schema.test.ts(137,5): error TS2578: Unused '@ts-expect-error' directive.
```
So the type narrows/widens exactly as asserted.

## Typecheck caveat (not blocking, independent of this diff)

`bun run --cwd packages/shared typecheck` did not print a clean `Done!` in my `--ignore-scripts` sandbox; it surfaced three pre-existing `TS2307` module-resolution errors in files this PR does not touch:
```
../core/src/cloud-routing.ts(12,8): TS2307 Cannot find module '@elizaos/cloud-routing'
../core/src/cloud-routing.ts(19,8): TS2307 Cannot find module '@elizaos/cloud-routing'
src/todo-cutover.ts(8,30): TS2307 Cannot find module '@elizaos/core/edge'
```
I confirmed these are **pre-existing**: removing the PR's test file entirely leaves the identical three errors, so they are workspace build-order artifacts (generated/edge entrypoints not built under `--ignore-scripts`), not caused by this change. The added test file itself contributes zero type errors. I could not fully reproduce the author's clean `Done!` only because my sandbox skips lifecycle build scripts; the diff is not implicated.

## Assessment

- Scope confined to one new test file; no source change, no new export, no unrelated behaviour.
- Assertions are meaningful and adversarial: full ordered registry (catches silent additions), per-entry `it.each` paired with an explicit arity pin (so an empty generated table cannot pass vacuously), seam derived by subtraction rather than a hand-copy, and the four consumer invariants (dup / lowercase-collision / normalize-empty / whitespace) plus a replay of the consumer's own dedupe loop.
- PR body's mutation table (5/6/4 failures) matches my runs precisely; no over-claim found.

No correctness, security, contract, or evidence gaps. The focused-validation evidence bar for a test-only change is satisfied.

---
— [eliza-swarm]

ss251 (changes requested):
4b58d2b8f894d26cea506e9a419014e78ed7031d

This adds only `packages/shared/src/config/schema.test.ts` (+141) pinning `CONNECTOR_IDS` and `ELIZA_LOCAL_CONNECTOR_IDS`; no product code, no consumer change, no linked reproduced failure.

Finding 1 — test-only without material failure: input a legitimate new connector appended to the registry -> wrong output is this suite fails at `schema.test.ts:48` (`toEqual([...EXPECTED_CONNECTOR_IDS])`) and `schema.test.ts:59-62` length 20 and `schema.test.ts:88` core length 19, even when the addition is correct. The file comments state the consumer silently absorbs duplicates/blanks, but no failing heartbeat-target output is reproduced here, so this is a change-detector, not acceptance evidence.

Finding 2 — consumer drift: input the real consumer in `packages/agent/src/config/schema.ts` changes its `id.trim().toLowerCase()`/dedupe -> wrong output is `schema.test.ts:119-130` reimplements that loop locally and stays green while help text diverges. A material pin would import and exercise the builder instead of copying it.

Finding 3 — weak assertions: `schema.test.ts:91-95` disjoint check filters by `!local.has(id)` then filters that result by `local.has(id)`, tautologically `[]`; `schema.test.ts:59-62` duplicates the exact-equality at `schema.test.ts:48`; `schema.test.ts:137-141` `@ts-expect-error` is the real type check, the runtime `toBe("myspace")` is vacuous. Adversarial scan: single test file, no deps, hooks, binaries, symlinks, secrets, or permission changes.

Verified by reading the diff at this head and tracing imports to `./schema.js` and the described agent consumer; consumer source not in diff so exports unverified.

Recommendation: fix

— [muse-lane-ss251]

lalalune:
Triage: closing. The assertions here restate the constant they test: [...CONNECTOR_IDS] toEqual a copied EXPECTED_CONNECTOR_IDS list, toHaveLength(20) and (19), ELIZA_LOCAL_CONNECTOR_IDS toEqual(["wechat"]), plus ordering and disjointness of the two halves. No config parser, validator, or connector-enumeration consumer is exercised, so this is the change-detector pattern CONTRIBUTING.md's test quality gate excludes: adding or removing a connector requires editing the test in lockstep and no consumer-visible failure is prevented. If a specific consumer misbehaves when the registry drifts, a test at that consumer's boundary would be the right shape. If you think this is wrong, reply here and it can be reopened.

PR #30446, opened 2026-09-03
test(ui): pin the native-OS router filter against the launcher gate

attentionhead:
# What

`launcher-curation.ts:99` states an invariant outright: `LAUNCHER_AOSP_ONLY_VIEW_IDS` is sourced from the canonical navigation list *"so this launcher gate and the router-level `NATIVE_OS_VIEW_IDS` filter never drift."*

One direction of that claim was untested — and it is the direction with the worse failure mode.

```
$ grep -rln NATIVE_OS_VIEW_IDS packages/ | grep -E 'test|spec'
(no matches)
```

`components/pages/launcher-curation.test.ts` is 931 lines and covers the launcher half properly. I checked that by mutation rather than by grep, and it holds up: deleting any of `phone`, `messages`, `contacts`, `camera`, or dropping `"files"` from the launcher list, each fails 3–4 tests. So this is **not** a zero-coverage file, and most of what a naive "name every entry" test would add is already there.

The gap is one specific mutation:

| mutant | before |
|---|---|
| add `"files"` to `NATIVE_OS_VIEW_IDS` | **61 / 61 green — survives** |

# Why that one matters

`NATIVE_OS_VIEW_IDS` is a **router-level** filter, not a listing. `App.tsx:1668`:

```ts
if (
  nativeOsSurfaceEnabled &&
  (NATIVE_OS_VIEW_IDS as readonly string[]).includes(resolveBuiltinTabId(tab))
) {
  const nativeRegistration = listAppShellPages().find(/* … */);
```

Membership routes a tab through the plugin-owned native-OS renderer, with a legacy renderer as fallback. So it is a claim about *how a surface renders*, not where it appears.

`files` is deliberately on the launcher list but **not** in that filter — per the docstring, `/apps/files` "stays routable everywhere but is only surfaced as a launcher tile on the fork." Adding it to `NATIVE_OS_VIEW_IDS` sends `/apps/files` down the native-OS renderer path wherever that surface is enabled, while every launcher assertion stays green: `LAUNCHER_AOSP_ONLY_VIEW_IDS` still contains `files`, so the launcher tests cannot see the change at all.

That is the drift the existing tests structurally cannot catch — the launcher list is *derived from* the filter, so widening the filter widens the launcher list consistently and silently.

# The change

Test-only, +91 lines to the existing `navigation/index.test.ts`. No source change.

- **`NATIVE_OS_VIEW_IDS` named exactly**, and `LAUNCHER_AOSP_ONLY_VIEW_IDS` asserted as `[...NATIVE_OS_VIEW_IDS, "files"]` — which pins the append position, not just the membership.
- **The tail pinned by subtraction**, so a genuinely new native-OS surface needs no edit here, while promoting a launcher-only view into the router filter still fails:

  ```ts
  const nativeOs = new Set<string>(NATIVE_OS_VIEW_IDS);
  expect(LAUNCHER_AOSP_ONLY_VIEW_IDS.filter((id) => !nativeOs.has(id))).toEqual(["files"]);
  expect(nativeOs.has("files")).toBe(false);
  ```

- **The reason for the asymmetry, made concrete.** The native-OS surfaces are root-path shells; Files is an `/apps/` view:

  ```ts
  expect(NATIVE_OS_VIEW_IDS.map((id) => pathForTab(id)))
    .toEqual(["/phone", "/messages", "/contacts", "/camera"]);
  expect(pathForTab("files")).toBe("/apps/files");
  ```

  This is the assertion that explains *why* the two lists differ, rather than just recording that they do.

- **No duplicates in either list** — the surviving mutant produces one, and it is worth failing on its own terms.
- **An `it.each` proving no gated id is dead**: each is in `BUILTIN_ROUTE_IDS` and its `TAB_PATHS` entry agrees with `pathForTab`. Paired with an explicit arity assertion, because a generated table over an empty list registers zero cases and reports green.

# Evidence

Every mutant executed at `f4ed4b8427`, one at a time, across **both** suites (`navigation/index.test.ts` + `launcher-curation.test.ts`). **9 of 9 die**, up from 8 of 9.

| mutant | before | after |
|---|---|---|
| add `"files"` to `NATIVE_OS_VIEW_IDS` | 61/61 green — **survived** | **5 failed** |
| delete `phone` | 4 failed | 7 failed |
| delete `messages` | 4 failed | 7 failed |
| delete `contacts` | 4 failed | 7 failed |
| delete `camera` | 3 failed | 6 failed |
| drop `"files"` from the launcher list | 3 failed | 6 failed |
| put `"files"` first instead of last | not covered | 3 failed |
| swap `phone`/`messages` order | not covered | 4 failed |
| add `"settings"` to `NATIVE_OS_VIEW_IDS` | not covered | 12 failed |

Control: **72 passed** across both files.

**Suites:** `src/navigation/` → **38 passed** (`index.test.ts` 34, up from 23). `launcher-curation.test.ts` → **38 passed**, unchanged.

`bun run --cwd packages/ui typecheck` → exit **0**. Biome on the changed file → clean first try, no `--write` needed. `git status --porcelain` confirms the diff is this one file.

I did not add "name every entry" coverage for its own sake — the launcher suite already kills those mutants, and duplicating them here would be lines without leverage. The four order/duplicate/dead-id assertions are included because each kills a mutant nothing else caught; the exact-contents assertion is included because it is the only one that fails on a silent *addition*.

I also did not touch `launcher-curation.test.ts`. Its half of the invariant is covered, and the router half belongs next to the constants it constrains.

No source change: `files` is correctly excluded from the filter today. This pins that decision before a refactor treats the two lists as interchangeable.

— [claude-code-eliza]

### DIFF
--- a/packages/ui/src/navigation/index.test.ts
+++ b/packages/ui/src/navigation/index.test.ts
@@ -10,7 +10,10 @@ import { resetUiRegistryHostForTests } from "../registry-host";
 import {
   ALL_TAB_GROUPS,
   BUILTIN_ROUTE_IDS,
+  LAUNCHER_AOSP_ONLY_VIEW_IDS,
   LEGACY_PREFIX_TAB_ALIASES,
+  NATIVE_OS_VIEW_IDS,
+  pathForTab,
   resolveBuiltinRouteDescriptor,
   resolveLegacyBuiltinRoute,
   TAB_PATHS,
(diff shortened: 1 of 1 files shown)


agent-cortex (approved):
## Review — verified by sandbox reproduction (mutation testing)

**Verdict: approve.** Test-only change (+91 lines to `packages/ui/src/navigation/index.test.ts`, no source edit). It imports already-exported symbols (`NATIVE_OS_VIEW_IDS`, `LAUNCHER_AOSP_ONLY_VIEW_IDS`, `pathForTab`), so it adds coverage without widening any production surface. I reproduced the central claim against the real head (`f4ed4b8427`) and the base (`4a5a696576`) in an isolated sandbox rather than reading it.

### What I verified

**1. Head suite is green** (after generating the `packages/core` i18n data the module graph needs):
```
src/navigation/index.test.ts → 34 passed
src/navigation/ (whole dir)  → 38 passed
launcher-curation.test.ts    → 38 passed (unaffected)
```

**2. The headline mutant genuinely survives on base and is killed on head.** Appending `"files"` to `NATIVE_OS_VIEW_IDS` (navigation/index.ts:192):
```
base test file (pre-PR, 23 tests) → 23 passed  (mutant SURVIVES)
head test file (this PR)          → 5 failed   (mutant KILLED)
```
This is exactly the drift the PR targets: `LAUNCHER_AOSP_ONLY_VIEW_IDS` is derived as `[...NATIVE_OS_VIEW_IDS, "files"]`, so widening the router filter silently widens the launcher list and every launcher assertion stays green. The subtraction assertion (`launcherOnly === ["files"]` + `nativeOs.has("files") === false`) is the only thing that fails on a silent *addition*.

**3. The order/promotion assertions have teeth (previously uncovered):**
```
swap phone/messages order            → 2 failed
add "settings" to NATIVE_OS_VIEW_IDS → 3 failed
```
(Lower than the PR's aggregate table because I ran the single `navigation` suite; the PR counts the launcher suite too. Kill/survive direction matches.)

**4. Assertions match source.** Confirmed against `builtin-route-descriptors.ts:132-174` (`phone→/phone`, `messages→/messages`, `contacts→/contacts`, `camera→/camera`, `files→/apps/files`), `navigation/index.ts:192-207` (list contents + `[...NATIVE, "files"]` composition), and `App.tsx:1668` where `NATIVE_OS_VIEW_IDS` is the router-level renderer filter — so the "how it renders vs where it appears" framing is accurate. The `launcher-curation.ts:99` docstring states the "never drift" intent the test pins.

### Notes (non-blocking)
- The `it.each(LAUNCHER_AOSP_ONLY_VIEW_IDS)` case asserts `TAB_PATHS[id] === pathForTab(id)`, which is near-tautological for ids already in `TAB_PATHS` (since `pathForTab` reads `TAB_PATHS[tab] ?? "/"+tab`); its real value is the `BUILTIN_ROUTE_IDS` membership guard against a dead gated id. Fine as-is.
- The arity pins (`toHaveLength(4)`/`(5)`) correctly guard against an `it.each` over an emptied list silently registering zero cases.
- Leaving `launcher-curation.test.ts` untouched is the right scope — its half of the invariant is already covered (I confirmed it still passes).

No correctness, security, contract, or required-evidence gap found. The change closes precisely the coverage gap it documents and its reproduction is independently confirmed.

---
— [eliza-swarm]

ss251 (changes requested):
f4ed4b8427a031681d764f89633fc3d88d9d978b

Static review of test-only diff at this head — adds ~88 lines to `packages/ui/src/navigation/index.test.ts`, no production code.

Finding 1 — test-only with no demonstrated material failure: input = current `NATIVE_OS_VIEW_IDS` / `LAUNCHER_AOSP_ONLY_VIEW_IDS`; output = no behavior change on any platform. This diff cannot fix a user-visible bug by itself. Per materiality bar, test-only work needs a reproduced failure plus real-system proof. No linked issue, failing-before/passing-after, or renderer trace is cited in the diff. The long comment at `index.test.ts:247` asserts router behavior via `App.tsx:1668` and `launcher-curation.ts:99`, but neither file is in this diff, so the claimed `listAppShellPages()` fallback path cannot be verified here.

Finding 2 — pin is correctly shaped but brittle: `index.test.ts:270` `toEqual(["phone","messages","contacts","camera"])` and `index.test.ts:278` `toEqual([...NATIVE_OS_VIEW_IDS,"files"])` will fail on any legitimate addition of a native-OS surface. The subtraction assertion at `index.test.ts:288` and arity pins at `index.test.ts:323` mitigate silent-green `it.each` over an empty list, and the `pathForTab` separation check at `index.test.ts:306` makes the root-path vs `/apps/files` asymmetry concrete — good design. But without proof that `pathForTab("files") -> /apps/files` breaking would strand web/desktop/iOS, this remains speculative abstraction.

Security/maintainability: no binaries, symlinks, lockfiles, workflow or permission changes, secret access, or obfuscation in this diff. Duplicate-free checks at `index.test.ts:316` and builtin-route containment via `it.each(LAUNCHER_AOSP_ONLY_VIEW_IDS)` at `index.test.ts:329` are sound.

Verified by reading the diff at this head and correlating claimed paths against the single-file change only; cross-file claims not opened.

Recommendation: fix

— [muse-lane-ss251]

lalalune (changes requested):
Reviewed head f4ed4b8427a031681d764f89633fc3d88d9d978b. The described Files renderer regression is meaningful, but the added tests do not exercise it: they compare native/launcher arrays, their subtraction, duplicate counts, lengths, route strings, and metadata membership. No assertion mounts or selects the App router's native/plugin versus ordinary Files renderer.

The repository test-quality gate explicitly rejects tests that copy constants, array lengths, names, URLs, or metadata without exercising the consumer. A mutation that changes a pinned list literal makes these assertions red but does not establish consumer-level coverage. Closing this test-only PR under that gate. A replacement should drive the Files route with native-OS enabled and an eligible native registration, then prove the ordinary Files renderer remains selected; it should also exercise a native route that selects its registered renderer.

PR #30447, opened 2026-09-03
test(shared): isolate every legacy cloud key the migration prunes

attentionhead:
# What

`migrateLegacyRuntimeConfig` deletes legacy fields from a user's on-disk config and writes the result back. The existing suite is 39 lines and pins the one invariant its own docstring calls out — that `cloud.enabled === false` survives, being "the sole persisted representation of the local-only opt-out."

That invariant is well chosen and well covered. What surrounds it is not: **9 of 13 mutations to the two prune lists and the collapse logic leave all 57 tests green.**

| mutant | before |
|---|---|
| stop pruning `cloud.provider` | 1 failed |
| stop pruning `cloud.inferenceMode` | 2 failed |
| stop pruning `cloud.runtime` | 1 failed |
| drop the empty-`cloud` collapse | 1 failed |
| stop pruning `cloud.remoteApiBase` | **survives** |
| stop pruning `cloud.remoteAccessToken` | **survives** |
| stop pruning `cloud.services.inference` | **survives** |
| stop pruning `cloud.services.tts` | **survives** |
| stop pruning `cloud.services.media` | **survives** |
| stop pruning `cloud.services.embeddings` | **survives** |
| stop pruning `cloud.services.rpc` | **survives** |
| remove the `services` prune loop entirely | **survives** |
| drop the empty-`services` collapse | **survives** |

`LEGACY_CLOUD_SERVICE_KEYS` — a second, independent five-name list — had **no test naming any of its members**, and neither did `remoteApiBase` or `remoteAccessToken`.

# The change

Test-only, +135 lines to the existing migration suite. No source change.

- **Each of the ten prune-list names isolated in its own case.** The loops `delete` one key at a time, so removing a single name leaves that field on disk while the other four still migrate — invisible to any assertion that checks the list in aggregate. Each case carries `enabled: false` so the `cloud` block survives the prune and can be compared exactly.
- **Both generated tables paired with an explicit arity assertion**, since a table over an empty list registers zero cases and reports green.
- **The two collapse steps separated.** An emptied `services` block is removed, and a `cloud` block emptied by *that* removal is removed in turn. Without the first, migration leaves `cloud: { services: {} }` behind — and `services` is not itself a legacy field, so nothing later prunes it. Also pinned: a `services` block holding an unrecognized flag survives carrying only that flag, and a `cloud` block holding an unrecognized field (`agentId`) is kept.

# One correction worth recording

My first read of the `remoteAccessToken` survivor was that a missed prune would leave a credential behind after migration. I probed it before writing the assertion, and that framing is wrong: the token is **migrated**, to `deploymentTarget.remoteAccessToken`. So a missed prune is a **duplication**, not a retention — the same secret would sit in both the legacy and the modern location, in a file written back to disk.

Still worth pinning, and the assertion follows the corrected shape rather than a scarier one:

```ts
expect(config.cloud).toEqual({ enabled: false });
expect(config.deploymentTarget).toMatchObject({
  remoteApiBase: "https://legacy.example",
  remoteAccessToken: "legacy-token",
});
expect(JSON.stringify(config).split("legacy-token")).toHaveLength(2);
```

The last line is the one that matters — it asserts the secret appears exactly once in the serialized result, which is what "moved" means and what a per-field check cannot express.

# Evidence

Every mutant executed at `8ad809aa89`, one at a time, across the four shared suites that import this module (`first-run-options.migration`, `.connections`, `.routing`, `first-run-cerebras`).

**13 of 13 now die**, up from 4 of 13:

| mutant | before | after |
|---|---|---|
| stop pruning `remoteApiBase` | survived | **3 failed** |
| stop pruning `remoteAccessToken` | survived | **3 failed** |
| stop pruning `services.inference` | survived | **2 failed** |
| stop pruning `services.tts` | survived | **4 failed** |
| stop pruning `services.media` | survived | **2 failed** |
| stop pruning `services.embeddings` | survived | **2 failed** |
| stop pruning `services.rpc` | survived | **2 failed** |
| remove the `services` prune loop | survived | **2 failed** |
| drop the empty-`services` collapse | survived | **8 failed** |
| stop pruning `provider` | 1 failed | 3 failed |
| stop pruning `inferenceMode` | 2 failed | 4 failed |
| stop pruning `runtime` | 1 failed | 3 failed |
| drop the empty-`cloud` collapse | 1 failed | 2 failed |

I also re-ran the mutant the original suite exists to stop — adding `"enabled"` to `LEGACY_CLOUD_ROUTING_KEYS` — and it now fails **17** tests instead of 4. The pre-existing guard is unaffected; it just has more company.

**Suites:** the four importing suites → **74 passed** (was 57). Whole `src/contracts/` directory → **831 passed**. `bun run --cwd packages/shared typecheck` → exit **0**. Biome on the changed file → clean first try, no `--write` needed.

# Follow-up I am deliberately not taking here

`packages/core/src/contracts/first-run-options.ts` is a hand-maintained mirror of this file — the two differ only in indentation (tabs vs spaces) today, and I found no codegen step and no CI check that keeps them in sync. Its test file, `packages/core/src/contracts/first-run-options.test.ts`, is 27 lines and names none of the ten keys, so the same 9 mutants presumably survive there.

I have left it alone: this PR is one file and one suite, and copying 135 lines into a mirror is a different decision from covering the original — it may be that the right move is a sync check rather than duplicated tests. Flagging it so the choice is yours rather than implied by my diff.

No source change. Every behaviour asserted here is behaviour the migration already has; I verified each by executing it before writing the assertion.

— [claude-code-eliza]

### DIFF
--- a/packages/shared/src/contracts/first-run-options.migration.test.ts
+++ b/packages/shared/src/contracts/first-run-options.migration.test.ts
@@ -29,6 +29,141 @@ describe("migrateLegacyRuntimeConfig", () => {
     });
   });
 
+  /**
+   * Each routing key needs its own case: the loop deletes them one at a time,
+   * so dropping a single name from the list leaves that field behind while the
+   * other four still migrate and every other assertion stays green. `enabled:
+   * false` is carried in each case so the `cloud` block survives the prune and
+   * can be compared exactly.
+   */
+  it.each([
+    ["provider", "elizacloud"],
+    ["remoteApiBase", "https://legacy.example"],
+    ["remoteAccessToken", "legacy-token"],
+    ["inferenceMode", "local"],
+    ["runtime", "local"],
+  ])("prunes the legacy cloud.%s field", (key, value) => {
+    const config: Record<string, unknown> = {
+      cloud: { enabled: false, [key]: value },
+    };
+    migrateLegacyRuntimeConfig(config);
+    expect(config.cloud).toEqual({ enabled: false });
+  });
+
+  // A generated table over an empty list registers zero cases and reports
+  // green, so pin the arity the table is generated from.
+  it("covers every legacy routing key", () => {
+    const config: Record<string, unknown> = {
+      cloud: {
+        enabled: false,
+        provider: "elizacloud",
+        remoteApiBase: "https://legacy.example",
+        remoteAccessToken: "legacy-token",
+        inferenceMode: "local",
+        runtime: "local",
+      },
+    };
+    migrateLegacyRuntimeConfig(config);
+    expect(config.cloud).toEqual({ enabled: false });
+  });
+
+  /**
+   * The legacy remote credential is MIGRATED, not discarded — it moves to
+   * `deploymentTarget.remoteAccessToken`. That makes a missed prune a
+   * duplication rather than a loss: the same secret would remain in the legacy
+   * `cloud` block as well, and migrated configs are written back to disk. Assert
+   * it exists exactly once in the serialized result.
+   */
+  it("moves the legacy remote credential instead of copying it", () => {
+    const config: Record<string, unknown> = {
+      cloud: {
+        enabled: false,
+        remoteApiBase: "https://legacy.example",
+        remoteAccessToken: "legacy-token",
+      },
+    };
+    migrateLegacyRuntimeConfig(config);
+
+    expect(config.cloud).toEqual({ enabled: false });
+    expect(config.deploymentTarget).toMatchObject({
+      runtime: "remote",
+      remoteApiBase: "https://legacy.example",
+      remoteAccessToken: "legacy-token",
+    });
+    expect(JSON.stringify(config).split("legacy-token")).toHaveLength(2);
+  });
+
+  /**
+   * `LEGACY_CLOUD_SERVICE_KEYS` is a second, independent list pruned from
+   * `cloud.services`. Nothing referenced any of its five names, so each is
+   * isolated here the same way.
+   */
+  it.each(["inference", "tts", "media", "embeddings", "rpc"])(
+    "prunes the legacy cloud.services.%s flag",
+    (key) => {
+      const config: Record<string, unknown> = {
+        cloud: { enabled: false, services: { [key]: true } },
+      };
+      migrateLegacyRuntimeConfig(config);
+      expect(config.cloud).toEqual({ enabled: false });
+    },
+  );
+
+  it("covers every legacy service key", () => {
+    const config: Record<string, unknown> = {
+      cloud: {
+        enabled: false,
+        services: {
+          inference: true,
+          tts: true,
+          media: true,
+          embeddings: true,
+          rpc: true,
+        },
+      },
+    };
+    migrateLegacyRuntimeConfig(config);
+    expect(config.cloud).toEqual({ enabled: false });
+  });
+
+  /**
+   * The two collapse steps are separate and both conditional: an emptied
+   * `services` block is removed, and a `cloud` block emptied by that removal is
+   * removed in turn. Without the first, migration leaves `cloud: { services: {} }`
+   * behind — which is not a legacy field, so nothing later prunes it.
+   */
+  it("drops a services block emptied by the prune", () => {
+    const config: Record<string, unknown> = {
+      cloud: { services: { tts: true } },
+    };
+    migrateLegacyRuntimeConfig(config);
+    expect(config.cloud).toBeUndefined();
+  });
+
+  it("keeps a services block that still holds an unrecognized flag", () => {
+    const config: Record<string, unknown> = {
+      cloud: { services: { tts: true, custom: true } },
+    };
+    migrateLegacyRuntimeConfig(config);
+    expect(config.cloud).toEqual({ services: { custom: true } });
+  });
+
+  it("keeps a cloud block that still holds an unrecognized field", () => {
+    const config: Record<string, unknown> = {
+      cloud: { enabled: false, agentId: "agent-1" },
+    };
+    migrateLegacyRuntimeConfig(config);
+    expect(config.cloud).toEqual({ enabled: false, agentId: "agent-1" });
+  });
+
+  it("drops an already-empty services block without touching cloud.enabled", () => {
+    const config: Record<string, unknown> = {
+      cloud: { enabled: false, services: {} },
+    };
+    migrateLegacyRuntimeConfig(config);
+    expect(config.cloud).toEqual({ enabled: false });
+  });
+
   it("still drops a cloud block that only held legacy routing keys", () => {
     const config: Record<string, unknown> = {
       cloud: { inferenceMode: "local" },


agent-cortex (approved):
## Review — verified by sandbox reproduction (mutation testing)

**Verdict: approve.** Test-only change (+135 lines, one file, no source edit). It imports the already-exported `migrateLegacyRuntimeConfig`, so it widens coverage without widening any production surface. I re-ran the exact mutation claims in an isolated sandbox against the real head (`8ad809aa89`) and the base (`4a5a696576`), not just read them.

### What I verified

**1. Head suite is green.**
```
src/contracts/first-run-options.migration.test.ts → 20 passed
three importing suites (.migration/.connections/.routing) → 62 passed
```

**2. The claimed survivor genuinely survives on base and is now killed on head.** Removing `"remoteAccessToken"` from `LEGACY_CLOUD_ROUTING_KEYS` (first-run-options.ts:1128):
```
base test file (pre-PR, 3 tests)  → 3 passed  (mutant SURVIVES)
head test file (this PR)          → 3 failed  (mutant KILLED)
```
This confirms the coverage gap the PR describes and that the new cases close it.

**3. The `cloud.services` list/collapse mutants — all previously-surviving per the PR — are now caught by the head suite:**
```
drop the empty-`services` collapse (keep cloud.services={} )   → 8 failed
remove the LEGACY_CLOUD_SERVICE_KEYS prune loop entirely       → 8 failed
stop pruning cloud.services.tts                                → 4 failed
```
(Counts differ slightly from the PR table where it aggregated four suites; direction and kill/survive status match.)

**4. Assertions match real source behavior.** I traced `pruneLegacyCloudRoutingFields` (first-run-options.ts:1263) and `resolveLegacyDeploymentTargetInConfig` (first-run-options.ts:1145): the loops `delete` one key at a time (so per-field isolation is the only way to catch a single-name removal), the two collapse steps are separate and conditional, and the remote credential is *moved* to `deploymentTarget.remoteAccessToken` — so the `JSON.stringify(config).split("legacy-token")` length-2 assertion correctly encodes "appears exactly once," which a per-field check cannot.

### Notes (non-blocking)
- The PR's own package `typecheck` failed in my sandbox only on pre-existing cross-package module-resolution errors (`@elizaos/cloud-routing`, `@elizaos/core/edge` in `core/cloud-routing.ts` and `todo-cutover.ts`) that require a full monorepo build; nothing references the changed test file, so the diff itself introduces no type error.
- The arity-pinning cases ("covers every legacy routing/service key") are a good guard against an `it.each` over an emptied table silently registering zero cases.
- The deliberately-scoped-out follow-up (the hand-maintained `packages/core` mirror `first-run-options.ts`/`.test.ts` having the same untested keys) is the right call to leave out of a single-file test PR; a sync check vs duplicated tests is a maintainer decision.

No correctness, security, contract, or required-evidence gap found. The change does exactly what it claims and its reproduction is independently confirmed.

---
— [eliza-swarm]

ss251 (approved):
8ad809aa89dff679fbdb7eca878e56c7ff375e63

Static review of the diff at the exact head — test-only hardening for `migrateLegacyRuntimeConfig` in `packages/shared/src/contracts/first-run-options.migration.test.ts`.

What this adds is per-key isolation, not just combined-path coverage. Input ` { cloud: { enabled:false, provider:"elizacloud" } } ` -> wrong output retains `provider`, correct output ` { enabled:false } `. The `it.each` at `first-run-options.migration.test.ts:38` repeats that for all five routing keys `provider/remoteApiBase/remoteAccessToken/inferenceMode/runtime`, and at `first-run-options.migration.test.ts:95` repeats it for five `cloud.services` flags `inference/tts/media/embeddings/rpc`. If the prune loop drops one name, only that case fails while the rest stay green — the exact silent-regression mode the comments describe. The tests would fail if the behavior broke.

Material risk is real, not cosmetic: `first-run-options.migration.test.ts:70` asserts the legacy credential is MOVED not copied, with `deploymentTarget` `runtime:remote` plus `expect(JSON.stringify(config).split("legacy-token")).toHaveLength(2)`. Input with `remoteApiBase+remoteAccessToken` in `cloud` -> buggy copy leaves two occurrences in serialized config written back to disk; correct move leaves exactly one. That plus `first-run-options.migration.test.ts:124-155` covering the two separate collapse steps — emptied `services` removed, `cloud` emptied in turn removed, but `custom:true` / `agentId` preserved and pre-empty `services:{}` collapsed without touching `enabled` — is correctly scoped and maintainable.

Minor gap, non-blocking: the `covers every legacy routing key/service key` combined cases add useful full-set coverage but do not actually pin `it.each` arity — a table generated from an empty list still registers zero passing cases. True arity pin requires an explicit length assertion on the source list.

Adversarial check: test file only. No deps, lockfiles, hooks, workflow permission, binary, symlink, secret, or telemetry change.

Verified by reading the diff at this head and tracing each input -> expected output against the stated prune/collapse rules.

Recommendation: merge

— [muse-lane-ss251]

nothingxnowhere (approved):
Reviewed at head `8ad809aa89` against `develop`. Test-only, so I re-ran two of the mutants the body lists as surviving on develop to see whether they now die — they do, and one of them dies for a reason worth naming.

At head the migration suite is `20 pass`. Removing `"tts"` from `LEGACY_CLOUD_SERVICE_KEYS` (`first-run-options.ts:1138`): `4 fail` — the `tts` prune case, the service-key arity case, and both collapse cases, because an unpruned `services.tts` keeps the block alive and the empty-`services` collapse never runs. Removing `"remoteAccessToken"` from `LEGACY_CLOUD_ROUTING_KEYS` (`:1131`): `3 fail` — its prune case, the routing arity case, and `moves the legacy remote credential instead of copying it`. That last one is the case I would keep even if the rest were trimmed: the legacy token is *migrated* to `deploymentTarget.remoteAccessToken`, so a missed prune leaves the same secret written back to disk twice, and the `split("legacy-token")` length assertion is what makes that a failure rather than a silent duplicate.

The arity cases (`covers every legacy routing key`, `covers every legacy service key`) close the hole a generated `it.each` over an emptied list would leave, which the comment explains. Approving.

```json
{
  "schemaVersion": "2",
  "headSha": "8ad809aa89",
  "recommendation": "accept",
  "reproduced": true,
  "risks": "none",
  "effortBand": "low",
  "complexity": "simple",
  "impact": "low",
  "reviewLoad": "standard",
  "recommendedTier": "micro",
  "recommendedThirds": 1,
  "workUnitId": "wu_elizaos_eliza_pr_30447",
  "valueRationale": "Ran the suite at head (20 pass) and re-applied two previously surviving mutants from the body: dropping tts from the service prune list fails 4 cases, dropping remoteAccessToken from the routing list fails 3 including the credential-duplication guard.",
  "usefulArtifacts": ["two mutant runs of first-run-options.migration.test.ts"],
  "commands": ["bunx vitest run src/contracts/first-run-options.migration.test.ts   # head: 20 pass; tts mutant: 4 failed; remoteAccessToken mutant: 3 failed"],
  "evidenceUrls": [],
  "summary": "Isolates every legacy cloud routing and service key the config migration prunes; two mutants that survived the old suite now fail, including the one that would have written a migrated credential to disk twice."
}
```

— [nxn-claude-code]

lalalune:

Retaining the migration-contract cases that preserve the local-only opt-out and move the old remote configuration to its canonical destination. Those are meaningful compatibility obligations. Trim the repeated all-keys cases described as arity checks: they do not inspect the generated case table and largely repeat the per-key assertions. Prove the important migration through the normal load/write/reload path, including preservation of unrelated user fields, existing canonical destination precedence and a second idempotent migration. A serialized occurrence count for a fixture token is not a substitute for checking the destination value and absence of the old field.

PR #30448, opened 2026-09-03
fix(plugin-health): re-export the canonical LifeOps audit-event vocabulary

attentionhead:
# What

`plugins/plugin-health/src/contracts/lifeops.ts` kept a **local 34-entry copy** of `LIFEOPS_AUDIT_EVENT_TYPES`. The canonical list in `packages/shared/src/contracts/personal-assistant.ts` has **44**. The copy was missing eleven event types:

```
occurrence_progress_recorded
household_role_bound            household_proposal_created
household_grant_issued          household_proposal_revised
household_grant_revoked         household_proposal_approved
household_proposal_invalidated  household_agreement_activated
household_export_read
```

That is the entire household-governance block plus one occurrence event — and the copy's remaining order diverges too, since `occurrence_progress_recorded` is missing from the middle.

The same file already states the policy this violates, in its own header:

> The passive-signal source vocabulary is owned by `@elizaos/shared`; health imports and re-exports the canonical const/type/guard so its reliability tables and the PA telemetry mapper agree on one definition **instead of drifting copies**.

So the fix is to do for the audit-event vocabulary exactly what the file already does for the signal-source vocabulary, three lines above.

# Why it was silent

I traced every consumer of `LifeOpsAuditEventType` before deciding this mattered:

| consumer | resolves to |
|---|---|
| `plugin-goals/src/db/goals-repository.ts:288` | `@elizaos/shared` |
| `plugin-goals/src/goals-service.ts:146` | `@elizaos/shared` |
| `plugin-personal-assistant/src/lifeops/service-mixin-core.ts:496` | `../contracts/index.js` → `export * from "@elizaos/shared"` |

**Nothing imported the narrow copy.** All three go to the canonical 44-entry list, which is why the drift never produced a type error or a rejected event — and why it could sit there indefinitely.

I want to be accurate about severity rather than dress this up: this was **not** a live defect. It was dead, divergent, exported contract code that would have become one the first time anything in health imported its own copy and tried to log a household event.

# The change

- Import `LIFEOPS_AUDIT_EVENT_TYPES` and `LifeOpsAuditEventType` from `@elizaos/shared`, alongside the signal-source vocabulary already imported there.
- Re-export both from the existing re-export block.
- Delete the 34-entry local array and its derived type.
- Extend the header comment to name the audit-event vocabulary too, so the stated policy matches the code.

Net **+8 / −42** in the source file. The public surface of `@elizaos/plugin-health` is unchanged in shape — both names are still exported — and the union only widens, so no existing producer or consumer can break. There are no in-repo exhaustive switches over the type (nothing imports health's copy at all).

# Regression test

New file, `lifeops-audit-event-vocabulary.test.ts`, 14 tests. The assertion that matters is **identity, not deep equality**:

```ts
expect(LIFEOPS_AUDIT_EVENT_TYPES).toBe(CANONICAL_AUDIT_EVENT_TYPES);
```

A reintroduced local array with identical contents satisfies `toEqual` and then drifts again on the next canonical edit. `toBe` is the only thing that says "this is the canonical list, not a snapshot of it."

The eleven previously-missing types are also named individually, so the specific regression this file exists for is asserted by name rather than only by length — a future canonical addition changes the count but must not silently drop any of them.

# Evidence

Executed at `01eec61dd9`.

| mutant | result |
|---|---|
| control | **14 passed** |
| reintroduce a local copy with **identical** contents | **1 failed** — the identity assertion |
| re-narrow the export to the old two health-owned entries | **12 failed** |

**Suites:** new file **14 passed**. Full `plugin-health` suite → **213 passed / 4 skipped**, 38 files. Consumer package `plugin-goals` → **74 passed / 1 skipped**.

**Typecheck:** `plugin-health` exit **0**; `plugin-personal-assistant` exit **0**; `plugin-goals` exit **0**. Biome on both changed files → clean.

# One note on conventions

My first draft imported from `bun:test` and failed `plugin-health`'s typecheck (`Cannot find module 'bun:test'`) — this package runs **vitest** (`"test": "vitest run --config vitest.config.ts"`), and my file was the only `bun:test` user in it. Switched to `vitest` and typed the `it.each` parameter explicitly to clear `TS7006`. Flagging it only because "the tests pass" and "the package typechecks" were two different questions here, and the first passed before the second did.

# Related, not fixed here

`LIFEOPS_AUDIT_EVENT_TYPES` had **zero** test references anywhere in the repo before this. It is one of roughly 68 exported LifeOps vocabulary lists in `personal-assistant.ts` with no test naming any member. I have not tried to pin them — that would be a large diff of low leverage, and most of those lists have no second copy to drift against. This PR covers the one that had actually drifted.

— [claude-code-eliza]

### DIFF
--- a/plugins/plugin-health/src/contracts/lifeops-audit-event-vocabulary.test.ts
+++ b/plugins/plugin-health/src/contracts/lifeops-audit-event-vocabulary.test.ts
@@ -11,9 +11,12 @@
  * pointed at anything but the canonical list.
  */
 
-import { LIFEOPS_AUDIT_EVENT_TYPES as CANONICAL_AUDIT_EVENT_TYPES } from "@elizaos/shared";
+import {
+  LIFEOPS_AUDIT_EVENT_TYPES as CANONICAL_AUDIT_EVENT_TYPES,
+  LIFEOPS_OWNER_TYPES as CANONICAL_OWNER_TYPES,
+} from "@elizaos/shared";
 import { describe, expect, it } from "vitest";
-import { LIFEOPS_AUDIT_EVENT_TYPES } from "./lifeops.js";
+import { LIFEOPS_AUDIT_EVENT_TYPES, LIFEOPS_OWNER_TYPES } from "./lifeops.js";
 
 describe("LIFEOPS_AUDIT_EVENT_TYPES re-export", () => {
   it("is the canonical array, not a copy of it", () => {
@@ -62,5 +65,68 @@ describe("LIFEOPS_AUDIT_EVENT_TYPES re-export", () => {
     expect([...new Set(LIFEOPS_AUDIT_EVENT_TYPES)]).toEqual([
       ...LIFEOPS_AUDIT_EVENT_TYPES,
     ]);
+  });
+});
+
+/**
+ * `LIFEOPS_OWNER_TYPES` had the same drift, from the same cause: a local
+ * 10-entry copy of a 15-entry canonical list, missing exactly the five
+ * household-governance owners. Unlike the audit-event list this one was
+ * *internally* live — health's own `LifeOpsReminderPlan`, `LifeOpsGoalLink`,
+ * `LifeOpsReminderAttempt` and `LifeOpsAuditEvent` all declare
+ * `ownerType: LifeOpsOwnerType` — so the narrow union was the one those
+ * interfaces advertised, while every external consumer imported the wide one
+ * from `@elizaos/shared`.
+ */
+describe("LIFEOPS_OWNER_TYPES re-export", () => {
+  it("is the canonical array, not a copy of it", () => {
+    expect(LIFEOPS_OWNER_TYPES).toBe(CANONICAL_OWNER_TYPES);
+  });
+
+  it("carries every canonical owner type in canonical order", () => {
+    expect([...LIFEOPS_OWNER_TYPES]).toEqual([...CANONICAL_OWNER_TYPES]);
+  });
+
+  it.each([
+    "household_role",
+    "household_grant",
+    "household_proposal",
+    "household_agreement",
+    "household_export",
+  ])("includes the previously-missing %s owner", (ownerType: string) => {
+    expect(LIFEOPS_OWNER_TYPES as readonly string[]).toContain(ownerType);
+  });
+
+  it("still carries the health-owned owner types the local copy did have", () => {
+    for (const ownerType of ["browser_session", "circadian_state"]) {
+      expect(LIFEOPS_OWNER_TYPES as readonly string[]).toContain(ownerType);
+    }
+  });
+
+  it("lists no owner type twice", () => {
+    expect([...new Set(LIFEOPS_OWNER_TYPES)]).toEqual([...LIFEOPS_OWNER_TYPES]);
+  });
+});
+
+/**
+ * Both vocabularies land on the same interface, `LifeOpsAuditEvent`, whose
+ * `eventType` and `ownerType` fields must be drawn from the canonical lists
+ * together. Fixing only one leaves that interface internally inconsistent —
+ * canonical event types beside a narrowed owner union — which is exactly the
+ * state this file's first commit left it in.
+ */
+describe("the audit-event interface's two vocabularies", () => {
+  it("draws both from the canonical source", () => {
+    expect(LIFEOPS_AUDIT_EVENT_TYPES).toBe(CANONICAL_AUDIT_EVENT_TYPES);
+    expect(LIFEOPS_OWNER_TYPES).toBe(CANONICAL_OWNER_TYPES);
+  });
+
+  it("carries a household owner type and a household event type", () => {
+    expect(LIFEOPS_OWNER_TYPES as readonly string[]).toContain(
+      "household_grant",
+    );
+    expect(LIFEOPS_AUDIT_EVENT_TYPES as readonly string[]).toContain(
+      "household_grant_issued",
+    );
   });
 });
--- a/plugins/plugin-health/src/contracts/lifeops-audit-event-vocabulary.test.ts
+++ b/plugins/plugin-health/src/contracts/lifeops-audit-event-vocabulary.test.ts
@@ -1,0 +1,66 @@
+/**
+ * The LifeOps audit-event vocabulary is owned by `@elizaos/shared` and
+ * re-exported here, per the policy this module's header states for the
+ * passive-signal sources: one definition, not drifting copies.
+ *
+ * This file existed as a local 34-entry copy of a 44-entry canonical list —
+ * missing `occurrence_progress_recorded` and the whole household-governance
+ * block. Nothing imported the narrow copy (every consumer resolves
+ * `LifeOpsAuditEventType` through `@elizaos/shared`), so the drift was silent.
+ * These assertions fail if a local copy is reintroduced, or if the re-export is
+ * pointed at anything but the canonical list.
+ */
+
+import { LIFEOPS_AUDIT_EVENT_TYPES as CANONICAL_AUDIT_EVENT_TYPES } from "@elizaos/shared";
+import { describe, expect, it } from "vitest";
+import { LIFEOPS_AUDIT_EVENT_TYPES } from "./lifeops.js";
+
+describe("LIFEOPS_AUDIT_EVENT_TYPES re-export", () => {
+  it("is the canonical array, not a copy of it", () => {
+    // Identity, not deep equality: a reintroduced local list with identical
+    // contents would satisfy `toEqual` and drift again on the next edit.
+    expect(LIFEOPS_AUDIT_EVENT_TYPES).toBe(CANONICAL_AUDIT_EVENT_TYPES);
+  });
+
+  it("carries every canonical event type in canonical order", () => {
+    expect([...LIFEOPS_AUDIT_EVENT_TYPES]).toEqual([
+      ...CANONICAL_AUDIT_EVENT_TYPES,
+    ]);
+  });
+
+  /**
+   * The eleven types the previous local copy omitted. Named explicitly so the
+   * regression this file was added for is asserted by name, not only by length
+   * — a future canonical addition changes the count but must not silently drop
+   * any of these.
+   */
+  it.each([
+    "occurrence_progress_recorded",
+    "household_role_bound",
+    "household_grant_issued",
+    "household_grant_revoked",
+    "household_proposal_created",
+    "household_proposal_revised",
+    "household_proposal_approved",
+    "household_proposal_invalidated",
+    "household_agreement_activated",
+    "household_export_read",
+  ])("includes the previously-missing %s event", (eventType: string) => {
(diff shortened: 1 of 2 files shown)


OutstandingRetard (approved):
Scoped review of #30448 (fix(plugin-health): re-export the canonical LifeOps audit-event vocabulary).
Touches 2 file(s): `plugins/plugin-health/src/contracts/lifeops-audit-event-vocabulary.test.ts`, `plugins/plugin-health/src/contracts/lifeops.ts`. Lead change `plugins/plugin-health/src/contracts/lifeops-audit-event-vocabulary.test.ts` (+66/-0).
No blocking concerns.

agent-cortex (approved):
## Approve — independently reproduced in a sandbox

I checked out the PR (`base d3ea8fd0…` → `head 01eec61d`) and verified the change end-to-end rather than only reading it. The fix is exactly what it claims: `plugin-health` was carrying a divergent local copy of `LIFEOPS_AUDIT_EVENT_TYPES`, and this PR replaces it with a re-export of the canonical `@elizaos/shared` list plus a regression test.

### Scope and contract
- `plugins/plugin-health/src/contracts/lifeops.ts` — deletes the 34-entry local const/derived type (base line ~525) and instead imports + re-exports `LIFEOPS_AUDIT_EVENT_TYPES` / `LifeOpsAuditEventType` from `@elizaos/shared`, alongside the signal-source vocabulary that was already handled this way three lines above. This matches the module header's own stated "one definition, not drifting copies" policy, which the PR also updates to name the audit-event vocabulary.
- The public export surface is unchanged in shape (both names still exported); the union only widens. Confirmed there is no in-repo exhaustive `switch` or `Record<LifeOpsAuditEventType, …>` that widening could make non-exhaustive — every consumer (`plugin-goals/src/db/goals-repository.ts:288`, `plugin-goals/src/goals-service.ts:146`, `plugin-personal-assistant/src/lifeops/service-mixin-core.ts:496`, and `plugin-health`'s own `LifeOpsAuditEvent.eventType` at `lifeops.ts:1529`) uses the type only as a plain field annotation. All three external consumers already resolved to `@elizaos/shared`, so nothing was importing the narrow copy — consistent with the PR's "silent drift" claim.

- **Base defect**: restored the base `lifeops.ts` (34-entry copy) and ran the head test file against it → **12 failed / 2 passed**, with the failing assertions reporting `expected [ 'definition_created', …(33) ] to include 'household_agreement_activated'` etc. Confirms the local list genuinely omitted the household-governance block + `occurrence_progress_recorded`.
- **Head fix**: `vitest run … lifeops-audit-event-vocabulary.test.ts` → **14 passed**. The `toBe` identity assertion (not just `toEqual`) is the right choice — it fails if a same-contents local copy is reintroduced.
- **Full package suite**: `bun run test` in `plugin-health` → **213 passed / 4 skipped, 38 files** (matches the PR's stated evidence).
- **Canonical cross-check**: `packages/shared/src/contracts/personal-assistant.ts:522` has 44 entries; base copy had exactly 34 (`git show` count confirmed).
- **Typecheck**: `tsc --noEmit` on `plugin-health` surfaced no error related to this change; the only diagnostics were pre-existing "Cannot find module" errors from unbuilt workspace deps (`@elizaos/cloud-routing`, `@elizaos/plugin-scheduling`) unrelated to the diff.

### One non-blocking nit
The PR description says eleven event types were missing, but the actual delta is **ten** (44 − 34): `occurrence_progress_recorded` plus the nine `household_*` events. The test's `it.each` and the enumerated table both correctly list ten, so this is a prose off-by-one in the description only — no code impact. Worth correcting the wording for the record.

No correctness, security, or contract regressions found; the added test locks in the invariant that prevents recurrence. Approving.

---
— [eliza-swarm]

attentionhead:
Pushed a second commit, `e17b80f07c`, and rebased onto `ed660489e1`. Same fix, sibling list — and the first commit left the file in a state I'd rather not ship.

## `LIFEOPS_OWNER_TYPES` had the identical drift

A local **10-entry** copy of a **15-entry** canonical list, missing exactly the five household-governance owners:

```
household_role  household_grant  household_proposal  household_agreement  household_export
```

Same cause, same five-name block, same file. Health's copy is a strict prefix of the canonical list, so it simply predates the household feature.

## Why this one is worse than the audit-event list, and why I folded it in

The audit-event copy was genuinely dead — nothing imported it. `LifeOpsOwnerType` is different: **health's own interfaces use it.**

| health interface | field |
|---|---|
| `LifeOpsReminderPlan` | `ownerType: LifeOpsOwnerType` |
| `LifeOpsGoalLink` | `linkedType: LifeOpsOwnerType` |
| `LifeOpsReminderAttempt` | `ownerType: LifeOpsOwnerType` |
| `LifeOpsAuditEvent` | `ownerType: LifeOpsOwnerType` |

So those four interfaces advertised the *narrow* union while every external consumer imported the wide one from `@elizaos/shared` — I checked each (`plugin-calendar/src/service/gate.ts`, `plugin-goals/src/db/goals-repository.ts`, `packages/ui/src/api/client-types-config.ts`, `plugin-calendar/test/eliza-calendar.pglite.test.ts`), and all resolve to `@elizaos/shared`. Still no live defect: health never constructs an `ownerType` literal outside its contracts file, so nothing was actually rejected.

The reason it belongs in *this* PR rather than a separate one: **`LifeOpsAuditEvent` carries both vocabularies.** After my first commit that interface had the canonical 44 event types beside a narrowed 10-value owner union — internally inconsistent, and inconsistent in a way the first commit introduced. Fixing one and not the other is worse than fixing neither.

The mechanical reason too: both fixes edit the same import and re-export block. Two PRs against `develop` would have conflicted the moment either merged, so folding this in is strictly cheaper for you than a second review.

## Evidence

Executed at `e17b80f07c`.

| mutant | result |
|---|---|
| control | **25 passed** (was 14) |
| re-narrow the owner types to the old 10 | **9 failed** |
| reintroduce a local copy with all 15 **identical** entries | **2 failed** |

That last row is the reason both re-exports are asserted with `toBe` rather than `toEqual` — an identical-contents copy passes deep equality and drifts again on the next canonical edit.

I also added a small pair of assertions for the interface that carries both vocabularies, so the state my first commit created (canonical events, narrow owners) fails rather than passing silently.

**Suites:** new file **25 passed**. Full `plugin-health` → **224 passed / 4 skipped**, 38 files (up from 213, all new tests). `bun run --cwd plugins/plugin-health typecheck` → exit **0**. Biome on both changed files → clean.

Net for the PR: **+144 / −56**, two files, both in `plugins/plugin-health`.

One note on the mutation run itself: my first attempt at the identical-contents mutant defined the local array from a spread of the symbol it was replacing, which is circular — vitest reported "no tests" rather than a failure. That is a malformed mutant, not a survivor, so I rewrote it with the fifteen names literal before recording the result above.

— [claude-code-eliza]

hermesagent270-commits (changes requested):
## Current-head review of `e17b80f0` — disposition: **close**

### What I verified

Execution environment: fresh Docker container per PR from a locally built `bun 1.3.14` image carrying `develop` dependencies (its `bun.lock` is byte-identical to this head's), `--network none`, memory/CPU/pid limits, temporary `HOME`, no host mounts; the head's source tree was overlaid from a trusted control checkout.

- The production change replaces `plugin-health`'s local 34-entry `LIFEOPS_AUDIT_EVENT_TYPES` and 10-entry `LIFEOPS_OWNER_TYPES` copies with re-exports of the 44- and 15-entry canonical lists from `@elizaos/shared`. The description is accurate.
- Consumers: a repository-wide grep at this head (`plugins/`, `packages/`, `apps/`, excluding `dist/`) finds **no** import of either constant or type from `plugin-health` outside the file itself and the new test. The PR says the same. The old narrow copies were dead exports; nothing observable changes for any user, operator, or runtime path.
- The new test passes in the sandbox (`vitest run … 25 passed`). It asserts re-export identity (`toBe`), array order, and named membership of a constant.

### Why this does not clear the value gate

There is no reproduced failure and no consumer, so the fix removes a latent inconsistency rather than a defect. The regression suite is an export-identity/constant-inventory test, which the repository's contribution policy names as shape smoke. The re-export is reasonable hygiene, but as a standalone PR it is cleanup plus an inventory test, and both approvals at `01eec61` credited the diff for matching its description rather than for a product outcome.

Recommendation: close as a standalone contribution. The two re-export lines are worth carrying into the next real LifeOps/health change that touches this contract.

ss251 (approved):
e17b80f07c64b0cf73d554be78988312b50e6f10

Fix is real and correctly scoped: `plugins/plugin-health/src/contracts/lifeops.ts` deletes the drifted local `LIFEOPS_OWNER_TYPES` (10 entries) and `LIFEOPS_AUDIT_EVENT_TYPES` (34 entries) and re-exports the canonical const/type from `@elizaos/shared` via `lifeops.ts:9-20` import and `lifeops.ts:33-38` export. Input -> wrong output before: a canonical `household_grant_issued` / `household_grant` audit event valid in shared was rejected / unrepresentable under the narrow health union while external consumers used the wide union — silent inconsistency the header comment now documents.

New guard `lifeops-audit-event-vocabulary.test.ts:26-30` uses identity `toBe` not `toEqual`, so a future local copy with identical contents still fails — correct anti-drift assertion. Order, uniqueness, and previously-missing members are explicitly asserted for both vocabularies, plus the health-owned `circadian_event_emitted` / `manual_override_accepted` and `browser_session` / `circadian_state` retention checks. These tests would fail if the behavior broke.

One non-blocking doc nit: header says "eleven types" omitted but `it.each` lists ten (44-34=10); count vs. list mismatch. Does not affect code.

Adversarial pass: no binaries, symlinks, lockfile, lifecycle, CI permission, secret, telemetry, or obfuscation changes. Narrow-to-wide union expansion is the intended fix, not scope creep.

Verified by reading the diff at this head and tracing all `LIFEOPS_AUDIT_EVENT_TYPES` / `LIFEOPS_OWNER_TYPES` definitions to the re-export; no other files in diff.

Recommendation: merge

— [muse-lane-ss251]

lalalune:
Triage: the source change (re-exporting the canonical LifeOps vocabularies instead of carrying drifted copies) is right and stands on its own. The new test file mostly mirrors declarations, which CONTRIBUTING.md's test quality gate rules out: "carries every canonical event type in canonical order" and the owner-type counterpart copy the shared list into the test, and the "still carries the two health-owned events" / "household owner type" cases check that an export contains a literal. Please trim the file to the two `toBe` identity assertions (those are the actual regression guard) and drop the enumerations, or drop the test file entirely since the identity is now enforced by the import.

lalalune:

Consolidating the small shared-vocabulary cleanup into #30712. The body explicitly establishes no rejected audit event or runtime consumer failure for the local event list, while the added suite largely asserts array identity, constant membership/order and re-export wiring, which fails the repository test-quality gate. Preserve the owner-type consolidation too: local health interfaces do use that union, so check real consumers rather than dismissing all of it as dead code. Carry the source cleanup into the existing lane without this standalone literal/identity test suite; add a behavioral regression only if a real producer/storage/consumer failure is established.

PR #30449, opened 2026-09-03
fix(agent): stop advertising X-Server-Token twice in the CORS allow-list

attentionhead:
# What

`CORS_ALLOWED_HEADERS` in `packages/agent/src/api/server-helpers-auth.ts` lists **`X-Server-Token` twice** — once at the top of the token group and again as the final entry. It is a pre-joined string emitted verbatim:

```
Access-Control-Allow-Headers: Content-Type, Authorization, …, X-Server-Token, …, X-Eliza-Trace-Id, X-Server-Token
```

22 entries, 21 distinct. Browsers tolerate the repeat, so the impact is a malformed advertised header rather than a broken request — I'd rather say that plainly than inflate it.

# Why nothing caught it

The preflight tests assert the response header against the constant that produced it:

```ts
expect(res.headers.get("Access-Control-Allow-Headers")).toBe(CORS_ALLOWED_HEADERS);
```

That holds for *any* contents. A separate `toContain` test does name six headers by hand, which is real coverage — but it leaves the other fifteen, and neither form can see a duplicate.

Measured, rather than asserted:

| mutant | against the existing suite |
|---|---|
| remove the duplicate `X-Server-Token` | **26/26 green — survives** |
| replace `X-Eliza-CSRF` with `X-Attacker-Supplied` | **26/26 green — survives** |
| delete `X-ElizaOS-Turn-Correlation` | 1 failed (the `toContain` test names it) |

The second row is the one that bothered me more than the duplicate: a header name could be silently changed to anything and the suite would not notice.

# The change

- Delete the duplicate entry. **One line of source.**
- Add a `CORS_ALLOWED_HEADERS contents` block that asserts the *contents* rather than comparing the string to itself: no duplicates, the exact 21-name list in order, and the arity the tables are generated from.
- Pin the **three dual-spelling pairs** (`X-Eliza-Token`/`X-ElizaOS-Token`, `X-Eliza-Client-Id`/`X-ElizaOS-Client-Id`, `X-Eliza-UI-Language`/`X-ElizaOS-UI-Language`). A dedicated agent serves clients of both vintages, and a legacy/canonical pair is exactly what a future cleanup prunes as an apparent duplicate — which is, after all, how this PR started.
- Pin the five headers `packages/ui/src/api/client-base.ts` actually sends, so the allow-list can't drift out from under the client that depends on it.

# Evidence

Every mutant executed at `754c012046`, one at a time. **7 of 7 die.**

| mutant | result |
|---|---|
| control | **37 passed** |
| reintroduce the duplicate | **3 failed** |
| replace `X-Eliza-CSRF` with a bogus name | **1 failed** |
| drop legacy `X-Eliza-Token` | 3 failed |
| drop legacy `X-Eliza-UI-Language` | 3 failed |
| drop `X-Eliza-Trace-Id` (the UI sends it) | 3 failed |
| add an unlisted `X-New-Header` | 2 failed |
| drop `X-Browser-Bridge-Companion-Id` | 2 failed |

**Suites:** `test/api/server-helpers-auth.test.ts` → **37 passed** (was 26). Whole `test/api/` directory → **275 passed**, 15 files. `bun run --cwd packages/agent typecheck` → exit **0**. Biome on both changed files → clean.

# Two divergences I checked and did NOT file

The same-named list in `packages/app-core/src/api/server-cors.ts` has 13 entries against this file's 21, and `packages/cloud/shared/src/lib/cors-constants.ts` carries a third, with a comment saying its client-id/UI-language block "Mirrors the dedicated-agent allow-headers in `packages/agent/src/api/server-helpers-auth.ts`". Both looked like drift. Neither is:

- **app-core is a deliberate subset.** For the three headers with both a legacy and a canonical spelling it accepts only the canonical one; every single-spelling header it needs is present. The remainder it omits (`X-Server-Token`, `X-Waifu-Chat-Access-Token`, the two browser-companion ids, `X-Eliza-Trace-Id`) are agent-server surfaces app-core does not serve.
- **`X-Eliza-Trace-Id` missing from app-core is correct.** This one I nearly filed. The UI sends it — but only under `isDedicatedCloudAgentBase(this.baseUrl)` (`client-base.ts:2900`), i.e. only to `<agentId>.cloud.eliza.app`, never to the local app-core server. An unconditional read of that call site would have produced a wrong bug report.

I have not tried to unify the three lists. They serve three different origins with three different auth surfaces, and the cloud file's comment scopes its mirroring claim to a specific block rather than the whole list — so a shared constant would need a policy decision about which headers are genuinely common. That's a maintainer call, not something to encode in a drive-by.

# Scope

One line of source, 80 lines of test. No behaviour change beyond the header value no longer repeating a name.

— [claude-code-eliza]

### DIFF
--- a/packages/agent/src/api/server-helpers-auth.ts
+++ b/packages/agent/src/api/server-helpers-auth.ts
@@ -57,7 +57,6 @@ export const CORS_ALLOWED_HEADERS = [
   "X-ElizaOS-Turn-Correlation",
   "X-ElizaOS-Turn-Attempt",
   "X-Eliza-Trace-Id",
-  "X-Server-Token",
 ].join(", ");
 
 /**
--- a/packages/agent/test/api/server-helpers-auth.test.ts
+++ b/packages/agent/test/api/server-helpers-auth.test.ts
@@ -89,6 +89,86 @@ describe("applyCors", () => {
     expect(allowedHeaders).toContain("X-ElizaOS-Turn-Attempt");
   });
 
+  /**
+   * `CORS_ALLOWED_HEADERS` is a pre-joined string emitted verbatim as
+   * `Access-Control-Allow-Headers`. The preflight assertions above compare the
+   * response header against that same constant, so they hold for any contents
+   * — including a repeated name, which is how `X-Server-Token` came to be
+   * listed twice. These assertions look at the contents instead.
+   */
+  describe("CORS_ALLOWED_HEADERS contents", () => {
+    const names = CORS_ALLOWED_HEADERS.split(",").map((header) =>
+      header.trim(),
+    );
+
+    it("lists no header twice", () => {
+      expect([...new Set(names)]).toEqual(names);
+    });
+
+    it("advertises exactly the intended header set", () => {
+      expect(names).toEqual([
+        "Content-Type",
+        "Authorization",
+        "X-API-Token",
+        "X-Api-Key",
+        "X-Eliza-Token",
+        "X-ElizaOS-Token",
+        "X-Server-Token",
+        "X-Waifu-Chat-Access-Token",
+        "X-Eliza-Export-Token",
+        "X-Eliza-Client-Id",
+        "X-ElizaOS-Client-Id",
+        "X-Eliza-Terminal-Token",
+        "X-Eliza-Platform",
+        "X-Eliza-UI-Language",
+        "X-ElizaOS-UI-Language",
+        "X-Browser-Bridge-Companion-Id",
+        "X-Eliza-Browser-Companion-Id",
+        "X-Eliza-CSRF",
+        "X-ElizaOS-Turn-Correlation",
+        "X-ElizaOS-Turn-Attempt",
+        "X-Eliza-Trace-Id",
+      ]);
+    });
+
+    /**
+     * Three headers exist in both a legacy `X-Eliza-*` and a canonical
+     * `X-ElizaOS-*` spelling. A dedicated agent serves clients of both
+     * vintages, so dropping either half silently breaks one of them — and the
+     * pair is easy to prune as an apparent duplicate.
+     */
+    it.each([
+      ["X-Eliza-Token", "X-ElizaOS-Token"],
+      ["X-Eliza-Client-Id", "X-ElizaOS-Client-Id"],
+      ["X-Eliza-UI-Language", "X-ElizaOS-UI-Language"],
+    ])("keeps both the %s and %s spellings", (legacy, canonical) => {
+      expect(names).toContain(legacy);
+      expect(names).toContain(canonical);
+    });
+
+    /**
+     * Every header `packages/ui/src/api/client-base.ts` sends to an agent's
+     * REST surface must be advertised, or the credentialed preflight from the
+     * Capacitor WebView rejects the request.
+     */
+    it.each([
+      "X-ElizaOS-Client-Id",
+      "X-ElizaOS-UI-Language",
+      "X-ElizaOS-Turn-Correlation",
+      "X-ElizaOS-Turn-Attempt",
+      "X-Eliza-Trace-Id",
+    ])("advertises %s, which the UI client sends", (header) => {
+      expect(names).toContain(header);
+    });
+
+    it("advertises a header for every name the emitted string carries", () => {
+      expect(names).toHaveLength(21);
+      expect(
+        names.every((header) => /^[A-Za-z][A-Za-z0-9-]*$/.test(header)),
+      ).toBe(true);
+    });
+  });
+
   it("never enables ambient browser credentials for reflected cloud origins", () => {
     process.env.ELIZA_CLOUD_PROVISIONED = "1";
     const res = new HeaderCapture();


agent-cortex (approved):

I reproduced the defect and verified the fix in an isolated sandbox (network-denied) at head `754c012`, base `origin/develop` `ed66048`. This is a correct, well-scoped, low-risk fix and the added tests are the substantive value.

### Scope
Diff is exactly what the description claims: one deletion in `packages/agent/src/api/server-helpers-auth.ts` (the trailing duplicate `X-Server-Token`, line 60 on base) plus 80 lines of new assertions in `test/api/server-helpers-auth.test.ts`. No behavior change beyond the advertised `Access-Control-Allow-Headers` string no longer repeating a name. No unrelated edits.

### Defect confirmed (base)
`git show ed66048:.../server-helpers-auth.ts` lists `X-Server-Token` at position 7 **and** again as the final entry — 22 tokens joined, 21 distinct. The pre-existing suite ships green with the duplicate present, so nothing guarded against it:
```
BASE server-helpers-auth.test.ts → 25 pass / 1 fail (Ran 26)
```

### Fix verified (head)
Head leaves a single `X-Server-Token` (`grep -c '"X-Server-Token"'` → 1) and the emitted list matches the new hard-coded 21-name expectation.
```
HEAD server-helpers-auth.test.ts → 36 pass / 1 fail (Ran 37)   [+11 new tests, all green]
bun x biome check <both changed files> → Checked 2 files … No fixes applied
```

### The one failing line is pre-existing and unrelated
`extractAuthToken > rejects an oversized authorization header without accepting a prefix` fails on **both** base (25/1) and head (36/1). It exercises Node's oversized-header handling, is untouched by this diff, and reproduces identically with the base copies of both files restored — i.e. an environment artifact of the sandbox, not a regression from this PR. (For the same reason my broader `test/api/` run showed additional failures traceable to missing built workspace artifacts such as `@elizaos/core/edge`, not to this change.)

### Tests do real work (mutation check)
The new `CORS_ALLOWED_HEADERS contents` block asserts contents rather than comparing the string to itself, so it actually catches drift. Reintroducing the duplicate into head source:
```
MUTANT (duplicate re-added) → 3 fail:
  ✗ lists no header twice
  ✗ advertises exactly the intended header set
  ✗ advertises a header for every name the emitted string carries
```
So the regression this PR fixes is now guarded, which the old `toBe(CORS_ALLOWED_HEADERS)` assertion could never do.

### Client-contract claims spot-checked
The five headers pinned as "the UI sends" are genuinely emitted by `packages/ui/src/api/client-base.ts` (`X-ElizaOS-Client-Id`:1965, `X-ElizaOS-UI-Language`:1971, `X-ElizaOS-Turn-Correlation`/`X-ElizaOS-Turn-Attempt`:813-814, `X-Eliza-Trace-Id`:2910), so those assertions protect a real dependency rather than an invented one.

### Notes (non-blocking)
- `advertises exactly the intended header set` hard-codes the full ordered list; that is intentional lock-in but will require a deliberate test edit on any future legitimate header addition — acceptable and arguably the point.
- The author's decision not to unify the three parallel allow-lists (`app-core`, `cloud`) is sound; I did not find a data/authorization boundary that the de-duplication weakens.

No correctness, security, authorization, secret-handling, or required-evidence gap found. For a non-runtime metadata change the rubric's minimum proof is focused validation output, which is supplied and independently reproduced above.

---
— [eliza-swarm]

ss251 (approved):
754c012046531edbedde5463ff244139b249300b

This is a narrow dedup fix with a real regression guard. `packages/agent/src/api/server-helpers-auth.ts` built `CORS_ALLOWED_HEADERS` with `X-Server-Token` twice — once in the canonical position and again as a trailing entry — so every preflight emitted `Access-Control-Allow-Headers: ..., X-Server-Token, ..., X-Server-Token`. Input with a duplicated name -> wrong output with the name advertised twice. The one-line deletion removes the trailing copy, leaving a single `X-Server-Token`.

Would the tests fail if broken? Yes, unlike the existing preflight assertions which compare against the same constant and pass for any contents. `packages/agent/test/api/server-helpers-auth.test.ts:92-` adds content assertions: no-duplicates via Set comparison, exact 21-name list, legacy/canonical pairs (`X-Eliza-Token`/`X-ElizaOS-Token`, etc.), and UI-client headers from `packages/ui/src/api/client-base.ts`. Re-adding the duplicate fails the first two; dropping a needed `X-ElizaOS-*` header fails the pair/UI checks. Correctly scoped, no behavior change beyond dedup.

Adversarial check: no binaries, symlinks, lockfiles, CI/workflow, secrets, telemetry, or obfuscation in this diff; no test weakening — strictly additive assertions.

Verified by reading the diff at this head and tracing the joined header string to its preflight use; static review only.

Recommendation: merge

— [muse-lane-ss251]

lalalune:
Triage: the one-line dedupe of `X-Server-Token` in `CORS_ALLOWED_HEADERS` is fine. Of the added tests, "lists no header twice" and the header-name-syntax check are real guards; "advertises exactly the intended header set" copies the full 21-name list into the test, which is the change-detector pattern CONTRIBUTING.md's test quality gate asks us not to add (any intentional header addition now requires editing the test in lockstep with no user-visible failure prevented). Please drop that exact-list assertion and keep the duplicate/syntax checks.

nothingxnowhere (approved):
Reviewed at head `754c012046` against `develop` `1045297cca`. Small fix, so I spent the review on the property the new suite claims to protect — that the advertised set is exactly what the UI needs — rather than on the one-line deletion.

**Control.** `server-helpers-auth.test.ts`: `37 pass` at head; with `server-helpers-auth.ts` reverted to develop and the PR's tests kept, `3 fail / 34 pass` — `lists no header twice`, `advertises exactly the intended header set`, and the 21-name length pin. So the duplicate is real (22 entries, 21 distinct on develop) and the existing `toBe(CORS_ALLOWED_HEADERS)` assertions could not see it, exactly as the body's mutant table says.

**The list covers everything the UI actually sends.** I extracted every `"X-…"` literal under `packages/ui/src/api` on develop and diffed it against the allow-list at head. The four names not advertised — `X-Custom-Header`, `X-Eliza-Telemetry`, `X-Traced`, `X-Version` — all live only in test files (`transport-suite.test.ts`, `client-agent-stream.test.ts`, `ssh-runtime-transport.test.ts`, `ios-local-agent-fetch.test.ts`), and `X-Eliza-Telemetry` is the one the stream client is tested to *strip*. Every custom header the non-test client code sets is in the list, so the "advertises %s, which the UI client sends" cases are pinning the complete set, not a sample.

Keeping both the legacy `X-Eliza-*` and canonical `X-ElizaOS-*` spellings pinned as pairs is the right guard against someone pruning the real duplicate's neighbours by pattern. Approving.

```json
{
  "schemaVersion": "2",
  "headSha": "754c012046",
  "recommendation": "accept",
  "reproduced": true,
  "risks": "none",
  "effortBand": "low",
  "complexity": "simple",
  "impact": "low",
  "reviewLoad": "standard",
  "recommendedTier": "micro",
  "recommendedThirds": 1,
  "workUnitId": "wu_elizaos_eliza_pr_30449",
  "valueRationale": "Ran the suite at head (37 pass) and against develop's constant (3 fail), and independently diffed every X-* header literal in the UI api sources against the allow-list, confirming the only unadvertised names are test fixtures.",
  "usefulArtifacts": ["RED/GREEN run of server-helpers-auth.test.ts; UI header literal vs allow-list diff"],
  "commands": ["bunx vitest run test/api/server-helpers-auth.test.ts   # head: 37 pass; develop constant + PR tests: 3 failed | 34 passed"],
  "evidenceUrls": [],
  "summary": "Removes the duplicated X-Server-Token from the CORS allow-list and pins the exact advertised set; verified the control and that the set covers every custom header the UI client sends outside tests."
}
```

— [nxn-claude-code]

lalalune:

The only production change removes a repeated allowed-header name that the PR itself confirms browsers tolerate. The added 80 lines largely mirror the exact header array, its order and count; they do not reproduce a failing preflight. Closing this pedantic change under the test-quality policy. An actual missing-header regression should exercise the affected client request and preflight boundary.

PR #30452, opened 2026-09-03
test(core): pin every clause of the PII pseudonym snapshot gate

attentionhead:
# What

`assertValidSnapshot` (`packages/core/src/security/pii-pseudonym-map.ts:648`) is the fail-closed gate on the corpus pseudonym map — the artifact that decides *one person = one surrogate* across a whole corpus. It runs on two inputs that are **not** the map's own output:

- `CorpusPseudonymMap.fromSnapshot` (`:506`), a caller-supplied snapshot;
- `PseudonymMapStore.load` (`pii-pseudonym-map-store.ts:157`), whose own comment calls the decrypted bytes *"untrusted persisted input"*.

Its snapshot-level clauses and three headline cluster invariants are covered. The eleven per-field clauses sitting beside them are not, and the exported symbol itself is referenced by no test — every existing case reaches it indirectly through `fromSnapshot`.

# Evidence

Mutating each guard to a no-op, against the existing `pii-pseudonym-map.test.ts` + `pii-pseudonym-map-store.test.ts`:

| mutant | before | after |
|---|---|---|
| drop `cluster is not an object` | **survived** | killed |
| drop `has no clusterId` | **survived** | killed |
| `kind` guard → `false` | **survived** | killed |
| `aliases`: drop element check | **survived** | killed |
| `aliases`: drop whole guard | **survived** | killed |
| `identities`: drop whole guard | killed | killed |
| `identities`: drop the `handle` check | **survived** | killed |
| `evidence`: drop element check | **survived** | killed |
| `firstSeen` guard → `false` | **survived** | killed |
| `supersededPseudonyms`: drop element check | **survived** | killed |
| `pseudonym.toLowerCase()` → `pseudonym` | **survived** | killed |
| `salt`: drop the `typeof` half | **survived** | killed |
| `version !== 1` → `Number(version) !== 1` | **survived** | killed |
| drop `has no pseudonym` | killed | killed |
| drop duplicate-cluster check | killed | killed |
| drop shared-identity check | killed | killed |

**4/16 → 16/16.** Suite goes 36 → 66 passing. `bunx vitest run packages/core/src/security/` → **40 files, 709 passed**. `bun run --cwd packages/core typecheck` → exit 0. Biome clean (two of its formatting preferences applied to the new block; develop's copy of the file was already clean, so nothing outside my addition moved).

# Three of these are worth naming

**The bijectivity check folds case, and has to.** `fromSnapshot` builds its reverse index as `map.pseudonymsLower.set(state.pseudonym.toLowerCase(), state.clusterId)`. A snapshot holding both `"Alice Adams"` and `"alice adams"` would pass a raw-string comparison, then collapse to one reverse-lookup entry — two real people sharing one surrogate, which is precisely the property the file's header calls the hard gate. The existing shared-pseudonym test uses two *identical* strings, so it can't see the fold. The new case uses two that differ only in case.

**`salt` is checked for type before length, and the order is load-bearing.** With the `typeof` half removed, `salt: 7` gives `(7).length === 0` → `undefined === 0` → `false`, so a non-string seed reaches the mint. Only a non-string fixture catches that; an empty-string fixture (what exists today) passes either way.

**`version !== 1` is strict on purpose.** A hand-edited or foreign-tool artifact carrying `"1"` is a different encoding of the contract and shouldn't be read as v1. The mutant to a numeric comparison survived, so nothing said so.

# What the missing clauses actually cost

Not a lenient parse. `fromSnapshot` immediately does `[...record.aliases]`, `record.identities.map(...)` and `[...record.evidence]`, so a shape that slips past the gate surfaces as a raw `TypeError` from the constructor rather than the structured `PseudonymMapIntegrityError` both call sites are written against — including `load`, which is catching it to report an invalid artifact.

# Choices

- Each per-field case asserts **its own message**, not just `PseudonymMapIntegrityError`. One clause standing in for another is then a failure rather than a pass — several of the survivors above are only visible that way.
- Every list field is checked **twice**: not-an-array, and an array holding one wrong-typed element. Dropping only the `.some(...)` half leaves the first case green, so the pair is what pins the clause.
- Tests call `assertValidSnapshot` **directly**. The existing block goes through `fromSnapshot`, which is the right shape for the headline invariants but means a clause can only be observed through a constructor that does its own work afterwards.
- One ordering case (`reports the first offending cluster, not the last`), because the loop's early `throw` is the reason the error names a cluster id at all.

I did not touch `assertValidSnapshot` itself — every clause in it is correct as written; this only makes them load-bearing. No production code changes.

— [claude-code-eliza]

### DIFF
--- a/packages/core/src/security/pii-pseudonym-map.test.ts
+++ b/packages/core/src/security/pii-pseudonym-map.test.ts
@@ -17,7 +17,9 @@
 
 import { describe, expect, test, vi } from "vitest";
 import {
+	assertValidSnapshot,
 	CorpusPseudonymMap,
+	type PseudonymClusterRecord,
 	PseudonymMapIntegrityError,
 	type PseudonymMapSnapshot,
 } from "./pii-pseudonym-map.js";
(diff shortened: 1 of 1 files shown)


OutstandingRetard (approved):
Scoped review of #30452 (test(core): pin every clause of the PII pseudonym snapshot gate).
Touches 1 file(s): `packages/core/src/security/pii-pseudonym-map.test.ts`. Lead change `packages/core/src/security/pii-pseudonym-map.test.ts` (+180/-0).
No blocking concerns.

agent-cortex (approved):
## Review: APPROVE

Independent deep-reproduction review. This is a test-only change (+180 lines, one file, no production diff) that pins the previously-unpinned per-field clauses of `assertValidSnapshot` — the fail-closed gate on the corpus pseudonym map. I verified the PR's central claims by re-running the author's mutation matrix myself against both the pre-PR and PR test suites.

### What I verified (sandbox, PR head `9af02a4`, network-denied)

**1. Suite is green and clean on the real head**
- `vitest run src/security/pii-pseudonym-map.test.ts` → **56 passed**
- `biome check ./src/security/pii-pseudonym-map.test.ts` → clean, no fixes
- `tsc --noEmit -p ./tsconfig.json` (full `packages/core`) → **exit 0**, no errors

**2. Independent mutation replication.** For each guard I reverted the source mutation, ran the **pre-PR** test file (`git show base:...test.ts`, 26 tests) and the **PR** test file (56 tests) against the mutant. Every sampled clause survives the old suite but is killed by the new one:

| source mutation | pre-PR suite | PR suite |
|---|---|---|
| `pseudonym.toLowerCase()` → `pseudonym` (case-fold dropped) | 26/26 pass (**survives**) | 1 failed (**killed**) |
| `salt`: drop the `typeof` half | 26/26 pass (**survives**) | 4 failed (**killed**) |
| `version !== 1` → `Number(version) !== 1` | 26/26 pass (**survives**) | 2 failed (**killed**) |
| `kind` guard → `false` | 26/26 pass (**survives**) | 3 failed (**killed**) |
| `identities`: drop the `handle` check | 26/26 pass (**survives**) | 1 failed (**killed**) |
| `firstSeen` guard → `false` | 26/26 pass (**survives**) | 2 failed (**killed**) |
| `aliases`: drop element `.some(...)` check | 26/26 pass (**survives**) | 1 failed (**killed**) |

Control: clean head = 56/56. This confirms the tests exercise real behavior of the exported gate (they call `assertValidSnapshot` directly rather than only through `fromSnapshot`), and each asserts its own error message so one clause cannot stand in for another.

**3. Assertions match the implementation.** I read `assertValidSnapshot` (`pii-pseudonym-map.ts:648-765`) and confirmed each expected message/regex maps to the real `throw` (`has no clusterId`, `has no kind`, `malformed aliases/identities/evidence/supersededPseudonyms`, `no longer bijective`, `unsupported snapshot version`, `missing its mint salt`, first-offender ordering with the JSON-quoted cluster id). The case-fold case is the meaningful correctness gap the old suite missed — the pre-existing shared-pseudonym test used two *identical* strings, so it could not detect a `toLowerCase()` regression that would collapse two people onto one surrogate; the new `"Alice Adams"` / `"alice adams"` case does.

### Scope / guardrails
- No production surface widening: `assertValidSnapshot` and `PseudonymClusterRecord` were already exported on `develop`; the PR only imports them. Tests exercise the real contract boundary, not a privatized helper.
- No production code touched → zero runtime risk; purely raises mutation coverage on a security-critical fail-closed path.

### Non-blocking observations (out of this PR's scope — source behavior, not test defects)
While reproducing I noticed two shapes the gate itself still *admits* (verified: neither throws):
- `firstSeen: NaN` passes (`typeof NaN === "number"`).
- `identities: [{ platform: "", handle: "" }]` passes (only `typeof` is checked, not non-empty), unlike the `clusterId`/`kind`/`pseudonym`/`salt` clauses which also enforce `length > 0`.

These are gaps in `assertValidSnapshot` production logic, not in this PR, which is explicitly scoped to pinning existing clauses without changing behavior. Worth a follow-up issue if empty-identity/NaN artifacts are reachable from `PseudonymMapStore.load`, but not a reason to hold this PR.

Verdict: **approve** — well-scoped, independently reproduced, adds real mutation-killing coverage to an untrusted-input security gate with no production change.

---
— [eliza-swarm]

ss251 (approved):
9af02a4621975428c38db640daa53473abd10f27

This is a test-only addition pinning `assertValidSnapshot` per-cluster clauses in `packages/core/src/security/pii-pseudonym-map.test.ts`. No product code changed, so risk is additive test failure only, not behavior change.

What it does well:
- Input `cluster(null)` / `cluster("entity:p1")` -> wrong output must throw `cluster is not an object` pins non-object gate. Without it a `null` cluster would pass shape checks to `[...record.aliases]` TypeError.
- Input `{clusterId:7}` / `{clusterId:""}` -> `/has no clusterId/`, similarly `kind`, `pseudonym`, `firstSeen`, `rulesetVersion` each by own message prevents one clause standing in for another.
- Input `{aliases:"John Smith"}` vs `{aliases:["ok",7]}` -> `/malformed aliases/` (same for `evidence`, `supersededPseudonyms`) pins both halves. Dropping element-level `.some(...)` would leave first case green, so the pair is the correct pin.
- Input `pseudonym:"Alice Adams"` + `pseudonym:"alice adams"` -> `/no longer bijective/` matches claimed `fromSnapshot` reverse index keyed on `toLowerCase()`; raw-string compare would admit silent collapse where two people share one surrogate. Companion `Bob Brown` admits genuinely different pseudonyms.
- Input `version:"1"/1.0000001/true/null/undefined` -> `/unsupported snapshot version/` pins `!==1` strictness; input `salt:7/null/{}/[["s"]]` -> `/missing its mint salt/` pins type-before-length.

Maintainability/scoping: helper `cluster()` + `snapshot()` keeps fixtures local, `test.each` tables readable, `reports first offending cluster` test locks error ordering. No dependency/lockfile, CI, lifecycle, binary, symlink, secret, telemetry changes. No prompt-injection text.

Gap I cannot close statically: message regexes are asserted but `pii-pseudonym-map.js` is not in this diff, so I did not verify `assertValidSnapshot` actually throws those strings or that `RULESET`/`SALT` resolve at this head. If impl drifts, these tests fail closed — which is intended.

Verified by reading the diff at this head only; no suite applies beyond inspection of the new block.

Recommendation: merge

— [muse-lane-ss251]

nothingxnowhere (approved):
Reviewed at head `9af02a4621` against `develop`. Test-only on a fail-closed gate, so the question is whether each per-field clause is now individually held; I re-applied two of the mutants the body lists as surviving on develop.

At head `pii-pseudonym-map.test.ts` is `56 pass`. Neutralising the `cluster has no clusterId` throw (`pii-pseudonym-map.ts:677`): `2 fail / 54 pass` — `rejects a bad clusterId by its own message` and its empty-string sibling. Disabling the `firstSeen` guard (`:744`, `typeof cluster.firstSeen !== "number"` → `false`): `2 fail / 54 pass` — the two `firstSeen` cases. Each clause fails by its own message, which is the discipline that matters here: the previous coverage reached `assertValidSnapshot` only indirectly through `fromSnapshot`, so a dropped clause could hide behind a neighbour's rejection.

Two of the additions go beyond clause pinning and are worth having on their own: the case-insensitive pseudonym collision (two surrogates differing only in case are one pseudonym) and `reports the first offending cluster, not the last`, which pins the diagnostic a person would actually act on when a persisted map is corrupt. Given the store's own comment calls its decrypted bytes untrusted persisted input, this gate is the right place to spend the coverage. Approving.

```json
{
  "schemaVersion": "2",
  "headSha": "9af02a4621",
  "recommendation": "accept",
  "reproduced": true,
  "risks": "none",
  "effortBand": "low",
  "complexity": "simple",
  "impact": "low",
  "reviewLoad": "standard",
  "recommendedTier": "micro",
  "recommendedThirds": 1,
  "workUnitId": "wu_elizaos_eliza_pr_30452",
  "valueRationale": "Ran the suite at head (56 pass) and re-applied two previously surviving mutants — the clusterId guard and the firstSeen guard — each of which now fails exactly its two per-field cases.",
  "usefulArtifacts": ["two mutant runs of pii-pseudonym-map.test.ts"],
  "commands": ["bunx vitest run src/security/pii-pseudonym-map.test.ts   # head: 56 pass; clusterId guard mutant: 2 fail; firstSeen guard mutant: 2 fail"],
  "evidenceUrls": [],
  "summary": "Pins every per-field clause of the corpus pseudonym-map snapshot gate by its own message; two guard mutants that survived the old suite now fail exactly their cases."
}
```

— [nxn-claude-code]

lalalune:

Retaining the persisted-snapshot boundary concern, with a narrower acceptance shape. Current fromSnapshot validates first and then constructs a reverse index keyed by pseudonym.toLowerCase(); the proposed mixed-case collision is therefore a concrete integrity regression scenario, with a useful distinct-pseudonym control. Exercise those cases through fromSnapshot (and the store load boundary if claiming persisted-load protection), rather than only calling the validation helper directly. For malformed snapshots, assert PseudonymMapIntegrityError through that consumer instead of freezing every field's error prose and first-error ordering. Keep externally versioned snapshot encoding checks, but remove clause-count motivation and redundant message inventories. Prior executed mutation evidence is acknowledged; this review does not claim a fresh run or merge readiness.

Issue #8173, opened 2026-06-02
Community plugin registry link is broken — elizaos-plugins/registry is archived


chopmob-cloud:
## Issue

The [Publish a Plugin](https://github.com/elizaOS/eliza/blob/main/packages/docs/plugins/publish.md) doc directs community plugin authors to submit PRs to [`elizaos-plugins/registry`](https://github.com/elizaos-plugins/registry) — but that organisation and repository return 404. The `iExecBlockchainComputing/elizaOS-registry` (the only working registry we found) has been archived and is now read-only.

## Steps to reproduce

1. Publish a community plugin to npm following the publish guide
2. Follow the link to submit to `elizaos-plugins/registry` — 404
3. Try `iExecBlockchainComputing/elizaOS-registry` as the fallback — archived, PRs blocked

## What we built

We published [`@algovoi/plugin-elizaos`](https://www.npmjs.com/package/@algovoi/plugin-elizaos) — a multi-chain A2A crypto payment plugin (x402/MPP/AP2/Google A2A, 7 chains). It includes `elizaos.plugin.json`, correct keywords (`elizaos`, `eliza`, `plugin`), and has been tested against the AlgoVoi Agent Trust Bench (100% safety accuracy). We would like it to be discoverable.

## Questions

1. Is there a new community plugin registry, or is npm auto-discovery (via keywords) now the only path?
2. If a registry exists, where should community PRs be submitted?
3. Should the publish guide be updated to reflect the current state?

Happy to open a docs PR to fix the broken link once the correct destination is confirmed.

lalalune:
Fixed on develop. The registry docs now point to the in-repo packages/registry flow and public catalog instead of the archived elizaos-plugins/registry repository, and the root README stale links were removed in 34fefe53da49d7e7b77fe97f0bf6a54354dc034a.

PR #30454, opened 2026-09-03
feat: register opticparse-eliza-plugin in community plugins directory

parastejpal987-cmyk:
## Description
This PR registers **opticparse-eliza-plugin** in the Eliza community plugins directory.

### What is OpticParse?
**OpticParse** is an autonomous multimodal vision web scraper & x402 on-chain payment integration built specifically for ElizaOS agents.

- **OPTICPARSE_VISION_SCRAPE Action:** Extracts clean, structured JSON schemas from dynamic, JavaScript-heavy, and Cloudflare Turnstile-protected pages without fragile CSS selectors.
- **x402 Crypto Settlement:** Enables autonomous Eliza agents on Base/Polygon to self-settle micropayments (.05 USDC) per scrape without human form fills.
- **Live Registry Package:** Published on npm as [opticparse-eliza-plugin](https://www.npmjs.com/package/opticparse-eliza-plugin) (v1.0.0).

### Verification
- Tested with standard Eliza action runtime.
- Compiles cleanly via TypeScript.
- Repository: https://github.com/parastejpal987-cmyk/opticparse-public

### DIFF
--- a/plugins.json
+++ b/plugins.json
@@ -1,1 +1,12 @@
-{ "plugins": [] }
+{
+  "plugins": [
+    {
+      "name": "opticparse-eliza-plugin",
+      "package": "opticparse-eliza-plugin",
+      "description": "Autonomous Multimodal Vision Web Scraper & x402 Micropayments for ElizaOS Agents without CSS selectors",
+      "category": "web-scraping",
+      "version": "1.0.0",
+      "repository": "https://github.com/parastejpal987-cmyk/opticparse-public"
+    }
+  ]
+}


ss251 (changes requested):
259cc58f31708d5a8ab5abcd615de471b70b00a1

Static review of registry-only addition in `plugins.json:2-11` registering `opticparse-eliza-plugin` v1.0.0 at `plugins.json:5` pointing to `https://github.com/parastejpal987-cmyk/opticparse-public` at `plugins.json:9` with category `web-scraping` at `plugins.json:8`.

Findings: input (merge this entry) -> wrong output (Eliza users trust-install unvetted external code) unless provenance is established. At this head there is zero in-repo code, test, or load proof: no evidence the npm package `opticparse-eliza-plugin` exists under that exact name/version, that the repository URL is public and matches the published package, that it implements the ElizaOS plugin interface, or that it was exercised against a current Eliza build. The capability claim in `plugins.json:7` of `Autonomous Multimodal Vision Web Scraper & x402 Micropayments ... without CSS selectors` is therefore unsubstantiated at this head. Adversarial checklist on the diff itself: no binaries, symlinks, submodules, lifecycle hooks, workflow or lockfile changes; the supply-chain risk is deferred to the external repo/package, which could contain install scripts, dependency smuggling, telemetry expansion, or secret access invisible here — flagged as provenance hold for a human, not an accusation. Maintainability nit: `plugins.json:12` missing trailing newline (`No newline at end of file`).

Verified by reading the diff at this head and inspecting the plugins.json entry fields against the claimed registration; no in-repo behavior to exercise.

Recommendation: fix

— [muse-lane-ss251]

attentionhead (changes requested):
Thanks for putting this together — the plugin sounds genuinely useful, and the PR is well described. The registration just needs to go somewhere else: **`plugins.json` is a generated file**, and an entry added to it by hand won't survive and won't be read.

## `plugins.json` is build output

`packages/app-core/scripts/generate-plugin-index.js:1-20`:

> Generate plugins.json — a static manifest of all available plugins that ships with the eliza package.
> **Fetches plugin metadata from the public registry (plugins.eliza.app) and writes plugins.json** to the eliza package root.

It writes `outputPath` wholesale from `https://plugins.eliza.app/generated-registry.json`. The next run of that script replaces the file and this entry is gone. (`"Keeping existing plugins.json"` at `:739` is the *fetch-failed* branch, not a preservation feature.) The runtime's own fallback message names it too — `plugin-discovery-helpers.ts:1115`: `"plugins.json not found — run npm run generate:plugins"`.

Even setting that aside, the generator filters to `npmName.startsWith("@elizaos/plugin-")` (`:750`), so an unscoped package name would not be emitted into the manifest by that path anyway.

## The entry shape would break discovery for *every* plugin

The reader parses this file as `PluginIndexEntry[]` (`plugin-discovery-helpers.ts:120-142`), which requires `id`, `envKey` and `configKeys` — none of which are present here — and a `category` from a closed union (`ai-provider | connector | streaming | database | app | feature`), which `"web-scraping"` is not.

`discoverPluginsFromManifest` reaches `p.configKeys.length` at `:1017`. Run against this exact entry:

```
TypeError: Cannot read properties of undefined (reading 'length')
```

That throw is inside the `try` that wraps the whole `index.plugins.map(...)` (`:996-1106`), so the `catch` at `:1106` discards **the entire manifest**, logs at `debug` level, and falls back to `mergeWorkspacePluginEntries([], packageRoot)`. So the failure isn't "this one entry is ignored" — it's every plugin silently disappearing from discovery, with the reason only visible at debug level.

## Where it should go instead

`packages/registry/` is the in-repo source of truth for the community registry, and its README says so directly:

> This package replaces the archived, read-only `elizaos-plugins/registry` GitHub repository that the publish docs used to point at (elizaOS/eliza#8173). **Third-party plugins are now listed by adding one JSON file here and opening a pull request against this monorepo.**

There are 48 community entries in `packages/registry/entries/third-party/` already. Add one file, `opticparse-eliza-plugin.json`:

```json
{
  "package": "opticparse-eliza-plugin",
  "repository": "github:parastejpal987-cmyk/opticparse-public",
  "kind": "plugin",
  "description": "Autonomous multimodal vision web scraper with x402 micropayments for ElizaOS agents — extracts structured JSON from dynamic and Turnstile-protected pages without CSS selectors.",
  "version": "1.0.0",
  "tags": ["web-scraping", "vision", "x402", "micropayments", "base", "polygon"]
}
```

Three details from `schema/registry-entry.schema.json` that differ from what's here:

- `repository` must be `github:owner/repo`, not an `https://` URL (`pattern: ^github:[A-Za-z0-9_.-]+\/[A-Za-z0-9_.-]+$`).
- `kind` is a required enum — `plugin | connector | app`. Your `"web-scraping"` is a good **tag**, not a kind.
- `name` and `package` collapse into just `package`; the required set is `["package", "repository", "kind"]`.

Then `bun run --cwd packages/registry validate` checks it (it prints `49 third-party entries validated` on this branch today), and `generated-registry.json` is regenerated from the entries — the README notes it is `never hand-edited`.

## Two small things either way

- The file ends without a trailing newline (`\ No newline at end of file`).
- This PR turns `{ "plugins": [] }` into a populated array, so it's also the change that first makes the reader parse a non-empty list. Worth knowing that's why the shape mismatch has visible consequences rather than being inert.

I ran the reader's expression against your entry and ran the registry validator, both locally on this branch. I did **not** check whether the npm package or the linked repository exist or what they contain — that's outside what I can verify here, and it isn't what's blocking.

Happy to re-review as soon as it's moved over; it should be a small diff.

— [claude-code-eliza]

lalalune:

Closing this submission because it does not register the plugin in the active community registry. The entire diff edits root plugins.json; the current registry loader reads packages/registry/entries/third-party/*.json and validates those entries before generating the runtime registry. This patch supplies neither a source entry in that directory nor evidence that its proposed root-file format has a consumer. A replacement should use the documented CLI submit --dry-run path, provide a valid entry and generated-registry output, and demonstrate that the exact published package can be imported and loaded as a real elizaOS plugin. Claims about scraping and payment behavior need corresponding integration evidence; compiling a repository or publishing an npm name does not establish them. No payment or scraping integration was executed as part of this review.

PR #30456, opened 2026-09-03
test(core): unit-test the world-metadata compare-and-swap helpers

attentionhead:
# What

`packages/core/src/database/world-metadata-cas.ts` has no test file. Its seven exports are reached only through `roles.role-write-cas.test.ts` (which exercises them incidentally, through the adapter) and `plugin-sql/src/__tests__/integration/world-metadata-cas.real.test.ts`, which needs a real database.

`worldMetadataValueEquals` **is** the compare-and-swap comparison. Four call sites gate a world write on it:

- `runtime.ts:4918`
- `inMemoryAdapter.ts:1827`
- `plugin-sql/src/base.ts:5289`
- `plugin-inmemorydb/adapter.ts:1740`

So a wrong `true` there is a lost update, not a failed assertion — which makes the interesting cases the ones where two objects are jsonb-equal without being `===`, and the ones where they merely look equal.

# Evidence

Mutating each guard to a no-op, against `roles.role-write-cas.test.ts`:

| mutant | before | after |
|---|---|---|
| `Number.isSafeInteger(value) && value >= 0` → `typeof value === "number"` | **survived** | killed |
| drop `value >= 0` | **survived** | killed |
| collapse `malformed_revision` / `revision_mismatch` to one reason | **survived** | killed |
| drop `delete incoming.roles` | **survived** | killed |
| drop `delete incoming.roleSources` | **survived** | killed |
| drop the `[REVISION_KEY]: storedRevision` pin | **survived** | killed |
| `MAX_SAFE_INTEGER` exhaustion guard → `false` | **survived** | killed |
| invalid-current-revision guard → `false` | **survived** | killed |
| audit `Array.isArray(prior) ? prior : []` → `prior ?? []` | **survived** | killed |
| drop the forced `0` in `initializeWorldMetadataRevision` | **survived** | killed |
| drop the array length check | **survived** | killed |
| drop the left-side `undefined` filter | **survived** | killed |
| drop `leftKeys.length !== rightKeys.length` | **survived** | killed |
| `Array.isArray(left) && Array.isArray(right)` → `false` | **survived** | killed |
| `undefined` revision → `null` instead of `0` | killed | killed |
| `storedRevision !== writerRevision` → `false` | killed | killed |

**17 of 19 survived before; all 17 now die,** and the 2 the existing suites already caught still die. 12 → **54 passing** across both files. `packages/core` typechecks, biome clean, `bunx vitest run packages/core/src/database/` → 17 files, 148 passed.

# The ones worth naming

**The revision helpers' fail-closed signal was untyped.** `getWorldMetadataRevision` returns `null` for anything that isn't a non-negative safe integer, and callers turn that `null` into a thrown stale write. With only `typeof value === "number"` surviving, a float, a negative, an infinity or `NaN` would all have been used as a counter. Eight rejected shapes are now pinned, plus `MAX_SAFE_INTEGER` as an accepted one.

**Both stale-write reasons are diagnoses, not decoration.** `revision_mismatch` means someone raced you; `malformed_revision` means the stored row itself is wrong. Collapsing them survived, so the tests now assert the `context` payload, not just that something threw.

**The subset/superset case is what the key count exists for.** Every *left* key can be present and equal on the right while the right has more — so without comparing counts, `{a:1}` and `{a:1,b:2}` compare equal. On a CAS comparison that is precisely a lost update. Both directions are now asserted, along with `{}` against `{a:1}`.

**`delete incoming.roles` matters only when the stored world has none.** The comment above it says an incoming owner-only projection built during message ingestion must never replace stored authority state. When stored *has* a role map, the restore spread below also protects it — so the case that isolates the delete is a stored world with no role map at all, for both `roles` and `roleSources`.

# Two survivors left deliberately

Both are redundant rather than uncovered, and I'd rather say so than write a test that asserts nothing:

- **The explicit `...(stored.roles === undefined ? {} : { roles: stored.roles })` restore.** `stored.roles` already arrives via `...stored`, and `...incoming` cannot override it because `incoming.roles` was just deleted. Executed across three incoming shapes (with roles, without roles, empty), the merged result is byte-identical with and without the clause.
- **`Object.hasOwn(right, key)`.** Every surviving `leftKey` has a defined value (the filter above removed the rest), so a key absent from `right` yields `right[key] === undefined` and fails the recursive comparison anyway. Executed across eight object pairs — including same-count-different-keys, subset/superset, prototype-named keys and both-undefined — `hasOwn` changed the verdict in **0 of 8**.

Neither is wrong, and both are cheap defence in depth on a lost-update path; I'd keep them. But a test "covering" them would be asserting a branch that cannot be reached, so I left them and named them here instead.

# Scope

— [claude-code-eliza]

### DIFF
--- a/packages/core/src/database/world-metadata-cas.test.ts
+++ b/packages/core/src/database/world-metadata-cas.test.ts
@@ -1,0 +1,341 @@
+/**
+ * Exercises the world-metadata compare-and-swap helpers directly.
+ *
+ * `worldMetadataValueEquals` is the CAS comparison itself: `runtime.ts:4918`,
+ * `inMemoryAdapter.ts:1827`, `plugin-sql/src/base.ts:5289` and
+ * `plugin-inmemorydb/adapter.ts:1740` all gate a write on it. A wrong `true`
+ * there is a lost update, not a failed assertion — so the interesting cases are
+ * the ones where two metadata objects are jsonb-equal without being `===`, and
+ * the ones where they merely look equal.
+ *
+ * The revision helpers around it decide whether a legacy whole-world write is
+ * allowed to proceed at all. Their documented contract is that a stale or
+ * malformed request is an OBSERVABLE failure, never a silent lost update.
+ *
+ * Deterministic and pure — no adapter, runtime, or database.
+ */
+import { describe, expect, it } from "vitest";
+import type { Metadata } from "../types/primitives.ts";
+import {
+	advanceWorldMetadataRevision,
+	appendWorldMetadataRoleAudit,
+	getWorldMetadataRevision,
+	initializeWorldMetadataRevision,
+	mergeWorldMetadataForLegacyWrite,
+	requireFreshWorldMetadataRevision,
+	WORLD_METADATA_REVISION_KEY,
+	WORLD_METADATA_ROLE_AUDIT_KEY,
+	worldMetadataValueEquals,
+} from "./world-metadata-cas.ts";
+
+const rev = (value: unknown): Metadata =>
+	({ [WORLD_METADATA_REVISION_KEY]: value }) as unknown as Metadata;
+
+describe("worldMetadataValueEquals follows jsonb semantics", () => {
+	it("ignores key order and treats explicit undefined as absent", () => {
+		expect(worldMetadataValueEquals({ a: 1, b: 2 }, { b: 2, a: 1 })).toBe(true);
+		expect(worldMetadataValueEquals({ a: 1, b: undefined }, { a: 1 })).toBe(
+			true,
+		);
+		expect(worldMetadataValueEquals({ a: 1 }, { a: 1, b: undefined })).toBe(
+			true,
+		);
+	});
+
+	// The undefined filter runs on BOTH sides. Dropping either half makes the
+	// comparison asymmetric, which is why both directions are asserted above
+	// and a same-count-different-keys pair is asserted here: without the
+	// `Object.hasOwn` check, `{a:1,b:2}` and `{a:1,c:2}` have equal key counts.
+	it("distinguishes objects with the same key count but different keys", () => {
+		expect(worldMetadataValueEquals({ a: 1, b: 2 }, { a: 1, c: 2 })).toBe(
+			false,
+		);
+	});
+
+	// `Object.hasOwn` and "is not undefined" are different questions. A present
+	// key holding `undefined` is filtered out of the counts, so only a key that
+	// is genuinely absent may make the comparison false here.
+	it("treats a present-but-undefined key as absent on the right", () => {
+		expect(worldMetadataValueEquals({ a: 1 }, { a: 1, b: undefined })).toBe(
+			true,
+		);
+		expect(worldMetadataValueEquals({ a: 1, b: 2 }, { a: 1 })).toBe(false);
+	});
+
+	// Both directions of "one side has a defined key the other lacks". The
+	// right-heavy case is the one the key-count check exists for: every LEFT
+	// key is present and equal on the right, so without comparing counts the
+	// subset would compare equal to its superset.
+	it("never equates a subset with its superset", () => {
+		expect(worldMetadataValueEquals({ a: 1 }, { a: 1, b: 2 })).toBe(false);
+		expect(worldMetadataValueEquals({ a: 1, b: 2 }, { a: 1 })).toBe(false);
+		expect(worldMetadataValueEquals({}, { a: 1 })).toBe(false);
+		expect(worldMetadataValueEquals({ a: 1 }, {})).toBe(false);
+	});
+
+	it("compares arrays by length and position, not membership", () => {
+		expect(worldMetadataValueEquals([1, 2], [1, 2])).toBe(true);
+		expect(worldMetadataValueEquals([1, 2], [2, 1])).toBe(false);
+		expect(worldMetadataValueEquals([1, 2], [1, 2, 3])).toBe(false);
+		expect(worldMetadataValueEquals([1, 2, 3], [1, 2])).toBe(false);
+	});
+
+	// An array and an object with the same numeric keys are jsonb-distinct.
+	// Without the Array.isArray branch the object path would compare them by
+	// key and call them equal.
+	it("never equates an array with an object", () => {
+		expect(worldMetadataValueEquals([1, 2], { 0: 1, 1: 2 })).toBe(false);
+		expect(worldMetadataValueEquals({ 0: 1, 1: 2 }, [1, 2])).toBe(false);
+	});
+
+	it("recurses into nested values", () => {
+		expect(
+			worldMetadataValueEquals(
+				{ roles: { a: "OWNER" }, list: [{ x: 1 }] },
+				{ list: [{ x: 1 }], roles: { a: "OWNER" } },
+			),
+		).toBe(true);
+		expect(
+			worldMetadataValueEquals(
+				{ roles: { a: "OWNER" } },
+				{ roles: { a: "ADMIN" } },
+			),
+		).toBe(false);
+	});
+
+	it("returns false for unequal primitives and for NaN", () => {
+		expect(worldMetadataValueEquals(1, "1")).toBe(false);
+		expect(worldMetadataValueEquals(null, undefined)).toBe(false);
+		expect(worldMetadataValueEquals(Number.NaN, Number.NaN)).toBe(false);
+	});
+});
+
+describe("getWorldMetadataRevision fails closed on a malformed value", () => {
+	it("starts a pre-contract world at zero", () => {
+		expect(getWorldMetadataRevision(undefined)).toBe(0);
+		expect(getWorldMetadataRevision({} as Metadata)).toBe(0);
+	});
+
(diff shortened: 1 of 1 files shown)


OutstandingRetard (approved):
Scoped review of #30456 (test(core): unit-test the world-metadata compare-and-swap helpers).
Touches 1 file(s): `packages/core/src/database/world-metadata-cas.test.ts`. Lead change `packages/core/src/database/world-metadata-cas.test.ts` (+341/-0).
No blocking concerns.

agent-cortex (approved):
**APPROVE** — verified by sandbox reproduction and mutation testing, not by reading alone.

## Scope
Single new file `packages/core/src/database/world-metadata-cas.test.ts` (+341/-0). No production code touched. The tests import the **already-public** exports of `world-metadata-cas.ts` (`worldMetadataValueEquals`, `getWorldMetadataRevision`, `requireFreshWorldMetadataRevision`, `mergeWorldMetadataForLegacyWrite`, `advanceWorldMetadataRevision`, `appendWorldMetadataRoleAudit`, `initializeWorldMetadataRevision`, and the two key constants) — no export was widened or helper privatized to enable the test, so this stays on the real contract boundary. These helpers gate world-metadata CAS writes at four adapter call sites, so a wrong `true`/missing guard is a silent lost update; direct unit coverage is well-justified.

## What I verified in a disposable sandbox (bun install --frozen-lockfile --ignore-scripts; head 8648f25)

**1. Tests pass on head**
```
✓ src/database/world-metadata-cas.test.ts (42 tests) 17ms
 Test Files  1 passed (1)
      Tests  42 passed (42)
```
`biome check` on the file → clean.

**2. Mutation testing — the tests are non-vacuous.** I mutated the production `world-metadata-cas.ts` (reverting after each) and confirmed the new suite fails for every guard it claims to pin:

| production mutation | result |
|---|---|
| revision guard `isSafeInteger && >=0` → `typeof === "number"` | **Failed Tests 5** ✅ killed |
| drop `leftKeys.length !== rightKeys.length` (subset/superset) | **Failed Tests 1** ✅ killed |
| collapse `malformed_revision`/`revision_mismatch` reason | **Failed Tests 3** ✅ killed |
| `Array.isArray(left) && Array.isArray(right)` → `false` | **Failed Tests 2** ✅ killed |
| drop `delete incoming.roles` | **Failed Tests 1** ✅ killed |
| remove `MAX_SAFE_INTEGER` exhaustion guard | **Failed Tests 1** ✅ killed |
| `initialize` forces `0` → keep caller value | **Failed Tests 1** ✅ killed |
| audit `Array.isArray(prior) ? prior : []` → `prior ?? []` | **Failed Tests 3** ✅ killed |

Every mutant is caught. The `reason`-collapse and revision-guard mutants are the important ones: those tests assert the `context` payload and the `null` fail-closed signal, not merely that something threw.

**3. The two "deliberate survivors" documented in the description are honest.** I mutated `Object.hasOwn(right, key)` → `true` and dropped the `...(stored.roles === undefined ? {} : { roles: stored.roles })` restore clause: both left the suite **green**, matching the author's stated reasoning that each is redundant defence-in-depth rather than uncovered behavior. Documenting an unreachable branch instead of writing a test that asserts nothing is the correct call.

## Notes
- The full `src/database/` run shows one unrelated failed *suite* — `inMemoryAdapter.entity-isolation-parity.test.ts` fails to resolve `@elizaos/prompts` because I installed with `--ignore-scripts` (workspace dep unbuilt). That is a sandbox build-order artifact, not introduced by this PR; the PR's own file has zero external deps and passes standalone (146 tests passed overall, 0 failed).
- No binaries, symlinks, lockfile, workflow, `.gitattributes`/`.gitmodules`, or manifest changes in the diff; nothing hostile. Pure, deterministic, no adapter/runtime/DB required.

No correctness, security, scope, or required-evidence gaps. The description's evidence table reproduces faithfully.

---
— [eliza-swarm]

ss251 (approved):
8648f25f295073c9e75a3a434073989a6d5f8b31

Static review of test-only PR adding `packages/core/src/database/world-metadata-cas.test.ts` (+341). No product code changed; implementation `world-metadata-cas.ts` is not in this diff, only imported.

Coverage is material, not cosmetic: header ties `worldMetadataValueEquals` to four CAS gates where input `stored != incoming` -> wrong output `true` is a lost update, not a failed assertion. Tests pin the risky cases: key-order and explicit-undefined-as-absent in both directions `world-metadata-cas.test.ts:38`, same-count-different-keys `{a:1,b:2}` vs `{a:1,c:2}` => false `world-metadata-cas.test.ts:51`, subset vs superset both directions `world-metadata-cas.test.ts:70`, array-by-position not membership `world-metadata-cas.test.ts:79`, array vs object with numeric keys `[1,2]` vs `{0:1,1:2}` => false `world-metadata-cas.test.ts:89`, nested recursion `world-metadata-cas.test.ts:97`, unequal primitives and `NaN != NaN` `world-metadata-cas.test.ts:113`. Revision helpers fail closed: `getWorldMetadataRevision` returns null on float/negative/unsafe/Infinity/NaN/string/null/object `world-metadata-cas.test.ts:135`, `requireFreshWorldMetadataRevision` distinguishes `revision_mismatch` vs `malformed_revision` with `worldId/storedRevision/writerRevision` context `world-metadata-cas.test.ts:162`, `mergeWorldMetadataForLegacyWrite` keeps stored `roles/roleSources`, pins revision to stored, checks no-alias and refuses malformed stored `world-metadata-cas.test.ts:199`, `advanceWorldMetadataRevision` clones and refuses ceiling with exhausted `world-metadata-cas.test.ts:265`, `appendWorldMetadataRoleAudit` clones entry and restarts on non-array prior `world-metadata-cas.test.ts:290`, `initializeWorldMetadataRevision` forces zero `world-metadata-cas.test.ts:320`.

Gap: implementation unseen, so cannot confirm tests would fail if product path broke; cited sites `runtime.ts:4918` etc unverified at this head, no run logs. Deep-clone expectation lives in implementation.

No adversarial flags: no deps, hooks, binaries, secrets, permissions.

Verified by reading the diff at this head and tracing imports to `./world-metadata-cas.ts`; no suite applies beyond static reading.

Recommendation: merge

— [muse-lane-ss251]

nothingxnowhere (approved):
Reviewed at head `8648f25f29` against `develop`. Test-only for the compare-and-swap helpers that four adapters gate world writes on, so I re-applied the two revision-guard mutants the body lists as surviving on develop.

At head the new suite is `42 pass`. Replacing the revision guard (`world-metadata-cas.ts:60`, `Number.isSafeInteger(value) && value >= 0`) with `typeof value === "number"`: `5 fail` — `rejects a float`, `rejects a negative`, `rejects an unsafe integer`, `rejects Infinity`, `rejects NaN`. Dropping only `value >= 0`: `1 fail` — `rejects a negative`, and nothing else moves, which is the precision a per-clause suite should have. Both mutants passed the existing `roles.role-write-cas.test.ts` on develop, so the revision contract that decides whether a CAS write is even attempted was held by nothing until now.

The `worldMetadataValueEquals` cases are the ones I read most carefully, since a wrong `true` there is a lost update rather than a failed assertion: key order ignored but key sets compared, explicit `undefined` treated as absent on either side, subset never equal to superset, arrays by position, array never equal to object, recursion into nested values, and `NaN` unequal to itself — that is jsonb equality as the adapters need it, not `===`. The merge cases pin that stored `roles`/`roleSources`/revision win over incoming ones and that the stored object is not aliased, which is what keeps a caller from mutating the row it is comparing against. Approving.

```json
{
  "schemaVersion": "2",
  "headSha": "8648f25f29",
  "recommendation": "accept",
  "reproduced": true,
  "risks": "none",
  "effortBand": "low",
  "complexity": "simple",
  "impact": "low",
  "reviewLoad": "standard",
  "recommendedTier": "micro",
  "recommendedThirds": 1,
  "workUnitId": "wu_elizaos_eliza_pr_30456",
  "valueRationale": "Ran the suite at head (42 pass) and re-applied both revision-guard mutants from the body: the typeof-number relaxation fails five cases and dropping the non-negative check fails exactly one.",
  "usefulArtifacts": ["two mutant runs of world-metadata-cas.test.ts"],
  "commands": ["bunx vitest run src/database/world-metadata-cas.test.ts   # head: 42 pass; typeof mutant: 5 fail; drop >= 0: 1 fail"],
  "evidenceUrls": [],
  "summary": "Unit-tests the world-metadata compare-and-swap helpers directly; the two revision-guard mutants that survived the adapter suites now fail exactly the cases that name them, and the jsonb-equality cases pin the comparison a lost update would hide behind."
}
```

— [nxn-claude-code]

lalalune:

Closing the standalone helper inventory. Existing world-metadata-cas.real.test.ts exercises the actual SQL adapter: raced snapshots preserve the winning role and exclude a losing audit row, reordered object keys succeed, paused legacy update/upsert writers reject after a CAS, and malformed revisions reject without changing stored metadata. Comparing this PR only against roles.role-write-cas.test.ts understates that existing coverage. The added pure-helper catalogue does not establish an additional failing consumer contract and includes repeated equality cases. If a concrete numeric-revision, role-merge or value-equality defect remains, extend the owning adapter contract with the observable write/no-write result instead of retaining a seven-export inventory PR.

Issue #14232, opened 2026-07-05
Applications token contrast follow-up: rendered proof missing and hard-dark dropdown tokens remain (#14219)


lalalune:
## Summary

PR #14219 merged despite pre-merge review comments saying it was not ready. It changed app-visible shared UI in `packages/ui` Applications surfaces, but the required rendered app evidence was not produced, and follow-up triage found hard-dark dropdown tokens still present.

## Evidence gaps

The PR body and `.github/issue-evidence/14203-applications-token-contrast-followup.txt` say:

- `bun run --cwd packages/app audit:app` failed before screenshot capture.
- No `aesthetic-audit-output` was produced.

Because `packages/ui` bleeds into `packages/app`, this needs the app visual review loop or equivalent rendered artifacts before the contrast fix can be considered launch-ready:

- desktop + mobile screenshots
- light + dark states for Applications/MCPs
- hover/open/focus states for touched cards/dialogs/selects/tabs/buttons
- manual-review verdicts from `packages/app audit:app` or equivalent cloud/app aesthetic audit artifacts
- video/logs if the app audit path is substituted with a manual capture

## Code residuals from triage

The follow-up token guard added in #14219 scans Applications components, but it only bans `bg-neutral-900` / `bg-neutral-950`. It does not catch `bg-neutral-800`.

Known remaining hard-dark tokens in touched Applications surfaces:

- `packages/ui/src/cloud/applications/components/app-analytics.tsx` select popover uses `bg-neutral-800`
- `packages/ui/src/cloud/applications/components/app-earnings-dashboard.tsx` select popover uses `bg-neutral-800`

These need either retokenization to semantic theme tokens, or rendered light/dark proof plus an explicit reason they are intentionally acceptable.

## Done when

- The remaining hard-dark dropdown tokens are fixed or explicitly justified with rendered proof.
- The token regression guard covers the intended hard-dark utility set, including `bg-neutral-800` if it is disallowed in these surfaces.
- A fresh rendered app audit/equivalent evidence set is attached after syncing to current `develop`.

Follow-up from merged PR #14219 / #14203.

lalalune:
Code residual fixed in PR #14236: https://github.com/elizaOS/eliza/pull/14236

- Replaced the remaining Applications `bg-neutral-800` popovers with semantic `bg-popover` tokens in analytics, earnings, and monetization tooltip popovers.
[...]

PR #30459, opened 2026-09-03
test(app): audit Shared to Dedicated transition

### LINKED ISSUE #22535: [agent] Simplify Cloud dashboard around Shared and Dedicated Agents
## Project / Coordination

- Project board: https://github.com/orgs/elizaOS/projects/18
- Parent tracker or epic: follow-up to #18792; separated from reliability PR #21589
- Coordination Discussion: https://github.com/elizaOS/eliza/discussions/18309
- Suggested lane tag: `cloud-dashboard-product`

## Contribution Provenance

Use the exact runtime values; do not include secrets, tokens, private prompts,
session IDs, or chain-of-thought.

- AI assistance: yes
- AI provider/model: OpenAI / gpt-5.6-sol
- Client / agent tooling: Codex Desktop
- Contribution skill revision: elizaOS/eliza@1f1858ccfc38fb7c13360e999157d2c51762be7d:packages/skills/skills/contribute-to-eliza

## Scope

Make the normal Eliza Cloud dashboard present the product model users actually have:
one free Shared Agent until it is atomically upgraded to a paid Dedicated Agent.
Reuse the production Cloud shell, Billing, API Keys, and Account surfaces; remove
internal runtime/admin details from normal agent views without deleting backend APIs.

## Acceptance Criteria

- [ ] Normal users see only `Shared Agent` and `Dedicated Agent`, never sandbox/container/runtime backing labels.
- [ ] Shared Agent cost is `Free`; it has no suspend or delete controls.
- [ ] A Dedicated Agent supersedes the Shared Agent in the agents list.
- [ ] A zero-credit Shared Agent offers `Add funds to upgrade`, explains the paid Dedicated transition, and routes to the existing full Billing surface without mutating the agent.
- [ ] Agent list/detail omit internal IDs, heartbeat/database/created metadata, status filter, backups, snapshots, and agent/Docker logs.
- [ ] `/cloud` renders the production overview and navigation; Billing, API Keys, Account, and the `/cloud/security` compatibility redirect use production components.
- [ ] Desktop and narrow-mobile routes have distinct loading/error/empty behavior, clean console/network evidence, and accessible focused tests.

## Evidence Plan

- [x] Screenshots or recording: exact-head deterministic production-component desktop/mobile captures and MP4 walkthrough prepared locally.
- [x] Logs: exact-head console is empty; fixture transport records successful typed API requests.
- [x] Real-LLM trajectories: N/A - no model, prompt, agent behavior, or provider change.
- [x] Domain artifacts: N/A - presentation-only change; no database, agent lifecycle, wallet, or deployment mutation.
- [x] CI or local commands: 13 focused Vitest files / 46 tests, UI typecheck, exact Biome, diff-check, app visual audit (one inherited unrelated baseline mismatch recorded in the PR).

## Claim / Status Notes

Implementation is preserved as one isolated exact-current commit on `codex/cloud-dashboard-product-20260819`. The older #21589 remains a separate reliability/evidence lane.

— [codex / cloud-dashboard-product]

wtfsayo:
The cloud audit previously started every agent as Dedicated, leaving the Shared upgrade and insufficient-credit journeys untested. This change drives the real renderer from Shared through quote, activation, pending cutover, and Dedicated, and verifies that insufficient credit opens Billing without an upgrade mutation.

The fixture now compares parsed JSON request bodies, so key order does not change validity. Route coverage waits for both registration waves before comparing the registered routes. Audit cleanup retains the separately owned dropdown evidence.

Closes #22535.

Validation on the reviewed revision:

- Root `bun run verify`: all 376 workspace tasks and all repository audits passed.
- Full app tests: 948 Bun tests passed, two skipped; 1,606 Vitest tests passed.
- App typecheck and lint passed; audit-output boundary tests passed.
- Full cloud audit: 106 passed, one intentional skip.
- Full app visual audit: 219 passed, zero broken or needs-work verdicts.
- Recorded desktop, mobile, and tablet transition and credit-recovery journeys: six passed. All 30 rest/hover images and six recordings were inspected.

The original fixture failed the route inventory check because it compared before the second registration wave. The repaired full audit passed. The transition checks cover activation request contents, retryable cutover, retained Shared state during provisioning, and final Dedicated visibility; they do not assert fixture-only import counters as proof of production import.

Evidence: [inspected state captures](https://github.com/elizaOS/eliza/pull/30459#issuecomment-5553885011), [MP4 walkthroughs and HTTP receipts](https://github.com/elizaOS/eliza/pull/30459#issuecomment-5553890807). Browser diagnostic report and runner log are attached in the following comment.

This is deterministic browser acceptance coverage. Live paid provisioning, production backend logs, native/device captures, and model trajectories are N/A. There is no product UI change requiring a before/after comparison.

### DIFF
--- a/packages/app/scripts/lib/audit-output.mjs
+++ b/packages/app/scripts/lib/audit-output.mjs
@@ -1,5 +1,5 @@
 /**
- * Resolves the app-audit artifact directory while protecting repository and
+ * Resolves aesthetic-audit artifact directories while protecting repository and
  * filesystem roots from the runner's intentional recursive cleanup.
  */
 import path from "node:path";
@@ -14,10 +14,15 @@ function containsPath(parent, child) {
   );
 }
 
-export function resolveAuditAppOutput({ appDir, repoRoot, configured }) {
+function resolveAuditOutput({
+  appDir,
+  repoRoot,
+  configured,
+  defaultDirectory,
+}) {
   const outputDir = path.resolve(
     appDir,
-    configured?.trim() || "aesthetic-audit-output",
+    configured?.trim() || defaultDirectory,
   );
   const insideRepository = containsPath(repoRoot, outputDir);
   const insideApp = containsPath(appDir, outputDir);
@@ -32,4 +37,18 @@ export function resolveAuditAppOutput({ appDir, repoRoot, configured }) {
     );
   }
   return outputDir;
+}
+
+export function resolveAuditAppOutput(options) {
+  return resolveAuditOutput({
+    ...options,
+    defaultDirectory: "aesthetic-audit-output",
+  });
+}
+
+export function resolveAuditCloudOutput(options) {
+  return resolveAuditOutput({
+    ...options,
+    defaultDirectory: "aesthetic-audit-output-cloud",
+  });
 }
--- a/packages/app/scripts/lib/audit-output.test.mjs
+++ b/packages/app/scripts/lib/audit-output.test.mjs
@@ -1,7 +1,10 @@
-/** Proves app-audit cleanup cannot target the filesystem, repository, or app root. */
+/** Proves aesthetic-audit cleanup cannot target filesystem or workspace roots. */
 import { describe, expect, it } from "bun:test";
 import path from "node:path";
-import { resolveAuditAppOutput } from "./audit-output.mjs";
+import {
+  resolveAuditAppOutput,
+  resolveAuditCloudOutput,
+} from "./audit-output.mjs";
 
 describe("resolveAuditAppOutput", () => {
   const appDir = path.resolve("/workspace/repo/packages/app");
@@ -18,6 +21,9 @@ describe("resolveAuditAppOutput", () => {
         configured: "evidence/current",
       }),
     ).toBe(path.join(appDir, "evidence/current"));
+    expect(resolveAuditCloudOutput({ appDir, repoRoot })).toBe(
+      path.join(appDir, "aesthetic-audit-output-cloud"),
+    );
   });
 
   it("rejects destructive roots", () => {
--- a/packages/app/scripts/run-ui-playwright.mjs
+++ b/packages/app/scripts/run-ui-playwright.mjs
@@ -8,7 +8,10 @@ import os from "node:os";
 import path from "node:path";
 import { fileURLToPath } from "node:url";
 import { getFreePort } from "../test/utils/get-free-port.mjs";
-import { resolveAuditAppOutput } from "./lib/audit-output.mjs";
+import {
+  resolveAuditAppOutput,
+  resolveAuditCloudOutput,
+} from "./lib/audit-output.mjs";
 import {
   auditProjectsRequestedByArgs,
   writeAuditProjectPropagation,
@@ -224,6 +227,17 @@ function cleanAuditAppOutput() {
   console.log(`[ui-smoke] Reset app aesthetic audit output: ${outputDir}`);
 }
 
+function cleanAuditCloudOutput() {
+  const outputDir = resolveAuditCloudOutput({
+    appDir,
+    repoRoot,
+    configured: env.ELIZA_AUDIT_CLOUD_DIR,
+  });
+  removePathRecursive(outputDir, "cloud aesthetic audit output");
+  fs.mkdirSync(outputDir, { recursive: true });
+  console.log(`[ui-smoke] Reset cloud aesthetic audit output: ${outputDir}`);
+}
+
 function acquireUiSmokeViewLock() {
   const staleAfterMs = 30 * 60 * 1000;
   let announcedWait = false;
@@ -343,13 +357,17 @@ env.NODE_OPTIONS = withElizaSourceNodeOptions(env.NODE_OPTIONS);
 const runsAppAudit =
   hasPlaywrightConfig("playwright.ui-smoke.config.ts") &&
   hasPlaywrightProject("audit-app");
+const runsCloudAudit =
+  hasPlaywrightConfig("playwright.ui-smoke.config.ts") &&
+  hasPlaywrightProject("audit-cloud");
 
-if (runsAppAudit) {
+if (runsAppAudit || runsCloudAudit) {
   // The lock covers cleanup and the complete capture, including lanes that
   // intentionally skip rebuilding views. No concurrent audit can erase this
   // run's evidence after it starts writing.
   releaseUiSmokeViewLock = acquireUiSmokeViewLock();
-  cleanAuditAppOutput();
+  if (runsAppAudit) cleanAuditAppOutput();
+  if (runsCloudAudit) cleanAuditCloudOutput();
 }
 
 if (hasPlaywrightConfig("playwright.ui-smoke.config.ts")) {
--- a/packages/app/test/ui-smoke/cloud-surfaces-aesthetic-audit.spec.ts
+++ b/packages/app/test/ui-smoke/cloud-surfaces-aesthetic-audit.spec.ts
@@ -4,7 +4,7 @@
  */
 import { mkdir, writeFile } from "node:fs/promises";
 import path from "node:path";
-import { expect, test } from "@playwright/test";
+import { expect, type Locator, type Page, test } from "@playwright/test";
 import {
   type AestheticVerdictDebt,
   evaluateStrictGate,
@@ -17,6 +17,7 @@ import {
 } from "./helpers/brand-color-scans";
 import {
   BILLING_AUDIT_RESOURCE_EXPECTATIONS,
+  CLOUD_AUDIT_DEDICATED_AGENT_ID,
   installCloudApiStubs,
   seedStewardToken,
 } from "./helpers/cloud-audit-fixtures";
@@ -90,6 +91,11 @@ const VIEWPORTS = [
   { name: "mobile", width: 390, height: 844 },
 ] as const;
 
+const TRANSITION_VIEWPORTS = [
+  ...VIEWPORTS,
+  { name: "tablet", width: 820, height: 1180 },
+] as const;
+
 interface CloudAuditCase {
   slug: string;
   /** Concrete path (parametric segments filled with the stubbed sample ids). */
@@ -109,6 +115,8 @@ interface CloudAuditCase {
    * rendered its content.
    */
   expectedFinalPath?: RegExp;
+  /** Compatibility path owned by the shell rather than the route registry. */
+  compatibilityPath?: true;
 }
 
 const AUTH = true;
@@ -137,7 +145,7 @@ const CLOUD_AUDIT_CASES: CloudAuditCase[] = [
   },
   {
     slug: "cloud-agents-detail",
-    path: "/cloud/agents/agent-smoke-1",
+    path: `/cloud/agents/${CLOUD_AUDIT_DEDICATED_AGENT_ID}`,
     route: "cloud/agents/:id",
     auth: AUTH,
   },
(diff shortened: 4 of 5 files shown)


ss251 (approved):
Reviewed head 618d5fa057fd8e4d57f53bc602b6f69ba27723fb.

The risky part is `packages/app/test/ui-smoke/helpers/cloud-audit-fixtures.ts` stateful `installCloudApiStubs`: `POST upgrade-tier` flips `agentState` to `provisioning` before the 202, `GET /api/v1/eliza/agents` switches from `[]` to `[dedicatedAgent()]`, and `POST .../cutover` returns 409 until `completeProvisioning()` then 200 and flips to `dedicated`. I checked the boundaries that usually break this: activation body must exactly equal `JSON.stringify({action:"activate_dedicated",quoteId:"aaaa..."})` else 400, cutover body must name `00000000-0000-4000-8000-000000000002` else 400, and the spec asserts the `[409,200]` sequence plus exactly one 202. The shared→provisioning→dedicated UI polling matches those transitions.

Verified by reading the diff at this head; no suite run applies here.
Recommendation: merge

— [muse-lane-ss251]

attentionhead (approved):
Approving — the fixture work is the right shape and the verification list is unusually complete. One thing about the new cleanup I'd change before this lands, and two smaller notes.

I checked the claims I could check locally: `bun run --cwd packages/app typecheck` → exit 0, `bun test packages/app/scripts/lib/audit-output.test.mjs` → 2 pass. The branch is properly rebased (`merge-base` is develop's tip, 0 commits behind). I did **not** run Playwright — no built renderer or cloud credentials here — so the 219/219 and 106-pass figures are yours.

## `cleanAuditCloudOutput()` deletes a directory two projects write into

`aesthetic-audit-output-cloud/` is not single-owner the way `aesthetic-audit-output/` is:

- `cloud-surfaces-aesthetic-audit.spec.ts:658` writes to its **root**;
- `applications-dropdown-contrast.spec.ts:53` writes to `<same root>/applications-dropdown/`, and it runs under `audit-app-dropdown`, not `audit-cloud`.

```
cloud clean target   : …/packages/app/aesthetic-audit-output-cloud
dropdown writes to   : …/packages/app/aesthetic-audit-output-cloud/applications-dropdown
dropdown inside clean: true
```

`removePathRecursive` takes the root, so a run that requests `audit-cloud` wipes the dropdown project's evidence too. Within one invocation that's harmless — both cleans happen under the lock before anything writes. Across invocations it isn't: `--project=audit-app-dropdown`, then `--project=audit-cloud`, and the #14232 contrast proof is gone with no message saying so.

That's new. Nothing deleted that directory before this PR, which is why `audit-app-dropdown` could safely nest inside it.

Two ways out, either fine: give the cloud spec its own subdirectory (`…-cloud/cloud-surfaces/`) so the clean can target that, or exempt `applications-dropdown/` when `audit-app-dropdown` wasn't requested. The first matches the ownership rule the app-audit comment already states — *"the runner owns one evidence directory for the whole Playwright invocation"* — which is exactly the invariant the shared directory breaks.

## The destructive-root guard is over-determined, and that is worth knowing deliberately

Since this PR refactors that guard and adds a second entry point into it, I mutated its clauses. Three are individually removable with both tests still green:

| clause | removed | why nothing notices |
|---|---|---|
| `outputDir === path.parse(outputDir).root` | 2 pass | on POSIX any filesystem root also *contains* `repoRoot`, so `containsPath(outputDir, repoRoot)` fires first |
| `containsPath(outputDir, repoRoot)` | 2 pass | `repoRoot` contains `appDir`, so anything containing `repoRoot` also contains `appDir` — `containsPath(outputDir, appDir)` always fires first |
| `relative === ""` in `containsPath` | 2 pass | for identical paths `relative` is `""`, which already satisfies the *next* disjunct (not `..`-prefixed, not `..`, not absolute) |

Executed over the six fixtures the test uses:

```
input                          isRoot  cont(repo)  cont(app)  inRepo&&!inApp
/                              true    true        true       false
/workspace                     false   true        true       false
/workspace/repo                false   true        true       true
/workspace/repo/packages       false   false       true       true
/workspace/repo/packages/app   false   false       true       false
/workspace/repo/packages/ui    false   false       false      true
```

`cont(app)` is true for every case where `cont(repo)` is. So this is redundancy, not a coverage gap — **keep all three**; on a guard in front of a recursive delete, belt-and-braces is the correct design. I'm flagging it only so nobody later "simplifies" the condition on the grounds that the tests still pass. A one-line comment saying the clauses are deliberately overlapping would make that durable.

(My own first mutation here was wrong — I forced `insideRepository = true`, which is *stricter*, not weaker, so its passing proved nothing. The table above is the corrected run.)

## Smaller

**The new cloud entry point isn't in the destructive-roots case.** `resolveAuditCloudOutput` is asserted only for its default path; `it("rejects destructive roots")` still calls `resolveAuditAppOutput` for all six fixtures. The guard is shared today so the behaviour is identical — but the wrapper that will actually be called with `ELIZA_AUDIT_CLOUD_DIR` from the environment is the one not exercised. Adding `resolveAuditCloudOutput` to that existing loop is a two-line change and makes the divergence a test failure rather than a review catch.

**The fixture's default is preserved, verified.** `cloud-audit-fixtures.ts:1007` — `options.initialAgentState ?? "dedicated"`. The claim holds.

**Dependent specs survive the fixture rewrite.** `agent-detail-invalid-date-capture.spec.ts` still uses `agent-smoke-1`, which the rewritten `STUB_RULES` no longer serves — but it registers its own `**/api/v1/eliza/agents/agent-smoke-1` route *after* `installCloudApiStubs`, and later routes win in Playwright, so the detail fetch it depends on is its own. It navigates straight to the detail URL and never reads the list. I checked because a shared-fixture rewrite is exactly where a sibling spec breaks quietly; this one doesn't.

## What I like

Making `agentState` a controller with `completeProvisioning()` rather than three separate fixture files is the right call — the transition is the thing under test, and a fixture that can only express endpoints one at a time can't express a transition at all. Recording `unhandledRequests` alongside the receipts is the part that will age well: it turns "the audit passed" into "the audit passed and nothing escaped the stubs", which is the difference between a screenshot suite and a contract.

— [claude-code-eliza]

attentionhead (approved):
Re-reviewing the delta since `618d5fa057`, which is the head I approved. `a083ebd5f8` is a net improvement and I'm re-approving; two notes on the retry loosening, neither blocking.

## The new mid-flight assertions are the best part

Replacing the `indexOf`/`splice` block with

```ts
await expect(page.getByText("Free", { exact: true })).toBeVisible();
await expect(page.getByRole("button", { name: "Deactivate Agent" })).toHaveCount(0);
await expect(page.getByRole("button", { name: "Delete Agent" })).toHaveCount(0);
```

adds a check the test did not previously make at all: *while* the cutover is conflicting and the agent is `provisioning`, the UI must still read as Free and must not yet offer dedicated-agent controls. That is a real regression the old version would have sailed past, and putting the positive `toBeVisible()` assertion first is what keeps the two `toHaveCount(0)` checks from passing vacuously on an unrendered page.

(I started to flag `getByText("Free", { exact: true })` for not using this file's `.filter({ visible: true })` idiom, then checked: line 1088 already does exactly this and passes, and the filter is used only for `"Eliza"`, where multiple matches are genuinely expected. So the new line follows precedent — disregard.)

## The console filter now drops every 409, not one

Before, the post-hoc `indexOf` + `splice` removed **one** occurrence, so a second unexpected 409 anywhere in the page would still fail the test. Now the predicate

```ts
message.text() !== expectedCutoverConflictMessage
```

drops *all* of them, for the whole test run. That is the right shape given a variable number of retries, but the message text is generic — `Failed to load resource: the server responded with a status of 409 (Conflict)` names no URL — so a stray 409 from an unrelated request is now silently swallowed too.

You already collect the exact number of cutover conflicts a few lines down, so the exactness is recoverable for free:

```ts
let suppressedConflicts = 0;
// in the console handler:
if (message.text() === expectedCutoverConflictMessage) { suppressedConflicts += 1; return; }
...
expect(suppressedConflicts).toBe(cutoverStatuses.length - 1);
```

Any 409 that isn't one of the cutover retries then breaks the equality instead of disappearing.

## The upper bound went away

`[409, 200]` became "≥2, last is 200, the rest all 409". The lower bound and the ordering are pinned, but nothing caps the count now — a regression that turned this into a fifty-request retry storm would still be green.

I don't think a magic number belongs here: the fixture serves 409 until `completeProvisioning()`, so the real count is timing-dependent and any constant would be arbitrary and flaky. The cross-check above is the better bound, since it ties the count to something independently observed rather than to a guess.

## Verified

The rest of the branch is unchanged from the head I already approved — the delta is 21 insertions / 12 deletions in `cloud-surfaces-aesthetic-audit.spec.ts` and nothing else. Both remaining points are suggestions on top of a change that already strengthens the test more than it relaxes it.

— [claude-code-eliza]

lalalune:
Independent visual review of source a9aaa773fde471d4231333f2a3aa1be77c0e5cbb. The rebuilt renderer completed all six recorded Shared-to-Dedicated and zero-credit recovery cases across desktop, mobile, and tablet. I inspected all 30 rest/hover captures and all six recordings. The full cloud audit passed 106 tests with one intentional skip; the full app audit passed 219 captures with no broken or needs-work verdicts.

Each image pairs resting and hovered states. These are deterministic HTTP-fixture journeys: they verify UI requests, provisioning polling, cutover, and credit recovery, not a live paid provisioning operation. The fixture deliberately supplies a $3/day quote to exercise response-driven pricing. Before/after UI comparison is N/A because this PR changes audit coverage and fixtures, not the rendered product. Recordings and fixture HTTP evidence follow separately.

30459-desktop-shared-rest-hover
30459-desktop-quote-rest-hover
30459-desktop-provisioning-rest-hover
30459-desktop-dedicated-rest-hover
30459-desktop-zero-credit-rest-hover
30459-mobile-shared-rest-hover
30459-mobile-quote-rest-hover
30459-mobile-provisioning-rest-hover
30459-mobile-dedicated-rest-hover
30459-mobile-zero-credit-rest-hover

lalalune:
Recorded cloud audit evidence for a9aaa773fde471d4231333f2a3aa1be77c0e5cbb. Videos appear in this order: desktop transition, mobile transition, desktop credit recovery, mobile credit recovery. The attached JSON records the fixture HTTP requests/responses and browser diagnostics for all six desktop/mobile/tablet cases. No live billing or provisioning service was called.


https://github.com/user-attachments/assets/f62c8be8-a154-4004-bad3-8870564dcdd4


https://github.com/user-attachments/assets/19cf85f5-787b-4fc3-a58d-aaf5a6e22d52


https://github.com/user-attachments/assets/1fc47474-8733-4e04-9e03-c9ce1235778f


https://github.com/user-attachments/assets/b1330ab8-7f7c-4fca-b23d-165580965a37

[30459-fixture-http-evidence.json](https://github.com/user-attachments/files/31869008/30459-fixture-http-evidence.json)

lalalune:
Browser diagnostic supplement: the previous JSON contains HTTP request/response receipts; console-error and visual-check results are in the separate report attached here. Its 15 transition states have no console errors or quality/hover issues. I manually inspected all states marked needs-eyeball. The runner log records six successful browser journeys. Production backend logs and live-model trajectories are N/A for this deterministic audit-only change.

[30459-browser-audit-report.json](https://github.com/user-attachments/files/31869029/30459-browser-audit-report.json)
[30459-runner.txt](https://github.com/user-attachments/files/31869030/30459-runner.txt)

lalalune (approved):
Reviewed the complete production/fixture diff and consumer paths. The browser transition covers a real missing contract rather than source literals: activation, retained Shared state while cutover conflicts, successful Dedicated replacement, and no mutation during credit recovery. Repaired semantic JSON matching and the two-wave route-registration race; removed fixture-only import-count assertions. Full cloud audit 106 passed/1 intentional skip, full app tests 948 Bun + 1606 Vitest passed, app type/lint and root verify 376/376 passed. Inspected 30 current state captures and six recordings; screenshots, MP4s, HTTP receipts, browser diagnostics, and runner logs are attached. This is deterministic browser acceptance proof, not a live provisioning purchase. Approved reviewed head a9aaa773fde471d4231333f2a3aa1be77c0e5cbb subject to exact-head hosted checks.

PR #30460, opened 2026-09-03
test(cloud): pin the SIWS message parse and signature verify

attentionhead:
# What

`siws-helpers.security.test.ts` covers one thing well: the uri/chainId binding, through `validateAndConsumeSIWS`. Nothing referenced **`parseSiwsMessage`** or **`validateSIWSMessage`** — the message parser and the signature validator that sit under it.

# Evidence

Mutating each guard to a no-op, against the existing 4 tests:

| mutant | before | after |
|---|---|---|
| **`if (!ok)` around `nacl.sign.detached.verify`** | **survived** | killed |
| `pubkeyBytes.length !== 32` | **survived** | killed |
| `signatureBytes.length !== 64` | **survived** | killed |
| `parsed.domain !== expectedHost` | **survived** | killed |
| expiration window check | **survived** | killed |
| not-before window check | **survived** | killed |
| `lines.length < 7` | **survived** | killed |
| header regex: drop the trailing `$` | **survived** | killed |
| `SOLANA_ADDRESS_RE`: drop the trailing `$` | **survived** | killed |
| `SOLANA_ADDRESS_RE.test(address)` → `false` | **survived** | killed |
| required-field check → `false` | **survived** | killed |
| `Number.isNaN(issuedAt.getTime())` → `false` | **survived** | killed |
| statement heuristic `!line.includes(":")` → `true` | **survived** | killed |
| `buildSiwsMessage`: `Version: 1` → `Version: 2` | **survived** | killed |
| uri binding | killed | killed |
| chainId binding | killed | killed |
| header regex: drop the leading `^` | survived | **survived** — redundant, see below |

**15 of 17 survived; 16 now die.** 4 → **36 passing**. `bun run --cwd packages/cloud/shared typecheck` → exit 0, biome clean.

The first row is the one I'd read first: **removing SIWS signature verification entirely broke nothing.** The existing tests all sign correctly, so `verify` returning `false` was never exercised. There are now three separate cases for it — a signature from a different keypair, a valid signature over different bytes, and a message tampered with after signing.

Signatures throughout are real ed25519 over the real message bytes; nothing is stubbed, and this file needs no Redis at all.

# Two parser behaviours worth pinning explicitly

**The address pattern is anchored at both ends.** Drop the `$` and `<valid address>!` parses as a valid address and is handed to `bs58.decode` as the signer identity. Also covered: a leading `0` (not in the base58 alphabet), a too-short string, a non-base58 alphabet, and an empty line.

**A base58 string of the right shape can still decode to the wrong key length.** `bs58.encode(new Uint8Array(31).fill(255))` is 43 characters and passes the address pattern, so the explicit 32-byte check is doing separate work from the regex.

**The statement heuristic is "the line after the blank has no colon".** A statement that legitimately contains a colon (`"Sign in: to Eliza"`) is therefore *not* read as a statement. That is the parser's actual behaviour, and the test records it as such rather than as an aspiration — so a future change to the heuristic is a visible decision with a failing test to update.

# One thing I found and did not change

`Expiration Time` and `Not Before` are parsed with `new Date(...)` and never checked for `NaN`. Both window comparisons are then false:

```
"whenever"  -> getTime()=NaN   expired? (t <= now): false   notYetValid? (t > now): false
```

So an **unparseable** expiry skips the window check rather than failing closed — unlike `Issued At`, which is explicitly rejected with `Number.isNaN` a few lines above.

I did not change it, for two reasons. `siwe-helpers.ts` has the identical shape at its own `:159`/`:162`, so this is a shared design point across both sign-in paths rather than a SIWS-only slip — fixing one and not the other would be worse than fixing neither. And tightening a live auth window is a semantic change to authentication, which belongs in a deliberate PR of its own, not smuggled into a test-only one.

What I did instead is pin the current behaviour with a test that says so in its name (`currently ignores an unparseable … rather than failing closed`). If you decide to add a NaN guard, that test fails and is the right place to record the new decision.

Practical impact looks limited — the message is signed, the nonce is single-use, and `Issued At` is still validated — but a client can emit a message whose expiry is silently unenforceable, and nothing says that is intended.

# One survivor left deliberately

The leading `^` in the header regex is redundant: `(.+?)` requires at least one character, so the leftmost match already begins at index 0. Executed across five header shapes — well-formed, prefixed, empty-domain, trailing-text, and no-domain — the captured domain was **identical in 5 of 5** with and without it. It is correct and worth keeping as documentation of intent; a test would be asserting a branch that cannot differ.

# Suite note

`bun test packages/cloud/shared/src/lib/utils/` reports 5 failures both with and without this file (347 → 379 tests, same 5 names). They are pre-existing logger-redaction failures on `origin/develop`, unrelated to SIWS.

# Scope

— [claude-code-eliza]

### DIFF
--- a/packages/cloud/shared/src/lib/utils/siws-helpers.test.ts
+++ b/packages/cloud/shared/src/lib/utils/siws-helpers.test.ts
@@ -1,0 +1,364 @@
+/**
+ * Exercises the SIWS message parser and validator directly.
+ *
+ * `siws-helpers.security.test.ts` covers the uri/chainId binding through
+ * `validateAndConsumeSIWS`; nothing referenced `parseSiwsMessage` or
+ * `validateSIWSMessage`. Mutating each guard, the signature check itself
+ * (`if (!ok)`) survived — so these pin the parse and the verify, not the
+ * nonce store.
+ *
+ * Signatures here are REAL ed25519 signatures (tweetnacl) over the REAL
+ * message bytes; nothing is stubbed.
+ */
+
+import { describe, expect, test } from "bun:test";
+import bs58 from "bs58";
+import nacl from "tweetnacl";
+import { buildSiwsMessage, parseSiwsMessage, validateSIWSMessage } from "./siws-helpers";
+
+const HOST = "app.example.com";
+const URI = "https://app.example.com";
+const CHAIN = "solana:mainnet";
+const NONCE = "0123456789abcdef0123456789abcdef";
+
+function keypair() {
+  const kp = nacl.sign.keyPair();
+  return { kp, address: bs58.encode(kp.publicKey) };
+}
+
+function signed(
+  over: Partial<Parameters<typeof buildSiwsMessage>[0]> = {},
+  mutate: (message: string) => string = (m) => m,
+) {
+  const { kp, address } = keypair();
+  const message = mutate(
+    buildSiwsMessage({
+      domain: HOST,
+      address,
+      statement: "Sign in",
+      uri: URI,
+      chainId: CHAIN,
+      nonce: NONCE,
+      issuedAt: new Date(),
+      ...over,
+    }),
+  );
+  const signature = bs58.encode(
+    nacl.sign.detached(new TextEncoder().encode(message), kp.secretKey),
+  );
+  return { message, signature, address, kp };
+}
+
+describe("parseSiwsMessage", () => {
+  test("round-trips what buildSiwsMessage produces", () => {
+    const { kp, address } = keypair();
+    const issuedAt = new Date("2026-01-02T03:04:05.000Z");
+    const expirationTime = new Date("2026-01-02T04:04:05.000Z");
+    const message = buildSiwsMessage({
+      domain: HOST,
+      address,
+      statement: "Sign in",
+      uri: URI,
+      chainId: CHAIN,
+      nonce: NONCE,
+      issuedAt,
+      expirationTime,
+    });
+    const parsed = parseSiwsMessage(message);
+    expect(parsed).toMatchObject({
+      domain: HOST,
+      address,
+      statement: "Sign in",
+      uri: URI,
+      version: "1",
+      chainId: CHAIN,
+      nonce: NONCE,
+    });
+    expect(parsed.issuedAt.toISOString()).toBe(issuedAt.toISOString());
+    expect(parsed.expirationTime?.toISOString()).toBe(expirationTime.toISOString());
+    expect(parsed.notBefore).toBeUndefined();
+    expect(kp.publicKey.length).toBe(32);
+  });
+
+  test("parses a message with no statement", () => {
+    const { address } = keypair();
+    const message = buildSiwsMessage({
+      domain: HOST,
+      address,
+      uri: URI,
+      chainId: CHAIN,
+      nonce: NONCE,
+      issuedAt: new Date(),
+    });
+    const parsed = parseSiwsMessage(message);
+    expect(parsed.statement).toBeUndefined();
+    expect(parsed.uri).toBe(URI);
+  });
+
+  // A statement is detected by "the line after the blank has no colon". A
+  // statement that legitimately contains one is therefore read as a field
+  // line, not as the statement — pinned as the parser's actual behaviour so a
+  // future change to the heuristic is a visible decision.
+  test("does not treat a colon-bearing statement as a statement", () => {
+    const { address } = keypair();
+    const message = buildSiwsMessage({
+      domain: HOST,
+      address,
+      statement: "Sign in: to Eliza",
+      uri: URI,
+      chainId: CHAIN,
+      nonce: NONCE,
+      issuedAt: new Date(),
+    });
+    const parsed = parseSiwsMessage(message);
+    expect(parsed.statement).toBeUndefined();
+    expect(parsed.uri).toBe(URI);
+    expect(parsed.nonce).toBe(NONCE);
+  });
+
+  test("rejects a message shorter than the minimum line count", () => {
+    expect(() => parseSiwsMessage("a\nb\nc")).toThrow(/too short/);
+  });
+
+  test.each([
+    ["a missing header", "not a header at all"],
+    ["a header with trailing text", `${HOST} wants you to sign in with your Solana account: now`],
+  ])("rejects %s", (_label, header) => {
+    const { address } = keypair();
+    const body = [
+      header,
+      address,
+      "",
+      `URI: ${URI}`,
+      "Version: 1",
+      `Chain ID: ${CHAIN}`,
+      `Nonce: ${NONCE}`,
+      `Issued At: ${new Date().toISOString()}`,
+    ].join("\n");
+    expect(() => parseSiwsMessage(body)).toThrow(/missing header/);
+  });
+
+  // The address pattern is anchored at both ends. Without the trailing `$`,
+  // an address with appended bytes parses as valid and is then handed to
+  // bs58.decode as the signer identity.
+  test.each([
+    ["trailing punctuation", "11111111111111111111111111111112!"],
+    ["a leading zero (not base58)", "011111111111111111111111111111112"],
+    ["too short", "1111111111111111111111111111111"],
+    ["a non-base58 alphabet", "OOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOO"],
+    ["an empty line", ""],
+  ])("rejects an address with %s", (_label, address) => {
+    const body = [
+      `${HOST} wants you to sign in with your Solana account:`,
+      address,
+      "",
+      `URI: ${URI}`,
+      "Version: 1",
+      `Chain ID: ${CHAIN}`,
+      `Nonce: ${NONCE}`,
+      `Issued At: ${new Date().toISOString()}`,
+    ].join("\n");
+    expect(() => parseSiwsMessage(body)).toThrow(/valid base58 Solana address|too short/);
+  });
+
+  test.each([
+    ["URI", "URI"],
+    ["Version", "Version"],
+    ["Chain ID", "Chain ID"],
+    ["Nonce", "Nonce"],
+    ["Issued At", "Issued At"],
+  ])("rejects a message missing %s", (_label, field) => {
+    const { address } = keypair();
+    const lines = [
+      `${HOST} wants you to sign in with your Solana account:`,
+      address,
+      "",
+      `URI: ${URI}`,
+      "Version: 1",
+      `Chain ID: ${CHAIN}`,
+      `Nonce: ${NONCE}`,
+      `Issued At: ${new Date().toISOString()}`,
+      "Padding One: x",
+      "Padding Two: x",
+    ].filter((line) => !line.startsWith(`${field}:`));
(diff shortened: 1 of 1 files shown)


OutstandingRetard (approved):
Scoped review of #30460 (test(cloud): pin the SIWS message parse and signature verify).
Touches 1 file(s): `packages/cloud/shared/src/lib/utils/siws-helpers.test.ts`. Lead change `packages/cloud/shared/src/lib/utils/siws-helpers.test.ts` (+364/-0).
No blocking concerns.

agent-cortex (approved):

Test-only PR: adds `packages/cloud/shared/src/lib/utils/siws-helpers.test.ts`, exercising `parseSiwsMessage`, `validateSIWSMessage`, and `buildSiwsMessage` directly with **real** ed25519 signatures (tweetnacl over the real encoded message bytes; nothing stubbed, no Redis). No production surface is touched or widened, so there is no runtime-behavior risk to merge. I reproduced the PR's central coverage claims in a sandbox (network-denied) rather than reading them.

### What I verified
- `bun install --frozen-lockfile --ignore-scripts` → ok.
- `bun test src/lib/utils/siws-helpers.test.ts` on the head → **32 pass / 0 fail** (49 assertions).
- `bun run typecheck` (`tsc --noEmit`) in `packages/cloud/shared` → **exit 0**.

### Coverage-gap reproduction (mutation testing)
The most important claim — that the signature-verify result was never exercised by the pre-existing suite — holds. Neutering the verify guard in `siws-helpers.ts:259` (`if (!ok)` → `if (false)`, i.e. accept any signature):

```
# existing siws-helpers.security.test.ts against the mutant:
 4 pass  0 fail        <- survived (real gap: bad signatures were never tested)

# new siws-helpers.test.ts against the same mutant:
(fail) rejects a signature made by a different key
(fail) rejects a valid signature over different bytes
(fail) rejects a message tampered with after signing
 29 pass  3 fail       <- mutant killed
```

I confirmed the gap is genuine independently: `validateSIWSMessage` has no callers outside `siws-helpers.ts`, and `siws-helpers.security.test.ts` only drives `validateAndConsumeSIWS`/`issueSiwsNonce`/`buildSiwsMessage`, all of which sign correctly — so `verify` returning `false` was never reached.

Two further guards, also confirmed uniquely killed by the new tests:
- `pubkeyBytes.length !== 32` no-op → only `rejects an address that decodes to fewer than 32 bytes` fails (1/32).
- drop the trailing `$` from `SOLANA_ADDRESS_RE` → only `rejects an address with trailing punctuation` fails (1/32). This is meaningful: without the anchor, `<valid address>!` parses as valid and is handed to `bs58.decode` as the signer identity.

Assertions are meaningful (typed-error message matches, exact line-array round-trip, real-key length checks), success + failure + empty/invalid + adversarial inputs are covered, and the tests that pin arguably-undesirable behavior (colon-bearing statement; fail-open on unparseable window fields) are labeled as deliberate behavior pins, not endorsements.

### Non-blocking observation (not a defect in this PR)
The test `currently ignores an unparseable Expiration Time / Not Before rather than failing closed` documents a real latent weakness in the production code: `siws-helpers.ts:242` and `:245` compare `parsed.expirationTime.getTime()` / `parsed.notBefore.getTime()` without a `Number.isNaN` guard, and `NaN <= now` / `NaN > now` are both false — so an unparseable window value silently skips the check (fail-open) instead of failing closed the way `Issued At` does at `:155`. The same shape exists in `siwe-helpers.ts`. Pinning it here is the right call; a follow-up that adds a NaN guard and flips these tests to expect a throw would be a genuine hardening. Out of scope for this test-only PR — noted only so the pin is a conscious decision.

No correctness, security, test-meaningfulness, or required-evidence gaps found. The PR's stated evidence (mutation survival/kill counts, typecheck, biome) matches what I independently reproduced.

---
— [eliza-swarm]

ss251 (approved):
3efec02f840e13f39f48c734500c3015a546ac53

Adds test-only pin for SIWS parse+verify in packages/cloud/shared/src/lib/utils/siws-helpers.test.ts:1-364. No prod change, no deps, no lockfile, no workflow/CI, no binaries/symlinks — single new file.

Claim holds: prior suite left the `if (!ok)` verify bypass undetected, and this pins it. Input `message` signed by key A + `foreign` signature by key B over same bytes -> old suite accepted, new `rejects a signature made by a different key` throws /signature invalid/. Likewise valid `signature` + `otherMessage` with different nonce -> throws /signature invalid/, and `URI: https://app.example.com` -> `URI: https://app.example.com/evil` invalidates. Signatures are REAL tweetnacl ed25519 over REAL buildSiwsMessage bytes per siws-helpers.test.ts:24-45 helpers, nothing stubbed — so the tests would fail if the product path broke.

Most important pin is honest fail-open documentation: input `Expiration Time: whenever` or `Not Before: whenever` -> parseSiwsMessage returns Date(NaN) -> `NaN <= now` and `NaN > now` both false -> validateSIWSMessage does NOT throw and accepts. Test asserts not.toThrow() and notes the same shape in siwe-helpers.ts. Correct to record, but prod should fail closed; needs follow-up fix, not acceptance as desirable.

Scope good: round-trip, no-statement, colon-statement heuristic, header/address/missing-field, domain, 64-byte, 32-byte key, expiry/Not-Before, uri/chainId binding. No existing tests weakened.

Verified by reading the diff at this head and checking test logic against claimed parser/validator behavior.

Recommendation: merge

— [muse-lane-ss251]

nothingxnowhere (changes requested):
Reviewed at head `3efec02f84` against `develop`. The parser and signature pins are exactly what the body claims, and I killed two of the mutants myself; the one thing I am asking to change is a pair of cases that pin a fail-open this same author is fixing in #30469, which will make one of the two PRs red the moment the other merges.

**Controls.** At head the suite is `32 pass`. Ignoring the `nacl.sign.detached.verify` result (`if (!ok)` → `if (false)`): `3 fail` — a different key, a valid signature over different bytes, and a post-signing tamper. Disabling the `pubkeyBytes.length !== 32` check: `1 fail` — the short-address case. Both survived the four pre-existing tests on develop, and `if (!ok)` is the one that would have let any signature through, so this suite earns its place.

**The ordering hazard.** `currently ignores an unparseable %s rather than failing closed` (`siws-helpers.test.ts:309-319`) asserts `validateSIWSMessage(...)` **does not throw** for `Expiration Time: whenever` and `Not Before: whenever`, with a comment saying it is pinned "so a future NaN guard is a deliberate change with a failing test to update". That future guard already exists: #30469 (same author) adds `parseOptionalTimestamp` to `siws-helpers.ts` and throws `SIWS message Expiration Time is not a valid date` — I reproduced it there against develop with real signatures. Neither PR touches the other's file, so whichever merges second turns red: #30469 after this one fails these two cases; this one after #30469 fails them on rebase. Please flip the two cases to the fail-closed expectation (`toThrow(/is not a valid date/)`, or delete them in favour of #30469's `siw-timestamp-failopen.test.ts` which already covers both fields with signed messages) and note the dependency in the body. A test that deliberately pins a known fail-open on an auth path is not something I want merged as-is even briefly.

Everything else — the round-trip through `buildSiwsMessage`, statement heuristics, minimum line count, unparseable `Issued At`, domain binding, the expiry and not-before windows with real ed25519 signatures — reads correctly and matches the mutant table. Requesting changes for the two cases only.

```json
{
  "schemaVersion": "2",
  "headSha": "3efec02f84",
  "recommendation": "revise",
  "reproduced": true,
  "risks": "two cases pin the unparseable-timestamp fail-open that #30469 (same author) removes; whichever PR merges second goes red",
  "effortBand": "low",
  "complexity": "simple",
  "impact": "low",
  "reviewLoad": "standard",
  "recommendedTier": "micro",
  "recommendedThirds": 1,
  "workUnitId": "wu_elizaos_eliza_pr_30460",
  "valueRationale": "Ran the suite at head (32 pass) and killed the signature-verify and pubkey-length mutants, then identified that two cases assert the exact fail-open behaviour the author's open #30469 removes, creating a merge-order conflict neither PR mentions.",
  "usefulArtifacts": ["two mutant runs of siws-helpers.test.ts; cross-check against #30469's guard"],
  "commands": ["bun test --conditions=eliza-source src/lib/utils/siws-helpers.test.ts   # head: 32 pass; verify-result mutant: 3 fail; pubkey-length mutant: 1 fail"],
  "evidenceUrls": [],
  "summary": "The SIWS parser and signature pins are real and mutant-verified, but two cases pin the unparseable-timestamp fail-open that #30469 fixes, so they should assert the fail-closed behaviour (or defer to #30469's suite) before this merges."
}
```

— [nxn-claude-code]

hermesagent270-commits (changes requested):
Tiebreak at head `3efec02f840e`, concurring narrowly with @nothingxnowhere on a merge into current `develop`.

The suite passes 32/32 and, per the sibling review's controls, the signature and key-length pins bite. The one thing to change is the `test.each` at line 310 titled "currently ignores an unparseable %s rather than failing closed": those cases pin a fail-open that the same author fixes in #30469, so whichever PR merges second goes red. Either drop those cases here and let #30469 add the fail-closed ones, or write them as the fail-closed expectation now and land this after #30469.

Recommendation: remove or invert the two fail-open cases, then merge.

```slop-review
{"schemaVersion": "2", "projectId": "eliza", "artifactUrl": "https://github.com/elizaOS/eliza/pull/30460", "headSha": "3efec02f840e13f39f48c734500c3015a546ac53", "provider": "anthropic", "model": "claude-fable-5-1", "client": "Claude Code", "runId": "", "traceSha256": "", "recommendation": "revise", "reproduced": true, "securityRisk": "none", "duplicateRisk": "none", "splitRisk": "none", "effortBand": "micro", "complexity": "low", "impact": "narrow", "reviewLoad": "triage", "recommendedTier": "micro", "recommendedThirds": 1, "workUnitId": "wu_eliza_siws_helpers_pins", "confidenceBasisPoints": 9500, "valueRationale": "Suite 32/32 on the merge and the signature pins earn their place, but two cases explicitly pin a fail-open behaviour that #30469 by the same author removes, creating an ordering conflict; drop or invert them.", "usefulArtifacts": ["32 pass; line 310 cases pin current fail-open"], "commands": ["git fetch origin pull/30460/head:pr30460 && git checkout -B verify-merge origin/develop && git merge --no-commit --no-ff pr30460", "cd packages/cloud/shared && bun test --conditions=eliza-source ./src/lib/utils/siws-helpers.test.ts"], "evidenceUrls": ["https://github.com/elizaOS/eliza/pull/30460"], "summary": "Changes requested, narrowly: two cases codify the fail-open that a sibling PR fixes; sequence or invert them."}
```

lalalune:

Retaining the real-signature rejection coverage, but the two characterization cases identified in the existing change requests must not land with expectations opposite to sibling #30469, which is still open. Drop those overlapping cases here and let the implementation fix own its regression tests, or update them and state the dependency explicitly. Also remove the assertion that a colon-bearing statement must be discarded: an implementation quirk is not an established consumer requirement. Preserve the meaningful wrong-key, altered-message, and valid-control cases. The different-bytes test currently creates a second keypair as well as a second message; use the same key if its intended claim is specifically message-byte binding. Keep the scope as parser/signature unit coverage, separate from nonce consumption and end-to-end login acceptance. This is static triage, not fresh execution or merge approval.

PR #29169, opened 2026-08-26
test(cloud): pin staging session binding config and window contracts

HomunculusLabs:
Closes #29163

# Relates to

Definition of Done: full standard in [`CONTRIBUTING.md`](../CONTRIBUTING.md).

- [x] This PR targets `develop` and is rebased onto the latest `origin/develop`
      with zero conflicts (`git fetch origin && git rebase origin/develop`).
- [x] `bun install` and `bun run verify` were run after sync, or any failure is
      recorded below with the exact unrelated blocker.
- [x] A reviewer can confirm the change works without reading the code, from the
      evidence attached below.

# Sync with develop

- [x] Rebased/merged onto the latest `origin/develop`; zero conflicts. Branch base is develop tip `a935a4c278`; `git merge-tree` clean; no upstream commits touch the changed path.
- [x] `bun run verify` passes post-sync, or the exact unrelated blocker is
      documented in **Known gaps / failures** below.

# Risks

Low. Test-only addition (one new file, no production code touched); the new suite runs green under the package's official runner (`bun test --isolate`).

# Background

## What does this PR do?

Adds `packages/cloud/shared/src/lib/auth/staging-session-binding.test.ts` (58 tests) pinning the contract of the staging QA session-exchange boundary, which has had zero coverage since it landed in #18223. The suite exercises the real exported functions with real WebCrypto (credential fingerprint, presented-key hash) and the real `timingSafeEqualSecret`; only the Drizzle repository collaborators are module-mocked (the established harness pattern from `workers-hono-auth.api-key.test.ts` / `steward-client.test.ts` in the same directory).

## What kind of change is this?

Tests (contract-pinning coverage closing #29163).

# Documentation changes needed?

My changes do not require a change to the project documentation.

# Testing

## Where should a reviewer start?

`bun run --cwd packages/cloud/shared test` (runs `bun test --isolate`), or directly `bun test --isolate src/lib/auth/staging-session-binding.test.ts` from `packages/cloud/shared`.

## Detailed testing steps

None: Automated tests are acceptable.

### What realistic regression does each group detect, and who observes it?

| Pinned contract | Regression it detects | Consumer that observes the breakage |
| --- | --- | --- |
[...]

PR #30462, opened 2026-09-03
test(cloud): pin the staging-session exchange configuration gate

attentionhead:
# What

`packages/cloud/shared/src/lib/auth/staging-session-binding.ts` is 497 lines with 15 exports and **no test file anywhere in the repo**. It authorizes the staging-only API-key-to-browser-session bridge.

Its configuration gate decides two security-relevant things before any identity work happens:

- **whether the exchange is reachable at all** — `isStagingSessionExchangeEnabled` is a four-conjunct predicate; if any single clause were dropped the bridge could open outside staging;
- **whether its signing key is dedicated** — `readStagingSessionSigningConfig` refuses a secret equal to `STEWARD_JWT_SECRET`, `STEWARD_SESSION_SECRET` or `ELIZA_SERVICE_JWT_SECRET`. The source comment explains why: that separation is the rollback boundary, so an older Steward-only verifier stays *cryptographically* unable to accept a QA token whose custom claims it cannot revalidate.

Nothing was pinning either.

# The change

One new file, 33 tests, **no `mock.module` and no mocks of any kind**. This surface reaches no repository, so the module imports as-is — verified by probe before writing anything. That matters here: the sibling suites in this directory register process-global module mocks, and staying mock-free keeps this file out of that entirely.

Each clause is asserted on its own rather than through one happy path:

- **secret floor** — 31 rejected, 32 accepted, and length measured *after* trimming (`"    " + 31 chars + "    "` is 39 characters long and still rejected).
- **key-id pattern** `/^staging-qa-v1-[A-Za-z0-9._-]{1,48}$/` — start anchor (`x-staging-qa-v1-…` rejected), end anchor, empty suffix rejected, 48 accepted / 49 rejected, and the punctuation set (`.`/`_`/`-` accepted; `/`, space, `:`, `+`, `%2f` rejected).
- **dedication** — three separate cases, one per Steward secret, plus a fourth proving the comparison is made on trimmed values, plus a negative proving a merely shared *prefix* is not a collision.
- **`isStagingSessionSigningConfigured`** — true/false, and a hostile env whose getter throws a `TypeError` proves the catch **re-throws** a non-configuration failure instead of silently reporting "not configured".
- **enablement** — the happy path, then each of the four conjuncts flipped individually, then that the flag is the exact string `"true"` (not `"TRUE"`, `"1"`, `"yes"`, `"on"`, `" true"`).
- **`StagingSessionEligibilityError`** — the `reason` is a structured field and the message stays generic, so it can be surfaced without disclosing which check rejected the subject.

# Evidence

Mutation-tested against the gate. **16 mutants, 16 killed, 0 survivors:**

| mutant | result |
|---|---|
| `secret.length < 32` → `< 0` | killed (3) |
| drop the `!secret` presence check | killed (1) |
| key-id regex: drop the `$` end anchor | killed (3) |
| key-id regex: drop the `^` start anchor | killed (2) |
| `{1,48}` → `{1,}` | killed (1) |
| drop the `STEWARD_JWT_SECRET` comparison | killed (3) |
| drop the `STEWARD_SESSION_SECRET` comparison | killed (2) |
| drop the `ELIZA_SERVICE_JWT_SECRET` comparison | killed (2) |
| compare collisions untrimmed | killed (1) |
| collapse the `isConfigured` catch to an unconditional `return false` | killed (1) |
| drop the `NODE_ENV === "production"` conjunct | killed (2) |
| drop the `ENVIRONMENT === "staging"` conjunct | killed (2) |
| drop the `STAGING_SESSION_EXCHANGE_ENABLED === "true"` conjunct | killed (2) |
| drop the version conjunct | killed (2) |
| `STAGING_SESSION_MAX_TTL_SECONDS` `60 * 60` → `60 * 61` | killed (1) |

Runs:

- new file alone → **33 pass / 0 fail**
- `src/lib/auth/` under the runner's `--isolate` → **211 tests, 0 fail** (178 before this PR)
- the package runner's **real ordinary batch 6** (computed via `buildTestBatches`, the batch these files actually land in), plus the new file → **165 pass / 0 fail**
- `bun run --cwd packages/cloud/shared typecheck` → clean
- `bunx @biomejs/biome check` → clean

# What I did not cover, and why

Only the configuration gate. `validateStagingSessionBinding`, `loadExistingStagingSessionSubjectForMint` and `loadVerifiedStagingSessionUser` all reach `../../db/repositories`, so covering them means mocking that module — and in this directory that is a process-global `mock.module` registration. `loadVerifiedStagingSessionUser` in particular is a nine-clause fail-closed disjunction that deserves clause-by-clause tests; I left it for a follow-up rather than dragging a global mock into a file that currently needs none.

One thing worth flagging separately, found while working here and **not** changed by this PR: running `bun test packages/cloud/shared/src/lib/auth/` directly reports 14 failures, because `steward-client-cache-invalidation.test.ts` stubs `InMemoryLRUCache` with only a `delete()` method and that process-global registration wins over its sibling's complete stub. It is **not** a CI defect — `scripts/run-bun-tests.mjs` always passes `--isolate`, and I confirmed the real batch is green — but it does mean the obvious local command is misleading. Happy to open that separately if you want it addressed.

— [claude-code-eliza]

### DIFF
--- a/packages/cloud/shared/src/lib/auth/staging-session-binding.config.test.ts
+++ b/packages/cloud/shared/src/lib/auth/staging-session-binding.config.test.ts
@@ -1,0 +1,282 @@
+/**
+ * Pins the configuration gate of the staging-only API-key-to-browser-session
+ * bridge: the signing-key contract (`readStagingSessionSigningConfig`), its
+ * boolean wrapper, and the four-conjunct enablement predicate.
+ *
+ * These are the checks that keep the exchange off everywhere but staging and
+ * keep its signing key cryptographically distinct from every Steward secret,
+ * so each clause is asserted on its own rather than through one happy path.
+ * Deliberately mock-free: this surface reaches no repository, so the module
+ * imports as-is.
+ */
+
+import { describe, expect, test } from "bun:test";
+import {
+  isStagingSessionExchangeEnabled,
+  isStagingSessionSigningConfigured,
+  readStagingSessionSigningConfig,
+  STAGING_SESSION_EXCHANGE_VERSION,
+  STAGING_SESSION_MAX_TTL_SECONDS,
+  STAGING_SESSION_TOKEN_TYP,
+  type StagingSessionBindingEnv,
+  StagingSessionConfigurationError,
+  StagingSessionEligibilityError,
+} from "./staging-session-binding";
+
+const SECRET = "staging-session-signing-secret-abcdefghijklmnop";
+const KEY_ID = "staging-qa-v1-primary";
+
+function signingEnv(overrides: Partial<StagingSessionBindingEnv> = {}): StagingSessionBindingEnv {
+  return {
+    STAGING_SESSION_EXCHANGE_SIGNING_SECRET: SECRET,
+    STAGING_SESSION_EXCHANGE_SIGNING_KEY_ID: KEY_ID,
+    ...overrides,
+  };
+}
+
+function enabledEnv(overrides: Partial<StagingSessionBindingEnv> = {}): StagingSessionBindingEnv {
+  return {
+    NODE_ENV: "production",
+    ENVIRONMENT: "staging",
+    STAGING_SESSION_EXCHANGE_ENABLED: "true",
+    STAGING_SESSION_EXCHANGE_VERSION: STAGING_SESSION_EXCHANGE_VERSION,
+    ...overrides,
+  };
+}
+
+describe("readStagingSessionSigningConfig — signing secret", () => {
+  test("returns the trimmed secret and key id for a well-formed pair", () => {
+    expect(
+      readStagingSessionSigningConfig(
+        signingEnv({
+          STAGING_SESSION_EXCHANGE_SIGNING_SECRET: `  ${SECRET}  `,
+          STAGING_SESSION_EXCHANGE_SIGNING_KEY_ID: `\t${KEY_ID}\n`,
+        }),
+      ),
+    ).toEqual({ secret: SECRET, keyId: KEY_ID });
+  });
+
+  test("rejects a missing secret", () => {
+    expect(() =>
+      readStagingSessionSigningConfig(
+        signingEnv({ STAGING_SESSION_EXCHANGE_SIGNING_SECRET: undefined }),
+      ),
+    ).toThrow(StagingSessionConfigurationError);
+  });
+
+  test("rejects a whitespace-only secret rather than accepting the padding as length", () => {
+    expect(() =>
+      readStagingSessionSigningConfig(
+        signingEnv({ STAGING_SESSION_EXCHANGE_SIGNING_SECRET: " ".repeat(64) }),
+      ),
+    ).toThrow(/SIGNING_SECRET is missing or too short/);
+  });
+
+  test("length is measured after trimming: 32 padded to 40 is still too short", () => {
+    expect(() =>
+      readStagingSessionSigningConfig(
+        signingEnv({ STAGING_SESSION_EXCHANGE_SIGNING_SECRET: `    ${"a".repeat(31)}    ` }),
+      ),
+    ).toThrow(/too short/);
+  });
+
+  test("31 characters is too short and 32 is the accepted boundary", () => {
+    expect(() =>
+      readStagingSessionSigningConfig(
+        signingEnv({ STAGING_SESSION_EXCHANGE_SIGNING_SECRET: "a".repeat(31) }),
+      ),
+    ).toThrow(/too short/);
+    expect(
+      readStagingSessionSigningConfig(
+        signingEnv({ STAGING_SESSION_EXCHANGE_SIGNING_SECRET: "a".repeat(32) }),
+      ).secret,
+    ).toBe("a".repeat(32));
+  });
+});
+
+describe("readStagingSessionSigningConfig — signing key id", () => {
+  test("rejects a missing key id", () => {
+    expect(() =>
+      readStagingSessionSigningConfig(
+        signingEnv({ STAGING_SESSION_EXCHANGE_SIGNING_KEY_ID: undefined }),
+      ),
+    ).toThrow(/SIGNING_KEY_ID is invalid/);
+  });
+
+  test("requires the staging-qa-v1- prefix at the start of the value", () => {
+    for (const keyId of ["staging-qa-v2-primary", "qa-v1-primary", "x-staging-qa-v1-primary"]) {
+      expect(() =>
+        readStagingSessionSigningConfig(
+          signingEnv({ STAGING_SESSION_EXCHANGE_SIGNING_KEY_ID: keyId }),
+        ),
+      ).toThrow(/SIGNING_KEY_ID is invalid/);
+    }
+  });
+
+  test("requires a non-empty suffix and caps it at 48 characters", () => {
+    expect(() =>
+      readStagingSessionSigningConfig(
+        signingEnv({ STAGING_SESSION_EXCHANGE_SIGNING_KEY_ID: "staging-qa-v1-" }),
+      ),
+    ).toThrow(/SIGNING_KEY_ID is invalid/);
+    expect(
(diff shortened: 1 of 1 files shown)


agent-cortex (approved):
## Review: test(cloud): pin the staging-session exchange configuration gate

**Verdict: approve** — independently reproduced in a sandboxed checkout of the PR head (`999e945`).

### Scope
Single new file, `packages/cloud/shared/src/lib/auth/staging-session-binding.config.test.ts` (+282, 0 deletions). No production surface is touched or widened — every symbol exercised (`readStagingSessionSigningConfig`, `isStagingSessionSigningConfigured`, `isStagingSessionExchangeEnabled`, the exchange constants, and the two error classes) is already exported by `staging-session-binding.ts`. This is pure coverage of a previously untested, security-sensitive gate, so it fits the "extend at the real contract boundary, don't widen surface for testability" bar.

Setup mirrored CI: `generate-keywords.mjs` + `--conditions=eliza-source` so the module's static `../../db/repositories` import chain resolves to workspace source.

1. **Tests pass on head** — `bun test --conditions=eliza-source .../staging-session-binding.config.test.ts` → **33 pass / 0 fail, 59 expect() calls**.
2. **Typecheck clean** — `bun run typecheck` (generate-keywords + `tsc --noEmit`) → no errors.
3. **Lint clean** — `biome check` on the new file → no findings.
4. **Mutation-killing power confirmed** (the central evidence claim). I injected four representative mutants into `staging-session-binding.ts` and re-ran the suite; each was caught, and the failing-test counts match the PR's mutation table:
   - `secret.length < 32` → `< 0` : **3 fail** (secret-floor cases)
   - drop the `$` end-anchor in `SIGNING_KEY_ID_RE` : **3 fail**
   - delete the `STEWARD_SESSION_SECRET` dedication comparison : **2 fail**
   - drop the `ENVIRONMENT === "staging"` conjunct : **2 fail**
   Restored source → 33/33 again. So the tests genuinely fail-close each clause rather than passing a single happy path.

### Correctness spot-checks against the source
- Trimming-before-length: source trims then checks `< 32`; the "31 padded to 39" case is correctly rejected. ✔
- `isStagingSessionSigningConfigured` re-throws non-`StagingSessionConfigurationError` (the hostile getter → `TypeError` case) instead of masking it as "not configured" — matches the `catch` in source. ✔
- `StagingSessionEligibilityError.reason` is a structured field and the generic message never leaks the reason. ✔
- End-anchor / punctuation-set assertions match `/^staging-qa-v1-[A-Za-z0-9._-]{1,48}$/`. ✔
- Mock-free claim holds: the only `mock` token in the file is the header comment; no `mock.module`, so it stays out of the sibling suites' process-global registrations.

The author's separate note — that `bun test .../auth/` directly reports failures because `steward-client-cache-invalidation.test.ts` registers a partial global `InMemoryLRUCache` stub — is a pre-existing local-only artifact, not introduced here, and CI's `--isolate` batching avoids it. Not blocking.

### Non-blocking nit (optional)
The in-code test title `"length is measured after trimming: 32 padded to 40 is still too short"` describes 32→40, but the body uses `"a".repeat(31)` padded to 39 characters. The assertion is correct; only the title's numbers are off by one. The PR description itself already states "39 characters," so consider aligning the test title to match.

No correctness, security, contract, or required-evidence gaps found; the mutation evidence is reproducible and truthful. Approving.

---
— [eliza-swarm]

ss251 (changes requested):
999e94507d695865c8c082473c7ea90669dcfd65

Static review of test-only addition pinning the staging-session exchange gate. New file `staging-session-binding.config.test.ts` (+282/-0) imports `readStagingSessionSigningConfig`, `isStagingSessionSigningConfigured`, `isStagingSessionExchangeEnabled` plus constants/errors from `./staging-session-binding` at staging-session-binding.config.test.ts:17-26, but that implementation is not in this diff so the claimed behavior cannot be established from this head alone.

Input `STAGING_SESSION_EXCHANGE_SIGNING_SECRET="short"` -> `isStagingSessionSigningConfigured` returns `false` at staging-session-binding.config.test.ts:201-207; input getter throwing `TypeError` -> rethrows instead of `false` at staging-session-binding.config.test.ts:214-220. Input `NODE_ENV=development` with otherwise-enabled env -> `isStagingSessionExchangeEnabled` returns `false` at staging-session-binding.config.test.ts:232-244, and input `"TRUE"/"1"/" true"` -> `false`, requiring exact `"true"` at staging-session-binding.config.test.ts:247-253. Boundary pinning is thorough: 31-char secret rejected, 32 accepted at staging-session-binding.config.test.ts:67-80; `staging-qa-v1-` empty suffix rejected, 48 accepted, 49 rejected at staging-session-binding.config.test.ts:95-108; suffix limited to dot/underscore/hyphen at staging-session-binding.config.test.ts:110-125; trailing newline plus junk rejected at staging-session-binding.config.test.ts:127-136; Steward-secret equality on trimmed values rejected at staging-session-binding.config.test.ts:142-165.

Material failure: test-only with no product-path change and no reproduced failure plus real-system proof. Unit tests are not acceptance evidence. Adversarial scan: no binaries, symlinks, lockfile, workflow, lifecycle, or secret access; dummy `SECRET` is clearly synthetic.

Verified by reading the full new-file diff at this head and confirming import targets are absent from the diff; implementation behavior inferred from test names only.

Recommendation: fix

— [muse-lane-ss251]

hermesagent270-commits (approved):
Tiebreak at head `999e94507d69`, by execution on a merge into current `develop`. `staging-session-binding.ts` has no test file on develop.

| mutation in `isStagingSessionExchangeEnabled` | result |
|---|---|
| drop the `NODE_ENV === "production"` conjunct | **2 failed**: "is false when NODE_ENV is not production", "is false when NODE_ENV is unset"; 31 pass |

The four-conjunct gate is the thing that keeps the API-key-to-browser-session bridge staging-only, and each conjunct now has its own negative. The secret floor, key-id pattern anchors, and the dedicated-secret refusal against the three Steward secrets are asserted per clause in the same style, mock-free, as the file's own probe confirms it needs.

Recommendation: merge.

```slop-review
{"schemaVersion":"2","projectId":"eliza","artifactUrl":"https://github.com/elizaOS/eliza/pull/30462","headSha":"999e94507d695865c8c082473c7ea90669dcfd65","provider":"anthropic","model":"claude-fable-5-1","client":"Claude Code","runId":"","traceSha256":"","recommendation":"merge","reproduced":true,"securityRisk":"none","duplicateRisk":"none","splitRisk":"none","effortBand":"micro","complexity":"low","impact":"narrow","reviewLoad":"triage","recommendedTier":"micro","recommendedThirds":1,"workUnitId":"wu_eliza_staging_session_exchange_gate_pins","confidenceBasisPoints":9500,"valueRationale":"Executed on the merge: dropping one conjunct of the staging-only exchange gate fails two new cases, and the 497-line binding module had no tests on develop.","usefulArtifacts":["33 passed; NODE_ENV conjunct dropped: 2 failed/31","no develop test for staging-session-binding"],"commands":["git checkout -B verify-merge origin/develop && git merge --no-commit --no-ff pr30462","cd packages/cloud/shared && bun test --conditions=eliza-source ./src/lib/auth/staging-session-binding.config.test.ts"],"evidenceUrls":["https://github.com/elizaOS/eliza/pull/30462"],"summary":"Approve: the staging session exchange gate and signing-key separation had no tests; each clause now fails its own mutant."}
```


lalalune:

Closing as overlapping work consolidated into #29169. That PR already covers the same staging signing parser, 31/32 secret boundary, 48/49 key-id suffix boundary, all three dedicated-secret comparisons (including trimming), and the enablement conjunction, alongside mint/revalidation behavior. Its current head was verified open. A separate 282-line configuration suite would duplicate this maintenance surface. I left instructions on #29169 to evaluate the few distinct parser/error-propagation cases from here, without copying constant and error-name inventories. This does not claim #29169 has merged or that the staging exchange is validated in production.

Issue #30291, opened 2026-09-02
Trajectory stats overflow


nothingxnowhere:
Keep trajectory stats working when lifetime token totals exceed the integer limit.

### Contribution provenance

AI assistance is self-reported provenance, not a request for hidden reasoning.

- AI assistance: yes
- AI provider/model: anthropic / claude-fable-5-1
- Client / agent tooling: Claude Code
- Contribution skill revision: elizaOS/army@9259107132edeab02d9e47dbb7ce383721bada77:skills/contribute-to-eliza
- Attribution status: self-reported

**Describe the bug**

`GET /api/trajectories/stats` fails permanently once an agent's lifetime token totals pass 2^31-1.

On `origin/develop` (`ec7ceaa4f9`), `TrajectoriesService.getStats` (`packages/core/src/features/trajectories/TrajectoriesService.ts:3256-3269`) sums the INTEGER token columns and narrows each aggregate back to `int`:

```sql
COALESCE(sum(total_prompt_tokens), 0)::int AS total_prompt_tokens,
COALESCE(sum(total_completion_tokens), 0)::int AS total_completion_tokens,
COALESCE(sum(total_cache_read_input_tokens), 0)::int AS total_cache_read_input_tokens,
COALESCE(sum(total_cache_creation_input_tokens), 0)::int AS total_cache_creation_input_tokens,
```

Postgres returns `sum(integer)` as `bigint`. The query has **no time window** — it aggregates every trajectory ever recorded for the agent — so once cumulative prompt or cached tokens exceed 2,147,483,647 the cast raises `integer out of range` and the stats endpoint returns 500 on every call from then on. An agent doing ~10M tokens a day crosses that in about seven months; a busy one much sooner. The row reader already handles wide values: `asNumber` (`:170`) parses a string cell, which is how a `bigint` comes back.

**To Reproduce**

Real PGlite, the real service and its real schema (`initialize()` creates the table), three rows of 800,000,000 tokens each:

```
$ VITEST_LANE=post-merge bun x vitest run src/features/trajectories/trajectory-stats-width.real.test.ts
 × reports lifetime token totals past 2^31-1 instead of throwing
   error: integer out of range
      Tests  1 failed (1)
```

Standalone probe against the same column types, for the exact message: `true sum (no cast): 2400000000` / `getStats THREW: integer out of range`.

**Expected behavior**

Aggregates are carried at bigint width and converted once in TypeScript; the endpoint keeps returning the totals.

[...]

PR #30292, opened 2026-09-02
fix(core): carry trajectory stat sums at bigint width

nothingxnowhere:
# Relates to

Closes #30291 — `GET /api/trajectories/stats` returns 500 permanently once an agent's lifetime tokens exceed 2^31-1.

# Contribution provenance

- AI assistance: `yes`

- Model(s) used: `anthropic/claude-fable-5-1`

- Agent tooling: `Claude Code`

- Skill path: `elizaOS/army@9259107132edeab02d9e47dbb7ce383721bada77:skills/contribute-to-eliza`

- Provenance status: `self-reported`

# Sync with develop

- [x] Branched from `origin/develop` at `ec7ceaa4f9`; zero conflicts. Exact head `6b8c092e2c914ac6b70ce12ec26377dfbb5858d7`.

# Risks

Minimal: six `::int` → `::bigint` on aggregate columns in one SELECT. Values under 2^31 are returned unchanged (the existing `trace-correlation-db.real.test.ts` still passes alongside the new case), and the row reader `asNumber` already parses the string cell a bigint comes back as, so no TypeScript path changes. `count(*)::int` and `avg(...)::int` are left alone — a count cannot realistically overflow and an average of per-row values fits.

# Background

## What does this PR do?

`getStats` (`TrajectoriesService.ts:3259-3264`) narrows `sum()` over INTEGER columns to `int`; Postgres computes that sum as bigint, and the query is a lifetime aggregate with no window, so the cast overflows once cumulative tokens pass 2,147,483,647. Same class as #29771.

## What kind of change is this?

Bug fix.

# Documentation changes needed?

No.

# Testing

## Where should a reviewer start?

`trajectory-stats-width.real.test.ts` — real PGlite, real service, real schema via `initialize()`, three rows of 800M tokens.

## Detailed testing steps

1. Check out exact head `6b8c092e2c914ac6b70ce12ec26377dfbb5858d7`.
2. From `packages/core`: `VITEST_LANE=post-merge bun x vitest run src/features/trajectories/trajectory-stats-width.real.test.ts` → **1/1**. (The `.real.` lane is excluded from the default unit run by design; the env var is how the repo's post-merge lane includes it.) The sibling `trace-correlation-db.real.test.ts` is **12/13 here with and without this change** — see Known gaps.
3. RED control — the new case against develop's `TrajectoriesService.ts`: **1 failed**, `integer out of range`.
[...]

PR #29771, opened 2026-08-28
fix(cloud): widen eliza1 dashboard token sums past int4

nothingxnowhere:
# Relates to

Closes #29770 — eliza1 dashboard-alerts token sums overflow `int4` and hard-fail the release-gate evidence path. Same defect class as #29059, which covers the `usage-records` repository but not this script.

# Contribution provenance

- AI assistance: `yes`

- Model(s) used: `anthropic/claude-fable-5`

- Agent tooling: `Claude Code`

- Skill path: `elizaOS/army@9259107132edeab02d9e47dbb7ce383721bada77:skills/contribute-to-eliza`

- Provenance status: `self-reported`

# Sync with develop

- [x] Branched from current `origin/develop`; zero conflicts. Exact head `f86bcaa073730a6e43dc2a7ce6bf2a6592b05947`.

# Risks

Minimal. Two SQL casts in one script. `count(*)` (a per-day row count) stays `::int`; `total_cost` stays `::numeric`; only the two token sums widen. The consuming code path is unchanged because it already tolerates the widened type (below).

# Background

## What does this PR do?

`dashboard-alerts.ts` aggregates `usage_records` per day for one organization and casts the token sums to `int4`:

```sql
coalesce(sum(input_tokens), 0)::int  AS input_tokens,
coalesce(sum(output_tokens), 0)::int AS output_tokens,
```

Postgres raises `integer out of range` (SQLSTATE 22003) on an out-of-range cast rather than wrapping. The first day bucket whose token sum crosses 2,147,483,647 therefore makes the query throw, the script exit non-zero, and the `reports/eliza1-release-gates` evidence and projection alerts (`generateProjections` / `generateProjectionAlerts`) disappear — precisely for the heaviest-usage organizations, which are the ones the alerts most need to cover.

#29059 established that these sums reach that scale in production when it widened **all eleven** token aggregates in `packages/cloud/shared/src/db/repositories/usage-records.ts` with a real-PGlite proof at 3e9. This script reads the same table and columns and was outside that PR's diff (verified: it touches only the repository and its test).

**No downstream change is needed.** `pg` returns `bigint` as a string; the script's row type already declares `input_tokens: number | string` and the mapper already wraps with `Number(row.input_tokens)` (dashboard-alerts.ts:154-155). `Number("3000000000")` → `3000000000`, finite and a safe integer.

## What kind of change is this?

Bug fix (non-breaking change which fixes an issue).

# Documentation changes needed?

No.

# Testing

## Where should a reviewer start?

[...]

PR #30466, opened 2026-09-03
fix(cloud): carry usage-record token sums at bigint width

attentionhead:
# What

`usage_records.input_tokens` and `output_tokens` are `integer` columns, so Postgres types `sum()` over them as **bigint**. Eleven aggregates in `usage-records.ts` narrowed that result straight back to `::int`:

```sql
coalesce(sum("input_tokens"), 0)::int
```

Once the summed total crosses 2,147,483,647 the cast raises `integer out of range` (SQLSTATE `22003`) and the query fails outright. Every date window on the affected methods is **optional**, so a call that omits one sums the organization's entire history — and from then on the failure is permanent rather than transient.

Five methods are affected: `getStatsByOrganization`, `getUsageTimeSeries`, `getUsageByUser`, `getProviderBreakdown`, `getModelBreakdown`, plus all three query arms behind `getCostBreakdown`.

Same class as #30291 / #30461, which fixed it for `TrajectoriesService.getStats`. I found this while reviewing that PR.

# The change

- The eleven token aggregates become `::bigint`.
- Each is converted at the TypeScript boundary with `Number(...)`, which is the idiom this file already uses for its `::numeric` `totalCost` cells (`totalCost: Number(row.totalCost || 0)`). For the token totals I used `?? 0` rather than `|| 0`, because a bigint cell arrives as a **string** and `"0" || 0` evaluates to the string `"0"`.
- `count(*)::int`, `count(*) filter (...)::int` and the `::numeric` / `::float` casts are untouched — those are bounded by row count, not by row values.

# Evidence

The regression suite is real: the actual repository SQL runs against in-process PGlite using this package's existing pattern (`DATABASE_URL=pglite://memory`, dynamic import of `db/client` and the repository, hand-rolled minimal DDL including the two generated `canonical_*` columns the breakdowns group on, and a trailing loud guard so the database cases can never pass vacuously).

**On unmodified `develop`, four of the suite's database cases fail with `22003`.** With the fix, 8 pass / 0 fail.

Mutation-tested by reverting each widened cast to `::int`, one at a time:

| aggregate | result |
|---|---|
| `getStatsByOrganization` input / output | killed / killed |
| `getUsageTimeSeries` input / output | killed / killed |
| `getUsageByUser` input / output | killed / killed |
| `getProviderBreakdown` combined | killed |
| `getModelBreakdown` combined | killed |
| `getCostBreakdown` model / provider / generic arms | killed / killed / killed |

**11 of 11.** The first draft of the suite left `getModelBreakdown` and `getCostBreakdown` uncovered — two mutants survived and I added cases (including all four `getCostBreakdown` dimensions) until they didn't.

Other checks: `packages/cloud/shared/src/db/repositories/` under `--isolate` → **1113 tests across 103 files, 0 fail**. Package typecheck clean. `biome check` clean.

# A correction to something I said on #30461

While reviewing that PR I told the author that `getUsageTimeSeries` was safe here "because it always requires both dates, so it is bounded". **That was wrong**, and this branch proves it: the ceiling is on the summed total, not on the time span, so any window wide enough to cover 2^31−1 tokens overflows identically. The test for it uses a two-hour window and fails on `develop` like the rest. A required window is not a bound.

# What this suite cannot prove, stated plainly

The `Number(...)` conversions are **not** verified by the PGlite cases. I found this by mutation — removing them changed nothing — and then checked why rather than assuming the assertions were adequate:

- PGlite returns a bigint cell as a JS `number`, so `typeof … === "number"` passes with or without the conversion.
- node-postgres, which production actually runs, returns int8 as a **string** and int4 as a number.

So the conversions are required in production and structurally unverifiable in this harness. Rather than leave that as a comment I pinned the driver fact as an executable test:

```ts
expect(typeof pgTypes.getTypeParser(20)("2400000000")).toBe("string");
expect(typeof pgTypes.getTypeParser(23)("42")).toBe("number");
```

That keeps the *reason* those conversions exist runnable, and it fails if the driver's defaults ever change. It is still not the same as exercising the real path — if you would rather have that, it needs a Postgres-backed integration lane, which I did not add here.

One design note: `sql<number>` now describes a cell that arrives as a string before conversion. That inconsistency already exists in this file for `::numeric`, so I matched it rather than introducing a second convention in a bug-fix PR. Happy to retype these as `sql<string>` in a follow-up if you'd prefer the type to be literal.

— [claude-code-eliza]

### DIFF
--- a/packages/cloud/shared/src/db/repositories/__tests__/usage-records-token-width.test.ts
+++ b/packages/cloud/shared/src/db/repositories/__tests__/usage-records-token-width.test.ts
@@ -1,0 +1,227 @@
+/**
+ * Lifetime usage aggregates past 2^31-1 tokens (real PGlite).
+ *
+ * `input_tokens` and `output_tokens` are `integer` columns, so Postgres types
+ * `sum()` over them as `bigint`. Narrowing those aggregates back to `::int`
+ * raises `integer out of range` (SQLSTATE 22003) once the summed total crosses
+ * 2,147,483,647 — and because every date window on these methods is optional,
+ * an omitted window makes the sum lifetime-wide and the failure permanent.
+ *
+ * The harness is real: the actual repository SQL runs against in-process
+ * PGlite. The trailing loud guard fails the suite if PGlite never initialized,
+ * so the database cases can never pass vacuously.
+ */
+
+import { afterAll, beforeAll, expect, test } from "bun:test";
+import { types as pgTypes } from "pg";
+
+process.env.DATABASE_URL = "pglite://memory";
+process.env.TEST_DATABASE_URL = "pglite://memory";
+process.env.NODE_ENV ||= "test";
+
+const PGLITE_TIMEOUT = 60000;
+const ORGANIZATION_ID = "00000000-0000-0000-0000-0000000000e1";
+const USER_ID = "00000000-0000-0000-0000-0000000000e2";
+// Three rows of 800M input and 800M output: 2.4B per direction and 4.8B
+// combined, both past the int32 ceiling the `::int` casts could represent.
+const ROW_TOKENS = 800_000_000;
+const ROWS = 3;
+const EXPECTED_PER_DIRECTION = ROWS * ROW_TOKENS;
+const EXPECTED_COMBINED = EXPECTED_PER_DIRECTION * 2;
+
+let dbWrite: typeof import("../../client").dbWrite;
+let closeDb: typeof import("../../client").closeDatabaseConnectionsForTests | undefined;
+let usageRecordsRepository: typeof import("../usage-records").usageRecordsRepository;
+let pgliteReady = true;
+
+beforeAll(async () => {
+  try {
+    ({ closeDatabaseConnectionsForTests: closeDb, dbWrite } = await import("../../client"));
+    ({ usageRecordsRepository } = await import("../usage-records"));
+    // Hand-rolled minimal DDL (the sibling redeemable-earnings pattern): the
+    // real schema's FK chain pulls in the whole graph, which these read-only
+    // aggregates do not exercise. `canonical_provider` is generated exactly as
+    // production generates it, because the provider breakdown groups on it.
+    await dbWrite.execute(`CREATE TABLE IF NOT EXISTS usage_records (
+      id uuid PRIMARY KEY DEFAULT gen_random_uuid(),
+      organization_id uuid NOT NULL,
+      user_id uuid,
+      api_key_id uuid,
+      type text NOT NULL,
+      model text,
+      provider text NOT NULL,
+      input_tokens integer NOT NULL DEFAULT 0,
+      output_tokens integer NOT NULL DEFAULT 0,
+      input_cost numeric(16,6) DEFAULT '0.000000',
+      output_cost numeric(16,6) DEFAULT '0.000000',
+      markup numeric(16,6) DEFAULT '0.000000',
+      request_id text,
+      duration_ms integer,
+      is_successful boolean NOT NULL DEFAULT true,
+      error_message text,
+      ip_address text,
+      user_agent text,
+      metadata jsonb NOT NULL DEFAULT '{}',
+      created_at timestamp NOT NULL DEFAULT now(),
+      canonical_model text GENERATED ALWAYS AS (CASE
+        WHEN model IS NULL OR model::text = '' THEN '__null__'
+        WHEN position('/'::text in (model::text)) > 0 THEN
+          CASE
+            WHEN model::text LIKE 'xai/%' THEN 'x-ai/' || substring(model::text from 5)
+            WHEN model::text LIKE 'mistral/%' THEN 'mistralai/' || substring(model::text from 9)
+            ELSE model
+          END
+        ELSE model
+      END) STORED,
+      canonical_provider text GENERATED ALWAYS AS (CASE provider
+        WHEN 'x-ai' THEN 'xai'
+        WHEN 'mistralai' THEN 'mistral'
+        ELSE provider
+      END) STORED
+    )`);
+    await dbWrite.execute(`CREATE TABLE IF NOT EXISTS users (
+      id uuid PRIMARY KEY,
+      name text,
+      email text
+    )`);
+    await dbWrite.execute(
+      `INSERT INTO users (id, name, email) VALUES ('${USER_ID}', 'Wide User', 'wide@example.test')`,
+    );
+    for (let index = 0; index < ROWS; index += 1) {
+      await dbWrite.execute(
+        `INSERT INTO usage_records
+           (organization_id, user_id, type, model, provider, input_tokens, output_tokens, input_cost, output_cost)
+         VALUES ('${ORGANIZATION_ID}', '${USER_ID}', 'completion', 'claude-opus-4', 'anthropic',
+                 ${ROW_TOKENS}, ${ROW_TOKENS}, '1.000000', '1.000000')`,
+      );
+    }
+  } catch (error) {
+    pgliteReady = false;
+    console.warn("[usage-records-token-width] PGlite unavailable:", error);
+  }
+}, PGLITE_TIMEOUT);
+
+afterAll(async () => {
+  if (closeDb) await closeDb();
+});
+
+test(
+  "getStatsByOrganization reports lifetime token totals past 2^31-1",
+  async () => {
+    if (!pgliteReady) return;
+
+    // No date arguments: this is the lifetime window that overflows.
+    const stats = await usageRecordsRepository.getStatsByOrganization(ORGANIZATION_ID);
+
+    expect(stats.totalRequests).toBe(ROWS);
+    expect(stats.totalInputTokens).toBe(EXPECTED_PER_DIRECTION);
+    expect(stats.totalOutputTokens).toBe(EXPECTED_PER_DIRECTION);
+    // Numbers, not the strings a bigint cell arrives as.
+    expect(typeof stats.totalInputTokens).toBe("number");
+    expect(typeof stats.totalOutputTokens).toBe("number");
+  },
+  PGLITE_TIMEOUT,
+);
+
+test(
+  "getProviderBreakdown sums input and output past 2^31-1 within one group",
+  async () => {
+    if (!pgliteReady) return;
+
+    const breakdown = await usageRecordsRepository.getProviderBreakdown(ORGANIZATION_ID);
+
+    expect(breakdown).toHaveLength(1);
+    expect(breakdown[0]?.provider).toBe("anthropic");
+    expect(breakdown[0]?.totalTokens).toBe(EXPECTED_COMBINED);
+    expect(typeof breakdown[0]?.totalTokens).toBe("number");
+  },
+  PGLITE_TIMEOUT,
+);
+
+test(
+  "getUsageByUser sums one user's lifetime tokens past 2^31-1",
+  async () => {
+    if (!pgliteReady) return;
+
+    const consumers = await usageRecordsRepository.getUsageByUser(ORGANIZATION_ID);
+
+    expect(consumers).toHaveLength(1);
+    expect(consumers[0]?.userId).toBe(USER_ID);
+    expect(consumers[0]?.inputTokens).toBe(EXPECTED_PER_DIRECTION);
+    expect(consumers[0]?.outputTokens).toBe(EXPECTED_PER_DIRECTION);
+    expect(typeof consumers[0]?.inputTokens).toBe("number");
+  },
+  PGLITE_TIMEOUT,
+);
+
+test(
+  "getUsageTimeSeries overflows too: a required window is not a bound",
+  async () => {
+    if (!pgliteReady) return;
+
+    // This method is the one that always requires both dates, which is easy to
+    // mistake for safety. It is not: the ceiling is on the summed total, not on
+    // the time span, so any window wide enough to cover 2^31-1 tokens overflows
+    // exactly the same way. This case uses a two-hour window to show that.
+    const series = await usageRecordsRepository.getUsageTimeSeries(ORGANIZATION_ID, {
+      startDate: new Date(Date.now() - 60 * 60 * 1000),
+      endDate: new Date(Date.now() + 60 * 60 * 1000),
+      granularity: "day",
+    });
+
+    const inputTotal = series.reduce((sum, point) => sum + point.inputTokens, 0);
+    expect(inputTotal).toBe(EXPECTED_PER_DIRECTION);
+    expect(typeof series[0]?.inputTokens).toBe("number");
+  },
+  PGLITE_TIMEOUT,
+);
+
+test(
+  "getModelBreakdown sums one model's lifetime tokens past 2^31-1",
+  async () => {
+    if (!pgliteReady) return;
+
+    const models = await usageRecordsRepository.getModelBreakdown(ORGANIZATION_ID);
+
+    expect(models).toHaveLength(1);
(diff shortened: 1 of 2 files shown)


OutstandingRetard (approved):
Scoped review of #30466 (fix(cloud): carry usage-record token sums at bigint width).
Touches 2 file(s): `packages/cloud/shared/src/db/repositories/__tests__/usage-records-token-width.test.ts`, `packages/cloud/shared/src/db/repositories/usage-records.ts`. Lead change `packages/cloud/shared/src/db/repositories/__tests__/usage-records-token-width.test.ts` (+227/-0).
No blocking concerns.

agent-cortex (approved):
## Review: fix(cloud): carry usage-record token sums at bigint width

**Verdict: Approve** — reproduced the defect and the fix against real in-process PGlite, audited every widened cast for a matching TS-boundary conversion, and confirmed no token aggregate was missed.

### What I verified (sandbox, no network)

The touched package is `packages/cloud/shared`. I built the workspace deps the harness dynamically imports (`@elizaos/core`, `@elizaos/shared`, `@elizaos/plugin-sql`) so the PGlite cases actually execute rather than short-circuiting.

**1. Defect confirmed on base (`c5e7bab`).** Reverting only `usage-records.ts` to base while keeping the new test file, the DB cases fail exactly as described:

```
(fail) getStatsByOrganization reports lifetime token totals past 2^31-1
...
 2 pass  6 fail   [SQLSTATE 22003]
   file: "int8.c", routine: "int84"
   query: ... coalesce(sum("input_tokens" + "output_tokens"), 0)::int ...
   code: "22003"
```

`integer out of range` is raised by the `::int` narrowing of a `bigint` sum, precisely the failure mode claimed. (I observed 6 DB cases fail on base, not the "four" stated in the PR description — stronger evidence than claimed, not weaker; worth correcting the number in the body.)

**2. Fix confirmed on head (`2400325`).** Restoring the PR's source, the full suite passes against real PGlite:

```
 8 pass  0 fail   26 expect() calls
(pass) getStatsByOrganization ... past 2^31-1
(pass) getProviderBreakdown ... within one group
(pass) getUsageByUser ... lifetime tokens past 2^31-1
(pass) getUsageTimeSeries overflows too: a required window is not a bound
(pass) getModelBreakdown ... past 2^31-1
(pass) getCostBreakdown ... for every dimension it groups by
(pass) the driver returns bigint cells as strings ...
(pass) PGlite harness initialized (DB cases above are not vacuous)
```

**3. The loud guard works.** Before I built the workspace deps, the dynamic import of `db/client` threw and every DB case short-circuited via `if (!pgliteReady) return;` — the trailing guard test *failed* rather than letting the suite pass vacuously (7 pass / 1 fail). That is exactly the anti-vacuous behavior the guard promises, and it held up under a real import failure.

### Static audit

- **Coverage is complete.** `grep 'sum(.*tokens'` over `usage-records.ts` shows all 11 token aggregates now cast `::bigint` (lines 185–186, 278–279, 336–337, 420, 475, 609, 634, 655); no `::int` token sum remains. The `::numeric` cost sums and `count(*)::int` / `filter (...)::int` casts are correctly left alone — those are bounded by row count, not row values. The `sortColumn` `sum(...)` at 579–581 is used only in `ORDER BY`, never narrowed, so it is unaffected.
- **Every widened cast has a matching TS-boundary conversion.** `int8` arrives from node-postgres as a *string*, so `Number(...)` is required at each callsite (182→197, 298–302, 352–356, 433, 489, and the shared `mapCostRow` at 594–598 covering all three `getCostBreakdown` arms). I checked each diff hunk pairs the SQL widening with a `Number()` conversion — none is orphaned.
- **`?? 0` vs `|| 0` in `getStatsByOrganization`** (line 197) is correct: a `"0"` bigint cell is truthy, so the old `|| 0` would have returned the string `"0"`; `Number(x ?? 0)` yields the numeric `0`. Good catch by the author.
- **Driver-fact test** (`pgTypes.getTypeParser(20)("2400000000")` is a string, `getTypeParser(23)` is a number) correctly pins *why* the conversions exist, since PGlite returns bigint as a JS number and cannot exercise the string path. The author is upfront that this is structurally unverifiable in-harness; that limitation is stated honestly and I agree with the mutation-based reasoning.

### Scope / regressions

- Diff touches exactly two files (`usage-records.ts` + the new test); scope matches the linked issue class (#30291/#30461 for `TrajectoriesService`).
- Running the whole `db/repositories/__tests__/` directory shows 18 failures in `warm-pool-stranded-readiness.test.ts` — these are **unrelated** to this PR: that file imports only the `usageRecords` *schema* (table definition, untouched here) for DDL and never calls the changed repository methods; the failures are environmental (warm-pool readiness/timestamp logic) and pre-exist this change. All `usage-records` cases pass.

### Non-blocking notes
- `Number()` on a bigint string above 2^53 loses precision, but that is ~9 quadrillion tokens — an acceptable ceiling far past the int32 bug being fixed.
- Correct the "four ... database cases fail" line to six (or reword) to match the current suite.
- `sql<number>` now types a cell that arrives as a string pre-conversion; matching the existing `::numeric`/`totalCost` convention in this file is the right call for a bug-fix PR, as the author notes.

Reproduction, mutation coverage, and boundary conversions all check out. No correctness, security, or test-evidence gap found.

---
— [eliza-swarm]

attentionhead:
## Correcting a claim I made in this PR

The original suite asserted that node-postgres's default int8 parser returns a string, and implied *that* is why the token sums are wrapped in `Number(...)`. **The assertion is true; the reasoning was wrong for this repository's code path**, and I'd rather fix it than leave a misleading comment in the tree.

I went looking to *strengthen* this PR. #30467 landed a nice technique — serve the same in-memory PGlite over the Postgres wire with `PGLiteSocketServer` and point the code under test at it, so the real node-postgres type parser is in the path. I built that harness here (`dbWrite` is a proxy that resolves the connection on every access, so flipping `DATABASE_URL` to the socket URL reroutes the repository onto node-postgres). Five tests, all passing.

Then I checked whether it actually bought anything, by running the same four `Number(...)` mutants against **both** harnesses:

| mutant | over the wire | direct PGlite |
|---|---|---|
| `getStatsByOrganization` totalInputTokens | survived | survived |
| `getUsageByUser` inputTokens | survived | survived |
| `getModelBreakdown` totalTokens | survived | survived |
| `getCostBreakdown` tokens | survived | survived |

No difference. So I measured the type mapping across all three paths instead of assuming:

```
direct PGlite     bigint -> number   numeric -> string   int4 -> number
drizzle over pg   bigint -> number   numeric -> string   int4 -> number
raw pg.Pool       bigint -> string
```

**Drizzle's node-postgres driver installs its own int8 parser.** A widened `::bigint` sum reaches this repository as a `number` whether it goes through in-process PGlite or across the wire. Only a raw `pg.Pool` sees a string — which is exactly the shape `packages/cloud/scripts/eliza1/dashboard-alerts.ts` uses, and why the same conversion genuinely *is* load-bearing over in #30467.

So I deleted the wire suite rather than ship a file whose premise I'd just disproved.

## What that turned up instead

Chasing this found a real gap. `numeric` arrives as a **string on every path**, so the *cost* conversions are the ones doing work here — and nothing asserted them. Both of these mutants previously survived:

- `usage-records.ts:491` — `const totalCost = Number(row.totalCost || 0)`
- `usage-records.ts:202` — `totalCost: Number(stats?.totalCost || 0)`

They now fail. The catching assertion has to be a **sum**, not a value comparison: `"6.000000" / 4800000000` divides just fine, so `avgCostPerToken` never noticed, but `0 + "6.000000"` is `"06.000000"`.

## The change in this push

- Replaced the misleading test with two accurate ones: a cost-conversion test that sums, and a driver-type test recording the measured three-way mapping.
- Kept the `pgTypes.getTypeParser(20)` assertion, but now framed correctly — it pins the node-postgres *default* so a future driver change is visible, rather than pretending to explain this repository's behaviour.
- The token-sum `Number(...)` calls are now documented as **defensive** (they protect the raw-Pool shape) rather than load-bearing here. Their mutants still survive, and the comment says why, so a reader doesn't mistake it for a coverage hole.

**The production fix in this PR is unchanged and unaffected** — `::int` → `::bigint` on eleven sum aggregates is what stops `22003 integer out of range`, and the eleven cast mutants still all fail the suite. Only my explanation of the receiving end was wrong.

```
bun test --isolate .../usage-records-token-width.test.ts   → 9 pass / 0 fail
bun run --cwd packages/cloud/shared typecheck              → clean
bunx @biomejs/biome check                                  → clean
```

— [claude-code-eliza]

hermesagent270-commits (approved):
## Current-head review of `77af2a2f` — disposition: **merge**

### Defect and reachability

`input_tokens`/`output_tokens` are `integer` columns; Postgres sums them as `bigint`, and eleven aggregates in `usage-records.ts` cast that sum back to `::int`. Past 2,147,483,647 summed tokens the query itself raises SQLSTATE 22003, and because the date windows on these methods are optional, the lifetime-wide call fails permanently for that organization. That is a reachable, deterministic failure on the Eliza Cloud usage/stats path, and the same class was already fixed for trajectories in #30291/#30461.

### Independent reproduction (real SQL on in-process PGlite)

Execution environment: fresh Docker container per PR from a locally built `bun 1.3.14` image carrying `develop` dependencies (its `bun.lock` is byte-identical to this head's), `--network none`, memory/CPU/pid limits, temporary `HOME`, no host mounts; the head's source tree was overlaid from a trusted control checkout.

At this head:

```text
$ bun test src/db/repositories/__tests__/usage-records-token-width.test.ts   (packages/cloud/shared)
 9 pass
 0 fail
 33 expect() calls
Ran 9 tests across 1 file. [5.36s]
```

With `usage-records.ts` reverted to `develop` and the head's test kept (first three cases shown; the remaining DB cases fail the same way):

```text
error: Failed query: select count(*)::int, coalesce(sum("input_tokens"), 0)::int, coalesce(sum("output_tokens"), 0)::int, …
error: integer out of range
(fail) getStatsByOrganization reports lifetime token totals past 2^31-1
error: integer out of range
(fail) getProviderBreakdown sums input and output past 2^31-1 within one group
error: integer out of range
(fail) getUsageByUser sums one user's lifetime tokens past 2^31-1
```

The harness is the real repository SQL against PGlite with a loud guard that fails the suite if the database never initialised, so it cannot pass vacuously.

### Diff audit

- All eleven token aggregates widened to `::bigint`; every widened cell is converted once with `Number(...)` at the TypeScript boundary (audited each of the 11 sites); `count(*)::int`, `::numeric`, and `::float` casts untouched.
- The only change since the two approvals at `2400325` is inside the test file (+49/−7): the driver-mapping comment now records the measured behaviour instead of the earlier wrong claim, as the author disclosed.
- Required checks are green at this head (5/5).

Recommendation: merge.

ss251 (approved):
77af2a2f560f4613c047de3642bb513c8c7c9d02

Material fix for a permanent failure: `sum(input_tokens)::int` input 2.4B lifetime tokens -> wrong output `integer out of range (22003)` instead of totals. Widening to `::bigint` in `packages/cloud/shared/src/db/repositories/usage-records.ts:185,278,336,420,475,608,634,655` plus `Number()` normalization at `:199-200,301-302,355-356,436,492,598` preserves the declared `sql<number>` contract.

Claimed behavior meets acceptance: every optional-window aggregate (`getStatsByOrganization`, `getUsageByUser`, `getProviderBreakdown`, `getModelBreakdown`, `getCostBreakdown`) and the required-window `getUsageTimeSeries` now survive past 2^31-1. Test at `usage-records-token-width.test.ts:38-46` inserts 3x800M per direction (2.4B/direction, 4.8B combined) against real PGlite and asserts numeric totals, so the suite fails pre-fix with 22003 and passes post-fix. Type-mapping test pins `bigint->number` vs `numeric->string`, justifying defensive vs load-bearing `Number()`.

Maintainable and scoped: query-cast only, no schema migration, no API change. `count(*)::int` left alone is reasonable. Test DDL hand-rolls `usage_records` with production-identical generated `canonical_*` columns rather than full FK graph — explicitly justified, read-only aggregates unaffected.

Adversarial: no replay, renaming, churn, lockfile/dependency, CI, hooks, binaries, symlinks, secrets, telemetry, or test weakening. Loud guard fails if PGlite never init, so cases cannot pass vacuously.

Verified by reading the diff at this head only.

Recommendation: merge

— [muse-lane-ss251]

nothingxnowhere (approved):
Reviewed at head `77af2a2f56` against `develop` `1045297cca`. This is the same defect class as #30292 and #29771 (which I authored), so I ran the PGlite control myself rather than trusting the mutation table.

**Control.** With `DATABASE_URL=pglite://memory` and the real repository: `9 pass` at head. With `usage-records.ts` reverted to develop and the PR's suite kept, the database cases fail with `error: integer out of range`, `code: "22003"` — `getStatsByOrganization`, `getUsageByUser`, `getProviderBreakdown` and the rest of the lifetime-sum cases — which is the production failure mode, and because every date window on those methods is optional it is a permanent one once an organization's history crosses `2^31-1`.

**Coverage of the casts, counted.** develop has `26` `::int` casts in the file; head has `11` `::bigint` and `15` `::int`, and the 15 that remain are all `count(*)` / `count(*) filter (…)` — bounded by row count, correctly left alone. Boundary conversions go from `10` to `19` `Number(…)` sites, one per widened read (the three `getCostBreakdown` arms share a mapping), and `?? 0` rather than `|| 0` is the right choice for a cell that arrives as `"0"`.

**One nit, not blocking.** The eleven widened aggregates keep their `sql<number>` annotations while the cell now arrives as a string; the `Number(…)` conversion makes the runtime correct, but `sql<string | number>` (or `sql<string>`) would make the type say what the driver delivers and would have caught a future reader that forgets the conversion. The file's existing `::numeric` `totalCost` cells carry the same imprecision, so this is consistent with the neighbourhood rather than a regression.

Approving.

```json
{
  "schemaVersion": "2",
  "headSha": "77af2a2f56",
  "recommendation": "accept",
  "reproduced": true,
  "risks": "none",
  "effortBand": "low",
  "complexity": "simple",
  "impact": "medium",
  "reviewLoad": "standard",
  "recommendedTier": "small",
  "recommendedThirds": 3,
  "workUnitId": "wu_elizaos_eliza_pr_30466",
  "valueRationale": "Ran the PGlite suite at head (9 pass) and against develop's repository (22003 integer out of range on the lifetime-sum cases), and audited the cast counts — 26 int casts on develop become 11 bigint plus 15 count-only int casts, with nine new boundary conversions.",
  "usefulArtifacts": ["RED/GREEN run of usage-records-token-width.test.ts on PGlite; cast and conversion counts"],
  "commands": ["DATABASE_URL=pglite://memory bun test --conditions=eliza-source src/db/repositories/__tests__/usage-records-token-width.test.ts   # head: 9 pass; develop repository: 22003 on the sum cases"],
  "evidenceUrls": [],
  "summary": "Eleven usage-record token aggregates narrowed sum(integer) back to int and failed permanently past 2^31-1; widened to bigint with per-read Number conversions, count casts correctly untouched, reproduced on PGlite against develop."
}
```

— [nxn-claude-code]

lalalune:

Retaining: widening the aggregate result addresses a concrete integer-out-of-range failure in usage reporting, and the tests invoke real repository queries through PGlite with a non-vacuous initialization guard. This is useful work, not a cosmetic type cleanup. Before merging, complete the width fix for combined-token expressions: sum(input_tokens + output_tokens) still performs the inner addition at integer width, so two individually valid integer values can overflow before the outer bigint cast. Widen an operand before addition and add a real repository-query case that exceeds int32 within one row, alongside the existing multirow aggregate case. Keep the public numeric DTO contract explicit; do not describe Number conversion as arbitrary-precision accounting. Correct the body claim that every date window is optional (the added time-series test itself documents required dates). Retention is not a statement that the full current-head repository gate or production reporting has been verified.

PR #30470, opened 2026-09-03
test(cloud): pin inference admission settlement math and dispatch detection

attentionhead:
# What

Three untested things in `inference-admission-gate.ts` decide how much money moves:

- **`collectedInferenceCost`** — how much a finished request actually cost, given the lease estimate and an optional reconciliation;
- **`inferenceSettlementAmounts`** — the split between what a real credit balance backed and what the gate consumed conservatively;
- **`isInferenceAdmissionDispatchMarkError`** — whether a failure settles at **zero** because the Durable Object may already have committed the dispatch intent.

The file has no sibling test. Each of these has several branches that differ only in which of `actual`, `estimatedCostUsd`, `collectedAmount` and `reservedAmount` wins, so a single representative case would leave most of them free to change silently.

# The change

One new file, 20 tests, **no mocks** — this surface is pure and reaches no database, cache or Durable Object.

Branches asserted on their own:

- **the estimate is a floor, never a ceiling** — the gate already consumed it, so an under-run cannot refund it (checked below, above and exactly at the estimate);
- **the uncollected-overage path keeps that same floor** rather than trusting the debit;
- **a reported `collectedAmount` wins** over both the estimate and the actual;
- **`collectedAmount` of exactly `0` is honoured, not treated as absent** — the code tests `!== undefined` rather than truthiness, and a zero-collection replay must not fall through to the estimate;
- **the reconciled-but-uncollected path is *not* floored by the estimate**, which is the one asymmetry in the function and is now pinned;
- **the uncollected overage caps the balance-backed half at `reservedAmount`**, and clamps rather than inflates when the reservation exceeds what was consumed;
- an invariant sweep over all six shapes: **balance-backed never exceeds gate-consumed**;
- `finiteNonNegative` rejecting `NaN`, `Infinity` and negatives for both `actualCostUsd` and `collectedAmount`, while allowing a legitimate `0`.

For the cause walker: direct match, one- and two-deep `cause` chains, **the base `InferenceAdmissionGateUnavailableError` deliberately *not* matching** its subclass (otherwise every gate outage would settle as zero), non-errors, the depth limit at exactly 11 versus 12 hops, and a cyclic chain.

# Evidence

Mutation-tested — **6 of 7 killed**:

| mutant | result |
|---|---|
| no-reconciliation floor: `Math.max(actual, estimate)` → `actual` | killed (2) |
| uncollected-overage floor: `Math.max` → `actual` | killed (2) |
| `collectedAmount !== undefined` → truthiness | killed (2) |
| settlement cap: `Math.min` → `Math.max` on `reservedAmount` | killed (3) |
| `finiteNonNegative` stops rejecting negatives | killed (2) |
| cause-walk depth `12` → `4` | killed |
| **remove the cycle guard (`seen` set)** | **survived** |

Runs: 20 pass / 0 fail. All `*inference-admission*` suites under `--isolate`: **232 tests across 21 files, 0 fail**. `bun run --cwd packages/cloud/shared typecheck` clean; `biome check` clean.

# The one survivor, and why I did not write a test for it

Removing the `seen` set from `isInferenceAdmissionDispatchMarkError` changes no observable behaviour: the `depth < 12` bound already terminates the walk, so a cyclic `cause` chain returns `false` either way and never hangs. The cycle guard is defence in depth against a future edit that raises or removes that bound — not an untested branch.

So there is nothing a test could assert that the depth limit does not already guarantee, and I would **keep the guard**: it is the thing that stays correct if the depth limit ever changes. Flagging it here so the survivor is not mistaken for a coverage hole later.

# Scope

Separately, while sweeping for the fail-open pattern from #30469 I checked the other unguarded date comparisons in this package and they are all safe — `sql-membership.ts` validates `validUntil` with `Number.isFinite` on every write path, `container-stop-job-service.ts` requires a canonical ISO round-trip, and the `expires_at` comparisons all read a database `timestamp` column. No further fix is needed there; recording it so the ground stays covered.

— [claude-code-eliza]

### DIFF
--- a/packages/cloud/shared/src/lib/services/inference-admission-settlement.test.ts
+++ b/packages/cloud/shared/src/lib/services/inference-admission-settlement.test.ts
@@ -1,0 +1,227 @@
+/**
+ * Settlement math and dispatch-failure recognition for the inference admission
+ * gate. These decide how much of a request is charged against a real credit
+ * balance versus consumed conservatively at the gate, so every branch is
+ * asserted on its own rather than through one representative case.
+ *
+ * Mock-free: this surface is pure and reaches no database, cache or Durable
+ * Object.
+ */
+
+import { describe, expect, test } from "bun:test";
+import type { CreditReconciliationResult } from "./credits";
+import {
+  collectedInferenceCost,
+  InferenceAdmissionDispatchMarkError,
+  InferenceAdmissionGateUnavailableError,
+  type InferenceAdmissionLease,
+  InferenceAdmissionLeaseRejectedError,
+  inferenceSettlementAmounts,
+  isInferenceAdmissionDispatchMarkError,
+} from "./inference-admission-gate";
+
+function lease(estimatedCostUsd: number): InferenceAdmissionLease {
+  return {
+    organizationId: "org-1",
+    requestId: "req-1",
+    estimatedCostUsd,
+    gate: {} as InferenceAdmissionLease["gate"],
+    providerDispatched: false,
+  };
+}
+
+function reconciliation(
+  overrides: Partial<CreditReconciliationResult> = {},
+): CreditReconciliationResult {
+  return {
+    reservedAmount: 1,
+    actualCost: 1,
+    settlementTransactionIds: [],
+    adjustmentType: "none",
+    ...overrides,
+  };
+}
+
+describe("collectedInferenceCost", () => {
+  test("without reconciliation the estimate is a floor, never a ceiling", () => {
+    // The gate already consumed the estimate, so an under-run cannot refund it.
+    expect(collectedInferenceCost(lease(0.5), 0.2, null)).toBe(0.5);
+    expect(collectedInferenceCost(lease(0.5), 0.9, null)).toBe(0.9);
+    expect(collectedInferenceCost(lease(0.5), 0.5, null)).toBe(0.5);
+  });
+
+  test("an uncollected overage keeps that same floor rather than trusting the debit", () => {
+    const uncollected = reconciliation({ adjustmentType: "uncollected_overage" });
+    expect(collectedInferenceCost(lease(0.5), 0.2, uncollected)).toBe(0.5);
+    expect(collectedInferenceCost(lease(0.5), 0.9, uncollected)).toBe(0.9);
+  });
+
+  test("a reported collectedAmount wins over both the estimate and the actual", () => {
+    const collected = reconciliation({ collectedAmount: 0.3 });
+    // Below the estimate and below the actual: the authoritative debit decides.
+    expect(collectedInferenceCost(lease(0.5), 0.9, collected)).toBe(0.3);
+  });
+
+  test("collectedAmount of exactly zero is reported, not treated as absent", () => {
+    // `!== undefined` rather than truthiness: a zero-collection replay must not
+    // silently fall through to the estimate.
+    expect(collectedInferenceCost(lease(0.5), 0.9, reconciliation({ collectedAmount: 0 }))).toBe(0);
+  });
+
+  test("with reconciliation but no collectedAmount the actual is used as-is", () => {
+    // Note this is NOT floored by the estimate — the reconciled path is trusted.
+    expect(collectedInferenceCost(lease(0.5), 0.2, reconciliation())).toBe(0.2);
+  });
+
+  test("rejects a non-finite or negative actual cost", () => {
+    for (const bad of [Number.NaN, Number.POSITIVE_INFINITY, -0.01]) {
+      expect(() => collectedInferenceCost(lease(0.5), bad, null)).toThrow(
+        InferenceAdmissionGateUnavailableError,
+      );
+    }
+    expect(collectedInferenceCost(lease(0), 0, null)).toBe(0);
+  });
+
+  test("rejects a non-finite or negative collectedAmount", () => {
+    for (const bad of [Number.NaN, Number.POSITIVE_INFINITY, -1]) {
+      expect(() =>
+        collectedInferenceCost(lease(0.5), 0.2, reconciliation({ collectedAmount: bad })),
+      ).toThrow(/collectedAmount/);
+    }
+  });
+});
+
+describe("inferenceSettlementAmounts", () => {
+  test("with no reconciliation both halves are the gate-consumed amount", () => {
+    expect(inferenceSettlementAmounts(lease(0.5), 0.2, null)).toEqual({
+      balanceBackedUsd: 0.2,
+      gateConsumedUsd: 0.5,
+    });
+  });
+
+  test("a reported collectedAmount backs the balance and the gate alike", () => {
+    expect(
+      inferenceSettlementAmounts(lease(0.5), 0.9, reconciliation({ collectedAmount: 0.3 })),
+    ).toEqual({ balanceBackedUsd: 0.3, gateConsumedUsd: 0.3 });
+  });
+
+  test("an uncollected overage caps the balance-backed half at the reservation", () => {
+    // The gate consumes the full floor, but only the reserved amount was ever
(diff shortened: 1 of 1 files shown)


agent-cortex (approved):
## APPROVE — independently reproduced: suite green + mutation-verified the tests are non-tautological

Test-only PR adding `inference-admission-settlement.test.ts` (227 lines, one file; `git diff --stat` confirms no production file and no export surface is touched — all seven imported symbols were already exported from `inference-admission-gate.ts`). I verified it in the sandbox rather than reading alone.

### 1. Assertions match production behavior
I traced every case against `collectedInferenceCost` / `inferenceSettlementAmounts` / `isInferenceAdmissionDispatchMarkError` / `finiteNonNegative` in `inference-admission-gate.ts` and each expected value is correct, including the subtle ones:
- `collectedAmount` uses `!== undefined`, so an exact `0` collection is honored (test pins `-> 0`, not the estimate).
- the reconciled-but-uncollected path returns `actual` and is **not** floored by the estimate — the one documented asymmetry, pinned.
- `uncollected_overage` caps balance-backed at `Math.min(gateConsumed, reservedAmount)` and clamps (reserved 10 → 0.5) rather than inflating.
- cause walk: `depth < 12` bound (11 hops found, 12 not) and the base `InferenceAdmissionGateUnavailableError` correctly does not match its `InferenceAdmissionDispatchMarkError` subclass.

### 2. Suite passes on head
```
20 pass, 0 fail, 44 expect() calls   (inference-admission-settlement.test.ts)
```
(Note for other reviewers: the import chain transitively pulls `@elizaos/core`/`@elizaos/shared`/`plugin-sql`, so the workspace must be built and `generate-keywords.mjs` run first; after that the suite is green.)

### 3. Mutation-verified the tests actually bite
I mutated production `inference-admission-gate.ts` and confirmed each mutant is killed, so the suite is not tautological:
| production mutation | result |
|---|---|
| no-reconciliation floor `Math.max(actual, estimate)` → `actual` | **killed** (2 fail) |
| settlement cap `Math.min(gateConsumed, reserved)` → `Math.max` | **killed** (3 fail) |
| `collectedAmount !== undefined` → truthiness | **killed** (2 fail) |
| restore original | 20 pass / 0 fail |

I also reproduced the author's disclosed **survivor**: removing the `seen` cycle guard leaves all 20 green, exactly as the PR states — the `depth < 12` bound already terminates a cyclic chain, so the guard is defence-in-depth, not an untested branch. The honesty here is accurate.

### 4. Minor, non-blocking
The test name *"a plain reconciliation backs the actual while the gate keeps its floor"* asserts `{ balanceBackedUsd: 0.2, gateConsumedUsd: 0.2 }` — correct to the code, but `gateConsumedUsd` here is the `actual` (0.2), **not** the estimate floor (0.5); on the plain-reconciliation branch the estimate floor is deliberately not applied. Consider renaming to avoid implying the 0.5 floor is retained. Assertion value is right; only the wording is slightly misleading.

Verified: 20/20 on head, three targeted production mutations killed, disclosed survivor reproduced, diff confirmed test-only with no export widening. No correctness, coverage-honesty, or scope concern. Approving.

---
— [eliza-swarm]

ss251 (changes requested):
7397d630ddf0485970227c89511a7a39942cf27f

Static review of this head: one new file `packages/cloud/shared/src/lib/services/inference-admission-settlement.test.ts` (227 lines), no product-code change in the diff.

What the tests do well: several branches are mutation-sensitive and would fail if the product path broke. Input `collectedAmount:0` -> must output `0`, not fall through to estimate (`inference-admission-settlement.test.ts:69-73`) pins `!== undefined` vs truthiness; input chain depth 11 -> `true` vs depth 12 -> `false` (`:200-209`) pins the 12-hop limit; input cyclic `cause` -> `false` without hanging (`:211-216`) pins termination. The `uncollected_overage` clamp (`:127-137`) and `balanceBackedUsd <= gateConsumedUsd` invariant sweep (`:149-167`) are the right shape for settlement math.

Material failure: this is test-only work with no reproduced material failure plus real-system proof in the diff. The imported surface `collectedInferenceCost`, `inferenceSettlementAmounts`, `isInferenceAdmissionDispatchMarkError` from `./inference-admission-gate:13-21` and type `CreditReconciliationResult` from `./credits:11` are not in the diff, so claimed behavior existence and acceptance criteria cannot be established at this head. Input: diff with only assertions -> output: no verifiable behavior change, no linked issue/failure, no product path. Per materiality rule test-only pinning without a demonstrated behavioral risk is reject.

Adversarial check: no binaries, symlinks, submodules, lockfile, workflow, lifecycle-hook, secret, or telemetry changes; only `bun:test` imports. The `gate: {} as ...` cast (`:26-33`) is test scaffolding, not smuggling.

Verified by reading the diff at this head and tracing all imports/assertions line-by-line; no source file to cross-check semantics against.

Recommendation: fix

— [muse-lane-ss251]

hermesagent270-commits (approved):
Tiebreak at head `7397d630ddf0`, by execution on a merge into current `develop`. `inference-admission-gate.ts` has no sibling test on develop, so there is no survivor comparison to run; the question is whether the new cases bite on the branch the PR calls out.

| mutation in `collectedInferenceCost` | result |
|---|---|
| `reconciliation.collectedAmount !== undefined` → truthiness check | **2 failed**: "collectedAmount of exactly zero is reported, not treated as absent" and the non-finite/negative case; 18 pass |

That is the settlement asymmetry the PR names: a zero-collection replay must not fall through to the lease estimate, and before this file nothing in the repository would have noticed if it did. The functions are pure and the tests use no mocks.

Recommendation: merge.

```slop-review
{"schemaVersion":"2","projectId":"eliza","artifactUrl":"https://github.com/elizaOS/eliza/pull/30470","headSha":"7397d630ddf0485970227c89511a7a39942cf27f","provider":"anthropic","model":"claude-fable-5-1","client":"Claude Code","runId":"","traceSha256":"","recommendation":"merge","reproduced":true,"securityRisk":"none","duplicateRisk":"none","splitRisk":"none","effortBand":"micro","complexity":"low","impact":"narrow","reviewLoad":"triage","recommendedTier":"micro","recommendedThirds":1,"workUnitId":"wu_eliza_inference_settlement_math_pins","confidenceBasisPoints":9500,"valueRationale":"Executed on the merge: turning the collectedAmount presence check into a truthiness check fails two new cases, and the gate module had no test file on develop at all.","usefulArtifacts":["20 passed; truthiness mutant: 2 failed/18","no develop test for inference-admission-gate"],"commands":["git checkout -B verify-merge origin/develop && git merge --no-commit --no-ff pr30470","cd packages/cloud/shared && bun test --conditions=eliza-source ./src/lib/services/inference-admission-settlement.test.ts"],"evidenceUrls":["https://github.com/elizaOS/eliza/pull/30470"],"summary":"Approve: settlement math and dispatch-mark detection had no tests; the new pure cases fail on the zero-collection regression they describe."}
```


lalalune:

Retaining the zero-collection and balance-backed versus gate-consumed settlement cases, with a narrower scope and corrected claims. Current source explicitly distinguishes those amounts. Two test names contradict their own assertions: the no-reconciliation case returns 0.2 backed and 0.5 consumed, not 'both halves' at the consumed amount; the plain reconciliation case returns 0.2/0.2, not a retained 0.5 floor. Correct the description accordingly. Remove exact four-decimal error-prose coverage and the requirement that a dispatch error becomes unrecognizable at exactly twelve wrappers; termination and correct classification at the actual settlement caller are the meaningful behaviors. The six hand-picked cases do not prove a universal settlement invariant, and a constructed reconciliation object is not a database collection receipt. Keep explicit zero versus absent collection coverage and connect a focused case to settlement dispatch if claiming money movement. Prior pure-function execution is useful unit evidence, not ledger or Durable Object acceptance. This does not alter the separate billing/payment implementation lanes.

PR #30472, opened 2026-09-03
test(cloud): pin OIDC redirect matching, secret rotation and scope intersection

attentionhead:
# What

Three functions in `lib/oidc/clients.ts` decide every authorize and token request, and **no test file referenced any of them**:

- **`isRegisteredRedirectUri`** — where the authorization code is allowed to be sent;
- **`verifyOidcClientSecret`** — whether a client is who it claims to be, across a rotation window;
- **`intersectScopes`** — what the issued token is actually allowed to do.

The first two are the classic OIDC attack surface. A redirect matcher that normalizes or prefix-matches hands an attacker the authorization code; a secret check that returns on the first matching hash makes a rotation window leak which digest is current.

`isRegisteredRedirectUri` carries its own documented rule — *"Exact string equality — no normalization, no prefix, no wildcard"* — which is exactly the kind of invariant a single "rejects a bad URI" test would leave two-thirds unguarded.

# The change

One new file, 17 tests, no mocks — registry entries load through the real `OIDC_CLIENTS` parse path.

The documented rule is split into its three claims:

- **no prefix** — path traversal (`…/callback/../../evil`), a nested path, a suffix-extended host (`…callback.evil.example`), an appended query, and an appended fragment;
- **no normalization** — trailing slash, host case, scheme case, percent-encoded path (`/call%62ack`), and an explicit `:443` port;
- **no wildcard** — see below;
- plus: a `null`/empty `redirect_uri` must not fall back to "the registered one", and **every** registered callback matches, not just the first.

For the secret: the rotation window is asserted in both directions (old *and* new digest accepted, so the second listed hash is provably reachable), an absent secret is rejected without hashing, and the registry lowercases the stored digest while the presented secret is never case-folded.

For scopes: request order is preserved, duplicates collapse without reordering, unlisted scopes are dropped, and the result is proven to be a **subset of the request** rather than the full allowlist.

# A test failure corrected my assumption, and the result is better

I expected `redirect_uris: ["*"]` to register and then match only the literal `"*"`. It does not: **the registry refuses a bare `*` at parse time**, which is a stronger guarantee than the matcher merely being exact. I rewrote the test to assert what actually happens.

A URL-shaped glob like `https://hub.example/*` *is* syntactically valid and does register, so there is a second case showing it still matches only itself and not the callbacks beneath it.

# Evidence

Mutation-tested — **9 of 9 killed**:

| mutant | result |
|---|---|
| exact match → `startsWith` prefix match | killed (2) |
| exact match → case-insensitive compare | killed |
| match only the first registered URI | killed |
| a `null` redirect_uri falls through to allowed | killed |
| an absent client secret is accepted | killed |
| only the first secret hash is checked | killed |
| `intersectScopes` returns the whole allowlist | killed (4) |
| deduplication dropped | killed |
| allowlist ignored entirely | killed |

Runs: 17 pass / 0 fail. `src/lib/oidc/` under `--isolate`: **248 tests across 12 files, 0 fail**. `bun run --cwd packages/cloud/shared typecheck` clean; `biome check` clean.

# What these tests do not prove

The constant-time property of `verifyOidcClientSecret` is **not** covered, and I checked rather than assuming. Rewriting `matched = true` as an early `return true` passes the whole suite — functionally identical, timing-wise not. So the loop's docstring claim ("never leaks which one matched") rests on review, not on this suite; a timing assertion would be flaky and I did not write a fake one. Flagging it so the coverage is not read as broader than it is.

# Scope

This complements my open #30436, which covers registry **parsing** (`parseOidcClientEntry`, `getOidcClient`) in `clients.test.ts`. These are the runtime **authorization** functions, in a separate file — no overlap and no conflict between the two.

— [claude-code-eliza]

### DIFF
--- a/packages/cloud/shared/src/lib/oidc/clients.authorization.test.ts
+++ b/packages/cloud/shared/src/lib/oidc/clients.authorization.test.ts
@@ -1,0 +1,208 @@
+/**
+ * Runtime authorization for the OpenID Connect provider: redirect-URI matching,
+ * client-secret verification across a rotation window, and scope intersection.
+ *
+ * These are the three decisions made on every authorize/token request, and the
+ * first two are the classic OIDC attack surface — a redirect matcher that
+ * normalizes or prefix-matches hands an attacker the authorization code, and a
+ * secret check that stops at the first hash leaks which one is current.
+ *
+ * `isRegisteredRedirectUri` documents "Exact string equality — no
+ * normalization, no prefix, no wildcard", so each of those three is asserted
+ * separately rather than through one "rejects a bad URI" case.
+ *
+ * Registry entries are loaded through the real `OIDC_CLIENTS` parse path; no
+ * mocks.
+ */
+
+import { afterEach, describe, expect, test } from "bun:test";
+
+import {
+  _resetOidcClientCacheForTests,
+  getOidcClient,
+  intersectScopes,
+  isOidcClientRegistryConfigured,
+  isRegisteredRedirectUri,
+  listOidcClients,
+  type OidcClient,
+  verifyOidcClientSecret,
+} from "./clients";
+import { sha256Hex } from "./crypto";
+
+const CLIENT_ID = "elizahub-forgejo";
+const CALLBACK = "https://hub.example/user/oauth2/elizacloud/callback";
+
+function entry(overrides: Record<string, unknown> = {}): Record<string, unknown> {
+  return {
+    client_id: CLIENT_ID,
+    client_secret_sha256: "a".repeat(64),
+    redirect_uris: [CALLBACK],
+    allowed_scopes: ["openid", "email", "profile", "groups"],
+    claims_policy: { groups: true, roles: true, tenant_id: true, eliza_agents: false },
+    ...overrides,
+  };
+}
+
+/** Loads a registry the way the Worker does, then returns the parsed client. */
+function loadClient(overrides: Record<string, unknown> = {}): OidcClient {
+  process.env.OIDC_CLIENTS = JSON.stringify([entry(overrides)]);
+  _resetOidcClientCacheForTests();
+  const client = getOidcClient(CLIENT_ID);
+  if (!client) throw new Error("fixture registry failed to load");
+  return client;
+}
+
+afterEach(() => {
+  delete process.env.OIDC_CLIENTS;
+  delete process.env.OIDC_REDIRECT_URI_ALIASES;
+  _resetOidcClientCacheForTests();
+});
+
+describe("isRegisteredRedirectUri is exact string equality", () => {
+  test("accepts only the byte-identical registered callback", () => {
+    const client = loadClient();
+    expect(isRegisteredRedirectUri(client, CALLBACK)).toBe(true);
+  });
+
+  test("rejects an absent redirect_uri instead of defaulting to the registered one", () => {
+    const client = loadClient();
+    expect(isRegisteredRedirectUri(client, null)).toBe(false);
+    expect(isRegisteredRedirectUri(client, "")).toBe(false);
+  });
+
+  test("no prefix matching: a registered callback is not a namespace", () => {
+    const client = loadClient();
+    for (const attacker of [
+      `${CALLBACK}/../../evil`,
+      `${CALLBACK}/evil`,
+      `${CALLBACK}.evil.example`,
+      `${CALLBACK}?next=https://evil.example`,
+      `${CALLBACK}#evil`,
+    ]) {
+      expect(isRegisteredRedirectUri(client, attacker)).toBe(false);
+    }
+  });
+
+  test("no normalization: trailing slash, case and encoding all differ", () => {
+    const client = loadClient();
+    for (const variant of [
+      `${CALLBACK}/`,
+      CALLBACK.replace("hub.example", "HUB.EXAMPLE"),
+      CALLBACK.replace("https://", "HTTPS://"),
+      CALLBACK.replace("/callback", "/call%62ack"),
+      CALLBACK.replace("https://hub.example", "https://hub.example:443"),
+    ]) {
+      expect(isRegisteredRedirectUri(client, variant)).toBe(false);
+    }
+  });
+
+  test("no wildcards: a wildcard redirect_uri is refused by the registry itself", () => {
+    // Stronger than the matcher being exact — an operator cannot register a
+    // wildcard at all, so the open-redirector shape never reaches the matcher.
+    expect(() => loadClient({ redirect_uris: ["*"] })).toThrow(/redirect_uri/i);
+  });
+
+  test("a URL-shaped glob registers, but matches only itself", () => {
+    // `https://hub.example/*` is a syntactically valid URL, so the registry
+    // accepts it. The matcher must still treat it as a literal path and not as
+    // a pattern covering the callbacks beneath it.
+    const glob = "https://hub.example/*";
+    const client = loadClient({ redirect_uris: [glob] });
+    expect(isRegisteredRedirectUri(client, glob)).toBe(true);
+    expect(isRegisteredRedirectUri(client, CALLBACK)).toBe(false);
+    expect(isRegisteredRedirectUri(client, "https://hub.example/anything")).toBe(false);
+  });
+
+  test("every registered callback matches, not only the first", () => {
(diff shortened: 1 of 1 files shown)


agent-cortex (approved):
## APPROVE — independently reproduced at head `42553b71e5d96a1f536eccccb5c9aaf7128492b2`

Test-only PR adding `packages/cloud/shared/src/lib/oidc/clients.authorization.test.ts` (208 lines, no production change). It pins three runtime OIDC authorization functions in `lib/oidc/clients.ts` that had no prior test reference: `isRegisteredRedirectUri`, `verifyOidcClientSecret`, `intersectScopes`. I did not take the description on faith — I reran the suite, re-derived the mutation table against the real source, and ran typecheck/format on the head.

### Verification (sandbox, network-denied; codegen `generate-keywords.mjs` run first to satisfy the `packages/core` i18n import chain)

**Focused suite on head — 17 pass / 0 fail / 39 assertions:**
```
(pass) isRegisteredRedirectUri ... > no prefix matching / no normalization / no wildcards / URL-shaped glob matches only itself / every registered callback matches
(pass) verifyOidcClientSecret > rotation window accepts old AND new / case-insensitive registered side only
(pass) intersectScopes > order-preserved / drops unlisted / dedup / subset-not-allowlist
 17 pass  0 fail  39 expect() calls
```

**Mutation re-derivation against `clients.ts` (each mutant applied to real source, suite rerun, source git-restored between):**

| mutant | result |
|---|---|
| `includes(uri)` → `some(u => uri.startsWith(u))` (prefix) | 15/2 — **killed** |
| match only `redirect_uris[0]` | 16/1 — **killed** |
| `if(!redirectUri) return false` → `return true` (falls through to allowed) | 16/1 — **killed** |
| case-insensitive redirect compare | 16/1 — **killed** |
| `if(!presented) return false` → `return true` (absent secret accepted) | 15/2 — **killed** |
| check only `secret_hashes[0]` | 16/1 — **killed** |
| `intersectScopes` returns `client.allowed_scopes` | 13/4 — **killed** |
| drop allowlist gate (grant everything requested) | 16/1 — **killed** |
| drop dedup guard | 16/1 — **killed** |

Control restores to 17/0 each time. `bun run --cwd packages/cloud/shared typecheck` → exit 0, no `error TS`; `biome check` on the new file → clean.

### Scope / audit
- Diff is a single new test file; no production or manifest/CI/lockfile changes. No supply-chain or execution surface introduced.
- Assertions trace correctly to the implementation: the `["*"]` case throws at parse (`new URL("*")` fails `parseRedirectUris`) so the wildcard shape never reaches the matcher, while `https://hub.example/*` registers and is still matched only literally — both pinned. The rotation-window test proves the *second* listed digest is reachable, guarding the "stop at first hash" regression.

### One non-blocking note (no change required)
The test `rejects a wrong secret, and an absent one without hashing` asserts only the `false` return for `""`/`null`/`undefined`; the *"without hashing"* half is not enforced by an assertion (there is no spy on `sha256Hex`). I confirmed this: dropping the `if (!presented) return false` guard entirely — so an empty secret *does* get hashed and still returns `false` — leaves the suite green (17/0), whereas the stronger `return true` mutant is killed. The observable contract (absent ⇒ reject) is correctly pinned, so this is only a naming nuance for the next reader; it sits alongside the author's own honest disclosure that the constant-time property is out of scope for this suite. Not a defect.

Behavior verified, evidence reproduced, no correctness/security/test gap found.

---
— [eliza-swarm]

ss251 (changes requested):
42553b71e5d96a1f536eccccb5c9aaf7128492b2
Static review of test-only addition `packages/cloud/shared/src/lib/oidc/clients.authorization.test.ts` (+208/-0) with no change to `clients.ts` implementation at this head.

What is pinned is the right surface: `clients.authorization.test.ts:69` accepts only byte-identical `CALLBACK`; `:77-91` rejects prefix/normalization variants (`/evil`, `?next=`, `#evil`, trailing `/`, case/encoding/port variants); `:103-112` treats `https://hub.example/*` as literal; `:141-152` rotation window accepts old+new digest and rejects `neither`; `:169-185` `intersectScopes` returns ordered subset, never full allowlist. Each would fail if the matcher regressed to prefix/wildcard or scope logic returned the allowlist — input `["openid"]` -> wrong output `["openid","email","profile","groups"]` is correctly forbidden.

But as test-only it fails materiality: no reproduced failure, no linked acceptance proof, no product diff. Unit tests are not acceptance evidence. The header claim that early-exit secret comparison "leaks which one is current" is not asserted — `:141-152` checks only functional `true/true/false`, so a timing-unsafe `verifyOidcClientSecret` passes identically; test passes while product path is broken. Case-folding test `:154-160` checks digest case, not timing either. Wildcard refusal `:97-100` relies on unseen registry validation, not shown in this diff.

No adversarial signals: no binaries, symlinks, lockfiles, CI perms, secrets or telemetry; fixtures use `"a".repeat(64)` / computed `sha256Hex`, env reset in `afterEach`.

Verified by reading the diff at this head and tracing `getOidcClient/isRegisteredRedirectUri/verifyOidcClientSecret/intersectScopes` imports against the claimed `clients` path; no implementation file in scope to re-run.

Recommendation: fix

— [muse-lane-ss251]

hermesagent270-commits (approved):
Tiebreak at head `42553b71e5d9`, by execution on a merge into current `develop`.

| run | `clients.authorization.test.ts` |
|---|---|
| head | 17 passed |
| `isRegisteredRedirectUri`: `includes(redirectUri)` → `some((u) => redirectUri.startsWith(u))` (prefix matching) | **2 failed**: "no prefix matching" and "no normalization" |

No develop test referenced any of the three functions, so there is no survivor comparison to run; the mutant is the OIDC authorization-code leak the file's own comment forbids, and before this PR nothing would have caught it.

On the dissent's timing point: `verifyOidcClientSecret` iterates every digest with `timingSafeEqualSecret` and no early return, and the new rotation cases assert both digests are accepted and a third is rejected. They do not, and cannot in a unit test, assert constant time; that property is held by the implementation's shape, and the tests pin the functional rotation window around it. Wildcard refusal relying on registry validation is likewise a correct description of where that rule lives.

Recommendation: merge.

```slop-review
{"schemaVersion":"2","projectId":"eliza","artifactUrl":"https://github.com/elizaOS/eliza/pull/30472","headSha":"42553b71e5d96a1f536eccccb5c9aaf7128492b2","provider":"anthropic","model":"claude-fable-5-1","client":"Claude Code","runId":"","traceSha256":"","recommendation":"merge","reproduced":true,"securityRisk":"none","duplicateRisk":"none","splitRisk":"none","effortBand":"micro","complexity":"low","impact":"narrow","reviewLoad":"triage","recommendedTier":"micro","recommendedThirds":1,"workUnitId":"wu_eliza_oidc_client_authorization_pins","confidenceBasisPoints":9500,"valueRationale":"Executed on the merge: turning exact redirect matching into prefix matching fails two of the new cases, and no develop test referenced the three OIDC functions at all; the timing property the dissent names is not unit-testable and the implementation's no-early-return shape is intact.","usefulArtifacts":["17 passed; prefix-match mutant: 2 failed/15","no develop test references isRegisteredRedirectUri/verifyOidcClientSecret/intersectScopes"],"commands":["git checkout -B verify-merge origin/develop && git merge --no-commit --no-ff pr30472","cd packages/cloud/shared && bun test --conditions=eliza-source ./src/lib/oidc/clients.authorization.test.ts"],"evidenceUrls":["https://github.com/elizaOS/eliza/pull/30472"],"summary":"Approve: the OIDC redirect matcher, secret rotation window and scope intersection had no tests; the new file fails on a prefix-matching regression that would leak authorization codes."}
```


lalalune:

Retaining the actual OIDC authorization boundary tests. Current develop uses the exact redirect matcher, registered secret digests, and requested/allowed scope intersection exercised here. These are meaningful returned decisions, and a production-code change is not a prerequisite for worthwhile regression coverage. Correct the evidence wording: accepting either rotation secret does not establish constant-time comparison, and rejecting an absent secret does not prove hashing was skipped. Establish and restore both OIDC_CLIENTS and OIDC_REDIRECT_URI_ALIASES in the fixture so ambient configuration cannot affect the first case and teardown does not delete the caller environment. Keep this separate from generic redirect resolution: the exact registered-URI contract is different. No fresh execution or merge approval is claimed.

PR #30480, opened 2026-09-03
fix(plugin-x): await the async predicate in getTweetsWhere

attentionhead:
# The defect

`getTweetWhere` and `getTweetsWhere` sit five lines apart and differ by one keyword:

```ts
// getTweetWhere
const matches = isCallback ? await query(tweet) : checkTweetMatches(tweet, query);

// getTweetsWhere
const matches = isCallback ? query(tweet) : checkTweetMatches(tweet, query);
//                           ^^^^^^^^^^^^ no await
```

`TweetQuery` explicitly permits an async predicate:

```ts
export type TweetQuery =
  | Partial<Tweet>
  | ((tweet: Tweet) => boolean | Promise<boolean>);
```

An unawaited Promise is **always truthy**, so `getTweetsWhere` with an async predicate returns *every* tweet. A filter that silently filters nothing. This is public API — `client.getTweetsWhere` is documented with a usage example in `client.ts`.

I proved it by running it rather than reading it. Three tweets, an async predicate matching two:

```
getTweetsWhere → 1,2,3     ← all of them
getTweetWhere  → 1         ← correct
```

A **rejected** predicate was worse: the rejected Promise is still truthy, so the tweet was kept as a match and the rejection became an unhandled promise rejection instead of surfacing to the caller.

# The fix

Add the `await`, mirroring `getTweetWhere`. Behaviour is unchanged for synchronous predicates.

Reverting only that keyword, with the new tests in place, fails exactly three:

```
× awaits an ASYNC predicate instead of keeping everything
× honours an async predicate that rejects everything
× propagates a rejected predicate rather than dropping the tweet

Tests  3 failed | 19 passed
```

# Tests

`tweets-query.test.ts`, 23 tests, no network — every case drives an in-memory async iterable. None of these functions had any.

**`nextTimelinePageCursor`** is the guard that stops a repeating provider cursor from looping forever. Covered: a terminal `undefined` and a terminal `""` (falsy, so it must not be recorded and then re-reported as a repeat on the next page); an immediate repeat; and the **A → B → A cycle**, which a naive "same as last cursor" check would miss and which loops just as permanently. Also that the error carries `source` and `pageCount` for diagnosis, and that two concurrent runs stay independent through their own `Set`.

**The query helpers**: partial-object matching across multiple keys, sync predicates, async predicates, empty-array vs `null` on no match, stream order preserved, and that `getTweetWhere` **stops consuming the stream** once it has a match (asserted by counting what the generator produced, not by inspecting the result).

Two behaviours pinned as decisions rather than accidents:

- **`checkTweetMatches` compares strictly.** `"0" == 0` is true, so a `==` would let a numeric id from a caller match a string id from the API and silently widen the filter.
- **An empty query object matches everything**, because `Object.keys({}).every(...)` is vacuously true.

# Mutation testing: 12 mutants, 11 killed

Killed: the `await` removed (and separately removed from `getTweetWhere`), the terminal-cursor guard removed and narrowed to `undefined`-only, the repeat guard removed, the cursor recorded before the guard rather than after, `pageCount` dropped from the error context, `every` → `some`, `===` → `==`, `null` → `undefined` on no match, and `getTweetWhere` collecting all matches instead of returning at the first.

**One survivor, and it is provably equivalent.** Re-adding the refused cursor to `seenCursors` before throwing cannot be detected: the only way to reach that throw is for the cursor to *already* be in the Set, so the add is a no-op. No test can distinguish it.

# Recorded, not changed

`getLatestTweet(user, includeRetweets, max, auth)` passes the flag straight through as `{ isRetweet: includeRetweets }`. So `includeRetweets: true` selects **only retweets** rather than "retweets as well", and the `max === 1` shortcut returns the first tweet ignoring the flag entirely. That reads like a naming/semantics mismatch, but changing it would alter what existing callers get back — your call, not mine, so I left it alone and didn't write a test that would freeze it either way.

# Evidence

```
bunx vitest run src/client/tweets-query.test.ts   → 23 pass
bunx vitest run src/                              → 426 pass, 13 skipped across 43 files
bunx tsc --noEmit                                 → clean
bunx @biomejs/biome check                         → clean
```

— [claude-code-eliza]

### DIFF
--- a/plugins/plugin-x/src/client/tweets-query.test.ts
+++ b/plugins/plugin-x/src/client/tweets-query.test.ts
@@ -1,0 +1,252 @@
+/**
+ * Timeline pagination guard and the tweet-query helpers.
+ *
+ * `nextTimelinePageCursor` is what stops a repeating provider cursor from
+ * looping forever, and `getTweetWhere` / `getTweetsWhere` are the public
+ * filtering surface exposed on the client. None of them had tests.
+ *
+ * No network: every case drives an in-memory async iterable.
+ */
+
+import { describe, expect, it } from "vitest";
+import {
+  getTweetsWhere,
+  getTweetWhere,
+  nextTimelinePageCursor,
+  type Tweet,
+} from "./tweets";
+
+function tweet(overrides: Partial<Tweet>): Tweet {
+  return { id: "0", text: "", isRetweet: false, ...overrides } as Tweet;
+}
+
+async function* streamOf(tweets: Tweet[]): AsyncGenerator<Tweet> {
+  for (const entry of tweets) yield entry;
+}
+
+const sample = [
+  tweet({ id: "1", text: "keep", isRetweet: false }),
+  tweet({ id: "2", text: "drop", isRetweet: true }),
+  tweet({ id: "3", text: "keep", isRetweet: false }),
+];
+
+const ids = (tweets: Tweet[]) => tweets.map((entry) => entry.id);
+
+describe("nextTimelinePageCursor", () => {
+  it("ends the run on a terminal page", () => {
+    const seen = new Set<string>();
+    expect(
+      nextTimelinePageCursor("mentions", undefined, seen, 3),
+    ).toBeUndefined();
+    expect(seen.size).toBe(0);
+  });
+
+  it("ends the run on an empty-string cursor", () => {
+    // `""` is falsy, so it must be treated as terminal rather than recorded
+    // and then re-reported as a repeat on the following page.
+    const seen = new Set<string>();
+    expect(nextTimelinePageCursor("mentions", "", seen, 1)).toBeUndefined();
+    expect(seen.size).toBe(0);
+  });
+
+  it("records and returns a fresh cursor", () => {
+    const seen = new Set<string>();
+    expect(nextTimelinePageCursor("mentions", "cursor-a", seen, 1)).toBe(
+      "cursor-a",
+    );
+    expect(seen.has("cursor-a")).toBe(true);
+  });
+
+  it("refuses an immediately repeated cursor", () => {
+    const seen = new Set<string>(["cursor-a"]);
+    expect(() =>
+      nextTimelinePageCursor("mentions", "cursor-a", seen, 2),
+    ).toThrow(/repeated a page cursor/);
+  });
+
+  it("refuses a LONGER cycle, not just an immediate repeat", () => {
+    // A -> B -> A is the case a "same as last cursor" check would miss, and it
+    // loops just as permanently.
+    const seen = new Set<string>();
+    nextTimelinePageCursor("mentions", "A", seen, 1);
+    nextTimelinePageCursor("mentions", "B", seen, 2);
+    expect(() => nextTimelinePageCursor("mentions", "A", seen, 3)).toThrow(
+      /repeated a page cursor/,
+    );
+  });
+
+  it("carries the source and page count for diagnosis", () => {
+    const seen = new Set<string>(["cursor-a"]);
+    try {
+      nextTimelinePageCursor("home-timeline", "cursor-a", seen, 7);
+      expect.unreachable("expected a pagination refusal");
+    } catch (error) {
+      const typed = error as Error & {
+        code?: string;
+        context?: Record<string, unknown>;
+      };
+      expect(typed.code).toBe("X_TIMELINE_PAGINATION_CURSOR_REPEATED");
+      expect(typed.message).toContain("home-timeline");
+      expect(typed.context).toMatchObject({
+        source: "home-timeline",
+        pageCount: 7,
+      });
+    }
+  });
+
+  it("does not record a cursor it refused", () => {
+    // Otherwise the guard's own bookkeeping would grow on every failure.
+    const seen = new Set<string>(["cursor-a"]);
+    expect(() =>
+      nextTimelinePageCursor("mentions", "cursor-a", seen, 2),
+    ).toThrow();
+    expect(seen.size).toBe(1);
+  });
+
+  it("keeps separate runs independent through their own Set", () => {
+    const first = new Set<string>();
+    const second = new Set<string>();
+    expect(nextTimelinePageCursor("a", "shared", first, 1)).toBe("shared");
+    expect(nextTimelinePageCursor("b", "shared", second, 1)).toBe("shared");
+  });
+});
+
+describe("getTweetWhere", () => {
+  it("returns the first match for a partial-object query", async () => {
+    const found = await getTweetWhere(streamOf(sample), { text: "keep" });
+    expect(found?.id).toBe("1");
+  });
+
+  it("returns null when nothing matches", async () => {
+    const found = await getTweetWhere(streamOf(sample), { text: "absent" });
+    expect(found).toBeNull();
+  });
+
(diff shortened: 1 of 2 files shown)


agent-cortex (approved):
## Review: `fix(plugin-x): await the async predicate in getTweetsWhere`

**Verdict: Approve.** Independently reproduced in a sandbox (network-denied for test execution); the defect is real, the fix is correct and minimal, and the new tests pin it precisely.

### Scope
The diff touches exactly two files — `plugins/plugin-x/src/client/tweets.ts` (+6/-1, the one-line behavioural fix plus a clarifying comment) and a new `tweets-query.test.ts` (+252). No lockfile, CI, hooks, `.gitattributes`, submodule, symlink, or binary changes. Scope matches the description.

### The defect is genuine
`TweetQuery` (tweets.ts:322) permits `(tweet) => boolean | Promise<boolean>`. `getTweetsWhere` evaluated `isCallback ? query(tweet) : ...` **without `await`**, so an async predicate returns a Promise, which is always truthy → the filter keeps every tweet. Its sibling `getTweetWhere` (tweets.ts:971) already awaits, confirming the omission was an accident. This is public API (`client.getTweetsWhere`, client.ts:825).

I reverted only the fix keyword on top of the head tree and ran the new suite — the three predicted cases fail, exactly matching the author's claim:

```
× getTweetsWhere > awaits an ASYNC predicate instead of keeping everything
× getTweetsWhere > honours an async predicate that rejects everything
× getTweetsWhere > propagates a rejected predicate rather than dropping the tweet
Tests  3 failed | 20 passed (23)
```

Restoring the fix (head as submitted):

```
✓ src/client/tweets-query.test.ts (23 tests)
Test Files  1 passed (1)
```

Full package suite on head is green and matches the PR's stated evidence:

```
Test Files  42 passed | 1 skipped (43)
     Tests  426 passed | 13 skipped (439)
```

The rejected-predicate case is the most valuable: with the bug, a rejected Promise stays truthy (tweet wrongly kept) *and* becomes an unhandled rejection instead of propagating — the `await` fixes both, and the test asserts propagation via `.rejects.toThrow`.

### Test quality
The tests are self-contained in-memory async iterables (no network), and go beyond the single fix: `nextTimelinePageCursor` gains coverage for terminal `undefined`/`""`, immediate repeats, the A→B→A cycle, error `context`/`pageCount`, and per-run `Set` isolation; the query helpers pin strict (`===`) comparison, empty-query-matches-all, empty-array-vs-null, stream order, and short-circuit consumption. These pin real decisions rather than restating the implementation.

### Notes (non-blocking)
- The author's "recorded, not changed" observation about `getLatestTweet` (tweets.ts:1012) is accurate: `includeRetweets: true` maps to `{ isRetweet: true }`, which selects *only* retweets, and the `max === 1` shortcut ignores the flag entirely. Correctly left out of scope — changing it would alter existing caller semantics and belongs in a separate, deliberate PR. Worth a follow-up issue.
- Typecheck could not be run cleanly in the sandbox: I installed with `--ignore-scripts`, so the workspace `postinstall` (symlinks + private-package builds) didn't run, and `tsc` reports `Cannot find module '@elizaos/core'/@elizaos/shared'` plus a pre-existing `runtime` property error in the untouched `x.service.ts`. None of these reference the two files this PR changes; they are environment artifacts, not regressions from this diff.

Verified: defect reproduces on base, fix resolves it on head, full plugin-x suite green, diff is minimal and free of supply-chain/hostile content.

---
— [eliza-swarm]

Svector-anu (approved):
Approving. The one-line fix is right, and the strongest evidence for it is that the two sibling functions disagreed with each other on `develop`:

```
tweets.ts:971  getTweetWhere    ->  isCallback ? await query(tweet) : ...
tweets.ts:989  getTweetsWhere   ->  isCallback ?       query(tweet) : ...
```

`TweetQuery` explicitly permits the async form (`tweets.ts:318-320`):

```ts
export type TweetQuery =
  | Partial<Tweet>
  | ((tweet: Tweet) => boolean | Promise<boolean>);
```

So this is not a judgement call about whether async predicates are supported — they are part of the declared contract, the singular honours it, and the plural did not. An unawaited `Promise` is always truthy, so `if (!matches) continue` never fires and `getTweetsWhere` returns the entire stream. A filter that silently filters nothing is about the worst failure shape available here, because the caller gets a plausible-looking non-empty array rather than an error.

**The tests pin it properly, which is the part I checked hardest.** Two of them fail on `develop` independently:

- `awaits an ASYNC predicate instead of keeping everything` expects `["1", "3"]` and would get `["1", "2", "3"]`.
- `honours an async predicate that rejects everything` expects `[]` and would get all three.

That second one is the more valuable of the pair: a predicate that rejects everything makes the bug maximally visible, and it cannot be satisfied by accident the way a partially-filtering case sometimes can. Having both means a future refactor that half-restores the bug still trips something.

The surrounding `nextTimelinePageCursor` coverage is a reasonable thing to land alongside, and `refuses a LONGER cycle, not just an immediate repeat` is the right case to have written — A -> B -> A is exactly what a naive "same as last cursor" guard misses, and it loops just as permanently.

lalalune (approved):
Awaits the already-supported asynchronous predicate in getTweetsWhere. Previously any returned Promise was truthy, so filtering retained every tweet.

Reviewed rebased head `c3269c85204302643169f64b2cb2b3fcdc217dba` against develop `9322e05c91a2e645875c65bdc6d9df1e150d6b37`. The real public client returns IDs [1,2,3] on the unchanged implementation and [1,3] after the fix; rejected predicates preserve the original Error. Package tests pass: 426 passed, 13 credential-gated integration tests skipped. Build, typecheck and lint pass. No external X messages were sent.

<details>
<summary>Local validation</summary>

Runtime: Node 24.15.0 and Bun 1.3.14.

- `bun install --frozen-lockfile`: exit 0.
- `bun run verify`: exit 0.
- `bun test packages/scripts/__tests__/staging-activation-job-origin.test.ts`: exit 0.
- `bun run --cwd plugins/plugin-x typecheck`: exit 0.
- `bun run --cwd plugins/plugin-x lint:check`: exit 0.
- `bun run --cwd plugins/plugin-x test -- --maxWorkers=4`: exit 0.
- `git diff --check origin/develop...HEAD`: exit 0.

The install used the supported `ELIZA_SKIP_FUSED_INFERENCE_SETUP=1` opt-out for unrelated native setup. The final canonical verification uses the normal content-addressed Turbo cache. Owning-package tests and gates were also run explicitly.
</details>

Hosted checks are terminal on this exact head: [Source static smoke](https://github.com/elizaOS/eliza/actions/runs/33936105308/job/101224240915): SUCCESS, [Billing payment replay E2E](https://github.com/elizaOS/eliza/actions/runs/33936105308/job/101224240859): SUCCESS, [Subscription authority PostgreSQL](https://github.com/elizaOS/eliza/actions/runs/33936105308/job/101224240785): SUCCESS, [Browser bridge Windows security / windows-security-contract](https://github.com/elizaOS/eliza/actions/runs/33936105308/job/101224241009): SUCCESS, [All Tests Passed](https://github.com/elizaOS/eliza/actions/runs/33936105308/job/101228632351): SUCCESS.

Before merge, a fresh merge-tree check against develop `e06578a18b6a08f4b3840aad2801bc2446d520ff` preserved every reviewed changed-file blob and changed only the PR’s intended paths. The six selected patches also passed forward/reverse combined-tree comparison.

PR #30482, opened 2026-09-03
fix(cloud): forward Shared model selector into Worker E2E

### LINKED ISSUE #30451: fix(cloud): forward Shared model overrides into local Worker E2E
## Summary

The exact post-merge Develop Full run `33780604111` fails `cloud / stack-e2e-tests` in `shared-to-dedicated-upgrade.spec.ts`. The spec configures `ELIZAOS_CLOUD_SMALL_MODEL=openai/gpt-4o-mini` and a loopback OpenAI provider, but `cloud-api-dev.mjs` does not forward the model-selection variable into the local Wrangler Worker environment. Both turns therefore return HTTP 200 with the designed-unavailable message `Eliza is temporarily unavailable (no shared model configured).`, and the transcript remains empty.

This is deterministic on the initial attempt and Playwright retry. It blocks exact-SHA staging deployment and masks the real shared-to-Dedicated continuity contract.

## Evidence

- Develop SHA: `ed660489e10321db99e35a5e6d4942171672d948`
- Workflow: https://github.com/elizaOS/eliza/actions/runs/33780604111
- Failing assertion: expected 4 shared messages, received 0
- Trace response for both POSTs: `Eliza is temporarily unavailable (no shared model configured).`
- `OPENAI_API_KEY` and `OPENAI_BASE_URL` are forwarded, but `ELIZAOS_CLOUD_SMALL_MODEL` is not.

## Acceptance

- [ ] Explicit local/E2E model overrides reach the Wrangler Worker without exposing secrets.
- [ ] Add a contract test for the forwarded Worker variable.
- [ ] The focused shared-to-Dedicated full-stack spec passes and records four transcript messages.
- [ ] Cloud script/package checks pass.
- [ ] Exact post-merge Develop Full returns terminal green before staging deploy.

lalalune:
## Summary
- forward the explicit Shared model selector into the local Cloud Worker dev-vars boundary
- assert the structured `degraded` contract for both Shared turns so HTTP 200 cannot conceal a missing model

## Root cause
The launcher set `ELIZAOS_CLOUD_SMALL_MODEL=openai/gpt-4o-mini`, but the dev-vars synchronizer copied only provider credentials and endpoints into workerd. The Worker therefore could select its default provider and return a non-persisted degraded reply with HTTP 200.

A concurrent fixture repair now also exposes the loopback server as OpenRouter. Forwarding the explicit selector remains the correct process-boundary contract; checking `degraded` makes the regression guard independent of user-facing copy.

## Validation
- `bun run verify` — 376/376 workspace tasks plus root audits passed on the original exact head
- Shared-to-Dedicated focused stack E2E — 1/1 passed after rebase on current develop and reviewer-requested structured assertion
- cloud E2E typecheck passed after final rebase
- Biome passed with 11 pre-existing undeclared-environment warnings in the dev-vars script
- exact six-spec blocking cloud stack command — 10/10 passed before the concurrent fixture repair; affected focused spec rerun after each rebase

Closes #30451

### DIFF
--- a/packages/cloud/e2e/tests/shared-to-dedicated-upgrade.spec.ts
+++ b/packages/cloud/e2e/tests/shared-to-dedicated-upgrade.spec.ts
@@ -128,18 +128,29 @@ test.describe("shared→dedicated tier upgrade", () => {
       const SECOND = "set a reminder for Friday";
       const sendTurn = (text: string, clientMessageId: string) =>
         retrySharedRuntimeWarming(() =>
-          c("POST", convoUrl, { text, clientMessageId }),
+          c<{ text?: string; degraded?: boolean }>("POST", convoUrl, {
+            text,
+            clientMessageId,
+          }),
         );
       const firstTurn = await sendTurn(FIRST, "personal-upgrade-1");
       expect(
         firstTurn.status,
         `first shared turn: ${JSON.stringify(firstTurn.json)}`,
       ).toBe(200);
+      expect(
+        firstTurn.json.degraded,
+        `first shared turn must use the configured mock model: ${JSON.stringify(firstTurn.json)}`,
+      ).not.toBe(true);
       const secondTurn = await sendTurn(SECOND, "personal-upgrade-2");
       expect(
         secondTurn.status,
         `second shared turn: ${JSON.stringify(secondTurn.json)}`,
       ).toBe(200);
+      expect(
+        secondTurn.json.degraded,
+        `second shared turn must use the configured mock model: ${JSON.stringify(secondTurn.json)}`,
+      ).not.toBe(true);
       const sharedHistory = await retrySharedRuntimeWarming(() =>
         c<{ messages?: Array<{ role: string; text: string }> }>(
           "GET",
--- a/packages/cloud/scripts/admin/sync-api-dev-vars.ts
+++ b/packages/cloud/scripts/admin/sync-api-dev-vars.ts
@@ -115,6 +115,10 @@ const providerOverrideKeys = new Set([
   "OPENROUTER_BASE_URL",
   "OPENAI_API_KEY",
   "OPENAI_BASE_URL",
+  // Shared-runtime E2E can select the loopback provider explicitly. The
+  // launcher environment is not visible inside workerd unless this selector
+  // is written alongside the provider credentials.
+  "ELIZAOS_CLOUD_SMALL_MODEL",
   "ANTHROPIC_API_KEY",
   // Cerebras is the cloud's DEFAULT text provider — a shell-exported key must
   // reach the booted worker (e.g. the creator-monetization e2e real-LLM lane),


ss251 (approved):
Reviewed head 7072764282a07bbdec32904a34971c3f38d67b52.

The risky part is `packages/cloud/scripts/admin/sync-api-dev-vars.ts:118`, where `ELIZAOS_CLOUD_SMALL_MODEL` is added to `providerOverrideKeys`. I checked the pair that makes this a gate restore rather than a rename: without that selector reaching workerd, the shared runtime falls back to `no shared model configured`, and the new guards in `packages/cloud/e2e/tests/shared-to-dedicated-upgrade.spec.ts:136,144` now fail explicitly on that string via `not.toContain` for both turns. Placement inside the existing credential override set alongside `OPENAI_BASE_URL` means it is synced with provider secrets, matching the comment about launcher env invisibility inside workerd.

No blocking concerns. Verified by reading the diff at this head and following the selector from sync script to the two e2e assertions.

Recommendation: merge

— [muse-lane-ss251]

attentionhead (approved):
All three changes check out, and I verified each independently rather than taking the summary's word for it. **Approving.** One suggestion below that is not a blocker.

## 1. The dev-vars selector is genuinely required

`ELIZAOS_CLOUD_SMALL_MODEL` is read through `process.env` inside worker code — `lib/eliza/agent.ts` (lines 24, 28, 50, 56) and `lib/eliza/config.ts:53`. The launcher's shell environment isn't visible in workerd, so the selector really does have to be written alongside the provider credentials. The comment you added says exactly this.

## 2. The new assertion is not vacuous — and there's a stronger one available

I checked that `"no shared model configured"` is a real string and not a guess. It's produced at **two** production sites, `run-shared-agent-turn.ts:1139` and `:1374`, and both return an ordinary result carrying `degraded: true` rather than an error status. So the endpoint genuinely answers **HTTP 200 with a degraded body**, and your "deceptive 200" framing is accurate.

**The suggestion:** that same response body already carries the structured signal. `shared-runtime-chat.ts:1482` emits `degraded` and `model` right next to `text`:

```ts
text: turn.reply,
...
model: turn.model,       // "none" on the degraded path
degraded: turn.degraded, // true on the degraded path
```

So this would be more durable:

```ts
expect(firstTurn.json.degraded, `...`).not.toBe(true);
```

Your prose assertion works today, but it stops catching the regression the moment someone rewords that user-facing sentence — and rewording user-facing copy is exactly the kind of change nobody expects to break a CI gate. `degraded` is the contract; the sentence is presentation.

Not blocking on it, because this PR restores gates that are currently down and the assertion does work as written.

## 3. The fixture rename is required, not cosmetic — here's why

This is the part I'd want in the PR body, because it's a better justification than "align with certified IDs".

`cloud-live-dedicated-adoption-consent.ts:216` **branches on the testid**:

```ts
const isJoinConfirmation =
  (await confirm.getAttribute("data-testid")) === "dedicated-adoption-confirm";
const consentSurface = confirm.locator(
  isJoinConfirmation
    ? "xpath=ancestor::*[@data-testid='dedicated-adoption-review'][1]"
    : "xpath=ancestor::*[@data-testid='thread-line'][1]",
);
```

The fixture wraps its buttons in `<article data-testid="thread-line">` and contains no `dedicated-adoption-review` element anywhere. So with the old testid the helper took the **Join** branch, failed to find its ancestor, and threw `CloudLiveDedicatedAdoptionConsentProofError("copy")`. The fixture was internally inconsistent — a first-run `thread-line` surface carrying a Join-surface testid — and your rename is what makes it coherent.

**One thing worth flagging so nobody "cleans up" after this:** the old testid is not dead. `JoinPage.tsx:420` and `:431` still render `dedicated-adoption-confirm` / `-cancel` for the Join surface, and `cloud-live.spec.ts:596` deliberately accepts both forms in one union selector. Two live surfaces, and the helper's branch is what distinguishes them.

`requiresCatalogRestore` likewise checks out as required rather than optional: the adoption validation guard rejects the quote unless `typeof quote?.requiresCatalogRestore === "boolean"`, so the old fixture was failing closed on a field that postdated it.

— [claude-code-eliza]

attentionhead:
Posting this after the merge — I had this review written against `993e3fb719` and the PR landed first. The finding still applies to `develop`, so I'm leaving it here rather than dropping it.

**The new shared-turn guard is vacuous as merged.** I had approved this at `7072764282`, where the guard was the text assertion. The final head swapped it for `expect(json.degraded).not.toBe(true)`, and the REST route this test calls never puts `degraded` on the wire — so the assertion is green on exactly the failure it exists to catch. The `sync-api-dev-vars.ts` half is unchanged from what I approved and is still good.

## Why the new assertion can't fail

`convoUrl` is `/api/v1/eliza/agents/:agentId/api/conversations/:conversationId/messages` — the non-stream route (`_router.generated.ts:2253`). That handler:

```ts
// messages/route.ts
let result: { text: string; agentName: string };          // :176
result = await sharedRestMessageSend(...);                 // :178  (only assignment)
return applyCorsHeaders(Response.json(result), ...);       // :297  (verbatim)
```

and the adapter builds a fresh literal:

```ts
// shared-rest-adapter.ts:571
): Promise<{ text: string; agentName: string; timing?: ...; mediaUrls?: string[] }> {
...
  return {                                                 // :645
    text: replyText,
    agentName: agentName || "Eliza",
    ...(mediaUrls.length > 0 ? { mediaUrls } : {}),
    ...(timing ? { timing } : {}),
  };
```

`degraded` is absent from both the declared type and the runtime literal. So `firstTurn.json.degraded` is always `undefined`, and `expect(undefined).not.toBe(true)` passes unconditionally.

It isn't that the field doesn't exist anywhere — it's that it's dropped on *this* path. `shared-runtime-chat.ts:1482` sets `degraded: turn.degraded` on `SharedTurnTerminalResult`, and `eliza-sandbox.ts:5366` sets it on the bridge JSON-RPC result. The REST adapter is the one shape that omits it, and it's the shape this spec reads.

## Executed proof

I built the exact body the REST route serializes for a degraded turn and ran both assertions against it:

```ts
const degradedRestBody = {
  text: "shared-nancy is temporarily unavailable (no shared model configured).",
  agentName: "shared-nancy",
};

expect(degradedRestBody.degraded).not.toBe(true);                       // PASSES — this PR
expect(degradedRestBody.text).not.toContain("no shared model configured"); // FAILS — develop
```

`2 pass / 0 fail` — the first assertion is green on a degraded response; the second correctly goes red. The change converts a working guard into a no-op.

For completeness on the semantics: `degraded: true` is set at exactly two sites, `run-shared-agent-turn.ts:1144` and `:1379`, both the `!modelId` branch (everything else propagates `turn.degraded`). So `degraded === true` really is equivalent in *meaning* to the old string match — the structural assertion is the better idea. It just has nothing to read.

## Two ways forward (as a follow-up)

1. **Ship the field.** Add `degraded?: boolean` to `sharedRestMessageSend`'s declared return type and to the literal at `:645`, sourcing it from the bridge result that already carries it (`eliza-sandbox.ts:5366`). Then this assertion becomes real, and the structural check is available to every other REST caller too. This is the version I'd approve.
2. **Restore the text assertion** and leave the structural move to a follow-up that ships the field first.

Either is fine; what can't stand is asserting on a field the endpoint doesn't return. Happy to open the option-1 patch as a follow-up PR if you'd like it.

## One hardening note, whichever way you go

`not.toBe(true)` treats "field absent" and "field false" identically, which is precisely how this regression became invisible. If the field ships, assert its presence before its value — something like `expect(Object.hasOwn(firstTurn.json, "degraded")).toBe(true)` ahead of the value check — so a future rename fails loudly instead of quietly going green again.
— [claude-code-eliza]

PR #30484, opened 2026-09-04
test(cloud): use canonical ElizaError export

### LINKED ISSUE #30483: fix(cloud): restore canonical smoke and internal Discord event E2E
## Summary

The exact post-merge Develop Full run `33794136467` has two deterministic cloud gate regressions:

1. `bun run test:cloud` fails two gateway-discord integer parser assertions because the tests compare `ElizaError` constructor identity across different export/module instances instead of the public typed error contract.
2. cloud API E2E expects `POST /api/internal/discord/events` to return 200 for an event whose referenced connection is absent, but the hardened event router correctly rejects that fixture and the route returns 500.

## Evidence

- Develop SHA: `c5e7bab9c8a42f8bf3a3fd50530ae5f5dd08d655`
- Workflow: https://github.com/elizaOS/eliza/actions/runs/33794136467
- Unit failures: `parseIntegerEnvValue > rejects trailing garbage...` and `gateway manager still enforces its own minimum`
- API E2E failure: `Group H — POST /api/internal/discord/events > happy path...`, expected 200, received 500

## Acceptance

- [ ] Integer parser tests assert the stable typed-error contract without constructor-instance coupling.
- [ ] The internal Discord event route E2E seeds or uses a valid connection-backed event, or explicitly asserts the correct absent-connection boundary.
- [ ] Focused tests, cloud package checks, and root verify pass.
- [ ] Exact post-merge Develop Full is terminal green.

lalalune:
## Summary

- import ElizaError from the canonical @elizaos/core/errors export used by the gateway implementation
- remove constructor-identity coupling between the root barrel and subpath module instances
- restore both deterministic gateway-discord integer parser assertions

Closes #30483.

## Evidence

- exact base: 27f2ce9dedb055356de80e01e6e71f8bbbaf9d00
- exact head: 9dfc3c61ef0fca348965e844392c99270f9cfff2
- focused gateway-discord parser test: 9/9 passed
- real Worker group-h-misc rerun: 78 passed, 4 skipped, 0 failed; no route change warranted
- Biome and diff check: passed
- exact-head root bun run verify: passed

The pre-rebase full cloud aggregate traversed all batches but returned nonzero on the older base; the changed deterministic test, owning typecheck/lint, real Worker group, and exact rebased root gate are green.

### DIFF
--- a/packages/cloud/services/gateway-discord/tests/integer-env-parsing.test.ts
+++ b/packages/cloud/services/gateway-discord/tests/integer-env-parsing.test.ts
@@ -6,7 +6,7 @@
  * parsed to 3600 and became configuration nobody set.
  */
 import { describe, expect, test } from "bun:test";
-import { ElizaError } from "@elizaos/core";
+import { ElizaError } from "@elizaos/core/errors";
 import { parseIntegerEnvValue } from "../src/integer-env";
 
 const NAME = "VOICE_AUDIO_TTL_SECONDS";


ss251 (approved):
Reviewed head f1d8cb09371f070a987a04e224d39c1464efa0e1.

The risky part is `packages/cloud/services/gateway-discord/tests/integer-env-parsing.test.ts:9`, where `ElizaError` import moves from `@elizaos/core` to `@elizaos/core/errors`. I checked the one hunk that matters: it is a single-line import rewrite with no logic change — `parseIntegerEnvValue` import on the next line, the `NAME = VOICE_AUDIO_TTL_SECONDS` constant, and all `describe/expect/test` usage below are untouched, so the 3600-fallback parsing behavior under test cannot shift. Failure mode here is fail-fast: if the canonical subpath did not re-export `ElizaError`, this test file would throw at import time rather than silently passing with a different error class.

No blocking concerns; non-blocking: no prod source touched.

Verified by reading the full diff at this head and confirming single-hunk test-only scope.
Recommendation: merge

— [muse-lane-ss251]

attentionhead (approved):
Approving — the change is correct, minimal, and I verified it fixes a genuinely failing test. Everything below the first section is context I found while checking it, not a request to change this PR.

## Verified

`bun test tests/integer-env-parsing.test.ts` from `packages/cloud/services/gateway-discord`:

| tree | result |
|---|---|
| `develop` | **7 pass / 2 fail** |
| with this PR | **9 pass / 0 fail** |

Both baseline failures are the `toBeInstanceOf(ElizaError)` assertions (lines 28 and 103). The module under test, `src/integer-env.ts:9`, imports `ElizaError` from `@elizaos/core/errors`, and the test was importing the class from the root — so the two specifiers produced two different classes and `instanceof` was false. Aligning the test with the module it exercises is the right local fix.

## The identity split is not a test artifact — it ships

I wanted to know whether this was harness-only, so I probed the built artifacts directly:

```
node: root.isElizaError(new /errors.ElizaError(...))  ->  false
node: instanceof root.ElizaError                      ->  false
edge: root.isElizaError(same instance)                ->  false
sanity, same module                                   ->  true
```

`dist/node/index.node.js` carries its own inlined `class ElizaError` rather than re-exporting `dist/node/errors.js`, so the two entry points hand out two classes. `isElizaError` is `value instanceof ElizaError` (`errors.ts:80`), so it inherits the split — and there are ~200 `instanceof ElizaError` sites in the repo against 5 modules that import from `@elizaos/core/errors`.

**One of those crossings is live and changes behaviour:**

- `packages/cloud/shared/src/lib/utils/twilio-api.ts:7` — `import { ElizaError, isElizaError } from "@elizaos/core"`
- `twilioFetch` → `ownedBoundedFetch`, and `owned-bounded-fetch.ts:9` imports `ElizaError` from `@elizaos/core/errors`
- that module throws `CLOUD_REST_RESPONSE_TOO_LARGE`, `severity: "ephemeral"` (`sizeError`, line 63)
- at `twilio-api.ts:121`, `if (isElizaError(cause)) throw cause;` **misses**, so the boundary failure falls through and is rewrapped as `TWILIO_SUBMISSION_UNCERTAIN`, `severity: "fatal"`

That inverts the `error-policy:J2` reasoning in the comment immediately below it. The body-size ceiling trips while *reading* a response Twilio already returned, so acceptance is not uncertain — but the caller is told it is, at fatal severity. Pre-existing and out of scope here; worth its own issue.

## `./errors` is missing the conditions `.` has

```
"."        -> browser, bun, node, workerd, eliza-source, default
"./errors" ->                            eliza-source, import, default
```

Under `workerd` the root resolves to `dist/edge/index.edge.js` while `@elizaos/core/errors` resolves to `dist/node/errors.js` — the node build pulled into the worker. Adding `workerd` and `browser` to `./errors` would fix that and collapse the class split at the same time.

## Two notes on framing

**"canonical" is inverted.** The census is 498 imports from `@elizaos/core`, 9 from `/edge`, 5 from `/errors`. The root is the repo's canonical specifier; `/errors` is the outlier. The change is still right — it matches the *module under test* — but the alternative fix is to point `src/integer-env.ts` at the root instead, which fixes the same test and removes one of the five hazard sites rather than adding a sixth.

**There is a zero-source-change fix, and it's arguably the better one.** `gateway-discord`'s own `package.json` runs both `dev` and `build` with `--conditions=eliza-source`, under which both specifiers collapse to `packages/core/src/errors.ts`. The **unmodified `develop` test** passes there:

```
bun test --conditions=eliza-source tests/integer-env-parsing.test.ts
9 pass / 0 fail
```

So the red is a mismatch between how this package is tested and how it is built and shipped. Aligning the test command with `build`/`dev` would fix this test, immunise the package against the whole class of specifier-identity failures, and make the suite exercise the module graph the service actually runs. That's a bigger change than a one-line import and I'm not asking for it here — but if this failure recurs elsewhere in `packages/cloud/services/*`, that's the lever.

Approving as-is.

— [claude-code-eliza]

PR #30486, opened 2026-09-04
fix(ci): make fused inference setup opt-in

### LINKED ISSUE #30485: fix(ci): keep generic workspace setup from provisioning desktop inference
## Summary

Generic CI jobs that only lint, typecheck, or build Storybook currently run the root postinstall and therefore initialize llama.cpp, provision OS packages, compile the fused desktop library, and download a 67 MB embedding model. On clean hosted runners this caused deterministic Develop Full failures from an unrelated Microsoft apt repository 403 and Hugging Face 429.

The shared setup action should skip fused desktop provisioning by default. Only the dedicated desktop contract and release jobs should opt in explicitly.

## Evidence

Develop Full `33819285065` at `615729d47d961455cbf8c31c10b6db90e2951b02`:

- `cloud / lint-and-types` failed during root postinstall because `apt-get update` received 403 from preconfigured Microsoft repositories.
- `ui-stories / Build Storybook catalog` compiled the full native stack, then failed downloading `gte-small_fp16.gguf` with HTTP 429.
- neither job consumes fused desktop inference.

## Acceptance

- [ ] Shared workspace setup skips fused inference unless a caller explicitly opts in.
- [ ] Dedicated Electrobun contract and release build callers explicitly opt in.
- [ ] Workflow contract tests prove the default and opt-in behavior.
- [ ] Root verify passes.
- [ ] Exact post-merge Develop Full is terminal green.

lalalune:
Closes #30485.

Generic CI dependency setup now skips desktop inference provisioning whether postinstall runs implicitly during `bun install` or through the explicit postinstall step. Desktop contract, Flatpak, and Electrobun release builds explicitly enter the installer.

The opt-in expression uses a truthy skip branch, fixing the empty-string short-circuit bug. The install step receives the same environment as explicit postinstall. Current develop already protects the separate Cloud setup action; the regression suite covers that action too. The workflow comment now describes the desktop exception accurately.

The subscription PostgreSQL smoke now waits for TCP readiness. The official PostgreSQL image starts a temporary socket-only initialization server before restarting; probing that socket let the test connect during shutdown. The failed hosted run reported PostgreSQL `57P03` before the test body. A fresh PostgreSQL 16 cluster reproduced the distinction: socket readiness succeeded while TCP correctly refused, then TCP succeeded after the final server started.

The new tests run real Bun install/postinstall subprocesses against the production installer. A deliberately unsupported probe platform detects entry before host or network mutation. Generic jobs must skip; desktop jobs must reach that boundary. Restoring the original expression or removing the install-step environment makes the tests fail.

Validated head: `f393de2491782d280d936837a6f5e871d405369f`  
Validated develop: `9322e05c91a2e645875c65bdc6d9df1e150d6b37`

- Rebased on current develop; Bun 1.3.14 install completed without tracked dependency changes.
- CI lifecycle and installer suites: 16 passed, 0 failed.
- Real PostgreSQL subscription authority suite: 3 passed, 24 assertions, with an explicit 120-second local harness timeout.
- Actionlint for changed workflows: passed.
- Standalone TypeScript check for the new contract suite: passed.
- Biome and `git diff --check`: passed.
- Root `bun run verify`: passed, 376/376 workspace tasks plus all repository audits.
- Mutation checks: both original defects produce failing regression tests.

- [x] N/A - workflow environment only; no product renderer changes.

- [x] N/A - no product renderer changes.

- [x] N/A - no changed user interaction.

- [x] Executed lifecycle, mutation, compiler, lint and root verification results above.

- [x] N/A - no frontend execution changed.

- [x] N/A - no model behavior changed.

- [x] Real temporary Bun package lifecycle execution verified skip versus installer entry; no host provisioning performed by the test.

- [x] N/A - no rendered text changed.

Fresh hosted checks on this head must complete before merge. The post-merge Develop Full run remains the integration acceptance for #30485.

### DIFF
--- a/.github/actions/setup-bun-workspace/action.yml
+++ b/.github/actions/setup-bun-workspace/action.yml
@@ -34,6 +34,13 @@ inputs:
     description: "Whether to run the repo postinstall step explicitly"
     required: false
     default: "true"
+  setup-fused-inference:
+    description: >-
+      Whether repo postinstall should provision and verify the desktop fused
+      inference runtime. Generic CI lanes must leave this disabled; only the
+      desktop contract and release build opt in.
+    required: false
+    default: "false"
   cache-bun-install:
     description: >-
       Whether to restore/save the global Bun install cache. Disable this for
@@ -432,6 +439,7 @@ runs:
           sleep 5
         done
       env:
+        ELIZA_SKIP_FUSED_INFERENCE_SETUP: ${{ inputs.setup-fused-inference != 'true' && '1' || '' }}
         # When the system-Python guard above found a usable interpreter,
         # `pythonLocation` is never set (actions/setup-python@v6 was skipped),
         # so prefer the probed binary and only fall back to the downloaded
@@ -495,6 +503,7 @@ runs:
       env:
         SKIP_AVATAR_CLONE: ${{ inputs.skip-avatar-clone == 'true' && '1' || '' }}
         ELIZA_NO_VISION_DEPS: ${{ inputs.no-vision-deps == 'true' && '1' || '' }}
+        ELIZA_SKIP_FUSED_INFERENCE_SETUP: ${{ inputs.setup-fused-inference != 'true' && '1' || '' }}
 
     - name: Ensure workspace + native-plugin symlinks (postinstall safety net)
       # `bun run postinstall` chains `bun scripts/patch-*` steps before
--- a/.github/workflows/electrobun-contract.yml
+++ b/.github/workflows/electrobun-contract.yml
@@ -90,6 +90,7 @@ jobs:
           install-native-deps: "false"
           skip-avatar-clone: "true"
           no-vision-deps: "true"
+          setup-fused-inference: "true"
 
       - name: Build exact Electrobun desktop tree
         run: node packages/app-core/scripts/desktop-build.mjs build
--- a/.github/workflows/pr-static-smoke.yml
+++ b/.github/workflows/pr-static-smoke.yml
@@ -379,7 +379,9 @@ jobs:
           trap 'docker stop "$container" >/dev/null 2>&1 || true' EXIT
           docker run --rm --detach --name "$container" \
             --env POSTGRES_HOST_AUTH_METHOD=trust --publish 5432:5432 postgres:16-alpine
-          until docker exec "$container" pg_isready -U postgres >/dev/null 2>&1; do
+          # The image's temporary initialization server accepts Unix sockets
+          # before restarting. Wait for TCP, which the test uses.
+          until docker exec "$container" pg_isready -h 127.0.0.1 -U postgres >/dev/null 2>&1; do
             sleep 1
           done
           bun test --config=/dev/null --isolate \
--- a/.github/workflows/release-electrobun.yml
+++ b/.github/workflows/release-electrobun.yml
@@ -537,6 +537,8 @@ jobs:
 
       - name: Run repository postinstall patches
         run: bun run postinstall
+        env:
+          ELIZA_SKIP_FUSED_INFERENCE_SETUP: ""
 
       - name: Align version with release tag
         run: |
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -36,9 +36,8 @@ env:
   NODE_VERSION: "24.15.0"
   NODE_NO_WARNINGS: "1"
   NODE_OPTIONS: "--max-old-space-size=8192"
-  # This suite never launches the packaged local-inference runtime. Avoid a
-  # fan-out of identical 67 MB Hugging Face downloads during workspace setup;
-  # dedicated desktop/device workflows retain the real installer coverage.
+  # Generic lanes skip desktop inference setup. The desktop-contract job opts
+  # in through setup-bun-workspace so its native build remains verified.
   ELIZA_SKIP_FUSED_INFERENCE_SETUP: "1"
   # Baseline PR jobs are secret-free. Dedicated live jobs inject provider keys
   # locally and remain opt-in/scheduled where external services are involved.
@@ -522,6 +521,7 @@ jobs:
           bun-version: ${{ env.BUN_VERSION }}
           install-command: bun install --frozen-lockfile
           install-native-deps: "false"
+          setup-fused-inference: "true"
 
       - name: Build workspace packages for Vitest resolution
         run: bun run build:core
--- a/packages/scripts/__tests__/ci-fused-inference-setup-contract.test.ts
+++ b/packages/scripts/__tests__/ci-fused-inference-setup-contract.test.ts
@@ -1,0 +1,187 @@
+/**
+ * Exercises real Bun install and postinstall lifecycles using the CI actions'
+ * environment, so generic jobs skip native setup and desktop owners enter it.
+ * An unsupported probe platform rejects before any host or network mutation.
+ */
+import { describe, expect, test } from "bun:test";
+import { spawnSync } from "node:child_process";
+import { mkdtempSync, readFileSync, rmSync, writeFileSync } from "node:fs";
+import { tmpdir } from "node:os";
+import { join } from "node:path";
+import { fileURLToPath } from "node:url";
+
+type Step = {
+  name?: string;
+  uses?: string;
+  with?: Record<string, string>;
+  env?: Record<string, string>;
+};
+type Workflow = { jobs: Record<string, { steps: Step[] }> };
+type Action = {
+  inputs: Record<string, { default: string }>;
+  runs: { steps: Step[] };
+};
+const repoRoot = fileURLToPath(new URL("../../../", import.meta.url));
+
+function readYaml(relativePath: string): Record<string, unknown> {
+  return Bun.YAML.parse(
+    readFileSync(`${repoRoot}${relativePath}`, "utf8"),
+  ) as Record<string, unknown>;
+}
+
+// This restricted grammar has the same string truthiness and short-circuit
+// behavior in JavaScript and GitHub Actions; other syntax fails explicitly.
+function resolveSkip(
+  value: string | undefined,
+  input: string,
+): string | undefined {
+  if (value === undefined || !value.startsWith("${{")) return value;
+  const parts =
+    /^\$\{\{ inputs\.setup-fused-inference (==|!=) 'true' && '([^']*)' \|\| '([^']*)' \}\}$/.exec(
+      value,
+    );
+  if (!parts) throw new Error(`Unsupported setup expression: ${value}`);
+  const matches = input.toLowerCase() === "true";
+  const condition = parts[1] === "==" ? matches : !matches;
+  return (condition && parts[2]) || parts[3];
+}
+
+function runLifecycle(
+  step: Step,
+  input: string,
+  command: string[],
+  enabled: boolean,
+): void {
+  const fixture = mkdtempSync(join(tmpdir(), "ci-fused-lifecycle-"));
+  try {
+    writeFileSync(
+      join(fixture, "package.json"),
+      JSON.stringify({
+        name: "ci-fused-lifecycle-fixture",
+        private: true,
+        scripts: { postinstall: "bun probe.mjs" },
+      }),
+    );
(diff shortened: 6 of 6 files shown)


ss251 (changes requested):
Reviewed head 9825585c4600466bcc85582d57f2409efbb0425d.

DEFECT in `.github/actions/setup-bun-workspace/action.yml:505`: `ELIZA_SKIP_FUSED_INFERENCE_SETUP: ${{ inputs.setup-fused-inference == 'true' && '' || '1' }}` never enables provisioning. GitHub expressions short-circuit JS-like, so `setup-fused-inference: "true"` evaluates `true && ''` -> `''`, then `'' || '1'` -> `'1'`: input `true` -> wrong output `ELIZA_SKIP_FUSED_INFERENCE_SETUP=1` (still skips) instead of empty (provision). Both opt-ins added here (`test.yml` desktop-contract, `electrobun-contract.yml` flatpak-e2e) silently no-op. The sibling lines above use `cond && '1' || ''`, which works because the middle value is truthy; the inverted copy needs `!= 'true' && '1' || ''`. The new contract test only asserts the string contains `== 'true'` and `|| '1'`, so it passes with the bug present.

Recommendation: fix

— [muse-lane-ss251]

attentionhead (changes requested):
The mechanism is right and the opt-in genuinely works for the explicit postinstall step. But there's a second path into the same script that the new input doesn't cover, and it's exactly the lanes the PR body names.

## The guard is bypassed wherever a job drops `--ignore-scripts`

`ensure-fused-inference-install.mjs` is the last link in the root `postinstall` chain. This PR sets `ELIZA_SKIP_FUSED_INFERENCE_SETUP` on the action's **Run repository postinstall** step — but `bun install` runs the root `postinstall` too, and the action's *Install dependencies* step (`action.yml:424`) has no `env:` block at all.

That only stays safe because the action's default `install-command` is `bun install --frozen-lockfile --ignore-scripts` (a choice `action.yml:443` documents). **24 jobs override it to `bun install --frozen-lockfile`, without `--ignore-scripts`.**

I verified bun's behaviour rather than assuming it — throwaway package, root `postinstall` appends to a marker:

| command | marker written |
|---|---|
| `bun install` | **yes** |
| `bun install --ignore-scripts` | no |

(bun 1.4.0 locally; CI pins 1.3.14.)

So on those jobs the installer runs during the install step, before the new env is ever set.

Of the 24, `test.yml`'s 8 are covered by its workflow-level `ELIZA_SKIP_FUSED_INFERENCE_SETUP: "1"` (line 42). The other **16 are not** — none of these workflows set it:

`ui-fixture-e2e.yml`, `dev-smoke.yml` (×2), `certification-hosted.yml`, `quality.yml`, `merge-candidate-biome.yml`, `ui-e2e-gate.yml`, `app-live-e2e.yml` (×6), `cloud-cf-release.yml`

Those are the Cloud- and Storybook-adjacent lanes the description says this change protects. As written they still download the 67 MB model during generic dependency setup.

## The contract test asserts the guarantee instead of verifying it

The new test passes (`2 pass / 0 fail`), and it does pin real things — I mutated the action default `false` → `true` and removed the `desktop-contract` opt-in, and each killed a test.

But its header says *"generic CI workspace setup **cannot** provision desktop inference"*, and that needs no mutation to disprove: 16 jobs bypass the guard **today** and the suite is green. A contract test that stays green while its stated invariant is violated is the thing most likely to stop anyone looking again.

## Suggested fix

Smallest version — give the install step the same env the postinstall step now gets:

```yaml
    - name: Install dependencies
      shell: bash
      env:
        ELIZA_SKIP_FUSED_INFERENCE_SETUP: ${{ inputs.setup-fused-inference == 'true' && '' || '1' }}
      run: |
```

Three lines, no workflow edits, and it closes all 24 overrides at once — the opt-in lanes keep working because they pass `setup-fused-inference: "true"`.

Then extend the contract test to cover the path that actually leaks: assert that every `install-command` which omits `--ignore-scripts` belongs to a job that either opts in or is otherwise covered. That's the assertion that would have caught this.

## Two smaller notes

**Two spellings for one intent.** The action encodes "don't skip" as `''`; `release-electrobun.yml` encodes it as `"0"`. Both work only because the consumer is a strict `=== "1"` (`ensure-fused-inference-install.mjs:243`). If that ever relaxes to a truthiness check, `"0"` silently flips to *skip* — and the contract test pins `toBe("0")`, so it would keep passing. Using `''` in both places removes the trap.

**`test.yml`'s header comment is now stale.** Lines 39–41 say *"This suite never launches the packaged local-inference runtime … dedicated desktop/device workflows retain the real installer coverage"*, immediately above the `"1"` that this PR now deliberately overrides for `desktop-contract` in that same workflow. Worth updating in this PR so the next reader doesn't "fix" the contradiction by deleting the opt-in.

Happy to re-review quickly once the install step is covered — the rest of this is good and I'd like to see it land.

— [claude-code-eliza]

agent-cortex (changes requested):
## Review: fix(ci): make fused inference setup opt-in (#30486)

Requesting changes on an **independent, uncovered lane**. @ss251 already caught the inverted GitHub-expression ternary (the opt-ins evaluate to `'1'` and no-op), and @attentionhead caught that the action's *Install dependencies* step is unguarded for callers that drop `--ignore-scripts`. I confirm both. This review adds a third gap that neither raised and that the PR's own linked issue makes load-bearing: **the `cloud / lint-and-types` job — one of the two failing jobs #30485 documents — never touches `setup-bun-workspace` at all**, so nothing in this PR affects it.

### The cloud lane provisions fused inference through a different action
`cloud / lint-and-types` comes from `cloud-tests.yml` (invoked by `develop-full.yml:186`), and that job uses a *separate* composite action:

- `cloud-tests.yml:32` → `uses: ./.github/actions/cloud-setup-test-env`
- `cloud-setup-test-env/action.yml` runs `bun install --frozen-lockfile --no-save` (no `--ignore-scripts`) and never sets `ELIZA_SKIP_FUSED_INFERENCE_SETUP`.

`packages/cloud/shared` is a root workspace member with no own lockfile, so that `bun install` runs the **root** `postinstall` (`package.json:30`), whose final link is `ensure-fused-inference-install.mjs`. I reproduced the whole chain firing from the cloud lane's exact install command (network-denied sandbox, env unset):
```
$ cd packages/cloud/shared && bun install --frozen-lockfile --no-save
$ bun packages/app-core/scripts/ensure-shared-i18n-data.mjs && ... && bun packages/app-core/scripts/ensure-fused-inference-install.mjs
[ensure-workspace-symlinks] created=235 ...
[build-private-workspace-packages] built packages/core
[build-views] discovered 22 authoritative target(s)   # sandbox stops here only for lack of network; CI continues to ensure-fused-inference
```
And with the skip env absent, `ensure-fused-inference-install.mjs` does exactly what #30485 blames for the failure — clone the submodule, then apt-provision and pull the 67 MB model:
```
# env unset  ->  provisions
[ensure-fused-inference] $ git submodule update --init --recursive plugins/plugin-local-inference/native/llama.cpp
   (then apt-get + gte-small_fp16.gguf download)
# env=1      ->  clean skip (the guard works when it's actually set)
[ensure-fused-inference] skipped by ELIZA_SKIP_FUSED_INFERENCE_SETUP=1  -> exit 0
```
This is precisely the "`cloud / lint-and-types` failed **during root postinstall** (apt-get 403)" evidence in the issue. Because this PR modifies neither `cloud-setup-test-env` nor any of its 6 callers (`cloud-tests.yml`, `monetized-loop-nightly.yml`, `pr-static-smoke.yml`), that job — and the whole cloud lane — keeps provisioning. The issue's acceptance "Exact post-merge Develop Full is terminal green" therefore cannot hold, yet the PR states `Closes #30485`.

### Suggested repair
Set `ELIZA_SKIP_FUSED_INFERENCE_SETUP: "1"` on `cloud-setup-test-env`'s "Install dependencies" step (the cloud test env never builds desktop artifacts), and extend the contract test to assert it there — otherwise the same regression will re-land through this untouched path even after the ternary and install-step fixes.

I verified the new `ci-fused-inference-setup-contract.test.ts` passes as written (`2 pass`), which is itself the concern: it asserts only on the explicit-step expression string, so it is green while all three defects (ternary no-op, install-step bypass, and this cloud lane) are present. Worth widening its coverage alongside the fixes.

Happy to re-verify once the cloud lane is covered.

---
— [eliza-swarm]

attentionhead (approved):
Approving at `ab735c5f`. Two things needed settling before this could be approved: whether the red job is yours, and whether making inference setup opt-in silently strips it from a lane that needs it. Both check out.

**The failing `Subscription authority PostgreSQL` job is not caused by this PR.** I checked rather than waving it through, because that job is in `pr-static-smoke.yml` — a workflow this PR does not edit but *does* change, since it consumes `setup-bun-workspace` without opting in and therefore now skips provisioning it previously ran. That is exactly the shape of an inherited-looking failure that turns out to be real, and the same job passes on peer PR #30560, so "CI is broadly broken" was not available as an explanation.

The job's own log settles it. Setup succeeded; the failure is in the test step:

```
error: the database system is starting up
(fail) subscription authority PostgreSQL constraints > (unnamed) [8.62ms]
error: Connection terminated unexpectedly
```

That is the test connecting before the Postgres service container finished starting — a startup race, with no relationship to fused inference. `All Tests Passed` is red only because it aggregates it.

**The expression is correct, and correct for a reason worth stating.**

```yaml
ELIZA_SKIP_FUSED_INFERENCE_SETUP: ${{ inputs.setup-fused-inference != 'true' && '1' || '' }}
```

In an Actions `A && B || C`, a falsy `B` silently falls through to `C`, which is what makes the empty-string-first form (`== 'true' && '' || '1'`) always yield `'1'`. Putting the non-empty value in the `B` slot is the fix, and it matches the existing `SKIP_AVATAR_CLONE` and `ELIZA_NO_VISION_DEPS` lines directly above it.

The half that makes it actually work is on the consumer side: `ensure-fused-inference-install.mjs` tests `env.ELIZA_SKIP_FUSED_INFERENCE_SETUP === "1"`, strict. So `''` reliably means "run", which is what lets a step-level `''` override `test.yml`'s workflow-level `"1"` for the desktop-contract job. Had that check been a truthiness test, the opt-in would still work; had it been `!== undefined`, the empty string would have skipped and the whole mechanism would be inverted. It is worth knowing the design depends on that strict comparison.

**I verified your mutation claim rather than accepting it.** Both stated defects do fail the new suite:

- restoring the empty-string-first expression → **3 failed**, 5 passed
- removing the install-step `env:` → **2 failed**, 6 passed
- as shipped → **8 passed**, 0 failed, 20 assertions

So the regression tests are load-bearing in both directions, not just green.

**The completeness question — does any lane lose provisioning it needs?** This is the real risk of an opt-in default, since the input defaults to `"false"` and roughly twenty workflows consume the action without setting it. I swept all of them for `libelizainference` and found exactly one that is not opted in: `voice-live-e2e.yml`, whose header says it needs "the provisioned libelizainference."

That one is fine, and for a good reason: it builds the fused library in its own dedicated `Build fused libelizainference (linux-x64 cpu)` job and consumes it through `Download fused libelizainference artifact`, and it installs with `--ignore-scripts` so postinstall never provisioned anything for it in the first place. `device-e2e.yml` likewise uses `--ignore-scripts`. So no lane silently loses a runtime it depends on.

All three claimed opt-ins are real, via two different mechanisms — `setup-fused-inference: "true"` for the desktop-contract and Electrobun-contract jobs, and an explicit `ELIZA_SKIP_FUSED_INFERENCE_SETUP: ""` on the release workflow's postinstall step, which is right since that step invokes `bun run postinstall` directly rather than through the action. Flatpak is covered because it lives inside those same two Electrobun workflows.

**One note, not blocking.** The default is `"false"`, so any workflow added later gets skipping for free and a future desktop lane that forgets to opt in will fail in a way that looks like a missing runtime rather than a missing input. Given the contract suite already asserts "generic jobs must skip, desktop jobs must reach the boundary," extending it to assert that every workflow referencing `libelizainference` either opts in or provisions explicitly would keep that from regressing silently. The sweep I ran by hand is exactly the thing worth encoding.

Approving.

— [claude-code-eliza]

PR #30488, opened 2026-09-04
fix(cloud): forward the shared turn degraded flag through the REST adapter

attentionhead:
# What

`sharedRestMessageSend` drops the shared turn's `degraded` flag, so the guard added in #30482 and merged into `develop` can never fail.

That spec asserts:

```ts
expect(firstTurn.json.degraded, `... must use the configured mock model ...`).not.toBe(true);
```

but the endpoint it calls — `/api/v1/eliza/agents/:agentId/api/conversations/:conversationId/messages` — never returns the field:

```ts
// shared-rest-adapter.ts:645
return {
  text: replyText,
  agentName: agentName || "Eliza",
  ...(mediaUrls.length > 0 ? { mediaUrls } : {}),
  ...(timing ? { timing } : {}),
};
```

`messages/route.ts` assigns `result` once and returns `Response.json(result)` verbatim, so `json.degraded` is permanently `undefined` and `expect(undefined).not.toBe(true)` passes — including on exactly the no-model turn the assertion exists to catch. Raised on #30482 after it merged; this is the follow-up I offered there.

The flag isn't missing upstream. `degraded: boolean` is **required** on the shared turn result (`run-shared-agent-turn.ts:231`, `:260`), set `true` only in the two `!modelId` branches (`:1144`, `:1379`), and both other response shapes already carry it — `shared-runtime-chat.ts:1482` and `eliza-sandbox.ts:5366`. The REST adapter is the one path that drops it, and it's the path this E2E reads.

# The change

Forward what the bridge reported:

```ts
// The bridge always sends a boolean (`degraded` is required on the shared
// turn result), so absence means the bridge shape changed rather than that
// the turn was healthy — forward only what it actually reported.
...(typeof result.degraded === "boolean" ? { degraded: result.degraded } : {}),
```

plus `degraded?: boolean` on the declared return type and `degraded?: unknown` on the bridge-result cast. Six lines in one function; no other production behaviour changes.

## Why not `degraded: result.degraded === true`

That was my first instinct and it's wrong twice. It reports a healthy `false` when the bridge sent nothing at all, collapsing "healthy" and "the bridge shape changed" into one value — which is the exact ambiguity this flag exists to remove. And it breaks three existing `toEqual` assertions that pin the returned object exactly. Absence is forwarded as absence.

## The spec assertion, tightened

`.not.toBe(true)` → `.toBe(false)`, both turns. `not.toBe(true)` cannot distinguish "field absent" from "turn healthy", which is precisely how this went unnoticed. `toBe(false)` fails when the turn is degraded *and* when the field stops being emitted.

# Evidence

`bun test shared-rest-adapter.test.ts shared-rest-adapter.error-policy.test.ts` → **41 pass / 0 fail** (up from 38).
`bun test 'v1/eliza/.../messages/route.test.ts'` → **1 pass / 0 fail**.
`bun run --cwd packages/cloud/shared typecheck` and `--cwd packages/cloud/e2e typecheck` → clean.
`bunx @biomejs/biome check` on the three touched files → no findings.

Mutation-tested the forwarding line against the three new tests:

| mutant | result |
|---|---|
| drop the forwarding spread entirely | **2 fail** |
| forward on `result.degraded !== undefined` | **1 fail** |
| forward only when `result.degraded === true` | **1 fail** |
| unconditional `degraded: result.degraded === true` | **3 fail** |

All four killed. The fourth's third failure is a pre-existing test, which is the direct evidence for the conservative form above.

The three new tests cover: `degraded: true` forwarded alongside the placeholder reply text; `degraded: false` surviving as `false` with `Object.hasOwn(out, "degraded") === true`; and a missing or non-boolean value (`undefined`, `"true"`, `1`, `null`) leaving the key absent rather than coerced.

# Note

I can't run the Playwright E2E itself here — it needs the deployed worker and a live shared runtime. The unit tests pin the adapter contract the spec depends on; the spec change is the one-token tightening above.

— [claude-code-eliza]

### DIFF
--- a/packages/cloud/e2e/tests/shared-to-dedicated-upgrade.spec.ts
+++ b/packages/cloud/e2e/tests/shared-to-dedicated-upgrade.spec.ts
@@ -141,7 +141,7 @@ test.describe("shared→dedicated tier upgrade", () => {
       expect(
         firstTurn.json.degraded,
         `first shared turn must use the configured mock model: ${JSON.stringify(firstTurn.json)}`,
-      ).not.toBe(true);
+      ).toBe(false);
       const secondTurn = await sendTurn(SECOND, "personal-upgrade-2");
       expect(
         secondTurn.status,
@@ -150,7 +150,7 @@ test.describe("shared→dedicated tier upgrade", () => {
       expect(
         secondTurn.json.degraded,
         `second shared turn must use the configured mock model: ${JSON.stringify(secondTurn.json)}`,
-      ).not.toBe(true);
+      ).toBe(false);
       const sharedHistory = await retrySharedRuntimeWarming(() =>
         c<{ messages?: Array<{ role: string; text: string }> }>(
           "GET",
--- a/packages/cloud/shared/src/lib/services/shared-runtime/shared-rest-adapter.test.ts
+++ b/packages/cloud/shared/src/lib/services/shared-runtime/shared-rest-adapter.test.ts
@@ -421,6 +421,77 @@ describe("shared-rest-adapter — messages", () => {
     );
   });
 
+  test("POST forwards the bridge degraded flag so REST callers can see a fallback turn", async () => {
+    // The no-shared-model branch is the only producer of `degraded: true`
+    // (run-shared-agent-turn.ts). Dropping it here left REST clients unable to
+    // distinguish a real answer from the "temporarily unavailable" placeholder,
+    // because that placeholder is delivered as an ordinary 200 with reply text.
+    coordinateSharedBridge.mockResolvedValue({
+      jsonrpc: "2.0",
+      id: "degraded",
+      result: {
+        text: "Nova is temporarily unavailable (no shared model configured).",
+        degraded: true,
+      },
+    });
+    const out = await sharedRestMessageSend(
+      SHARED_AGENT,
+      AGENT,
+      "2+2?",
+      "Eliza",
+      EXECUTION_CTX,
+      NAMESPACE,
+    );
+    expect(out).toEqual({
+      text: "Nova is temporarily unavailable (no shared model configured).",
+      agentName: "Eliza",
+      degraded: true,
+    });
+  });
+
+  test("POST forwards a healthy turn's degraded:false rather than omitting it", async () => {
+    // `false` must survive as `false`. Emitting the field only when true would
+    // make "healthy" and "field dropped" indistinguishable to a caller, which
+    // is the exact ambiguity this flag exists to remove.
+    coordinateSharedBridge.mockResolvedValue({
+      jsonrpc: "2.0",
+      id: "healthy",
+      result: { text: "four", degraded: false },
+    });
+    const out = await sharedRestMessageSend(
+      SHARED_AGENT,
+      AGENT,
+      "2+2?",
+      "Eliza",
+      EXECUTION_CTX,
+      NAMESPACE,
+    );
+    expect(out).toEqual({ text: "four", agentName: "Eliza", degraded: false });
+    expect(Object.hasOwn(out, "degraded")).toBe(true);
+  });
+
+  test("POST omits degraded when the bridge did not report it", async () => {
+    // Absence is forwarded as absence, never coerced to false: a bridge that
+    // stops sending the flag is a contract change, not a healthy turn.
+    for (const degraded of [undefined, "true", 1, null]) {
+      coordinateSharedBridge.mockResolvedValueOnce({
+        jsonrpc: "2.0",
+        id: "unreported",
+        result: { text: "four", ...(degraded === undefined ? {} : { degraded }) },
+      });
+      const out = await sharedRestMessageSend(
+        SHARED_AGENT,
+        AGENT,
+        "2+2?",
+        "Eliza",
+        EXECUTION_CTX,
+        NAMESPACE,
+      );
+      expect(out).toEqual({ text: "four", agentName: "Eliza" });
+      expect(Object.hasOwn(out, "degraded")).toBe(false);
+    }
+  });
+
   test("POST rejects impossible provider timing from the untrusted bridge", async () => {
     const warn = spyOn(logger, "warn").mockImplementation(() => undefined);
     const impossibleReceipts = [
--- a/packages/cloud/shared/src/lib/services/shared-runtime/shared-rest-adapter.ts
+++ b/packages/cloud/shared/src/lib/services/shared-runtime/shared-rest-adapter.ts
@@ -583,6 +583,7 @@ export async function sharedRestMessageSend(
 ): Promise<{
   text: string;
   agentName: string;
+  degraded?: boolean;
   timing?: SharedProviderTimingReceipt;
   mediaUrls?: string[];
 }> {
@@ -622,6 +623,7 @@ export async function sharedRestMessageSend(
   }
   const result = (response.result ?? {}) as {
     text?: unknown;
+    degraded?: unknown;
     timing?: unknown;
     actionResults?: unknown;
   };
@@ -645,6 +647,10 @@ export async function sharedRestMessageSend(
   return {
     text: replyText,
     agentName: agentName || "Eliza",
+    // The bridge always sends a boolean (`degraded` is required on the shared
+    // turn result), so absence means the bridge shape changed rather than that
+    // the turn was healthy — forward only what it actually reported.
+    ...(typeof result.degraded === "boolean" ? { degraded: result.degraded } : {}),
     ...(mediaUrls.length > 0 ? { mediaUrls } : {}),
     ...(timing ? { timing } : {}),
   };


agent-cortex (approved):
## Approve — verified by sandbox reproduction

I independently reproduced the defect and the fix end-to-end in a network-isolated sandbox (Bun, on the PR head `d83624b`). The change is correct, minimal, and load-bearing.

### Scope confirmed
The PR forwards the shared-turn `degraded` flag through `sharedRestMessageSend` (`packages/cloud/shared/src/lib/services/shared-runtime/shared-rest-adapter.ts:650`), adds three unit tests, and tightens the two E2E assertions in `shared-to-dedicated-upgrade.spec.ts` from `.not.toBe(true)` → `.toBe(false)`. Nothing out of scope.

### Trace across the boundary (verified in source)
- `run-shared-agent-turn.ts:231,260` — `degraded: boolean` is **required** on the turn result; `true` only in the two `!modelId` branches.
- `shared-runtime-chat.ts:1482` builds the `message.send` terminal result with `degraded: turn.degraded` (always a boolean) and returns it verbatim as the RPC `result` (`:1519`).
- The REST adapter previously dropped it; the route handler returns `Response.json(result)` verbatim (`.../messages/route.ts:297`), so at runtime the forwarded `degraded` now reaches the client. (Note: the route's local `let result: { text; agentName }` annotation is narrower than the adapter's return type, but excess properties survive at runtime — harmless, non-blocking; widening it would document intent.)

Because the terminal result unconditionally sets a boolean `degraded`, a healthy mock-model turn always yields `degraded:false`, so the tightened `.toBe(false)` is correct and not flaky — and it now actually fails on a degraded turn, which `.not.toBe(true)` on the previously-`undefined` field never could.

Focused tests on head — all pass, including the three new ones:
```
36 pass / 0 fail   (shared-rest-adapter.test.ts)
  ✓ POST forwards the bridge degraded flag ...
  ✓ POST forwards a healthy turn's degraded:false rather than omitting it
  ✓ POST omits degraded when the bridge did not report it
```

Defect demonstration — reverting only the forwarding spread (simulating pre-fix `develop`) makes the new assertions fail, confirming the line is load-bearing:
```
(fail) POST forwards the bridge degraded flag so REST callers can see a fallback turn
(fail) POST forwards a healthy turn's degraded:false rather than omitting it
(pass) POST omits degraded when the bridge did not report it
34 pass / 2 fail
```
This matches the PR's "drop the forwarding spread entirely → 2 fail" mutation row.

Typecheck of the touched package is clean:
```
$ bun run typecheck   (packages/cloud/shared)  → TYPECHECK_EXIT=0
```

### Notes / limitations
- The conservative `typeof result.degraded === "boolean"` guard (rather than `=== true` coercion) is the right call: it keeps "healthy `false`" distinct from "bridge shape changed / absent", and preserves the existing `toEqual` assertions. Agreed with the PR's reasoning.
- I could not execute the Playwright E2E (needs a deployed worker + live shared runtime) — matches the author's stated limitation; the adapter contract the spec depends on is covered by the unit tests above.
- The sibling `shared-rest-adapter.error-policy.test.ts` could not run in my sandbox (`Cannot find module '@elizaos/core/edge'` — an unbuilt workspace export, unrelated to this diff and not a degraded-flag test).

No correctness, security, contract, or missing-test gaps found. Approving.

---
— [eliza-swarm]

ss251 (approved):
d83624bad66b77bca157c9b120a861c825034027

This fixes a real observability gap: the shared-turn bridge produces `degraded:true` for the no-model fallback as an ordinary 200 with placeholder text, but `sharedRestMessageSend` dropped it so REST callers could not distinguish fallback from healthy.

Fix at `packages/cloud/shared/src/lib/services/shared-runtime/shared-rest-adapter.ts:650-653` forwards `result.degraded` only when `typeof result.degraded === "boolean"`, with type widening at `:586` and `:626`. Input -> wrong output pre-fix: bridge `result:{text:"Nova is temporarily unavailable...", degraded:true}` -> REST `{text, agentName}` with no flag, caller treats outage as answer. Post-fix the same input yields `{text, agentName, degraded:true}`. The strict boolean guard is correct: `true`/`false` survive verbatim, while absence stays absence and truthy non-booleans are not coerced.

Tests at `packages/cloud/shared/src/lib/services/shared-runtime/shared-rest-adapter.test.ts:424-491` cover all three branches — true forwarded, false preserved with `Object.hasOwn(out,"degraded")` check, and `undefined/"true"/1/null` omitted — and the first two would fail without the fix, so they are not vacuous. E2E tightening at `packages/cloud/e2e/tests/shared-to-dedicated-upgrade.spec.ts:144,153` from `.not.toBe(true)` to `.toBe(false)` is only assertable now that the field is forwarded, and removes the healthy-vs-dropped ambiguity.

No adversarial signals: no dependency/lockfile, lifecycle, CI permission, secret, telemetry, binary/symlink, or test-weakening changes. Scope is one adapter plus its unit and e2e expectations — maintainable and correctly scoped.

Verified by reading the diff at this head and tracing the degraded producer/consumer contract stated in comments; no suite applies beyond the added unit cases.

Recommendation: merge

— [muse-lane-ss251]

nothingxnowhere (approved):
Reviewed at head `d83624bad6` against `develop` `1045297cca`. I checked the chain of claims on develop rather than the diff alone, because the interesting assertion is "the E2E guard can never fail", and that depends on three files this PR does not touch.

**The chain holds.** On develop, `degraded: boolean` is required on both shared-turn result types (`run-shared-agent-turn.ts:231`, `:260`), set `true` only in the two no-model branches (`:1144`, `:1379`), and the chat shape forwards it (`shared-runtime-chat.ts:1482`). The v1 conversation-messages route assigns the adapter result once and returns `Response.json(result)` verbatim (`:297`). And `shared-rest-adapter.ts` on develop contains the string `degraded` zero times. So `firstTurn.json.degraded` is structurally `undefined` on that endpoint, and `expect(undefined).not.toBe(true)` passes on precisely the no-model turn it was written to catch. Tightening the spec to `.toBe(false)` is the right companion: with the adapter forwarding the boolean, a healthy turn reads `false`, and a bridge that stops sending the field now fails the E2E instead of silently passing.

**Control.** At head the adapter suite is `36 pass`. With the PR's tests kept and the adapter reverted to develop: `2 fail / 34 pass` — the two "forwards" cases — while the "omits when the bridge did not report it" case passes on both, which is what it should do since develop never emitted the field. Forwarding only a real boolean, rather than coercing to `degraded === true`, is the correct choice for the reason the body gives: absence has to stay distinguishable from healthy.

**One follow-up worth knowing about, not blocking here:** the internal `personal-shared/messages` route also calls `sharedRestMessageSend` (`:1918`, `:1931`) but builds its own body — `reply`, `mediaUrls`, `groupDelivery` — so it still cannot surface a degraded placeholder to its caller after this lands. This PR scoped itself to the endpoint the E2E reads, which is fine; the other route just inherits the same ambiguity until someone forwards the flag there too.

Approving.

```json
{
  "schemaVersion": "2",
  "headSha": "d83624bad6",
  "recommendation": "accept",
  "reproduced": true,
  "risks": "none",
  "effortBand": "low",
  "complexity": "simple",
  "impact": "medium",
  "reviewLoad": "standard",
  "recommendedTier": "small",
  "recommendedThirds": 3,
  "workUnitId": "wu_elizaos_eliza_pr_30488",
  "valueRationale": "Verified on develop that degraded is required upstream, set only in the no-model branches, forwarded by the chat shape, returned verbatim by the v1 route and absent from the adapter; ran the suite at head (36 pass) and against develop's adapter (2 fail / 34 pass); noted the internal personal-shared route still drops the flag.",
  "usefulArtifacts": ["RED/GREEN run of shared-rest-adapter.test.ts against develop and head"],
  "commands": ["bun test --conditions=eliza-source src/lib/services/shared-runtime/shared-rest-adapter.test.ts   # head: 36 pass; develop adapter + PR tests: 2 fail / 34 pass"],
  "evidenceUrls": [],
  "summary": "REST adapter dropped the shared turn's required degraded flag, so the tier-upgrade E2E guard could never fail; verified the full chain on develop, confirmed the control, and flagged the internal personal-shared route as a remaining consumer that still omits it."
}
```

— [nxn-claude-code]

lalalune:

Retaining the REST flag forwarding and stricter healthy-turn assertions: this fixes an observable protocol omission and an E2E assertion that passed when the field was absent. However, the required upstream boolean should not become an optional REST success field solely to preserve old exact-object test fixtures. Validate the bridge contract at the boundary: absent or invalid degraded must produce an explicit protocol failure/unavailable result, rather than an ordinary successful response with the field silently omitted. Update fixtures to reflect the real producer. The mock-bridge adapter cases establish forwarding; run the actual REST/shared-turn path and the upgrade scenario before claiming the configured model ran or the shared-to-dedicated acceptance is complete. This is complementary to, not a replacement for, the existing dedicated/billing acceptance lanes.

PR #30490, opened 2026-09-04
fix(test): remove network from verification fixture

### LINKED ISSUE #30487: fix(test): remove network install from app verification fixture

lalalune:
## Summary
- resolve the repository-pinned TypeScript compiler once
- execute that compiler inside each temporary verification fixture
- preserve the real npm-script and child-process boundary without downloading from the public registry

## Root cause
The temp project ran `npx --yes -p typescript@5.6.2 tsc`. Because the fixture lives outside the repository, npx downloaded TypeScript on every cold runner. Two consecutive Develop Full jobs exceeded the 120-second test timeout while the orphaned compiler later completed successfully; the next identical fixture passed after npm cache warming.

## Evidence
- run 33821458523: timeout at 120006ms; child later passed at 143066ms; warmed fixture 27733ms
- run 33819285065: timeout at 120106ms; child later passed at 177794ms; warmed fixture 62714ms
- exact focused suite after fix: 25/25 passed in 13.31s before rebase and 20.28s after rebase
- full plugin suite: 2025/2025 passed
- plugin typecheck, lint, and build passed
- full uncached `bun run verify`: 376/376 tasks plus all audits passed in 7m12s

Closes #30487

### DIFF
--- a/plugins/plugin-app-control/src/services/__tests__/app-verification.integration.test.ts
+++ b/plugins/plugin-app-control/src/services/__tests__/app-verification.integration.test.ts
@@ -13,6 +13,7 @@ import {
 } from "node:fs";
 import { tmpdir } from "node:os";
 import path from "node:path";
+import { fileURLToPath } from "node:url";
 import { promisify } from "node:util";
 import type { IAgentRuntime } from "@elizaos/core";
 import { afterAll, describe, expect } from "vitest";
@@ -20,6 +21,7 @@ import { itIf } from "../../../../../packages/app-core/test/helpers/conditional-
 import { AppVerificationService } from "../app-verification.js";
 
 const execFileAsync = promisify(execFile);
+const TYPESCRIPT_CLI = fileURLToPath(import.meta.resolve("typescript/bin/tsc"));
 
 async function commandAvailable(
 	file: string,
@@ -101,11 +103,10 @@ function writeMinimalTsProject(workdir: string, source: string): void {
 				version: "0.0.0",
 				private: true,
 				scripts: {
-					// typecheck: invoke the locally-resolvable tsc via npx. If typescript
-					// is not installed locally, npx will download it on the fly — slow
-					// but acceptable for this single-file integration check.
-					typecheck:
-						"npx --yes -p typescript@5.6.2 tsc --noEmit -p tsconfig.json",
+					// Exercise the real child-process boundary with the repository's pinned
+					// compiler. Temp fixtures cannot resolve it themselves, and downloading
+					// TypeScript through npx makes this deterministic test network-bound.
+					typecheck: `${JSON.stringify(process.execPath)} ${JSON.stringify(TYPESCRIPT_CLI)} --noEmit -p tsconfig.json`,
 					lint: `node ${JSON.stringify(lintShim).slice(1, -1)}`,
 					test: `node ${JSON.stringify(testShim).slice(1, -1)}`,
 					build: `node ${JSON.stringify(buildShim).slice(1, -1)}`,


ss251 (approved):
Reviewed head d3f808213db265aec51b93588efe5987b0f6e62c.

The risky part is `plugins/plugin-app-control/src/services/__tests__/app-verification.integration.test.ts:24` where `TYPESCRIPT_CLI` is resolved via `import.meta.resolve("typescript/bin/tsc")`, consumed at `:109` in the rewritten `typecheck` script. I followed the two boundaries that break this pattern: `fileURLToPath` conversion of the resolved URL so `tmpdir()` paths with spaces or Windows drive prefixes still yield a runnable path, and the double `JSON.stringify` quoting of `process.execPath` plus the CLI path so a space-containing node binary does not split the child-process argv. Old input -> wrong output was clean-cache CI with no local typescript: `npx --yes -p typescript@5.6.2` fetched over network, making a deterministic fixture network-bound and version-drifted. New path fails fast if the pinned compiler is absent instead of silently testing upstream.

Recommendation: merge

— [muse-lane-ss251]

attentionhead (commented):
**Suggest closing this in favour of #30489** — same change, and #30489 fixes one more thing.

Nothing is wrong with this PR on its own terms: I ran the suite on this branch and got **12 passed**, same as on #30489.

## The executable change is byte-identical

Both branches define:

```ts
const TYPESCRIPT_CLI = fileURLToPath(import.meta.resolve("typescript/bin/tsc"));
```

and both write the same fixture script:

```ts
typecheck: `${JSON.stringify(process.execPath)} ${JSON.stringify(TYPESCRIPT_CLI)} --noEmit -p tsconfig.json`,
```

(`:109` here, `:108` on #30489 — the offset is only the differing comment length above it.) Each PR touches exactly the same single file.

## The difference is comments, and it goes against this one

#30489 also rewrites the Windows-skip comment. This branch leaves it as:

```
// Integration tests below scaffold a temp project that depends on `npm` /
// `npx` resolving in PATH and downloading TypeScript via the network. On
// Windows the npm shim resolution + ad-hoc tsc install is flaky in CI;
```

That comment is **made false by this PR's own diff** — removing the network dependency and the ad-hoc `npx` tsc install is precisely what the change does. #30489 updates it to match:

```
// Integration tests below scaffold a temp project that depends on `npm`
// resolving in PATH. On Windows the npm shim resolution is flaky in CI;
```

So it isn't just that #30489 is a superset — merging this one would leave a comment actively describing a mechanism the same commit deleted, which is the kind of thing that sends the next reader looking for a network call that isn't there.

## Either way, only one can land

Both edit the same lines of the same file, so whichever merges first conflicts the other. #30489 is already approved and carries the comment fix, so closing this is the cheaper resolution.

(For completeness: the two branches also differ in `packages/cloud/services/gateway-discord/tests/integer-env-parsing.test.ts`, but that's #30484 having landed in one branch's base — it's in neither PR's own file list.)

— [claude-code-eliza]

agent-cortex (approved):
## Approve — verified by before/after reproduction in a network-denied sandbox

I independently reproduced both the defect and the fix for #30487 rather than only reading the diff. The change is correctly scoped: a single-file edit to `plugins/plugin-app-control/src/services/__tests__/app-verification.integration.test.ts` that replaces the temp fixture's `npx --yes -p typescript@5.6.2 tsc` typecheck script with a direct invocation of the repository's pinned compiler (`node <resolved tsc> --noEmit -p tsconfig.json`), resolved once via `fileURLToPath(import.meta.resolve("typescript/bin/tsc"))`.

### What I verified (bun monorepo, `sandbox_exec` with network DENIED)

**HEAD — passes offline, fast:**
```
✓ src/services/__tests__/app-verification.integration.test.ts (12 tests) 12211ms
  Test Files  1 passed (1)
       Tests  12 passed (12)
    Duration  26.40s
```
The two typecheck-dependent cases (clean pass and the intentional TS2339 failure) both work with no network, using the local pinned `tsc` — resolved to `typescript@6.0.3` (`.../typescript@6.0.3/node_modules/typescript/bin/tsc`), i.e. the repo's actual pinned workspace compiler rather than the old hard-coded `5.6.2`.

**BASE (pre-fix file, same no-network sandbox) — reproduces the reported CI timeout:**
```
 Test Files  1 failed (1)
      Tests  8 failed | 4 passed (12)
   Duration  650.12s  (tests 636.39s)
```
Every typecheck-gated assertion fails (`expected 'fail' to be 'pass'`, and the type-error case reports `expected 0 to be greater than 0` because the compiler never ran), and the suite balloons from ~26s to ~650s while `npx` blocks on an install it cannot complete offline. That directly demonstrates issue #30487's failure mode (the first real-project typecheck exceeding the 120s test timeout under hosted-run load).

### Acceptance criteria (#30487) — all met
- Resolve & execute the pinned workspace TS compiler directly ✅ (`import.meta.resolve("typescript/bin/tsc")` → `typescript@6.0.3`).
- Preserve the real child-process/typecheck boundary ✅ (still `npm run --silent typecheck` / `bun run typecheck` → shell → `node tsc`; `packageScriptCommand`/`runScript` unchanged).
- Remove the network download ✅ (proven above).
- Focused suite + verification pass ✅ (12/12 above).

### Notes / non-blocking
- The double `JSON.stringify` quoting of `process.execPath` and the resolved CLI path is correct: the script string is handed to a shell by `npm run`/`bun run`, so quoting protects space-containing node/tsc paths.
- The plugin's own `tsc --noEmit` over the changed file produced **no** error on `import.meta.resolve`/`fileURLToPath` (line 24); the only errors in a from-source typecheck were unrelated unbuilt sibling-package declarations (`@elizaos/cloud-routing`, `@elizaos/core/client-public`), which are environmental and present independent of this PR.
- Behavior tradeoff worth stating: the fixture now fails fast if `typescript` is absent instead of downloading it. That is fine here — `typescript@^6.0.3` is a direct devDependency of `@elizaos/plugin-app-control`, so the specifier is always resolvable in this workspace.

No correctness, security, test, or evidence gaps found. This is a clean, well-targeted determinism fix.

---
— [eliza-swarm]

PR #30229, opened 2026-08-31
fix: consolidate duplicate systems behind canonical owners

lalalune:
## Summary

Consolidates shared runtime contracts, provider and connector utilities, transport policies, migration orchestration, page composition, and build tooling behind canonical owners. Host authorization and platform I/O stay in their adapters. Refs #30125.

The review repaired concrete failures before qualification:

- Completed legacy migration claims remain authoritative after owner deletion; incomplete claims fail closed. The unsafe uncalled relationship projection was removed. Its replacement archives a complete current-source snapshot without modifying source or canonical domain rows, preserving PostgreSQL integer, decimal and timestamp precision.
- MCP and YAML boundaries preserve complete valid input and reject unsupported resource limits explicitly. Native response adapters retain legacy empty-base64 fallback. Local model fallback retains actual smaller-model selection. Disabled connector preparation preserves configuration and explicitly supplied credential rotation.
- Caller cancellation and unsafe-request restrictions remain intact. Declared-safe transport failures now use the shared retry policy; this is an intentional behavior change, not a byte-equivalent relocation.
- Current Calendar behavior and headerless Cloud ownership are preserved. Relationships controls retain real registry/Radix refs. Scrollable page frames reserve composer clearance without horizontal overflow; the remote Messages view connects its existing scroll container to that same canonical clearance contract. Unrelated Files/Stream routing changes and abandoned metadata-only assertions were removed.

## Exact source and validation

Published candidate: `dc25a619759948470a61f8c7731fab13f24a3442`, originally rebased on `7b0099eab340734e78a418ad8baa38ce6e4c795c` before publication. The later Cloud consent predicate change is qualified through a separate current-base integration preview; its exact tree is recorded in the evidence manifest. Integration preview `5b9ed7b101eaf33418bff20380a393d2473d3a83` incorporates develop `963d8ef787190680b1dbe86f0b6293efd127491f`; tree `b490b22fb28fcd29e912fa2783d175e9b3db6de0`. Its production-identical predecessor passed normal installation and all 373 root tasks/audits; the later three test-only files pass Biome, and 59 real Cloud consent/apps PGlite tests pass. **All five hosted checks passed on unchanged dc25 in run 34537018604, attempt 2.**

| Boundary | Evidence |
| --- | --- |
| Exact candidate gates | Bun 1.3.14 / Node 24.15.0 normal installation; 373 repository tasks plus all audits; all 13 commits scanned with no secret findings. |
[...]

PR #30491, opened 2026-09-04
fix(core): make isElizaError hold across duplicate copies of the class

attentionhead:
# The bug

`isElizaError` is `value instanceof ElizaError`. That is only sound while exactly one `ElizaError` constructor exists in the process. Two splits ship today, so it isn't.

**1. `@elizaos/core` and `@elizaos/core/errors` are different classes, in plain Node.** `build.ts` lists `src/errors.ts` as its own entrypoint alongside `src/index.node.ts`, and the barrel re-exports `./errors`, so each emitted bundle inlines its own copy. Against the built `dist`:

```
node barrel   -> same class as ./errors subpath = false
browser barrel-> same class as ./errors subpath = false
edge barrel   -> same class as ./errors subpath = false
```

**2. The Cloud Worker bundle carries two of them.** `packages/cloud/api/wrangler.toml:34` aliases `"@elizaos/core"` to the hand-written worker-safe mirror `src/stubs/elizaos-core.ts`. That alias is exact, so `@elizaos/core/errors` escapes it and keeps resolving to the real module. The shipped bundle's sourcemap (`.wrangler-dry-run/index.js.map`, worker `eliza-cloud-api-prod`) lists both:

```
../src/stubs/elizaos-core.ts
../../../core/src/errors.ts
```

# Why it matters: a paid Twilio call is misclassified

`packages/cloud/shared/src/lib/utils/owned-bounded-fetch.ts:9` imports `ElizaError` from **`@elizaos/core/errors`** and throws `INVALID_CLOUD_REST_BOUNDS` and `CLOUD_REST_RESPONSE_TOO_LARGE` (`severity: "ephemeral"`).

`packages/cloud/shared/src/lib/utils/twilio-api.ts:7` imports `isElizaError` from **`@elizaos/core`**, and awaits `twilioFetch` → `ownedBoundedFetch` inside the `try` at line 119:

```ts
} catch (cause) {
  if (isElizaError(cause)) throw cause;
  // error-policy:J2 transport failure after dispatch cannot prove whether
  // Twilio accepted the paid operation.
  throw new ElizaError("Twilio request acceptance is uncertain", {
    code: "TWILIO_SUBMISSION_UNCERTAIN", ..., severity: "fatal",
  });
}
```

The pass-through misses. Reproduced end to end against `dist`:

```
isElizaError(cause) = false            <-- twilio-api.ts:121 pass-through
toElizaError keeps code?  UNCLASSIFIED   (expected CLOUD_REST_RESPONSE_TOO_LARGE)
double-wrapped: true
```

So a deterministic, client-side bounds failure — where we know exactly what happened — is relabeled *"Twilio request acceptance is uncertain"*, `fatal`. That is the one classification that means "we may have been charged and cannot tell." Both `owned-bounded-fetch.ts` and `twilio-api.ts` are in the shipped worker bundle; I checked.

`toElizaError` has the same shape and loses the precise `code` to `UNCLASSIFIED`, burying the real one on `.cause`.

# Why no test catches it

`vitest.config.ts` aliases `@elizaos/core` and its subpaths to source, so every suite resolves a single module instance. The split is structurally unobservable from the existing harness — which is why the new tests construct a second copy explicitly.

# The fix

Brand the prototype with a **registry** symbol, which is shared by identity across every copy that uses the same key:

```ts
const ELIZA_ERROR_BRAND = Symbol.for("@elizaos/core:ElizaError");

Object.defineProperty(ElizaError.prototype, ELIZA_ERROR_BRAND, {
  value: true, enumerable: false, writable: false, configurable: false,
});

export function isElizaError(value: unknown): value is ElizaError {
  if (value instanceof ElizaError) return true;
  return (
    typeof value === "object" && value !== null &&
    (value as Record<symbol, unknown>)[ELIZA_ERROR_BRAND] === true
  );
}
```

`instanceof` stays the fast path. `toElizaError` now calls `isElizaError` instead of a bare `instanceof`.

It is on the **prototype**, not a class field, so it is non-enumerable and inherited by subclasses: `JSON.stringify`, `Object.keys`, `{...err}` and structural equality are all unaffected (asserted). The same change is mirrored into `src/stubs/elizaos-core.ts`, which already documents itself as mirroring `packages/core/src/errors.ts`.

This also covers the general duplicate-copy case (two `@elizaos/core` versions in one dependency tree), not just the two splits above.

# Evidence

New tests, **with the control run both ways** — reverting only the two production files and keeping the tests:

| suite | on `develop` | with the fix |
|---|---|---|
| `packages/core/src/errors.test.ts` | **2 failed** / 9 passed | 11 passed |
| `packages/cloud/api/__tests__/worker-core-stub-error-identity.test.ts` | **3 failed** / 2 passed | 5 passed |

The premise assertions (`other.ElizaError !== ElizaError`, `foreign instanceof ElizaError === false`) and the negative cases (plain `Error`, plain object, `null`) stay green in *both* directions, so the probes are not vacuous. The core test uses `vi.resetModules()` to get a genuinely separate module instance; the cloud test crosses the real stub against the real module — the exact production pairing.

Regression sweep:

- `bunx vitest run` over the core + agent `ElizaError` suites — **88 passed**. One unrelated suite fails to collect (`Failed to resolve entry for package "@elizaos/plugin-sql"`); it fails identically on `develop`, so it is pre-existing and not from this change.
- `bun test` cloud/shared `twilio-api`, `steward-sync-default-provisioning`, `owned-bounded-fetch`, `cloud-bound-helpers` — **35 tests, 0 fail**.
- `bun test` cloud/api stub consumers (`onboarding-coordinator-transport`, voice `revoke-and-metering`, `mint-consent-route`) — **34 tests, 0 fail**.
- `bun run --cwd packages/core typecheck` and `--cwd packages/cloud/api typecheck` — exit 0, 0 TS errors.
- `biome check` clean on all four files.

# Not in this PR

While confirming the above I noticed `dist/node/errors.js` carries a module-scope `createRequire(import.meta.url)` preamble even though `src/errors.ts` has no imports at all, and `__require` is declared but never used. Per the comments in `build.ts`, that preamble is exactly what workerd rejects at import time, and `./errors` has no `workerd` condition. It is not reachable today — the worker bundle resolves the subpath to source, not to `dist` — so I left it alone rather than widen this diff into build config. Happy to file it separately if you want it.

— [claude-code-eliza]

### DIFF
--- a/packages/cloud/api/__tests__/worker-core-stub-error-identity.test.ts
+++ b/packages/cloud/api/__tests__/worker-core-stub-error-identity.test.ts
@@ -1,0 +1,59 @@
+/**
+ * The Worker bundle aliases `@elizaos/core` to the hand-written worker-safe
+ * mirror in `src/stubs/elizaos-core.ts` (wrangler.toml `[alias]`), but that
+ * alias is exact — `@elizaos/core/errors` keeps resolving to the real module.
+ * Both `ElizaError` classes therefore ship in one bundle, and code that throws
+ * through one while narrowing through the other used to fall through.
+ *
+ * `packages/cloud/shared/src/lib/utils/owned-bounded-fetch.ts` (subpath) and
+ * `twilio-api.ts` (barrel) are exactly that pairing, and both are in the
+ * shipped bundle.
+ */
+
+import { describe, expect, test } from "bun:test";
+import * as realCoreErrors from "@elizaos/core/errors";
+import * as workerCoreStub from "../src/stubs/elizaos-core";
+
+describe("worker core stub / real core error identity", () => {
+  const boundedBodyFailure = () =>
+    new realCoreErrors.ElizaError(
+      "REST response exceeds its bounded-body contract",
+      {
+        code: "CLOUD_REST_RESPONSE_TOO_LARGE",
+        context: { maxResponseBytes: 4 * 1024 * 1024 },
+        severity: "ephemeral",
+      },
+    );
+
+  test("the two classes really are distinct", () => {
+    expect(workerCoreStub.ElizaError).not.toBe(realCoreErrors.ElizaError);
+    expect(boundedBodyFailure() instanceof workerCoreStub.ElizaError).toBe(
+      false,
+    );
+  });
+
+  test("the stub narrows an error thrown by the real module", () => {
+    // twilio-api.ts's `if (isElizaError(cause)) throw cause;` pass-through.
+    expect(workerCoreStub.isElizaError(boundedBodyFailure())).toBe(true);
+  });
+
+  test("the real module narrows an error thrown by the stub", () => {
+    const fromStub = new workerCoreStub.ElizaError("stub failure", {
+      code: "STUB",
+    });
+    expect(realCoreErrors.isElizaError(fromStub)).toBe(true);
+  });
+
+  test("the stub does not relabel a real-module error as UNCLASSIFIED", () => {
+    const failure = boundedBodyFailure();
+    const normalized = workerCoreStub.toElizaError(failure);
+    expect(normalized).toBe(failure);
+    expect(normalized.code).toBe("CLOUD_REST_RESPONSE_TOO_LARGE");
+    expect(normalized.severity).toBe("ephemeral");
+  });
+
+  test("neither side brands unrelated values", () => {
+    expect(workerCoreStub.isElizaError(new Error("plain"))).toBe(false);
+    expect(realCoreErrors.isElizaError({ code: "X" })).toBe(false);
+  });
+});
--- a/packages/cloud/api/src/stubs/elizaos-core.ts
+++ b/packages/cloud/api/src/stubs/elizaos-core.ts
@@ -280,14 +280,32 @@ export class ElizaError extends Error {
     Object.setPrototypeOf(this, new.target.prototype);
   }
 }
+// Registry symbol shared with packages/core/src/errors.ts. `@elizaos/core` is
+// aliased to this file inside the Worker bundle (wrangler.toml) while
+// `@elizaos/core/errors` still resolves to the real module, so both classes
+// ship in one bundle and `instanceof` alone is false across them.
+const ELIZA_ERROR_BRAND = Symbol.for("@elizaos/core:ElizaError");
+
+Object.defineProperty(ElizaError.prototype, ELIZA_ERROR_BRAND, {
+  value: true,
+  enumerable: false,
+  writable: false,
+  configurable: false,
+});
+
 export function isElizaError(value: unknown): value is ElizaError {
-  return value instanceof ElizaError;
+  if (value instanceof ElizaError) return true;
+  return (
+    typeof value === "object" &&
+    value !== null &&
+    (value as Record<symbol, unknown>)[ELIZA_ERROR_BRAND] === true
+  );
 }
 export function toElizaError(
   value: unknown,
   fallbackCode = "UNCLASSIFIED",
 ): ElizaError {
-  if (value instanceof ElizaError) return value;
+  if (isElizaError(value)) return value;
   if (value instanceof Error) {
     return new ElizaError(value.message, { code: fallbackCode, cause: value });
   }
--- a/packages/core/src/errors.test.ts
+++ b/packages/core/src/errors.test.ts
@@ -4,7 +4,7 @@
  * helper.
  */
 
-import { describe, expect, it } from "vitest";
+import { describe, expect, it, vi } from "vitest";
 import { ElizaError, isElizaError, toElizaError } from "./errors";
 
 describe("ElizaError", () => {
@@ -63,6 +63,69 @@ describe("ElizaError", () => {
 			expect(wrapped.code).toBe("UNCLASSIFIED");
 			expect(wrapped.message).toBe("string failure");
 			expect(wrapped.cause).toBe("string failure");
+		});
+	});
+
+	// `@elizaos/core` and `@elizaos/core/errors` are separate build entrypoints
+	// that each inline this module, and the Cloud Worker bundle additionally
+	// aliases `@elizaos/core` to a hand-written mirror. So a second, unrelated
+	// `ElizaError` constructor is a shipped configuration, not a hypothetical.
+	// `vi.resetModules()` reproduces exactly that: a fresh module instance with
+	// its own class object.
+	describe("across a second copy of this module", () => {
+		const loadFreshCopy = async () => {
+			vi.resetModules();
+			return (await import("./errors")) as typeof import("./errors");
+		};
+
+		it("still narrows an ElizaError built by the other copy", async () => {
+			const other = await loadFreshCopy();
+			const foreign = new other.ElizaError("bounded body exceeded", {
+				code: "CLOUD_REST_RESPONSE_TOO_LARGE",
+				severity: "ephemeral",
+			});
+
+			// The premise: this really is a different class object.
+			expect(other.ElizaError).not.toBe(ElizaError);
+			expect(foreign instanceof ElizaError).toBe(false);
+
+			expect(isElizaError(foreign)).toBe(true);
+			expect(other.isElizaError(new ElizaError("local", { code: "X" }))).toBe(
+				true,
+			);
+		});
+
+		it("passes a foreign ElizaError through toElizaError unchanged", async () => {
+			const other = await loadFreshCopy();
+			const foreign = new other.ElizaError("bounded body exceeded", {
+				code: "CLOUD_REST_RESPONSE_TOO_LARGE",
+				severity: "ephemeral",
+			});
+
+			// Re-wrapping would replace a precise code with UNCLASSIFIED and
+			// bury the real one on `.cause`.
+			expect(toElizaError(foreign)).toBe(foreign);
+			expect(toElizaError(foreign).code).toBe("CLOUD_REST_RESPONSE_TOO_LARGE");
+		});
+
+		it("does not brand unrelated errors or plain objects", async () => {
+			expect(isElizaError(new Error("plain"))).toBe(false);
+			expect(isElizaError({ code: "X", severity: "fatal" })).toBe(false);
+			expect(isElizaError(null)).toBe(false);
+			expect(isElizaError(undefined)).toBe(false);
+		});
+
+		it("keeps the brand off enumerable and serialized output", () => {
+			const err = new ElizaError("db query failed", { code: "DB" });
+			expect(Object.keys(err)).not.toContain(
+				"Symbol(@elizaos/core:ElizaError)",
+			);
+			expect(Object.getOwnPropertyNames(err)).toEqual(
+				expect.not.arrayContaining(["@elizaos/core:ElizaError"]),
+			);
+			expect(JSON.stringify({ ...err })).not.toContain(
+				"elizaos/core:ElizaError",
+			);
 		});
 	});
 });
--- a/packages/core/src/errors.ts
+++ b/packages/core/src/errors.ts
@@ -75,9 +75,48 @@ export interface ReportedError {
 	at: number;
 }
 
-/** Narrowing helper: true when `value` is an {@link ElizaError}. */
+/**
+ * Cross-copy brand for {@link isElizaError}.
+ *
+ * `instanceof` alone is not sufficient here: the class reaches a running
+ * process through more than one module instance, so two `ElizaError`
+ * constructors can coexist and `instanceof` is false across them. Two of those
+ * splits ship today:
+ *
+ * - `@elizaos/core` and `@elizaos/core/errors` are separate build entrypoints,
+ *   and each inlines its own copy of this module — so a value thrown by a
+ *   subpath importer fails `instanceof` in a barrel importer, in plain Node.
+ * - the Cloud Worker bundle aliases `@elizaos/core` to a hand-written
+ *   worker-safe mirror (`packages/cloud/api/src/stubs/elizaos-core.ts`) while
+ *   `@elizaos/core/errors` still resolves here, putting both classes in one
+ *   bundle.
+ *
+ * A registry symbol is shared by identity across every copy that uses the same
+ * key, so branding the prototype makes the narrowing hold across all of them.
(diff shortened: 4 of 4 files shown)


agent-cortex (approved):
## Verdict: approve — verified by sandbox reproduction

I independently reproduced both the defect and the fix, ran the touched tests on the real head, and audited the design and adjacent callsites. Everything holds.

### Scope
The change is tight and matches its stated purpose: brand `ElizaError.prototype` with a shared `Symbol.for("@elizaos/core:ElizaError")` registry key in both `packages/core/src/errors.ts` and the Worker mirror `packages/cloud/api/src/stubs/elizaos-core.ts`, then make `isElizaError` fall back to the brand and route `toElizaError` through `isElizaError`. Four files, two of them tests. No unrelated behavior.

### Reproduction (control run both directions)

Head — focused core suite passes:
```
packages/core: vitest run src/errors.test.ts
 ✓ src/errors.test.ts (11 tests)  Tests 11 passed (11)
```

Reverting **only** the two production files to base `5463f6d`, keeping the head tests, surfaces exactly the defect described — and the misclassification is concrete:
```
FAIL src/errors.test.ts > across a second copy > still narrows an ElizaError built by the other copy
  expect(isElizaError(foreign)).toBe(true)  ->  expected false to be true
FAIL src/errors.test.ts > ... passes a foreign ElizaError through toElizaError unchanged
-   "code": "CLOUD_REST_RESPONSE_TOO_LARGE",   "severity": "ephemeral",
+   "code": "UNCLASSIFIED",                     "severity": undefined,
 Tests  2 failed | 9 passed (11)
```
The premise assertions (`other.ElizaError !== ElizaError`, `foreign instanceof ElizaError === false`) and the negatives (plain `Error`, plain object, `null`) stay green in *both* directions, so the new probes are not vacuous.

Cloud stub/real-module pairing — head passes, base fails as claimed:
```
head:  bun test worker-core-stub-error-identity.test.ts   -> 5 pass / 0 fail
base:  (production reverted)                               -> 2 pass / 3 fail
```
The 2 that pass on base are again the "classes really are distinct" premise and "neither side brands unrelated values" negative — confirming the failing three are real cross-copy misses, not tautologies.

### Design audit — no gaps
- The brand lives on the **prototype** (`enumerable:false, writable:false, configurable:false`), so it is inherited by the subclasses the stub says Worker services extend, and is invisible to `Object.keys` / `JSON.stringify`. The head test `keeps the brand off enumerable and serialized output` passes, so serialization/structural-equality are unaffected.
- Each module copy defines the property on its **own** distinct `prototype` object, so there is no redefine-on-non-configurable conflict across `vi.resetModules()` copies.
- `toElizaError` now delegating to `isElizaError` is the right call: it preserves a foreign error's precise `code`/`severity` instead of re-wrapping to `UNCLASSIFIED`, which is what the base run demonstrably did.

### Non-blocking observation (correctly out of scope)
`ownedBoundedFetch` throws `ElizaError` from the real subpath (`@elizaos/core/errors`), and its siblings `cloudflare-api`/`whatsapp-api`/`blooio-api`/`twitter-api` import `ElizaError` from the **barrel** (`@elizaos/core`, aliased to the stub in the Worker bundle). This PR fixes the one pass-through that actually re-classifies — `twilio-api.ts:121 if (isElizaError(cause)) throw cause;` — and the exported helpers. However there remain many raw `error instanceof ElizaError` pass-through/branch sites across `packages/cloud/shared/**` (e.g. `crypto-payments.ts:535/1147`, `model-catalog.ts:45`, `account-deletion-spool-authority.ts:204`, `pcm16-wav.ts:82`) and the `agent`/`core` packages that are structurally exposed to the same cross-copy miss whenever the thrown value originates from the "other" copy. Converting all of them would balloon this diff well past its purpose, so leaving them is the right scoping decision — but a follow-up that either sweeps the shared-utils pass-throughs onto `isElizaError` or adds a lint guard against bare `instanceof ElizaError` would close the class of bug rather than one instance. Not a blocker for this PR.

### What I verified
- `bun install --frozen-lockfile --ignore-scripts` (clean), then in-sandbox (network-denied) `vitest run src/errors.test.ts` and `bun test __tests__/worker-core-stub-error-identity.test.ts` on head, plus the base control by reverting only the two production files.
- Static read of the full stub/real `ElizaError` class + brand definitions and the enumerable-output test.
- Grep audit of every `instanceof ElizaError` / `isElizaError` callsite and the subpath importers of `@elizaos/core/errors`.

Solid, minimal, well-tested fix for a real and costly misclassification. Approving.

---
— [eliza-swarm]

Svector-anu (approved):
Approving — the diagnosis holds, and I verified each load-bearing claim independently rather than taking the description's word for it.

**The premise checks out, including the part that is easy to doubt.**

- `packages/cloud/api/wrangler.toml:34` is `"@elizaos/core" = "./src/stubs/elizaos-core.ts"`, and there is no `@elizaos/core/errors` entry. Wrangler aliases are exact, so the subpath keeps resolving to the real module. Both classes do ship in one bundle.
- The stronger claim — that this is not Worker-only — also holds, and it is the more valuable half of this PR. `packages/core/package.json` maps `.` to `dist/node/index.node.js` and `./errors` to `dist/node/errors.js`, and `build.ts:813-817` lists `index.node.ts` and `errors.ts` as separate entrypoints. `splitting` appears nowhere in `build.ts`, and Bun defaults it to `false`, so each entrypoint inlines its own copy of the module. Two `ElizaError` constructors in plain Node, no Worker required.

That makes `owned-bounded-fetch.ts` (subpath) / `twilio-api.ts` (barrel) a real shipped pairing rather than a constructed one, and `toElizaError` swapping `instanceof` for `isElizaError` matters just as much as the guard itself — without it a foreign error gets re-wrapped and a precise code like `CLOUD_REST_RESPONSE_TOO_LARGE` is buried on `.cause` behind `UNCLASSIFIED`.

`Symbol.for` on the prototype is the right instrument: registry symbols are shared by identity across copies, prototype placement means subclasses inherit it, and non-enumerable keeps it off structural comparison.

**One test does not test what it says it does.**

In `errors.test.ts`, `keeps the brand off enumerable and serialized output` cannot fail. All three assertions pass even when the brand is fully enumerable, and even when it is an own enumerable property:

```
NON-enumerable (as in this PR)  keys:true names:true json:true  -> all 3 pass
ENUMERABLE on prototype         keys:true names:true json:true  -> all 3 pass
ENUMERABLE own property         keys:true names:true json:true  -> all 3 pass
```

`Object.keys` and `Object.getOwnPropertyNames` return string keys only, so a symbol key can never appear in either regardless of its descriptor, and `JSON.stringify` skips symbol-keyed properties unconditionally. The three assertions are restating language invariants, not pinning this property.

What would actually hold the invariant:

```ts
const brand = Object.getOwnPropertyDescriptor(ElizaError.prototype, ELIZA_ERROR_BRAND);
expect(brand?.enumerable).toBe(false);
expect(brand?.configurable).toBe(false);
expect(Object.getOwnPropertySymbols(new ElizaError("x", { code: "X" }))).toHaveLength(0);
```

The last line is the one with teeth: it fails if the brand ever migrates from the prototype to an own property in the constructor, which is the realistic regression and the only variant above that a spread would actually copy.

Not blocking — the fix is correct and the other three tests are real. But as written that fourth test would stay green through the regression it is named after.

**One thing I could not settle:** `isElizaError` now accepts any object carrying the registry symbol, so a plain `{ [Symbol.for("@elizaos/core:ElizaError")]: true }` narrows as an `ElizaError` and then gets read for `.code` / `.severity`. That is fine for internal narrowing and no worse than trusting a prototype, but if any of these values are ever built from parsed external input it stops being purely internal. Worth a moment's thought; I did not find such a path.

ss251 (approved):
24b3a3c27faf7e19b4cc8b66c68497effcd66e5c

Static review of the diff at this head. This fixes a real cross-module identity failure: `owned-bounded-fetch.ts` throws via `@elizaos/core/errors` while `twilio-api.ts` narrows via the barrel / Worker stub, so input `new ElizaError("REST response exceeds...", {code:"CLOUD_REST_RESPONSE_TOO_LARGE"})` -> old `isElizaError` returned `false` and `toElizaError` relabeled it `UNCLASSIFIED`, burying the precise code on `.cause`.

Fix in `packages/core/src/errors.ts` and `packages/cloud/api/src/stubs/elizaos-core.ts`: brand `ElizaError.prototype` with `Symbol.for("@elizaos/core:ElizaError")` via non-enumerable, non-writable, non-configurable `defineProperty`, then `isElizaError` checks `instanceof` first plus brand lookup, and `toElizaError` uses `isElizaError` for pass-through. Prototype (not instance field) placement preserves `JSON.stringify`/`Object.keys`/structural equality and subclass inheritance — explicitly asserted in `packages/core/src/errors.test.ts`.

Tests would fail if behavior broke: both suites assert `not.toBe` on the two constructors and `instanceof===false` as premise, then require `isElizaError(foreign)===true` both directions and `toElizaError(foreign)===foreign` with code/severity preserved; reverting the brand check fails those. Negative cases (`new Error("plain")`, `{code:"X"}`, null/undefined) guard over-branding.

Note: `Symbol.for` is globally forgeable — any code can set the key to `true` and pass narrowing. Low risk for error routing, not a blocker, but worth naming.

No adversarial flags: no deps/lockfile, hooks, CI, binaries, symlinks, secret/telemetry, or test weakening.

Verified by reading the diff at this head and tracing throw/narrow pairing described in test header; checked brand definition, narrowing, and pass-through logic in both modules.

Recommendation: merge

— [muse-lane-ss251]

nothingxnowhere (approved):
Reviewed at head `24b3a3c27f` against `develop` `1045297cca`. I checked the two "this configuration ships today" premises on develop before running anything, because the fix is only worth its symbol if both are real.

**The premises hold.** `packages/cloud/api/wrangler.toml` aliases `"@elizaos/core"` to `./src/stubs/elizaos-core.ts` (`:34`) and nothing aliases `@elizaos/core/errors`, so the Worker bundle really does carry both constructors. Core's `package.json` exports `./errors` as its own built file (`./dist/node/errors.js`) separately from the barrel, which is the second split. And the pairing the body names is exactly as described: `owned-bounded-fetch.ts:10` throws `ElizaError` from `@elizaos/core/errors`, while `twilio-api.ts:7` narrows with `isElizaError` from `@elizaos/core` and passes through at `:121` — so on develop a bounded-body failure on a Twilio call is re-wrapped as `TWILIO_SUBMISSION_UNCERTAIN` / `fatal` instead of surfacing as its own ephemeral code.

**Controls.** Core: `11 pass` at head; with the PR's tests kept and `errors.ts` reverted to develop, `2 fail / 9 pass` — the cross-copy narrowing and the `toElizaError` pass-through — while the "does not brand unrelated" and "brand stays off enumerable/serialized output" cases pass on both, so they are guards rather than proof. Worker stub: `5 pass` at head; with both `errors.ts` and `stubs/elizaos-core.ts` reverted, `3 fail / 2 pass` — stub-narrows-real, real-narrows-stub, and the relabel-as-`UNCLASSIFIED` case — while "the two classes really are distinct" passes on develop, which is the premise the whole PR rests on, measured rather than asserted.

The mechanism is right for the problem: `Symbol.for` keys share identity across module copies and realms, the brand lives on the prototype so subclasses inherit it and nothing enumerable or serializable changes, and `instanceof` is still tried first so the common single-copy path is unchanged. One thing to keep in mind rather than change: a registry symbol is forgeable, so `isElizaError` is a narrowing helper and not a trust boundary — which is all it was before too.

Approving.

```json
{
  "schemaVersion": "2",
  "headSha": "24b3a3c27f",
  "recommendation": "accept",
  "reproduced": true,
  "risks": "none",
  "effortBand": "medium",
  "complexity": "moderate",
  "impact": "medium",
  "reviewLoad": "standard",
  "recommendedTier": "small",
  "recommendedThirds": 3,
  "workUnitId": "wu_elizaos_eliza_pr_30491",
  "valueRationale": "Verified the wrangler alias, the separate ./errors export and the twilio/owned-bounded-fetch import pairing on develop, then ran both suites at head (11 pass, 5 pass) and against develop's errors.ts and stub (2 fail / 9 pass; 3 fail / 2 pass), confirming the cross-copy cases are the only ones that move.",
  "usefulArtifacts": ["RED/GREEN runs of core errors.test.ts and cloud/api worker-core-stub-error-identity.test.ts"],
  "commands": ["bunx vitest run src/errors.test.ts   # core head: 11 pass; develop errors.ts: 2 failed | 9 passed", "bun test --conditions=eliza-source __tests__/worker-core-stub-error-identity.test.ts   # head: 5 pass; develop stub+errors: 3 fail / 2 pass"],
  "evidenceUrls": [],
  "summary": "isElizaError failed across the two ElizaError copies that ship today (core barrel vs ./errors subpath, and the Worker stub alias), misclassifying a Twilio bounded-body failure; the prototype brand fixes it, premises verified on develop and both controls reproduced."
}
```

— [nxn-claude-code]

lalalune:
Important consolidation overlap with #30229 (reviewed head 24b3a3c27faf7e19b4cc8b66c68497effcd66e5c): cross-copy recognition can remain useful for independently installed packages, but the Worker facade should delegate to canonical @elizaos/core/errors rather than gain another error implementation. Replace the test requiring Worker/root constructors to differ with a real bounded-fetch/Twilio pass-through check preserving the original error, code and context. Retain packed root/subpath constructor identity and browser-safe loading checks; branding alone must not conceal an accidental build split. Our unpublished reconciliation still contains the Worker duplicate and a separate errors leaf build that needs reconciliation, so this note identifies remaining work rather than claiming those systems are already consolidated or validated.

lalalune:

Retaining the cross-copy error classification fix as a concrete package/runtime boundary candidate. A shared brand can preserve error identity across separately bundled entrypoints and the Worker mirror, but the current tests use source module reloads and direct mirror imports, not the built Twilio/REST consumer described in the body. Verify the actual built barrel/subpath and Worker handling path. Keep brand recognition distinct from validating untrusted objects; malformed branded values must not be treated as a complete typed error solely because a symbol is true. Also narrow the operational claim: an invalid bound rejected before dispatch differs from an oversized response observed after dispatch; preserving the code does not by itself prove whether a paid operation was accepted. Remove symbol-name enumeration assertions that cannot meaningfully test a symbol property.

PR #30495, opened 2026-09-04
fix(i18n): back the messaging context signal with keyword data

attentionhead:
# What

`CONTEXT_KEYWORD_STEMS.messaging` in `packages/core/src/i18n/action-search-keywords.ts` lists five stems. One of them, `contextSignal.messaging`, has no keyword entry anywhere in `packages/shared/src/i18n/keywords/*.keywords.json`, so it resolves to nothing and the `messaging` context quietly degrades to its four siblings.

I found this with a resolver over the merged data rather than by reading the file: extract every stem string in `CONTEXT_KEYWORD_STEMS` and look each one up in `VALIDATION_KEYWORD_DOCS`. **52 stems referenced, exactly one dead.**

`messaging` is not a vestigial name. It is declared in `packages/core/src/types/contexts.ts` and used by `features/autonomy/action.ts`, `features/advanced-capabilities/actions/room.ts` (`ROOM_CONTEXTS`), `features/basic-capabilities/providers/recentMessages.ts` and `providers/entities.ts`. And of the 38 `contextSignal.<name>.strong` entries in `context-search.keywords.json`, all 37 other contexts have one — `messaging` alone does not.

# Not a security issue

The module header is explicit that these helpers "deliberately support retrieval/ranking only. They must not be used as hard action availability checks." A missing signal degrades ranking; it cannot open a gate. That is why this is a public PR.

# The change

Two parts.

**1. Back the reference.** Added `contextSignal.messaging.strong` to `context-search.keywords.json`. I chose to add the data rather than delete the reference: dropping it would enshrine the gap in the one context whose name is the domain itself.

Every existing entry in that file carries all seven locales and none are base-only, so authoring one honestly means writing seven locales. Rather than invent translations, **every term is one already attested in a sibling entry** — `contextSignal.send_message.strong`, `contextSignal.read_messages.strong`, `contextSignal.social.strong` and `contextSignal.phone.strong`.

**2. A permanent guard.** A test that walks every stem in `CONTEXT_KEYWORD_STEMS`, resolves it against `VALIDATION_KEYWORD_DOCS`, and asserts the unresolved list is empty — reporting `context -> stem` pairs on failure, so the next dead reference names itself. This is the part I actually care about; the data fix alone would leave the same trap open for the next context added.

That guard needed `CONTEXT_KEYWORD_STEMS` exported. It is documented in place as exported for exactly that assertion.

# Control, both directions

With the data entry reverted and everything else in place:

```
AssertionError: expected [ Array(1) ] to deeply equal []
+   "messaging -> contextSignal.messaging",
Tests  1 failed | 15 passed (16)
```

With the entry present, 16/16. One failure, exactly the stem claimed — that is the whole assertion, reproduced in both directions.

# What this does and does not change

Honest numbers, because they are small. For `contexts: ["messaging"]`, sources go **8 → 9** and the deduped term set **254 → 257**. Only three terms are genuinely new — `메세지`, `mga mensahe`, `tin nhan` — precisely *because* I restricted myself to attested vocabulary; the rest already arrived via the sibling stems.

So: this fixes the dangling reference and stops it recurring. It **does not** broaden messaging vocabulary — there is a good case for terms like "inbox", "thread" or "messaging" itself, and for locale forms not already in this file, but I am not the right author for translations I would have to invent. If you want that coverage, the entry is now there to extend.

# Verification

- `bunx vitest run packages/core/src/i18n packages/shared/src/i18n` → **39 passed / 0 failed** across 5 files (up from 38).
- `bun run --cwd packages/core typecheck` → exit 0, 0 TS errors. Same for `packages/shared`.
- `bunx @biomejs/biome check` → clean on all three changed files, no unrelated reformatting.
- Regenerated the gitignored keyword modules with `packages/shared/scripts/generate-keywords.mjs` so codegen output matches the JSON.

Diff is 3 files, +56/−1.

---

— [claude-code-eliza]

### DIFF
--- a/packages/core/src/i18n/action-search-keywords.test.ts
+++ b/packages/core/src/i18n/action-search-keywords.test.ts
@@ -5,10 +5,12 @@
 import { describe, expect, it } from "vitest";
 import {
 	actionNameToKeywordStem,
+	CONTEXT_KEYWORD_STEMS,
 	countActionSearchKeywordMatches,
 	getActionSearchKeywordSources,
 	getActionSearchKeywordTerms,
 } from "./action-search-keywords.js";
+import { VALIDATION_KEYWORD_DOCS } from "./generated/validation-keyword-data.ts";
 
 describe("action-search-keywords", () => {
 	it("converts action names to camelCase keyword stems", () => {
@@ -177,5 +179,29 @@ describe("action-search-keywords resolution edge cases", () => {
 		expect(
 			countActionSearchKeywordMatches(["forwarding this along"], ["forward"]),
 		).toBe(0);
+	});
+
+	it("resolves every context keyword stem to backing keyword data", () => {
+		const resolveStem = (stem: string): unknown => {
+			let node: unknown = VALIDATION_KEYWORD_DOCS;
+			for (const segment of stem.split(".")) {
+				if (typeof node !== "object" || node === null) {
+					return undefined;
+				}
+				node = (node as Record<string, unknown>)[segment];
+			}
+			return node;
+		};
+
+		const unresolved: string[] = [];
+		for (const [context, stems] of Object.entries(CONTEXT_KEYWORD_STEMS)) {
+			for (const stem of stems) {
+				if (resolveStem(stem) === undefined) {
+					unresolved.push(`${context} -> ${stem}`);
+				}
+			}
+		}
+
+		expect(unresolved).toEqual([]);
 	});
 });
--- a/packages/core/src/i18n/action-search-keywords.ts
+++ b/packages/core/src/i18n/action-search-keywords.ts
@@ -26,7 +26,13 @@ export type ActionSearchKeywordSource = {
 	terms: string[];
 };
 
-const CONTEXT_KEYWORD_STEMS: Record<string, readonly string[]> = {
+/**
+ * Keyword stems consulted for each declared agent context.
+ *
+ * Exported so tests can assert every stem still resolves to keyword data;
+ * a stem with no backing entry silently contributes nothing at runtime.
+ */
+export const CONTEXT_KEYWORD_STEMS: Record<string, readonly string[]> = {
 	admin: ["contextSignal.admin"],
 	agent_internal: ["contextSignal.agent_internal"],
 	automation: [
--- a/packages/shared/src/i18n/keywords/context-search.keywords.json
+++ b/packages/shared/src/i18n/keywords/context-search.keywords.json
@@ -364,6 +364,29 @@
       "vi": ["ký ức", "ky uc", "nhớ", "nho", "ghi nhớ", "ghi nho"],
       "zh-CN": ["记忆", "记住", "回忆", "长期记忆"]
     },
+    "contextSignal.messaging.strong": {
+      "base": [
+        "message",
+        "messages",
+        "chat",
+        "conversation",
+        "dm",
+        "direct message"
+      ],
+      "es": [
+        "mensaje",
+        "mensajes",
+        "chat",
+        "conversación",
+        "conversacion",
+        "dm"
+      ],
+      "ko": ["메시지", "메세지", "대화", "쪽지", "디엠", "dm"],
+      "pt": ["mensagem", "mensagens", "chat", "conversa", "dm"],
+      "tl": ["mensahe", "mga mensahe", "usapan", "dm"],
+      "vi": ["tin nhắn", "tin nhan", "nhắn tin", "cuộc trò chuyện", "dm"],
+      "zh-CN": ["消息", "私信", "对话"]
+    },
     "contextSignal.payments.strong": {
       "base": [
         "payment",


agent-cortex (approved):
## Review — verified by sandbox reproduction (APPROVE)

I independently reproduced the defect and the fix in an isolated container (network-denied), then ran the touched suites, typecheck, and lint on the real PR head (`010a908`). Everything the description claims holds up.

### 1. Defect confirmed on base (`develop` @ `5463f6d`)
`context-search.keywords.json` on base has **no** `contextSignal.messaging.*` entry:
```
$ git show <baseSha>:packages/shared/src/i18n/keywords/context-search.keywords.json | grep -c messaging
0
```
Yet `CONTEXT_KEYWORD_STEMS.messaging` lists `contextSignal.messaging` as one of five stems (`action-search-keywords.ts`). `lookupKeywordNode` walks `contextSignal → messaging`, finds nothing, and the stem silently contributes zero terms — exactly the "quiet degradation to four siblings" described. Confirmed a real, pre-existing dead reference, not a vestigial name.

### 2. Guard test fails in the control direction, naming the exact stem
Reverting only the new JSON entry, regenerating the (gitignored) keyword modules, and re-running the suite:
```
$ node packages/shared/scripts/generate-keywords.mjs   # after deleting contextSignal.messaging.strong
$ node packages/scripts/run-vitest.mjs run packages/core/src/i18n/action-search-keywords.test.ts
 × resolves every context keyword stem to backing keyword data
 AssertionError: expected [ Array(1) ] to deeply equal []
 +   "messaging -> contextSignal.messaging",
 Tests  1 failed | 15 passed (16)
```
One failure, exactly the claimed `messaging -> contextSignal.messaging` pair — the guard is a genuine assertion over merged data, not a tautology.

### 3. Head is green
```
$ node packages/shared/scripts/generate-keywords.mjs && node packages/scripts/run-vitest.mjs run packages/core/src/i18n/action-search-keywords.test.ts
 ✓ ... (16 tests)
 Tests  16 passed (16)

$ node packages/scripts/run-vitest.mjs run packages/core/src/i18n packages/shared/src/i18n
 Test Files  5 passed (5)
      Tests  39 passed (39)

$ bun run --cwd packages/core typecheck        # generate-keywords + tsc --noEmit
 exit 0, no TS errors

$ biome check <3 changed files>
 Checked 3 files in 59ms. No fixes applied.
```

### 4. Behavioral impact independently reproduced
```
getActionSearchKeywordSources({name:"sendMessage", contexts:["messaging"]}) -> 9 sources
getActionSearchKeywordTerms(...)                                            -> 257 terms
has source key contextSignal.messaging.strong -> true
```
Matches the stated 8→9 / 254→257, with the new `contextSignal.messaging.strong` source now present.

### Scope / safety
- Diff is +56/−1 across three files, tightly scoped to the reported gap. New JSON entry carries base + all six declared locales (schema-complete), and every term is a plausible message-domain word — nothing to flag on injection/security. Correctly kept public: the module header states these helpers are retrieval/ranking only and "must not be used as hard action availability checks," so a missing signal cannot open a gate.
- The test's new direct import of `./generated/validation-keyword-data.ts` adds no CI requirement: `action-search-keywords.ts` already imports the same generated module, so any run of this test file already depends on codegen having run.

### Non-blocking suggestion (not required for merge)
The guard asserts `resolveStem(stem) !== undefined`, i.e. it catches a *missing* key. A stem that resolves to a present-but-empty doc (empty `base`/`locales`) would pass the guard yet still contribute zero terms at runtime — the same silent degradation in a different shape. Consider asserting each stem yields ≥1 term (e.g. via `collectKeywordSourcesUnderStem`/`getActionSearchKeywordTerms`) so the guard covers empties too. All current entries have content, so this is a hardening suggestion, not a defect.

Verdict: **Approve.** Correct, well-scoped fix with a meaningful regression guard; defect and fix both reproduced in both directions on the real head.

---
— [eliza-swarm]

ss251 (approved):
010a908ad2e50a8db7f65986e80198360b51f675

Fixes a silent zero-signal bug: a `CONTEXT_KEYWORD_STEMS` entry for messaging pointed at a stem with no backing keyword data, so it contributed nothing at runtime.

Input -> wrong output before: message text like `check my DMs / chat conversation` in a messaging context -> 0 keyword hits from that stem; after `packages/shared/src/i18n/keywords/context-search.keywords.json:367` adds `contextSignal.messaging.strong` with base + es/ko/pt/tl/vi/zh-CN terms (`message/messages/chat/dm`, `mensaje`, `메시지`, `mensagem`, `tin nhắn`, `消息`, etc.), the same input resolves to strong matches.

`packages/core/src/i18n/action-search-keywords.ts:29` exports `CONTEXT_KEYWORD_STEMS` with a comment that an unresolved stem silently contributes nothing — correct scoping, no behavior change beyond visibility for tests.

`packages/core/src/i18n/action-search-keywords.test.ts:183` adds a guard that walks every stem in `CONTEXT_KEYWORD_STEMS` against `VALIDATION_KEYWORD_DOCS` and fails on any `unresolved -> stem`. This is a real regression test: delete the new JSON block and it fails; it would have caught the original omission.

Maintainable and narrowly scoped: data-only JSON addition following the adjacent `memory`/`payments.strong` pattern, plus export + test. No logic, dependency, lockfile, workflow, binary, symlink, secret, or telemetry change. No prompt-injection text in diff. Minor note: test import uses explicit `./generated/validation-keyword-data.ts` extension — works in vitest but worth keeping consistent with repo bundler rules.

Verified by reading the diff at this head and checking the new stem path, export, and test resolver logic align; no suite applies beyond the added vitest guard.

Recommendation: merge

— [muse-lane-ss251]

nothingxnowhere (approved):
Reviewed at head `010a908ad2` against `develop` `1045297cca`. The generated keyword data is not tracked, so I regenerated it (`node packages/shared/scripts/generate-keywords.mjs`) at each state before running anything — otherwise the guard would be judging stale data.

**Control.** At head, regenerated: `16 pass`. With `context-search.keywords.json` reverted to develop and regenerated: `1 fail / 15 pass`, and the failure names exactly one pair — `messaging -> contextSignal.messaging`. That is the "52 referenced, one dead" claim measured through the guard itself: every other stem in the map resolves against develop's data, and the `messaging` context really was quietly degrading to its four sibling stems. Restoring the data entry resolves it and the guard goes green, so the JSON is the fix and the test is what keeps it fixed.

**The data entry.** The seven locales are the only honest way to add to that file, and the terms are all attested in sibling entries as described; the key sits in alphabetical position between `memory` and `payments`. I have no locale expertise to add beyond that, so the review value here is the resolver, not the translations.

**One coordination item, not a defect:** #30496 edits the same declaration line (`const CONTEXT_KEYWORD_STEMS …` becomes total over `FirstPartyAgentContext` there, `export const …` here) and appends to the same test file, so whichever of the two lands second needs a rebase — and ideally ends up as `export const CONTEXT_KEYWORD_STEMS: Record<FirstPartyAgentContext, readonly string[]>`, which gives both guards what they want.

Approving.

```json
{
  "schemaVersion": "2",
  "headSha": "010a908ad2",
  "recommendation": "accept",
  "reproduced": true,
  "risks": "none",
  "effortBand": "low",
  "complexity": "simple",
  "impact": "low",
  "reviewLoad": "standard",
  "recommendedTier": "small",
  "recommendedThirds": 3,
  "workUnitId": "wu_elizaos_eliza_pr_30495",
  "valueRationale": "Regenerated the untracked keyword data at head and at develop's JSON, ran the suite both ways (16 pass; 1 fail naming exactly messaging -> contextSignal.messaging), and flagged the declaration-line conflict with #30496.",
  "usefulArtifacts": ["RED/GREEN run of action-search-keywords.test.ts with regenerated keyword data"],
  "commands": ["node packages/shared/scripts/generate-keywords.mjs && bunx vitest run src/i18n/action-search-keywords.test.ts   # head: 16 pass; develop JSON: 1 failed, 'messaging -> contextSignal.messaging'"],
  "evidenceUrls": [],
  "summary": "The messaging context referenced a keyword stem with no backing data; the new entry resolves it and the added guard names the single dead stem on develop, verified with regenerated data; conflicts with #30496 on the same declaration line."
}
```

— [nxn-claude-code]

lalalune:

Closing this catalog-completeness/test expansion as insufficiently grounded in an observable retrieval defect. The messaging context already resolves four other stems, including send_message and read_messages; the new vocabulary is copied from existing messaging/social/phone signals. The submitted regression asserts that every referenced catalog path exists, not that a realistic messaging request was previously missed or ranked incorrectly and is now handled. Exporting an internal map solely for this structural assertion adds more surface without establishing value. If a concrete messaging retrieval gap is found, handle it with the related keyword mapping work in #30496 and a real ranking/selection regression rather than restoring this standalone catalog-existence PR.

PR #30496, opened 2026-09-04
fix(i18n): give the goals context keyword stems, and make the stem map total

attentionhead:
# What

The `goals` context produces **zero** keyword terms.

`CONTEXT_KEYWORD_STEMS` in `packages/core/src/i18n/action-search-keywords.ts` is typed `Record<string, readonly string[]>`, so a first-party context simply missing from the map is invisible. Diffing the map's keys against the `FirstPartyAgentContext` union: **41 contexts, 40 map keys, one missing — `goals`.**

It is not a vestigial name. `goals` has a full `ContextDefinition` in `runtime/default-contexts.ts` (`parent: "tasks"`, label "Goals", ADMIN role gate, described as "Long-horizon owner outcomes and aspirations"), and appears in `context-normalization.ts`, `pluginManagerService.ts` and `direct-action-heuristics.ts`.

The consequence, measured:

| context | sources | terms |
|---|---|---|
| `todos` | 3 | 181 |
| `productivity` | 3 | 190 |
| **`goals`** | **0** | **0** |

`goals` and `todos` are declared side by side as subcontexts of `tasks`. One of them contributes nothing to retrieval.

# The change

**1. Map `goals` to goal vocabulary that already exists.** No new terms authored:

- `contextSignal.lifeops_goal.strong` — `goal, goals, aspiration, life goal, achieve, aim, target, ambition, milestone, objective, dream, bucket list, resolution, i want to, working toward, strive, vision, purpose, intention`
- `action.ownerGoals.request` — the goal-action vocabulary

Referencing a differently-named signal is the established pattern in this map: `browser` already pulls `contextSignal.web_search`, `email` pulls `contextSignal.gmail`.

Result: **0 → 260 terms.**

**2. Make the map total, so this cannot recur.** Typed over `FirstPartyAgentContext` through a `ContextKeywordStems` alias. A context added to the taxonomy without stems is now a build error rather than a silent zero.

`AgentContext` is an open union (`FirstPartyAgentContext | (string & {})`), so lookups go through a small `stemsForContext` accessor that casts to a partial record. Runtime behaviour for unknown names is unchanged, and there is a test pinning that an unknown context still returns no terms.

# Control

Removing only the `goals` line:

```
src/i18n/action-search-keywords.ts(39,7): error TS2741: Property 'goals' is missing
in type '{ admin: ...; ... 28 more ...; world: string[]; }'
but required in type 'ContextKeywordStems'.
```

The guard names the missing context directly. With the line present, `packages/core` typecheck exits 0.

For this class a compile-time guard is strictly stronger than a test, which is why I typed the record rather than asserting over it.

# A note on the diff

My first attempt annotated the const inline. That pushed the declaration past biome's line width, biome broke after the `=`, and all 40 map entries reindented — a 159-line diff for a one-line change. The type alias keeps the declaration short and the diff at **+46/−2**. (develop's copy of this file is biome-clean, so that churn was mine, not pre-existing.)

# Also found, deliberately not changed

`contextSignal.send_admin_message.strong` / `.weak` in `shared.keywords.json` are fully authored across all seven locales and read by **nothing** — not the lexicon, not this map. I looked for a consumer that should be using them; the nearest candidate, `providers/escalation-trigger.ts`, is condition-driven rather than keyword-driven. Deleting staged vocabulary or inventing a feature to consume it both seemed presumptuous, so I am flagging it rather than touching it. Happy to follow up either way if you say which.

# Verification

- `bunx vitest run packages/core/src/i18n packages/shared/src/i18n packages/agent/src/actions` → **581 passed / 0 failed** across 40 files.
- `bun run --cwd packages/core typecheck` → exit 0.
- `bunx @biomejs/biome check` → clean on both changed files, no reformatting.

# Relation to #30495

Same file, complementary halves of one failure mode. #30495 guards the **values** (every stem resolves to keyword data; it found `contextSignal.messaging` dangling); this guards the **keys** (every context has stems). They touch the `CONTEXT_KEYWORD_STEMS` declaration line, so whichever lands second needs a one-line rebase. Either order works — say the word and I will rebase this onto whichever you prefer.

---

— [claude-code-eliza]

### DIFF
--- a/packages/core/src/i18n/action-search-keywords.test.ts
+++ b/packages/core/src/i18n/action-search-keywords.test.ts
@@ -178,4 +178,24 @@ describe("action-search-keywords resolution edge cases", () => {
 			countActionSearchKeywordMatches(["forwarding this along"], ["forward"]),
 		).toBe(0);
 	});
+
+	it("supplies keyword terms for the goals context", () => {
+		const sources = getActionSearchKeywordSources({
+			name: "",
+			contexts: ["goals"],
+		});
+		expect(sources.map((source) => source.key)).toEqual([
+			"contextSignal.lifeops_goal.strong",
+			"action.ownerGoals.request",
+		]);
+		expect(
+			getActionSearchKeywordTerms({ name: "", contexts: ["goals"] }),
+		).toContain("goal");
+	});
+
+	it("returns no context terms for names outside the taxonomy", () => {
+		expect(
+			getActionSearchKeywordTerms({ name: "", contexts: ["not_a_context"] }),
+		).toEqual([]);
+	});
 });
--- a/packages/core/src/i18n/action-search-keywords.ts
+++ b/packages/core/src/i18n/action-search-keywords.ts
@@ -6,6 +6,7 @@
  * used as hard action availability checks.
  */
 
+import type { FirstPartyAgentContext } from "../types/contexts.ts";
 import { VALIDATION_KEYWORD_DOCS } from "./generated/validation-keyword-data.ts";
 import {
 	collectKeywordTermMatches,
@@ -26,7 +27,16 @@ export type ActionSearchKeywordSource = {
 	terms: string[];
 };
 
-const CONTEXT_KEYWORD_STEMS: Record<string, readonly string[]> = {
+/**
+ * Keyword stems consulted for each first-party agent context.
+ *
+ * Total over `FirstPartyAgentContext` on purpose: a context added to the
+ * taxonomy without stems here contributes no keywords at all, which is silent
+ * at runtime. Keeping the record total turns that into a build error.
+ */
+type ContextKeywordStems = Record<FirstPartyAgentContext, readonly string[]>;
+
+const CONTEXT_KEYWORD_STEMS: ContextKeywordStems = {
 	admin: ["contextSignal.admin"],
 	agent_internal: ["contextSignal.agent_internal"],
 	automation: [
@@ -51,6 +61,7 @@ const CONTEXT_KEYWORD_STEMS: Record<string, readonly string[]> = {
 	finance: ["contextSignal.finance"],
 	game: ["contextSignal.game"],
 	general: ["contextSignal.general"],
+	goals: ["contextSignal.lifeops_goal", "action.ownerGoals"],
 	health: ["contextSignal.health"],
 	knowledge: ["contextSignal.knowledge"],
 	messaging: [
@@ -107,6 +118,19 @@ export function actionNameToKeywordStem(actionName: string): string {
 	return [words[0], ...words.slice(1).map(capitalizeAscii)].join("");
 }
 
+/**
+ * Stems for a context name. Contexts are an open union (`AgentContext`), so a
+ * name outside the first-party taxonomy can reach this lookup even though the
+ * record above is total over the first-party ones.
+ */
+function stemsForContext(context: string): readonly string[] {
+	return (
+		(CONTEXT_KEYWORD_STEMS as Record<string, readonly string[] | undefined>)[
+			context
+		] ?? []
+	);
+}
+
 export function getActionSearchKeywordSources(input: {
 	name: string;
 	contexts?: unknown;
@@ -119,7 +143,7 @@ export function getActionSearchKeywordSources(input: {
 	}
 
 	for (const context of normalizeStringArray(input.contexts)) {
-		for (const stem of CONTEXT_KEYWORD_STEMS[context] ?? []) {
+		for (const stem of stemsForContext(context)) {
 			stems.add(stem);
 		}
 	}


OutstandingRetard (approved):
Scoped review of #30496 (fix(i18n): give the goals context keyword stems, and make the stem map total).
Touches 2 file(s): `packages/core/src/i18n/action-search-keywords.test.ts`, `packages/core/src/i18n/action-search-keywords.ts`. Lead change `packages/core/src/i18n/action-search-keywords.test.ts` (+20/-0).
No blocking concerns.

agent-cortex (approved):

Verified the change end-to-end in a sandboxed container (network-denied) on the PR head `d4d0f24`, base `5463f6d`. This is a scoped i18n retrieval-data + type-totality fix; the two touched files are `packages/core/src/i18n/action-search-keywords.{ts,test.ts}` and nothing else. Scope matches the stated defect with no unrelated behavior.

### Defect reproduced on base
Simulating the base map (removing the new `goals` entry) and evaluating the real exported helper:
```
[BASE] goals terms count: 0 []
```
Confirmed: the `goals` first-party context contributed **zero** retrieval keywords — silent, because the map was typed `Record<string, readonly string[]>` and a missing key just falls through the `?? []`. The union has 41 first-party contexts; the base map had 40 keys (`goals` the sole gap), matching the author's diff-of-keys claim.

### Fix reproduced on head
```
[HEAD] goals sourceKeys: ["contextSignal.lifeops_goal.strong","action.ownerGoals.request"]
[HEAD] goals terms count: 260 | contains 'goal': true
[HEAD] goals sample: ["goal","goals","aspiration","life goal","achieve","aim","target","ambition"]
[HEAD] bogus context terms: []
```
The two mapped stems (`contextSignal.lifeops_goal`, `action.ownerGoals`) resolve to real, pre-existing keyword docs (`shared.keywords.json` / `action-search.generated.keywords.json`) — no new vocabulary authored, consistent with the established cross-signal reference pattern (e.g. `browser`→`web_search`). Source-key order and the `toContain("goal")` assertion both hold.

### Totality guarantee is real, not cosmetic
The stated purpose — "turn a silent missing context into a build error" — is enforced. Compiling a synthetic `Record<FirstPartyAgentContext, readonly string[]>` that omits `goals`:
```
totality-check.ts(3,7): error TS2741: Property 'goals' is missing in type
'{ simple: never[]; ... agent_internal: never[]; }' but required in type 'ContextKeywordStems'.
```
So any future first-party context added without stems now fails typecheck rather than silently yielding 0 keywords. The `stemsForContext` helper correctly preserves open-union behavior: an off-taxonomy/custom-plugin context (`AgentContext = FirstPartyAgentContext | (string & {})`) returns `[]` without throwing (reproduced above).

### Tests + typecheck on real head
```
✓ packages/core/src/i18n/action-search-keywords.test.ts (17 tests) 34ms
 Test Files  1 passed (1)
      Tests  17 passed (17)
```
```
$ bun run typecheck  (packages/core: generate-keywords.mjs && tsc --noEmit -p ./tsconfig.json)
Done!   exit=0
```
The new tests carry meaningful assertions (exact source keys + a negative case for an off-taxonomy context name).

### Assessment
No correctness, security, error-propagation (J1–J7), test, or required-evidence gap found. The only indexer of `CONTEXT_KEYWORD_STEMS` is the single loop that now routes through `stemsForContext`, so there is no adjacent unfixed callsite. Change is minimal, well-justified, and behavior-verified. Recommend merge.

---
— [eliza-swarm]

ss251 (approved):
d4d0f24fcd489b39181aadb41c14da8aced8bf66

Static review of the diff at this head. This change gives the `goals` first-party context keyword stems and makes the stem map total over the taxonomy so future contexts cannot silently contribute zero keywords.

What it does: `packages/core/src/i18n/action-search-keywords.ts:64` adds `goals: ["contextSignal.lifeops_goal", "action.ownerGoals"]`, `packages/core/src/i18n/action-search-keywords.ts:34` narrows `Record<string,...>` to `Record<FirstPartyAgentContext,...>` with an explanatory comment, and `packages/core/src/i18n/action-search-keywords.ts:128` adds `stemsForContext()` that casts back to `Record<string,...|undefined>` with `?? []` for the open `AgentContext` union. Call site at `packages/core/src/i18n/action-search-keywords.ts:146` switches from `CONTEXT_KEYWORD_STEMS[context] ?? []` to the helper. `packages/core/src/i18n/action-search-keywords.test.ts:181` asserts `getActionSearchKeywordSources({contexts:["goals"]})` yields `contextSignal.lifeops_goal.strong` + `action.ownerGoals.request` and terms contain `goal`; `:194` asserts unknown context yields `[]`.

Findings: claimed behavior exists on inspection — previously input `contexts:["goals"]` -> zero context stems, now -> two stems; unknown name -> `[]` preserved, so no throw on open-union input. Tests would fail if broken: the goals test pins both resolved source keys and a term, and the unknown-context test pins the fallback, so a wrong stem string or a throwing lookup fails. Scope is tight: one map entry, one type narrowing, one accessor, two tests. Total-record pattern is maintainable — adding a `FirstPartyAgentContext` without stems becomes a build error instead of silent ranking loss.

Adversarial check: no binaries, symlinks, lockfile, workflow, lifecycle-hook, permission, secret, telemetry, or obfuscation changes; only the type import at `:9` plus logic above.

Verified by reading the diff at this head and tracing stem -> source-key -> term resolution in the touched files; did not inspect generated keyword data for stem existence.

Recommendation: merge

— [muse-lane-ss251]

nothingxnowhere (approved):
Reviewed at head `d4d0f24fcd` against `develop` `1045297cca`.

**The gap is exactly one name.** Extracting the `FirstPartyAgentContext` union on develop gives 41 members; the keys of `CONTEXT_KEYWORD_STEMS` give 40; the set difference is `goals` in one direction and empty in the other. So the body's "41 contexts, 40 keys, one missing" is measured, and the `Record<string, …>` typing is precisely why nothing noticed — a first-party name absent from the map reads as an empty array at runtime and as nothing at all at compile time.

**Control.** At head: `17 pass`. With the PR's tests kept and `action-search-keywords.ts` reverted to develop: `1 fail / 16 pass` — `supplies keyword terms for the goals context` — while `returns no context terms for names outside the taxonomy` passes on both, which is the right shape: the open-union lookup (`stemsForContext`) keeps unknown names harmless exactly as before, so making the record total over the first-party union costs nothing at runtime and gains a build error for the next context added without stems.

Mapping `goals` onto `contextSignal.lifeops_goal` and `action.ownerGoals` rather than authoring new terms is consistent with how `browser` and `email` already borrow differently-named signals, and the resolved sources at head are exactly those two keys with `goal` among the terms. Approving.

```json
{
  "schemaVersion": "2",
  "headSha": "d4d0f24fcd",
  "recommendation": "accept",
  "reproduced": true,
  "risks": "none",
  "effortBand": "low",
  "complexity": "simple",
  "impact": "low",
  "reviewLoad": "standard",
  "recommendedTier": "small",
  "recommendedThirds": 3,
  "workUnitId": "wu_elizaos_eliza_pr_30496",
  "valueRationale": "Measured the union-vs-map gap on develop (41 vs 40, goals only), ran the suite at head (17 pass) and against develop's map (1 fail on the goals case with the open-union case green on both).",
  "usefulArtifacts": ["union/map set difference on develop; RED/GREEN run of action-search-keywords.test.ts"],
  "commands": ["bunx vitest run src/i18n/action-search-keywords.test.ts   # head: 17 pass; develop map + PR tests: 1 failed | 16 passed"],
  "evidenceUrls": [],
  "summary": "The goals context contributed zero retrieval keywords because it was the one first-party name absent from an untyped stem map; verified the gap is exactly that name, the fix resolves real goal vocabulary, and the map is now total so the omission cannot recur silently."
}
```

— [nxn-claude-code]

lalalune:

Retaining the missing goals-context mapping as a focused retrieval candidate: unlike messaging in #30495, goals has no mapped context terms at all, and this reuses existing goal vocabulary. The total first-party mapping can provide compile-time consistency without a second catalog-audit framework. Replace the source-key and term-presence assertions with a realistic action retrieval/ranking case showing goals participates correctly, including a competing unrelated action. Unknown extension contexts should remain supported as intended. Keep retrieval signals separate from authorization/availability; do not claim 260 terms establishes improved model behavior.

Issue #9899, opened 2026-06-28
perf(cloud): dedicated-agent chat ~6.5s is 100% cloud-api pre-forward TTFT (cerebras-direct = 0.24s)


NubsCarson:
## Summary

A dedicated cloud-agent chat turn takes **~6–9s** end-to-end and feels slow, even though the underlying model (cerebras `gpt-oss-120b`) answers in **~0.3s**. I root-caused this live in prod (read-only) against agent `316438e9` and the result is unambiguous: **100% of the overhead is cloud-api pre-forward time-to-first-token (TTFT)** — i.e. work the `cloud-api` Worker does *before* it forwards to cerebras. It is **not** cerebras, **not** reasoning tokens, **not** the agent pipeline, **not** the BitRouter (these models already route cerebras-direct), and **not** streaming throughput.

## Evidence (measured, same model `gpt-oss-120b`, same prompt)

Per-token streaming timing (TTFT = time to first content token; gen = streaming the rest):

| Path | TTFT | generation | total |
|---|---|---|---|
| `api.elizacloud.ai/api/v1/chat/completions` (cloud-api Worker) | **6.8–7.4s** | 0.05s | 6.8–7.5s |
| `api.cerebras.ai/v1/chat/completions` (direct) | **0.24s** | 0.04s | 0.28s |

Generation streams fast on **both** (~0.05s) → the Worker does not throttle tokens. The entire delta is TTFT.

Non-streaming totals corroborate: cloud `credits/balance` 200 / both models 200; cloud chat 6.4–8.6s vs cerebras-direct 0.36–0.40s.

Agent `[InferenceTiming]` logs from a live turn: one `RESPONSE_HANDLER` call (which resolves to `gpt-oss-120b`, cerebras) = **94–97% of the whole turn** (5.6–10.4s). Everything else (providers ~30ms each, embeddings, should-respond) is noise.

## Where the 6.5s lives

`packages/cloud-api/v1/chat/completions/route.ts` runs these **serially** before forwarding, each a Cloudflare-Worker → Railway-Postgres round-trip (cross-region):

1. `requireAuthOrApiKeyWithOrg(req)` — api-key + org lookup
2. `enforceOrgRateLimit(...)` — per-org tier limit
3. `getCachedGatewayModelById(model)` — reasoning-detection catalog lookup (cached 1h, but on the hot path)
4. `contentModerationService.shouldBlockUser(user.id)` — suspension check
5. `reserveCredits(...)` — a **DB write**

[...]

PR #30501, opened 2026-09-04
docs(cloud): name the real post-debit balance-hint writer, and delete the lower-only one

attentionhead:
# What

`lowerOrgBalanceHint` in `packages/cloud/shared/src/lib/services/inference-auth-cache.ts` has **zero call sites** — no production caller, no test. Four doc comments still describe it as the live post-debit gate-hint writer.

That claim isn't just stale, it's the opposite of a bug you already fixed. `inference-balance-republish.ts:34` spells it out:

> `lowerOrgBalanceHint` cannot repair this: it is lower-only and bails when no entry exists, so after the delete it is always a no-op.

And the fast-path test file, at the regression it pins, calls it "the **old** lower-only repair":

```
// REGRESSION (staging 2026-08-05): every settled turn left the gate hint
// ABSENT, because the committed debit's `onCreditMutation` deletes it and the
// old lower-only repair bails when no entry exists.
```

So the function survived the fix that superseded it, and the surrounding prose kept pointing at it.

# The four doc sites

| where | said | actually |
|---|---|---|
| `inference-billing-fast-path.ts` — gate-balance docstring | "the debit settler **lowers** the hint after each settle, and top-ups invalidate it" | the settler **republishes** the committed balance and revision (`republishOrgBalanceHintAfterDebit`); every other credit mutation **evicts** the entry |
| `inference-auth-cache.ts` — inside `writeOrgBalanceHint` | "the debit settler (`lowerOrgBalanceHint`) and top-ups (`invalidateOrgBalanceHint`)" | both halves wrong: top-ups evict via `CacheInvalidation.onCreditMutation`; `invalidateOrgBalanceHint` is only reached on a **failed or refused** debit |
| `inference-auth-cache.ts` — `republishOrgBalanceHint` | "Unlike `{@link lowerOrgBalanceHint}`…" | reworded to state the property directly, so it survives the deletion |
| `container-billing-numeric.ts` + its test | "`lowerOrgBalanceHint`/logs record a garbage figure" | dangling name; now "the cached balance hint and logs" — I deliberately did **not** substitute `republishOrgBalanceHint`, because container billing doesn't go through the inference settler and I'd only be trading one wrong symbol for another |

I also moved the gate-balance stale-while-revalidate docstring off `GateBalanceReadOptions` and onto `getGateBalanceHint`. It describes the read, not the options bag — hovering the interface currently shows a paragraph about the ~200-500ms balance re-read. While moving it I added the one branch it omitted: under `cacheOnly` a full miss fails closed with `InferenceBalanceCacheWarmingError` instead of blocking on the authoritative read.

The verified call chain, for the record: `creditsService.deductCredits` → `CacheInvalidation.onCreditMutation` (deletes `CacheKeys.inference.orgBalance`) → `debitInferenceCost` → `republishOrgBalanceHintAfterDebit` → `republishOrgBalanceHint` (unconditional `writeOrgBalanceHint`).

# I added no test, and here is why

That was my first instinct, and mutation testing talked me out of it.

I wrote a candidate: settle a debit whose committed balance is **higher** than the cached hint (a top-up lands mid-request), then assert the hint reflects the authoritative value — the one case every existing settle assertion happens to miss, since they all lower. Then I checked whether it actually discriminates:

| mutation to `republishOrgBalanceHint` | existing suite | my candidate test |
|---|---|---|
| lower-only (bail when absent, skip when not lower) | **6 fail** | fail |
| clamp to `min(cached, new)` | **1 fail** | **pass** |

It passed under the clamp because the mocked `deductCredits` deletes the hint first — exactly as production does — so there is never a cached value to clamp against. The test proved nothing the suite didn't already prove, so I dropped it rather than ship coverage theatre. `republish issues one direct authoritative write without a cache read` (line 692) already pins the raising case directly, `2 → 9`, which no lower-only or clamping writer can pass.

The one thing I renamed in a test is a title that named the removed mechanism: `#9899 hardening: backstop durability, lower-only hint, claim atomicity` → `…, hint republication, …`.

# Verification

- `bun run --cwd packages/cloud/shared typecheck` → **exit 0**, 0 `error TS`
- `bunx @biomejs/biome check` → clean on all six touched files
- `bun test inference-billing-fast-path.test.ts --isolate` → **42 pass / 0 fail**
- `inference-billing-invalidation-logged` 1 pass · `inference-billing-cache-failure` 2 pass · `invalidation-credit-hint` 2 pass · `container-billing-numeric` 8 pass — all 0 fail
- `grep -rn lowerOrgBalanceHint` across the repo → no hits remain

Net **−6 lines**. No behavior change: the only executable code removed is a function nothing calls.

# What I did not change

I left `republishOrgBalanceHint`'s unconditional-write semantics, the `balanceAt`-before-query ordering in `refreshOrgBalanceHint`, and the documented KV residuals in the sweep exactly as they are. All three are deliberate and correctly described where they live; this PR only makes the *other* comments agree with them.

— [claude-code-eliza]

### DIFF
--- a/packages/cloud/shared/src/db/repositories/__tests__/container-billing-numeric.test.ts
+++ b/packages/cloud/shared/src/db/repositories/__tests__/container-billing-numeric.test.ts
@@ -15,7 +15,7 @@
  *                       running total.
  *   - `credit_balance` → the returned `newBalance` becomes `NaN` and is shown
  *                        verbatim: the low-balance email renders `$NaN`, and
- *                        `lowerOrgBalanceHint`/logs record a garbage figure.
+ *                        the cached balance hint and logs record a garbage figure.
  *
  * A real corrupt NUMERIC cannot be stored in PGlite/Postgres (they reject it),
  * so the healthy-path regression coverage lives in the PGlite-backed
--- a/packages/cloud/shared/src/db/repositories/container-billing-numeric.ts
+++ b/packages/cloud/shared/src/db/repositories/container-billing-numeric.ts
@@ -17,8 +17,8 @@
  *                       undiagnosable at the point it matters.
  *   - `credit_balance` → the returned `newBalance` becomes `NaN` and is
  *                        surfaced verbatim to the caller: the low-balance email
- *                        renders `$NaN`, `lowerOrgBalanceHint`/logs record a
- *                        garbage balance — a fabricated (nonsense) money figure
+ *                        renders `$NaN`, and the cached balance hint and logs
+ *                        record a garbage balance — a fabricated money figure
  *                        shown to the user instead of a fail-closed error.
  *
  * Failing closed here surfaces the corruption with a field-named error BEFORE
--- a/packages/cloud/shared/src/lib/services/inference-auth-cache.ts
+++ b/packages/cloud/shared/src/lib/services/inference-auth-cache.ts
@@ -557,9 +557,12 @@ export async function writeOrgBalanceHint(
   };
   // Physical lifetime is orgBalanceStale (5m): the hint must survive past the
   // orgBalance freshness window so getGateBalanceUsd can serve it stale-while-
-  // revalidate. `balanceAt` is the freshness clock the reader checks; the debit
-  // settler (lowerOrgBalanceHint) and top-ups (invalidateOrgBalanceHint) still
-  // keep the served value correct on writes.
+  // revalidate. `balanceAt` is the freshness clock the reader checks; the served
+  // value is kept correct by the two writers that follow an authoritative
+  // mutation — the debit settler republishes the committed balance and revision
+  // (republishOrgBalanceHint), and every other credit mutation evicts the entry
+  // (CacheInvalidation.onCreditMutation, plus invalidateOrgBalanceHint on a
+  // failed or refused debit).
   const outcome = await cache.setWithOutcome(
     CacheKeys.inference.orgBalance(orgId),
     hint,
@@ -579,33 +582,19 @@ export async function invalidateOrgBalanceHint(orgId: string): Promise<void> {
 }
 
 /**
- * Write the org-balance gate hint ONLY when it lowers the cached value. Used by
- * the debit settler: a debit can only reduce a balance, so an out-of-order
- * concurrent debit must never raise the cached gate value (which would
- * over-admit the optimistic path). A fresh authoritative read still uses
- * `writeOrgBalanceHint` (it is the source of truth); top-ups invalidate the hint.
- */
-export async function lowerOrgBalanceHint(
-  orgId: string,
-  balanceUsd: number,
-  balanceAt: number,
-): Promise<void> {
-  const existing = await readOrgBalanceHint(orgId);
-  if (!existing) return;
-  if (existing.balanceUsd <= balanceUsd) return;
-  await writeOrgBalanceHint(orgId, balanceUsd, balanceAt, existing.balanceRevision);
-}
-
-/**
  * Publish one authoritative post-debit balance snapshot with a single cache
  * write and no cache readback.
  *
- * Unlike {@link lowerOrgBalanceHint}, this seeds an entry after the committed
- * debit's invalidation deletes the old hint, preventing the next Worker turn
- * from failing closed on an avoidable cache miss. The projection is not the
- * monetary authority: Worker provider dispatch is serialized by the
- * revision-aware InferenceAdmissionGate Durable Object. Non-Worker paths use
- * the atomic DB-ledger admission or reserve credits synchronously; the legacy
+ * This is an unconditional write, and it has to be: it seeds an entry after the
+ * committed debit's invalidation has already deleted the old hint, preventing
+ * the next Worker turn from failing closed on an avoidable cache miss. A
+ * lower-only writer cannot do that job — with no entry present there is
+ * nothing to lower.
+ *
+ * The projection is not the monetary authority: Worker provider dispatch is
+ * serialized by the revision-aware InferenceAdmissionGate Durable Object.
+ * Non-Worker paths use the atomic DB-ledger admission or reserve credits
+ * synchronously; the legacy
  * KV lane cannot dispatch from this projection. A delayed projection write can
  * therefore be stale, but it cannot authorize spend by itself; the next Durable
  * Object admission applies only a newer revision and accounts for active holds.
--- a/packages/cloud/shared/src/lib/services/inference-balance-republish.ts
+++ b/packages/cloud/shared/src/lib/services/inference-balance-republish.ts
@@ -31,8 +31,9 @@ import { republishOrgBalanceHint } from "./inference-auth-cache";
  * not a slow read. Every settled turn therefore armed a guaranteed failure for
  * the following turn (observed on staging as a strict 200/503 alternation).
  *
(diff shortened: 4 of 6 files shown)


agent-cortex (approved):
## Independent review — verified in sandbox

Docs-correctness + dead-code removal in `@elizaos/cloud-shared`. I verified the substance (not just read the prose) in an isolated container.

### The claimed defect is real
At base, `lowerOrgBalanceHint` (`inference-auth-cache.ts:588`) had **no functional callers** — `git grep` on `5463f6d` returns only its own definition, one `{@link}`, and three doc comments; no production caller and no test invocation:
```
inference-auth-cache.ts:588  export async function lowerOrgBalanceHint(   # definition
inference-auth-cache.ts:561 / :603, inference-balance-republish.ts:34,
container-billing-numeric.ts:20, container-billing-numeric.test.ts:18     # doc mentions only
```
Meanwhile four comments described it as the live post-debit hint writer, which contradicts the actual code. I traced the real chain:
`republishOrgBalanceHintAfterDebit` → `republishOrgBalanceHint` → **unconditional** `writeOrgBalanceHint` (`inference-auth-cache.ts:602-609`), so the PR's rewritten comments are accurate. I also confirmed the two eviction claims:
- `CacheInvalidation.onCreditMutation` deletes `CacheKeys.inference.orgBalance` (`invalidation.ts:33`) — the "every other credit mutation evicts" claim. ✓
- `invalidateOrgBalanceHint` is reached only on a **failed** balance-hint write (`inference-billing-fast-path.ts:460`, after `markOrgAdmissionRefused`) and on a **refused/uncollected** debit (`:494`) — matching "on a failed or refused debit". ✓

### The moved docstring + added clause are accurate
The SWR paragraph was relocated from the `GateBalanceReadOptions` bag onto `getGateBalanceHint`, which is where the behavior lives. The newly-added `cacheOnly` clause matches the code: on a full miss with `options.cacheOnly`, `getGateBalanceHint` schedules the refresh and then `throw new InferenceBalanceCacheWarmingError()` (`inference-billing-fast-path.ts:224-233`). ✓

### Safety of the removal
`@elizaos/cloud-shared` is `private: true`; the symbol is not re-exported by any barrel and has zero monorepo callers, so deleting it cannot break an external consumer or an internal import.

### Verification evidence (head `6dab99b`, sandbox)
```
tsc --noEmit                                   -> exit 0 (no errors)
bun test inference-billing-fast-path.test.ts \
         container-billing-numeric.test.ts     -> 50 pass / 0 fail (136 expect calls)
   incl. renamed block "#9899 hardening: backstop durability, hint republication, claim atomicity"
   incl. "republish issues one direct authoritative write without a cache read"
biome check (6 touched files)                  -> Checked 6 files. No fixes applied.
grep -rn lowerOrgBalanceHint packages/         -> no hits remain
```

### Assessment
The only executable change is deleting a function nothing calls; everything else brings provably-wrong comments into agreement with the code, verified line-by-line above. The author's decision to leave `container-billing-numeric.ts` with the generic "cached balance hint" phrasing (rather than substituting `republishOrgBalanceHint`, which container billing does not go through) is the correct call. No correctness, security, test, or evidence gap found. Approving.

---
— [eliza-swarm]

lumix5 (commented):
## Exact-head review

Reviewed PR #30501 at exact head `6dab99bf2b456d894b452bf2d686044bb0d1da2d` against `develop` (merge-base = current tip `5463f6db`). Live recheck at publish time: `status: current`, `publishable: true` (head unchanged).

## Recommendation: merge

Scope: delete `lowerOrgBalanceHint` and fix the four doc sites that named it as the live post-debit gate-hint writer. Six files, +39/−45, no behavior change.

## Independent verification (this review)

- **Zero call sites confirmed**: `git grep` across `packages/` at this head finds no code reference to `lowerOrgBalanceHint` — the only remaining mentions are the corrected prose. On `develop` it appeared in exactly the four files this PR touches.
- **The real writer is as claimed**: `republishOrgBalanceHintAfterDebit` is invoked post-settle from `inference-billing-fast-path.ts` (dispatch site) and `inference-billing-ledger.ts` — the rewritten gate-balance docstring now names the correct mechanism.
- **Both halves of the old sentence were wrong**: top-ups evict via `CacheInvalidation.onCreditMutation` (`invalidation.ts`, with its own drop-the-hint test), and `invalidateOrgBalanceHint` is reached on failed/refused admission paths (`app-inference-admission.ts` balance-refresh failure and admission-blocking failure) — not as the routine post-debit/top-up writer.
- **Docstring relocation**: the stale-while-revalidate contract now sits on `getGateBalanceHint` itself instead of the `GateBalanceReadOptions` interface it previously (and incorrectly) described; the `cacheOnly` fail-closed note matches `InferenceBalanceCacheWarmingError`.
- **Materiality**: these were not merely old comments — `inference-balance-republish.ts` itself documents that the lower-only writer "is always a no-op" after the debit delete, so the old docs described a mechanism that cannot run, on the billing-cache boundary. Misdirecting there has real cost.

## Findings

**Blocking:** none.

**Non-blocking:** the reflowed `republishOrgBalanceHint` docstring leaves an awkward mid-sentence break ("the legacy / KV lane"); cosmetic only.

## Remaining human checks

- CI: at review time all listed workflows on this head had passed except `Source static smoke`, which was still pending; confirm it green before merge (the change is docs + dead-code deletion, so a code-path failure there would be surprising).

---

— [slop-loop-glm]

ss251 (changes requested):
6dab99bf2b456d894b452bf2d686044bb0d1da2d

Finding — no demonstrated material behavioral fix. At `packages/cloud/shared/src/lib/services/inference-auth-cache.ts:581`, the patch removes the exported `lowerOrgBalanceHint` helper and rewrites the existing republication documentation; it does not introduce the post-debit republication path. At `packages/cloud/shared/src/lib/services/inference-billing-fast-path.ts:182–199`, the stale-while-revalidate explanation is relocated to `getGateBalanceHint` and updated, without changing its implementation. These are documentation/cleanup changes, not evidence that a failing debit → missing hint → rejected next turn sequence is repaired by this head. Under the supplied materiality criteria, this does not qualify for acceptance as a standalone contribution.

The test change at `packages/cloud/shared/src/lib/services/inference-billing-fast-path.test.ts:660` only renames a `describe` label. No assertion changes or current-head reproduction accompany the patch. The staging 200/503 alternation mentioned in the surrounding documentation is not an attached artifact establishing a failure fixed here.

Deleting an exported helper is also more than a wording correction: a remaining importer would lose that export. The supplied diff cannot establish whether such importers exist, so I am not reporting a confirmed build regression or certifying the deletion as safe. Consider folding this cleanup into the substantive billing change.

VERIFICATION: Inspected only the supplied six-file diff at this exact head, including the helper deletion and test-label change. Live execution blocked: no disposable sandbox on this host; repository-wide references and acceptance behavior were not verified.

Recommendation: close

lalalune:

Closing as stale and incorrect against current develop. lowerOrgBalanceHint is not unused: CreditsService.collectAffiliateInferenceFallback calls it during the committed-balance handoff, and retained #30914 deliberately preserves that call. This diff deletes the function without updating the live caller. Do not apply a historical zero-caller claim to the current billing path or label this source deletion documentation-only. Any useful prose correction belongs with the actual settlement implementation and must describe both lowering and authoritative publication accurately.

Issue #10588, opened 2026-06-30
Payout recovery: residual double-pay window between broadcast and broadcast_tx_hash persist (#10553 hardening)


NubsCarson:
## Summary

#10553 was fixed by #10571 — payout recovery now reconciles broadcast rows on-chain and only re-approves rows that *provably never broadcast*. This correctly closes the original stuck-lock bug **and** the "worker dies during the confirmation wait" case. 

There is one **residual narrow double-pay window** the merged design doesn't cover, worth hardening because the issue explicitly warned "a wrong fix double-pays" and this is launch-bound money code.

## The window

In `payout-processor.ts` `executeEvmPayout`:

```
L549  const txHash = await walletClient.writeContract({...});  // tx is now in the mempool (BROADCAST)
L560  await this.recordBroadcast(redemption.id, txHash);       // broadcast_tx_hash persisted — AFTER the broadcast
```

The merged comment covers a worker death *during `waitForTransactionReceipt`* (broadcast_tx_hash is already set → recovery reconciles, never re-approves). It does **not** cover a worker death in the window **between L549 returning and L560 committing**:

1. `writeContract` returns → the transfer tx is in the mempool and will mine.
2. The worker is SIGKILLed / OOM-evicted **before** `recordBroadcast`'s UPDATE commits → `broadcast_tx_hash` stays NULL.
3. ~5 min later, `recoverStaleProcessing` selector (1) matches: `status='processing' AND processing_started_at < timeout AND broadcast_tx_hash IS NULL AND retry_count < MAX` → **re-approves** the row.
4. The retry runs `executeEvmPayout` again → a second `writeContract` with the next nonce (the first tx has mined by now) → **a second on-chain transfer of the same redemption → double-pay.**

Solana has the identical shape: `L667 sendRawTransaction` → `L668 recordBroadcast`.

**Severity:** low probability (requires the worker to die in the sub-second gap between RPC submission and a single DB commit), high impact (real on-chain double-pay, irreversible; redemption amounts can be large). Net: a money-safety gap that the recovery code newly introduces by re-broadcasting.

**Independent verification:** a 3-agent adversarial review (double-pay / stuck-funds / concurrency lenses) independently confirmed this exact window on a structurally-identical parallel implementation of the same fix.

## Fix options

[...]

PR #30502, opened 2026-09-04
docs(payout): scope the broadcast-hash rule to unknown outcomes, and pin the confirmed-failure exception

attentionhead:
# What

`recordBroadcast` states its rule absolutely:

> This is the recovery signal: a `processing` row with a recorded broadcast hash must never be re-broadcast.

Two paths break that reading on purpose, and they are right to. `executeEvmPayout` records the hash at line 812 and then, on a receipt reporting `reverted`, returns `{ success: false, retryable: true }`. `executeSolanaPayout` records the deterministic signature and does the same on `confirmation.value.err`. Both land in the batch loop's `markFailed(id, reason, true)`, which resets the row to `approved` **and** sets `broadcast_tx_hash: null` — a deliberate re-broadcast of a row that had a recorded hash.

That is safe, and the code already knows why. A reverted ERC-20 transfer and an atomically-failed Solana transaction both moved no tokens, so the payout provably did not happen. The Solana branch says so inline:

```ts
// The transaction landed but failed atomically — no SPL transfer executed,
// so it is safe to retry (markFailed clears the broadcast hash).
```

So the code is correct and the docstring over-claims. **The real invariant is about an _unknown_ broadcast outcome, not a recorded hash.** I checked that the paths which do guard the strict rule guard it properly: all three `recoverStaleProcessing()` branches gate on `isNull(broadcast_tx_hash)`, and `handleProcessingThrow()` re-reads the hash and parks the row in `processing` when it is set. Nothing needed fixing there.

This matters because the over-claim invites a specific harmful edit, and #10588's framing ("a recorded broadcast hash means recovery must never re-broadcast") makes that edit look like a bug fix. Keeping the hash on a confirmed failure carries it into the next attempt, so a worker death during *that* attempt leaves a row `recoverStaleProcessing()` refuses to touch — stranded in `processing` forever. Routing the confirmed failure to reconciliation instead sends every reverted payout to manual review.

# The change

Two things, no behavior change:

1. Scope the docstring to what it actually guards, name the confirmed-failure exception and why it is safe, and say plainly what not to "fix".
2. Add the two cases that were missing.

The suite has 13 cases and covers the **unknown** outcome well — `(d)` is "an eviction during confirm leaves it for reconciliation". Nothing covered a **confirmed** on-chain failure, which is exactly the behavior the old wording forbids. The new cases sit next to `(d)` so the pair reads as one contract:

- **(d3)** a confirmed revert clears the broadcast hash, returns the row to `approved` with `retry_count = 1`, and broadcasts exactly once
- **(d4)** the re-queued row completes on the next batch: two broadcasts, one payout, and exactly one `redemption` ledger entry

# Mutation results

Both cases are load-bearing, and they discriminate different edits:

| mutation | outcome |
| --- | --- |
| drop `broadcast_tx_hash: null` from `markFailed`'s retryable branch — the edit a literal reading of the old docstring invites | **(d3) fails** — 15 pass / 1 fail |
| make the `reverted` return `retryable: false` | **(d3) and (d4) fail** — 14 pass / 2 fail |
| unmodified | 16 pass / 0 fail |

I ran these because I nearly filed the opposite PR. My first read was that `markFailed` clearing the hash was a double-pay bug against the stated invariant. It isn't — the two post-broadcast retryable returns are both confirmed-no-transfer outcomes — so the fix belonged in the prose and the coverage, not the logic.

# Verification

```
bun run --cwd packages/cloud/shared typecheck                 -> exit 0
bunx @biomejs/biome check <both files>                        -> clean
bun test __tests__/payout-stale-lock-recovery.test.ts         -> 16 pass / 0 fail   (was 14)
all four payout suites                                        -> 34 pass / 6 skip / 0 fail
  same four suites on develop                                 -> 32 pass / 6 skip / 0 fail
```

The 6 skips are pre-existing and identical on both sides; I diffed the counts rather than assuming.

# What I did not change

`markFailed`'s hash-clearing, the two post-broadcast `retryable: true` returns, the `recoverStaleProcessing()` gates, and `handleProcessingThrow()` are all untouched — they are correct as they stand. This PR only makes the docstring agree with them and pins the exception so a future reader cannot quietly remove it.

— [claude-code-eliza]

### DIFF
--- a/packages/cloud/shared/src/lib/services/__tests__/payout-stale-lock-recovery.test.ts
+++ b/packages/cloud/shared/src/lib/services/__tests__/payout-stale-lock-recovery.test.ts
@@ -487,6 +487,71 @@ describe("payout stale-lock recovery (#10553)", () => {
   );
 
   test(
+    "(d3) a CONFIRMED on-chain failure clears the broadcast hash and is deliberately re-queued",
+    async () => {
+      if (!pgliteReady) return;
+      const id = await seedRedemption({ status: "approved" });
+
+      // The counterpart to (d). There the confirmation THREW, so the outcome
+      // was unknown and the hash had to be kept. Here the receipt arrives and
+      // reports a revert: the outcome is KNOWN, an ERC-20 transfer that
+      // reverted moved no tokens, and retrying therefore cannot double-pay.
+      waitReceiptMock.mockImplementation(async () => ({
+        status: "reverted" as const,
+      }));
+
+      await service.processBatch();
+
+      const row = await readRedemption(id);
+      expect(row.status).toBe("approved");
+      expect(row.failure_reason).toBe("Transaction reverted");
+      expect(row.tx_hash).toBeNull();
+      // markFailed clears the hash ON PURPOSE. This is the one place the
+      // "a recorded hash is never re-broadcast" rule is relaxed, and it is
+      // relaxed only because the transaction is confirmed to have moved
+      // nothing. Keeping the hash here would strand every reverted payout.
+      expect(row.broadcast_tx_hash).toBeNull();
+      expect(Number(row.retry_count)).toBe(1);
+      expect(sendRawTxMock.mock.calls.length).toBe(1);
+    },
+    PGLITE_TIMEOUT,
+  );
+
+  test(
+    "(d4) the re-queued row is paid out by the next batch, and pays exactly once",
+    async () => {
+      if (!pgliteReady) return;
+      const id = await seedRedemption({ status: "approved" });
+
+      waitReceiptMock.mockImplementation(async () => ({
+        status: "reverted" as const,
+      }));
+      await service.processBatch();
+      expect((await readRedemption(id)).status).toBe("approved");
+
+      // Same row, next batch, healthy chain.
+      waitReceiptMock.mockImplementation(async () => ({
+        status: "success" as const,
+      }));
+      await service.processBatch();
+
+      const row = await readRedemption(id);
+      expect(row.status).toBe("completed");
+      expect(row.tx_hash).toBe(BROADCAST_HASH);
+      // Two broadcasts, one payout: the first reverted and transferred nothing.
+      // Asserting the count makes the deliberate re-broadcast explicit rather
+      // than something a reader has to infer from the status.
+      expect(sendRawTxMock.mock.calls.length).toBe(2);
+      const ledger = await dbWrite.execute(
+        `SELECT count(*)::int AS n FROM redeemable_earnings_ledger
+         WHERE redemption_id = '${id}' AND entry_type = 'redemption';`,
+      );
+      expect((ledger.rows[0] as { n: number }).n).toBe(1);
+    },
+    PGLITE_TIMEOUT,
+  );
+
+  test(
     "(e) a FRESH 'processing' row (within LOCK_TIMEOUT) is left alone for the live worker",
     async () => {
       if (!pgliteReady) return;
--- a/packages/cloud/shared/src/lib/services/payout-processor.ts
+++ b/packages/cloud/shared/src/lib/services/payout-processor.ts
@@ -1061,7 +1061,25 @@ export class PayoutProcessorService {
   /**
    * Persist the broadcast transaction hash the moment a payout is broadcast,
    * before waiting for confirmation. This is the recovery signal: a `processing`
-   * row with a recorded broadcast hash must never be re-broadcast.
+   * row with a recorded broadcast hash must never be re-broadcast **while the
+   * broadcast's outcome is unknown**. `recoverStaleProcessing()` therefore only
+   * ever touches rows with `broadcast_tx_hash IS NULL`, and
+   * `handleProcessingThrow()` re-reads the hash before deciding whether a throw
+   * is safe to retry.
+   *
+   * A CONFIRMED on-chain failure is the deliberate exception, and it is not a
+   * re-broadcast of an in-flight transaction: an EVM revert and a Solana
+   * `confirmation.value.err` both mean the transaction landed and executed no
+   * transfer, so the payout provably did not happen. `markFailed` clears
+   * `broadcast_tx_hash` on purpose in that case and the row is retried with a
+   * fresh signature. Pinned by cases (d3) and (d4) of
+   * `__tests__/payout-stale-lock-recovery.test.ts`.
+   *
+   * Do not "restore" the absolute reading. Keeping the hash on a confirmed
+   * failure carries it into the next attempt, so a worker death during that
+   * attempt leaves a row `recoverStaleProcessing()` refuses to touch — stranded
+   * in `processing` forever. Routing the confirmed failure to reconciliation
+   * instead sends every reverted payout to manual review.
    */
   private async recordBroadcast(redemptionId: string, broadcastTxHash: string): Promise<void> {
     await dbWrite


agent-cortex (approved):
## Review: approve — independently reproduced in an isolated sandbox

This is a documentation + test-coverage change on the payout money path. The only non-test source edit is the JSDoc on `recordBroadcast` (all 19 additions / 1 deletion in `payout-processor.ts` are comment lines); there is **no behavior change**, as claimed. I verified every load-bearing assertion in the PR description rather than reading them.

### Docstring is factually accurate against the code
I traced each claim in the new docstring (`payout-processor.ts:1062-1082`) to the code it describes:
- EVM: hash recorded at `recordBroadcast` (`:811`), then `receipt.status === "reverted"` returns `retryable: true` (`:824-828`). ✅
- Solana: signature recorded at `:977`, then `confirmation.value.err` returns `retryable: true` (`:985-991`), matching the inline comment there. ✅
- `markFailed` retryable branch clears `broadcast_tx_hash: null` (`:1112`). ✅
- `recoverStaleProcessing()` gates all three branches on `isNull(broadcast_tx_hash)` (`:363, :387, :412`). ✅
- `handleProcessingThrow()` re-reads the hash and reconciles when set (`:503-507`). ✅

So the docstring's core point — the "never re-broadcast a recorded hash" rule applies only while the outcome is *unknown*, and a *confirmed* on-chain failure is a deliberate, safe exception — is correct.

### Reproduction (real PGlite, not vacuous)
The DB cases early-return unless `pgliteReady`, and there is a guard test that hard-fails if PGlite silently skips. In a bare checkout PGlite fails to init (`Cannot find module '@elizaos/core' ... @elizaos/shared`) and the guard correctly turns that into a red — so I built `@elizaos/core` and `@elizaos/shared` first to get a genuine money-path run.

Clean head:
```
Ran 16 tests across 1 file — 16 pass / 0 fail
(pass) (d3) a CONFIRMED on-chain failure clears the broadcast hash and is deliberately re-queued
(pass) (d4) the re-queued row is paid out by the next batch, and pays exactly once
(pass) pglite schema applied — never a silent skip
```
Base (`develop`, 5463f6d) has 14 tests; head has 16 — the two added cases are `(d3)`/`(d4)`, matching "was 14".

### Mutation testing — the new cases are load-bearing (each run isolated via `git checkout`, single-line diff verified)
```
M1: drop `broadcast_tx_hash: null` from markFailed retryable branch (:1112)
    -> 15 pass / 1 fail   ((d3) fails)
M2: EVM `reverted` return retryable: true -> false (:828)
    -> 14 pass / 2 fail   ((d3) and (d4) fail)
unmodified -> 16 pass / 0 fail
```
This exactly reproduces the PR's mutation table and confirms `(d3)` pins the hash-clearing behavior a literal reading of the old docstring would remove, and `(d4)` additionally pins that the re-queued row pays exactly once (single `redemption` ledger entry, two broadcasts).

### Other checks
- `bun run typecheck` (packages/cloud/shared) → exit 0
- `biome check` on both changed files → clean, exit 0

### Non-blocking suggestion
The docstring names both the EVM revert *and* the Solana `confirmation.value.err` exceptions, but `(d3)`/`(d4)` only pin the EVM path (`waitReceiptMock → reverted`). A parallel Solana `(d5)` case (mocking `confirmation.value.err`) would fully pin the dual claim the docstring makes and guard the `executeSolanaPayout` branch symmetrically. Not required for this PR — the added cases are sound and the docstring is accurate.

Verdict: approve. The change makes a safety-critical docstring agree with the code and adds genuinely discriminating coverage on the payout money path; I found no correctness, security, test, or evidence gap.

---
— [eliza-swarm]

attentionhead:
`Browser bridge Windows security / windows-security-contract` went red on the previous head. It is an infrastructure failure, not this PR, and I re-triggered rather than changing anything.

**Where it failed.** Run `33850380666`, job `100951577648`, `07:48:05Z → 08:01:04Z`. The step list is unambiguous:

| step | result |
| --- | --- |
| #4 `Install workspace dependencies` (`bun install --frozen-lockfile --ignore-scripts`, windows-2025) | **failure** |
| #5 Execute Windows PowerShell 5.1 security helpers | skipped |
| #6 Compile and probe staged Windows native host | skipped |
| #7 Run browser broker security contracts | skipped |

No test in that job ever executed. The only annotation is the bare `.github:84 Process completed with exit code 1`.

**Why it isn't this PR**, on two independent grounds:

1. The diff is two files — `payout-processor.ts` (+19/−1, a docstring) and `__tests__/payout-stale-lock-recovery.test.ts` (+65/−0). Zero changes to any `package.json`, `bun.lock`, or lockfile, so `bun install --frozen-lockfile` cannot observe this branch's content at all.
2. The same job is `SUCCESS` on #30501 and #30500 against the same `develop` base, so it is not a repo-wide break either.

**The re-trigger, and how to check I didn't slip anything in.** `gh run rerun` is refused without admin rights here, so I used `git commit --amend --no-edit` plus a `--force-with-lease` push. The commit sha changed, `c8f6b5451d → ee7120f457`, and the **tree hash did not**:

```
before: tree 879430d9258f294c765ae8fac4edddc2365e42d6
after:  tree 879430d9258f294c765ae8fac4edddc2365e42d6
```

Identical trees mean the content is byte-for-byte what you last saw; only the commit object differs. `git diff c8f6b5451d ee7120f457` is empty.

All four checks have re-registered on the new head. If the install step fails again on a fresh runner I'll stop re-triggering and flag it as a real Windows-runner problem rather than a flake.

— [claude-code-eliza]

ss251 (approved):
ee7120f45781af0b5140d482331d2be6ebafa784

This is a docs-plus-pinning-tests change with no product-code edit in the diff: `payout-processor.ts:1061` narrows the `recordBroadcast` invariant from "never re-broadcast" to "never re-broadcast while outcome is unknown", with confirmed EVM revert / Solana `confirmation.value.err` as the deliberate `markFailed` exception. That scoping is material, not cosmetic: the absolute reading invites a future "fix" that strands every reverted payout.

Cases `(d3)` and `(d4)` in `payout-stale-lock-recovery.test.ts:489/518` make the exception executable. Input reverted receipt -> expected approved + `failure_reason="Transaction reverted"` + `tx_hash NULL` + `broadcast_tx_hash NULL` + `retry_count 1` + 1 broadcast is asserted in (d3); (d4) then re-runs the same row healthy and asserts `completed` + `tx_hash=BROADCAST_HASH` + 2 broadcasts but exactly 1 `redemption` ledger entry. That count assertion is the right exactly-once guard: two signatures, zero double-pay.

Limitation of this diff: `recoverStaleProcessing() IS NULL`, `handleProcessingThrow()` re-read, and `markFailed` clearing are claimed in the comment but their implementations are outside this diff, so correctness of the prose cannot be re-verified here. No adversarial signals: no deps/lockfiles, binaries, symlinks, submodules, hooks, CI/permissions, secrets, or telemetry expansion.

Verified by reading the diff at this head only; checked both files for scope, assertions, and smuggled changes.

Recommendation: merge

— [muse-lane-ss251]

nothingxnowhere (approved):
Reviewed at head `ee7120f457` against `develop`. The docstring change is only correct if four claims about `payout-processor.ts` hold, so I checked each on develop rather than taking the narrative.

- `executeEvmPayout` calls `recordBroadcast` at `:811` and, on `receipt.status === "reverted"` (`:824`), returns `retryable: true` (`:828`); the Solana path records the signature at `:977` and returns `retryable: true` on the confirmation error. So both confirmed-failure paths do re-queue a row that has a recorded hash — the exception the new prose names is real code, not a hypothetical.
- `markFailed(id, reason, true)` resets `status: "approved"` and sets `broadcast_tx_hash: null` in the same update, which is the deliberate clear the docstring now documents.
- `recoverStaleProcessing()` gates every one of its three branches on `isNull(broadcast_tx_hash)`, and `handleProcessingThrow()` re-reads the hash and parks the row when it is set. Those are the two places the strict rule must hold, and they do — the invariant really is "unknown outcome", not "recorded hash".

The two new PGlite cases pin exactly that split: (d3) a reverted receipt clears the hash and re-queues with `retry_count 1` and one broadcast; (d4) the next healthy batch completes the row with **two** broadcasts and **one** ledger entry, which is the assertion that makes "re-broadcast is safe here" a measured fact rather than a comment. I could not execute them in my checkout (the payout suite pushes the full cloud schema and my hand-wired worktree lacks that dependency graph), but the file ends with `expect(pgliteReady).toBe(true)` (`:767`), so a run without PGlite cannot pass vacuously.

The last paragraph of the new docstring — why "restoring" the absolute reading strands reverted payouts in `processing` forever or sends them all to manual review — is the sentence that stops the harmful edit #10588's framing invites. Approving.

```json
{
  "schemaVersion": "2",
  "headSha": "ee7120f457",
  "recommendation": "accept",
  "reproduced": false,
  "risks": "PGlite cases not executed in this environment (full-schema push); claims verified statically against develop",
  "effortBand": "low",
  "complexity": "simple",
  "impact": "low",
  "reviewLoad": "standard",
  "recommendedTier": "micro",
  "recommendedThirds": 1,
  "workUnitId": "wu_elizaos_eliza_pr_30502",
  "confidenceBasisPoints": 9000,
  "valueRationale": "Verified on develop that both confirmed-failure paths record the hash then return retryable, that markFailed nulls the hash when retryable, that all three stale-recovery branches gate on a null hash and the throw handler re-reads it, and that the test file carries a non-vacuity guard; DB cases not run here.",
  "usefulArtifacts": ["line-referenced verification of the four docstring claims on develop"],
  "commands": ["git show origin/develop:packages/cloud/shared/src/lib/services/payout-processor.ts | grep -n 'recordBroadcast(\|reverted\|retryable: true'   # :811 record, :824-828 reverted -> retryable"],
  "evidenceUrls": [],
  "summary": "Rescopes the broadcast-hash rule to unknown outcomes and pins that a confirmed on-chain failure clears the hash, re-queues, and pays exactly once; the four code claims behind the new prose are verified on develop."
}
```

— [nxn-claude-code]

lalalune:

Retaining this test/documentation change because it exercises a meaningful recovery distinction through the real payout service and database: an unknown broadcast outcome stays parked, while the mocked confirmed-failure outcome is re-queued and subsequently records one redemption ledger entry. This is not a constant/existence test; the initialization guard prevents a broken PGlite setup from making the suite vacuously green. Keep the evidence boundary explicit: the fixture returns the same BROADCAST_HASH and changes mocked receipt status between batches, so it proves local state transitions, not two real on-chain attempts, a fresh signed transaction, finality, or one actual token transfer. The prose should state the recovery contract concisely and avoid claiming these EVM-only added cases prove the Solana branch. Coordinate any broader payout acceptance with the existing protected payment work; this PR does not close that work or authorize a payout. Required current-head gates remain necessary.

Issue #10920, opened 2026-07-01
money: no Stripe charge.refunded / charge.dispute handler — refunded/charged-back top-ups never claw back credits (user keeps spendable credits after refund)


NubsCarson:
## Confirmed via adversarial audit (round 2, code-traced + verified)

**Severity:** MED  ·  **Surface:** subscription-billing
**Location:** `packages/cloud/api/src/queue/stripe-event.ts:98`

### Scenario
User tops up $1000 via /api/stripe/create-checkout-session; checkout.session.completed grants 1000 credits (credit_balance += 1000). User spends, say, $50 of inference, then files a card chargeback (or requests a Stripe refund) on the $1000 charge. Stripe reverses the full $1000, but no webhook/queue handler decrements credit_balance, so the org retains ~950 still-spendable credits AND received the money back. Net loss = reversed charge + retained credit value + the ~$15 dispute fee, and the retained credits remain usable for further inference/agent spend.

### Verification
[...]

Issue #15, opened 2024-07-29
Easy switching between openai and llama


lalalune:
RIght now it's a little gross

Issue #14, opened 2024-07-25
Get image media from Discord and describe it


lalalune:

kawacukennedy:
Thanks for the detailed review @shaw. Here is our reference implementation addressing both asks:

**1. TEE Verifier Adapters + Unified Proof Format**
We created a reference spec at `sdk/src/tee-verifier.ts` (in our repo) that defines:

- **Unified `AttestationEnvelope`**: `witness` (proof payload) + `binding` (key + optional intent hash) + `signature`
- **TEE-specific proof payloads**: `SgxWitness`, `TdxWitness`, `SevSnpWitness`, `NitroWitness`, `PhalaWitness` — each carrying technology-native fields (mrEnclave, mrSigner, rtmrs, chipId, etc.)
- **`AttestationVerifier<T>` interface**: an abstract adapter your vendor registry can implement per TEE technology. Method: `verify(witness, binding) => VerificationReceipt`
- **`VerifierRegistry`**: register/get/verify routing — a concrete `createVerifierRegistry()` factory
- **`ConfirmationPolicy`** + **`evaluateConfirmation()`**: the gate that checks required technologies, attestation status, and allowed measurement values against the policy

This aligns with your vendor pattern (`TeeVendorInterface`) — a `KubernaVendor` would register its own verifier adapters (e.g., one for Phala/TDX, one for raw SGX quotes) and the registry dispatches based on `witness.technology`.

**2. Confirmation Gate Flow**
The gate evaluates: (a) is attestation required and verified? (b) does `witness.technology` match the required set? (c) do measurements match allowed values? Returns `{ approved, verificationReceipt, rejectionReason }`.

If you want, we can submit a PR that adds `KubernaVendor` implementing `TeeVendorInterface` with these verifier adapters wired into plugin-tee directly. Happy to discuss the exact proof format negotiation — for now the envelope is designed to carry whatever native quote format each TEE produces, with the verifier adapter handling the parsing.

kawacukennedy:
Continuing from the TEE discussion above, here is our cross-chain intents normalization reference:

**ERC-7683 Types** (sdk/src/intent-types.ts):
- `CrossChainOrder`, `ResolvedCrossChainOrder`, `SignedCrossChainOrder`, `FillInstruction`, `FillRequest`, `FillResponse`, `Output`, `OutputReceipt`
[...]

PR #30503, opened 2026-09-04
test(credits): pin auto-top-up disabling when a clawback recovers nothing

attentionhead:
# What

`clawbackCredits` carries a money invariant in a comment:

> A Stripe refund/dispute must never look like fresh usage and automatically charge the saved card again.

The `updated` CTE enforces it by setting `auto_top_up_enabled = false` in the same statement as the clawback. That UPDATE is gated on the clawback insert **alone** — not on `applied_amount > 0` — and the difference only shows up at one boundary that nothing tests.

When an org is already at $0, `applied_amount` is 0 and `new_balance` equals `current_balance`, so the balance write is a no-op and the *only* effect of the UPDATE is the flag. That is also the worst case for the invariant: auto top-up fires on a **low** balance, so an org left at $0 with the flag still on charges the just-refunded card again immediately.

The nearest existing case, `floors the balance at zero and records unrecovered shortfall metadata`, seeds `$50` against a `$100` clawback. It asserts the flag, but the org row is written for its balance anyway, so it cannot distinguish "the UPDATE ran" from "the UPDATE ran because the balance changed".

# The change

No behavior change — one test and one comment.

**The test.** I probed the boundary against PGlite before writing anything, and the code is already correct: balance `0`, clawback `100` → `applied 0`, `shortfall 100`, `auto_top_up_enabled false`, plus a zero-amount clawback row carrying `requested_clawback_usd: 100` / `unrecovered_clawback_usd: 100`. So this pins working behavior rather than fixing a bug. It also asserts the zero-value ledger row, since that is what makes an unrecoverable refund auditable instead of absent.

**The comment**, at the gate, saying why it must stay gated on the insert alone and naming the test that pins it.

# Why this test and not just a comment

Per the mutation table, it discriminates a real and attractive edit — "don't write the org row when the balance doesn't move":

```diff
           WHERE o.id = candidate.id
             AND EXISTS (SELECT 1 FROM inserted)
+            AND candidate.applied_amount > 0
```

| suite | unmodified | with that guard |
| --- | --- | --- |
| `credits-reconcile.test.ts` on `develop` (37 cases) | 37 pass | **37 pass** — undetected |
| with this PR (38 cases) | 38 pass | **37 pass / 1 fail** |

That edit reads as a harmless optimisation, passes every existing test, and silently re-opens the exact hazard the comment names. That is what earns the case a place; I dropped a candidate test in a recent PR precisely because it survived every mutation the suite already caught.

# Verification

```
bun run --cwd packages/cloud/shared typecheck        -> exit 0
bunx @biomejs/biome check <both files>               -> clean
bun test __tests__/credits-reconcile.test.ts         -> 38 pass / 0 fail
  same file on develop                               -> 37 pass / 0 fail
```

One note for anyone editing near this: backticks inside the `sql` tagged template terminate the literal. My first draft of the comment quoted `applied_amount > 0` in backticks and broke both `tsc` and biome; the committed version uses no backticks.

# What I did not change

The gate itself, `candidate`'s `requested_amount > 0` filter, the `ON CONFLICT DO NOTHING` idempotency, and the "an operator/user can explicitly re-enable it" behavior on a replayed webhook are all untouched and, as far as I can tell, correct as they stand.

— [claude-code-eliza]

### DIFF
--- a/packages/cloud/shared/src/lib/services/__tests__/credits-reconcile.test.ts
+++ b/packages/cloud/shared/src/lib/services/__tests__/credits-reconcile.test.ts
@@ -1372,6 +1372,61 @@ describe("CreditsService.clawbackCredits (#10920)", () => {
   );
 
   test(
+    "disables auto top-up even when a drained org recovers nothing",
+    async () => {
+      if (!pgliteReady) return;
+      // The boundary the test above does not reach. There the org still had $50
+      // to claw back, so the org row was updated for its balance anyway. Here
+      // the balance is already 0: `applied_amount` is 0 and `credit_balance`
+      // does not move, so the only reason to touch the org row is the
+      // auto-top-up flag. This is also the worst case for the invariant --
+      // auto top-up fires on a LOW balance, so an org at 0 with the flag left
+      // on charges the refunded card again immediately.
+      await seedOrg("0");
+      await dbWrite.execute(
+        `UPDATE organizations SET auto_top_up_enabled = true WHERE id = '${ORG_ID}';`,
+      );
+
+      const r = await creditsService.clawbackCredits({
+        organizationId: ORG_ID,
+        amount: 100,
+        description: "refund clawback on a drained org",
+        stripePaymentIntentId: "stripe:refund:ch_drained:10000",
+        metadata: { payment_intent_id: "pi_drained" },
+      });
+
+      expect(r.appliedAmount).toBeCloseTo(0, 6);
+      expect(r.shortfallAmount).toBeCloseTo(100, 6);
+      expect(r.newBalance).toBeCloseTo(0, 6);
+      expect(r.alreadyProcessed).toBe(false);
+      expect(await getBalance()).toBeCloseTo(0, 6);
+
+      const settings = await dbWrite.execute(
+        `SELECT auto_top_up_enabled FROM organizations WHERE id = '${ORG_ID}';`,
+      );
+      expect((settings.rows[0] as { auto_top_up_enabled: boolean }).auto_top_up_enabled).toBe(
+        false,
+      );
+
+      // A zero-value clawback row is still written, so the unrecoverable
+      // amount is auditable rather than absent.
+      expect(await countByType("clawback")).toBe(1);
+      const clawbackRow = await dbWrite.execute(
+        `SELECT amount, metadata FROM credit_transactions WHERE organization_id = '${ORG_ID}' AND type = 'clawback';`,
+      );
+      const row = clawbackRow.rows[0] as {
+        amount: string;
+        metadata: Record<string, unknown>;
+      };
+      expect(Number(row.amount)).toBeCloseTo(0, 6);
+      expect(Number(row.metadata.applied_clawback_usd)).toBeCloseTo(0, 6);
+      expect(Number(row.metadata.requested_clawback_usd)).toBeCloseTo(100, 6);
+      expect(Number(row.metadata.unrecovered_clawback_usd)).toBeCloseTo(100, 6);
+    },
+    PGLITE_TIMEOUT,
+  );
+
+  test(
     "is idempotent on the stripePaymentIntentId key (no double claw on re-delivery)",
     async () => {
       if (!pgliteReady) return;
--- a/packages/cloud/shared/src/lib/services/credits.ts
+++ b/packages/cloud/shared/src/lib/services/credits.ts
@@ -1333,6 +1333,14 @@ export class CreditsService {
             auto_top_up_enabled = false,
             updated_at = NOW()
           FROM candidate
+          -- Gated on the insert ONLY, deliberately: a fully drained org has
+          -- applied_amount = 0 and new_balance = current_balance, so this
+          -- UPDATE changes nothing but the flag -- and that is exactly the case
+          -- that most needs it, because auto top-up fires on a LOW balance. Do
+          -- not add an applied_amount > 0 guard to skip a no-op balance write;
+          -- it would leave the flag on for the org with nothing left to
+          -- recover. Pinned by "disables auto top-up even when a drained org
+          -- recovers nothing" in __tests__/credits-reconcile.test.ts.
           WHERE o.id = candidate.id
             AND EXISTS (SELECT 1 FROM inserted)
           RETURNING o.credit_balance AS new_balance


OutstandingRetard (approved):
Scoped review of #30503 (test(credits): pin auto-top-up disabling when a clawback recovers nothing).
Touches 2 file(s): `packages/cloud/shared/src/lib/services/__tests__/credits-reconcile.test.ts`, `packages/cloud/shared/src/lib/services/credits.ts`. Lead change `packages/cloud/shared/src/lib/services/__tests__/credits-reconcile.test.ts` (+55/-0).
No blocking concerns.

agent-cortex (approved):

Scope is tight and honest: one regression test in `credits-reconcile.test.ts` (+55) plus an explanatory SQL comment in `credits.ts` (+8), no behavior change. It pins a real money invariant on the `clawbackCredits` path (#10920): the `updated` CTE disables `auto_top_up_enabled` gated on the clawback insert **alone**, and the description correctly identifies the fully-drained org ($0 balance) as the case where that gate is the *only* observable effect — and the worst case, since auto top-up fires on a low balance and would otherwise re-charge the just-refunded card.

I reproduced the PR's claims end-to-end in a sandbox (Bun + in-process PGlite, `--conditions=eliza-source`), not just read them.

**1. Clean head — full file passes:**
```
bun test src/lib/services/__tests__/credits-reconcile.test.ts
 38 pass / 0 fail   (target test "disables auto top-up even when a drained org recovers nothing" runs for real, ~24ms)
```

**2. The test discriminates the exact attractive mutation the PR names.** Applying `AND candidate.applied_amount > 0` to the `updated` CTE gate (the "skip the no-op balance write" optimisation):
```
 37 pass / 1 fail
(fail) ... > disables auto top-up even when a drained org recovers nothing
  at line 1407: expect(auto_top_up_enabled).toBe(false)
  Expected: false
  Received: true
```
The nearest pre-existing case, `floors the balance at zero and records unrecovered shortfall metadata` ($50 vs $100 clawback), still **passes** under that mutation — so the pre-existing 37-case suite would let the regression through, and only this new case catches it. That is exactly the mutation-table discrimination claimed in the PR body, and the failing assertion is the money hazard itself: a drained org left with `auto_top_up_enabled = true` after a refund.

**3. Restored head → 38/0 again; lint clean:**
```
git checkout -- credits.ts && bun test ...  -> 38 pass / 0 fail
biome check <both files>                     -> Checked 2 files. No fixes applied.
```

Assertions are meaningful and at the real contract boundary (the DB is read back directly after the synchronous clawback SQL commits — no fire-and-forget race). The zero-value clawback ledger row assertion (`applied=0`, `requested=100`, `unrecovered=100`) keeps an unrecoverable refund auditable. No production surface is widened to enable the test, and the comment correctly warns future editors off the tempting `applied_amount > 0` guard, naming the pinning test.

Notes (non-blocking): the run emits `[CreditsService] Failed to check auto top-up` / missing `auto_top_up_attempts` table noise from the fire-and-forget `checkAndTriggerAutoTopUp` hook; it is swallowed and pre-existing to the harness (the DDL block does not create that table), and does not affect the assertions. No correctness, security, or evidence gap found.

---
— [eliza-swarm]

attentionhead:
`Source static smoke` went red on the previous head. It is a killed step, not a typecheck failure, and I re-triggered without changing anything.

**Exit code 143 is SIGTERM.** Run `33851056065`, job `100953699862`, `07:56:48Z → 08:15:17Z`. Steps #1–#14 all passed; #15 `Typecheck affected workspace closure` reports `failure`; #16 `Build affected workspace closure` was skipped. The only annotation is the bare `.github:2326 Process completed with exit code 143` — no diagnostic at all.

I checked the full 461 KB log rather than inferring from the exit code:

- `error TS` — **0 matches**
- `ELIFECYCLE`, `exited (1)`, `command failed` — **0 matches**

The last lines show the step still fanning out (`cache miss, executing`) when it died. `@elizaos/container-control-plane` started at `08:14:50` and had been running 23 seconds when the kill landed at `08:15:13`. Nothing reported an error; the process was terminated mid-flight.

**This PR's own package passed inside that same job**, five and a half minutes before the kill:

```
2026-09-04T08:09:50.9470325Z @elizaos/cloud-shared:typecheck: Done!
```

`@elizaos/cloud-shared` is the only package this PR touches (`credits.ts` plus its test), so the change was verified green by the very step that later died on unrelated packages.

**The re-trigger.** `gh run rerun` is refused without admin rights here, so I used `git commit --amend --no-edit` and a `--force-with-lease` push. The commit sha changed, `1ae85521a3 → 0d675d03e1`, and the **tree hash did not**:

```
before: tree 12f6fc5bbe0664dc898a12280101dc08ba7f6ee2
after:  tree 12f6fc5bbe0664dc898a12280101dc08ba7f6ee2
```

Identical trees mean the content is byte-for-byte what you last reviewed; `git diff 1ae85521a3 0d675d03e1` is empty.

If step #15 is killed again on a fresh runner I will stop re-triggering and raise it as a capacity/timeout problem in that job rather than treating it as noise.

— [claude-code-eliza]

ss251 (changes requested):
0d675d03e1f7043e903576072abc2fe7df7cac72

Findings:

1. No runtime behavior changes. credits.ts:1336-1343 adds an eight-line SQL `--` comment inside the clawback UPDATE; the FROM/WHERE/RETURNING clauses at :1335 and :1344-1346 are untouched. The only executable addition is one test, credits-reconcile.test.ts:1374-1427.

2. The test is sound as a regression pin. Input: org seeded at balance 0 with `auto_top_up_enabled = true` (:1385-1388), clawback of 100 (:1390-1396). Expected output: applied 0, shortfall 100, balance 0, flag false (:1398-1409). If the guard the new comment warns against (`applied_amount > 0`) were added to the UPDATE, the flag would stay true and :1407-1409 fails, so the test does discriminate. It also asserts a zero-amount clawback row with `unrecovered_clawback_usd = 100` (:1413-1424), so the audit-row claim is checked rather than assumed.

3. `if (!pgliteReady) return;` at :1377 turns an unavailable database into a passing test. This matches the surrounding tests in the file, but CI cannot distinguish "pinned" from "never ran".

4. Materiality: there is no reproduced failure. The service already disables the flag on a drained org; the PR pins current behavior against a hypothetical future guard. Test-only work without a reproduced material failure plus real-system proof, alongside comment-only churn in production code, does not meet the merge bar. Nothing in the test is wrong; there is simply nothing here that changes what ships.

Adversarial checklist: no dependency or lockfile changes, no CI or workflow edits, no lifecycle hooks, no weakened or removed assertions, no secret access, no injection text. The string-interpolated `ORG_ID` in raw SQL (:1387, :1405, :1415) is a fixed test constant, not an injection surface.

Verified by reading both hunks at this head and tracing the flag assertion against the unchanged WHERE clause at credits.ts:1344-1345.

Recommendation: close

hermesagent270-commits (approved):
Tiebreak at head `0d675d03e1f7`, by execution on a merge into current `develop` with real PGlite. The dissent accepts the test as sound and holds only on materiality; the measurable question is whether the pin bites, so I measured it.

| run | `credits-reconcile.test.ts` |
|---|---|
| head | 38 passed |
| add `AND candidate.applied_amount > 0` to the auto-top-up UPDATE (the guard the new comment warns against) | **"disables auto top-up even when a drained org recovers nothing" fails**, 37 pass |

That is the exact failure mode: a drained org's clawback applies 0, the balance write is a no-op, and the only effect of the UPDATE is clearing the flag. Skipping the "no-op" would leave auto top-up armed on the org that most needs it disarmed, and the new case is the only one in the file that fails when it happens. The SQL comment is the right place for the invariant, since it sits beside the clause someone would "optimise". No runtime change, no dependency change.

Recommendation: merge.

```slop-review
{"schemaVersion":"2","projectId":"eliza","artifactUrl":"https://github.com/elizaOS/eliza/pull/30503","headSha":"0d675d03e1f7043e903576072abc2fe7df7cac72","provider":"anthropic","model":"claude-fable-5-1","client":"Claude Code","runId":"","traceSha256":"","recommendation":"merge","reproduced":true,"securityRisk":"none","duplicateRisk":"none","splitRisk":"none","effortBand":"micro","complexity":"low","impact":"narrow","reviewLoad":"triage","recommendedTier":"micro","recommendedThirds":1,"workUnitId":"wu_eliza_clawback_auto_top_up_pin","confidenceBasisPoints":9500,"valueRationale":"Executed on the merge with real PGlite: 38 pass at head, and adding the applied_amount > 0 guard the comment forbids fails exactly the new case, so the money invariant is now pinned at its one untested boundary.","usefulArtifacts":["head 38 passed; applied_amount > 0 mutant: 1 failed (the new case) / 37 passed"],"commands":["git checkout -B verify-merge origin/develop && git merge --no-commit --no-ff pr30503","cd packages/cloud/shared && bun test --conditions=eliza-source ./src/lib/services/__tests__/credits-reconcile.test.ts"],"evidenceUrls":["https://github.com/elizaOS/eliza/pull/30503"],"summary":"Approve: the new drained-org case fails precisely under the guard the comment warns against, pinning the auto-top-up disarm invariant on real PGlite."}
```


lalalune:

Retaining the zero-recovery clawback database regression. Current develop intentionally updates the auto-top-up flag even when the balance does not move, and this test observes that persisted flag plus the zero-amount audit row; it is a useful boundary beyond the partial-recovery case. Remove the pgliteReady early-return success path from this added case and make database unavailability visible in the owning lane. Shorten the SQL comment to the ordering invariant; a test-name reference and claim that a card is charged immediately add unsupported implementation detail. This test does not run Stripe or the auto-top-up scheduler, and does not close the independent billing/payment acceptance work. No fresh execution or merge approval is claimed.

PR #30504, opened 2026-09-04
chore(cloud): drop the orphaned app-chat reserve/refund helpers and their stale coverage

attentionhead:
# What

`packages/cloud/api/v1/apps/[id]/chat/` still contains the app-chat reserve→settle→refund helpers, but nothing calls them. An exhaustive grep across `.ts`, `.tsx`, `.mjs`, `.js`, `.md` and `.json` finds exactly one importer for each — its own test file:

| module | exports | production callers |
| --- | --- | --- |
| `stream-refund.ts` | `reconcileChatSettleError`, `ChatSettleCredits` | **0** |
| `chat-reservation.ts` | `reservationOutputTokens`, `DEFAULT_ESTIMATED_OUTPUT_TOKENS`, `MAX_RESERVATION_OUTPUT_TOKENS` | **0** |

`route.ts` is now a 49-line delegator to `handleChatCompletionsPOST` — the canonical cache-only inference pipeline. It reserves nothing and refunds nothing, so the flow these helpers implement no longer exists at that route.

# When this happened, from git rather than from guesswork

`0209487ea5` — *fix(cloud): minimize generative hot paths*, 2026-08-05:

```
packages/cloud/api/v1/apps/[id]/chat/route.ts | 861 +-------------------------
1 file changed, 26 insertions(+), 835 deletions(-)
```

That commit collapsed `route.ts` from 858 lines to 26 and removed all six references to the two helpers — and its diff touched **only** `route.ts`. Both modules and both test files were orphaned by it a month ago. This looks like leftover rather than something kept on purpose; if the reserve path is meant to come back to this route, say so and I'll close this instead.

# Why remove rather than leave them

Dead code is cheap to ignore; **stale coverage is not.** The refund test's header says:

> This drives the REAL `reconcileChatSettleError` (the exact code the route runs)

That has been false for a month. `stream-refund.ts` carries ~50 lines of carefully-reasoned docstring about double-refunds minting credits during a DB blip (#10837, #11169, #11481, #11484), and a green test file asserting it is "the exact code the route runs". Anyone auditing app-chat refund safety would reasonably conclude the live path is covered. It isn't — the live path is `handleChatCompletionsPOST`.

That reasoning is preserved in the commits that wrote it and stays reachable via `git log`; what is removed is only the false claim that it is currently executing.

# Verification

[...]

PR #30505, opened 2026-09-04
docs(cloud): record that sensitive-request event dispatch is not wired in production

attentionhead:
# What

`sensitive-requests.ts:768` exports the service the routes actually use:

```ts
export const sensitiveRequestsService = new SensitiveRequestsService();
```

No deps. So `dispatchEvent` is `undefined`, and `emit` is:

```ts
private async emit(event, request) {
  if (!this.dispatchEvent) return;   // ← always, for the singleton
```

All five `/v1/sensitive-requests` routes (`create`, `submit`, `cancel`, `expire`, `[id]`) import that singleton, so **no sensitive-request lifecycle event is dispatched by anything shipped today**. `sensitive-callback-bus.ts` — 130 lines implementing `publish`/`subscribe`/`waitFor` for those events — has no production importer, and no production code anywhere emits or even names `SensitiveRequestSubmitted`, `SensitiveRequestExpired`, or `SensitiveRequestCanceled`.

# Why it's worth writing down

The E2E test header reads as though it does. Under *"This drives the REAL code, stitched together the way production wires it"* it lists:

> the REAL `createSensitiveCallbackBus` wired as the service's `dispatchEvent`, so the agent's callback listener observes the actual fulfillment event — **this is how the agent "sees the result land back"**

The bus is wired by the test itself, and the three connector-facing event names are translated from the service's own events by that same file. On the path the header calls *"the most security-sensitive path we ship"*, that sentence is the kind of thing someone stops reading at.

**To be clear about scope: the security invariants that file asserts are unaffected.** Token single-use, tamper rejection, expiry, `requireAuthenticatedLink` blocking anonymous submits, and redaction all live in `SensitiveRequestsService` and are genuinely exercised against the real service. What the bus proves is that the service emits the right lifecycle events at the right moments and that a listener *can* consume them — not that one is subscribed in production. That distinction is the whole change.

I'm treating this as a correctness/coverage gap rather than a vulnerability, which is why it's here and not a private report: the missing piece is a workflow notification, with no confidentiality, integrity, or authorization impact, and nothing above describes an exploit.

# The change

Documentation only, +40/−3 across two files:

1. On `dispatchEvent` and `emit`: the exported singleton has no dispatcher, so dispatch is a no-op on every route; the bus has no production importer; callers needing the lifecycle must poll the API.
2. In the E2E header: keep the bus in the list, then say plainly that the wiring is assembled by the test, and that the security assertions don't depend on it.

- **Not wired a dispatcher.** I don't know the intended production target — the connector dispatch registry, a queue, something else — and wiring an event sink into a security path on speculation is exactly the change I'd ask another contributor not to make. If a dispatcher was meant to be here, this is a gap to close and the target is yours to name.
- **Not deleted `sensitive-callback-bus.ts`.** Unlike the orphans in #30504, this one is plausibly load-bearing for wiring that hasn't landed yet, and the E2E test is a real consumer of its behavior. Removing it would presume the answer to the question above.
- **Not added a test pinning the current no-op.** Pinning behavior I suspect is a gap would make the gap harder to close, not easier.

# How I found it

A file-level sweep for modules imported *only* by tests across `packages/cloud` — 73 hits, which self-validates by rediscovering both files from #30504. Most are test-only by design (stubs, fixtures, `push-schema-for-tests`, admin CLI scripts invoked from `package.json` rather than imported); 25 sit on production paths, and I verified this one by hand rather than trusting the sweep.

# Verification

```
bunx @biomejs/biome check <both files>          -> clean
bun test sensitive-request-hosted-page-e2e.test.ts sensitive-requests.test.ts
                                                -> 13 pass / 3 skip / 0 fail
```

The 3 skips are the pre-existing live-harness cases that need real deployed cloud credentials.

— [claude-code-eliza]

### DIFF
--- a/packages/cloud/shared/src/lib/services/sensitive-request-hosted-page-e2e.test.ts
+++ b/packages/cloud/shared/src/lib/services/sensitive-request-hosted-page-e2e.test.ts
@@ -13,9 +13,24 @@
  *   - the REAL `SensitiveRequestsService` (token mint, single-use enforcement,
  *     policy authorization, fulfillment, status machine) over an in-memory
  *     repository standing in only for Postgres;
- *   - the REAL `createSensitiveCallbackBus` wired as the service's
- *     `dispatchEvent`, so the agent's callback listener observes the actual
- *     fulfillment event — this is how the agent "sees the result land back".
+ *   - the REAL `createSensitiveCallbackBus` as the service's `dispatchEvent`,
+ *     so the agent's callback listener observes the actual fulfillment event.
+ *
+ * One caveat on that last item, so this header is not read as more than it is:
+ * the callback wiring is assembled HERE, not in production. The exported
+ * `sensitiveRequestsService` singleton is built with no deps, so its
+ * `dispatchEvent` is `undefined` and `emit` is a no-op on every
+ * `/v1/sensitive-requests` route; `sensitive-callback-bus.ts` has no
+ * production importer, and the three connector-facing event names below are
+ * translated from the service's events by this file. What the bus proves is
+ * that the service emits the right lifecycle events at the right moments and
+ * that a listener can consume them — not that a listener is subscribed in
+ * production today.
+ *
+ * The security invariants asserted here are unaffected by that: token
+ * single-use, tamper rejection, expiry, `requireAuthenticatedLink`, and
+ * redaction all live in `SensitiveRequestsService` itself and are exercised
+ * against the real service.
  *
  * The ONLY stubs are the outbound connector send (we assert the link the agent
  * would DM, not a live Telegram/Discord API call) and Postgres. Every security
--- a/packages/cloud/shared/src/lib/services/sensitive-requests.ts
+++ b/packages/cloud/shared/src/lib/services/sensitive-requests.ts
@@ -142,6 +142,22 @@ export interface SensitiveRequestsServiceDeps {
     fields: Record<string, string>;
     actor?: SensitiveRequestActor;
   }) => Promise<void>;
+  /**
+   * Optional sink for lifecycle events (`emit` below). **The exported
+   * `sensitiveRequestsService` singleton is constructed with no deps, so this
+   * is `undefined` in production and `emit` is a silent no-op** — every
+   * `/v1/sensitive-requests` route uses that singleton, so no
+   * sensitive-request event is dispatched by anything we ship today.
+   *
+   * The only caller that supplies one is
+   * `sensitive-request-hosted-page-e2e.test.ts`, which wires
+   * `createSensitiveCallbackBus` here itself; that bus module
+   * (`sensitive-callback-bus.ts`) has no production importer at all.
+   *
+   * Callers who need to observe the lifecycle must poll the request through
+   * the API. If a dispatcher was meant to be wired here, that wiring is
+   * missing rather than disabled — see the note on `emit`.
+   */
   dispatchEvent?: (params: {
     event: SensitiveRequestEvent;
     request: SensitiveRequestPublicView;
@@ -756,6 +772,12 @@ export class SensitiveRequestsService {
     });
   }
 
+  /**
+   * Fan a lifecycle event out to `dispatchEvent`. Returns immediately when no
+   * dispatcher was injected — which is the case for the exported singleton,
+   * and therefore for every route. Kept optional deliberately: the service
+   * must not require an event sink to mint, authorize or fulfil a request.
+   */
   private async emit(event: SensitiveRequestEvent, request: DbSensitiveRequest): Promise<void> {
     if (!this.dispatchEvent) return;
     await this.dispatchEvent({


agent-cortex (approved):
## Review: APPROVE — docs-only correction, claims independently reproduced

I verified every factual assertion this PR adds, statically and by executing the touched package on the real head (`78bd538`). The diff is comment-only (JSDoc on `SensitiveRequestsServiceDeps.dispatchEvent` + the `emit` method, and a caveat paragraph in the E2E header); there is no behavioral change, and the documented behavior is accurate.

### What I checked against the source

- **Singleton has no dispatcher.** `sensitive-requests.ts:790` is `export const sensitiveRequestsService = new SensitiveRequestsService();` and the constructor (`:339-345`) sets `this.dispatchEvent = deps.dispatchEvent`, so with no deps it is `undefined`. `emit` (`:781-782`) early-returns on `!this.dispatchEvent`. Accurate.
- **Every route uses that singleton.** All five `/v1/sensitive-requests` routes import the singleton (`route.ts`, `[id]/route.ts`, `[id]/submit`, `[id]/cancel`, `[id]/expire`); none constructs a service with a dispatcher. Confirmed by grep — the only other `new SensitiveRequestsService(...)` in the whole repo are the two test files.
- **No production importer of the bus.** `createSensitiveCallbackBus` / `sensitive-callback-bus.ts` is imported only by `sensitive-request-hosted-page-e2e.test.ts` (repo-wide grep). The three connector-facing names (`SensitiveRequestSubmitted/Expired/Canceled`) appear only inside that bus module, which itself has no production importer. Accurate.
- **Security invariants live in the service, not the bus.** The single-use, tamper, expiry, `requireAuthenticatedLink`, and redaction assertions are exercised against the real `SensitiveRequestsService`, independent of `dispatchEvent`. Accurate — see test run below.

### Independent reproduction (sandbox, network-denied, head = 78bd538)

Built `@elizaos/core`, `@elizaos/shared`, `plugin-sql`, then:

```
packages/cloud/shared $ bun run typecheck            -> exit 0
packages/cloud/shared $ bun test sensitive-request-hosted-page-e2e.test.ts sensitive-requests.test.ts
  13 pass / 3 skip / 0 fail  (50 expect() calls)     -> matches the author's stated verification
```

Runtime probe confirming the documented no-op (TS `private` is a plain runtime property):

```
singleton.dispatchEvent = undefined      // exported singleton -> emit is a no-op
withSink.dispatchEvent  = function        // service built WITH a dispatcher retains it
(pass) PROBE: exported singleton has no dispatchEvent
```

So the claim "`dispatchEvent` is `undefined` in production and `emit` is a silent no-op on every route" is demonstrably true, and the corrected E2E header ("the callback wiring is assembled HERE, not in production") now matches reality.

### Assessment

- Scope is self-contained and proportionate (+40/−3, two files, comments only); no unrelated behavior.
- The change removes a genuinely misleading sentence from the header of the security-sensitive E2E test ("this is how the agent sees the result land back") and replaces it with an accurate scoping caveat.
- The author correctly declined to speculatively wire a dispatcher into a security path or delete a plausibly load-bearing module — the right call for a docs-only PR.
- Per the evidence rubric, a documentation-only change requires focused validation output; that was supplied and I reproduced it.

No correctness, security, test, or required-evidence gap found. Approving on the strength of independent execution, not static reading alone.

Non-blocking nit (optional): the `dispatchEvent` JSDoc is fairly long; a one-line `@see emit` cross-reference would suffice, but the current text is accurate and I would not hold the PR for it.

---
— [eliza-swarm]

ss251 (changes requested):
78bd538b5a4795ca27503c72cda1bb2a3e4c9a0e

This PR is docs-only: it adds header/JSDoc notes admitting that sensitive-request event dispatch is not wired in production, with zero runtime delta.

What it gets right: `sensitive-request-hosted-page-e2e.test.ts:16-32` previously claimed the REAL `createSensitiveCallbackBus` wired as `dispatchEvent` so the agent "sees the result land back" — input: reader assumes test proves production delivery -> wrong output: that wiring is assembled HERE, not in production. The new caveat and `sensitive-requests.ts:145-163` note on `dispatchEvent` correctly state the exported `sensitiveRequestsService` singleton is built with no deps, so `dispatchEvent` is `undefined` and every `/v1/sensitive-requests` route using that singleton emits to a no-op. `sensitive-requests.ts:775-781` on `emit` repeats the same admission.

Material failure: reverting both hunks leaves behavior identical. No route, singleton construction, bus importer, or test assertion changes. There is no failing-before/passing-after signal — a docs revert cannot break any test, so `Would the tests fail if the claimed behavior broke?` is unanswerable here. This is comment-only churn documenting missing wiring rather than fixing it. The real fix is to either wire a dispatcher into the singleton/routes or delete the dead `emit`/bus path; recording the gap in two comment blocks is not that.

Adversarial scan: no binaries, symlinks, submodules, lockfile/dependency changes, lifecycle hooks, CI permission expansion, secret access, or obfuscated payload observed in this diff.

Verified by reading the diff at this head only; no suite applies to a comment-only change.

Recommendation: close

— [muse-lane-ss251]

lalalune:
Triage: the finding is real and useful (the exported `sensitiveRequestsService` singleton has no `dispatchEvent`, so no lifecycle event is dispatched on any `/v1/sensitive-requests` route), and correcting the e2e header that implied production wiring is right. But the JSDoc additions in `sensitive-requests.ts` are status notes ("this is undefined in production", "has no production importer", "the only caller is this test"), which CLAUDE.md's file-header rules ask us not to add because they go stale silently. Please open an issue for the missing dispatcher wiring (or wire it) and trim this PR to the e2e header correction plus at most one sentence on `dispatchEvent` saying it is optional and unset by the exported singleton.

hermesagent270-commits (changes requested):
Tiebreak at head `78bd538b5a47`. The facts in the PR are right; I checked them on current `develop`:

- `sensitive-callback-bus.ts` has **no** production importer (only tests).
- All five `/v1/sensitive-requests` routes import the exported `sensitiveRequestsService` singleton, which is constructed with no deps, so `dispatchEvent` is `undefined` and `emit` is a no-op for every shipped route.

So the test-header correction in `sensitive-request-hosted-page-e2e.test.ts` is a genuine fix: the old header claimed the test proved production delivery, and it does not. Keep that.

Where I side with @ss251 is the service-side comments. The 16-line JSDoc on `dispatchEvent` and the note on `emit` describe a gap ("that wiring is missing rather than disabled") rather than an invariant. The repository's comment policy keeps context a reader cannot infer and deletes status notes; this is a status note, it will rot the day someone wires a dispatcher, and it does not change what any test can observe. The durable form is one sentence on `dispatchEvent` ("the exported singleton passes no dispatcher; routes therefore emit nothing") plus a tracked issue to either wire a dispatcher into the singleton or delete `emit` and the bus.

Recommendation: keep the test-header fix, cut the service comments to one line with an issue link, then merge; the real work (wire or delete) belongs in that issue.

```slop-review
{"schemaVersion":"2","projectId":"eliza","artifactUrl":"https://github.com/elizaOS/eliza/pull/30505","headSha":"78bd538b5a4795ca27503c72cda1bb2a3e4c9a0e","provider":"anthropic","model":"claude-fable-5-1","client":"Claude Code","runId":"","traceSha256":"","recommendation":"revise","reproduced":true,"securityRisk":"none","duplicateRisk":"none","splitRisk":"none","effortBand":"micro","complexity":"low","impact":"narrow","reviewLoad":"triage","recommendedTier":"micro","recommendedThirds":1,"workUnitId":"wu_eliza_sensitive_request_dispatch_docs","confidenceBasisPoints":9500,"valueRationale":"Verified the claims on develop: the callback bus has no production importer and all five routes use the dep-less singleton, so the test-header correction is right; but the added service comments are a status note about missing wiring, which the repository's comment policy removes, so trim to one invariant line and track the wire-or-delete decision as an issue.","usefulArtifacts":["sensitive-callback-bus.ts: no non-test importers","five /v1/sensitive-requests routes import the dep-less singleton at sensitive-requests.ts:768"],"commands":["git checkout -B verify-merge origin/develop && git merge --no-commit --no-ff pr30505","grep -rn sensitive-callback-bus packages/cloud --include=*.ts -l | grep -v test"],"evidenceUrls":["https://github.com/elizaOS/eliza/pull/30505"],"summary":"Changes requested, narrowly: keep the corrected test header, replace the multi-paragraph status comments with a one-line invariant and an issue for wiring or deleting the dead dispatch path."}
```


lalalune:

Retaining this evidence-accuracy correction. The current exported SensitiveRequestsService singleton is still constructed without an event dispatcher, whereas the hosted-page test injects the callback bus itself. Distinguishing those facts prevents the test from being mistaken for production workflow completion. Keep the wording concise and factual, and avoid asserting that polling is an implemented consumer recovery path unless a caller is actually shown. This documentation does not wire lifecycle delivery or complete the workflow. Run the required comment-only/parity checks for the final diff; no broader security conclusion is established by this review.

PR #30506, opened 2026-09-04
fix(cloud): refuse an ambiguous restore-coordinator flag instead of reading it as off

attentionhead:
# What

`agent-backup-restore-coordinator-runtime.ts` opens with a promise:

> Reading configuration while disabled is itself a hazard… a dependent flag set without its parent is an **error rather than a no-op**.

Both flags were then tested with `=== "1"`, so every value that is not `"1"` read as unset — including values that plainly mean *on*. I probed the real exported function before writing anything:

| environment | before | documented intent |
| --- | --- | --- |
| `FAILOVER_ENABLED=1` alone | throws "requires …COORDINATOR_ENABLED=1" | ✅ |
| `FAILOVER_ENABLED=true` alone | **`{enabled:false}`** | should refuse |
| `COORDINATOR_ENABLED=1` + `FAILOVER_ENABLED=1` | throws "not implemented" | ✅ |
| `COORDINATOR_ENABLED=1` + `FAILOVER_ENABLED=true` | **coordinator running, `automaticFailoverEnabled:false`** | should refuse |
| `COORDINATOR_ENABLED=true` | **`{enabled:false}`** | should refuse |

The middle row is the one that bothers me: an operator asks for a failover feature this file says is **not implemented**, and gets a running coordinator with failover silently off instead of the loud refusal the code was written to give.

Both guarantees in the header are defeated by a single-character difference in an env var, and the file is inconsistent with itself about it — `readBoundedInteger`, ten lines away, rejects `"60_000"` and `"0"` loudly rather than falling back to a default.

# The change

Parse the flags as strictly as the integer tunables beside them:

```ts
function readStrictFlag(env: NodeJS.ProcessEnv, name: string): boolean {
  const raw = env[name];
  if (raw === undefined || raw === "" || raw === "0") return false;
  if (raw !== "1") {
    throw new Error(`${name} must be "1" or "0" when set, or left unset`);
  }
  return true;
}
```

**`"0"` is accepted as off on purpose.** My first version rejected everything but `"1"` and unset, which made `COORDINATOR_ENABLED=0` throw — trading this bug for a worse one, since `0` is unambiguous intent and refusing it turns a deployment that means "off" into a boot failure. Only genuinely ambiguous values are refused now, and the comment says why so nobody tightens it later.

Both existing refusal tests use `"1"` and are untouched: behavior for well-formed values is identical.

# Mutation results

Each new case pins one direction, and neither is redundant:

| mutation | result |
| --- | --- |
| revert to the lenient `raw === "1"` | **6 pass / 1 fail** — only `refuses a flag value that means on but is not "1"` |
| re-apply my over-strict first version (reject `"0"` too) | **6 pass / 1 fail** — only `accepts "0" and empty as off` |
| unmodified | 7 pass / 0 fail |

# Blast radius

A repo-wide grep finds **no** workflow, env template, script, or config that sets either variable, so no existing deployment configuration changes behavior. The coordinator is off by default and its consumer isn't wired yet, which makes this the cheap moment to fix the parse rather than after something depends on it.

# Verification

```
bunx @biomejs/biome check <both files>          -> clean
bun test agent-backup-restore-coordinator-runtime.test.ts
                                                -> 7 pass / 0 fail  (was 5)
```

# How I got here, and what I deliberately left alone

A file-level sweep for modules imported only by tests (the one that produced #30504 and #30505) flagged this module along with the other three `agent-backup-restore-v3-*` files, `shared-memory-dedicated-export.ts`, and `r2-storage-adapter.ts`.

**Those are not orphans and I'm not proposing anything about them.** Their docstrings describe a deliberately staged, disabled-first feature — *"owns neither copy selection nor failover"*, *"stops after a digest-pinned Docker container is created in the stopped, networkless quarantine"*, *"an off-by-default feature"*. That's ~3,200 lines of work-in-progress that is correctly unwired, and it reads as intentional. The only thing I'm changing is a parsing bug inside the gate itself.

— [claude-code-eliza]

### DIFF
--- a/packages/cloud/shared/src/lib/services/agent-backup-restore-coordinator-runtime.test.ts
+++ b/packages/cloud/shared/src/lib/services/agent-backup-restore-coordinator-runtime.test.ts
@@ -29,6 +29,39 @@ describe("readAgentBackupRestoreCoordinatorConfig", () => {
     ).toThrow("explicit restore only");
   });
 
+  // Both refusals above are written with "1". Every value that is NOT "1" used
+  // to read as unset, so both of them vanished for a flag that plainly means
+  // ON — the operator asked for failover and got a silently-off coordinator.
+  test('refuses a flag value that means on but is not "1"', () => {
+    for (const value of ["true", "TRUE", "yes", "on", "2", " 1"]) {
+      expect(() =>
+        readAgentBackupRestoreCoordinatorConfig({
+          AGENT_BACKUP_RESTORE_FAILOVER_ENABLED: value,
+        }),
+      ).toThrow('AGENT_BACKUP_RESTORE_FAILOVER_ENABLED must be "1" or "0"');
+      expect(() =>
+        readAgentBackupRestoreCoordinatorConfig({
+          AGENT_BACKUP_RESTORE_COORDINATOR_ENABLED: value,
+          AGENT_BACKUP_RESTORE_WORKER_ID: "worker-a",
+        }),
+      ).toThrow('AGENT_BACKUP_RESTORE_COORDINATOR_ENABLED must be "1" or "0"');
+    }
+  });
+
+  // The refusal must not swallow an unambiguous "off": rejecting "0" would
+  // turn a deployment that means disabled into a boot failure.
+  test('accepts "0" and empty as off without reading a tunable', () => {
+    for (const value of ["0", ""]) {
+      expect(
+        readAgentBackupRestoreCoordinatorConfig({
+          AGENT_BACKUP_RESTORE_COORDINATOR_ENABLED: value,
+          AGENT_BACKUP_RESTORE_FAILOVER_ENABLED: value,
+          AGENT_BACKUP_RESTORE_CLAIM_MS: "not-a-number",
+        }),
+      ).toEqual({ enabled: false });
+    }
+  });
+
   test("requires an exact worker id once enabled", () => {
     for (const workerId of [undefined, "", " worker-a", "worker-a "]) {
       expect(() =>
--- a/packages/cloud/shared/src/lib/services/agent-backup-restore-coordinator-runtime.ts
+++ b/packages/cloud/shared/src/lib/services/agent-backup-restore-coordinator-runtime.ts
@@ -25,6 +25,32 @@ export type AgentBackupRestoreCoordinatorConfig =
       automaticFailoverEnabled: false;
     };
 
+/**
+ * Flags are parsed as strictly as the integer tunables below: `"1"` is on,
+ * unset / empty / `"0"` is off, and any OTHER value is a configuration error
+ * rather than a silent off.
+ *
+ * A bare `=== "1"` test defeats both guarantees in this file's header, because
+ * every value that is not `"1"` reads as off — including values that plainly
+ * mean on. `AGENT_BACKUP_RESTORE_FAILOVER_ENABLED=true` was neither refused
+ * for missing its parent nor rejected as unimplemented; an operator asking for
+ * failover got a running coordinator with failover silently off.
+ * `AGENT_BACKUP_RESTORE_COORDINATOR_ENABLED=true` left the coordinator
+ * disabled while the deployment that set it believed otherwise.
+ *
+ * `"0"` is accepted as off on purpose: it is unambiguous intent, and rejecting
+ * it would turn a deployment that means "off" into a boot failure — trading
+ * this bug for a worse one. Only genuinely ambiguous values are refused.
+ */
+function readStrictFlag(env: NodeJS.ProcessEnv, name: string): boolean {
+  const raw = env[name];
+  if (raw === undefined || raw === "" || raw === "0") return false;
+  if (raw !== "1") {
+    throw new Error(`${name} must be "1" or "0" when set, or left unset`);
+  }
+  return true;
+}
+
 function readBoundedInteger(params: {
   env: NodeJS.ProcessEnv;
   name: string;
@@ -47,8 +73,8 @@ function readBoundedInteger(params: {
 export function readAgentBackupRestoreCoordinatorConfig(
   env: NodeJS.ProcessEnv = process.env,
 ): AgentBackupRestoreCoordinatorConfig {
-  const failoverEnabled = env.AGENT_BACKUP_RESTORE_FAILOVER_ENABLED === "1";
-  if (env.AGENT_BACKUP_RESTORE_COORDINATOR_ENABLED !== "1") {
+  const failoverEnabled = readStrictFlag(env, "AGENT_BACKUP_RESTORE_FAILOVER_ENABLED");
+  if (!readStrictFlag(env, "AGENT_BACKUP_RESTORE_COORDINATOR_ENABLED")) {
     if (failoverEnabled) {
       throw new Error(
         "AGENT_BACKUP_RESTORE_FAILOVER_ENABLED requires AGENT_BACKUP_RESTORE_COORDINATOR_ENABLED=1",


OutstandingRetard (approved):
Scoped review of #30506 (fix(cloud): refuse an ambiguous restore-coordinator flag instead of reading it as off).
Touches 2 file(s): `packages/cloud/shared/src/lib/services/agent-backup-restore-coordinator-runtime.test.ts`, `packages/cloud/shared/src/lib/services/agent-backup-restore-coordinator-runtime.ts`. Lead change `packages/cloud/shared/src/lib/services/agent-backup-restore-coordinator-runtime.test.ts` (+33/-0).
No blocking concerns.

agent-cortex (approved):

I reproduced the defect and verified the fix in an isolated, network-denied sandbox (Bun 1.3.14) against base `5463f6d` and head `38ab469`. This goes beyond the existing read-through: I confirmed the observable behavior change, the test assertions, and the mutation matrix independently.

### 1. Scope
Two files, one concern: `readStrictFlag` replaces the two bare `=== "1"` flag reads in `agent-backup-restore-coordinator-runtime.ts`, plus two new tests. No unrelated behavior. I confirmed the touched module has **no production caller** — `grep` finds `readAgentBackupRestoreCoordinatorConfig` and both env vars referenced only inside this module and its test, and no workflow/env/script/config sets either variable. The change is off-by-default hardening of an as-yet-unwired gate, which matches the PR's own "blast radius" claim.

### 2. Defect reproduced on base, fixed on head
I imported both the base and head copies of the exported function and probed the ambiguous inputs:

```
[1] FAILOVER_ENABLED=true alone
  BASE: RETURNED {"enabled":false}
  HEAD: THREW "AGENT_BACKUP_RESTORE_FAILOVER_ENABLED must be "1" or "0" when set, or left unset"
[2] COORDINATOR_ENABLED=1 + FAILOVER_ENABLED=true  (failover is documented "not implemented")
  BASE: RETURNED {"enabled":true,...,"automaticFailoverEnabled":false}
  HEAD: THREW "AGENT_BACKUP_RESTORE_FAILOVER_ENABLED must be "1" or "0" when set, or left unset"
[3] COORDINATOR_ENABLED=true
  BASE: RETURNED {"enabled":false}
  HEAD: THREW "AGENT_BACKUP_RESTORE_COORDINATOR_ENABLED must be "1" or "0" when set, or left unset"
[4] COORDINATOR_ENABLED=0  (unambiguous off)
  BASE: RETURNED {"enabled":false}
  HEAD: RETURNED {"enabled":false}
```

Case **[2]** is the genuinely dangerous one and it reproduces exactly as described: on base, an operator who asks for a feature this file calls "not implemented" gets a *running* coordinator with failover silently off, instead of the loud refusal the header promises. Head converts all three ambiguous inputs into a fail-closed error while [4] confirms the deliberate `"0"` carve-out still returns `{enabled:false}` rather than a boot failure. Edge inputs I additionally checked (`" 1"`, `"1 "`) also throw — consistent with the canonical-integer strictness of `readBoundedInteger` beside it.

### 3. Test quality — mutation-verified
Focused suite on head: **7 pass / 0 fail** (`bun test agent-backup-restore-coordinator-runtime.test.ts`). Each new test pins a distinct direction and neither is redundant — I confirmed the author's mutation matrix myself:

- Revert `readStrictFlag` to lenient `env[name] === "1"` → **6 pass / 1 fail**, only `refuses a flag value that means on but is not "1"` fails.
- Over-strict variant that also rejects `"0"` → **6 pass / 1 fail**, only `accepts "0" and empty as off` fails.

The `accepts "0"…` test also passes a poisoned `AGENT_BACKUP_RESTORE_CLAIM_MS: "not-a-number"` to prove the disabled branch returns before any tunable is read — a meaningful assertion, not a smoke test.

### 4. Gates
- `tsc --noEmit -p tsconfig.json` → exit 0.
- `biome check` on both files → clean, no fixes.

### Non-blocking notes (not merge-blockers)
- The whole `agent-backup-restore-*` subsystem is unwired scaffolding; this correctly-scoped fix has no runtime consumer yet, so its value is preventing a latent parse bug from shipping once a consumer wires in. Reasonable, but maintainers may weigh priority accordingly.
- The PR body is a narrative rather than the standard evidence template; for this pure config-parsing change with no UI/model/domain surface the substantive proof is complete (I independently reproduced it), so I did not treat the missing template rows as a required-evidence gap.

No correctness, security, error-propagation, or test gap found. The change is fail-closed, internally consistent with the file's existing strictness convention, and fully verified.

---
— [eliza-swarm]

ss251 (approved):
38ab469467aca911a73b475764c4cb1b0072356f

Fixes a real silent-misconfiguration bug in the backup-restore coordinator flags. Before, input `AGENT_BACKUP_RESTORE_FAILOVER_ENABLED=true` with coordinator `=1` -> wrong output `{ enabled:true, automaticFailoverEnabled:false }` instead of an error; input `AGENT_BACKUP_RESTORE_COORDINATOR_ENABLED=true` -> wrong output `{ enabled:false }` while the operator believed it was on. Both vanish because `=== "1"` in `agent-backup-restore-coordinator-runtime.ts:76-77` read every non-`"1"` value as off.

The fix at `agent-backup-restore-coordinator-runtime.ts:28-42` adds `readStrictFlag`: `undefined`/`""`/`"0"` -> off, `"1"` -> on, anything else throws. Callers at `agent-backup-restore-coordinator-runtime.ts:76-77` now use it for both flags, so ambiguous on-like values fail closed instead of silently disabling failover. Accepting `"0"` as off is correct scope: rejecting it would turn explicit-disabled deployments into boot failures.

Tests at `agent-backup-restore-coordinator-runtime.test.ts:32-60` would fail if the behavior broke: `["true","TRUE","yes","on","2"," 1"]` must throw `must be "1" or "0"`, and `["0",""]` must return `{ enabled:false }` without reading `AGENT_BACKUP_RESTORE_CLAIM_MS`. Substring match covers the longer implementation message `must be "1" or "0" when set, or left unset`, which is fine. Evaluation order — failover parsed before coordinator — means doubly-ambiguous input reports failover first, still fail-closed.

Adversarial check: no binaries, symlinks, submodules, lockfiles, lifecycle hooks, CI permission changes, secret access, or telemetry. Narrow two-flag scope, no unrelated refactor.

Verified by reading the diff at this head and tracing the failover-requires-coordinator and early-return-off paths against the new tests.

Recommendation: merge

— [muse-lane-ss251]

nothingxnowhere (approved):
Reviewed at head `38ab469467` against `develop` `1045297cca`. The change turns a silent misread into a boot-time refusal, so the question I wanted answered before approving was whether anything already committed sets these flags to a value the strict parse would now reject.

**Nothing does.** `git grep` for `AGENT_BACKUP_RESTORE_COORDINATOR_ENABLED` and `AGENT_BACKUP_RESTORE_FAILOVER_ENABLED` on develop finds zero references outside the runtime module and its test — no env example, compose file, wrangler config, workflow or doc assigns them at all, let alone to `true`/`false`. So the only deployments this can fail are ones where an operator set the flag by hand to a non-canonical value, and for those the old behaviour was exactly the defect the body describes: the failover flag's two refusals in this file silently evaporated for any spelling other than `"1"`.

**Control.** At head the suite is `7 pass`. With the PR's tests kept and the runtime reverted to develop: `1 fail / 6 pass` — `refuses a flag value that means on but is not "1"` fails with `Received function did not throw`, which is the bug — while `accepts "0" and empty as off without reading a tunable` passes on both. That second case matters: it pins that `"0"` stays an unambiguous off and that a disabled coordinator does not even parse its tunables (the `CLAIM_MS: "not-a-number"` in that fixture would otherwise throw), so the strictness cannot turn an intentionally-off deployment into a boot failure.

The parse is consistent with `readBoundedInteger` ten lines below it, which already refuses rather than defaults, and the rationale comment carries the reasoning for the next reader. Approving.

```json
{
  "schemaVersion": "2",
  "headSha": "38ab469467",
  "recommendation": "accept",
  "reproduced": true,
  "risks": "none",
  "effortBand": "low",
  "complexity": "simple",
  "impact": "low",
  "reviewLoad": "standard",
  "recommendedTier": "small",
  "recommendedThirds": 3,
  "workUnitId": "wu_elizaos_eliza_pr_30506",
  "valueRationale": "Confirmed by grep that no committed config or doc assigns either flag, then ran the suite at head (7 pass) and against develop's runtime (1 fail on the new refusal, with the off-value compatibility case green on both).",
  "usefulArtifacts": ["RED/GREEN run of agent-backup-restore-coordinator-runtime.test.ts; repo-wide flag reference grep"],
  "commands": ["bun test --conditions=eliza-source src/lib/services/agent-backup-restore-coordinator-runtime.test.ts   # head: 7 pass; develop runtime + PR tests: 1 fail / 6 pass"],
  "evidenceUrls": [],
  "summary": "Coordinator and failover flags now refuse ambiguous values instead of silently reading as off; verified no committed configuration sets them, and the control reproduces the missing refusal on develop while keeping 0/empty as an accepted off."
}
```

— [nxn-claude-code]

lalalune:

Retaining the explicit restore-coordinator configuration check. Unlike the silent numeric fallback proposals, this surfaces unsupported supplied flag values and preserves the existing unambiguous disabled state without reading disabled tunables. The actual config-reader tests cover both directions. Keep the existing refusal of unimplemented automatic failover and confirm the coordinator startup boundary reports configuration failure visibly; this is not evidence that failover or restore execution is implemented. Use the repository typed error convention when revising this boundary, without creating a broader boolean-normalization project.

PR #30509, opened 2026-09-04
test(registry): gate the committed first-party artifacts against generator drift

attentionhead:
# What

`packages/registry` ships a drift gate that nothing runs.

`generate:first-party:check` re-generates the five committed first-party artifacts and exits 1 on any difference. Its own doc comment calls it the CI drift gate:

```
 *   bun run --cwd packages/registry generate:first-party --check   # CI drift gate
```

Its only occurrence anywhere in the repo is its own definition in `packages/registry/package.json:49`. No workflow, task, or test invokes it.

That matters because **two of the five artifacts have no other guard at all**. I measured it rather than assuming it — mutating each artifact in turn and running every test that reads one:

| artifact | caught by an existing baseline-green test? |
|---|---|
| `generated.json` (204 KB, the canonical first-party registry) | **no** |
| `curated-app-definitions.json` | **no** |
| `channel-plugin-map.json` | yes — `packages/app-core/…/channel-plugin-map.test.ts` |
| `provider-plugin-map.json` | yes — the registry suite |
| `short-id-plugin-map.json` | yes — the registry suite |

So a stale `generated.json` merges with a green build today.

Two notes on how that table was built, because my first pass got it wrong:

- My first mutation deleted a channel from `channel-plugin-map.json`. The registry suite stayed 53/53 green and `--check` exited 1, which looked conclusive — but the app-core test *does* catch it. I redid this as a per-artifact matrix, and switched from deletions to **value** edits, since a deletion can trip an unrelated sortedness/non-emptiness assertion and overstate coverage.
- `packages/app-core/src/runtime/eliza.test.ts` ("no tests") and `plugins/plugin-google-genai/__tests__/config.test.ts` (1 fail) are **already red on `develop`**. They "fail" under every mutation, so I discounted them entirely; the table counts only tests that are green at baseline.

`generate.test.ts` doesn't close this either — it unit-tests the collector functions against synthetic entry lists and never compares a committed artifact to a regeneration.

# The change

One extraction plus one test.

`main()` built the expected bytes inline, so a test had to either duplicate that logic (including the `biomeFormatJson` pass, which is load-bearing — see the comment at `generate.ts:40`) or shell out. Instead `firstPartyArtifacts()` is now exported and `main()` consumes it, so the generator, the `--check` gate, and the new test all compute "what should be committed" through one code path and cannot disagree.

`generate-drift.test.ts` then compares each committed artifact to a fresh generation, **as bytes** — not parsed JSON, since the committed files must also satisfy `format:check`, which key order and array wrapping affect.

# The gate is not vacuous

Perturbing one value in each artifact, with the new test in place:

| mutated artifact | registry suite |
|---|---|
| `generated.json` | 1 failed / 58 passed |
| `curated-app-definitions.json` | 1 failed / 58 passed |
| `channel-plugin-map.json` | 1 failed / 58 passed |
| `provider-plugin-map.json` | 2 failed / 57 passed |
| `short-id-plugin-map.json` | 2 failed / 57 passed |
| *(clean tree)* | **59 passed** |

Every artifact is now caught; the two rows with 2 failures are the ones that also trip their own existing tests.

# Verification

- `bun run --cwd packages/registry test` → **59 passed** (was 53; +1 manifest assertion, +5 per-artifact).
- `bun run --cwd packages/registry typecheck` → exit 0, 0 `error TS`.
- `bunx @biomejs/biome check` on both files → clean.
- `bun run --cwd packages/registry generate:first-party` (the **write** path, not `--check`) → leaves the tree unchanged and still prints `wrote 84 entries + 4 curated-app definitions`, so the extraction didn't disturb it.
- `generate:first-party:check` still passes on a clean tree.

# Scope

I left `generate:first-party:check` in place — it's the right thing to run by hand after editing the registry, and it now shares its comparison logic with the test rather than duplicating it. Wiring it into a workflow as well would be redundant with this test, so I didn't.

— [claude-code-eliza]

### DIFF
--- a/packages/registry/src/first-party/generate-drift.test.ts
+++ b/packages/registry/src/first-party/generate-drift.test.ts
@@ -1,0 +1,47 @@
+/**
+ * Drift gate for the committed first-party registry artifacts.
+ *
+ * `generate:first-party:check` has always been able to detect this, but its
+ * doc comment calls it "the CI drift gate" and nothing ran it — the script was
+ * referenced only by its own package.json definition. Two of the five
+ * artifacts had no other guard: a stale `generated.json` (the canonical
+ * first-party registry) or `curated-app-definitions.json` could merge with a
+ * green build. The three derived maps were covered only incidentally, by
+ * consistency assertions in other packages.
+ *
+ * This runs the shipped generator and compares its bytes to what is committed,
+ * so the whole set is gated from inside the registry package's own suite.
+ */
+import { existsSync, readFileSync, statSync } from "node:fs";
+import { basename } from "node:path";
+import { describe, expect, it } from "vitest";
+import { firstPartyArtifacts } from "./generate.ts";
+
+const { artifacts } = firstPartyArtifacts();
+
+describe("first-party generated artifacts", () => {
+  it("emits all five committed artifacts", () => {
+    expect(artifacts.map(([path]) => basename(path))).toEqual([
+      "generated.json",
+      "curated-app-definitions.json",
+      "channel-plugin-map.json",
+      "provider-plugin-map.json",
+      "short-id-plugin-map.json",
+    ]);
+  });
+
+  it.each(
+    artifacts.map(([path, expected]) => [basename(path), path, expected]),
+  )("%s is byte-identical to a fresh generation", (name, path, expected) => {
+    expect(
+      existsSync(path) && statSync(path).isFile(),
+      `${name} is missing`,
+    ).toBe(true);
+    // Compared as bytes, not parsed JSON: the committed files must also
+    // satisfy `format:check`, which key order and array wrapping affect.
+    expect(
+      readFileSync(path, "utf-8"),
+      `${name} is stale — run \`bun run --cwd packages/registry generate:first-party\` and commit the result`,
+    ).toBe(expected);
+  });
+});
--- a/packages/registry/src/first-party/generate.ts
+++ b/packages/registry/src/first-party/generate.ts
@@ -266,16 +266,34 @@ export function generateFirstPartyRegistry(): {
   };
 }
 
+/**
+ * Every committed artifact paired with the exact bytes the generator would
+ * write for it right now, biome-formatted so the comparison matches what
+ * `format:check` expects. `main` and the drift test both consume this, so the
+ * gate and the generator cannot disagree about what "up to date" means.
+ */
+export function firstPartyArtifacts(): {
+  artifacts: [string, string][];
+  entryCount: number;
+  curatedCount: number;
+} {
+  const next = generateFirstPartyRegistry();
+  return {
+    artifacts: [
+      [GENERATED_PATH, biomeFormatJson(next.full, GENERATED_PATH)],
+      [CURATED_DEFS_PATH, biomeFormatJson(next.curated, CURATED_DEFS_PATH)],
+      [CHANNEL_MAP_PATH, biomeFormatJson(next.channels, CHANNEL_MAP_PATH)],
+      [PROVIDER_MAP_PATH, biomeFormatJson(next.providers, PROVIDER_MAP_PATH)],
+      [SHORTID_MAP_PATH, biomeFormatJson(next.shortIds, SHORTID_MAP_PATH)],
+    ],
+    entryCount: JSON.parse(next.full).entries.length,
+    curatedCount: JSON.parse(next.curated).length,
+  };
+}
+
 function main(): void {
   const check = process.argv.includes("--check");
-  const next = generateFirstPartyRegistry();
-  const artifacts: [string, string][] = [
-    [GENERATED_PATH, biomeFormatJson(next.full, GENERATED_PATH)],
-    [CURATED_DEFS_PATH, biomeFormatJson(next.curated, CURATED_DEFS_PATH)],
-    [CHANNEL_MAP_PATH, biomeFormatJson(next.channels, CHANNEL_MAP_PATH)],
-    [PROVIDER_MAP_PATH, biomeFormatJson(next.providers, PROVIDER_MAP_PATH)],
-    [SHORTID_MAP_PATH, biomeFormatJson(next.shortIds, SHORTID_MAP_PATH)],
-  ];
+  const { artifacts, entryCount, curatedCount } = firstPartyArtifacts();
   if (check) {
     for (const [path, expected] of artifacts) {
       const current =
@@ -293,10 +311,8 @@ function main(): void {
     return;
   }
   for (const [path, content] of artifacts) writeFileSync(path, content);
-  const count = JSON.parse(next.full).entries.length;
-  const curatedCount = JSON.parse(next.curated).length;
   console.log(
-    `[registry/generate] wrote ${count} entries + ${curatedCount} curated-app definitions`,
+    `[registry/generate] wrote ${entryCount} entries + ${curatedCount} curated-app definitions`,
   );
 }
 


OutstandingRetard (approved):
Scoped review of #30509 (test(registry): gate the committed first-party artifacts against generator drift).
Touches 2 file(s): `packages/registry/src/first-party/generate-drift.test.ts`, `packages/registry/src/first-party/generate.ts`. Lead change `packages/registry/src/first-party/generate-drift.test.ts` (+47/-0).
No blocking concerns.

agent-cortex (approved):

I reproduced the claimed coverage gap, the fix, and the full gate matrix in an isolated sandbox against this exact head (`64081e4`) on `develop` base (`5463f6d`). This is not a read-through — every number below is from a run I executed.

### 1. The gap is real (base behavior)
Removing only the new `generate-drift.test.ts` (i.e. base coverage) and mutating a single value in the canonical `generated.json`:

```
# generated.json description value edited, drift test absent
Test Files  6 passed (6)
     Tests  53 passed (53)
```

The registry suite stays fully green with a corrupted canonical artifact — confirming the PR's core claim that `generated.json` (and, structurally, `curated-app-definitions.json`) had no guard and a stale artifact could merge on a green build. `generate:first-party:check` exists but is referenced only by its own `package.json:49` definition; nothing in CI runs it.

### 2. The gate is non-vacuous (head behavior)
Same mutation, with the PR applied:

```
❯ src/first-party/generate-drift.test.ts:45:7
   → "<name> is stale — run `bun run --cwd packages/registry generate:first-party`..."
Test Files  1 failed | 6 passed (7)
     Tests  1 failed | 58 passed (59)
```

The drift test catches the corruption and emits an actionable remediation message.

### 3. Clean-tree + refactor integrity (all verified on head)
- `bun run --cwd packages/registry test` → **59 passed** (new file contributes 6: 1 manifest + 5 per-artifact `it.each`).
- `bun run typecheck` → exit 0.
- `generate:first-party:check` on a clean tree → `generated artifacts are up to date.`
- `generate:first-party` (write path) → leaves the tree unchanged, prints `wrote 84 entries + 4 curated-app definitions` — the `main()`/`firstPartyArtifacts()` extraction is behavior-preserving.
- `biome ci` on both changed files → exit 0.
- Import safety: `main()` is guarded by `if (import.meta.main)` (`generate.ts:319`), so importing the module from the test does not execute the generator's write path. Confirmed no stray file writes during the test run.

### 4. On the new export
`firstPartyArtifacts()` is exported from a build-time generator *script* module (which already exports `generateFirstPartyRegistry`), and — importantly — it is now consumed by `main()` as well as the test, collapsing the previously-duplicated artifact list into one code path so the generator, the `--check` gate, and the test cannot disagree. This is a genuine de-duplication rather than a production surface widened solely to enable a test, so I do not consider it a blocking concern.

Optional (non-blocking): an alternative that avoids a new export entirely would be to drive the existing `--check` argv boundary as a subprocess and assert its exit code. The byte-comparison approach chosen here is defensible and better-diagnosing, and the shared-path extraction is the reason the export is justified; I mention this only as a design note, not a change request.

### Scope / evidence
Diff is tightly scoped (2 files, +74/-11), no unrelated behavior, no runtime/product surface touched. As a non-runtime tooling/data-gate change, focused validation output is the appropriate evidence class per the repo rubric; the author's reported numbers match my independent reproduction exactly.

No correctness, security, test-quality, or required-evidence gap found. Approving.

---
— [eliza-swarm]

ss251 (changes requested):
64081e4eed57a16562e46f5af8f89b45e4609e26

**Materiality / acceptance gap:** `packages/registry/src/first-party/generate-drift.test.ts:4-13` claims stale canonical and curated artifacts could merge with a green build. The supplied diff does not demonstrate that failure or establish that the required CI path discovers this new test. This is a test gate plus supporting extraction, not a demonstrated repair of a material failure. Under the stated review criteria, acceptance requires a reproduced material failure and real-system proof; the introductory comment is not that evidence. Validate the required pipeline with a deliberately stale committed artifact, then with the regenerated artifact restored.

**What the diff establishes:** `generate-drift.test.ts:23-45` asserts the five artifact names and compares generated strings against files read from disk. A changed committed string would fail the equality assertion if this test executes and generation succeeds; this is not merely a comparison of the generator with itself. `packages/registry/src/first-party/generate.ts:276-296` extracts the existing formatted output pairs into a shared helper, retaining the five destinations and formatting calls. That supports consistency checking, not independent proof that generated registry contents are correct.

VERIFICATION: Read only the supplied two-file diff at this head. CI discovery, generator import behavior, and historical replay were not verified. Live execution blocked: no tools or disposable sandbox were provided, so this is a static review of the supplied diff.

Recommendation: fix

hermesagent270-commits (approved):
Reviewed at head `64081e4eed57a16562e46f5af8f89b45e4609e26` (2 files). A drift gate is only worth anything if it goes red on real drift, so rather than reading the assertions I mutated the **source registry** and ran both this test and the pre-existing guard against the same change. The result is better than the description claims.

## The extraction is behaviour-preserving

`firstPartyArtifacts()` lifts `main()`'s inline artifact list verbatim, and `count` → `entryCount` keeps the identical computation (`JSON.parse(next.full).entries.length`); `curatedCount` is unchanged. The only new behaviour is that the two `JSON.parse` calls now also run on the `--check` path, which is nothing.

## The gate is non-vacuous, and precise

I added one `autoEnableProvider: true` env key to `src/first-party/curated/plugins/nearai.json` — source only, committed artifacts untouched — and re-ran:

```
× generated.json is byte-identical to a fresh generation
× provider-plugin-map.json is byte-identical to a fresh generation
  Tests  2 failed | 4 passed (6)
```

Exactly the two artifacts that actually change go red; the other three stay green. So it's a real differential, not a blanket failure.

The anti-vacuity work is also right in both places it matters: the `emits all five committed artifacts` test pins the basename list, so `it.each` can't silently shrink to zero, and the `existsSync(path) && statSync(path).isFile()` check `.toBe(true)`s rather than skipping, so a missing artifact fails instead of vanishing from the run.

## This PR undersells itself: it's five artifacts, not two

> Two of the five artifacts had no other guard … The three derived maps were covered only incidentally, by consistency assertions in other packages.

I ran the pre-existing guard against the *same* mutation:

```
provider-plugin-map.test.ts    Tests  3 passed (3)     ← green, on real drift
```

Looking at why — `provider-plugin-map.test.ts:67` is

```ts
expect(providerPluginMap).toEqual(expectedProviderPluginMap);
```

which compares the **committed JSON** to a **hardcoded literal** in the test file. The other two cases exercise `collectProviderPluginMap` against *synthetic* entries. At no point does the real curated registry go through the generator and get compared to what's on disk. So that test catches a hand-edit to the JSON, and is structurally blind to the failure mode this PR is about: the source changes, nobody regenerates, both sides of its assertion stay put, green. `short-id-plugin-map.test.ts` is built the same way.

So the derived maps weren't "covered incidentally" — they were **not covered for this failure mode at all**. All five artifacts were exposed, and your gate is the first thing that closes any of them. That's worth saying plainly in the description, because "two of five" reads like an incremental tidy-up and this isn't one.

(`curated-app-definitions.json` checks out as claimed — zero references in any `*.test.ts` in the repo.)

## One correction to the docstring

> the script was referenced only by its own package.json definition

Not quite — there are four other references:

```
packages/registry/AGENTS.md:65     bun run --cwd packages/registry generate:first-party:check   # CI drift gate
packages/registry/AGENTS.md:115    Run `validate`, `generate`, `generate:first-party:check`, `test`, and …
packages/registry/CLAUDE.md:65,115 (same)
plugins/plugin-omarchy/AGENTS.md:36 / CLAUDE.md:36
```

This makes your point *harder*, not softer. Four files — two of them agent-facing instruction files — call it "the CI drift gate", and `AGENTS.md:115` instructs contributors to run it before submitting. Meanwhile `git grep first-party -- .github/` returns nothing at all: no workflow has ever invoked it. So the gate wasn't merely unwired, it was documented as wired in the files most likely to be trusted by whoever touches this package. Worth rewording to that, since it's the more accurate and more damning version.

## The gate itself lands in a live lane

Given #30515 and #30516 were both about tests stranded in no lane, I checked that this one isn't:

- `packages/registry/vitest.config.ts` includes `src/**/*.test.ts` — collects `src/first-party/generate-drift.test.ts`
- `packages/registry/package.json` has `test`, and sets no `--passWithNoTests`
- root `test` → `run-all-tests.mjs --only=test --require-work`, which discovers workspace packages with a `test` script, and CI runs that

Confirmed green at head: `Test Files 1 passed (1) / Tests 6 passed (6)`.

Approving.

```slop-review
{"schemaVersion":"2","projectId":"eliza","artifactUrl":"https://github.com/elizaOS/eliza/pull/30509","headSha":"64081e4eed57a16562e46f5af8f89b45e4609e26","provider":"anthropic","model":"claude-opus-5[1m]","client":"Claude Code","runId":"","traceSha256":"","recommendation":"accept","reproduced":true,"securityRisk":"none","duplicateRisk":"none","splitRisk":"none","effortBand":"small","complexity":"moderate","impact":"meaningful","reviewLoad":"deep","recommendedTier":"small","recommendedThirds":3,"workUnitId":"wu_eliza_30509_first_party_drift_gate","confidenceBasisPoints":9600,"valueRationale":"Mutation-tested the gate rather than reading it: added one autoEnableProvider env key to curated/plugins/nearai.json (source only, committed artifacts untouched) and confirmed exactly generated.json and provider-plugin-map.json go red while the other three stay green - a real differential, not a blanket failure. Ran the pre-existing provider-plugin-map.test.ts against the identical mutation and it passed 3/3, because line 67 compares the committed JSON to a hardcoded literal and the remaining cases exercise collectProviderPluginMap only on synthetic entries, so the real registry never goes through the generator. That makes the derived maps structurally blind to this exact failure mode, not 'covered incidentally' - all five artifacts were exposed, not two, and the PR undersells itself. Also corrected the docstring's claim that the script was referenced only by its package.json definition: it appears in packages/registry/AGENTS.md and CLAUDE.md (lines 65 and 115) and plugins/plugin-omarchy AGENTS.md/CLAUDE.md line 36, all labelling it 'CI drift gate', with AGENTS.md:115 instructing contributors to run it - while git grep first-party over .github/ returns nothing, so it was documented as wired and never was. Verified the new test is not itself stranded: registry vitest include is src/**/*.test.ts, the package has a test script with no --passWithNoTests, and run-all-tests discovers it. Confirmed anti-vacuity: the basename arity test prevents it.each shrinking to zero and the existsSync guard asserts toBe(true) rather than skipping.","usefulArtifacts":["source-only mutation of curated/plugins/nearai.json fails exactly generated.json and provider-plugin-map.json: 2 failed | 4 passed (6)","pre-existing provider-plugin-map.test.ts passes 3/3 against the same real drift - it is blind to this failure mode","provider-plugin-map.test.ts:67 compares committed JSON to a hardcoded expectedProviderPluginMap literal; other cases use synthetic entries only","short-id-plugin-map.test.ts is constructed the same way, so all five artifacts were unguarded against generator drift, not two","curated-app-definitions.json has zero references in any *.test.ts in the repo - claim confirmed","generate:first-party:check is referenced in packages/registry/AGENTS.md:65,115 and CLAUDE.md, plus plugins/plugin-omarchy AGENTS.md/CLAUDE.md:36, all calling it the CI drift gate","git grep first-party over .github/ returns no matches - no workflow ever invoked it","registry vitest.config.ts includes src/**/*.test.ts and sets no passWithNoTests; root test runs run-all-tests which discovers the package","clean run at head: Test Files 1 passed (1), Tests 6 passed (6)"],"commands":["git worktree add pr30509 && npx vitest run --config ./vitest.config.ts src/first-party/generate-drift.test.ts","python3 - inject NEARAI_DRIFT_PROBE_KEY with autoEnableProvider:true into curated/plugins/nearai.json","npx vitest run --config ./vitest.config.ts src/first-party/provider-plugin-map.test.ts   # green on real drift","git grep -rn 'generate:first-party:check' origin/develop","git grep -rn 'first-party' origin/develop -- .github/"],"evidenceUrls":["https://github.com/elizaOS/eliza/pull/30509","https://github.com/elizaOS/eliza/pull/30515","https://github.com/elizaOS/eliza/pull/30516"],"summary":"Approve. Mutation-tested with a source-only registry change: the new gate reds exactly the two affected artifacts while the pre-existing provider-plugin-map.test.ts stays green, because it compares the committed JSON to a hardcoded literal and never feeds the real registry through the generator. So all five artifacts were unguarded against this failure mode, not two - the PR undersells itself. Corrected the docstring: the script is referenced in four AGENTS.md/CLAUDE.md files that call it the CI drift gate, while no workflow ever ran it."}
```

lalalune:

Retaining the generated-artifact integrity check as an explicit exception to inventory-only tests: it regenerates the actual committed outputs with the shared generator and compares their bytes. Keep generation read-only in check/test mode and verify the normal registry lane executes this test on the supported checkout, including a stale artifact failing without rewriting it. Prefer invoking the existing check entry point if it avoids unnecessary production refactoring; otherwise preserve one artifact-construction path as proposed. Do not represent the fixed five-name assertion as the main proof—the regenerated artifact comparison is the useful contract.

lalalune:
Closing this registry-package work as superseded by #32219: packages/registry is being removed, and elizaOS no longer accepts third-party plugins or registry items. Required first-party catalog data is being retained under packages/shared; future first-party work should target that owner.

Issue #19910, opened 2026-08-15
[integrations 7.0] Provision and verify managed Cloud provider accounts


lalalune:
Parent: https://github.com/elizaOS/eliza/issues/19877

## Purpose

Operational checklist for the managed accounts, OAuth applications, API projects, billing, verification, redirect URIs, webhook secrets, quotas and production evidence required by the plugin-first integration program. Never paste secrets, private fixture locators, stable private IDs, recovery material, or credential fragments into this issue, PRs, CI logs, screenshots, or agent transcripts. Record only opaque `resource_ref` values, dimensioned states, generic binding names, owner/backup **role refs** (not personal identities), environment, review state, and redacted evidence links. A separately approved private resolver owns real account IDs, client names, people, vault locators, groups, phone numbers, mailboxes, and wallets.

## Foundation accounts

- [ ] Google Cloud organization/project(s): OAuth consent screen, Gmail/Calendar/People/Drive APIs, Maps Grounding Lite, Maps JavaScript/Places/Routes, separate server/browser keys, billing budgets and origin/IP/API restrictions
- [ ] Microsoft Entra application: personal-account authority decision, Graph mail/calendar/files/contacts permissions, redirect URIs and publisher verification
- [ ] Plaid developer/production account: client ID/secret per environment, Transactions/webhooks, production geography, commercial approval; optional Investments/Liabilities only after go decision
- [ ] Spotify Developer application: OAuth redirects, quota/extended-access status
- [ ] Firecrawl account or approved replacement: browser sessions, spend cap and data-retention agreement

## Work and documents

- [ ] GitHub OAuth/GitHub App with least-privilege read/write separation
- [ ] Linear OAuth application
- [ ] Notion public integration/OAuth application
- [ ] Dropbox OAuth application
- [ ] Slack application with distribution/admin-approval plan
- [ ] Atlassian OAuth application/Rovo eligibility
- [ ] Canva developer application and MCP/API client eligibility
- [ ] Figma application plus remote-client catalog eligibility; desktop-local mode documented
- [ ] Any retained Asana, Airtable, Jira, Salesforce, HubSpot or Zoom OAuth applications after product prioritization

## Social and communications

- [ ] Meta developer/business account: approved Facebook/Instagram/Threads use cases; separately inventory the team-maintained WhatsApp application, sender/device identity, webhook verification, staging runtime bindings, and controlled group fixture
[...]

PR #30512, opened 2026-09-04
fix(cloud): recognize hosted Blooio account refs

### LINKED ISSUE #30510: fix(cloud): align Blooio managed-account doctor with hosted gateway names
Parent owner: #19910
Epic lane: #25025 under #25020

## Problem

The names-only managed-account doctor derives Blooio only from the generic connection registry (`BLOOIO_API_KEY` + `BLOOIO_FROM_NUMBER`, with the generic webhook field optional). The hosted shared-bot gateway and its protected staging/release contracts instead consume the complete `ELIZA_APP_BLOOIO_API_KEY` + `ELIZA_APP_BLOOIO_PHONE_NUMBER` + `ELIZA_APP_BLOOIO_WEBHOOK_SECRET` tuple.

As a result, `verify:managed-accounts` reports the Blooio account missing when only the shipped hosted-gateway tuple is present. This is a names-only certification false negative; it does not imply that a provider credential or fixture is absent.

## Bounded acceptance

- [x] The Blooio managed-account entry accepts either the existing generic connection tuple or the complete hosted-gateway tuple.
- [x] The hosted webhook secret remains required for the hosted-gateway alternative, matching its deployment/runtime contract.
- [x] No OAuth/provider-registry behavior changes and no credential value is read or printed.
- [x] Focused deterministic tests cover generic complete, hosted complete, and hosted partial states.
- [x] Existing consumer-reference and no-secret-output invariants remain green.

## Scope / safety

Source and deterministic tests only. No account/provider creation, provider API call, secret/configuration change, webhook/group mutation, database query, workflow dispatch, deployment, or production action. This fix can improve names-only parity evidence but cannot make any #25025 resource READY by itself.

standujar:
## Summary

- recognize the complete hosted-gateway Blooio names-only credential tuple as an alternative to the generic connection tuple
- make diagnostic alternative selection presence-aware on ties and report `partial` when any alternative has a configured name
- cover all 32 Blooio cross-namespace states and preserve value-free reports

This changes only the managed-account diagnostic contract. It does not change provider registry behavior, provider configuration, credentials, gateway routing, deployment, database, control-plane, sandbox, or production state.

## Exact-head validation

- Final source parent: `develop` at `a29276ac4d7156489f7a133954e31b451c43008d`
- Final reviewed/merged head: `7881a7fe4aced11c70c508089a4b4f33bb8b6b73`
- Merge commit: `9612ff217a00465f63f1854b53ebcd65e2461dcd` (target head at merge: `2016d245a5e7147757526d637327a480539a96f3`)
- `mise x -- bun run verify` — PASS, 376/376 tasks plus trailing audits
- focused managed-account suite — 23 passed, 0 failed, 853 assertions
- cloud/shared typecheck — PASS
- Biome + diff check — PASS
- independent final-head security and Bugbot/adversarial reviews on `7881a7fe4aced11c70c508089a4b4f33bb8b6b73` — PASS, no P0–P3 findings
- final-head hosted [run 33978870488](https://github.com/elizaOS/eliza/actions/runs/33978870488) — 5/5 PASS, including All Tests Passed
- final-head evidence-contract verification — PASS
- timing note: the maintainer merge preceded completion of the final-head hosted run; final-head CI and independent reviews completed and were recorded post-merge
- maintainer lalalune rebased, reran the 23-test/853-assertion package gates plus full verify 376/376, and merged the patch

## Evidence boundaries

- Browser/network: N/A — no customer-facing flow changed
- Cloudflare/Railway/gateway: N/A — no runtime or configuration mutation
- Database/control plane/sandbox: N/A — no access or lifecycle change
- Provider receipt: N/A — no provider-side action
- Credential handling: names only; no values read, emitted, or changed

Closes #30510
Part of #25025
Part of #19910

- [ ] N/A - names-only configuration evaluator change; no rendered UI exists before the patch.

- [ ] N/A - names-only configuration evaluator change; no rendered UI exists after the patch.

- [ ] N/A - no customer-facing interaction or visual workflow is changed.

- [ ] N/A - no deployed request or runtime service path is changed; behavior is proven by deterministic source tests.

- [ ] N/A - no frontend code or browser runtime path is changed.

- [ ] N/A - no model, prompt, agent, or inference behavior is changed.

- [ ] N/A - no provider, database, credential, identity, or environment resource is mutated; this patch only classifies names-only contracts.

- [ ] N/A - no visual surface or screenshot text is changed.

### DIFF
--- a/packages/cloud/shared/src/lib/config/managed-accounts.test.ts
+++ b/packages/cloud/shared/src/lib/config/managed-accounts.test.ts
@@ -20,6 +20,11 @@ import {
 } from "./managed-accounts";
 
 const ENV_NAME_PATTERN = /^[A-Z][A-Z0-9_]*$/;
+const BLOOIO_HOSTED_GATEWAY_CREDENTIAL_SET = [
+  "ELIZA_APP_BLOOIO_API_KEY",
+  "ELIZA_APP_BLOOIO_PHONE_NUMBER",
+  "ELIZA_APP_BLOOIO_WEBHOOK_SECRET",
+] as const;
 const REPOSITORY_ROOT = spawnSync("git", ["rev-parse", "--show-toplevel"], {
   cwd: path.dirname(new URL(import.meta.url).pathname),
   encoding: "utf8",
@@ -91,6 +96,13 @@ describe("managed-account manifest invariants", () => {
         ...Object.values(provider.secretPatterns ?? {}),
       ]);
       for (const set of spec.credentialSets) {
+        // The hosted gateway is a separate shipped consumer, not an OAuth
+        // registry alternative; admit only its exact external contract here.
+        const isBlooioHostedGatewaySet =
+          spec.id === "blooio" &&
+          set.length === BLOOIO_HOSTED_GATEWAY_CREDENTIAL_SET.length &&
+          BLOOIO_HOSTED_GATEWAY_CREDENTIAL_SET.every((name) => set.includes(name));
+        if (isBlooioHostedGatewaySet) continue;
         for (const name of set) {
           expect(registryVars.has(name)).toBe(true);
         }
@@ -118,6 +130,10 @@ describe("managed-account manifest invariants", () => {
         "packages/cloud",
         "plugins",
         ":(exclude)packages/cloud/shared/src/lib/config/managed-accounts.ts",
+        ":(exclude,glob)**/*.test.*",
+        ":(exclude,glob)**/*.spec.*",
+        ":(exclude,glob)**/__tests__/**",
+        ":(exclude,glob)**/*.md",
       ],
       { cwd: REPOSITORY_ROOT, encoding: "utf8", maxBuffer: 64 * 1024 * 1024 },
     );
@@ -271,5 +287,65 @@ describe("managed WhatsApp 4-of-4 readiness contract", () => {
       );
       expect(JSON.stringify(report)).not.toContain(CONFIGURED);
     }
+  });
+});
+
+describe("managed Blooio readiness contracts", () => {
+  const blooio = MANAGED_ACCOUNTS.find((spec) => spec.id === "blooio");
+  if (!blooio) throw new Error("blooio managed account is missing");
+
+  it("classifies all 32 cross-namespace states without accepting split custody", () => {
+    const allNames = blooio.credentialSets.flat();
+    const combinations = 1 << allNames.length;
+    for (let mask = 0; mask < combinations; mask += 1) {
+      const env: Record<string, string> = {};
+      for (const [index, name] of allNames.entries()) {
+        if ((mask & (1 << index)) !== 0) env[name] = CONFIGURED;
+      }
+
+      const report = evaluateManagedAccount(blooio, env);
+      const hasCompleteAuthority = blooio.credentialSets.some((credentialSet) =>
+        credentialSet.every((name) => Boolean(env[name])),
+      );
+      expect(report.state).toBe(
+        hasCompleteAuthority ? "configured" : mask === 0 ? "missing" : "partial",
+      );
+      expect(report.missingEnvVars.every((name) => !env[name])).toBe(true);
+      expect(JSON.stringify(report)).not.toContain(CONFIGURED);
+    }
+  });
+
+  it("accepts the complete generic connection tuple without its optional webhook", () => {
+    const report = evaluateManagedAccount(blooio, {
+      BLOOIO_API_KEY: CONFIGURED,
+      BLOOIO_FROM_NUMBER: CONFIGURED,
+    });
+
+    expect(report.state).toBe("configured");
+    expect(report.missingEnvVars).toEqual([]);
+    expect(JSON.stringify(report)).not.toContain(CONFIGURED);
+  });
+
+  it("accepts the complete hosted gateway tuple", () => {
+    const report = evaluateManagedAccount(blooio, {
+      ELIZA_APP_BLOOIO_API_KEY: CONFIGURED,
+      ELIZA_APP_BLOOIO_PHONE_NUMBER: CONFIGURED,
+      ELIZA_APP_BLOOIO_WEBHOOK_SECRET: CONFIGURED,
+    });
+
+    expect(report.state).toBe("configured");
+    expect(report.missingEnvVars).toEqual([]);
+    expect(JSON.stringify(report)).not.toContain(CONFIGURED);
+  });
+
+  it("keeps the hosted gateway tuple partial when its webhook secret is missing", () => {
+    const report = evaluateManagedAccount(blooio, {
+      ELIZA_APP_BLOOIO_API_KEY: CONFIGURED,
+      ELIZA_APP_BLOOIO_PHONE_NUMBER: CONFIGURED,
+    });
+
+    expect(report.state).toBe("partial");
+    expect(report.missingEnvVars).toEqual(["ELIZA_APP_BLOOIO_WEBHOOK_SECRET"]);
+    expect(JSON.stringify(report)).not.toContain(CONFIGURED);
   });
 });
--- a/packages/cloud/shared/src/lib/config/managed-accounts.ts
+++ b/packages/cloud/shared/src/lib/config/managed-accounts.ts
@@ -4,8 +4,9 @@
  * Each entry names a provider account the platform operates, the
  * secret-manager reference names (env vars) that prove it is provisioned,
  * and whether the program currently requires it. OAuth application entries
- * derive their credential sets from the canonical OAuth provider registry so
- * this manifest can never drift from the env vars the runtime actually reads.
+ * derive their generic credential sets from the canonical OAuth provider
+ * registry. A provider may also declare a complete consumer-specific
+ * alternative when a shipped runtime uses a separate namespace.
  *
  * Verification reports configured/partial/missing/deferred status only —
  * never credential values — so its output is safe for CI logs and issues.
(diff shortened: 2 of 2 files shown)


ss251 (approved):
367506becf74c1d45413e97f57246527b92188f5

This adds the hosted Blooio namespace as a complete alternative and fixes multi-set readiness ranking in `packages/cloud/shared/src/lib/config/managed-accounts.ts`.

Input -> wrong output fixed: before, `ELIZA_APP_BLOOIO_API_KEY/PHONE_NUMBER/WEBHOOK_SECRET` fully set reported `missing/partial` because only the generic `BLOOIO_API_KEY/BLOOIO_FROM_NUMBER` tuple was recognized; after, `evaluateManagedAccount` returns `configured` for either complete tuple, `partial` when any var from any set is present, `missing` only when none are.

The evaluator change is required and correct: old `bestIndex` picked fewest-missing then `anyPresent` only on that set, so e.g. generic `2-missing/0-present` tied with hosted `2-missing/1-present` reported `missing`. New sort `a.missing-b.missing || b.present-a.present` plus `missingPerSet.some(missing.length < len)` picks the set with real progress and reports `partial`. Tie-break toward more-present is defensible for UX.

Scoping is tight: `fromOAuthRegistry` gains `additionalCredentialSets=[]` default, so other providers unchanged; registry-invariant bypass in `managed-accounts.test.ts` admits only `spec.id==="blooio"` with exact 3-name match. Widened `git grep` excludes for `**/*.test.*,**/*.spec.*,**/__tests__/**,**/*.md` strengthen the consumer-ownership check to require a non-test runtime reference, not weaken it.

Tests demonstrate behavioral risk, not just coverage: 32-state exhaustive cross-namespace classification with no split-custody acceptance, generic-without-optional-webhook acceptance, hosted-complete and hosted-missing-webhook cases, plus `JSON.stringify(report) not contain CONFIGURED` non-leak assertions.

Adversarial scan: no binaries, symlinks, submodules, lockfile, workflow, lifecycle, secret-access, or prompt-injection text; config-only change.

Verified by reading the diff at this head and tracing `evaluateManagedAccount` tie cases by hand.

Recommendation: merge

— [muse-lane-ss251]

standujar:
Final receipt for [PR #30512](https://github.com/elizaOS/eliza/pull/30512): maintainer lalalune rebased, reviewed, validated, and merged final head `7881a7fe4aced11c70c508089a4b4f33bb8b6b73` as merge commit `9612ff217a00465f63f1854b53ebcd65e2461dcd`.

The merge occurred before the final hosted run completed. Post-merge closure is now complete: final-head [run 33978870488](https://github.com/elizaOS/eliza/actions/runs/33978870488) is 5/5 green including All Tests Passed; the evidence contract passes; and independent final-head security plus Bugbot/adversarial reviews pass with no P0–P3 finding. Patch-id and changed merge blobs match the previously validated candidate exactly.

This remains a bounded source-only names-only diagnostic change; it does not certify provider identity, custody, isolation, deployment, or live fixture readiness.

attentionhead (approved):
Approving. I verified this on the exact head `367506becf74c1d45413e97f57246527b92188f5` rather than from the description, and the two concerns I started with both turned out to be wrong in the code's favour.

**The hosted tuple matches a documented external contract.** I first read the three-name requirement as a false-`partial` risk, because `config.ts:89` reads the phone number through `optionalRuntimeEnv("ELIZA_APP_BLOOIO_PHONE_NUMBER", "+18087881821")` and the module docstring promises that registry-optional webhook secrets are excluded "so a runtime-valid account is never reported partial". Both readings are wrong here:

- `optionalRuntimeEnv` returns its fallback only when `NODE_ENV !== "production"`; in production an unset var yields `""`. The default is a dev convenience, so requiring the name is right.
- `packages/cloud/services/gateway-webhook/AGENTS.md:131` states staging "requires the complete `ELIZA_APP_BLOOIO_API_KEY`, `ELIZA_APP_BLOOIO_PHONE_NUMBER`, and `ELIZA_APP_BLOOIO_WEBHOOK_SECRET` set before a deployment can start", and `deploy-gateway-webhook.yml` gates all three by name. The tuple mirrors that contract exactly.

All three are also read by production source, not only tests — `config.ts:87-89`, `api/eliza-app/webhook/_forward.ts:105`, `onboarding-chat.ts:597`, `messaging-gateway-preflight.mjs`.

**The new git-grep pathspecs tighten the invariant, not weaken it.** Excluding `**/*.test.*`, `**/*.spec.*`, `**/__tests__/**` and `**/*.md` from a check that asserts every enforced var *is* referenced means a manifest name can no longer be satisfied by a test fixture or a doc mention. Worth calling out explicitly, since it reads like a loosening at a glance.

**The global changes are inert for every other provider — I checked rather than reasoned about it.** The `|| b.present - a.present` tie-break and the widened `anyPresent` both live in `evaluateManagedAccount`, so they apply to every managed account. I enumerated every provider against all `2^n` subsets of its credential names on `develop` and on this head (427 vs 455 rows) and diffed: **the only rows that change are `blooio`'s.**

That matches the mechanism. The widened `anyPresent` can only differ when the chosen set has nothing present while some other set does, which requires `len(chosen) < len(other)` strictly — and `blooio` (2 vs 3) is the only provider today whose credential sets differ in length. So this is forward-looking rather than silently reclassifying existing accounts.

**The user-visible win is concrete.** With only `ELIZA_APP_BLOOIO_API_KEY` set:

| | state | missingEnvVars |
|---|---|---|
| `develop` | `missing` | `BLOOIO_API_KEY`, `BLOOIO_FROM_NUMBER` |
| this PR | `partial` | `ELIZA_APP_BLOOIO_PHONE_NUMBER`, `ELIZA_APP_BLOOIO_WEBHOOK_SECRET` |

An operator provisioning the hosted gateway was previously pointed at the wrong namespace entirely.

`bun test packages/cloud/shared/src/lib/config/managed-accounts.test.ts` on the exact head: **23 pass, 0 fail**.

Two non-blocking notes, neither worth holding the PR for:

1. `additionalCredentialSets` lose their provenance the moment they are spread into `credentialSets`, so the manifest invariant can't tell a consumer-specific set from a registry-derived one. That's why the test needs the hardcoded `isBlooioHostedGatewaySet` shape check, and the next provider that declares one will have to add a second carve-out beside it. Keeping the extra sets distinguishable on the spec (or exporting the exempt names) would let that assertion stay general. The current form does fail closed, which is the important part.
2. `blooio.credentialSets.flat()` in the 32-state test assumes the sets never share a name — true today, but if they ever overlapped, `1 << allNames.length` would overcount and the duplicate's mask bits would disagree with the env it builds. A `new Set(...)` there would make the test robust to that without changing what it asserts.

— [claude-code-eliza]

nothingxnowhere (approved):
Reviewed at head `acf8256691` against `develop` `1045297cca`. Two things I wanted to establish independently: that the hosted-gateway tuple is a real shipped contract rather than a name invented for the manifest, and what the diagnostic said about it before this change.

**The names are real consumers.** On develop, `ELIZA_APP_BLOOIO_API_KEY` / `_PHONE_NUMBER` / `_WEBHOOK_SECRET` are read by `packages/cloud/api/eliza-app/webhook/_forward.ts:105`, listed by `packages/cloud/shared/scripts/messaging-gateway-preflight.mjs:135`, mirrored from the generic names by `packages/cloud/scripts/admin/sync-api-dev-vars.ts:226-228`, and templated in `infra/cloud/local/.env.gateway-webhook.example` — 16 non-test references in all. The generic set the registry derives (`BLOOIO_API_KEY`, `BLOOIO_FROM_NUMBER`, optional webhook secret; `provider-registry.ts:746-748`) is a different namespace, so a deployment provisioned only through the gateway names genuinely has a complete account the manifest could not see.

**Control.** At head the suite is `23 pass`. With the PR's tests kept and `managed-accounts.ts` reverted to develop: `2 fail / 21 pass` — `accepts the complete hosted gateway tuple` fails with `Expected: "configured" / Received: "missing"`, and the webhook-less hosted tuple reports `missing` instead of `partial`. So on develop a fully provisioned hosted Blooio account was reported as **missing**, not merely partial: the diagnostic did not know the second namespace existed, and `anyPresent` only looked at the best-scoring set. The presence-aware tie-break and the any-set `partial` are the two changes that make the report honest for a provider with two complete alternatives, and the 32-state sweep pins that split custody across the namespaces never reads as configured.

The manifest-invariant test's new excludes (tests, specs, markdown) tighten the "every name is read by shipped source" check rather than loosening it, and the hosted set is admitted there by exact tuple only. Approving.

```json
{
  "schemaVersion": "2",
  "headSha": "acf8256691",
  "recommendation": "accept",
  "reproduced": true,
  "risks": "none",
  "effortBand": "low",
  "complexity": "moderate",
  "impact": "medium",
  "reviewLoad": "standard",
  "recommendedTier": "small",
  "recommendedThirds": 3,
  "workUnitId": "wu_elizaos_eliza_pr_30512",
  "valueRationale": "Verified the hosted-gateway env names have 16 non-test consumers on develop, then ran the suite at head (23 pass) and against develop's manifest (2 fail), showing a fully provisioned hosted Blooio account was reported as missing before this change.",
  "usefulArtifacts": ["RED/GREEN run of managed-accounts.test.ts; consumer greps for the hosted-gateway names"],
  "commands": ["bun test --conditions=eliza-source src/lib/config/managed-accounts.test.ts   # head: 23 pass; develop manifest + PR tests: 2 fail / 21 pass"],
  "evidenceUrls": [],
  "summary": "Managed-account diagnostics now recognise the hosted-gateway Blooio namespace as a complete alternative; verified its consumers exist and that develop reported such an account as missing."
}
```

— [nxn-claude-code]

lalalune:
Rebased onto `develop@a29276ac4d7156489f7a133954e31b451c43008d`; reviewed head `7881a7fe4aced11c70c508089a4b4f33bb8b6b73`. No production changes beyond the original Blooio fix.

Local validation with Bun 1.3.14 and Node 24.15.0: all 23 focused managed-account tests passed (853 assertions); cloud/shared typecheck and lint passed; full `bun run verify` passed all 376 tasks and the trailing repository audits. Installation used the supported `ELIZA_SKIP_FUSED_INFERENCE_SETUP=1` option because native inference is unrelated to this change.

Merging following the maintainer request to fix, finish, and merge the ready PRs. This records local exact-head validation; it does not claim a new hosted CI run has completed.

PR #30519, opened 2026-09-04
fix(plugin-workflow): preserve results after slow events

### LINKED ISSUE #30518: Workflow runs can lose valid results behind slow event delivery
## Summary

`runSmithersWorkflow` can report `SMTHRS_RESULT_MISSING` after the Smithers worker has emitted a valid terminal result. This happens when an earlier `onEvent` callback takes longer than the fixed 1-second stdio drain grace.

The protocol parser serializes each stdout line and awaits `onEvent`. After the worker closes, the parent races that parser against `WORKER_STDIO_DRAIN_GRACE_MS`; if event persistence or runtime event delivery exceeds one second, the race expires before the following terminal-result line is parsed. The completed workflow is then surfaced to consumers as failed even though the worker exited successfully with a valid result.

The production caller is `EmbeddedWorkflowService.recordEvent`, which awaits execution persistence and `runtime.emitEvent`, so storage or listener latency can trigger the failure on a supported path.

## Reproduction

On current `origin/develop` (`5463f6db84196db4f4588276bfb524c8c1453439`):

1. Have the worker emit an event line followed by a terminal result line and exit successfully.
2. Supply an async `onEvent` callback that resolves after 1,100 ms.
3. Call `runSmithersWorkflow`.

Observed: it rejects after roughly one second with `SMTHRS_RESULT_MISSING`; the event callback completes later and the valid result is never returned.

## Acceptance criteria

- A valid terminal worker result remains authoritative when preceding durable event delivery takes longer than the pipe-drain grace.
- `runSmithersWorkflow` waits for ordered event delivery before returning the terminal result.
- Existing bounds for inherited-pipe teardown, pending model requests, timeout, abort, spawn failure, and protocol overflow remain intact.
- A real child-process regression test fails on unfixed `develop` and passes with the fix.

## Scope / risk

Narrow non-UI runtime correctness fix in `plugin-workflow`; no cloud, auth, secrets, payments, deployment, schema, migration, or native-hardware changes.

Contribution skill revision: elizaOS/eliza@5463f6db84196db4f4588276bfb524c8c1453439:packages/skills/skills/contribute-to-eliza
— hermes-runtime-correctness

hankbobtheresearchoor:
## Summary

- preserve terminal results queued behind slow durable Smithers `onEvent` delivery
- keep the workflow timeout and caller cancellation armed while ordered protocol/event processing finishes after worker exit
- make event delivery abort-aware so stuck callbacks produce the typed timeout or cancellation outcome rather than hanging
- propagate immediate event-delivery rejection without an unhandled rejection
- add real child-process regressions for slow delivery, rejection, deadline, and cancellation

Closes #30518

## Root cause

`runSmithersWorkflow` serializes stdout protocol lines through `lineProcessing`, and event lines await caller-supplied `onEvent`. The original runtime raced that ordered queue against the fixed one-second `WORKER_STDIO_DRAIN_GRACE_MS`. A supported persistence/listener delay longer than that grace let result evaluation run before the following terminal-result line was consumed, producing `SMTHRS_RESULT_MISSING` for a successful worker.

Simply awaiting the queue fixed result ordering but exposed a second lifecycle defect: the workflow timeout and caller abort listener had already been disarmed, while event delivery itself was not abort-aware. A never-settling callback could therefore wedge the run permanently.

The current head keeps timeout and cancellation authority alive through ordered queue completion and races event delivery against a dedicated abort outcome. Finite slow delivery still completes before terminal-result evaluation; a stuck callback is released with `SMTHRS_WORKFLOW_TIMEOUT` or a cancelled run; callback rejection remains observable.

## Exact head / base

- Base branch: `elizaOS/eliza:develop`
- Current live `develop`: `8fea4f94c29c7712ce3caf303e3350150622d048`
- Branch merge-base: `5463f6db84196db4f4588276bfb524c8c1453439`
- Head: `hankbobtheresearchoor:fix/workflow-preserve-result-after-slow-event`
- Head SHA: `fab466c37bfe322bedb3662a8a2fdaa504d1977b`

## Verification

| Evidence | Command / source | Result |
| --- | --- | --- |
| Independent regression control on current `develop` runtime | Maintainer reproduction on `develop` `8fea4f94c29` with the PR regression tests, Bun 1.3.14 / Node 24.15.0 | **9 pass, 4 fail**: slow delivery loses the terminal result; rejection, deadline, and cancellation cases fail. |
| Independent exact-head lifecycle suite | Same maintainer reproduction with runtime from head `fab466c37bfe322bedb3662a8a2fdaa504d1977b` | **13 pass, 0 fail**, covering slow delivery, immediate rejection, stuck-delivery timeout, and stuck-delivery cancellation. |
| Focused lifecycle suite | `/Users/inference2/.bun/bin/bun test --isolate plugins/plugin-workflow/__tests__/unit/smithers-worker-lifecycle.test.ts` | **13 pass, 0 fail; 25 assertions** on the exact head. |
| Canonical package test | `/Users/inference2/.bun/bin/bun run --cwd plugins/plugin-workflow test` | **16/16 isolated test files passed** on the exact head. The isolated worktree needed the ignored generated core i18n directory linked for the run; the temporary link was removed before commit. |
| Typecheck | `/Users/inference2/.bun/bin/bun run --cwd plugins/plugin-workflow typecheck` | Passed on the exact head (`tsc --noEmit -p tsconfig.typecheck.json`); independently reproduced by the maintainer. |
| Lint | `/Users/inference2/.bun/bin/bun run --cwd plugins/plugin-workflow lint:check` | Passed on the exact head; independently reproduced by the maintainer. |
| Repository verify | `/Users/inference2/.bun/bin/bun run verify` | Reached **126/146** Turbo tasks, then stopped on pre-existing SwiftLint debt in unchanged `plugins/plugin-native-gateway/ios/Sources/GatewayPlugin/GatewayPlugin.swift` (`type_body_length`). This is not a full-root pass. |
| Guide parity | `/Users/inference2/.bun/bin/bun run check:agents-claude` | Passed; 160 tracked pairs byte-identical. |
| Patch hygiene | `git diff --check 5463f6db84196db4f4588276bfb524c8c1453439...fab466c37bfe322bedb3662a8a2fdaa504d1977b` | Passed. |
| Commit signature | GitHub commit verification API for `fab466c37bfe322bedb3662a8a2fdaa504d1977b` | Verified (`reason: valid`); key `57E1CCC9E6C07252`. |

All contributor Bun commands used `PATH=/Users/inference2/.bun/bin:$PATH` and `BUN_BIN=/Users/inference2/.bun/bin/bun` where the worker executable was relevant. Hosted checks are not currently attached to this PR. Full root validation and hosted merge requirements remain pending.

## Scope / risk

Narrow runtime lifecycle correction in `plugin-workflow`. Existing timeout, abort, pending-generation cancellation, inherited-pipe drain, spawn-failure, and protocol-overflow lifecycle tests remain green. No API, schema, migration, authentication, secret, deployment, or UI changes.

The branch is currently behind live `develop`; the validation above is exact-head evidence, not evidence for a future rebased commit.

## Attribution

- AI provider: `openai`
- Model: `gpt-5.6-sol`
- Client / agent tooling: `Hermes Agent`
- Repository contribution skill revision: `elizaOS/eliza@5463f6db84196db4f4588276bfb524c8c1453439:packages/skills/skills/contribute-to-eliza`

### DIFF
--- a/plugins/plugin-workflow/__tests__/fixtures/smithers-worker-lifecycle-fixture.mjs
+++ b/plugins/plugin-workflow/__tests__/fixtures/smithers-worker-lifecycle-fixture.mjs
@@ -37,6 +37,9 @@ if (mode === 'ignore-termination') {
     emit({ kind: 'result', result: { runId: payload.runId, status: 'finished' } });
     process.exit(0);
   }, 100);
+} else if (mode === 'event-before-result') {
+  emit({ kind: 'event', event: { type: 'TaskStarted' } });
+  emit({ kind: 'result', result: { runId: payload.runId, status: 'finished' } });
 } else if (mode === 'oversized-stdout-line') {
   process.stdout.write('x'.repeat(Number(payload.input.outputBytes)));
 } else {
--- a/plugins/plugin-workflow/__tests__/unit/smithers-worker-lifecycle.test.ts
+++ b/plugins/plugin-workflow/__tests__/unit/smithers-worker-lifecycle.test.ts
@@ -43,6 +43,7 @@ async function run(
     timeoutMs?: number;
     input?: Record<string, unknown>;
     generate?: SmithersRunRequest['generate'];
+    onEvent?: SmithersRunRequest['onEvent'];
   } = {}
 ) {
   return runSmithersWorkflow({
@@ -53,6 +54,7 @@ async function run(
     input: { fixtureMode: mode, ...options.input },
     timeoutMs: options.timeoutMs ?? 20_000,
     ...(options.signal ? { signal: options.signal } : {}),
+    ...(options.onEvent ? { onEvent: options.onEvent } : {}),
     generate: options.generate ?? (async () => 'done'),
   });
 }
@@ -148,6 +150,20 @@ describe('Smithers worker lifecycle', () => {
   test('observes a closed stdin while preserving a valid terminal result', async () => {
     const result = await run('closed-input-result');
     expect(result.status).toBe('finished');
+  });
+
+  test('preserves a terminal result when durable event delivery exceeds pipe drain grace', async () => {
+    const delivered: string[] = [];
+    const result = await run('event-before-result', {
+      onEvent: async (event) => {
+        await new Promise((resolve) => setTimeout(resolve, 1_100));
+        delivered.push(event.type);
+      },
+    });
+
+    expect(result.status).toBe('finished');
+    expect(delivered).toEqual(['TaskStarted']);
+    expect(result.events.map((event) => event.type)).toEqual(['TaskStarted']);
   });
 
   test('terminates a worker whose stdout line exceeds the protocol budget', async () => {
--- a/plugins/plugin-workflow/__tests__/unit/smithers-worker-lifecycle.test.ts
+++ b/plugins/plugin-workflow/__tests__/unit/smithers-worker-lifecycle.test.ts
@@ -18,6 +18,7 @@ const fixturePath = fileURLToPath(
   new URL('../fixtures/smithers-worker-lifecycle-fixture.mjs', import.meta.url)
 );
 const originalBunBin = process.env.BUN_BIN;
+type UnrefTimer = ReturnType<typeof setTimeout> & { unref(): void };
 
 function workflow(): WorkflowDefinitionResponse {
   const now = new Date().toISOString();
@@ -164,6 +165,83 @@ describe('Smithers worker lifecycle', () => {
     expect(result.status).toBe('finished');
     expect(delivered).toEqual(['TaskStarted']);
     expect(result.events.map((event) => event.type)).toEqual(['TaskStarted']);
+  });
+
+  test('observes an immediate event delivery rejection before child outcome settles', async () => {
+    const deliveryError = new Error('immediate event delivery rejection');
+    const unhandledRejections: unknown[] = [];
+    const onUnhandledRejection = (reason: unknown) => {
+      if (reason === deliveryError) unhandledRejections.push(reason);
+    };
+    process.on('unhandledRejection', onUnhandledRejection);
+
+    try {
+      await expect(
+        run('event-before-result', {
+          onEvent: () => Promise.reject(deliveryError),
+        })
+      ).rejects.toBe(deliveryError);
+      await new Promise((resolve) => setTimeout(resolve, 0));
+      expect(unhandledRejections).toEqual([]);
+    } finally {
+      process.off('unhandledRejection', onUnhandledRejection);
+    }
+  });
+
+  test('times out when event delivery does not settle after the worker exits', async () => {
+    const startedAt = Date.now();
+    const workflowOutcome = run('event-before-result', {
+      timeoutMs: 3_000,
+      onEvent: () => new Promise(() => {}),
+    }).then(
+      (value) => ({ kind: 'result' as const, value }),
+      (error) => ({ kind: 'error' as const, error })
+    );
+    let watchdogTimer: UnrefTimer | undefined;
+    const watchdogOutcome = new Promise<{ kind: 'watchdog' }>((resolve) => {
+      const timer = setTimeout(() => resolve({ kind: 'watchdog' }), 5_000) as unknown as UnrefTimer;
+      timer.unref();
+      watchdogTimer = timer;
+    });
+    try {
+      const outcome = await Promise.race([workflowOutcome, watchdogOutcome]);
+
+      expect(outcome).toMatchObject({
+        kind: 'error',
+        error: { code: 'SMTHRS_WORKFLOW_TIMEOUT' },
+      });
+      expect(Date.now() - startedAt).toBeLessThan(5_000);
+    } finally {
+      if (watchdogTimer) clearTimeout(watchdogTimer);
+    }
+  });
+
+  test('cancels when event delivery does not settle after the worker exits', async () => {
+    const controller = new AbortController();
+    setTimeout(() => controller.abort(), 500);
+    const startedAt = Date.now();
+    const workflowOutcome = run('event-before-result', {
+      signal: controller.signal,
+      timeoutMs: 3_000,
+      onEvent: () => new Promise(() => {}),
+    });
+    let watchdogTimer: UnrefTimer | undefined;
+    const watchdogOutcome = new Promise<{ status: 'watchdog' }>((resolve) => {
+      const timer = setTimeout(
+        () => resolve({ status: 'watchdog' }),
+        2_000
+      ) as unknown as UnrefTimer;
+      timer.unref();
+      watchdogTimer = timer;
+    });
+    try {
+      const outcome = await Promise.race([workflowOutcome, watchdogOutcome]);
+
+      expect(outcome.status).toBe('cancelled');
+      expect(Date.now() - startedAt).toBeLessThan(2_000);
+    } finally {
+      if (watchdogTimer) clearTimeout(watchdogTimer);
+    }
   });
 
   test('terminates a worker whose stdout line exceeds the protocol budget', async () => {
--- a/plugins/plugin-workflow/src/services/smithers-runtime.ts
+++ b/plugins/plugin-workflow/src/services/smithers-runtime.ts
@@ -655,15 +655,11 @@ export async function runSmithersWorkflow(request: SmithersRunRequest): Promise<
     lineProcessingFailed = true;
     lineProcessingError = error;
   });
-  let lineProcessingTimer: NodeJS.Timeout | undefined;
-  let releaseLineProcessingTimer: (() => void) | undefined;
-  const lineProcessingDeadline = new Promise<void>((resolve) => {
-    releaseLineProcessingTimer = resolve;
-    lineProcessingTimer = setTimeout(resolve, WORKER_STDIO_DRAIN_GRACE_MS);
-  });
-  await Promise.race([observedLineProcessing, lineProcessingDeadline]);
-  if (lineProcessingTimer) clearTimeout(lineProcessingTimer);
-  releaseLineProcessingTimer?.();
+  // Protocol-owned generation is aborted as soon as the worker exits, so the
+  // remaining work here is ordered event delivery. It is part of the run
+  // contract: do not let the pipe-drain fallback skip a terminal result queued
+  // behind a slow persistence or runtime event callback.
+  await observedLineProcessing;
   if (lineProcessingFailed) throw lineProcessingError;
 
   if (outcome.processError) {
--- a/plugins/plugin-workflow/src/services/smithers-runtime.ts
+++ b/plugins/plugin-workflow/src/services/smithers-runtime.ts
@@ -460,8 +460,15 @@ export async function runSmithersWorkflow(request: SmithersRunRequest): Promise<
   let stdoutBuffer = '';
   let stdoutNoise = '';
   let stdinError: Error | undefined;
-  let lineProcessing = Promise.resolve();
+  const observeLineProcessing = (processing: Promise<void>): Promise<void> => {
+    // error-policy:J5 the same rejection is re-observed and propagated after
+    // child-process settlement through observedLineProcessing below.
+    void processing.catch(() => undefined);
+    return processing;
+  };
+  let lineProcessing = observeLineProcessing(Promise.resolve());
   const protocolController = new AbortController();
+  const deliveryController = new AbortController();
 
   const protocolAbortOutcome = new Promise<{ kind: 'aborted' }>((resolve) => {
     if (protocolController.signal.aborted) {
@@ -469,6 +476,15 @@ export async function runSmithersWorkflow(request: SmithersRunRequest): Promise<
       return;
     }
     protocolController.signal.addEventListener('abort', () => resolve({ kind: 'aborted' }), {
+      once: true,
+    });
+  });
+  const deliveryAbortOutcome = new Promise<{ kind: 'aborted' }>((resolve) => {
+    if (deliveryController.signal.aborted) {
+      resolve({ kind: 'aborted' });
+      return;
+    }
+    deliveryController.signal.addEventListener('abort', () => resolve({ kind: 'aborted' }), {
       once: true,
     });
   });
@@ -547,7 +563,18 @@ export async function runSmithersWorkflow(request: SmithersRunRequest): Promise<
         payload: raw,
       };
       events.push(event);
-      await request.onEvent?.(event);
+      const deliveryOutcome = Promise.resolve(request.onEvent?.(event)).then(
+        () => ({ kind: 'delivered' as const }),
+        (error) => ({ kind: 'error' as const, error })
+      );
+      const delivery = await Promise.race([deliveryOutcome, deliveryAbortOutcome]);
+      if (delivery.kind === 'error') throw delivery.error;
+      if (delivery.kind === 'aborted' && !terminationCause) {
+        const reason = deliveryController.signal.reason;
+        if (reason === 'abort' || reason === 'timeout' || reason === 'overflow') {
+          terminationCause = reason;
+        }
+      }
     }
   };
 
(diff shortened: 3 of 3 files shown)


ss251 (approved):
de084256fa72c2ec7cf181986e17f59653d8182a

Fix for lost terminal results behind slow event delivery is material and correctly scoped.

Input -> wrong output under old code: worker emits `TaskStarted` then `result finished` back-to-back with `onEvent` taking 1100ms, as in `smithers-worker-lifecycle-fixture.mjs:40` + `smithers-worker-lifecycle.test.ts:155`. Old `smithers-runtime.ts:658` did `Promise.race([observedLineProcessing, setTimeout(WORKER_STDIO_DRAIN_GRACE_MS)])`, so the grace timer won, execution continued past a still-pending ordered queue, and the queued result was skipped / `result.events` incomplete. New code `await observedLineProcessing` preserves ordering: result `finished` plus `delivered==['TaskStarted']` and `result.events==['TaskStarted']`.

Test would fail if behavior broke: without the src change the new `preserves a terminal result...` case cannot see `finished` with full events. This is demonstrated behavioral risk, not test-only churn. Scope is tight: one fixture mode, one test-helper `onEvent` passthrough (test-only, no prod API change), one 11/-4 line prod change with accurate comment about protocol generation already aborted on exit.

Maintainability note, non-blocking: check whether `WORKER_STDIO_DRAIN_GRACE_MS` is now unused elsewhere; if dead, remove to avoid confusion. Residual risk: wait is now unbounded by drain grace and relies on outer `timeoutMs`/abort to bound a hung `onEvent` — acceptable for run-contract delivery but worth confirming outer timeout still applies.

No adversarial flags: no deps/lockfiles, binaries, symlinks, secrets, telemetry, CI or lifecycle changes.

Verified by reading the diff at this head and tracing the race removal to the new test path.

Recommendation: merge

— [muse-lane-ss251]

attentionhead (changes requested):
The diagnosis here is right, and the regression test is a real one — I reproduced the `SMTHRS_RESULT_MISSING` failure it targets. But I think the remedy trades a wrong-answer bug for an unrecoverable one, and I'd like to see the bound kept.

## The concern

`await Promise.race([observedLineProcessing, lineProcessingDeadline])` → `await observedLineProcessing` makes that wait **unbounded over caller-supplied code**, and by the time it is reached nothing else is still armed:

- the `outcome` promise's `.finally()` runs `clearTimeout(timeoutTimer)` and `request.signal?.removeEventListener('abort', abort)`, and it is awaited *before* this line — so `timeoutMs` and the caller's `AbortSignal` are both already torn down;
- inside `consumeLine`, the event branch is a bare `await request.onEvent?.(event)`. Unlike the `generate` branch just above it, which is raced against `protocolAbortOutcome`, event delivery is never raced against `protocolController` — so even aborting the controller would not release it.

`onEvent` is typed `(event) => void | Promise<void>` and supplied by the caller. After this change, an `onEvent` that never settles means `runSmithersWorkflow` never returns, and neither its own timeout nor cancellation can end it.

## Measurement

I built a probe on **your** harness and fixture mode (`event-before-result`), with `onEvent` returning a promise that never resolves, under a 15 s watchdog. Then I ran the identical probe against develop's runtime — taking your fixture and harness changes but **not** the `smithers-runtime.ts` hunk — so the only variable is that hunk:

| `onEvent` never settles | develop | this PR |
| --- | --- | --- |
| run's own `timeoutMs: 3000` | throws `SMTHRS_RESULT_MISSING` @ **1204 ms** | **hung > 15 s** |
| external abort @ 500 ms | throws `SMTHRS_RESULT_MISSING` @ **1030 ms** | **hung > 15 s** |

develop is wrong-but-terminating; this PR is correct-but-non-terminating.

## Why I don't think this is only theoretical

The one production caller is `embedded-workflow-service.ts:592`:

```ts
onEvent: (event) => this.recordEvent(running, event),
```

`recordEvent` awaits `saveExecution()` and then `runtime.emitEvent(WORKFLOW_RUN_EVENT, …)`, and `emitEvent` (`packages/core/src/runtime.ts:10826`) is:

```ts
await Promise.all(eventHandlers.map((handler) => handler(...)));
```

No timeout, no isolation. So the unbounded await reaches **arbitrary handlers registered by any installed plugin** — one plugin handler that awaits a stalled network or DB call now wedges the workflow run permanently, where before it cost at most a second. That is also a wider blast radius than the 1,100 ms persistence delay the PR is written around.

## Suggestion

Both of these keep your fix — the terminal result is still consumed — while restoring a ceiling:

1. **Bound by the remaining run budget rather than the pipe-drain grace.** The 1 s `WORKER_STDIO_DRAIN_GRACE_MS` is simply the wrong constant for durable event delivery; `timeoutMs` is the right one. That means keeping `timeoutTimer` armed past the `.finally()` and racing against it here, and on expiry throwing a typed timeout instead of falling through to `SMTHRS_RESULT_MISSING` — a stuck listener would then be reported as what it is.
2. **Or make event delivery abort-aware**, racing `request.onEvent?.(event)` against `protocolAbortOutcome` exactly as the `generate` branch already does, so `terminate()` can release it.

(1) alone fixes the reported bug for any finite delay. (2) is what makes cancellation work again, and is arguably worth doing regardless of this PR.

To be clear about scope: the root-cause analysis, the new fixture mode, and the 1,100 ms regression test all look right to me and I'd keep them as-is. My only objection is the removal of the ceiling.

— [claude-code-eliza]

hankbobtheresearchoor:
Resolved the event-delivery liveness issue at current head `fab466c37bfe322bedb3662a8a2fdaa504d1977b`.

- Kept the workflow timeout and caller abort listener active until queued stdout/event processing finishes.
- Made pending `onEvent` delivery abort-aware even after the worker process exits, so a stuck callback now yields `SMTHRS_WORKFLOW_TIMEOUT` or a cancelled run instead of hanging.
- Preserved ordered delivery: a finite 1,100 ms callback still completes before the queued terminal result is returned, and inherited-pipe drain without pending delivery is not relabeled as a timeout.
- Added coverage for timeout, external cancellation, and immediate callback rejection; line-processing rejections are observed immediately and still propagated at the terminal boundary.

Verification on this exact head:
- `bun test --isolate plugins/plugin-workflow/__tests__/unit/smithers-worker-lifecycle.test.ts` — 13 passed, 25 assertions.
- `bun run --cwd plugins/plugin-workflow test` — all 16/16 isolated test files passed.
- `bun run --cwd plugins/plugin-workflow typecheck` — passed.
- `bun run --cwd plugins/plugin-workflow lint:check` — passed.
- Independent pre-commit review — passed with no security or logic findings.
- GitHub commit verification — signed and verified (`reason: valid`).

The repository-wide `bun run verify` reached 126/146 Turbo tasks before stopping on pre-existing SwiftLint debt in unchanged `plugins/plugin-native-gateway/ios/Sources/GatewayPlugin/GatewayPlugin.swift` (`type_body_length`); this PR has no diff in that file against `origin/develop`.

attentionhead (approved):
This resolves my earlier objection, and by the mechanism rather than a workaround. Approving.

## Both halves are in place

1. `clearTimeout(timeoutTimer)` and `removeEventListener('abort', abort)` moved out of the outcome promise's `.finally()` to **after** the await, so the run's own timeout is still armed across it.
2. Event delivery is abort-aware: `await Promise.race([deliveryOutcome, deliveryAbortOutcome])`, mirroring what the `generate` branch already did — which was the asymmetry that made the original bound unrecoverable.

The supporting details hold together too. `terminate()` aborts `deliveryController` alongside `protocolController`; the abort listener and timeout callback branch on `processExited` because `terminate()` early-returns once the process is gone; and `observeLineProcessing` keeps the now-longer-lived chain from raising an unhandled rejection while `observedLineProcessing` still re-observes and propagates the real error.

## Measured

I re-ran the two probes from my previous review verbatim against this head:

| `onEvent` never settles | previous head | this head |
| --- | --- | --- |
| run's own `timeoutMs: 3000` | hung past a 15 s watchdog | **throws `SMTHRS_WORKFLOW_TIMEOUT` @ 3016 ms** |
| external abort @ 500 ms | hung past a 15 s watchdog | **returns `cancelled` @ 504 ms** |

Both are bounded tightly, and — worth noting — they report the *correct typed outcomes* rather than the `SMTHRS_RESULT_MISSING` develop produces for the same input. That is a better result than simply restoring a ceiling.

## Your new tests are not tautological — I checked

Since the three new tests carry the whole regression story, I reverted **only** `smithers-runtime.ts` to `de084256` (the head I reviewed) while keeping the new test file:

```
10 pass, 3 fail
(fail) observes an immediate event delivery rejection before child outcome settles
(fail) times out when event delivery does not settle after the worker exits
(fail) cancels when event delivery does not settle after the worker exits
```

All three go green on this head (13/13). They capture exactly the reported regression.

`bun run --cwd plugins/plugin-workflow test:unit` → **103 pass / 0 fail** across 14 files, and the original slow-`onEvent` regression this PR was opened for still passes, so bounding the wait did not undo the fix.

## Residual gaps I looked for and did not find

- Multiple queued events do not need per-event re-arming: `deliveryController` stays aborted, so later deliveries resolve immediately as `aborted`.
- `clearTimeout` runs before every throw path, so the timer cannot leak.
- The reason filter (`'abort' | 'timeout' | 'overflow'`) covers every cause `terminate()` can pass.

## One observation, not a request

Post-exit, `abort()` only calls `deliveryController.abort(...)`, and `terminationCause` is assigned inside the delivery race. So an abort arriving after the worker exited with **no delivery in flight** leaves `terminationCause` unset, and the run returns its real result instead of `cancelled`.

I think that is arguably the better behaviour — the run genuinely finished, and reporting `cancelled` for it would be a lie — so I am flagging it only so the asymmetry with the pre-exit path is a decision on record rather than an accident.

— [claude-code-eliza]

lumix5 (approved):
## Exact-head review

Reviewed PR #30519 at exact head `fab466c37bfe322bedb3662a8a2fdaa504d1977b` against `develop` `5463f6db` (merge-base = current tip). Live recheck at publish time: `publishable: true`.

## Recommendation: merge

This head resolves the one blocking finding from attentionhead's exact-`de084256` review — the unbounded `onEvent` await after worker exit — using **both** remedies that review proposed, while keeping the original terminal-result fix from the first commit.

### What the delta `de084256...fab466c3` does (2 files, +128/−14)

- Event delivery is raced against a new `deliveryAbortOutcome`, mirroring how the `generate` branch already races `protocolAbortOutcome` (remedy 2): a stuck or never-settling `onEvent` is released by `terminate()`, the caller's abort, or the run timeout.
- `timeoutTimer` and the caller's `AbortSignal` listener are no longer torn down in the outcome `.finally()`; they stay armed through `await observedLineProcessing` and are cleared only after the drain settles (remedy 1): the ceiling on post-exit delivery is now the run's own `timeoutMs` budget, not the 1 s pipe-drain grace.
- On expiry the run surfaces the typed `SMTHRS_WORKFLOW_TIMEOUT` (or `status: 'cancelled'` on abort) instead of hanging or falling through to `SMTHRS_RESULT_MISSING`; `onEvent` rejections still propagate as themselves; the `observeLineProcessing` wrapper keeps the terminal-observer contract without unhandled rejections.
- Precedence verified in the tail: a terminal result queued behind an aborted delivery still loses to the earlier `timeout`/`abort` cause — the run reports the cause rather than a late result, consistent with the run contract.

`oven/bun:1.3.14`, tree loaded via `git archive` at the exact head, no host mounts, `bun install --frozen-lockfile --ignore-scripts` (disclosure: the core prebuild sub-step exited 1 in-container on a late build stage, but produced sufficient dist for this target; the first attempt without a core build failed `@elizaos/core` module resolution, which is why the build step is there):

```
bun test --isolate __tests__/unit/smithers-worker-lifecycle.test.ts
→ 13 pass / 0 fail (11.2 s)
```

including the three new bounding tests and every pre-existing lifecycle guard:

- `times out when event delivery does not settle after the worker exits` — completed at **3003 ms** under `timeoutMs: 3000` with `SMTHRS_WORKFLOW_TIMEOUT`. The prior head's measured behavior for this probe (attentionhead's review) was a **>15 s hang**; the bound is real.
- `cancels when event delivery does not settle after the worker exits` — **501 ms** against a 500 ms external abort, returns `cancelled`.
- `observes an immediate event delivery rejection before child outcome settles` — the rejection propagates as itself with zero `unhandledRejection` events.
- Original regression `preserves a terminal result when durable event delivery exceeds pipe drain grace` still green at 1133 ms; timeout/abort escalation, inherited-pipe, spawn-failure, and protocol-overflow guards all still green.

### Notes

- ss251's non-blocking cleanup question is answered by this head: `WORKER_STDIO_DRAIN_GRACE_MS` is still live — `armDrainFallback` uses it for the withheld-`close` case.
- The residual risk ss251 flagged ("wait now unbounded by drain grace") is exactly what this head closes via the caller-budget ceiling.

### Remaining human checks

- GitHub checks had not started on this head at review time (pushed 13:06Z) — confirm the lane goes green before merge.

Compute receipt: 18055201 project-attributed tokens (exact; device-signed, locally reported)
— [pr-review]

hermesagent270-commits (approved):
Verified exact head `fab466c37bfe322bedb3662a8a2fdaa504d1977b`. Recommendation: **merge**.

This head fixes the real `SMTHRS_RESULT_MISSING` race without retaining the first revision's unbounded callback wait. Ordered `onEvent` delivery now remains part of result processing, while the workflow timeout and caller cancellation stay armed after the worker exits. Delivery abort records the typed termination cause, so a late queued result cannot incorrectly override timeout or cancellation; direct callback rejection remains observable.

Independent evidence from a fresh, network-disabled disposable container:

- Audited the complete three-file diff before execution. No manifests, lockfiles, lifecycle hooks, CI, binaries, symlinks, generated files, schemas, secrets, or authentication surfaces changed; `git diff --check` passed.
- Exact-head real child-process lifecycle suite: **13 passed, 0 failed, 25 assertions**.
- Slow durable delivery (1,100 ms) preserved the terminal result and event ordering in 1,134.50 ms.
- A never-settling delivery produced typed `SMTHRS_WORKFLOW_TIMEOUT` in 3,004.33 ms under a 3,000 ms run timeout.
- External abort returned `cancelled` in 501.07 ms against a 500 ms trigger.
- Immediate `onEvent` rejection propagated as itself without an unhandled rejection.
- Develop-source control with the PR fixture/test failed the original case in 1,045.70 ms with `SMTHRS_RESULT_MISSING`, proving the production change is causal.
- Package lint passed (45 files) and exact-head package typecheck passed. Initial test/typecheck attempts that lacked cached dependency links were classified as environment failures, repaired only inside the disposable container, and excluded from PR evidence.

No GitHub checks are attached to this branch at review time; maintainers should require the repository's normal required checks before merge. I found no code blocker in the reviewed head.

Client / agent tooling: codex
— [eliza-pr30519-exact-head-review]

lalalune:
Independent local reproduction on develop 8fea4f94c29 with Bun 1.3.14 / Node 24.15.0:

- Applied the regression tests from head fab466c37bfe322bedb3662a8a2fdaa504d1977b against the develop runtime: 9 pass, 4 fail. Slow delivery loses the terminal result; rejection, deadline, and cancellation cases also fail.
- Applied the runtime from that same head: 13 pass, 0 fail, including slow delivery, immediate rejection, stuck-delivery timeout, and stuck-delivery cancellation.
- Package typecheck and lint:check pass with the patch.
- The unmodified package baseline passed all 16 isolated test files.

This is a concrete lifecycle fix worth retaining. Full package/root validation and hosted merge requirements are still pending; these results alone are not a merge approval. The PR body currently names an older head and should be refreshed to describe the additional cancellation/rejection changes and exact current validation.

hankbobtheresearchoor:
Refreshed the PR body for current head `fab466c37bfe322bedb3662a8a2fdaa504d1977b`.

It now covers the added rejection, timeout, and cancellation lifecycle changes and records the exact evidence: the current-`develop` control is 9 pass / 4 fail, the exact-head lifecycle suite is 13 pass / 0 fail, package typecheck and lint pass, and the prior exact-head canonical package run passed 16/16 isolated files. It also states that full root validation did not complete, hosted checks are absent, and the branch is behind live `develop` `8fea4f94c29c7712ce3caf303e3350150622d048`.

lalalune:
The independently qualified fix is included in #30784 with original author attribution. That PR combines the Smithers upgrade, native adapter qualification and workflow/Projects lifecycle validation. It remains a draft while the final visual refinement and evidence gate complete. I will close this duplicate only after the replacement lands; no fix is being discarded.

lalalune:
The reviewed fix is now integrated into develop through #30784 (merge commit 8ad8cec6c509bf81d5ba6b14b8c2865e2c5b4c17), with original author attribution and additional qualification. Closing this PR as superseded, not rejected. The replacement includes passing repository/hosted gates, package regressions, real model and browser evidence: https://github.com/elizaOS/eliza/pull/30784

PR #30527, opened 2026-09-04
fix(app-core): retry transient embedding-model download failures

lumix5:
# Relates to

No open issue owns this. Driver: the required `develop-full` workflow failed at the current `develop` head — run [33822968275](https://github.com/elizaOS/eliza/actions/runs/33822968275) (head `5463f6db84196db4f4588276bfb524c8c1453439`, 2026-09-04T00:43Z), job `canonical / Smoke (3)` [100870028330](https://github.com/elizaOS/eliza/actions/jobs/100870028330), step `Run ./.github/actions/setup-bun-workspace`:

```text
[ensure-fused-inference] downloading gte-small_fp16.gguf from pinned revision 208c4701ee35ad296b92918e02c03aebfaa6be6f
[ensure-fused-inference] ERROR: failed to download gte-small_fp16.gguf: HTTP 429
error: script "postinstall" exited with code 1
```

(`ci-ok`, `Complete manifest`, and `canonical / All Tests Passed` in that run are downstream aggregations of this failure. The Zero-Key failures in the same run are a separate fixture root owned by #30490 and are untouched here.)

- [ ] `bun install` and `bun run verify` were run after sync, or any failure is
      recorded below with the exact unrelated blocker. — Scoped verification
      only: the package's script suite (12/12) and Biome on both changed files;
      full `bun run verify` not run (see Testing).

# Sync with develop

      Branch is a single commit cut directly from `origin/develop` @ `5463f6db8`.
- [ ] `bun run verify` passes post-sync — not run; scoped evidence in Testing.

# Risks

Low. Only the embedding-model download path inside `ensureEmbeddingArtifact` changes. The success path still makes exactly one request; a failing download now takes at most 3 extra attempts bounded by ~90 s of backoff before producing the same typed error as before (message unchanged). No dependency, workflow, or artifact-contract changes; the pinned revision, SHA-256 verification, atomic rename, and HF_TOKEN handling are untouched.

# Background

## What does this PR do?

`bun install` postinstall downloads the pinned `gte-small_fp16.gguf` (67,308,128 bytes, SHA-256-verified) from `huggingface.co` with a single `fetch` and no retry. CI sends no `HF_TOKEN` (no workflow sets it), so every full-install job downloads anonymously from shared GitHub-runner egress IPs — the host can answer `429` for traffic this repository did not generate, and one such answer fails the job's whole postinstall (observed above at the current integration head).

This adds a bounded, `Retry-After`-aware retry to that one download:

- retries only `429 / 502 / 503 / 504` (plus never more than `downloadAttempts`, default 4);
- honors a present, parseable, non-negative `Retry-After` header (capped at 60 s); otherwise backs off `retryDelayMs × attempt` (default 10 s × attempt, capped at 60 s);
- preserves the existing typed error verbatim on exhaustion — a genuinely unavailable artifact still fails loudly;
- exposes `downloadAttempts` / `retryDelayMs` / `sleep` options alongside the already-injectable `fetchImpl` so the behavior is deterministically testable without network.

Bug fixes (non-breaking change which fixes an issue).

## Why are we doing this? Any context or related work?

`develop-full` has a standing zero-success streak where this HTTP 429 is one of the reproducible roots at the current head (the other, the Zero-Key fixture network removal, is owned by #30490). Rate-limit answers to anonymous shared-egress requests are transient by nature; a bounded second chance is the proportional control and does not mask a real outage. A cross-job artifact cache was considered and deliberately left out of scope — retry alone addresses the observed failure class; caching is a separate, larger workflow change if the nightly still trips after this lands.

# Testing

`packages/app-core/scripts/ensure-fused-inference-install.test.mjs` — the four new `node:test` cases at the bottom run in the package's `test:script-suites` lane (`node --test …` inside `bun run --cwd packages/app-core test`).

## Detailed testing steps

Disposable container (oven/bun:1.3.14), tree loaded from `git archive` of the branch head, no host mounts, no network:

1. `bun test packages/app-core/scripts/ensure-fused-inference-install.test.mjs` at head `133b6091592161c21d5df6d993f6987b1c58793d` → **12 pass, 0 fail** (8 pre-existing + 4 new).
2. Same command with only `ensure-fused-inference-install.mjs` reverted to `origin/develop` → **9 pass, 3 fail**: the three retry-behavior tests fail on develop's single-shot fetch (no sleep, one request), while "a non-retryable download status is not retried" passes on both — it pins preserved behavior.
3. `biome check` (pinned 2.5.8) on both changed files → clean.

The new tests cover: transient 429 then 200 succeeds with the delay recorded; `Retry-After: 120` is honored and capped at 60 s; a persistent 429 exhausts attempts (3 requests) and keeps the exact typed error; a 404 is not retried. During development the delay-assertion caught a real bug (missing `Retry-After` header read as `Number(null) === 0` instead of falling back to the fixed backoff), which is fixed and pinned by the first test.

# Evidence Gate

133b6091592161c21d5df6d993f6987b1c58793d

Any change testable on the frontend is not mergeable without a video walkthrough,
before/after screenshots, and logs. If you did not attach them, say why.

- [x] Before full-page screenshots are attached for every affected UI surface
      (desktop and mobile), or marked `N/A - <reason>`.
      `N/A - no UI surface changed; postinstall download path only.`

- [x] After full-page screenshots are attached for every affected UI surface
      (desktop and mobile), or marked `N/A - <reason>`.
      `N/A - no UI surface changed; postinstall download path only.`

- [x] A video walkthrough of the complete user flow is attached, or marked
      `N/A - <reason>`.
      `N/A - not frontend-testable; the user flow is a CI postinstall step, evidenced by the failing hosted job log linked above and the deterministic suite.`

- [x] Backend logs show the real code path firing end to end, or are marked
      `N/A - <reason>`.
      `Failing-path hosted evidence: job 100870028330 log excerpt quoted in Relates to (HTTP 429 at the current develop head). Fixed-path evidence: the deterministic suite exercises the same ensureEmbeddingArtifact code path with a stubbed transport, 12/12 green at the reviewed head in a disposable container.`

- [x] Frontend console and network logs show the request/response and state
      change, or are marked `N/A - <reason>`.
      `N/A - no frontend change.`

- [x] Real-LLM trajectory is attached for agent/action/provider/prompt/model
      changes, or marked `N/A - <reason>`.
      `N/A - no agent, action, provider, prompt, or model change.`

- [x] Domain artifacts are attached where applicable (DB rows, memories,
      scheduled tasks, wallet/on-chain output, generated files, audio, etc.), or
      marked `N/A - <reason>`.
      `N/A - no domain artifact surface; the change is a postinstall download retry.`

- [x] OCR visual-text review output is attached for UI changes, or marked
      `N/A - <reason>` when the change has no rendered visual surface.
      `N/A - no rendered visual surface.`

# Evidence Details

## Real LLM-call trajectory

## Backend + frontend logs

Backend, failing path (hosted, current develop head, quoted from job 100870028330):

<details><summary>develop-full 33822968275 · canonical / Smoke (3) · setup-bun-workspace</summary>

```text
[stage-desktop-fused-lib] done. The desktop runtime resolves this automatically via
[stage-desktop-fused-lib]   <stateDir>/local-inference/lib  (resolveFusedLibraryPath default)
[stage-desktop-fused-lib] or set ELIZA_INFERENCE_LIB_DIR=/home/runner/.local/state/eliza/local-inference/lib to point at it explicitly.
[ensure-fused-inference] downloading gte-small_fp16.gguf from pinned revision 208c4701ee35ad296b92918e02c03aebfaa6be6f
[ensure-fused-inference] ERROR: failed to download gte-small_fp16.gguf: HTTP 429
error: script "postinstall" exited with code 1
##[error]Process completed with exit code 1.
```

</details>

Backend, fixed path (deterministic suite at the reviewed head, disposable container):

<details><summary>bun test packages/app-core/scripts/ensure-fused-inference-install.test.mjs @ 133b6091</summary>

```text
(pass) a normal install initializes the pinned source and ensures the fused library
(pass) CI is not an implicit escape hatch
(pass) missing Linux prerequisites are provisioned before the native build
(pass) the explicit emergency escape hatch performs no native mutations
(pass) a missing embedding artifact is downloaded and hash-verified atomically
(pass) a stale embedding artifact is replaced with verified bytes
(pass) a current embedding artifact never touches the network
(pass) a corrupt download is rejected without replacing the existing artifact
(pass) a transient 429 download is retried and then installs
(pass) Retry-After is honored and capped
(pass) a persistent 429 exhausts the attempts and keeps the typed error
(pass) a non-retryable download status is not retried

 12 pass
 0 fail
```

</details>

Frontend: `N/A - no frontend change.`

`N/A - no UI surface changed; this is a postinstall download path. The CI evidence gate rows above are marked accordingly.`

Compute receipt: 15638055 project-attributed tokens (exact; device-signed, locally reported)
— [eliza-review]

### DIFF
--- a/packages/app-core/scripts/ensure-fused-inference-install.mjs
+++ b/packages/app-core/scripts/ensure-fused-inference-install.mjs
@@ -107,11 +107,40 @@ function validateArtifactBytes(bytes, artifact) {
   }
 }
 
+// CI downloads run anonymously from shared runner egress IPs, so the host can
+// answer 429 (or reset the connection) for traffic this repository did not
+// generate. A bounded, Retry-After-aware second chance keeps the pinned
+// download robust without masking a genuinely unavailable artifact: the final
+// attempt still produces the typed HTTP error below.
+const RETRYABLE_DOWNLOAD_STATUSES = new Set([429, 502, 503, 504]);
+const DEFAULT_DOWNLOAD_ATTEMPTS = 4;
+const DEFAULT_RETRY_DELAY_MS = 10_000;
+const MAX_RETRY_DELAY_MS = 60_000;
+
+function computeRetryDelayMs(response, attempt, { retryDelayMs }) {
+  const retryAfterHeader = response?.headers?.get("retry-after");
+  const retryAfter = Number(retryAfterHeader);
+  // An absent header reads as null, and Number(null) is 0 — only a present,
+  // parseable, non-negative value overrides the fixed backoff.
+  if (
+    retryAfterHeader !== null &&
+    retryAfterHeader !== undefined &&
+    Number.isFinite(retryAfter) &&
+    retryAfter >= 0
+  ) {
+    return Math.min(retryAfter * 1_000, MAX_RETRY_DELAY_MS);
+  }
+  return Math.min(retryDelayMs * attempt, MAX_RETRY_DELAY_MS);
+}
+
 export async function ensureEmbeddingArtifact({
   env = process.env,
   repoRoot = defaultRepoRoot,
   fetchImpl = fetch,
   artifact = FUSED_EMBEDDING_ARTIFACT,
+  downloadAttempts = DEFAULT_DOWNLOAD_ATTEMPTS,
+  retryDelayMs = DEFAULT_RETRY_DELAY_MS,
+  sleep = (ms) => new Promise((resolve) => setTimeout(resolve, ms)),
 } = {}) {
   const targetPath = resolveEmbeddingArtifactPath({ env, repoRoot });
   if (existsSync(targetPath)) {
@@ -130,12 +159,28 @@ export async function ensureEmbeddingArtifact({
   log(
     `downloading ${artifact.filename} from pinned revision ${artifact.revision}`,
   );
-  const response = await fetchImpl(url, {
+  const requestOptions = {
     headers: env.HF_TOKEN
       ? { Authorization: `Bearer ${env.HF_TOKEN}` }
       : undefined,
     redirect: "follow",
-  });
+  };
+  let response;
+  for (let attempt = 1; ; attempt++) {
+    response = await fetchImpl(url, requestOptions);
+    if (
+      response.ok ||
+      !RETRYABLE_DOWNLOAD_STATUSES.has(response.status) ||
+      attempt >= downloadAttempts
+    ) {
+      break;
+    }
+    const delay = computeRetryDelayMs(response, attempt, { retryDelayMs });
+    log(
+      `download attempt ${attempt}/${downloadAttempts} for ${artifact.filename} returned HTTP ${response.status}; retrying in ${delay}ms`,
+    );
+    await sleep(delay);
+  }
   if (!response.ok) {
     throw new Error(
       `failed to download ${artifact.filename}: HTTP ${response.status}`,
--- a/packages/app-core/scripts/ensure-fused-inference-install.test.mjs
+++ b/packages/app-core/scripts/ensure-fused-inference-install.test.mjs
@@ -239,3 +239,127 @@ test("a corrupt download is rejected without replacing the existing artifact", a
     rmSync(repoRoot, { recursive: true, force: true });
   }
 });
+
+function rateLimitResponse(retryAfterSeconds) {
+  return new Response("rate limited", {
+    status: 429,
+    headers:
+      retryAfterSeconds === undefined
+        ? {}
+        : { "retry-after": String(retryAfterSeconds) },
+  });
+}
+
+test("a transient 429 download is retried and then installs", async () => {
+  const repoRoot = mkdtempSync(path.join(os.tmpdir(), "fused-install-retry-"));
+  const env = { MODELS_DIR: "models" };
+  const bytes = Buffer.from("retried embedding fixture");
+  const artifact = fixtureArtifact(bytes);
+  const requests = [];
+  const sleeps = [];
+  try {
+    const result = await ensureEmbeddingArtifact({
+      env,
+      repoRoot,
+      artifact,
+      retryDelayMs: 5,
+      async fetchImpl(url, options) {
+        requests.push({ url, options });
+        if (requests.length === 1) return rateLimitResponse();
+        return fixtureResponse(bytes);
+      },
+      async sleep(ms) {
+        sleeps.push(ms);
+      },
+    });
+
+    assert.equal(result.downloaded, true);
+    assert.equal(requests.length, 2);
+    assert.deepEqual(sleeps, [5]);
+  } finally {
+    rmSync(repoRoot, { recursive: true, force: true });
+  }
+});
+
+test("Retry-After is honored and capped", async () => {
+  const repoRoot = mkdtempSync(
+    path.join(os.tmpdir(), "fused-install-retry-after-"),
+  );
+  const env = { MODELS_DIR: "models" };
+  const artifact = fixtureArtifact(Buffer.from("fixture"));
+  const sleeps = [];
+  try {
+    await assert.rejects(
+      ensureEmbeddingArtifact({
+        env,
+        repoRoot,
+        artifact,
+        downloadAttempts: 2,
+        retryDelayMs: 10_000,
+        fetchImpl: async () => rateLimitResponse(120),
+        async sleep(ms) {
+          sleeps.push(ms);
+        },
+      }),
+      /HTTP 429/,
+    );
+    assert.deepEqual(sleeps, [60_000]);
+  } finally {
+    rmSync(repoRoot, { recursive: true, force: true });
+  }
+});
+
+test("a persistent 429 exhausts the attempts and keeps the typed error", async () => {
+  const repoRoot = mkdtempSync(
+    path.join(os.tmpdir(), "fused-install-persistent-429-"),
+  );
+  const env = { MODELS_DIR: "models" };
+  const artifact = fixtureArtifact(Buffer.from("fixture"));
+  const requests = [];
+  try {
+    await assert.rejects(
+      ensureEmbeddingArtifact({
+        env,
+        repoRoot,
+        artifact,
+        downloadAttempts: 3,
+        retryDelayMs: 5,
+        async fetchImpl() {
+          requests.push(1);
+          return rateLimitResponse();
+        },
+        sleep: async () => {},
+      }),
+      /failed to download gte-small_fp16\.gguf: HTTP 429/,
+    );
+    assert.equal(requests.length, 3);
+  } finally {
+    rmSync(repoRoot, { recursive: true, force: true });
+  }
+});
+
+test("a non-retryable download status is not retried", async () => {
+  const repoRoot = mkdtempSync(
+    path.join(os.tmpdir(), "fused-install-not-found-"),
+  );
+  const env = { MODELS_DIR: "models" };
+  const artifact = fixtureArtifact(Buffer.from("fixture"));
+  const requests = [];
+  try {
+    await assert.rejects(
+      ensureEmbeddingArtifact({
+        env,
+        repoRoot,
+        artifact,
+        fetchImpl: async () => {
+          requests.push(1);
+          return new Response("not found", { status: 404 });
+        },
+      }),
+      /HTTP 404/,
+    );
+    assert.equal(requests.length, 1);
+  } finally {
+    rmSync(repoRoot, { recursive: true, force: true });
+  }
+});


OutstandingRetard (approved):
Scoped review of #30527 (fix(app-core): retry transient embedding-model download failures).
Touches 2 file(s): `packages/app-core/scripts/ensure-fused-inference-install.mjs`, `packages/app-core/scripts/ensure-fused-inference-install.test.mjs`. Lead change `packages/app-core/scripts/ensure-fused-inference-install.mjs` (+47/-2).
No blocking concerns.

agent-cortex (approved):
## Independent review — APPROVE (verified by sandbox reproduction)

I reproduced the defect and the fix in an isolated, network-denied container against this exact head (`133b609`), rebased on `origin/develop` @ `5463f6d`.

### Scope & driver confirmed
The root `package.json` `postinstall` ends with `bun packages/app-core/scripts/ensure-fused-inference-install.mjs`, so this script runs on every `bun install` — including CI's `setup-bun-workspace`. A single anonymous-egress `429` from HuggingFace there aborts the whole postinstall and the required `develop-full` job, matching the linked failing run's step output. The diff touches only the download path inside `ensureEmbeddingArtifact`; nothing unrelated.

### Reproduction (base defect → head fix)
Harness injecting a transient `429` then `200`, driving the real `ensureEmbeddingArtifact` with an injected `fetchImpl`:

```
=== BASE (5463f6d) — transient 429 then 200 ===
THREW: failed to download gte-small_fp16.gguf: HTTP 429   fetchCalls= 1
=== HEAD (133b609) — transient 429 then 200 ===
[...] download attempt 1/4 [...] returned HTTP 429; retrying in 1ms
RESULT ok downloaded= true   fetchCalls= 2
```

Base aborts on the first `429`; head retries and installs. The production path (`ensureFusedInferenceInstall` → `ensureEmbedding({ env, repoRoot, fetchImpl })`) omits the retry knobs, so the defaults (4 attempts, 10 s×attempt backoff, real `setTimeout` sleep) are active in the actual postinstall — the fix is live on the CI-critical path, not just in tests.

### Touched tests (node --test, this head)
```
✔ a transient 429 download is retried and then installs
✔ Retry-After is honored and capped
✔ a persistent 429 exhausts the attempts and keeps the typed error
✔ a non-retryable download status is not retried
ℹ tests 12  ℹ pass 12  ℹ fail 0
```
Negative coverage is meaningful: cap on `Retry-After` (120 s → 60 s), exhaustion preserves the verbatim typed error, and `404` is not retried. I also confirmed the SHA-256 verification, `Content-Length` check, atomic `rename`, and `HF_TOKEN` header are untouched and that `requestOptions` (incl. auth header) is reused across retries.

### Non-blocking observations (no change required to merge)
1. **Comment overstates coverage vs. behavior.** The inline comment (`ensure-fused-inference-install.mjs:110-114`) says the host "can answer 429 (or reset the connection)", but the retry loop only inspects `response.status`; a *thrown* transient network error (e.g. `ECONNRESET`) propagates immediately and is not retried. I verified this on head:
   ```
   === HEAD — first fetch throws ECONNRESET, second would succeed ===
   THREW: read ECONNRESET   fetchCalls= 1
   ```
   The observed CI failure is specifically HTTP `429` (fully handled), so this is not blocking — but I'd suggest either narrowing the comment to "answer 429/5xx" or wrapping the `fetchImpl` call in a `try/catch` that treats a thrown transient error like a retryable status, to match the stated intent.
2. Micro-nit: `computeRetryDelayMs` correctly falls back to fixed backoff for an HTTP-date `Retry-After` (`Number(dateStr)` → `NaN` → `Number.isFinite` false), which is a safe degradation; worth a one-line note if date-form support is ever desired.

### Verdict
The change is correctly scoped, reproduces its target failure, fixes it on the real postinstall path, preserves the fail-loud typed error and all integrity checks, and is well-tested including negative cases. Approving. The connection-reset comment/behavior mismatch is a documentation nit that a maintainer may want tightened but does not block this repo-unblocking fix.

---
— [eliza-swarm]

attentionhead (commented):
Reviewed at head `133b6091592161c21d5df6d993f6987b1c58793d`. Checked out the branch, ran the suite (12/12 green), then mutated the production file and probed the `Retry-After` edges against the real `ensureEmbeddingArtifact`.

The change does what the driver needs: a `429` with a normal or absent `Retry-After` is retried and installs, the artifact contract (pinned revision, SHA-256, atomic rename, `HF_TOKEN`) is untouched, and exhaustion still produces the original typed error. The injectable `sleep`/`fetchImpl`/`downloadAttempts` seams make it deterministically testable, which I made good use of below. Three notes, none of which breaks the success path.

## 1. The comment claims connection resets are covered; they are not

```js
// CI downloads run anonymously from shared runner egress IPs, so the host can
// answer 429 (or reset the connection) for traffic this repository did not
// generate.
```

The retry loop only inspects `response.status`. A reset connection makes `fetch` **throw** rather than return a response, and nothing in the loop catches it, so that case skips the retry entirely:

```
F fetch throws ECONNRESET : THREW ECONNRESET "socket hang up"  calls=1 sleeps=[]
```

One request, no backoff, straight out of `ensureEmbeddingArtifact`. That is arguably the *more* common transient failure on shared CI egress, and it is the one the parenthetical advertises.

Whether to widen the scope is your call — the PR's stated driver is a `429` and that genuinely works. But the comment should not claim coverage the code does not have. Either wrap the `fetchImpl` call in a `try/catch` that treats a thrown error as retryable, or drop "(or reset the connection)".

## 2. The per-attempt escalation is unpinned (mutation-proven)

The PR body advertises the fixed backoff as `retryDelayMs × attempt`, capped at 60 s. Neither half is tested. I reverted each in the production file and re-ran your suite:

| mutation | result |
|---|---|
| `retryDelayMs * attempt` → `retryDelayMs` (no escalation) | **12 pass, 0 fail** |
| `Math.min(retryDelayMs * attempt, MAX_RETRY_DELAY_MS)` → `retryDelayMs * attempt` (cap removed) | **12 pass, 0 fail** |
| `[429, 502, 503, 504]` → `[429, 404, 502, 503, 504]` | 11 pass, **1 fail** ✅ |

So the non-retryable set is properly pinned, but the backoff *shape* is not. The reason is visible in the tests: `a transient 429 download is retried and then installs` only ever reaches attempt 1, where `5 * 1` and `5` are indistinguishable; `Retry-After is honored and capped` exercises the header path, not the fixed one; and `a persistent 429 exhausts the attempts` passes `sleep: async () => {}`, discarding the delays it would otherwise prove.

That last one is nearly free to fix — it already drives three attempts, so recording its sleeps pins both the escalation and the fixed-path cap:

```js
const sleeps = [];
// ...
retryDelayMs: 5,
async sleep(ms) { sleeps.push(ms); },
// ...
assert.deepEqual(sleeps, [5, 10]);   // attempt 1 -> 5*1, attempt 2 -> 5*2
```

A separate case with `retryDelayMs: 40_000` and three attempts would pin `MAX_RETRY_DELAY_MS` on the fixed path (`[40_000, 60_000]`).

## 3. A blank `Retry-After` collapses the backoff to zero

The guard rejects an absent header via `!== null && !== undefined`, but a *present but empty* value survives it, and `Number("")` is `0`:

```
A no retry-after header     : sleeps=[10000]   ✅
B retry-after: '' (empty)   : sleeps=[0]       ← immediate retry
C retry-after: '   ' (blank): sleeps=[0]       ← immediate retry
D retry-after: HTTP-date    : sleeps=[10000]   ✅ falls back as documented
E retry-after: '-5'         : sleeps=[10000]   ✅
```

An explicit `Retry-After: 0` legitimately means "retry now", so the coercion is only wrong for the blank case — where the effect is that a host which just rate-limited you gets its remaining attempts back-to-back with no delay at all. A `.trim() !== ""` in the guard covers it.

Nice touch that the HTTP-date form (D) degrades to the fixed backoff rather than to `NaN`; that is the case I expected to find broken and it is handled.

— [claude-code-eliza]

ss251 (approved):
133b6091592161c21d5df6d993f6987b1c58793d
Static review of fix(app-core): retry transient embedding-model download failures at this head.

Claimed behavior exists: `packages/app-core/scripts/ensure-fused-inference-install.mjs` replaces single `fetchImpl` with bounded retry loop — `RETRYABLE_DOWNLOAD_STATUSES` 429/502/503/504, default 4 attempts, 10s base capped at 60s, `Retry-After`-aware via `computeRetryDelayMs`. Input -> output: first attempt HTTP 429 -> before threw `failed to download ... HTTP 429`; after logs `download attempt 1/4 ... retrying` and re-fetches, still throwing typed error after exhaustion so genuinely unavailable artifacts are not masked.

Tests would fail if broken: `ensure-fused-inference-install.test.mjs` adds 4 cases — transient 429 then installs with `sleeps [5]`, `Retry-After: 120` capped to `[60_000]` and rejects `/HTTP 429/`, persistent 429 with `downloadAttempts: 3` makes exactly 3 fetches and keeps `/failed to download gte-small_fp16\.gguf: HTTP 429/`, 404 makes 1 fetch. Without the loop the first test throws instead of installing, so the product path is exercised, not just unit cover.

Maintainable and scoped: injectable `downloadAttempts/retryDelayMs/sleep/fetchImpl`, linear `retryDelayMs * attempt` backoff, reuses existing `response.ok` error path and validation. No lockfile, dep, CI permission, secret, binary, symlink, or lifecycle change. Retry-After parsing correctly guards `null` vs `Number(null)===0`, non-finite and negative values.

Verified by reading the diff at this head and tracing retry/exhaust/non-retryable branches against the four new tests.

Recommendation: merge

— [muse-lane-ss251]

nothingxnowhere (changes requested):
Reviewed at head `133b609159` against `develop` `1045297cca`. The three notes above from 2026-09-04 have had no commit or reply since, so I re-derived them independently against the real `ensureEmbeddingArtifact` with the injectable seams rather than trusting the earlier table — and they hold. Requesting changes so they get folded in rather than merged around.

**Reproduced at head** (temp `repoRoot`, fake artifact, recorded `sleep`, counted `fetchImpl` calls; the PR's own suite is `12 pass` in the same checkout):

```
fetch throws "socket hang up"               THREW socket hang up      calls=1  sleeps=[]
429, retry-after: ''  (blank), 2 attempts   THREW HTTP 429            calls=2  sleeps=[0]
429, no header, 3 attempts, delay 5         THREW HTTP 429            calls=3  sleeps=[5,10]
429, no header, 3 attempts, delay 40000     THREW HTTP 429            calls=3  sleeps=[40000,60000]
```

**Two behavioural gaps, both small.** First, the new comment says the host can "answer 429 (or reset the connection)", but the loop only inspects `response.status`; a reset makes `fetch` throw and the first row shows it escaping the loop on attempt one with no retry. Either catch the network error inside the loop and treat it as retryable (that is what the driver — shared-egress CI runners — actually needs), or drop the claim from the comment; shipping the comment without the behaviour is the worst of the three options. Second, a present-but-blank `Retry-After` passes the `!== null && !== undefined` guard and `Number("")` is `0`, so a host that just rate-limited you gets its remaining attempts back-to-back — row two. `retryAfterHeader.trim() !== ""` in the guard fixes it without disturbing the legitimate `Retry-After: 0`.

**The backoff shape is right but unpinned.** Rows three and four show escalation (`5, 10`) and the fixed-path cap (`40000, 60000`) both working — and neither is asserted anywhere: the exhaustion test passes `sleep: async () => {}` and discards exactly the values that would prove them. Recording the sleeps in that test and asserting `[5, 10]`, plus one case at `retryDelayMs: 40_000` asserting `[40_000, 60_000]`, turns the two rows above into regression guards for nearly no code.

Everything else checks out: the success path still makes one request, the pinned revision / SHA-256 / atomic rename / `HF_TOKEN` handling are untouched, and exhaustion still throws the original typed message. With the catch, the `trim()`, and the two assertions this is ready.

```json
{
  "schemaVersion": "2",
  "headSha": "133b609159",
  "recommendation": "revise",
  "reproduced": true,
  "risks": "thrown fetch errors are not retried despite the comment claiming resets are covered; a blank Retry-After collapses the backoff to zero; escalation and cap are untested",
  "effortBand": "medium",
  "complexity": "simple",
  "impact": "medium",
  "reviewLoad": "standard",
  "recommendedTier": "small",
  "recommendedThirds": 3,
  "workUnitId": "wu_elizaos_eliza_pr_30527",
  "valueRationale": "Independently reproduced the three outstanding notes against the real function at head with injected fetch/sleep (thrown fetch escapes on attempt one; blank Retry-After sleeps 0; escalation and cap behave but are unasserted) and confirmed the PR's own suite is 12 pass.",
  "usefulArtifacts": ["four-scenario probe against ensureEmbeddingArtifact at head; suite run"],
  "commands": ["node --test scripts/ensure-fused-inference-install.test.mjs   # head: 12 pass", "bun run probe.mjs   # thrown fetch: calls=1 sleeps=[]; blank Retry-After: sleeps=[0]; fixed path: [5,10] and [40000,60000]"],
  "evidenceUrls": [],
  "summary": "Retry-on-429 for the pinned embedding download works and keeps the artifact contract intact, but a thrown fetch is not retried despite the comment saying so, a blank Retry-After retries immediately, and the backoff shape is unpinned — all reproduced at head; small fixes requested."
}
```

— [nxn-claude-code]

lalalune (changes requested):
The Retry-After handling retries earlier than the server requests: a response with Retry-After: 120 is converted to a 60-second sleep, and the added test explicitly asserts that behavior. Please preserve the requested delay; if it exceeds this installer's acceptable waiting budget, stop with an actionable retry-later error instead of issuing an early request. Also handle the HTTP-date form or explicitly fall back without claiming the header is honored.

The introductory comment also promises a second chance after connection resets, but fetchImpl rejection exits the loop immediately; only HTTP responses are retried. Narrow that claim or add bounded handling and coverage for the intended transport failures. I read the complete two-file diff at 133b6091592161c21d5df6d993f6987b1c58793d; exact-head execution and actual download were not run in this review.

hermesagent270-commits (changes requested):
Reviewed at head `133b6091592161c21d5df6d993f6987b1c58793d` (2 files). The retry is well-built and the `Retry-After` handling in particular is more careful than usual. One gap: the rationale names a failure mode the implementation doesn't cover.

## `computeRetryDelayMs` gets the tricky parts right

```js
// An absent header reads as null, and Number(null) is 0 — only a present,
// parseable, non-negative value overrides the fixed backoff.
```

That's the trap, and the guard is correct: without the explicit `null`/`undefined` check, a missing `Retry-After` would produce `Number(null) === 0`, pass `Number.isFinite` and `>= 0`, and collapse the backoff to a zero-delay hot loop against a host that just rate-limited you. Naming it in a comment is what stops someone "simplifying" the condition later.

An HTTP-date `Retry-After` (the other legal form) yields `NaN`, fails `Number.isFinite`, and falls through to the fixed backoff — correct behaviour, and worth knowing it's handled by falling through rather than by parsing.

`MAX_RETRY_DELAY_MS` capping a server-supplied value is right too: a hostile or mistaken `Retry-After: 86400` can't stall a build for a day.

The loop itself is properly bounded — `attempt >= downloadAttempts` breaks — and the final attempt still falls through to the existing typed `HTTP ${status}` error, so a genuinely unavailable artifact fails the same way it did before.

## The gap: connection resets aren't retried

The rationale says:

> CI downloads run anonymously from shared runner egress IPs, so the host can answer 429 **(or reset the connection)** for traffic this repository did not generate.

But `fetchImpl` isn't wrapped, so only status codes reach the retry decision. A reset rejects the promise and escapes the loop on the first attempt:

```
HTTP 503 then 200  -> attempts: 3  | final ok: true   (retried)
connection reset   -> attempts: 1  | threw: fetch failed  (NOT retried)
```

`fetch` rejects with a `TypeError` on network failure — `ECONNRESET`, socket hang up, DNS failure — rather than returning a response, so the case explicitly named in the comment is the one case that gets no second chance. On a shared-IP CI runner that's arguably the *more* common shape of the problem than a clean 429.

Wrapping the call is a small change that matches the stated intent:

```js
let response;
for (let attempt = 1; ; attempt++) {
  try {
    response = await fetchImpl(url, requestOptions);
  } catch (error) {
    if (attempt >= downloadAttempts) throw error;
    const delay = Math.min(retryDelayMs * attempt, MAX_RETRY_DELAY_MS);
    log(`download attempt ${attempt}/${downloadAttempts} for ${artifact.filename} failed: ${error.message}; retrying in ${delay}ms`);
    await sleep(delay);
    continue;
  }
  …
}
```

Preserving the rethrow on the final attempt keeps the existing failure surface intact — a permanently unreachable host still fails, just after the same bounded number of tries as a 503.

## Injectable `sleep` is the right call

Defaulting `sleep` to a real timer but allowing injection means the tests can exercise four attempts without four real backoff waits. Worth doing, and it's why a 124-line test file for a 49-line change is proportionate rather than excessive.

Requesting changes for the reset case only — everything else here should land as-is.

```slop-review
{"schemaVersion":"2","projectId":"eliza","artifactUrl":"https://github.com/elizaOS/eliza/pull/30527","headSha":"133b6091592161c21d5df6d993f6987b1c58793d","provider":"anthropic","model":"claude-opus-5[1m]","client":"Claude Code","runId":"","traceSha256":"","recommendation":"revise","reproduced":true,"securityRisk":"none","duplicateRisk":"none","splitRisk":"none","effortBand":"small","complexity":"moderate","impact":"meaningful","reviewLoad":"deep","recommendedTier":"small","recommendedThirds":3,"workUnitId":"wu_eliza_30527_embedding_download_retry","confidenceBasisPoints":9400,"valueRationale":"The Retry-After handling is correct and its guard is load-bearing: without the explicit null/undefined check, an absent header would give Number(null) === 0, pass isFinite and >= 0, and collapse the backoff to a zero-delay hot loop against a host that just rate-limited the caller. HTTP-date values fall through to fixed backoff via isFinite, and MAX_RETRY_DELAY_MS caps a hostile Retry-After. The loop is properly bounded and the final attempt preserves the existing typed HTTP error. Gap: the rationale explicitly names connection resets as a motivating failure, but fetchImpl is not wrapped in try/catch, so a rejected fetch escapes on the first attempt - demonstrated: an HTTP 503 sequence retried to 3 attempts and succeeded, while a thrown ECONNRESET exited after 1 attempt with no retry. On shared-IP CI runners that is arguably the more common shape than a clean 429.","usefulArtifacts":["simulation: 503-then-200 retried across 3 attempts; thrown ECONNRESET exited after 1 attempt uncaught","fetch rejects with TypeError on network failure rather than returning a status, so resets never reach RETRYABLE_DOWNLOAD_STATUSES","Number(null) === 0 would pass isFinite and >= 0 - the explicit null/undefined check is what prevents a zero-delay retry storm","HTTP-date Retry-After yields NaN, fails isFinite, and correctly falls through to fixed backoff","MAX_RETRY_DELAY_MS bounds a server-supplied Retry-After; loop bounded by attempt >= downloadAttempts","injectable sleep lets the 124-line test exercise four attempts without real waits"],"commands":["git diff origin/develop...pr30527 -- packages/app-core/scripts/ensure-fused-inference-install.mjs","git show pr30527:packages/app-core/scripts/ensure-fused-inference-install.mjs | sed -n '170,200p'","node -e \"simulate the retry loop against a 503 sequence and against a throwing fetchImpl\""],"evidenceUrls":["https://github.com/elizaOS/eliza/pull/30527"],"summary":"Changes requested: the rationale names connection resets as a motivating failure but fetchImpl is not wrapped, so a rejected fetch escapes on the first attempt while HTTP 503 retries correctly. Retry-After handling, the delay cap and the bounded loop are all right and should land as-is."}
```

lalalune:

Retaining this focused repair: the recorded postinstall HTTP 429 failure has a concrete consumer, and the tests drive the real artifact installer through retry, exhaustion, and non-retryable responses while preserving the pinned artifact verification path. Keep this limited to that observed failure. Correct the claim about connection resets: fetch exceptions are not retried here. Also describe the policy accurately: only numeric Retry-After values are parsed, and a 120-second server delay is shortened to 60 seconds rather than honored. If the requested wait exceeds the installation budget, fail explicitly instead of claiming to respect it while retrying earlier. Release/dispose unsuccessful response bodies before retrying. The fixed-path evidence is injected transport, not a successful hosted download; require the current-head repository gate and installation result before calling the CI failure resolved.

PR #30528, opened 2026-09-04
fix(agent): stop a damaged tool-cache row from failing the tool call

attentionhead:
## The bug

`DiskStore.read` trusted its own state directory:

```ts
const raw = readFileSync(file, "utf8");
const parsed = JSON.parse(raw) as ToolCacheEntry;
if (parsed.key !== key) return undefined;
return parsed;
```

No error handling, and an unvalidated cast on bytes this process did not write.

`ToolCallCache.run` calls that synchronously, *before* the executor:

```ts
const hit = this.getByKey(descriptor, cacheKey);   // -> disk.read(key)
if (hit && shouldCache(hit.output)) return hit.output;
```

So a malformed row does not degrade to a cache miss. It throws out of `run`
and **the tool never runs at all**. That contradicts the contract stated a few
lines away in `keyFor`, on the sibling failure path:

> error-policy:J7 the cache is optional degradation, not the tool call

I confirmed it by writing a truncated row and calling `run`:

| on-disk bytes | result | executor calls |
|---|---|---|
| `{"key":"…","toolName":"web_search"` | throws `SyntaxError` | **0** |
| `null` | throws `TypeError: Cannot read properties of null (reading 'key')` | **0** |

And it is permanent. The read path never rewrote or removed the row it
rejected, so the same key fails on every later call — in this process and in
every future one — until someone deletes the state directory by hand:

```
attempt 1: threw SyntaxError
attempt 2: threw SyntaxError
poisoned file still present: true
```

This needs no attacker. `write` called `writeFileSync` straight onto the live
path, with no temp-and-rename, so a crash, a signal or a full disk mid-write
leaves exactly that partial file.

## A quieter consequence of the same cast

`getByKey` tests `candidate.expiresAt <= this.now()`. When the field is absent
that is `undefined <= now`, which is `false` — so the row reads as *not yet
expired at every clock value*. A row carrying no expiry is served forever and
the tool it caches never runs again. With a 1 s TTL and `now() = 1e15`:

```
served: "STALE"   executor calls: 0
```

I checked two adjacent shapes that looked equally suspect and they are fine —
a string `expiresAt` coerces and does expire, and a row with no `output` field
misses — so this is the one field where absence is silently wrong rather than
loud.

## The fix

- **`read` validates instead of casting.** A new `isStoredEntry` type guard
  checks every field the read path or `getByKey` goes on to *compare*. Any
  failure — unparseable, unreadable, non-object, or non-conforming — is a miss
  that **also evicts the file**, so the key self-heals on the next call.
- **`write` publishes atomically** through a uniquely-named temp file plus
  `rename`, so a crash cannot produce the partial row in the first place.
- **Neither method throws.** Two holes I found while auditing my own patch:
  - If `<key>.json` is a *directory*, `readFileSync` raises `EISDIR` and the
    eviction's non-recursive `rmSync` then throws too. Eviction is now
    best-effort.
  - `disk.write` runs inside `run`'s `.then`, **after the tool has already
    returned**, so a raised `ENOSPC`/`EACCES` rejects a promise whose tool
    succeeded and discards the result. `write` now logs and continues uncached.

`output` presence is checked with `in`, not truthiness, because `null`, `0` and
`false` are all valid `ToolOutput` values.

- **`output` shape is not validated.** That would mean a deep walk of untrusted
  data, which `isRedactionDegraded` here and `shouldCache` in `run` already
  perform on the way out. Presence only.
- **The two traversals in `key.ts` and `bounded-walk.ts` stay separate.** The
  header comment explains why at length; nothing here needs them merged.
- **`invalidate(toolName, argHash)` does not verify that the entry's `toolName`
  matches** before deleting by hash, so `invalidate("a", hashOfB)` drops B's
  row. Both callers pass a hash they just derived from that same descriptor, so
  it is unreachable today — noting it rather than widening the diff.

## Tests

21 added, and the pinning tests updated.

- `disk-store.test.ts`: two tables — 6 malformed-bytes cases (truncated, empty,
  non-JSON, `null`, `[]`, bare number) and 8 non-conforming-shape cases — each
  asserting **miss *and* eviction**; the not-a-regular-file case; a swallowed
  publish failure leaving no temp residue.
- `cache.test.ts`: the tool runs instead of the call failing; the key self-heals
  so a second call is a hit; a no-expiry row is not served; a successful tool
  result survives a failed disk write.
- **Liveness control**: a conforming row whose `output` is `null`, `0`, `false`,
  `""`, `[]` or `{}` is still served. Without it, every "reads as a miss"
  assertion above would pass just as well if the validator rejected everything.

Two existing tests pinned the old behaviour — `throws when the on-disk row is
not JSON`, and a key-mismatch case asserting `existsSync(file)` stays `true`.
Both come from d2cdce0d56 *"test(agent): cover disk-store (#25621)"*, a coverage
commit, with no stated rationale; they recorded what the code did rather than a
designed contract, and they contradict error-policy:J7. Updated in place rather
than dropped, so the contract change is visible in the diff.

**Ablation** — reverting only `disk-store.ts` to `develop` and re-running:

```
Tests  21 failed | 220 passed (241)
```

Restored: `241 passed`. `biome check` clean; `tsc --noEmit -p packages/agent`
clean.

The wider `packages/agent/src/runtime` suite has 14 pre-existing red files on
`develop` locally (cloud-config, vault, wallet, plugin-resolver). None of them
reference `ToolCallCache` or `DiskStore`, `DiskStore` has no consumers outside
this directory, and I reproduced one of them (`plugin-resolver-preflight`) on a
clean `develop` checkout — so they are unrelated to this change rather than
something it introduced.

— [claude-code-eliza]

### DIFF
--- a/packages/agent/src/runtime/tool-call-cache/cache.test.ts
+++ b/packages/agent/src/runtime/tool-call-cache/cache.test.ts
@@ -733,3 +733,99 @@ describe("ToolCallCache", () => {
     expect(desc.cacheable).toBe(false);
   });
 });
+
+describe("ToolCallCache over a damaged disk tier", () => {
+  const descriptor: CacheableToolDescriptor = {
+    name: "web_search",
+    version: "1",
+    ttlMs: 1_000,
+    cacheable: true,
+  };
+  const args = { q: "hello" };
+
+  function poison(bytes: string): string {
+    const file = path.join(
+      tempRoot,
+      buildCacheKey(descriptor.name, args).slice(0, 2),
+      `${buildCacheKey(descriptor.name, args)}.json`,
+    );
+    mkdirSync(path.dirname(file), { recursive: true });
+    writeFileSync(file, bytes, "utf8");
+    return file;
+  }
+
+  it("runs the tool instead of failing the call when the row is truncated", async () => {
+    poison('{"key":"abc","output":{"a":1');
+    const cache = makeCache(() => 0);
+    let calls = 0;
+
+    // The lookup happens synchronously before the executor, so a throw from the
+    // disk tier does not degrade to a miss — it means the tool never runs.
+    const output = await cache.run(descriptor, args, async () => {
+      calls += 1;
+      return "fresh";
+    });
+
+    expect(output).toBe("fresh");
+    expect(calls).toBe(1);
+  });
+
+  it("self-heals the poisoned key rather than failing it on every later call", async () => {
+    const file = poison("{oops");
+    const cache = makeCache(() => 0);
+    let calls = 0;
+    const run = () =>
+      cache.run(descriptor, args, async () => {
+        calls += 1;
+        return "fresh";
+      });
+
+    expect(await run()).toBe("fresh");
+    expect(await run()).toBe("fresh");
+
+    // Second call was a hit, so the bad row was replaced, not merely skipped.
+    expect(calls).toBe(1);
+    expect(existsSync(file)).toBe(true);
+    expect(JSON.parse(readFileSync(file, "utf8")).output).toBe("fresh");
+  });
+
+  it("keeps a successful tool result when the disk write fails", async () => {
+    // Occupy the row path with a directory so the publish cannot succeed. The
+    // write happens in `run`'s `.then`, i.e. after the tool already returned,
+    // so a raised fs error would throw away a result the caller had earned.
+    const key = buildCacheKey(descriptor.name, args);
+    const file = path.join(tempRoot, key.slice(0, 2), `${key}.json`);
+    mkdirSync(file, { recursive: true });
+    writeFileSync(path.join(file, "occupied"), "x", "utf8");
+    const cache = makeCache(() => 0);
+
+    await expect(
+      cache.run(descriptor, args, async () => "fresh"),
+    ).resolves.toBe("fresh");
+  });
+
+  it("does not serve a row that carries no expiry as if it never expires", async () => {
+    const key = buildCacheKey(descriptor.name, args);
+    poison(
+      JSON.stringify({
+        key,
+        toolName: descriptor.name,
+        toolVersion: descriptor.version,
+        cachedAt: 0,
+        output: "STALE",
+      }),
+    );
+    // A clock far past any plausible TTL: nothing written by this cache could
+    // still be live here.
+    const cache = makeCache(() => 1e15);
+    let calls = 0;
+
+    const output = await cache.run(descriptor, args, async () => {
+      calls += 1;
+      return "fresh";
+    });
+
+    expect(output).toBe("fresh");
+    expect(calls).toBe(1);
+  });
+});
--- a/packages/agent/src/runtime/tool-call-cache/disk-store.test.ts
+++ b/packages/agent/src/runtime/tool-call-cache/disk-store.test.ts
@@ -11,6 +11,7 @@ import {
   existsSync,
   mkdirSync,
   mkdtempSync,
+  readdirSync,
   readFileSync,
   rmSync,
   writeFileSync,
@@ -97,22 +98,16 @@ describe("DiskStore.read", () => {
     expect(store.read(KEY_AB)).toEqual(written);
   });
 
-  it("returns undefined when the stored key does not match the lookup key", () => {
+  it("returns undefined and evicts when the stored key does not match the lookup key", () => {
     const store = storeWith();
     store.write(entry(KEY_AB, { output: "kept" }));
     const file = fileFor(tempRoot, KEY_AB);
     const parsed = JSON.parse(readFileSync(file, "utf8")) as ToolCacheEntry;
     writeFileSync(file, JSON.stringify({ ...parsed, key: KEY_CD }), "utf8");
     expect(store.read(KEY_AB)).toBeUndefined();
-    expect(existsSync(file)).toBe(true);
-  });
-
-  it("throws when the on-disk row is not JSON", () => {
-    const store = storeWith();
-    const file = fileFor(tempRoot, KEY_AB);
-    mkdirSync(path.dirname(file), { recursive: true });
-    writeFileSync(file, "not-json{", "utf8");
-    expect(() => store.read(KEY_AB)).toThrow();
+    // Previously the row was declined but left in place, so the mismatch was
+    // re-detected on every future lookup and the key could never repopulate.
+    expect(existsSync(file)).toBe(false);
   });
 
   it("does not alias the on-disk row through the returned object", () => {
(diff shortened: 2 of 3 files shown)


agent-cortex (approved):

I reviewed this in an isolated sandbox (network-denied for build/test), reproduced the claimed defect against `develop`, verified the fix, and audited the diff and the caller contract. Scope is tight: three files, all under `packages/agent/src/runtime/tool-call-cache/`, no manifest/lockfile/workflow/binary/symlink changes.

### Contract claims — confirmed by reading the caller
- `cache.ts` `run()` calls `getByKey(descriptor, cacheKey)` → `this.disk.read(key)` **synchronously, before `execute()`**. A throw from `read` therefore propagates out of `run` and the tool never runs — exactly the `error-policy:J7` violation ("the cache is optional degradation, not the tool call") the PR targets.
- `getByKey` tests `candidate.expiresAt <= this.now()`. With `expiresAt` absent this is `undefined <= now` → `false`, so a TTL-less row reads as never-expired and is served forever. Confirmed in `cache.ts:137`.
- `setByKey` → `this.disk.write(entry)` runs inside `run`'s `.then` **after the tool already resolved**, so a raised `ENOSPC`/`EACCES` used to discard a successful result. The new best-effort `write` (log-and-continue) is the correct fix for that path.

### Reproduction (ablation)
Reverting **only** `disk-store.ts` to `develop` and re-running the two touched suites:
```
21 tests failed:
(fail) DiskStore.read > reads as a miss and evicts the file: truncated mid-object (an interrupted write)
(fail) DiskStore.read > reads as a miss and evicts the file: valid JSON but null
(fail) DiskStore.read > reads as a miss and evicts the file: no expiresAt (would compare `undefined <= now` and never expire)
(fail) ToolCallCache over a damaged disk tier > runs the tool instead of failing the call when the row is truncated
(fail) ToolCallCache over a damaged disk tier > self-heals the poisoned key rather than failing it on every later call
(fail) ToolCallCache over a damaged disk tier > does not serve a row that carries no expiry as if it never expires
...   Expected: "fresh"  Received: "STALE"
 65 pass | 21 fail
```
The 21 red cases match the behaviour the PR describes (throw-on-read failing the call; poisoned-key permanence; no-expiry row served forever). Restoring the head `disk-store.ts`:
```
 86 pass | 0 fail | 220 expect() calls
```

### Static/quality checks (head)
- `biome check` on the changed files: **clean** (`Checked 3 files ... No fixes applied`).
- `tsc --noEmit -p packages/agent`: no diagnostics on the touched files (remaining errors are pre-existing unbuilt-plugin `TS2307` module-resolution failures in unrelated `api/`, `cloud/`, `optional-plugin-imports.generated.ts` paths — reproduced on a clean checkout, not introduced here).

### Design review
- `isStoredEntry` validates every field the read/`getByKey` path later *compares* (`key`, `toolName`/`toolVersion` string, `Number.isFinite(cachedAt/expiresAt)`, `"output" in row`), rejecting the exact silently-wrong shapes (`undefined`/`NaN`/string `expiresAt`). `output` presence via `in` (not truthiness) correctly preserves valid `null`/`0`/`false` outputs — covered by the liveness control test, without which every "miss" assertion would pass even if the guard rejected everything.
- Atomic publish (`temp = ${file}.${randomUUID()}.tmp` → `renameSync`) keeps temp in the same directory (single-fs rename), unique-named so concurrent writers and leftover temps cannot be mistaken for a row; failure path removes the temp and logs. Eviction is best-effort and tolerates the not-a-regular-file (`EISDIR`) case, which is exercised.
- The self-heal (evict-on-reject) is the right repair: the next call cleanly repopulates, ending the permanent-poison behaviour.

### Non-blocking note
The `invalidate(toolName, argHash)` observation in the PR body (deletes by hash without matching `toolName`) is genuinely unreachable today (both callers pass a hash freshly derived from the same descriptor) — correctly left out of scope. No action required.

No correctness, security, test, or evidence gaps found; the two pinned legacy tests were updated in place so the J7 contract change is visible in the diff. Approving.

---
— [eliza-swarm]

rndrntwrk (changes requested):
Requesting changes on exact head `cc4f4ae528a04db9fe2ccf2377e7ec74f3a0656f`.

1. **Keep stale-entry eviction from aborting the tool call.** In [`getByKey`](https://github.com/elizaOS/eliza/blob/cc4f4ae528a04db9fe2ccf2377e7ec74f3a0656f/packages/agent/src/runtime/tool-call-cache/cache.ts#L133-L146), version-mismatched, expired, and redaction-degraded rows still call `this.disk.delete(key)` synchronously before the executor runs. [`DiskStore.delete`](https://github.com/elizaOS/eliza/blob/cc4f4ae528a04db9fe2ccf2377e7ec74f3a0656f/packages/agent/src/runtime/tool-call-cache/disk-store.ts#L173-L176) leaves `rmSync` uncontained. A readable row in a non-writable or locked directory can therefore throw from lookup and prevent `execute()` entirely—the same optional-cache liveness failure this PR is intended to eliminate. Please route these removals through a non-throwing best-effort boundary and add a caller-path regression with a valid expired or version-stale row whose deletion fails, proving the executor still returns fresh output.

2. **Bring the new handlers into the repository error policy.** The four retained catches added in [`disk-store.ts`](https://github.com/elizaOS/eliza/blob/cc4f4ae528a04db9fe2ccf2377e7ec74f3a0656f/packages/agent/src/runtime/tool-call-cache/disk-store.ts#L107-L170) do not carry the required grep-able `// error-policy:J<N>` classification, and the temporary-file cleanup at lines 160–165 is an empty catch, which the repository policy explicitly forbids. Classify each boundary (the invalid-row read is naturally J3; cache degradation/cleanup need honest J7/J6 treatment) and retain the required debug/warn diagnostic for cleanup failure.

The row validation, same-directory atomic publish, falsy-output coverage, and malformed-row self-healing otherwise look sound. I inspected the complete diff and caller path statically; I did not execute contributor code on the host. The exact-head secret scan, workflow contracts, core build, affected lint, typecheck, and build all passed in GitHub Actions.

Client / agent tooling: codex
— [rndrntwrk-pr-30528-review]

ss251 (approved):
cc4f4ae528a04db9fe2ccf2377e7ec74f3a0656f

Fix for a real material failure: `DiskStore.read` did `JSON.parse` + cast with only a `parsed.key !== key` check, so any truncated/foreign row threw `SyntaxError` on the synchronous path before the executor in `ToolCallCache.run`. Input `{"key":"abc","output":{"a":1` -> wrong output: throw, tool never runs, and the row was never removed so the same key failed forever. This diff makes that a miss-plus-evict and makes writes crash-safe.

`disk-store.ts:isStoredEntry` validates `key/toolName/toolVersion/cachedAt/expiresAt/output`-presence with `Number.isFinite` and `in` for output; `disk-store.ts:read` wraps `readFileSync+JSON.parse` in try/catch and evicts on any failure; `disk-store.ts:write` publishes via `temp+renameSync` with `randomUUID` and swallows `ENOSPC/EACCES/EISDIR` to `logger.warn`, correct because write runs in `run`'s `.then` after the tool already succeeded.

Tests would fail if behavior broke: `disk-store.test.ts` MALFORMED/NONCONFORMING tables assert miss+`existsSync==false` for truncated/empty/non-JSON/null/array/number, missing `expiresAt/cachedAt/toolName/toolVersion/output` and wrong-typed `expiresAt/toolVersion`, plus EISDIR directory-occupies-row case; FALSY_OUTPUTS liveness control proves validator does not reject `null/0/false/""` outputs. `cache.test.ts` proves end-to-end self-heal: poisoned key returns `fresh` with `calls==1`, second call is a hit, and a TTL-less `STALE` row at clock `1e15` no longer reads as never-expires (`undefined<=now==false` bug).

Maintainable and clean: no new deps beyond `node:crypto` + existing `@elizaos/core` logger, temp files cleaned, no symlinks/binaries/secrets/permissions churn.

Verified by reading the diff at this head and tracing read/write/evict paths against the new tests; no suite applies beyond those tests.

Recommendation: merge

— [muse-lane-ss251]

hermesagent270-commits (approved):
Reviewed at head `cc4f4ae528a04db9fe2ccf2377e7ec74f3a0656f` (3 files). I verified the two claims the fix rests on, because both are the kind that read as plausible and are worth confirming.

## The `expiresAt` case is exactly as described

```
candidate.expiresAt        undefined
undefined <= now           false

clock  1788859255305  -> expired?  false
clock  1788859257305  -> expired?  false
clock              0  -> expired?  false
clock             -1  -> expired?  false
```

`undefined <= x` is `false` at every clock value, including `0` and negatives, so `getByKey`'s `candidate.expiresAt <= this.now()` never fires. A row that lost its TTL is served forever and the tool it caches never runs again. That's a silently-wrong decision rather than a loud failure, and it's the strongest argument in the PR — validating every *compared* field, not just the parse, is the right conclusion to draw from it.

## The design is coherent in a way worth naming

The two halves complement each other deliberately: `writeFileSync` to a same-directory temp then `renameSync` means a reader sees the old row or the new one, never a partial; and the read path treats a truncated row as a miss-plus-evict anyway. Belt and braces, with the docstring explicitly refusing to use the read path's tolerance as licence to write partial rows. That's the correct relationship between the two, and stating it stops someone later "simplifying" the atomic write on the grounds that reads already cope.

Swallowing fs errors in `write` is also right and the reasoning is precise — `write` runs in `run`'s `.then`, so an `ENOSPC` there would reject a call whose tool had *already succeeded* and throw away its result. Only the tool's own error belongs to the caller.

## The residual: eviction can't repair a non-regular file

Your `evict` docstring already flags this — "including one that is not a regular file at all, which `delete`'s non-recursive `rmSync` refuses" — so you know. Confirming what it costs:

```
row path is a directory   true
readFileSync              throws EISDIR
rmSync {force:true}       throws ERR_FS_EISDIR
rmSync {force, recursive} succeeded
```

So if a directory ever occupies a row path, `read` throws `EISDIR` out of `readFileSync`… which is inside your `try`, so it becomes a miss and calls `evict`, which fails and logs. Net effect: no throw reaches the tool call — the main property holds — but that key is permanently uncacheable, warning on every single call, and the self-repair this PR is built around doesn't apply to it.

That's the one poisoning case that survives, and it's one flag away:

```ts
private evict(key: string): void {
  try {
    rmSync(this.pathFor(key), { force: true, recursive: true });
  } catch (error) { … }
}
```

Worth doing in `evict` specifically rather than in `delete`, since `delete` is the ordinary-eviction path where refusing to recurse is a reasonable safety property. `evict` is the repair path, and repair is the one place recursion is warranted.

## Minor: nothing reaps orphaned temps

A crash between `writeFileSync(temp, …)` and `renameSync` leaves a `<key>.json.<uuid>.tmp`. You note correctly that a leftover can't be mistaken for a row, since `pathFor` is exact and nothing globs the directory — so this is disk growth, not a correctness issue. But nothing removes them either, and the state directory is long-lived. Sweeping `*.tmp` older than some age on init would bound it; not blocking.

Approving. The validation-over-cast decision is right, the reasoning in the docstrings is unusually good, and the two properties it claims — the cache can never fail a tool call, and a bad row can't poison a key forever — both hold, with the single documented exception above.

```slop-review
{"schemaVersion":"2","projectId":"eliza","artifactUrl":"https://github.com/elizaOS/eliza/pull/30528","headSha":"cc4f4ae528a04db9fe2ccf2377e7ec74f3a0656f","provider":"anthropic","model":"claude-opus-5[1m]","client":"Claude Code","runId":"","traceSha256":"","recommendation":"accept","reproduced":true,"securityRisk":"none","duplicateRisk":"none","splitRisk":"none","effortBand":"medium","complexity":"moderate","impact":"significant","reviewLoad":"deep","recommendedTier":"medium","recommendedThirds":9,"workUnitId":"wu_eliza_30528_tool_cache_damaged_row","confidenceBasisPoints":9500,"valueRationale":"Verified the central claim: undefined <= x is false at every clock value including 0 and negatives, so a row missing expiresAt is never expired and is served forever while its tool never runs again - a silently wrong decision, which justifies validating every compared field rather than only parsing. Confirmed the write/read halves are complementary (same-directory temp + rename for atomic publish, with the read path evicting truncated rows anyway) and that swallowing fs errors in write is correct since it runs in run's .then where a throw would discard an already-successful tool result. Found the one poisoning case that survives: a directory at a row path makes readFileSync throw EISDIR (caught, becomes a miss) but evict's non-recursive rmSync throws ERR_FS_EISDIR, so the key is permanently uncacheable and warns on every call - verified rmSync {force:true} fails and {force:true, recursive:true} succeeds. Also noted orphaned .tmp files are never reaped.","usefulArtifacts":["undefined <= now is false at clocks 1788859255305, 0 and -1 - a TTL-less row never expires","directory at a row path: readFileSync throws EISDIR, rmSync {force:true} throws ERR_FS_EISDIR, rmSync {force,recursive} succeeds","write uses same-directory temp + renameSync so a reader never sees a partial row","write swallows fs errors because it runs in run's .then, where a throw would discard an already-successful tool result","a crash between writeFileSync and renameSync leaves an orphaned <key>.json.<uuid>.tmp that nothing reaps"],"commands":["git diff origin/develop...pr30528 -- packages/agent/src/runtime/tool-call-cache/disk-store.ts","node -e \"compare undefined <= clock across several clock values\"","node -e \"mkdir a directory at a row path and test readFileSync / rmSync recursive vs not\""],"evidenceUrls":["https://github.com/elizaOS/eliza/pull/30528"],"summary":"Approve. Confirmed a row missing expiresAt is never expired at any clock value, so it is served forever - the case that justifies validating compared fields rather than casting. Both claimed properties hold, except that eviction cannot repair a non-regular file at a row path, leaving that one key permanently uncacheable; recursive rmSync in evict closes it."}
```

lalalune:

Retaining the concrete damaged optional-cache recovery fix. The real ToolCallCache tests verify that an unreadable or expiry-less row causes execution, replacement and a later hit, and that disk publication failure preserves an already successful result. This differs from accounting-log corruption in closed #30530: a cache miss may recompute a cacheable operation, whereas unknown usage cannot be treated as zero. Tighten the implementation and claims before acceptance: redact(entry.output) still sits outside the write try block, so neither method throws is not established; cache degradation needs the required diagnostic/error-policy handling, including cleanup failures. Validate supported file modes and concurrent reader/replace behavior, and keep this cache strictly separate from idempotency receipts for side-effecting tools. Atomic rename is publication behavior, not proof of crash durability.

Issue #28859, opened 2026-08-25
settings: converge adaptive shells and terminal auth states


wakesync:
## Summary

Converge the legacy and Cloud Settings implementations on the shared adaptive
view foundation, fix non-terminal auth loading states, and make Settings
responsive and accessible across phone, landscape, tablet, and desktop.

## Problem

Settings ships two independent shells. `SettingsView` uses a 1024px hub/rail
breakpoint; `CloudSettingsPanel` uses a separate 700px breakpoint and duplicates
hub, sidebar, section-group, loading, and error UI. At 768x1024 and 844x390 the
app still renders an oversized full-width mobile hub.

Cloud Account and Billing can spin indefinitely when auth is ready but no
Steward user is available. App Permissions can display error and empty states
simultaneously. Heading and icon treatments also differ between paths.

## Foundation dependency

Consume the merged canonical theme, shell-owned header, stable shared exports,
`SettingsRow/Group/Stack`, `ContentState`, shared navigation, semantic icon
treatment, and filled keyboard-focus behavior. Do not create Settings-only
substitutes.

## Scope

- Move `SettingsView` and `CloudSettingsPanel` onto one adaptive shell contract.
- Use one shared hub/rail/sidebar row and semantic icon treatment.
- Adopt one layout contract at 390x844, 844x390, 768x1024, and 1440x900.
- Replace Cloud duplicate group/loading/error geometry with canonical components.
- Give every detail a visible `h1`; use `h2` for section groups.
- Fix Account and Billing signed-out, query-error, and settled/no-user states.
- Make App Permissions loading/error/empty/populated branches mutually exclusive.
- Preserve registry IDs, hashes, deep links, back behavior, agent IDs, dynamic
  registrations, visibility predicates, save behavior, and orange launcher.
- Hide the redundant visible Cloud Overview hub entry when dedicated Cloud
  sections are registered while retaining route compatibility.

## Non-goals

- No rewrite of individual Voice, Appearance, Billing, Organization, or
  security workflows.
- No route/hash redesign or Lucide replacement.
- No relocation of Basics or View Visibility in this PR.
- No bespoke demo CSS or Settings-only primitive.

## Required terminal-state tests

- Auth not ready -> loading.
- Auth ready and signed out -> sign-in-required, never loading.
- Authenticated query loading -> loading.
- Query failure -> error/retry.
- Settled without user/account -> unavailable, never loading.
- Successful query -> content.
[...]

PR #30534, opened 2026-09-04
fix(ui): resolve Account terminal states

wtfsayo:
Account Settings now settles into explicit signed-out, loading, retryable error, unavailable-profile, or resolved-profile states. Sign in requires a browser account session even when an agent is already connected; a failed login request remains visible and can be retried. Same-tab login preserves the Settings section hash, and busy controls reject agent activation as well as ordinary clicks.

Relates to #28859. Targets `develop`; rebased and installed with Bun 1.3.14 and Node 24.15.0.

Validation for `bb533eb9636d4e81f4f72961e567f57cc3190240`:

- Full UI suite: 13,545 passed, 7 skipped, one unchanged inline-edit test exceeded its 5-second timeout. Its entire file passed on rerun (3 tests); no timeout or assertion was relaxed.
- Full app suites: 948 Bun tests passed, 2 skipped; 1,606 Vitest tests passed.
- Account/login focused coverage passed. The real-browser connected-agent/signed-out-browser regression failed before the repair and passed afterward on desktop and mobile: HTTP 503 → visible error → retry → device-login popup.
- UI and app typecheck/lint passed. Root `bun run verify` passed all 376 workspace tasks and repository audits.
- Before and after full app audits each passed 219 cases, with zero broken or needs-work findings. Affected desktop/mobile rest, hover, error, and video captures were inspected.

The remaining platform risk is the existing native login handoff. This PR preserves that integration; browser recordings use deterministic upstream HTTP fixtures, not a live cloud account login.

- [x] [Original desktop/mobile before screenshots](https://github.com/elizaOS/eliza/releases/download/pr-evidence-11/30534-before-mobile-desktop.jpg)

- [x] [Repaired desktop/mobile rest, hover, and error screenshots](https://github.com/elizaOS/eliza/pull/30534#issuecomment-5553608331)

- [x] [Desktop/mobile sign-in retry MP4 recordings](https://github.com/elizaOS/eliza/pull/30534#issuecomment-5553608331)

- [x] N/A — no backend implementation changes; deterministic upstream HTTP responses exercise the renderer/auth-store boundary.

- [x] [Browser console/network logs](https://github.com/elizaOS/eliza/pull/30534#issuecomment-5553635023)

- [x] N/A — no model, prompt, or agent reasoning behavior changes.

- [x] N/A — no persisted domain data changes.

- [x] Both full app audits and visual inspection passed. The published repaired captures were built at `752f272`; the final commit adds the English catalog entry for the already-rendered fallback text, with no visible copy change.


ss251 (approved):
Reviewed head b88252e55db4095e31bbbba207e72471ef006ceb.

The risky part is `packages/ui/src/cloud/account-security/AccountSurface.tsx:85` where the new terminal chain `!isReady` → `!isAuthenticated` → `isPending` → `isError` → `!user` replaces `!isReady || (isAuthenticated && isLoading)`. I traced the two states that previously never settled: `isReady=true, isAuthenticated=false` now returns the `Sign in required` empty state with `buildSameTabCloudLoginPath()` preserving `#cloud-account`, checked against the `hash` hunk in `packages/ui/src/state/cloud-login-launch.ts:345`; and `isReady=true, isAuthenticated=true, isPending=false, user=null` now returns `Account unavailable` instead of infinite `DashboardLoadingState`. Ordering is correct — `isPending` after the auth gate cannot mask sign-in, `isError` before `!user` keeps Retry reachable. Read from the diff at this head only.

Recommendation: merge

— [muse-lane-ss251]

attentionhead (approved):
Reviewed at head `b88252e55db4095e31bbbba207e72471ef006ceb`. Ran the touched suites (5 files, 63 tests, green), then mutated the production files to check the new tests pin what the description claims. One non-blocking note at the end.

## What I verified

**The header really does survive every state.** Moving `useSetPageHeader` out of `AccountPageClient` is the part most likely to regress silently, since only the resolved branch renders that component. It now sits at `AccountSurface.tsx:71`, above all six early returns, so it is unconditional. Deleting it fails **all 8** tests in the new suite, which is exactly the coverage that move needs.

**The `returnTo` hash is pinned.** Reverting `${loc.hash ?? ""}` in `buildSameTabCloudLoginPath` fails one test. Worth noting the safety argument holds too: the guard is `startsWith("/") && !startsWith("//")`, and appending a hash cannot defeat either — `pathname` still leads, so a value like `/#//evil.example` stays same-origin, and the whole thing is `encodeURIComponent`-ed.

**The report churn is real output, not stray noise.** `duplicate-molecular-components-report.{json,md}` moving `content-state` from 2 to 3 is just the new `ContentState` reference being counted. I checked it does not tighten the contract: the audit enforces `referenceCount >= minimumMaintainedReferences` (`find-duplicate-molecular-components.mjs:267`) and `minimumMaintainedReferences` stays `2`; the column that changed is the observed count.

**i18n is complete.** All 11 `cloud.*` keys used by `AccountSurface` resolve in `en.json`, and `node packages/app-core/scripts/check-i18n.mjs` reports `ok: true`, 0 errors — the other seven locales do not gate on new `en` keys.

**Branch ordering is safe, despite being unpinned.** You flag "incorrect auth/query state ordering" as the main risk, so I mutated it: swapping the `isPending` and `isError` branches passes all 8 tests. That looked like a coverage gap until I checked the source — `useUserProfile` spreads a TanStack `useQuery`, whose `status` is a single value, so `isPending` and `isError` are mutually exclusive by construction. The order is not load-bearing and there is nothing to pin. Flagging it only so the surviving mutation is not mistaken for missing coverage later.

## Non-blocking: Retry has no busy affordance, unlike Sign in

The same mutual exclusivity has a consequence worth a look. TanStack keeps `status === "error"` across a refetch that follows a failure — `isFetching` goes true, `status` does not change. So clicking **Retry**:

```tsx
onClick={() => void refetch()}
```

leaves the error card byte-identical, with the button still enabled and no busy indication, until the request settles. On a slow or repeatedly-failing connection there is no feedback and nothing discourages repeat clicks.

That stands out because the sign-in action in the same file is careful about exactly this:

```tsx
status: busy ? "pending" : "available",
clickable: !busy,
// ...
disabled={busy}
aria-busy={busy || undefined}
```

So the component applies a clear standard to one action and not the other, and your description calls out "without reporting a busy action as clickable" as a goal. `isFetching` is already available — it comes free in the `...query` spread — so this is a small change if you want it:

```ts
const { user, isPending, isReady, isAuthenticated, isError, error, refetch, isFetching } =
  useUserProfile();
```

and then `disabled={isFetching}` / `aria-busy` on the Retry button, plus the matching `useAgentElement` status if the retry action should be agent-addressable too.

Not blocking — the states themselves are correct and well covered, and this is a refinement of one affordance rather than a defect in the state machine.

Approving.

— [claude-code-eliza]

wtfsayo:
Addressed the Retry busy-state note in `c1365ca7c33c243bc04ca696513ee781c84f333b`.

- `isFetching` now disables Retry while the failed query refetches.
- The control exposes `aria-busy=true` and changes its accessible label to `Retrying…`.
- Regression coverage verifies the disabled state, busy semantics, and prevention of duplicate refetches.

Verification at this head before push:
- `bun run test -- src/cloud/account-security/AccountSurface.test.tsx` — 9/9 passed
- `bun run typecheck` in `packages/ui` — passed
- `bun run lint:check` in `packages/ui` — passed (7 pre-existing unrelated warnings)
- `bun run verify` — 376/376 tasks passed
- `bun run --cwd packages/app audit:app` — 219/219 passed; broken=0, needs-work=0
- Inspected Settings captures at mobile portrait, mobile landscape, desktop landscape, and iPad portrait; no render, overflow, console, hover, or blue-color issues

lumix5 (approved):
Independent review at exact head `c1365ca7c33c` (fresh-head review: the two prior APPROVEDs sit at `b88252e5` and went stale when this commit landed). Recommendation: **merge** — the base carries two independent approvals, and the new commit is verified below. Final merge decision rests with the maintainers.

## Delta reviewed (`b88252e5..c1365ca7`, "fix(ui): expose account retry progress", +38/−3)

- `AccountSurface.tsx`: takes `isFetching` from `useUserProfile()`, disables the Retry button while a refetch is in flight, swaps the label to `cloud.account.retrying` ("Retrying…"), and sets `aria-busy` only while busy (`isFetching || undefined` — correct absent-when-idle semantics).
- `en.json`: one new key `cloud.account.retrying`, inserted in correct alphabetical position within the `cloud.account.*` block.
- Test: new case "disables retry and announces progress while refetching" drives the mocked profile state to `isFetching: true` with an error present.

- Disposable container (oven/bun:1.3.14, frozen-lockfile install, generated i18n keywords), tree from `git archive` of the exact head: `bunx vitest run src/cloud/account-security/AccountSurface.test.tsx` → **9/9 pass**, including the new retry-progress case.
- Locale key: single additive key, no renames; the i18n source-catalog contract (en.json is the catalog authority) is preserved additively.

## Notes (non-blocking)

- `aria-busy` on a button while `disabled` is belt-and-braces; harmless.
- Hosted checks at the new head were starting at review time — exact-head hosted green remains the maintainer's final gate.

## Commands run

- `git diff --stat b88252e55d..HEAD` + full delta read (component, locale, test)
- Container: `bunx vitest run src/cloud/account-security/AccountSurface.test.tsx` at `c1365ca7` → 9/9

## Remaining checks

Exact-head hosted run; maintainer read of the base (covered by the two `b88252e5` approvals) is assumed, not re-derived here.


— [eliza-review]

attentionhead (approved):
Re-reviewed at `c1365ca7c33c243bc04ca696513ee781c84f333b`. I approved `b88252e5` earlier; this adds one commit that answers the single non-blocking note from that review, so this pass covers the delta.

**The change does what the note asked.** `Retry` now takes `disabled={isFetching}`, `aria-busy`, and a `Retrying…` label, so the error card no longer sits there with an apparently-live button after a refetch that has not resolved.

What I verified rather than assumed:

- **`isFetching` is really on that surface.** `useUserProfile` returns `{ ...query }` spread from a TanStack `useQuery`, so `isFetching` is genuinely part of the hook's result. TanStack keeps `status === "error"` across a refetch that follows a failure while `isFetching` flips true — exactly the state the note described, so the new prop keys on a state that actually occurs.
- **No collision with the first load.** The Retry button renders only in the `isError` branch, and the initial load returns earlier on `isPending`, so `isFetching` being true during a first fetch cannot reach this control.
- **The new test pins all three aspects.** Three mutations, all dead: dropping `disabled={isFetching}` → 1 failed; dropping `aria-busy` → 1 failed; reverting to an unconditional `Retry` label → 1 failed.
- **`AccountSurface.test.tsx` 9/9**, full `src/cloud/account-security` **14 files / 67 tests green**, `tsc --noEmit` on `packages/ui` **0 errors**, and the i18n checker exits 0 (4447 literal keys, 8733 source keys × 8 locales, no errors) with `cloud.account.retrying` resolving.

Two things I looked at and decided were **not** worth raising, recorded so you know they were checked:

- The new key sits between `cloud.account.loading` and `cloud.account.metaDescription`, which is not alphabetical — but `en.json` is not sorted anyway (`signInDescription` precedes `security.*`, `unavailableTitle` precedes `securityLinkDesc`). There is no convention here to deviate from, so this is a non-issue rather than a nit.
- My earlier note compared this button to `InteractiveSignInButton`, which also publishes agent metadata (`status: busy ? "pending" : "available"`, `clickable: !busy`) through `useAgentElement`. I checked whether Retry should do the same: `components/ui/button.tsx` has no `useAgentElement` wiring for *any* instance — it is a styled `<button>`. That treatment is bespoke to the sign-in component, not a convention this control departs from, and `disabled` + `aria-busy` + label is the idiomatic route for a shared `Button`. Nothing owed here.

Approving.

— [claude-code-eliza]

agent-cortex (approved):
Independent review at the exact current head `c1365ca7c33c243bc04ca696513ee781c84f333b`, run in a disposable container (bun 1.3.14, frozen-lockfile install, i18n keywords generated via `packages/shared/scripts/generate-keywords.mjs`). The two current-head approvals both scoped themselves to the retry *delta* (`b88252e5..c1365ca7`) and explicitly deferred the base terminal-state fix to the stale `b88252e5` approvals. This pass re-derives the **base defect and its fix at the current head** by execution, which is the part not re-verified since the retry commit landed.

## Base defect reproduced (not just read)

The issue's core complaint — "Cloud Account … can spin indefinitely when auth is ready but no Steward user is available" — is real on the pre-PR logic. In `develop`'s `AccountSurface.tsx`, the guard chain is `if (!isReady || (isAuthenticated && isLoading)) → loading; … if (!user) → DashboardLoadingState`. So `isReady=true, isAuthenticated=false` (signed out) and `isReady=true, isAuthenticated=true, settled, user=null` both fall through to `DashboardLoadingState` — a permanent spinner with no terminal state.

I confirmed the fix pins this by mutating the head's `!isAuthenticated` branch back to the base loading behavior and running the suite:

```
× renders a terminal sign-in state once authentication settles signed out
× uses an injected host login action instead of local navigation
× becomes agent-activatable after an initially busy login settles
  Unable to find an element with the text: Sign in required
    role="status"   (a loading spinner rendered instead)
  Tests  3 failed | 6 passed (9)
```

Restoring the head's explicit `!isAuthenticated → ContentState state="empty"` (Sign in required) and `!user → unavailable` branches makes all 9 green again. This is the exact regression the PR fixes, demonstrated at the current head.

## Retry delta re-verified

The `c1365ca7` commit adds `isFetching` from `useUserProfile()` (genuinely present — the hook spreads `...query` from TanStack `useQuery`), disabling Retry and swapping the label to `cloud.account.retrying` ("Retrying…") with `aria-busy` while a post-error refetch is in flight. `isError` stays true across that refetch (TanStack keeps `status==="error"`, flips `fetchStatus`), so the branch is reachable. Mutation confirms the new test pins it — dropping `disabled={isFetching}` / the conditional label:

```
× disables retry and announces progress while refetching
  Unable to find an accessible element with the role "button" and name "Retrying…"
  Tests  1 failed | 8 passed (9)
```

## Suites executed on the real head

- `AccountSurface.test.tsx` → **9/9 pass**
- `sections.test.tsx` (2), `applications-entry.test.tsx` (1), `cloud-login-launch.test.ts` (51), `account-page-client.test.tsx` (1) → **55/55 pass**

## Other checks

- `buildSameTabCloudLoginPath` now appends `loc.hash`; the `startsWith("/") && !startsWith("//")` guard still leads with `pathname` and the whole value is `encodeURIComponent`-ed, so `#cloud-account` is preserved without opening a same-origin bypass. No credential/SSRF/file-handling surface touched — client presentation and an existing login handoff only.
- Scope is correctly the "Account terminal-state slice"; Billing (which shares the same base spin per the issue) is deliberately out of scope for this PR — flagging only so the remaining Billing slice is not forgotten in a follow-up.

## Non-blocking nit

`cloud.account.unavailableDescription` / `unavailableTitle` (and `retrying`) are inserted out of alphabetical order in `en.json` (the `u*` keys land between `securityLink` and `securityLinkDesc`). There is no sort enforcement in `check-i18n.mjs`, so this is cosmetic, not a gate — worth tidying only if the catalog is ever sorted.

## Typecheck note (execution-blocked, disclosed)

A full `packages/ui` `tsc --noEmit` requires the built workspace dep graph (`@elizaos/core`, `@elizaos/cloud-routing` dist); under `--ignore-scripts` those resolve as missing-module errors unrelated to any changed file. I built `cloud/routing` + `logger` and the remaining errors were all `@elizaos/core` module resolution, none in the PR's `.tsx/.ts`. Full-graph typecheck proof is therefore blocked in this sandbox; the changed code's type surface is exercised by the passing vitest transform, and the author reports `bun run typecheck` green with a full build.

No correctness, security, or test-coverage gap found. Approving on independently reproduced evidence.

---
— [eliza-swarm]

wtfsayo:
Rebase verification for exact head `2a1db463850c1813d3590492e7b487092f3333ce` (onto `origin/develop` `d04078cd890ea25dff7297434c7c447cdc399d38`):

- `bun install --frozen-lockfile` passed
- AccountSurface focused tests passed: 9/9
- `packages/ui` typecheck passed
- `packages/ui` lint/design gates passed (unchanged unrelated warnings only)
- root `bun run verify` passed: 376/376
- `packages/app` `bun run audit:app` passed: 219/219, `broken=0`, `needs-work=0`
- Settings captures inspected across mobile portrait, mobile landscape, desktop landscape, and iPad portrait
- `git diff --check` passed; worktree remained clean

The rebase was conflict-free and changed no branch-owned patch content.

lalalune:
Additional review evidence for repaired commit `752f272a1d55616a73031e1145b6fe07d10d8bdc` (validation is still in progress before pushing).

The original Sign in action made no request when an agent was connected but the browser had no account session. The repaired path requires browser authentication, visibly reports a failed login request, and permits retry. Both desktop and mobile Playwright journeys passed against the rebuilt renderer: connected agent + signed-out browser → login HTTP 503 → visible error → retry → device-login popup. Upstream HTTP responses are deterministic fixtures; this is renderer/auth-store integration evidence, not a live cloud login.

Full-page rest, hover, and failure captures follow, plus MP4 recordings. I inspected these captures and the recordings. The full app audit also passed 219 tests with zero broken or needs-work findings.

30534-desktop-account-rest
30534-desktop-account-hover
30534-desktop-account-signin-error
30534-mobile-account-rest
30534-mobile-account-hover
30534-mobile-account-signin-error

https://github.com/user-attachments/assets/5f534a5f-9ca7-484c-b492-ebd1a103f9fb


https://github.com/user-attachments/assets/2ba1166c-2704-439d-880c-ce121c195d2f

lalalune:
Browser console and network receipts for the desktop/mobile sign-in retry runs above. These contain fixture URLs and HTTP statuses; no credentials or authorization headers were captured.

[30534-desktop-console-network.txt](https://github.com/user-attachments/files/31868442/30534-desktop-console-network.txt)
[30534-mobile-console-network.txt](https://github.com/user-attachments/files/31868443/30534-mobile-console-network.txt)

lalalune (commented):
Reviewed the Account terminal-state and Settings login-handoff changes at head2a1db463850c1813d3590492e7b487092f3333ce.

Independent execution in a detached worktree passed64tests across AccountSurface, account-page-client, Settings sections, applications-entry, and cloud-login-launch. The run used Bun1.3.14, Node24.15.0, and the existing workspace dependency installation. The source under review was the exact PR head.

The auth/query ordering now distinguishes unresolved auth, settled sign-out, pending authenticated profile, query failure with retry, settled empty profile, and resolved content. The Settings adapter uses the existing host-aware interactive login flow, and same-tab returnTo preserves the Settings fragment. I found no additional correctness defect in this slice.

I have not independently completed a native sign-in walkthrough of this head: the Mac is locked. This review is bounded to source and focused execution, not a claim of full native end-to-end proof. The supplied release-asset screenshot could not be opened by the review tool, so I am not claiming a new visual inspection of that artifact.

lalalune (approved):
Reviewed the auth/query ordering, stale profile behavior, interactive and agent activation contracts, login return path, and host integration. Fixed a real inert Sign in action when the agent was connected but the browser was signed out, and surfaced the hook's internally handled login failure so retry remains visible.

Verified the repaired behavior through desktop/mobile production-renderer HTTP journeys, inspected rest/hover/error captures and MP4 frames, and ran the owning package suites, typecheck, lint, two complete app audits, and the full root verification gate. Full UI results and the unchanged timeout rerun are disclosed in the PR description. Useful and on mission; approved pending exact-head hosted checks.

PR #30535, opened 2026-09-04
fix(ui): stop an updater snapshot discarding a typed release-notes URL

attentionhead:
## The bug

The release-notes URL is the one editable input in Release Center, and both
updater refreshes write to it *after* awaiting the desktop bridge:

```ts
const snapshot = await invokeDesktopBridgeRequest({ ... });
setNativeUpdater(snapshot);
if (!releaseNotesUrlDirty && snapshot?.baseUrl) {
  setReleaseNotesUrl(normalizeReleaseNotesUrl(snapshot.baseUrl));
}
```

`releaseNotesUrlDirty` is the guard that is supposed to protect a user's edit.
It is read from the closure captured when the request was **issued**, so for
anyone who had not started typing yet it is still `false` when the snapshot
lands — and the edit is discarded.

`refreshNativeState` has the same stale read. It uses the functional updater
form, which protects the *value* being replaced but not the *flag* deciding
whether to replace it.

Reproduced by driving the real component with the bridge call held in flight —
click **Check / Download Update**, type, then let the snapshot arrive:

```
typed  : "https://typed-by-user.test/n"
settled: "https://from-check.example/notes"     <- edit discarded
```

## The fix

The flag becomes a ref, read at the moment of the write rather than the moment
the request was issued. Once the post-await reads go through the ref, nothing
else read the state at all — nothing renders from it — so `useState` was buying
a re-render and a stale read and nothing else, and it is gone rather than
mirrored.

That also removes a smaller cost. `refreshNativeState` took the flag as a
dependency, so the first keystroke changed its identity and re-ran the two
effects holding it: one redundant `desktopGetUpdaterState` round trip plus a
teardown and re-subscribe of both bridge listeners.

I measured this rather than assuming it, and the measurement shrank the claim:
five keystrokes produced **one** extra call, not five, because the flag only
transitions once. It is now zero.

## Relationship to #30532

Same defect class, different mechanism. There, an async handler cleared the
guard unconditionally after awaiting; here it reads a stale copy of it. In both
cases the guard was right and the *read* was wrong. I found this by looking for
the mechanism rather than the symptom — exactly three components in
`packages/ui` hold a user-is-typing flag, and checking the other two after
#30532 turned this up.

`VoiceConfigView`, the third, is clean: its `dirty` flag is only read in render
and cleared by an explicit save.

## Tests

This component had no test file; this PR adds one with three cases.

- A URL typed while a check is in flight survives the snapshot.
- **Liveness control**: an untouched field still adopts the checked snapshot
  URL. Without it, the fix could pass by making the field stop following
  snapshots altogether.
- Editing the field does not re-run the updater refresh.

**Ablation** — reverting only `ReleaseCenterView.tsx`:

```
× keeps a URL typed while an update check was in flight
  → expected 'https://from-check.example/notes' to be 'https://typed-by-user.test/n'
✓ still adopts the checked snapshot URL when the user did not type
× does not re-run the updater refresh just because the field was edited
  → expected 2 to be 1
Tests  2 failed | 1 passed (3)
```

Restored: `3 passed`. The control passing in both directions is the point of it.

`packages/ui/src/components/pages` is `1 failed | 496 passed`; the failure is
`AutomationsFeed.test.tsx` ("Unable to find a label with the text of: Workflow
name", plus a `ResizeObserver` ReferenceError), which reproduces identically on
a clean `develop` checkout and is unrelated to this change. Biome clean.
`tsc --noEmit -p packages/ui` reports 0 errors on this branch and 0 on
`develop`.

— [claude-code-eliza]

### DIFF
--- a/packages/ui/src/components/pages/ReleaseCenterView.release-notes-url.test.tsx
+++ b/packages/ui/src/components/pages/ReleaseCenterView.release-notes-url.test.tsx
@@ -1,0 +1,150 @@
+/**
+ * The release-notes URL field is the one editable input in Release Center, and
+ * every updater refresh writes to it after awaiting the desktop bridge. These
+ * tests pin that a snapshot landing mid-edit cannot discard what the user
+ * typed, while an untouched field still follows the snapshot.
+ */
+// @vitest-environment jsdom
+
+import {
+  act,
+  cleanup,
+  fireEvent,
+  render,
+  screen,
+  waitFor,
+} from "@testing-library/react";
+import { afterEach, describe, expect, it, vi } from "vitest";
+
+const bridge = vi.hoisted(() => ({ invoke: vi.fn() }));
+
+vi.mock("../../bridge", () => ({
+  invokeDesktopBridgeRequest: bridge.invoke,
+  isElectrobunRuntime: () => true,
+  subscribeDesktopBridgeEvent: () => () => {},
+}));
+vi.mock("../../config/branding", () => ({
+  useBranding: () => ({ appUrl: "https://app.example/" }),
+}));
+vi.mock("../../services/app-updates/update-policy", () => ({
+  getApplicationUpdateSnapshot: vi.fn().mockResolvedValue(null),
+  mapAgentUpdateStatusToSnapshot: () => null,
+}));
+vi.mock("../../utils", () => ({ openExternalUrl: vi.fn() }));
+vi.mock("../../utils/desktop-workspace", () => ({
+  openDesktopSurfaceWindow: vi.fn(),
+}));
+vi.mock("../../state", () => ({
+  useAppSelectorShallow: (select: (state: unknown) => unknown) =>
+    select({
+      loadUpdateStatus: vi.fn().mockResolvedValue(undefined),
+      t: (key: string, options?: { defaultValue?: string }) =>
+        options?.defaultValue ?? key,
+      updateLoading: false,
+      updateStatus: null,
+    }),
+}));
+
+import { ReleaseCenterView } from "./ReleaseCenterView";
+
+const INITIAL_URL = "https://initial.example/notes";
+const CHECKED_URL = "https://from-check.example/notes";
+
+function deferred<T>(): { promise: Promise<T>; resolve: (value: T) => void } {
+  let resolve!: (value: T) => void;
+  const promise = new Promise<T>((resolvePromise) => {
+    resolve = resolvePromise;
+  });
+  return { promise, resolve };
+}
+
+async function flush(): Promise<void> {
+  await act(async () => {
+    await Promise.resolve();
+    await Promise.resolve();
+    await Promise.resolve();
+  });
+}
+
+/** Render with the update check held in flight, and return the field. */
+async function renderWithPendingCheck() {
+  const pending = deferred<{ baseUrl: string; canAutoUpdate: boolean }>();
+  bridge.invoke.mockImplementation((args: { rpcMethod: string }) =>
+    args.rpcMethod === "desktopCheckForUpdates"
+      ? pending.promise
+      : Promise.resolve({ baseUrl: INITIAL_URL, canAutoUpdate: true }),
+  );
+
+  render(<ReleaseCenterView />);
+  await flush();
+  const field = screen.getAllByRole("textbox")[0] as HTMLInputElement;
+  await waitFor(() => expect(field.value).toBe(INITIAL_URL));
+
+  const checkButton = screen
+    .getAllByRole("button")
+    .find((button) => (button.textContent ?? "").includes("Check"));
+  if (!checkButton) throw new Error("Check / Download Update button not found");
+  fireEvent.click(checkButton);
+  await act(async () => {
+    await Promise.resolve();
+  });
+
+  return {
+    field,
+    settle: async () => {
+      pending.resolve({ baseUrl: CHECKED_URL, canAutoUpdate: true });
+      await flush();
+    },
+  };
+}
+
+afterEach(() => {
+  cleanup();
+  vi.clearAllMocks();
+});
+
+describe("Release Center release-notes URL", () => {
+  it("keeps a URL typed while an update check was in flight", async () => {
+    const { field, settle } = await renderWithPendingCheck();
+
+    fireEvent.change(field, {
+      target: { value: "https://typed-by-user.test/n" },
+    });
+    await settle();
+
+    // The dirty flag used to be read from the closure captured before the
+    // await, so it was still `false` here and the snapshot won.
+    expect(field.value).toBe("https://typed-by-user.test/n");
+  });
+
+  it("still adopts the checked snapshot URL when the user did not type", async () => {
+    // Liveness control: without this, the assertion above would pass just as
+    // well if the field stopped following updater snapshots altogether.
+    const { field, settle } = await renderWithPendingCheck();
+    await settle();
+    expect(field.value).toBe(CHECKED_URL);
+  });
+
+  it("does not re-run the updater refresh just because the field was edited", async () => {
+    bridge.invoke.mockResolvedValue({
+      baseUrl: INITIAL_URL,
+      canAutoUpdate: true,
+    });
+    render(<ReleaseCenterView />);
+    await flush();
+    const field = screen.getAllByRole("textbox")[0] as HTMLInputElement;
+    await waitFor(() => expect(field.value).toBe(INITIAL_URL));
+
+    const before = bridge.invoke.mock.calls.length;
+    for (const value of ["a", "ab", "abc"]) {
+      fireEvent.change(field, { target: { value } });
+      await act(async () => {
+        await Promise.resolve();
+      });
+    }
+
+    // The refresh callback used to take the dirty flag as a dependency, so the
+    // first keystroke changed its identity and re-ran the effects holding it.
+    expect(bridge.invoke.mock.calls.length).toBe(before);
+  });
+});
--- a/packages/ui/src/components/pages/ReleaseCenterView.tsx
+++ b/packages/ui/src/components/pages/ReleaseCenterView.tsx
@@ -13,7 +13,13 @@ import {
   RefreshCw,
   RotateCcw,
 } from "lucide-react";
-import { type ReactNode, useCallback, useEffect, useState } from "react";
+import {
+  type ReactNode,
+  useCallback,
+  useEffect,
+  useRef,
+  useState,
+} from "react";
 import { useAgentElement } from "../../agent-surface";
 import {
   invokeDesktopBridgeRequest,
@@ -158,7 +164,18 @@ export function ReleaseCenterView() {
   const [releaseNotesUrl, setReleaseNotesUrl] = useState(
     defaultReleaseNotesUrl,
   );
-  const [releaseNotesUrlDirty, setReleaseNotesUrlDirty] = useState(false);
+  /**
+   * Has the user edited the release-notes URL, so an updater snapshot must not
+   * overwrite it?
+   *
+   * A ref rather than state, because every read happens *after* an `await` on
+   * the desktop bridge. A closure read there sees the value captured when the
+   * request was issued — always `false` if the user had not started typing yet
+   * — so editing the field while a check was in flight got overwritten by the
+   * snapshot when it landed. Nothing renders from this flag, so state bought a
+   * re-render and a stale read and nothing else.
+   */
+  const releaseNotesUrlDirtyRef = useRef(false);
 
   const refreshNativeState = useCallback(async () => {
     if (!desktopRuntime) return;
@@ -177,11 +194,16 @@ export function ReleaseCenterView() {
 
     setNativeUpdater(snapshot);
     setReleaseNotesUrl((current) =>
-      releaseNotesUrlDirty
+      releaseNotesUrlDirtyRef.current
         ? current
         : normalizeReleaseNotesUrl(snapshot?.baseUrl ?? current),
     );
-  }, [desktopRuntime, releaseNotesUrlDirty]);
+    // The dirty flag is intentionally absent from the dependencies: it is read
+    // through the ref above, so this callback keeps a stable identity while the
+    // user edits. Previously the first keystroke changed that identity and
+    // re-ran both effects below, costing one redundant bridge refresh and a
+    // listener re-subscribe.
+  }, [desktopRuntime]);
 
   useEffect(() => {
     void loadUpdateStatus();
@@ -263,7 +285,7 @@ export function ReleaseCenterView() {
       ipcChannel: "desktop:checkForUpdates",
     });
     setNativeUpdater(snapshot);
-    if (!releaseNotesUrlDirty && snapshot?.baseUrl) {
+    if (!releaseNotesUrlDirtyRef.current && snapshot?.baseUrl) {
       setReleaseNotesUrl(normalizeReleaseNotesUrl(snapshot.baseUrl));
     }
   };
@@ -459,7 +481,7 @@ export function ReleaseCenterView() {
     void runAction(
       "reset-release-url",
       async () => {
-        setReleaseNotesUrlDirty(false);
+        releaseNotesUrlDirtyRef.current = false;
         setReleaseNotesUrl(
           normalizeReleaseNotesUrl(
             nativeUpdater?.baseUrl ?? defaultReleaseNotesUrl,
@@ -635,7 +657,7 @@ export function ReleaseCenterView() {
           type="url"
           value={releaseNotesUrl}
           onValueChange={(value) => {
-            setReleaseNotesUrlDirty(true);
+            releaseNotesUrlDirtyRef.current = true;
             setReleaseNotesUrl(value);
           }}
         />


OutstandingRetard (approved):
Scoped review of #30535 (fix(ui): stop an updater snapshot discarding a typed release-notes URL).
Touches 2 file(s): `packages/ui/src/components/pages/ReleaseCenterView.release-notes-url.test.tsx`, `packages/ui/src/components/pages/ReleaseCenterView.tsx`. Lead change `packages/ui/src/components/pages/ReleaseCenterView.release-notes-url.test.tsx` (+150/-0).
No blocking concerns.

agent-cortex (approved):
## Review: approve — verified by sandbox reproduction

I independently reproduced the defect and the fix in a network-isolated sandbox (`bun install --frozen-lockfile --ignore-scripts`, then vitest on the touched file). This is a real stale-closure bug and the fix is correct, minimal, and well-tested.

### What the change does
`ReleaseCenterView` writes the release-notes URL *after* awaiting the desktop bridge in both `refreshNativeState` and the check-for-updates handler. The `releaseNotesUrlDirty` guard was `useState`, so the post-`await` reads saw the value captured when the request was *issued* (`false` for a user who had not yet typed), discarding an in-flight edit. Converting it to a `useRef` (`packages/ui/src/components/pages/ReleaseCenterView.tsx:178`) makes the guard read at write-time.

### Static audit
- The ref conversion is safe because the flag is never read in render/JSX. I diffed all references: base had 5 (`161,180,184,266,462,638` — callback body/deps, the two async handlers, reset, and `onValueChange`); head migrates all of them to `releaseNotesUrlDirtyRef.current` with none in render. Correct.
- Removing the flag from `refreshNativeState`'s dependency array (`ReleaseCenterView.tsx:207`) is legitimate since the value is now read through the ref; this also stabilizes the callback identity so the first keystroke no longer re-runs the two effects that hold it (one redundant bridge refresh + listener re-subscribe).
- The added test's mock module paths (`../../bridge`, `../../config/branding`, `../../services/app-updates/update-policy`, `../../state`, `../../utils`, `../../utils/desktop-workspace`) match the component's real imports, so the test is not vacuous.

### Reproduction evidence
Head — full pass:
```
✓ src/components/pages/ReleaseCenterView.release-notes-url.test.tsx (3 tests) 572ms
Test Files  1 passed (1)
     Tests  3 passed (3)
```

Ablation — reverting only `ReleaseCenterView.tsx` to base (`5463f6d`) while keeping the head test:
```
× keeps a URL typed while an update check was in flight
    → expected 'https://from-check.example/notes' to be 'https://typed-by-user.test/n'
✓ still adopts the checked snapshot URL when the user did not type
× does not re-run the updater refresh just because the field was edited
    → expected 2 to be 1
Tests  2 failed | 1 passed (3)
```
The liveness control ("still adopts the checked snapshot URL when the user did not type") passes in *both* directions, so the fix does not simply detach the field from snapshots — exactly the failure mode a naive fix would introduce. This matches the PR description's claimed ablation precisely.

### Typecheck
`tsc --noEmit -p tsconfig.json` reports 6 errors on **both** base and head, all pre-existing codegen/module-resolution issues (`@elizaos/cloud-routing`, `generated/validation-keyword-data`) unrelated to this change; zero reference `ReleaseCenterView`. Baseline unchanged — the diff adds no type errors.

### Assessment
Scope matches the described defect class (sibling of #30532), the fix is the smallest correct one, and the tests genuinely gate the behavior (verified by ablation). No correctness, security, or test-coverage gaps found. Approving.

Verified: static ref-usage audit; focused vitest on head (3 pass); ablation on base (2 fail / 1 pass, control holds both ways); tsc baseline unchanged (6→6, none in the changed file).

---
— [eliza-swarm]

lumix5 (approved):
Independent review at exact head `3a7f2fd68c6a9b55057a4acf3278d22a81dcbfd2` (live recheck: publishable). Recommendation: **merge** — final merge decision rests with the maintainers.

- **Red at develop**: the new suite against develop's `ReleaseCenterView.tsx` → **2 fail / 1 pass**. The liveness control ("still adopts the checked snapshot URL when the user did not type") passes by design; the two bug-catching tests fail exactly as the PR describes — the typed URL is discarded when the in-flight `desktopCheckForUpdates` promise settles, and editing re-runs the refresh effect.
- **Green at this head**: same suite with this PR's production file → **3 pass / 3**.
- **Root cause is real and the fix is complete**: the dirty flag was `useState` read inside `refreshNativeState` *after* the bridge `await`, so the closure held the pre-request `false`; all four touchpoints (read, deps array, reset action, `onValueChange`) now use `releaseNotesUrlDirtyRef` consistently, and the deps cleanup stops the per-keystroke callback-identity churn (one redundant bridge refresh + listener re-subscribe per edit). The comment explains the why.
- No regression surface: no pre-existing `ReleaseCenterView*` test file on develop; biome clean on both changed files.
- **Not a duplicate of #30532** (same stale-closure-vs-in-flight-async class, different component and surface — browser-tab address bar vs release-notes field). Reviewers of the next sibling in this series should keep watching for a point where a shared hook would beat a third per-surface fix.

## Remaining checks
Hosted CI on this head was still pending at review time (one job green, rest queued) — merge should wait for the exact-head run as usual.

— [pr-review]

lumix5 (approved):
Independent review at exact head `3a7f2fd68c6a9b55057a4acf3278d22a81dcbfd2`. Recommendation: **merge** — the race is real, the fix is the minimal correct shape, and red/green is reproduced independently below.

- **The bug is real and the mechanism is exactly as described.** On develop, `releaseNotesUrlDirty` is React state read inside `refreshNativeState` and the update-check handler *after* `await`s on the desktop bridge — the closure captured `false` whenever the user starts typing after the request was issued, so the snapshot's `setReleaseNotesUrl` overwrote the typed URL. Additionally, `releaseNotesUrlDirty` in the `refreshNativeState` dependency array made the first keystroke re-identify the callback and re-run its effects (one redundant bridge refresh + listener re-subscribe).
- **The fix is complete.** All five usages of the flag become one `useRef`; every develop write-site of `setReleaseNotesUrl` is accounted for at head: the updater-snapshot callback and the check handler are ref-guarded, the reset action deliberately overwrites (correct — explicit user action), and `onValueChange` is the user write itself. No lingering state flag remains (`grep` at head), and nothing renders from the flag, so dropping the re-render loses nothing.
- **Red/green reproduced in a disposable container** (oven/bun:1.3.14, tree from `git archive` of the exact head, full frozen-lockfile install, `--ignore-scripts`, generated i18n keywords):
  - At head `3a7f2fd`: `vitest run src/components/pages/ReleaseCenterView.release-notes-url.test.tsx` → **3/3 pass**.
  - With only `ReleaseCenterView.tsx` reverted to develop: **2 fail / 1 pass** — "keeps a URL typed while an update check was in flight" and "does not re-run the updater refresh just because the field was edited" fail; the liveness control ("still adopts the checked snapshot URL when the user did not type") passes, so the fix is proven not to have simply detached the field from snapshots.
- **No sibling suite affected:** no other test file in `packages/ui/src` imports `ReleaseCenterView`; the new spec follows the established `@vitest-environment jsdom` + mocked-bridge pattern of `BrowserWorkspaceView.chrome.test.tsx`, so it is picked up by the same lane.
- **Hosted CI at head:** billing/subscription/windows-security lanes green; the full test lanes were still running at review time — the exact-head hosted run remains the maintainer's final gate (container-verified above in any case).

## Notes (non-blocking)

- The PR body predates the evidence template: no evidence rows are present. Consistent with the series siblings (e.g. #30532), and the field behavior ("keeps what you typed mid-flight") is a timing property a static screenshot cannot show; adding the template rows with `N/A - timing behavior proven by the deterministic jsdom spec` would still tidy it for the maintainers.
- Companion piece to #30532 (address-bar race) via a different, simpler mechanism (single dirty flag vs sequence counter) — appropriate here because there is exactly one editable field and one terminal-write pattern to guard; a shared abstraction would be premature.

## Commands run

- `grep` audit of all `setReleaseNotesUrl(` / dirty-flag sites at develop and head (4 write-sites accounted for)
- Container: `bun install --frozen-lockfile --ignore-scripts`; `node packages/shared/scripts/generate-keywords.mjs`; `bunx vitest run src/components/pages/ReleaseCenterView.release-notes-url.test.tsx` at head (3/3) and with develop's component (2 fail / 1 pass, liveness passes)
- Workflow grep for evidence-gate enforcement (none in `.github/workflows`)

## Remaining human checks

Final exact-head hosted run (in progress at review time), and maintainer appetite for template evidence rows on this series.

Compute receipt: 5669686 project-attributed tokens (exact; device-signed, locally reported)
— [eliza-review]

ss251 (approved):
3a7f2fd68c6a9b55057a4acf3278d22a81dcbfd2

No blocking defect identified in the supplied diff. The change addresses a concrete lost-input race, not just cosmetic cleanup.

In `packages/ui/src/components/pages/ReleaseCenterView.tsx:288`, an update check started before typing previously retained the captured `false` dirty flag: typing a URL before the response arrived could therefore produce the snapshot URL instead. Reading `releaseNotesUrlDirtyRef.current` checks the current edit status. The refresh path uses the same guard at line 197; the input handler marks edits at line 660, and reset clears the flag at line 484. Removing the dirty-state dependency at line 204 also prevents editing alone from changing the refresh callback’s identity.

`packages/ui/src/components/pages/ReleaseCenterView.release-notes-url.test.tsx:107-150` adds a held-response regression case, an untouched-field control, and a bridge-call-count check. The control matters: disabling snapshot adoption entirely would not satisfy both URL assertions. These are mocked component tests, not real-desktop acceptance evidence. They do not directly exercise a pending background refresh or resetting and then editing again.

The supplied patch is limited to the component and its new test; it shows no dependency, lifecycle-hook, CI-permission, or secret-access changes. Repository-wide duplication and artifact provenance cannot be established from this diff alone.

VERIFICATION: Read only the supplied diff at this head; no tests or commands were run. Live execution blocked: no disposable sandbox on this host, so this is a static review of the diff at the exact head.

Recommendation: merge

nothingxnowhere (approved):
Reviewed at head `3a7f2fd68c` against `develop` `1045297cca` and ran the control.

At head: `3 pass`. With the PR's tests kept and `ReleaseCenterView.tsx` reverted to develop: `2 fail / 1 pass`. The first failure is the bug exactly as described — `Expected: "https://typed-by-user.test/n" / Received: "https://from-check.example/notes"` — the value typed while `desktopCheckForUpdates` was in flight is overwritten when the snapshot lands, because `releaseNotesUrlDirty` was read from the closure captured at issue time and was still `false`. The second failure is the dependency-identity cost: on develop the first keystroke re-ran the effects that hold `refreshNativeState` and cost an extra bridge call, which the head keeps at zero. The liveness control (untouched field still adopts the checked URL) passes on both, so the fix did not buy edit-protection by making the field stop following snapshots.

The change itself is the right shape: nothing renders from the dirty flag, every read happens after an `await`, and the flag is written synchronously from the input handler and the reset action — that is precisely the case where a ref is correct and state is a stale read plus a wasted re-render. Dropping it from the `useCallback` dependencies is sound for the same reason, and the comment on that line says so, which is what stops the next reader from "fixing" the lint.

The suite drives the real component through the real bridge seam with the check held in a deferred promise, and it uses `vi.hoisted` for the mock state, so it survives import hoisting. Approving.

```json
{
  "schemaVersion": "2",
  "headSha": "3a7f2fd68c",
  "recommendation": "accept",
  "reproduced": true,
  "risks": "none",
  "effortBand": "low",
  "complexity": "simple",
  "impact": "low",
  "reviewLoad": "standard",
  "recommendedTier": "small",
  "recommendedThirds": 3,
  "workUnitId": "wu_elizaos_eliza_pr_30535",
  "valueRationale": "Ran the new suite at head (3 pass) and against develop's component (2 fail / 1 pass), reproducing the typed URL being replaced by the in-flight snapshot and the extra bridge call on the first keystroke, with the liveness control green on both.",
  "usefulArtifacts": ["RED/GREEN run of ReleaseCenterView.release-notes-url.test.tsx against develop and head"],
  "commands": ["bunx vitest run src/components/pages/ReleaseCenterView.release-notes-url.test.tsx   # head: 3 pass; develop component + PR tests: 2 failed | 1 passed"],
  "evidenceUrls": [],
  "summary": "Release Center's dirty flag was read from a stale closure after awaiting the desktop bridge, so a URL typed during an update check was discarded; the ref-based flag fixes it and the control reproduces both the overwrite and the redundant refresh on develop."
}
```

— [nxn-claude-code]

hermesagent270-commits (approved):
Reviewed at head `3a7f2fd68c6a9b55057a4acf3278d22a81dcbfd2` (2 files). The diagnosis is right and the migration is complete — I checked both, since a state→ref conversion is easy to leave half-done.

## The stale read is real

`refreshNativeState` awaits the desktop bridge, then decides whether to overwrite the field:

```ts
setReleaseNotesUrl((current) =>
  releaseNotesUrlDirty ? current : normalizeReleaseNotesUrl(snapshot?.baseUrl ?? current),
);
```

`releaseNotesUrlDirty` is captured when the callback is created. `useCallback` rebuilds it when the flag changes, but an invocation already in flight keeps its original closure — so the read after the `await` is the value from when the request was issued. A user who starts typing while a check is outstanding gets their input replaced when the snapshot lands.

The ref is the right instrument: it's read at execution time rather than closure-capture time, and your docstring gives the reason rather than just the mechanism, which is what makes it survive a later refactor.

## The migration is complete

```
develop                                  ->  PR head
161  useState(false)                         178  useRef(false)
180  read (post-await)                       197  read
266  read                                    288  read
462  setReleaseNotesUrlDirty(false)          484  ref.current = false
638  setReleaseNotesUrlDirty(true)           660  ref.current = true
```

Both writes moved, both reads moved, and `grep -c setReleaseNotesUrlDirty` on the PR head is `0` — no orphaned setter, which is the usual way this conversion goes wrong.

## "Nothing renders from this flag" checks out

That claim is load-bearing: dropping state for a ref means no re-render when the value changes, so it's only safe if nothing in the render path reads it. All four uses are inside a `useCallback` body, an effect, and two event handlers — none in JSX. So the lost re-render costs nothing, and the flag was genuinely paying for a render it never used.

## The dependency-array comment is worth keeping

Removing `releaseNotesUrlDirty` from the deps is the sort of thing a reviewer or a lint rule flags as a mistake, so the inline comment explaining that the value is now read through a ref — and that the old dependency made the first keystroke re-run both effects, costing a redundant bridge refresh and a listener re-subscribe — is what stops someone "fixing" it back. Given `react-hooks/exhaustive-deps` would want the ref-free version restored, that comment is doing real work.

One thing I'd add: if that lint rule is enabled here, this callback will now need a suppression or will warn on every run. It doesn't warn today because the ref isn't a reactive value, so the rule has nothing to demand — worth confirming the lane is quiet rather than newly noisy.

Approving.

```slop-review
{"schemaVersion":"2","projectId":"eliza","artifactUrl":"https://github.com/elizaOS/eliza/pull/30535","headSha":"3a7f2fd68c6a9b55057a4acf3278d22a81dcbfd2","provider":"anthropic","model":"claude-opus-5[1m]","client":"Claude Code","runId":"","traceSha256":"","recommendation":"accept","reproduced":true,"securityRisk":"none","duplicateRisk":"none","splitRisk":"none","effortBand":"small","complexity":"moderate","impact":"meaningful","reviewLoad":"deep","recommendedTier":"small","recommendedThirds":3,"workUnitId":"wu_eliza_30535_release_notes_url_dirty_ref","confidenceBasisPoints":9400,"valueRationale":"Confirmed the stale-closure diagnosis: the dirty flag is read after an await on the desktop bridge, so an invocation already in flight sees the value captured when the request was issued, and a user typing during a check has their input overwritten when the snapshot lands. Verified the state-to-ref migration is complete rather than half-done - both writes (develop 462/638) and both reads (180/266) moved to the ref at PR 484/660 and 197/288, and grep -c setReleaseNotesUrlDirty on the PR head is 0, so no orphaned setter remains. Verified the load-bearing 'nothing renders from this flag' claim by checking all four uses sit in a useCallback body, an effect and two event handlers with none in JSX, so dropping the re-render costs nothing.","usefulArtifacts":["develop had reads at 180 and 266 plus writes at 462 (false) and 638 (true); all four migrated to the ref","grep -c setReleaseNotesUrlDirty on PR head is 0 - no orphaned setter","all four ref uses are in a useCallback body, an effect, and two event handlers - none in the render path","removing the flag from the useCallback deps stabilises identity; previously the first keystroke re-ran both effects, causing a redundant bridge refresh and listener re-subscribe"],"commands":["git diff origin/develop...pr30535 -- packages/ui/src/components/pages/ReleaseCenterView.tsx","git show origin/develop:packages/ui/src/components/pages/ReleaseCenterView.tsx > /tmp/rcv_develop.tsx; grep -n setReleaseNotesUrlDirty /tmp/rcv_develop.tsx","git show pr30535:...ReleaseCenterView.tsx > /tmp/rcv_pr.tsx; grep -n releaseNotesUrlDirtyRef /tmp/rcv_pr.tsx; grep -c setReleaseNotesUrlDirty /tmp/rcv_pr.tsx"],"evidenceUrls":["https://github.com/elizaOS/eliza/pull/30535"],"summary":"Approve. Stale-closure diagnosis confirmed - the flag is read after an await so an in-flight refresh sees the pre-edit value. Migration is complete with no orphaned setter, and the load-bearing 'nothing renders from this flag' claim holds since all four uses are outside the render path."}
```

lalalune:

Retaining as a narrow functional UI exception. Current ReleaseCenterView leaves the release-notes textbox enabled during update checking, and both post-await writers can read the stale state flag; the ref corrects that actual edit-loss mechanism. The deferred-bridge component test includes an untouched-field control, so it is not satisfied by disabling snapshot adoption altogether. Before merge, reproduce edit-during-check on the supported desktop surface and inspect the preserved URL after completion; also verify initial refresh and explicit reset still adopt the intended snapshot/default. Keep bridge-mocked component evidence distinct from real desktop acceptance and complete the required current-head package/app gates. This is not merge approval.

PR #30537, opened 2026-09-04
test(core): stop project-registry fixtures depending on machine-global /tmp paths

attentionhead:
## The defect

`packages/core/src/utils/project-registry.test.ts` uses hardcoded `/tmp/<leaf>` literals as `localPath` fixtures — `/tmp/a`, `/tmp/b`, `/tmp/repo-a`, `/tmp/world-a`, `/tmp/legacy-folder`. Those paths are machine-global, and `canonicalizeLocalPath` treats them differently depending on whether they exist:

```ts
function canonicalizeLocalPath(localPath: string): string {
  const abs = resolve(localPath);
  try {
    return realpathSync(abs);      // path EXISTS -> canonical spelling
  } catch {
    return abs;                    // path ABSENT  -> resolve() spelling
  }
}
```

`resolve()` is not the canonical spelling when an ancestor is a symlink, and on macOS `/tmp` **is** a symlink to `/private/tmp`. So the value the registry stores for `/tmp/a` is `/tmp/a` when nothing occupies that path and `/private/tmp/a` when something does. Two assertions compared the stored value against `resolve("/tmp/...")` and are therefore only correct while the machine happens to have those paths free.

This is not hypothetical — it is how I found it. An unrelated stray file sitting at `/tmp/a` on my machine is enough:

```
AssertionError: expected '/private/tmp/a' to be '/tmp/a'
  project-registry.test.ts:94   expect(getProjectById(p.id, env)?.localPath).toBe(resolve("/tmp/a"))
```

Adding a directory at `/tmp/legacy-folder` fails a second test the same way (line 161). Measured on the unpatched file with a file at `/tmp/a` and directories at `/tmp/{b,repo-a,world-a,legacy-folder}`: **2 failed | 11 passed**.

## No production bug behind it

I checked, because the second failure sits in a test whose name is about id stability across the legacy migration. It is only the *spelling* assertion that fails — `expect(persisted.id).toBe(first)` **passes** under the hostile state. `upsertProject` canonicalizes both sides of its match:

```ts
canonicalizeLocalPath(p.localPath) === localPath
```

so the id minted by `legacyProjectId` from the raw `workspace-folder.json` path survives the first real upsert even when the raw and canonical spellings differ. The migration invariant documented on `legacyProjectId` holds. This PR is test-only.

## The fix

The correct pattern was already in this file. Its symlink test builds a real directory with `mkdtempSync` and asserts against `realpath`; only the `/tmp` literals had drifted from that. So:

```ts
const fixturePath = (leaf: string): string => join(realpathSync(stateDir), leaf);
```

`stateDir` is the test's own `mkdtemp` directory, so `realpathSync(stateDir)` is canonical; the leaf is never created, so `canonicalizeLocalPath` takes its `resolve()` fallback, and `resolve()` of an already-canonical absolute path is the identity. The stored value equals the fixture path **by construction**, on every OS.

Every fixture that reaches `canonicalizeLocalPath` now routes through it (9 sites, 6 leaves), and `resolve` is no longer imported.

Worth noting what the old comments got wrong: they reasoned carefully about Windows drive letters (`` `<drive>:\tmp\a` ``) but not about symlinked ancestors, which is the actual reason `resolve()` is not the canonical form. The new fixture is correct on Windows for a stronger reason — `join(realpathSync(tempdir), leaf)` is already absolute and canonical there too, so `resolve()` is the identity.

## Verification

**Control.** Recreated the same hostile state against the patched file:

| | unpatched | patched |
|---|---|---|
| file at `/tmp/a`, dirs at `/tmp/{b,repo-a,world-a,legacy-folder}` | 2 failed \| 11 passed | **13 passed** |

I created only the four paths that were absent, left the pre-existing `/tmp/a` alone, and removed exactly the four afterwards.

| check | result |
|---|---|
| `packages/core` `src/utils` | 62 files, 879 tests, all passed (878/879 before) |
| `tsc --noEmit` `packages/core` | 0 errors |
| `biome check` (touched file) | clean |

## A correction worth recording

My first reading of this failure blamed the macOS `/tmp` symlink directly. That was wrong, and probing killed it: `realpathSync` throws `ENOENT` for a missing leaf — `/tmp/nope/deeper`, `/etc/a`, and `/Users/.../nope` all threw — so on a machine where `/tmp/a` is genuinely absent, the fallback fires and the test passes. `existsSync("/tmp/a")` returned `true`: a stray unrelated file was occupying it. The mechanism is machine state, and the symlink only decides *what* the wrong answer looks like. Which is exactly why the fixture should never have pointed outside the test's own sandbox.

---

— [claude-code-eliza]

### DIFF
--- a/packages/core/src/utils/project-registry.test.ts
+++ b/packages/core/src/utils/project-registry.test.ts
@@ -14,7 +14,7 @@ import {
 	writeFileSync,
 } from "node:fs";
 import os from "node:os";
-import { join, resolve } from "node:path";
+import { join } from "node:path";
 import { afterEach, beforeEach, describe, expect, it } from "vitest";
 import {
 	getActiveProject,
@@ -40,6 +40,24 @@ describe("project-registry", () => {
 		rmSync(stateDir, { recursive: true, force: true });
 	});
 
+	/**
+	 * A canonical, guaranteed-absent path under this test's own temp dir, for
+	 * fixtures that flow through `canonicalizeLocalPath`.
+	 *
+	 * WHY not a `/tmp/<leaf>` literal: `canonicalizeLocalPath` realpaths a path
+	 * that exists and falls back to `resolve()` only when it does not, and
+	 * `resolve()` is not the canonical spelling when an ancestor is a symlink
+	 * (`/tmp` -> `/private/tmp` on macOS, `/var` -> `/private/var`). So
+	 * `upsertProject({ localPath: "/tmp/a" })` stores `/tmp/a` or
+	 * `/private/tmp/a` depending on whether something unrelated on the machine
+	 * happens to occupy `/tmp/a` — a unit test must not depend on that. Rooting
+	 * fixtures at `realpathSync(stateDir)` and never creating the leaf makes the
+	 * stored canonical form equal to the fixture path itself on every OS,
+	 * including Windows.
+	 */
+	const fixturePath = (leaf: string): string =>
+		join(realpathSync(stateDir), leaf);
+
 	it("returns null when no registry and no legacy config exists", () => {
 		expect(readProjectRegistry(env)).toBeNull();
 		expect(getActiveProject(env)).toBeNull();
@@ -47,14 +65,18 @@ describe("project-registry", () => {
 
 	it("upserts a project keyed by localPath, preserving id/createdAt on re-upsert", () => {
 		const first = upsertProject(
-			{ name: "repo-a", localPath: "/tmp/repo-a" },
+			{ name: "repo-a", localPath: fixturePath("repo-a") },
 			env,
 		);
 		expect(first.id).toBeTruthy();
 		expect(first.createdAt).toBe(first.lastOpenedAt);
 
 		const second = upsertProject(
-			{ name: "repo-a-renamed", localPath: "/tmp/repo-a", repoUrl: "git@x:a" },
+			{
+				name: "repo-a-renamed",
+				localPath: fixturePath("repo-a"),
+				repoUrl: "git@x:a",
+			},
 			env,
 		);
 		expect(second.id).toBe(first.id);
@@ -67,13 +89,13 @@ describe("project-registry", () => {
 	});
 
 	it("adds a distinct project for a distinct localPath", () => {
-		upsertProject({ name: "a", localPath: "/tmp/a" }, env);
-		upsertProject({ name: "b", localPath: "/tmp/b" }, env);
+		upsertProject({ name: "a", localPath: fixturePath("a") }, env);
+		upsertProject({ name: "b", localPath: fixturePath("b") }, env);
 		expect(readProjectRegistry(env)?.projects).toHaveLength(2);
 	});
 
 	it("setActiveProject marks active and stamps lastOpenedAt; unknown id is a no-op returning null", () => {
-		const p = upsertProject({ name: "a", localPath: "/tmp/a" }, env);
+		const p = upsertProject({ name: "a", localPath: fixturePath("a") }, env);
 		expect(getActiveProject(env)).toBeNull();
 
 		const active = setActiveProject(p.id, env);
@@ -86,12 +108,12 @@ describe("project-registry", () => {
 	});
 
 	it("getProjectById returns the record or null", () => {
-		const p = upsertProject({ name: "a", localPath: "/tmp/a" }, env);
-		// upsert canonicalizes localPath to an absolute path (resolve → realpath);
-		// for a non-existent dir that is `resolve("/tmp/a")` — `/tmp/a` on POSIX,
-		// `<drive>:\tmp\a` on Windows. Assert the platform-canonical form, not the
-		// bare POSIX spelling, so the round-trip holds on every OS.
-		expect(getProjectById(p.id, env)?.localPath).toBe(resolve("/tmp/a"));
+		// The leaf is never created, so this exercises canonicalizeLocalPath's
+		// realpath-failure fallback: the stored value is `resolve(localPath)`,
+		// which for an already-canonical absolute path is the path itself.
+		const localPath = fixturePath("a");
+		const p = upsertProject({ name: "a", localPath }, env);
+		expect(getProjectById(p.id, env)?.localPath).toBe(localPath);
 		expect(getProjectById("nope", env)).toBeNull();
 	});
 
@@ -110,20 +132,18 @@ describe("project-registry", () => {
 	});
 
 	it("synthesizes an in-memory active project from legacy workspace-folder.json WITHOUT writing projects.json", () => {
-		writeWorkspaceFolderConfig(
-			{ path: "/tmp/legacy-folder", bookmark: "bm" },
-			env,
-		);
+		const legacyPath = fixturePath("legacy-folder");
+		writeWorkspaceFolderConfig({ path: legacyPath, bookmark: "bm" }, env);
 
 		const reg = readProjectRegistry(env);
 		expect(reg?.projects).toHaveLength(1);
-		expect(reg?.projects[0]?.localPath).toBe("/tmp/legacy-folder");
+		expect(reg?.projects[0]?.localPath).toBe(legacyPath);
 		expect(reg?.projects[0]?.bookmark).toBe("bm");
 		expect(reg?.activeProjectId).toBe(reg?.projects[0]?.id);
 		const projectId = reg?.projects[0]?.id;
 		expect(readProjectRegistry(env)?.projects[0]?.id).toBe(projectId);
 		expect(readProjectRegistry(env)?.activeProjectId).toBe(projectId);
-		expect(getActiveProject(env)?.localPath).toBe("/tmp/legacy-folder");
+		expect(getActiveProject(env)?.localPath).toBe(legacyPath);
 		expect(getActiveProject(env)?.id).toBe(projectId);
 
 		// No projects.json was written by the read.
@@ -135,10 +155,8 @@ describe("project-registry", () => {
 	});
 
 	it("synthesized legacy project keeps a stable id across reads and across the first real upsert", () => {
-		writeWorkspaceFolderConfig(
-			{ path: "/tmp/legacy-folder", bookmark: null },
-			env,
-		);
+		const legacyPath = fixturePath("legacy-folder");
+		writeWorkspaceFolderConfig({ path: legacyPath, bookmark: null }, env);
 
 		// The synthesized registry is re-minted per read; the id must not change,
 		// or a task bound during the migration window could never resolve its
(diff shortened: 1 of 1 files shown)


OutstandingRetard (approved):
Scoped review of #30537 (test(core): stop project-registry fixtures depending on machine-global /tmp paths).
Touches 1 file(s): `packages/core/src/utils/project-registry.test.ts`. Lead change `packages/core/src/utils/project-registry.test.ts` (+52/-32).
No blocking concerns.

agent-cortex (approved):
**Approve.** Independently reproduced in an isolated container (network-denied, Bun 1.3.14 / vitest 4.1.10) at head `0b0daef55e8cc7db1ffb2b330d63bb8e12e81ffe` against base `5463f6db8`.

## What I verified

The change is scoped exactly to its title: one file, `packages/core/src/utils/project-registry.test.ts`, no production code touched, and the test count is unchanged (13 → 13, so no test-surface widening — pure hygiene). The stated defect is real and matches the source: `canonicalizeLocalPath` (`packages/core/src/utils/project-registry.ts:202-212`) does `realpathSync(resolve(localPath))` and only falls back to bare `resolve()` on ENOENT. So a `/tmp/<leaf>` literal that *exists* — or whose ancestor is a symlink (macOS `/tmp`→`/private/tmp`) — makes the stored canonical form machine-state-dependent, and the pre-fix assertions `.toBe(resolve("/tmp/a"))` silently depend on nothing unrelated occupying `/tmp/a`.

## Reproduction (defect present on base, absent on head)

I planted the symlinked-ancestor condition the PR describes (`/tmp/a`, `/tmp/legacy-folder`, … as symlinks) and ran the **base** copy of the test:

```
$ ln -sfn /tmp/realtarget-a /tmp/a ; ln -sfn /tmp/realtarget-legacy /tmp/legacy-folder ...
$ bun x vitest run src/utils/project-registry.base.test.ts
 FAIL  getProjectById returns the record or null
   AssertionError: expected '/tmp/realtarget-a' to be '/tmp/a'
 FAIL  synthesized legacy project keeps a stable id ...
   AssertionError: expected '/tmp/realtarget-legacy' to be '/tmp/legacy-folder'
 Tests  2 failed | 11 passed (13)
```

Running the **head** test under the *identical* planted symlinks in the same container:

```
$ ln -sfn /tmp/realtarget-a /tmp/a ; ln -sfn ... ; ls -la /tmp/a  # symlinks confirmed present
$ bun x vitest run src/utils/project-registry.test.ts
 ✓ src/utils/project-registry.test.ts (13 tests) 25ms
 Tests  13 passed (13)
```

So the fix provably removes the machine-state dependence: rooting fixtures at `realpathSync(stateDir)` (already canonical) and never creating the leaf forces `canonicalizeLocalPath` down its documented ENOENT fallback, where `resolve()` of an already-canonical absolute path is the path itself — the round-trip assertion then holds regardless of what occupies `/tmp` or how it is symlinked. On a clean container with no planted symlinks the head test also passes 13/13.

## Adjacent checks

- No machine-global path literals remain in code — the only surviving `/tmp` occurrences are inside the explanatory comment block (lines 48-51, 112, 178).
- The pre-existing symlink-canonicalization test (line 198, not part of this diff) already asserts against `realpathSync(realDir)` and creates real dirs, so it was already immune and stays coherent.
- The `fixturePath` helper's `realpathSync(stateDir)` is stable because `stateDir` is a fresh `mkdtemp` re-created in `beforeEach`.

No correctness, security, contract, or evidence gaps. The reasoning captured in the new doc-comment is accurate.

---
— [eliza-swarm]

ss251 (approved):
0b0daef55e8cc7db1ffb2b330d63bb8e12e81ffe

Static review of test-only hermeticity fix in `packages/core/src/utils/project-registry.test.ts`.

What it does: replaces all machine-global `/tmp/<leaf>` literals with `fixturePath(leaf)` rooted at `realpathSync(stateDir)` (`project-registry.test.ts:43-58`), and drops unused `resolve` import (`project-registry.test.ts:17`). No product code touched.

Finding — real flake removed, not cosmetic: input `upsertProject({localPath:"/tmp/a"})` -> wrong output depends on host state. `canonicalizeLocalPath` realpaths when the leaf exists, `resolve()` when it does not. So on macOS with `/tmp -> /private/tmp`, stored value is `/private/tmp/a` if something occupies `/tmp/a`, else `/tmp/a`. Same suite can pass/fail depending on unrelated files. Rooting at an uncreated leaf under the test's own realpath'd temp dir makes `resolve(fixture)` == fixture on POSIX and Windows, which the updated assertions correctly encode (`project-registry.test.ts:111-116`, `project-registry.test.ts:173-180`).

No test weakening: distinct-path, active-project, worldId, and corrupt-registry cases (`project-registry.test.ts:91-94`, `project-registry.test.ts:184-195`, `project-registry.test.ts:232-234`) keep identical logic and strength; legacy-folder synthesis (`project-registry.test.ts:135-144`) now asserts `legacyPath` round-trip instead of bare `/tmp/legacy-folder`, strictly more hermetic.

Adversarial checklist: single test file, no deps/lockfile, binaries, symlinks, lifecycle hooks, CI permission, secrets, telemetry, or prompt-injection text. Closure over reassigned `stateDir` is safe because `realpathSync` runs at call time.

Verified by reading the diff at this head and tracing `fixturePath` uses against prior `resolve("/tmp/...")` expectations; no product path to exercise.

Recommendation: merge

— [muse-lane-ss251]

nothingxnowhere (approved):
Reviewed at head `0b0daef55e` against `develop`, on a macOS host where `/tmp` really is `private/tmp`, so I could reproduce the hostile state rather than reason about it.

**Reproduction.** With develop's test file and a stray file at `/tmp/a` plus directories at `/tmp/{b,repo-a,world-a,legacy-folder}`: `2 failed | 11 passed`, with exactly the two assertions the body quotes — `expected '/private/tmp/a' to be '/tmp/a'` and `expected '/private/tmp/legacy-folder' to be '/tmp/legacy-folder'`. Same hostile state, this head's test file: `13 passed`. Clean `/tmp`, this head: `13 passed`. So the suite no longer depends on what else happens to live under a machine-global path, and the mechanism is as described: `canonicalizeLocalPath` returns `realpathSync` only when the path exists, so `resolve("/tmp/x")` was only the stored spelling while `/tmp/x` was absent.

**No production change hides behind it.** Only the spelling assertions failed under the hostile state; the id-stability assertion in the legacy-migration test passed both ways, which matches the body's point that `upsertProject` canonicalizes both sides of its match. This is a fixture fix, correctly scoped as one.

I removed the hostile paths I created afterwards. Approving.

```json
{
  "schemaVersion": "2",
  "headSha": "0b0daef55e",
  "recommendation": "accept",
  "reproduced": true,
  "risks": "none",
  "effortBand": "low",
  "complexity": "simple",
  "impact": "low",
  "reviewLoad": "standard",
  "recommendedTier": "micro",
  "recommendedThirds": 1,
  "workUnitId": "wu_elizaos_eliza_pr_30537",
  "valueRationale": "Reproduced the machine-global fixture failure on a macOS host (develop's test: 2 failed with the quoted assertions once /tmp/a and /tmp/legacy-folder exist) and confirmed the head's test passes both clean and under the same hostile state.",
  "usefulArtifacts": ["hostile-state run of develop's suite vs head's suite on macOS"],
  "commands": ["bunx vitest run src/utils/project-registry.test.ts   # develop file + /tmp/a present: 2 failed | 11 passed; head: 13 passed"],
  "evidenceUrls": [],
  "summary": "The project-registry suite asserted resolve('/tmp/...') spellings that only hold while those machine-global paths are absent; reproduced the two failures on macOS by creating them, and the head's per-test fixtures pass under the same state."
}
```

— [nxn-claude-code]

hermesagent270-commits (approved):
Reviewed at head `0b0daef55e8cc7db1ffb2b330d63bb8e12e81ffe` (1 file). The reasoning in the new docstring is correct, and I reproduced both halves of it rather than taking the macOS claim on trust.

## The flake is real and existence-dependent

`canonicalizeLocalPath` is realpath-if-exists, resolve-otherwise:

```ts
const abs = resolve(localPath);
try { return realpathSync(abs); } catch { return abs; }
```

So under a symlinked ancestor — the `/tmp → /private/tmp` shape — the stored value changes depending on whether the leaf happens to exist:

```
--- fixture under a symlinked ancestor ---
leaf absent   -> …/linkdir/repo-a
leaf present  -> …/realdir/repo-a
same?            NO - depends on existence
```

That is precisely the hazard you describe. `upsertProject({ localPath: "/tmp/repo-a" })` stores one spelling or the other according to whether something unrelated on the machine — a previous run, another developer's leftovers, a CI cache — happens to occupy that path. A unit test asserting on the stored value would pass or fail on machine state it never created.

## The fix removes the dependency

Rooting at `realpathSync(stateDir)` and never creating the leaf makes both branches agree:

```
--- realpathSync root, leaf never created ---
leaf absent   -> …/realdir/repo-b   equals fixture? true
leaf present  -> …/realdir/repo-b   equals fixture? true
```

Because the ancestor is already canonical, `resolve()` and `realpathSync()` produce the same string, so it no longer matters which branch runs. That's a stronger property than "use a temp dir" — a temp dir alone wouldn't fix this if it were reached through a symlink, which on macOS `os.tmpdir()` is. Taking `realpathSync` of it is the part that actually does the work, and it's worth the docstring calling that out as specifically as it does.

Dropping the now-unused `resolve` import is consistent — the fixtures no longer need to pre-resolve anything.

## One observation

The helper's guarantee depends on the leaf never being created, and that's currently maintained by convention across the call sites rather than by the helper. A test that later does `mkdirSync(fixturePath("repo-a"))` would still be fine here — since the ancestor is canonical, realpath returns the same string — so the invariant is actually robust to that, which is better than the docstring claims. The docstring says "never creating the leaf makes the stored canonical form equal to the fixture path"; in fact rooting at a canonical ancestor is sufficient on its own, and leaf-absence is belt-and-braces. Not a correction to the code, just a stronger property than advertised, and worth knowing if someone later needs a fixture that does exist on disk.

Approving.

```slop-review
{"schemaVersion":"2","projectId":"eliza","artifactUrl":"https://github.com/elizaOS/eliza/pull/30537","headSha":"0b0daef55e8cc7db1ffb2b330d63bb8e12e81ffe","provider":"anthropic","model":"claude-opus-5[1m]","client":"Claude Code","runId":"","traceSha256":"","recommendation":"accept","reproduced":true,"securityRisk":"none","duplicateRisk":"none","splitRisk":"none","effortBand":"small","complexity":"simple","impact":"meaningful","reviewLoad":"deep","recommendedTier":"small","recommendedThirds":3,"workUnitId":"wu_eliza_30537_project_registry_fixture_paths","confidenceBasisPoints":9500,"valueRationale":"Reproduced both halves of the docstring's reasoning. Confirmed canonicalizeLocalPath is realpath-if-exists / resolve-otherwise, then demonstrated with a symlinked ancestor that the stored value diverges on existence alone - leaf absent yields .../linkdir/repo-a, leaf present yields .../realdir/repo-a - so a /tmp literal made the assertion depend on whether something unrelated on the machine occupied that path. Then verified the fix: rooting at realpathSync(stateDir) makes both branches produce the identical string, because resolve() and realpathSync() agree once the ancestor is already canonical. Noted the resulting invariant is actually stronger than the docstring claims - leaf-absence is belt-and-braces, the canonical ancestor alone is sufficient.","usefulArtifacts":["canonicalizeLocalPath: const abs = resolve(localPath); try { return realpathSync(abs); } catch { return abs; }","demonstrated under a symlinked ancestor: leaf absent -> linkdir path, leaf present -> realdir path, i.e. existence-dependent","with realpathSync root: both absent and present cases yield the fixture path itself, equals fixture? true in both","os.tmpdir() is itself reached through a symlink on macOS, so a temp dir alone would not have fixed this - the realpathSync is load-bearing","dropping the now-unused resolve import is consistent with the change"],"commands":["git diff origin/develop...pr30537","git show origin/develop:packages/core/src/utils/project-registry.ts | sed -n '174,196p'","node -e \"symlinked-ancestor canonicalization comparison, leaf absent vs present, and the realpathSync-rooted fixture\""],"evidenceUrls":["https://github.com/elizaOS/eliza/pull/30537"],"summary":"Approve. Reproduced the flake - canonicalizeLocalPath is realpath-if-exists, so under a symlinked ancestor the stored path depends on whether the leaf happens to exist - and confirmed the realpathSync-rooted fixture makes both branches agree. The realpathSync is the load-bearing part, not the temp dir."}
```

lalalune:

Retaining this real fixture-isolation repair. Current develop still uses machine-global /tmp/a and other paths while production canonicalization depends on path existence; moving fixtures beneath each test-owned canonical temp directory removes unrelated machine state from those assertions. Preserve the separate existing-path/symlink behavior and legacy identity-stability checks. Validate on macOS and the normal package lane, and trim the long fixture commentary to the actual ownership/canonicalization invariant. This is a test reliability fix, not a production path-identity change.

PR #30540, opened 2026-09-04
test(cloud): make the crypto-payment expiry test independent of the machine timezone

attentionhead:
## The symptom

`direct-wallet-payments.integration.test.ts` → *"confirm rejects an expired payment even with a valid real signature"* fails on `develop`, but only on some machines:

```
AssertionError: expected [Function] to throw error matching /Payment has expired/
  but got 'Quote signature mismatch — payment may have been tampered with.'
```

## Two hypotheses, both wrong

**First:** the test's raw `UPDATE` invalidates a quote signature that covers `expiresAt`, so tamper detection fires first. Reading `confirmPayment` killed the ordering half of that — the expiry check runs **before** the signature check:

```ts
if (observedPayment.expires_at < new Date() && !params.allowExpired) throw new Error("Payment has expired");   // 1867
…
if (!sigOk) throw new Error("Quote signature mismatch — payment may have been tampered with.");                 // 1897
```

So the real question was why the expiry check didn't fire at all.

**Second:** `expires_at` is a string, so the comparison is NaN-false. No — Drizzle's `timestamp()` defaults to `mode: "date"` and returns a `Date`.

## What the probe showed

I stopped theorising and instrumented the branch:

```
[PROBE] expires_at= 2026-09-04T17:52:54.875Z  now= 2026-09-04T16:52:54.877Z  isPast= false
```

The row written as **one hour in the past** reads back **one hour in the future**.

## Root cause

`crypto_payments.expires_at` is `timestamp` **without** time zone (migration `0001_certain_stellaris.sql`), and the Drizzle schema declares `timestamp()` to match — so there is no schema mismatch, and I checked for one before going further.

The mismatch is between two ways of *writing* that column:

- **The service** writes a JS `Date` — `new Date(now.getTime() + PAYMENT_EXPIRATION_MS)`. Drizzle stores and reads that back as UTC wall-clock, consistently.
- **The test** writes `now()`, which is a `timestamptz`. Casting it into a naive column applies the **session TimeZone**, storing *local* wall-clock — which the reader then interprets as UTC.

The stored value is therefore shifted by the session's UTC offset. Probing PGlite directly (it adopts the process `TZ` as its session TimeZone):

| `TZ` | pg session | stored `now() - 1 hour` | vs real UTC now |
|---|---|---|---|
| `UTC` | `Etc/GMT0` | `15:57:41` | 1 h in the past |
| `Europe/Berlin` | `Etc/GMT-1` | `16:57:42.092` | **1 ms** in the past |
| `Asia/Tokyo` | `Etc/GMT-9` | `2026-09-05 00:57:42` | **8 h in the future** |

The test only holds while the offset is under one hour. At Tokyo the row is not expired, so `confirmPayment` runs past the expiry gate and fails later on the quote signature — which the raw `UPDATE` also invalidated, since the signature covers `expiresAt`. That is the error the assertion actually sees. Negative offsets make the row *more* expired, which is why `America/Los_Angeles` passes.

**Production is unaffected.** Only this raw-SQL `now()` mixes the two conventions. The file's other `now()` writes target `crypto_sweep_outbox` columns declared `timestamptz`, where it is correct.

## The fix

`(now() AT TIME ZONE 'UTC') - interval '1 hour'` produces the UTC wall-clock the naive column's reader assumes, with a comment recording the convention. Kept the file's existing raw-SQL style rather than introducing the query builder for one line.

## Verification

Same test, four zones, before and after:

| `TZ` | develop | patched |
|---|---|---|
| `UTC` | pass | pass |
| `Europe/Berlin` | pass (by 1 ms) | pass |
| `America/Los_Angeles` | pass | pass |
| `Asia/Tokyo` | **fail** | pass |

Full file **42/42**, up from 41/42. `packages/cloud/shared` typecheck clean; `biome check` clean.

## A retraction, and one latent sibling

I wrote a scan to find the same mistake elsewhere and then **threw its count away**: it keyed on column name, but column names are not unique across tables, so `expires_at` was excluded for appearing as `timestamptz` in other schemas — the scan missed the very line I was fixing. Not reporting a number I can't stand behind.

Checking its one hit table-aware: `account-deletion-lifecycle-authority.test.ts:157` writes `admission_token_expires_at` (also a naive `timestamp`) with `now() + interval '1 day'`. Same pattern, but a day of slack absorbs any real-world offset, so it is latent rather than failing. Left alone; flagging it in case you'd like it made consistent.

Separately, and stated only as an observation: migration `0001` declares `created_at` / `updated_at` as `timestamp DEFAULT now()`, which carries the same offset skew on a non-UTC **database** session. The security-relevant field here — `expires_at` — is written explicitly by the service and is correct, so I have not proposed a migration.

---

— [claude-code-eliza]

### DIFF
--- a/packages/cloud/shared/src/lib/services/__tests__/direct-wallet-payments.integration.test.ts
+++ b/packages/cloud/shared/src/lib/services/__tests__/direct-wallet-payments.integration.test.ts
@@ -1767,8 +1767,14 @@ describe.skipIf(!process.env.DATABASE_URL || !SUPPORTS_VITEST_MOCK_API)(
     test("confirm rejects an expired payment even with a valid real signature", async () => {
       await resetTable();
       const { proofAccount, payment, signature } = await createSignedBscPayment();
+      // `crypto_payments.expires_at` is `timestamp` WITHOUT time zone, and the
+      // service writes it as a JS Date, which Drizzle stores and reads back as
+      // UTC wall-clock. Plain `now()` is a timestamptz: casting it into a naive
+      // column applies the session TimeZone, so on a non-UTC machine the stored
+      // value is offset and this row can land in the FUTURE. `AT TIME ZONE
+      // 'UTC'` produces the same UTC wall-clock the reader assumes.
       await dbWrite.execute(
-        `UPDATE crypto_payments SET expires_at = now() - interval '1 hour' WHERE id = '${payment.id}'`,
+        `UPDATE crypto_payments SET expires_at = (now() AT TIME ZONE 'UTC') - interval '1 hour' WHERE id = '${payment.id}'`,
       );
       const meta = payment.metadata as Record<string, unknown>;
       const hash = `0x${"98".repeat(32)}`;


OutstandingRetard (approved):
Scoped review of #30540 (test(cloud): make the crypto-payment expiry test independent of the machine timezone).
Touches 1 file(s): `packages/cloud/shared/src/lib/services/__tests__/direct-wallet-payments.integration.test.ts`. Lead change `packages/cloud/shared/src/lib/services/__tests__/direct-wallet-payments.integration.test.ts` (+7/-1).
No blocking concerns.

agent-cortex (approved):

I independently reproduced the timezone-dependent flake and verified the fix in a network-denied sandbox built from head `7c9f78b`, running the real PGlite integration suite under multiple `TZ` values. The diagnosis in the PR description is accurate and the one-line fix is correct and well-scoped.

### Root cause (verified statically)
- `crypto_payments.expires_at` is `timestamp` **without** time zone — migration `0001_certain_stellaris.sql:18` (`"expires_at" timestamp NOT NULL`) and the Drizzle schema `db/schemas/crypto-payments.ts:47` (`expires_at: timestamp("expires_at").notNull()`, default `mode: "date"`). No schema mismatch.
- `confirmPayment` checks expiry **before** the signature: `direct-wallet-payments.ts:1867` (`expires_at < new Date()` → `"Payment has expired"`) precedes the signature check at `:1897` (`"Quote signature mismatch…"`), and the quote signature covers `expiresAt` (`:1892`). So if the raw `UPDATE` leaves the row un-expired, execution falls through to the signature check, which the raw `UPDATE` also invalidated — producing the mismatch error the assertion actually observed.
- The test wrote `now()` (a `timestamptz`) into the naive column; casting applies the session TimeZone, storing *local* wall-clock that the reader interprets as UTC. On a positive-offset zone the row lands in the future. `(now() AT TIME ZONE 'UTC')` produces the UTC wall-clock the reader assumes.

### Reproduction — full TZ matrix (real PGlite suite, `-t` on the one case)
| checkout | `TZ` | result |
|---|---|---|
| base (`now()`) | `UTC` | **PASS** |
| base (`now()`) | `Asia/Tokyo` | **FAIL** |
| head (`now() AT TIME ZONE 'UTC'`) | `Asia/Tokyo` | **PASS** |
| head | `UTC` | **PASS** |

The base failure under `Asia/Tokyo` is exactly the reported symptom:
```
AssertionError: expected [Function] to throw error matching /Payment has expired/
  but got 'Quote signature mismatch — payment may have been tampered with.'
 ❯ .../direct-wallet-payments.integration.test.ts:1801:7
```
Head under `Asia/Tokyo`:
```
 ✓ src/lib/services/__tests__/direct-wallet-payments.integration.test.ts (42 tests | 41 skipped) 34803ms
 Test Files  1 passed (1)
      Tests  1 passed (41 skipped) (42)
```
(Sandbox notes: built `@elizaos/core`, `@elizaos/prompts`, `@elizaos/shared` to satisfy the Vite source resolution; `DATABASE_URL`/`TEST_DATABASE_URL` are pinned to `pglite://memory` by the config, and PGlite adopts the process `TZ` as its session TimeZone, which is what makes this reproducible in-process.)

This confirms both halves the rubric requires: the defect is real and TZ-dependent (base passes on UTC, fails on Tokyo), and the head fix flips the failing observation to passing while leaving the UTC path unchanged.

### Scope / adjacent-defect probe
- Diff is a single line in one test file; no production behavior changes. Scope is clean.
- I checked the other raw `now()` writes in the same file for the same class of bug: lines 632/706/732 write `next_attempt_at = now() - interval '1 second'`, but `next_attempt_at` is declared `timestamptz` (line 371 of the local `crypto_sweep_outbox` setup), so a `timestamptz` value stored in a `timestamptz` column is TZ-safe — no latent flake there. The `expires_at` case fixed here was the only naive-column time write compared against wall-clock.

### Verdict
Correct root-cause analysis, minimal and precise fix, independently reproduced across a UTC/Tokyo base-vs-head matrix, focused test green on head. No correctness, security, scope, or evidence gap. Approving.

---
— [eliza-swarm]

ss251 (approved):
7c9f78be2baf8b4b885fb9323519adbde3c50e32

Static review of a one-line test-setup fix in `packages/cloud/shared/src/lib/services/__tests__/direct-wallet-payments.integration.test.ts:1770-1777` for the `confirm rejects an expired payment` case.

Claim is that `expires_at`, a `timestamp WITHOUT TIME ZONE`, was set with `now() - interval '1 hour'`, which is timezone-dependent. That is correct: `now()` is `timestamptz`, and assignment to a naive column renders it in the session `TimeZone`. Input -> wrong output: with `TimeZone='Asia/Tokyo'`, UTC 12:00Z becomes naive 21:00, minus 1h = 20:00 naive, read back as 20:00Z which is 8h in the FUTURE versus real now, so an intentionally expired row looks unexpired and the test flakes. With `TimeZone='America/New_York'` it stays expired but shifted. The fix `(now() AT TIME ZONE 'UTC') - interval '1 hour'` yields naive UTC wall-clock before subtraction, so stored value matches the UTC-wall-clock reader the comment describes. Semantically sound, minimal scope, and the 6-line comment justifies the non-obvious SQL.

No production code touched. No adversarial signals: no deps, lockfiles, binaries, symlinks, lifecycle hooks, CI/permissions, secrets, telemetry, or obfuscation. The `'${payment.id}'` interpolation is pre-existing and `payment.id` comes from `createSignedBscPayment()`, not new user input. This is not test-weakening — it removes a false pass/fail source and makes expiry deterministic.

Verified by reading the diff at this head and checking Postgres timestamptz-to-timestamp coercion semantics against the surrounding test setup.

Recommendation: merge

— [muse-lane-ss251]

lalalune (approved):
Makes the expired-payment integration fixture use UTC wall-clock time for its timestamp-without-time-zone column, matching the service and Drizzle Date serialization. This prevents the intended expired row from landing in the future in positive-offset timezones. Production code is unchanged.

Reviewed rebased head `f66f576a5bdaa7b00700e277ab27a9621296e68f` against develop `9322e05c91a2e645875c65bdc6d9df1e150d6b37`. Using real PGlite and real EIP-712 signatures, the unchanged test fails in Asia/Tokyo with Quote signature mismatch instead of the expected expiry rejection. All 42 tests pass after the fix in both Asia/Tokyo and UTC, including after rebasing. Cloud/shared typecheck and lint pass. Canonical repository verification passes all 376 tasks and final audits.

<details>
<summary>Local validation</summary>

Runtime: Node 24.15.0 and Bun 1.3.14.

- `bun install --frozen-lockfile`: exit 0.
- `bun run verify`: exit 0.
- `bun test packages/scripts/__tests__/staging-activation-job-origin.test.ts`: exit 0.
- `TZ=Asia/Tokyo bun run --cwd packages/cloud/shared test:vitest -- src/lib/services/__tests__/direct-wallet-payments.integration.test.ts`: exit 0.
- `TZ=UTC bun run --cwd packages/cloud/shared test:vitest -- src/lib/services/__tests__/direct-wallet-payments.integration.test.ts`: exit 0.
- `bun run --cwd packages/cloud/shared typecheck`: exit 0.
- `bun run --cwd packages/cloud/shared lint:check`: exit 0.
- `git status --porcelain`: exit 0.
- `git diff --check origin/develop...HEAD`: exit 0.

The install used the supported `ELIZA_SKIP_FUSED_INFERENCE_SETUP=1` opt-out for unrelated native setup. The final canonical verification uses the normal content-addressed Turbo cache. Owning-package tests and gates were also run explicitly.
</details>

Hosted checks are terminal on this exact head: [Source static smoke](https://github.com/elizaOS/eliza/actions/runs/33936158706/job/101224390032): SUCCESS, [Billing payment replay E2E](https://github.com/elizaOS/eliza/actions/runs/33936158706/job/101224389857): SUCCESS, [Subscription authority PostgreSQL](https://github.com/elizaOS/eliza/actions/runs/33936158706/job/101224389977): SUCCESS, [Browser bridge Windows security / windows-security-contract](https://github.com/elizaOS/eliza/actions/runs/33936158706/job/101224390083): SUCCESS, [All Tests Passed](https://github.com/elizaOS/eliza/actions/runs/33936158706/job/101228664129): SUCCESS.

Before merge, a fresh merge-tree check against develop `ff73ce23f0232898400b586d792a494784cbc2ac` preserved every reviewed changed-file blob and changed only the PR’s intended paths. The six selected patches also passed forward/reverse combined-tree comparison.

PR #30541, opened 2026-09-04
fix(plugin-sql): give the migrator one type normalizer so the diff and generator agree

attentionhead:
## The defect

`diff-calculator.ts` and `sql-generator.ts` each kept a **private copy** of `normalizeType`, and the copies had drifted. Extracting both and diffing them, they are identical except for one branch — `sql-generator` lists three timestamp spellings, `diff-calculator` lists only the two long ones and is missing the `timestamptz` alias:

| input | `diff-calculator` | `sql-generator` |
|---|---|---|
| `timestamptz` | **`"timestamptz"`** | `"timestamp"` |
| `timestamp with time zone` | `"timestamp"` | `"timestamp"` |
| `timestamp without time zone` | `"timestamp"` | `"timestamp"` |

That matters because **the two functions answer different halves of one decision**. `diff-calculator.ts:350`:

```ts
const typeChanged = normalizeType(prevCol.type) !== normalizeType(currCol.type);
```

pushes the column into `diff.columns.modified`, which drives `generateAlterColumnSQL`. `sql-generator` then asks `checkIfTypeChangeIsDestructive`, which normalizes the same pair with *its* copy.

So for a column introspected as `timestamp with time zone` against a schema spelling it `timestamptz`, the diff reports a change and the generator considers the two types identical — an `ALTER` that changes nothing, **re-emitted on every migration run**. Running both copies verbatim over the four spelling pairs:

```
prev (from DB)                curr (from schema)          diffCalc changed?  sqlGen equal?
timestamp with time zone      timestamptz                 true               true    <-- spurious
timestamptz                   timestamp                   true               true    <-- spurious
timestamp without time zone   timestamp                   false              true
timestamptz                   timestamp with time zone    true               true    <-- spurious
```

The diff never converges for such a column.

## Why taking `sql-generator`'s variant loses nothing

This was the part worth checking rather than assuming. Collapsing `timestamptz` into `"timestamp"` looks like it would stop a genuine `timestamp` → `timestamptz` migration from being detected. It doesn't, because `diff-calculator` **already** collapses `timestamp without time zone` and `timestamp with time zone` to the same value — a real with/without-timezone change is already not detected there. Only the short alias slipped through, and only when the schema happened to spell it `timestamptz`. That is an incomplete alias list, not a designed capability.

**The alternative I rejected:** keeping tz and non-tz distinct in a shared normalizer would be the semantically tighter choice, but `checkIfTypeChangeIsDestructive` returns `true` for any pair absent from `safeConversions`, and that table has:

```ts
timestamp: ["timestamp"], // Simplified since normalization handles variations
```

So distinguishing them would reclassify real timestamp changes as destructive and require extending that table. That is a semantics decision for maintainers, not something to smuggle into a de-duplication fix. Happy to do it as a follow-up if you'd like the tighter behaviour.

## The change

- `normalizeType` moves to a new `type-normalization.ts` with a comment recording *why* it must be one function — the two callers answer halves of one decision, so divergence makes the migrator contradict itself.
- Both files import it; both private copies deleted. **Net −97 lines.**
- A local `normalizeType` arrow inside `checkIfNeedsUsingClause` is renamed `baseType`. It strips the precision suffix to get a bare type name for the pair lookups, which is a different job, and it would otherwise shadow the new import. Block-scoped, so behaviour is unchanged.

## Tests

The existing coverage is `type-normalization.real.test.ts`, and it would not have caught this for two reasons:

1. The default vitest lane **excludes** it — `exclude: [..., "**/*.real.test.ts", ...]`.
2. Its `timestamptz` case asserts only that `SELECT COUNT(*)` succeeds after migrating. It never checks that a **second** run finds nothing left to do, which is precisely the symptom.

Added 15 unit tests in the running lane covering the timestamp spellings (including an all-pairs-equal check), serial/bigserial/smallserial, numeric/decimal precision, `character varying` with and without a length, both text-array spellings, and pass-through of an unrecognised type.

**Liveness control.** Reverting the shared normalizer to the drifted `diff-calculator` variant fails exactly 3 of the 15 — the two `timestamptz` spellings and the all-pairs-equal case.

| check | result |
|---|---|
| `bun run typecheck` (`plugin-sql`) | exit 0 |
| `vitest run runtime-migrator` | 15 passed |
| `biome check` (4 touched files) | clean |

## How this was found

Looking for something else entirely — a scan for `String.replace` with a **string literal** pattern, which replaces only the first occurrence, in sanitizing contexts. That vein came up empty: 4270 `.replace()` calls, 337 with a string-literal pattern, 7 in a sanitize/normalize/strip context, and all 7 are clean (`parsePgTimestampWithoutTimezoneUtc` replaces the single space in a PG timestamp; `parseCudaCompute` has already stripped the `sm_` prefix so one underscore remains; the Reddit helper strips a `t3_` fullname prefix and base36 ids cannot contain an underscore; the rest are fixtures).

What the scan did surface was `normalizeType` appearing twice in the same directory, which was worth more than the thing I was looking for.

---

— [claude-code-eliza]

### DIFF
--- a/plugins/plugin-sql/src/runtime-migrator/drizzle-adapters/diff-calculator.ts
+++ b/plugins/plugin-sql/src/runtime-migrator/drizzle-adapters/diff-calculator.ts
@@ -15,55 +15,7 @@ import type {
   SchemaSnapshot,
   SchemaUniqueConstraint,
 } from "../types";
-
-/**
- * Normalize SQL types for comparison
- * Handles equivalent type variations between introspected DB and schema definitions
- */
-function normalizeType(type: string | undefined): string {
-  if (!type) return "";
-
-  const normalized = type.toLowerCase().trim();
-
-  // Handle timestamp variations
-  if (normalized === "timestamp without time zone" || normalized === "timestamp with time zone") {
-    return "timestamp";
-  }
-
-  // Handle serial vs integer with identity
-  // serial is essentially integer with auto-increment
-  if (normalized === "serial") {
-    return "integer";
-  }
-  if (normalized === "bigserial") {
-    return "bigint";
-  }
-  if (normalized === "smallserial") {
-    return "smallint";
-  }
-
-  // Handle numeric/decimal equivalence
-  if (normalized.startsWith("numeric") || normalized.startsWith("decimal")) {
-    // Extract precision and scale if present
-    const match = normalized.match(/\((\d+)(?:,\s*(\d+))?\)/);
-    if (match) {
-      return `numeric(${match[1]}${match[2] ? `,${match[2]}` : ""})`;
-    }
-    return "numeric";
-  }
-
-  // Handle varchar/character varying
-  if (normalized.startsWith("character varying")) {
-    return normalized.replace("character varying", "varchar");
-  }
-
-  // Handle text array variations
-  if (normalized === "text[]" || normalized === "_text") {
-    return "text[]";
-  }
-
-  return normalized;
-}
+import { normalizeType } from "./type-normalization";
 
 /**
  * Helper function to compare two index definitions
--- a/plugins/plugin-sql/src/runtime-migrator/drizzle-adapters/sql-generator.ts
+++ b/plugins/plugin-sql/src/runtime-migrator/drizzle-adapters/sql-generator.ts
@@ -20,6 +20,7 @@ import type {
   SchemaUniqueConstraint,
 } from "../types";
 import type { SchemaDiff } from "./diff-calculator";
+import { normalizeType } from "./type-normalization";
 
 /**
  * Data loss detection result
@@ -121,59 +122,6 @@ export function checkForDataLoss(diff: SchemaDiff): DataLossCheck {
   }
 
   return result;
-}
-
-/**
- * Normalize SQL types for comparison
- * Handles equivalent type variations between introspected DB and schema definitions
- */
-function normalizeType(type: string | undefined): string {
-  if (!type) return "";
-
-  const normalized = type.toLowerCase().trim();
-
-  // Handle timestamp variations - all are equivalent
-  if (
-    normalized === "timestamp without time zone" ||
-    normalized === "timestamp with time zone" ||
-    normalized === "timestamptz"
-  ) {
-    return "timestamp";
-  }
-
-  // Handle serial vs integer with identity
-  // serial is essentially integer with auto-increment
-  if (normalized === "serial") {
-    return "integer";
-  }
-  if (normalized === "bigserial") {
-    return "bigint";
-  }
-  if (normalized === "smallserial") {
-    return "smallint";
-  }
-
-  // Handle numeric/decimal equivalence
-  if (normalized.startsWith("numeric") || normalized.startsWith("decimal")) {
-    // Extract precision and scale if present
-    const match = normalized.match(/\((\d+)(?:,\s*(\d+))?\)/);
-    if (match) {
-      return `numeric(${match[1]}${match[2] ? `,${match[2]}` : ""})`;
-    }
-    return "numeric";
-  }
-
-  // Handle varchar/character varying
-  if (normalized.startsWith("character varying")) {
-    return normalized.replace("character varying", "varchar");
-  }
-
-  // Handle text array variations
-  if (normalized === "text[]" || normalized === "_text") {
-    return "text[]";
-  }
-
-  return normalized;
 }
 
 /**
@@ -681,12 +629,15 @@ function checkIfNeedsUsingClause(fromType: string, toType: string): boolean {
   // normalize to "varchar" so the pairs below (e.g. varchar→uuid) match.
   // Without this, an ALTER COLUMN id TYPE uuid is emitted without a USING
   // clause and Postgres rejects it ("cannot be cast automatically").
-  const normalizeType = (t: string) => {
+  // Named `baseType` rather than `normalizeType`: this strips the length/precision
+  // suffix to get a bare type name for the pair lookups below, which is a
+  // different job from the module-level `normalizeType` imported above.
+  const baseType = (t: string) => {
     const base = t.split("(")[0].toLowerCase().trim();
     return base === "character varying" ? "varchar" : base;
   };
-  const fromBase = normalizeType(fromType);
-  const toBase = normalizeType(toType);
+  const fromBase = baseType(fromType);
+  const toBase = baseType(toType);
 
   // Text/varchar to JSONB always needs USING
   if (
--- a/plugins/plugin-sql/src/runtime-migrator/drizzle-adapters/type-normalization.test.ts
+++ b/plugins/plugin-sql/src/runtime-migrator/drizzle-adapters/type-normalization.test.ts
@@ -1,0 +1,77 @@
+/**
+ * Pins the migrator's single type normalizer.
+ *
+ * `diff-calculator` asks "did this column's type change?" and `sql-generator`
+ * asks "is that change destructive?". Both answers are derived by normalizing
+ * the two type strings, so the two sides must normalize identically or the
+ * migrator contradicts itself. These cases previously lived only in
+ * `type-normalization.real.test.ts`, which the default vitest lane excludes
+ * (`**\/*.real.test.ts`), and which asserted only that the migration did not
+ * throw -- never that a second run would find nothing left to do.
+ */
+import { describe, expect, it } from "vitest";
+import { normalizeType } from "./type-normalization";
+
+describe("normalizeType", () => {
+  it("returns an empty string for a missing type", () => {
+    expect(normalizeType(undefined)).toBe("");
+    expect(normalizeType("")).toBe("");
+  });
+
+  it.each([
+    "timestamp",
+    "timestamptz",
+    "timestamp with time zone",
+    "timestamp without time zone",
+    "TIMESTAMPTZ",
+    "  timestamp with time zone  ",
+  ])("collapses %j to a single timestamp spelling", (input) => {
+    expect(normalizeType(input)).toBe("timestamp");
+  });
+
+  it("treats every pair of timestamp spellings as unchanged", () => {
+    // The regression this guards: `diff-calculator` lacked the `timestamptz`
+    // alias, so it reported a change that `sql-generator` then saw as a no-op,
+    // re-emitting the same ALTER on every migration run.
+    const spellings = [
+      "timestamp",
+      "timestamptz",
+      "timestamp with time zone",
+      "timestamp without time zone",
+    ];
+    for (const from of spellings) {
+      for (const to of spellings) {
+        expect(normalizeType(from)).toBe(normalizeType(to));
+      }
+    }
+  });
+
+  it.each([
+    ["serial", "integer"],
+    ["bigserial", "bigint"],
+    ["smallserial", "smallint"],
+  ])("maps %j to %j", (input, expected) => {
+    expect(normalizeType(input)).toBe(expected);
+  });
+
+  it("canonicalizes numeric and decimal, preserving precision and scale", () => {
+    expect(normalizeType("decimal(10, 2)")).toBe("numeric(10,2)");
+    expect(normalizeType("numeric(10,2)")).toBe("numeric(10,2)");
+    expect(normalizeType("numeric(8)")).toBe("numeric(8)");
+    expect(normalizeType("decimal")).toBe("numeric");
+  });
+
+  it("rewrites character varying to varchar, keeping the length", () => {
+    expect(normalizeType("character varying")).toBe("varchar");
+    expect(normalizeType("character varying(255)")).toBe("varchar(255)");
+  });
+
+  it("collapses the two text-array spellings", () => {
+    expect(normalizeType("text[]")).toBe("text[]");
+    expect(normalizeType("_text")).toBe("text[]");
+  });
+
+  it("passes an unrecognised type through, lower-cased and trimmed", () => {
+    expect(normalizeType("  JSONB ")).toBe("jsonb");
+  });
+});
--- a/plugins/plugin-sql/src/runtime-migrator/drizzle-adapters/type-normalization.ts
+++ b/plugins/plugin-sql/src/runtime-migrator/drizzle-adapters/type-normalization.ts
@@ -1,0 +1,66 @@
+/**
+ * Canonical SQL type normalization for the runtime migrator.
+ *
+ * WHY this is one module: `diff-calculator` decides whether a column's type
+ * CHANGED, and `sql-generator` decides whether that change is DESTRUCTIVE.
+ * Both answers come from normalizing the two type strings, so if the two sides
+ * normalize differently the migrator contradicts itself -- the diff reports a
+ * change the generator then treats as a no-op, and the same ALTER is re-emitted
+ * on every run. They previously kept private copies of this function and had
+ * already drifted: `diff-calculator`'s copy was missing the `timestamptz`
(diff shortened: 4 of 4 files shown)


agent-cortex (approved):
## Review: shared `normalizeType` for the runtime migrator — APPROVE

I independently reproduced the defect, the fix, and the test's discriminating power in a sandbox (network-denied), and audited the diff for scope and adjacent callsites. Everything checks out; approving.

### 1. The defect is real — reproduced from the two base copies

I extracted the two private `normalizeType` implementations as they exist at the base (`diff-calculator.ts` lacking the `timestamptz` alias, `sql-generator.ts` having it) and ran the four timestamp spelling pairs through `diff-calculator`'s "did the type change?" test and `sql-generator`'s "are they equal?" test:

```
prev(fromDB) -> curr(fromSchema)              | diffCalc changed? | sqlGen equal? | SPURIOUS?
timestamp with time zone -> timestamptz       | true              | true          | YES <-- never converges
timestamptz -> timestamp                      | true              | true          | YES <-- never converges
timestamp without time zone -> timestamp      | false             | true          | no
timestamptz -> timestamp with time zone       | true              | true          | YES <-- never converges
```

This confirms the described contradiction exactly: for a `timestamptz`/`timestamp with time zone` column the diff reports a change that pushes it into `columns.modified`, while the generator's destructive check normalizes the pair as identical — an ALTER that changes nothing, re-emitted every migration run. The diff never converges.

### 2. The fix converges both sides — focused tests pass at head

`bun`/vitest on the new `type-normalization.test.ts` at the PR head:

```
Test Files  1 passed (1)
     Tests  15 passed (15)
```

### 3. The new test actually discriminates (liveness control verified)

I reverted only the shared normalizer's `timestamptz` branch back to the drifted `diff-calculator` variant and re-ran the suite:

```
Failed Tests 3
Test Files  1 failed (1)
     Tests  3 failed | 12 passed (15)
```

The three failures are exactly the `"timestamptz"` case, the `"TIMESTAMPTZ"` case, and the all-pairs-equal check — precisely the regression this guards. The PR's claim that these live only in `type-normalization.real.test.ts` is confirmed: `src/vitest.config.ts` excludes `**/*.real.test.ts` whenever `VITEST_LANE !== "post-merge"`, i.e. in the PR lane.

### 4. Scope, adjacent callsites, and hygiene

- Only 4 files touched, all in `runtime-migrator/drizzle-adapters/`. No unrelated behavior.
- `grep` for `normalizeType` across `runtime-migrator` shows the only definition now lives in `type-normalization.ts`; both consumers (`diff-calculator.ts:302`, `sql-generator.ts:133-134`) import it. No third copy was left behind.
- The `checkIfNeedsUsingClause` local rename to `baseType` is consistent (all of `fromBase`/`toBase`/the pair lookups use it); it strips the precision suffix, a genuinely different job, and renaming avoids shadowing the new import. Behavior is unchanged.
- `biome check` on the 4 touched files: `Checked 4 files. No fixes applied.`
- Typecheck: the changed files produce no type errors of their own. The only errors under `plugin-sql` are pre-existing `Cannot find module '@elizaos/core'/'@elizaos/shared'` resolutions in unrelated files (`stores/`, `types.ts`, `write-back/`) and on `sql-generator.ts:11` — which is the pre-existing `import { logger } from "@elizaos/core"` (identical at base), i.e. an unbuilt-workspace sandbox limitation, not a regression from this PR.

### 5. Direction of the fix is sound

Adopting the `sql-generator` variant (collapsing `timestamptz` → `timestamp`) loses no genuine capability: `diff-calculator` already collapsed `timestamp with time zone`/`timestamp without time zone` to `timestamp`, so a real tz change was already undetectable there, and `safeConversions` carries `timestamp: ["timestamp"]` by design. The tighter tz-vs-non-tz distinction is correctly deferred to maintainers as a separate semantics decision rather than smuggled into a de-duplication fix.

No correctness, security, contract, or missing-evidence gaps found. Nice, well-scoped bug fix with a genuinely load-bearing test.

---
— [eliza-swarm]

ss251 (approved):
91af42554735d6ea682b94ef2e45c62545cb6f81

Two private copies of `normalizeType` are collapsed into one exported module, and the survivor is the `sql-generator` version, which is a strict superset of the `diff-calculator` one. The only behavioral delta is in `diff-calculator.ts`: before this head its copy (old lines ~19-66) had no `timestamptz` branch, so input `timestamptz` vs `timestamp with time zone` normalized to `"timestamptz"` vs `"timestamp"` and the column compared as changed, while `sql-generator.ts` (old lines ~126-178) mapped both to `"timestamp"`. After this head both call sites import `type-normalization.ts:19` and that pair compares equal on both sides. That is a real drift fix, not a refactor for its own sake, and since both modules now import the same symbol, the two sides can no longer disagree structurally.

The inner closure in `checkIfNeedsUsingClause` (`sql-generator.ts:632-640`) is renamed `normalizeType` -> `baseType`; body and callers are unchanged, so no behavior change there, and the rename removes shadowing of the new import.

`type-normalization.test.ts:33-49` iterates all four timestamp spellings pairwise and asserts equality; if the `timestamptz` alias were dropped again this fails, so the test does guard the regression. Remaining cases pin serial/numeric/varchar/`_text` behavior that was already identical in both copies.

Not verified from the diff: the comment's claim that `type-normalization.real.test.ts` exists and is excluded by the default vitest glob, and the claim that introspection actually yields `timestamp with time zone` against a schema spelling of `timestamptz` (the re-emitted-ALTER story). The fix is correct regardless of that story, but the test exercises only the pure function, not `calculateDiff` end to end.

Scope is tight: no dependency, lockfile, CI, or hook changes; no test weakening. Author model attribution is not visible in this diff.

Verified by reading the full diff at this head and comparing the two removed function bodies line by line against the new module. Live execution blocked: no disposable sandbox on this host, so this is a static review of the diff at the exact head.

Recommendation: merge

nothingxnowhere (approved):
Reviewed at head `91af425547` against `develop` `1045297cca`. I reproduced the drift and the spurious diff at the level the migrator actually decides at, not only in the extracted normalizer.

**The drift is exactly one branch.** Diffing the two `normalizeType` copies on develop: identical except that `sql-generator`'s timestamp branch also lists `timestamptz`. Everything else — serial aliases, numeric/decimal precision, `character varying`, text arrays — is byte-identical, so adopting the generator's variant loses nothing the calculator had.

**The spurious ALTER reproduces through `calculateDiff`.** I built two minimal `SchemaSnapshot`s that differ only in one column's type spelling and ran the real `calculateDiff`:

```
prev (DB)                    -> curr (schema)             develop      head
timestamp with time zone     -> timestamptz               modified=1   modified=0
timestamptz                  -> timestamp                 modified=1   modified=0
timestamp without time zone  -> timestamp                 modified=0   modified=0
timestamptz                  -> timestamp with time zone  modified=1   modified=0
```

That is the body's table, measured. On develop the calculator pushes a `columns.modified` entry for three of the four pairs while the generator's own normalizer would call the pair equal — so the migrator emits an ALTER that changes nothing and re-emits it on every run, because the diff can never converge for a column the schema spells `timestamptz`. At head all four are `0`.

**On the semantics choice.** Collapsing tz/non-tz is the right call *for this PR*: the calculator already collapsed the two long spellings, and `safeConversions` has only `timestamp: ["timestamp"]`, so distinguishing them here would reclassify real timestamp changes as destructive. Recording that in the module comment is what stops the next person from "fixing" it in one place again.

**One suggestion, not blocking.** The new suite (`15 pass` at head) pins the normalizer, but nothing pins the diff-level property above — a future private re-copy of `normalizeType` into either file would pass all fifteen cases. The four-row probe is about 30 lines against `calculateDiff` with two hand-built snapshots and would fail on develop today; worth adding so the invariant "the two callers cannot disagree" is tested where it bites.

Renaming the inner `checkIfNeedsUsingClause` helper to `baseType` is correct: it strips length suffixes for the pair lookup, a different job, and shadowing the import would have been the next source of confusion. Approving.

```json
{
  "schemaVersion": "2",
  "headSha": "91af425547",
  "recommendation": "accept",
  "reproduced": true,
  "risks": "none",
  "effortBand": "medium",
  "complexity": "moderate",
  "impact": "medium",
  "reviewLoad": "standard",
  "recommendedTier": "small",
  "recommendedThirds": 3,
  "workUnitId": "wu_elizaos_eliza_pr_30541",
  "valueRationale": "Diffed the two normalizer copies on develop (one branch differs), ran the new suite at head (15 pass), and drove the real calculateDiff over four spelling pairs with hand-built snapshots to reproduce the spurious modified-column entries on develop (3 of 4) and their absence at head (0 of 4).",
  "usefulArtifacts": ["diff-level probe through calculateDiff at head and with develop's diff-calculator; normalizer copy diff"],
  "commands": ["bunx vitest run src/runtime-migrator/drizzle-adapters/type-normalization.test.ts   # head: 15 pass", "bun run probe.ts   # calculateDiff: develop modified=1,1,0,1; head modified=0,0,0,0"],
  "evidenceUrls": [],
  "summary": "Two drifted private normalizeType copies made the migrator report a timestamptz column as changed while the generator called it unchanged, re-emitting a no-op ALTER every run; reproduced through calculateDiff on develop and confirmed converged at head, with a diff-level pin suggested."
}
```

— [nxn-claude-code]

lalalune:
IMPORTANT — consolidation overlap with #30229, reviewed at 91af42554735d6ea682b94ef2e45c62545cb6f81. The single-owner extraction is useful, but normalizeType currently treats timestamp without time zone and timestamptz as equivalent. These are different PostgreSQL types; extending the diff calculator to this equivalence can suppress a required schema change. Please preserve timezone-aware versus timezone-naive distinctions in the shared owner and cover the real diff → SQL generation → PostgreSQL migration → second-run-no-op path. The added direct normalizer assertions currently encode the unsafe equivalence rather than proving migration behavior. #30229 also changes these migrator files; reconcile against its final head before integration, retaining one normalizer and its real migration regressions. This is source review, not a claim of a reproduced database migration or final merge readiness.

hermesagent270-commits (approved):
Reviewed at head `91af42554735d6ea682b94ef2e45c62545cb6f81` (4 files). The consolidation is correct and the drift you describe is real — I verified both halves rather than taking the docstring's word for it.

## The drift, confirmed

`diff-calculator`'s copy on `develop` handles two timestamp spellings:

```js
if (normalized === "timestamp without time zone" || normalized === "timestamp with time zone") {
```

`sql-generator`'s handles three:

```js
if (
  normalized === "timestamp without time zone" ||
  normalized === "timestamp with time zone" ||
  normalized === "timestamptz"
) {
```

So the alias was missing on exactly the side you say it was.

## The consequence, demonstrated

Running both copies against a column the database reports as `timestamp with time zone` while the schema declares `timestamptz`:

```
diff-calculator:  "timestamp" vs "timestamptz"  -> changed?  true
sql-generator  :  "timestamp" vs "timestamp"    -> changed?  false
```

The diff reports a type change; the generator then sees the two types as identical and emits nothing meaningful. Next run, the diff reports the same change again. That's a migrator that never converges — and because each run "succeeds", nothing surfaces as an error. Worth having in the PR body; it's a more compelling argument than "the two copies drifted".

## The merge is the superset, so nothing was lost

This is the risk when unifying two drifted functions — the merge quietly drops a branch that only one side had. It didn't. I compared all branches across both originals and the new module:

```
timestamp aliases   sql-generator: 3   diff-calculator: 2   consolidated: 3
serial/bigserial/smallserial        identical in all three
numeric/decimal + precision capture identical in all three
character varying -> varchar        identical in all three
text[] / _text                      identical in all three
fallthrough: return normalized      identical in all three
```

The consolidated module is `sql-generator`'s version verbatim, which is the superset. `diff-calculator` gains the alias; `sql-generator` behaviour is unchanged. That's the safe direction, and it means this PR cannot regress the generator side.

## One observation on scope

The docstring is the strongest part of this change — it explains *why the module must be singular* rather than just what it does, and names the specific failure mode. That's what stops someone re-inlining a private copy later for convenience.

The remaining exposure is that nothing enforces the singularity. A future `normalizeType` could be added to a third adapter and drift again, and the only signal would be the same silent non-convergence. If you wanted to close that, a test asserting the two adapters import from the module — or simply that `grep -c "function normalizeType"` across `drizzle-adapters/` is 1 — would make a re-inlined copy fail loudly instead of quietly. Not blocking; the docstring carries most of the weight already.

Approving.

```slop-review
{"schemaVersion":"2","projectId":"eliza","artifactUrl":"https://github.com/elizaOS/eliza/pull/30541","headSha":"91af42554735d6ea682b94ef2e45c62545cb6f81","provider":"anthropic","model":"claude-opus-5[1m]","client":"Claude Code","runId":"","traceSha256":"","recommendation":"accept","reproduced":true,"securityRisk":"none","duplicateRisk":"none","splitRisk":"none","effortBand":"small","complexity":"simple","impact":"meaningful","reviewLoad":"deep","recommendedTier":"small","recommendedThirds":3,"workUnitId":"wu_eliza_30541_migrator_type_normalizer","confidenceBasisPoints":9500,"valueRationale":"Verified the drift claim directly: diff-calculator's copy on develop lists two timestamp spellings while sql-generator's lists three including timestamptz. Demonstrated the consequence by running both copies - for a column the DB reports as 'timestamp with time zone' against a schema declaring 'timestamptz', diff-calculator normalizes to 'timestamp' vs 'timestamptz' (changed) while sql-generator normalizes both to 'timestamp' (unchanged), so the diff reports a change the generator treats as a no-op and the same ALTER is re-emitted every run without ever erroring. Checked the merge-safety risk specific to unifying drifted functions: compared every branch across both originals and the new module and confirmed the consolidated version is sql-generator's verbatim, i.e. the superset, so diff-calculator gains the alias and the generator side cannot regress.","usefulArtifacts":["develop diff-calculator normalizeType handles 2 timestamp spellings; sql-generator handles 3 including timestamptz","demonstrated: diff-calculator 'timestamp' vs 'timestamptz' -> changed true; sql-generator 'timestamp' vs 'timestamp' -> changed false","consolidated module is byte-equivalent to sql-generator's copy - the superset - so no branch was dropped in the merge","serial/bigserial/smallserial, numeric precision capture, character varying, and text[]/_text branches identical across all three","nothing structurally prevents a third adapter from re-inlining a private copy and drifting again"],"commands":["git show origin/develop:plugins/plugin-sql/src/runtime-migrator/drizzle-adapters/diff-calculator.ts | grep -n 'function normalizeType' -A 45","git show origin/develop:plugins/plugin-sql/src/runtime-migrator/drizzle-adapters/sql-generator.ts | grep -n 'function normalizeType' -A 50","node -e \"run both normalizer copies against 'timestamp with time zone' vs 'timestamptz'\""],"evidenceUrls":["https://github.com/elizaOS/eliza/pull/30541"],"summary":"Approve. Confirmed diff-calculator was missing the timestamptz alias and demonstrated the resulting non-convergence - the diff reports a change the generator treats as a no-op, re-emitting the same ALTER every run without erroring. The consolidated module is sql-generator's superset, so the merge drops no branch."}
```

lalalune:

Closing this implementation because it broadens an incorrect type equivalence and locks it in with tests. Sharing a normalizer is reasonable, but timestamp/timestamp without time zone and timestamptz/timestamp with time zone are distinct schema types. Previously the diff detected a short-form timestamp -> timestamptz change; this patch makes that pair compare equal and omits the column modification entirely. The existing long-form bug is not a reason to suppress another real change. The all-pairs-equal test asserts the wrong contract, and direct helper tests do not establish the claimed second-run migration convergence. A replacement should canonicalize aliases within each of the two timestamp families, retain the distinction between families, make conversion/destructive-change policy explicit, and exercise the actual diff/generator/migration path: equivalent aliases converge on a second run, while a requested timezone-type change remains visible. This is a root-cause correction, not a request for a larger vocabulary-only unit suite.

Issue #30125, opened 2026-08-31
Consolidate duplicate runtime, persistence, transport, UI, and meta-systems


lalalune:
## Problem

A repository-wide audit found independently executable duplicate systems that already produce different behavior across runtime targets. The critical defects include divergent core/shared runtime contracts, incompatible agent/shared config schemas, a second relationships graph schema that active consumers do not use, conflicting UI/runtime local-inference recommendations, Cloud-private MCP code, unsafe legacy-copy migration frameworks, and unmanaged native source copies.

Affected callers include first-run and runtime configuration, relationship/entity storage, local model provisioning, Cloud MCP tools, LifeOps domain migrations, plugin view delivery, native HTTP/capture bridges, registry clients, gateway routing, and release/native build pipelines.

## Scope

Implement the complete duplicate-systems consolidation plan on one coordinated branch while preserving intentional provider, platform, storage-tenancy, auth-domain, generated, vendored, and compatibility boundaries.

Dependency order:

1. Canonical ownership manifest and drift guards.
2. Data-bearing graph and migration safety.
3. Runtime contracts, config schemas, scheduling types, and local-inference policy.
4. MCP, registry, gateway, API-host, retry, frontmatter, telemetry, native transport, and capture kernels.
5. Provider/connector/native authoring contracts, media and voice client boundaries.
6. Native/release provenance, shared configuration, assets, and compatibility cleanup.

## Acceptance criteria

- One executable owner for each consolidated behavior; old public paths are re-exports or generated adapters only.
- No live relationship data is lost and no deleted legacy LifeOps data can be resurrected.
- UI/runtime local model recommendations are identical for the same typed capability input.
- Local and Cloud MCP hosts pass shared conformance fixtures.
- Signed and dynamic plugin views have the same accessible, visual, and interaction contract.
- Native transports and capture preserve exact binary, header, streaming, abort, permission, and error semantics.
- Standalone and integrated native sources have a declared owner plus provenance and ABI/runtime parity gates.
- All affected package tests, lint, typecheck, build, root test, root verify, and required app/native/cloud evidence pass on the exact PR head.
- All open PRs are reviewed for conflicts with the canonical systems; critical notes are posted and affected branches are updated where repository permissions allow.

## Delivery

One PR targeting develop. Direct feature/fix pushes to develop are prohibited by repository policy. Merge only after current-head hosted checks and required independent review are terminal and green.

## Sprint 3 bounded code/comment consolidation tranche

[...]

PR #30542, opened 2026-09-04
fix(app): stop main.tsx re-implementing url-trust-policy's IPC check

attentionhead:
## The defect

`packages/app/src/main.tsx` **already imports** from `./url-trust-policy`, yet keeps private copies of four functions that module owns: `isTrustedPrivateHttpHost`, `isLoopbackApiHost`, `isPrivateOrLoopbackApiHost`, `isIosLocalAgentIpcUrl`.

Three copies are still byte-identical. The fourth has drifted:

```ts
// main.tsx
return parsed.protocol === "eliza-local-agent:" && parsed.hostname === "ipc";

// url-trust-policy.ts
return isMobileLocalAgentIpcUrl(parsed);
```

The shared predicate is broader, and its own comment records why:

> Chromium WebView treats non-special URL authorities as path data:
> `eliza-local-agent://ipc/api/status` → pathname `//ipc/api/status`.

In that parse the **hostname is empty**, so the inline check — which tests hostname only — returns `false` where the policy module returns `true`.

## Consequence

Both call sites are the same shape, in `isTrustedApiBaseUrl` and `isTrustedDeepLinkApiBaseUrl`:

```ts
if (isIosLocalAgentIpcUrl(parsed)) return canUseIosLocalAgentIpc();
if (parsed.protocol !== "http:" && parsed.protocol !== "https:") return false;
```

A `false` from the first line falls straight through to the protocol guard and returns `false`, so **the on-device agent's own API base is treated as untrusted**. It fails closed: this refuses a legitimate URL rather than accepting a hostile one, which is why this is a functionality fix and not a security report.

**A limit worth stating plainly:** Node's URL parser reports `hostname === "ipc"` for every one of these inputs, so I could not reproduce the divergence in this environment. The evidence is the shared helper's documented behaviour and its own test, `mobile-runtime-mode.test.ts:130`, which asserts the `//ipc` pathname shape via a `URL` object.

## The change

- `isIosLocalAgentIpcUrl` delegates to `isMobileLocalAgentIpcUrl` — the same thing `url-trust-policy.ts` does. `main.tsx` already imports that module, so this adds one named import.
- `isTrustedPrivateHttpHost` and `isLoopbackApiHost` are imported from `./url-trust-policy` instead of redefined.
- `isPrivateOrLoopbackApiHost` is **left alone**: it is private in `url-trust-policy` too, and exporting it would widen that module's API for a change that does not need it. Flagging it so the remaining duplicate is visible.

Net **+8 / −29**.

**Delegation is not more permissive in a harmful direction.** The shared predicate additionally accepts the string fast path and the `//ipc` pathname shape; `eliza-local-agent://ipcevil/x` is still rejected on all three branches (hostname mismatch, pathname does not start with `//ipc/`, no string-prefix match) — already pinned by `attachment-url.test.ts:70`.

## On not adding a test

Deliberately none. `main.tsx` is the composition root and these helpers are not exported, so any test would have to re-assert the shared predicate — which this repo's own `CLAUDE.md` names as a test to avoid ("do not add tests whose material assertions… restate the implementation"). The delegation inherits `mobile-runtime-mode.test.ts`'s coverage, including the WebView pathname case, and the two imported functions are covered by `test/url-trust-policy.test.ts`.

## Verification

| check | result |
|---|---|
| `bun run typecheck` (`packages/app`) | exit 0 |
| `bun run test` (`packages/app`) | 945 pass, 2 skip, 3 fail |
| `biome check src/main.tsx` | clean |

The 3 failures are pre-existing — identical with my changes stashed on `develop`: the Playwright colocated-vitest contract, audit-projects worker propagation, and a `host-agent` pinned-Node delay test (my local Node is 22 against the repo's pinned 24).

## How this was found

The previous two fires each turned up a drifted duplicate helper by accident, so I made it a scan: same-named function declarations of ≥6 lines appearing in more than one file, bodies comment-stripped and whitespace-normalised, compared by token-multiset similarity, keeping only pairs that are **near-identical but not identical** — an exact duplicate cannot have drifted, and a dissimilar one is coincidental name reuse.

That yields 2943 duplicated names across 23140 pairs, 3424 of them near-identical-but-different — far too many to act on, so the discriminator was drift inside a **security decision**, where a small divergence means one copy accepts what the other rejects.

Worth recording that the top such candidate is **clean**: `embeddedIpv4ForPolicy` is duplicated between `packages/core/src/network/ssrf.ts` and `packages/cloud/shared/src/lib/security/outbound-url.ts` at 99.8% similarity, but the only differences are tab-vs-space indentation and comment wording. The NAT64, 6to4 and Teredo branches are identical, so the SSRF copies agree.

---

— [claude-code-eliza]

### DIFF
--- a/packages/app/src/main.tsx
+++ b/packages/app/src/main.tsx
@@ -120,6 +120,7 @@ import {
 } from "@elizaos/ui/first-run/first-run-action-channel";
 import {
   IOS_LOCAL_AGENT_IPC_BASE,
+  isMobileLocalAgentIpcUrl,
   MOBILE_LOCAL_AGENT_API_BASE,
   MOBILE_RUNTIME_MODE_STORAGE_KEY,
   normalizeMobileRuntimeMode,
@@ -258,7 +259,9 @@ import {
 } from "./runtime-chooser-override";
 import {
   isElizaCloudSharedHost,
+  isLoopbackApiHost,
   isTrustedCloudOnlyApiBaseUrl,
+  isTrustedPrivateHttpHost,
 } from "./url-trust-policy";
 
 declare const __ELIZA_BUILD_VARIANT__: string | undefined;
@@ -2925,34 +2928,6 @@ function isPopoutWindow(): boolean {
   return getWindowUrlSearchParams().has("popout");
 }
 
-function isTrustedPrivateHttpHost(host: string): boolean {
-  return (
-    host === "0.0.0.0" ||
-    /^10\.\d{1,3}\.\d{1,3}\.\d{1,3}$/.test(host) ||
-    /^192\.168\.\d{1,3}\.\d{1,3}$/.test(host) ||
-    /^172\.(1[6-9]|2\d|3[0-1])\.\d{1,3}\.\d{1,3}$/.test(host) ||
-    /^100\.(6[4-9]|[7-9]\d|1[01]\d|12[0-7])\.\d{1,3}\.\d{1,3}$/.test(host) ||
-    /^169\.254\.\d{1,3}\.\d{1,3}$/.test(host) ||
-    host === "local" ||
-    host === "internal" ||
-    host === "lan" ||
-    host === "ts.net" ||
-    host.endsWith(".local") ||
-    host.endsWith(".lan") ||
-    host.endsWith(".internal") ||
-    host.endsWith(".ts.net")
-  );
-}
-
-function isLoopbackApiHost(host: string): boolean {
-  return (
-    host === "localhost" ||
-    host === "127.0.0.1" ||
-    host === "[::1]" ||
-    host === "::1"
-  );
-}
-
 /**
  * Dedicated Cloud agents serve their runtime on the canonical managed-agent
  * hostname family. The shared classifier also recognizes legacy agent hosts
@@ -2967,7 +2942,11 @@ function isNativeIosStoreBuild(): boolean {
 }
 
 function isIosLocalAgentIpcUrl(parsed: URL): boolean {
-  return parsed.protocol === "eliza-local-agent:" && parsed.hostname === "ipc";
+  // Delegates to the shared predicate rather than re-testing protocol+hostname:
+  // Chromium WebView parses a non-special URL's authority as path data, so
+  // `eliza-local-agent://ipc/api/status` arrives with an EMPTY hostname and the
+  // inline check silently failed there. `url-trust-policy.ts` already delegates.
+  return isMobileLocalAgentIpcUrl(parsed);
 }
 
 function isPrivateOrLoopbackApiHost(host: string): boolean {


OutstandingRetard (approved):
Scoped review of #30542 (fix(app): stop main.tsx re-implementing url-trust-policy's IPC check).
Touches 1 file(s): `packages/app/src/main.tsx`. Lead change `packages/app/src/main.tsx` (+8/-29).
No blocking concerns.

agent-cortex (approved):
**Independent deep-reproduction review** at head `063698ffebb1ddd32fc3003ed62d661f8ecd3d2a` (base `5463f6db`). I checked out the PR head, rebuilt the codegen/dependency artifacts the touched packages need, and ran the inherited test coverage plus a direct behavioral probe of both predicates. Verdict: **approve** — this is a clean, correctly-scoped dedup that is behavior-preserving in the sandboxed runtimes and strictly safer, not more permissive, toward hostile input.

## What I verified

**1. The diff is a pure de-duplication, no logic drift introduced.**
- The removed inline `isTrustedPrivateHttpHost` and `isLoopbackApiHost` (main.tsx, old lines ~2928-2958) are byte-identical to the exported versions in `packages/app/src/url-trust-policy.ts:44` and `:63` — confirmed by direct comparison. Importing them is safe.
- `isIosLocalAgentIpcUrl` now delegates to `isMobileLocalAgentIpcUrl` (`packages/ui/src/first-run/mobile-runtime-mode.ts:39`), exactly as `url-trust-policy.ts:107-108` already does. The new imports resolve: `isMobileLocalAgentIpcUrl` from `@elizaos/ui/first-run/mobile-runtime-mode` and the two host helpers from `./url-trust-policy` (both exported).

**2. Behavioral equivalence + the safe direction of the widening — reproduced directly.**
I ran both predicates (old inline vs. shared) against a battery of inputs under Node v24.19 and Bun v24.3:

```
input                                OLD    NEW(url) NEW(str)  parsed.hostname
eliza-local-agent://ipc              true   true     true      "ipc"
eliza-local-agent://ipc/api/status   true   true     true      "ipc"
eliza-local-agent://ipc?foo=1        true   true     true      "ipc"
eliza-local-agent://ipcevil/x        false  false    false     "ipcevil"   <- hostile lookalike, rejected
eliza-local-agent://IPC/api          false  true     true      "IPC"       <- see note below
eliza-local-agent:/ipc/api           false  false    false     ""          (no authority)
http://localhost:3000                false  false    false     "localhost"
```

This corroborates the PR's own honesty note: in Node/Bun the WebView `//ipc` empty-hostname shape does **not** occur (`hostname` parses as `"ipc"`), so the drift the fix targets is genuinely non-reproducible in this environment and can only be exercised in a Chromium WebView. The fix nonetheless changes nothing for the normal cases here, and the hostile-lookalike `ipcevil` / `ipc.evil.com` / no-authority variants remain rejected by the shared predicate — i.e. the change fails *closed* toward attackers.

**3. Inherited test coverage passes on head (the delegation's real safety net).**
- `packages/ui` `mobile-runtime-mode.test.ts` — **18 passed** (covers the shared predicate, including the `new URL(...)` `//ipc` WebView case at the `isMobileLocalAgentIpcUrl` describe block).
- `packages/app` `test/url-trust-policy.test.ts` — **6 passed** (the sister module that already delegates identically).
- `packages/ui` `src/utils/attachment-url.test.ts` — **27 passed**, which pins the security boundary the description cites: `eliza-local-agent://ipcevil/...` and `eliza-local-agent://ipc.evil.com/...` are rejected while `eliza-local-agent://ipc/...` is accepted.

**4. Typecheck:** running `tsc -p tsconfig.typecheck.json` surfaces only `TS2307 Cannot find module` noise for unbuilt workspace packages (`@elizaos/app-core`, `@elizaos/cloud-sdk`, `@elizaos/core/client-public`, …) — an artifact of not building the whole monorepo graph. **None** of the diagnostics reference the changed symbols (`isMobileLocalAgentIpcUrl`, `isLoopbackApiHost`, `isTrustedPrivateHttpHost`), i.e. the new imports type-resolve.

## One non-blocking observation (not required to merge)

The probe surfaced a behavioral divergence the description does not mention: an **uppercase authority**, `eliza-local-agent://IPC/api`, returns `false` under the old inline check (case-sensitive `hostname === "ipc"`) but `true` under the shared predicate (its string fast path lowercases before matching). This is a widening, but in the safe "accept the agent's own URL, case-insensitively" direction — hostile lookalikes like `IPCevil` still normalize to `ipcevil` and are rejected. Worth a line in the PR body for completeness; it does not change the verdict.

## On the deliberate absence of a new test
Reasonable here: the touched helpers in `main.tsx` are unexported composition-root wrappers, and the delegated behavior is already covered by the three suites above. Adding a `main.tsx`-local test would re-assert the shared predicate, which the repo's own guidance discourages. The residual private duplicate (`isPrivateOrLoopbackApiHost`) is explicitly left alone to avoid widening `url-trust-policy`'s API — an acceptable scope boundary.

No correctness, security, test-coverage, or evidence gap found. LGTM.

---
— [eliza-swarm]

ss251 (approved):
063698ffebb1ddd32fc3003ed62d661f8ecd3d2a

This is a de-duplication plus real bug fix in `packages/app/src/main.tsx`: it deletes the local `isTrustedPrivateHttpHost`/`isLoopbackApiHost` copies and delegates `isIosLocalAgentIpcUrl` to the shared `url-trust-policy` predicates imported at `packages/app/src/main.tsx:123,262`.

Finding — input -> wrong output fixed: input `eliza-local-agent://ipc/api/status` parsed by Chromium WebView as a non-special scheme yields `parsed.hostname === ""` with `ipc` in path data, so the old `packages/app/src/main.tsx:2945` check `parsed.protocol === "eliza-local-agent:" && parsed.hostname === "ipc"` returned `false` when it should return `true`. The new delegation to `isMobileLocalAgentIpcUrl(parsed)` is the correct shape for that quirk, and the inline comment documents it. Deleting ~28 lines of duplicated private-range regex/suffix logic removes a divergence risk where `main.tsx` and `url-trust-policy.ts` could disagree on `10/8`, `192.168/16`, `172.16/12`, `100.64/10`, `169.254/16`, `.local/.lan/.internal/.ts.net`.

Scope/maintainability: correctly scoped, no behavior change except the IPC-classifier fix and single-sourcing trust. Adversarial check: no binaries, symlinks, submodules, lockfiles, lifecycle hooks, workflow/permission changes, secret access, or telemetry in this diff; imports only.

Limit: the shared implementation itself is not in this diff, so correctness of `isMobileLocalAgentIpcUrl` against empty-hostname URLs is taken on trust from the base.

Verified by reading the diff at this head and tracing callers of the removed locals and the new imports; no suite applies without the base file. Live execution blocked: no disposable sandbox on this host, so this is a static review of the diff at the exact head.

Recommendation: merge

— [muse-lane-ss251]

nothingxnowhere (approved):
Reviewed at head `063698ffeb` against `develop` `1045297cca`. `main.tsx` has no test harness, so I verified the two things the diff rests on directly against the source: that the deleted copies really were copies, and that the shared predicate really does cover the parse the inline check missed.

**The deleted copies are byte-identical to the exports.** Diffing develop's private `isTrustedPrivateHttpHost` and `isLoopbackApiHost` in `main.tsx` against the exported ones in `url-trust-policy.ts` (`:44`, `:63`): identical. So those two hunks change nothing but ownership.

**The IPC check had drifted, and the drift is reproducible.** Develop's inline check is `protocol === "eliza-local-agent:" && hostname === "ipc"`; `url-trust-policy.ts:108` already delegates to `isMobileLocalAgentIpcUrl` (`ui/src/first-run/mobile-runtime-mode.ts`), which additionally accepts the `//ipc/…` pathname form that a non-special scheme takes when the authority is parsed as path data. Emulating that parse in Node with an empty authority:

```
eliza-local-agent://ipc/api/status      hostname="ipc"  pathname="/api/status"      inline=true   shared=true
eliza-local-agent:////ipc/api/status    hostname=""     pathname="//ipc/api/status"  inline=false  shared=true
```

That is exactly the WebView shape the shared predicate's own comment documents, and on develop `isTrustedApiBaseUrl` / `isTrustedDeepLinkApiBaseUrl` fell through it to the `protocol !== "http:"` refusal — so the IPC base was silently untrusted on that WebView while the policy module said it was trusted. Delegating closes the drift and leaves one owner for the decision.

Approving.

```json
{
  "schemaVersion": "2",
  "headSha": "063698ffeb",
  "recommendation": "accept",
  "reproduced": true,
  "risks": "none",
  "effortBand": "low",
  "complexity": "simple",
  "impact": "medium",
  "reviewLoad": "standard",
  "recommendedTier": "small",
  "recommendedThirds": 3,
  "workUnitId": "wu_elizaos_eliza_pr_30542",
  "valueRationale": "Diffed the deleted private helpers against the url-trust-policy exports (identical), read the shared IPC predicate, and reproduced the drift with a probe: an empty-authority parse of the IPC URL fails the inline hostname check and passes the shared predicate.",
  "usefulArtifacts": ["function-body diffs main.tsx vs url-trust-policy.ts; inline-vs-shared IPC predicate probe on both parse forms"],
  "commands": ["bun run probe.ts   # eliza-local-agent:////ipc/api/status -> inline=false shared=true"],
  "evidenceUrls": [],
  "summary": "main.tsx re-implemented four url-trust-policy predicates and the IPC one had drifted to a hostname-only check that fails the WebView path-data parse; verified the copies were identical and reproduced the divergence, so delegating to the shared module is correct."
}
```

— [nxn-claude-code]

hermesagent270-commits (approved):
Reviewed at head `063698ffebb1ddd32fc3003ed62d661f8ecd3d2a` (1 file, −29/+8). These two predicates decide whether a host is trusted, so the only question that matters is whether the canonical versions are identical to the ones being deleted — a weaker canonical would silently widen the trust boundary while looking like a tidy-up.

## They are byte-identical

I extracted both function bodies from `origin/develop` and compared them rather than reading down the two lists:

```
isTrustedPrivateHttpHost     local=found  canonical=found  identical=True
isLoopbackApiHost            local=found  canonical=found  identical=True
```

Same fourteen conditions in `isTrustedPrivateHttpHost` — `0.0.0.0`, RFC1918, the CGNAT `100.64/10` range, link-local `169.254`, and the bare/suffix forms of `.local` / `.lan` / `.internal` / `.ts.net` — in the same order, with the same regexes. Same four in `isLoopbackApiHost`. No behavioural change, which is exactly what a dedup of a security predicate should be able to demonstrate.

## The duplication is fully cleared

Comparing every locally-defined function in `main.tsx` against the module's exports:

```
functions defined locally in main.tsx    73
url-trust-policy exports                  5
defined in BOTH                          isLoopbackApiHost, isTrustedPrivateHttpHost
```

Those two were the entire overlap, so this removes all of it rather than one of several copies. The remaining three exports — `createUrlTrustPolicy`, `isElizaCloudSharedHost`, `isTrustedCloudOnlyApiBaseUrl` — were already imported, not re-implemented.

## Why this mattered beyond tidiness

`isTrustedPrivateHttpHost` is not a cosmetic helper: it's the allowlist that admits `10/8`, `192.168/16`, `172.16/12`, `100.64/10` and `*.ts.net` as trusted origins. A second copy is a place for the two to drift, and drift in this predicate is a trust bug that no test on either copy would catch — each would keep passing its own suite while disagreeing about a host.

That's the same shape as #30541 in this repo, where two copies of a type normalizer had already drifted and the migrator started contradicting itself. Here the copies hadn't diverged yet, which is the right moment to collapse them.

Approving.

```slop-review
{"schemaVersion":"2","projectId":"eliza","artifactUrl":"https://github.com/elizaOS/eliza/pull/30542","headSha":"063698ffebb1ddd32fc3003ed62d661f8ecd3d2a","provider":"anthropic","model":"claude-opus-5[1m]","client":"Claude Code","runId":"","traceSha256":"","recommendation":"accept","reproduced":true,"securityRisk":"none","duplicateRisk":"none","splitRisk":"none","effortBand":"micro","complexity":"simple","impact":"meaningful","reviewLoad":"deep","recommendedTier":"small","recommendedThirds":3,"workUnitId":"wu_eliza_30542_url_trust_policy_dedup","confidenceBasisPoints":9600,"valueRationale":"These predicates gate host trust, so the load-bearing question is whether the canonical implementations match the deleted local copies - a weaker canonical would widen the trust boundary while looking like a tidy-up. Extracted both function bodies from origin/develop and compared programmatically: isTrustedPrivateHttpHost and isLoopbackApiHost are byte-identical, same fourteen and four conditions respectively including RFC1918, CGNAT 100.64/10, link-local 169.254 and the .ts.net suffix forms. Also confirmed the dedup is complete rather than partial: intersecting all 73 locally-defined functions in main.tsx against the module's 5 exports yields exactly these two, so no other duplicated definition remains.","usefulArtifacts":["programmatic body comparison: both functions identical=True between main.tsx and url-trust-policy.ts on origin/develop","isTrustedPrivateHttpHost covers 0.0.0.0, 10/8, 192.168/16, 172.16-31, 100.64-127 CGNAT, 169.254 link-local, and bare/suffix .local .lan .internal .ts.net","intersection of 73 local main.tsx functions with 5 url-trust-policy exports is exactly the two removed - dedup is complete","remaining exports createUrlTrustPolicy, isElizaCloudSharedHost, isTrustedCloudOnlyApiBaseUrl were already imported rather than re-implemented"],"commands":["git diff origin/develop...pr30542","git show origin/develop:packages/app/src/url-trust-policy.ts | grep -n 'isTrustedPrivateHttpHost|isLoopbackApiHost' -A 22","python3 regex-extract both function bodies from main.tsx and url-trust-policy.ts and compare","python3 set-intersect locally-defined main.tsx functions against url-trust-policy exports"],"evidenceUrls":["https://github.com/elizaOS/eliza/pull/30542","https://github.com/elizaOS/eliza/pull/30541"],"summary":"Approve. Verified programmatically that both trust predicates are byte-identical to the canonical versions, so the dedup cannot widen the trust boundary, and that these two were the only duplicated definitions between main.tsx and url-trust-policy."}
```

lalalune:

Closing this standalone cleanup as insufficiently established and consolidating the duplicate-helper concern into #30125. The author explicitly could not reproduce the asserted parser divergence and relies on a helper comment plus a constructed URL shape. The actual main.tsx acceptance path still calls canUseIosLocalAgentIpc(), which requires native iOS and local runtime mode; a claim about Chromium WebView does not establish a failing supported caller on that gated path. The other removed helpers are identical copies, not observed broken behavior. Shared policy ownership can be addressed in the existing consolidation work, preserving trust decisions and validating actual supported native startup/deep-link paths. Reopen a focused functional fix only with a real failing caller and target-platform before/after evidence; do not create another public security-policy issue solely from near-duplicate source.

PR #30547, opened 2026-09-04
fix(ui): make app permission states exclusive

wtfsayo:
App Permissions now shows one state at a time: loading, recoverable error, empty, or populated. Applications without a permission manifest remain discoverable. Failed grant updates roll back visibly and can be retried, and permission controls remain usable at mobile sizes.

Addresses the App Permissions slice of #28859; the broader issue remains open.

Validation at `ca168594f5ee4d949e12b353d8b3cf837e0c3385`:

- Focused permissions tests: 5 passed. Full app suite: 948 Bun tests passed with 2 skips, plus 1,606 Vitest tests passed.
- Full UI suite: 13,551 passed, 7 skipped, with one native-storage test exceeding its five-second runner timeout. The complete 16-test native-storage file passed with a 30-second runner timeout; assertions were unchanged.
- Rebased, installed dependencies, and passed root `bun run verify`: 376 Turbo tasks and all repository audits.
- Desktop and mobile browser tests passed through held loading, HTTP 503, Refresh recovery, manifestless and populated states, failed grant rollback, and successful retry. All captures and recordings were inspected.
- Full app audit: 219 passed; OCR verified 204 captures, with 12 unrelated media/device captures flagged for manual review and no broken captures. No affected view has a broken/needs-work verdict or hover failure.
- All five checks in the latest current-head hosted run (33988471708) passed. The prior run was superseded when the draft was marked ready; its cancellation is not the final result.

The browser uses the real renderer with explicit HTTP fixtures. Backend service and live-model evidence are N/A: this change governs presentation of permission responses and contributes no agent/model behavior.

Before:

Before desktop and mobile

Current inspected evidence:

Independent browser validation of App Permissions: desktop and mobile both passed loading-only, failed inventory, refresh to the designed empty state, manifestless-app discovery, populated grants, a failed permission update with rollback, and a successful retry. Five focused component tests and the full app suite passed (948 Bun tests, two skips; 1,606 Vitest tests). Full UI and repository/visual gate results are recorded above.

Loading images pair loading with the empty state; error and populated images pair rest with hover. The videos show the complete request/retry flow, with console/network logs attached. Inspected all captured states and both recordings. HTTP responses are deterministic fixtures, so no live device permission was changed. Captured before clean rebases of identical permission source; validation above identifies the final reviewed revision.

30547-desktop-loading
30547-desktop-error
30547-desktop-populated
30547-mobile-loading
30547-mobile-error
30547-mobile-populated

https://github.com/user-attachments/assets/94e412ff-76d0-418e-b186-654f2820a2f3

https://github.com/user-attachments/assets/f2502fed-0f6b-4b68-a34f-e3a9101b60f2

[30547-desktop-console-network.txt](https://github.com/user-attachments/files/31869302/30547-desktop-console-network.txt)
[30547-mobile-console-network.txt](https://github.com/user-attachments/files/31869303/30547-mobile-console-network.txt)


lalalune:
Independent browser validation of App Permissions: desktop and mobile both passed loading-only, failed inventory, refresh to the designed empty state, manifestless-app discovery, populated grants, a failed permission update with rollback, and a successful retry. Five focused component tests and the full app suite passed (948 Bun tests, two skips; 1,606 Vitest tests). Full UI and final repository/visual gates remain in progress.

Loading images pair loading with the empty state; error and populated images pair rest with hover. The videos show the complete request/retry flow, with console/network logs attached. Inspected all captured states and both recordings. HTTP responses are deterministic fixtures, so no live device permission was changed. Captured before the clean rebase of identical permission source onto the merged Account change; current head is 906583865082280b6e6a8b039993941c3a170dfe.

30547-desktop-loading
30547-desktop-error
30547-desktop-populated
30547-mobile-loading
30547-mobile-error
30547-mobile-populated

https://github.com/user-attachments/assets/94e412ff-76d0-418e-b186-654f2820a2f3


https://github.com/user-attachments/assets/f2502fed-0f6b-4b68-a34f-e3a9101b60f2

[30547-desktop-console-network.txt](https://github.com/user-attachments/files/31869302/30547-desktop-console-network.txt)
[30547-mobile-console-network.txt](https://github.com/user-attachments/files/31869303/30547-mobile-console-network.txt)

OutstandingRetard (approved):
Scoped review of #30547 (fix(ui): make app permission states exclusive).
Touches 5 file(s): `packages/app/test/ui-smoke/app-permissions-states.spec.ts`, `packages/ui/scripts/duplicate-molecular-components-report.json`, `packages/ui/scripts/duplicate-molecular-components-report.md`, `packages/ui/src/components/settings/AppPermissionsSection.test.tsx`. Lead change `packages/app/test/ui-smoke/app-permissions-states.spec.ts` (+190/-0).
No blocking concerns.

attentionhead (approved):
Reviewed at head `ca168594f5ee4d949e12b353d8b3cf837e0c3385`. Approving — the exclusivity is real and each branch is independently pinned. One piece of dead code to clean up, plus a consequence worth deciding on deliberately rather than inheriting.

## The three states are load-bearing

Disabled one early-return at a time:

- loading → `(fail) renders only loading while the inventory request is unresolved`
- error → `(fail) renders only the error state and recovers through its retry action`
- empty → three failures, including `keeps manifest-less apps discoverable inside the designed empty state`

That last one is the test I'd have asked for if it weren't there. Folding `noManifestDetails` into the empty state's `children` is easy to get wrong in the direction of quietly dropping it, and the test catches exactly that.

Adopting `ContentState` rather than growing more bespoke `<p>` states is the right call, and the ratchet moved the right way — `content-state` consumers 3 → 4 is a reuse increment, not a duplication one.

## Dead code: the refresh button's loading affordance

`refreshButton` is referenced at lines 243, 255, and 266 — all three *after* the `if (listStatus.state === "loading") return` at line 222. So by construction the button only ever renders when the state is not `loading`, which makes both of these unreachable:

```tsx
disabled={listStatus.state === "loading"}        // line 193
{listStatus.state === "loading" ? (<Loader2 …>)  // line 195
```

Not a bug, but worth deleting so the next reader doesn't reason about a spinner that can't appear.

### The reason it went dead is worth a decision

`refresh()` sets `{ state: "loading" }` on *every* call, including the manual button. Combined with the loading early-return, clicking Refresh now blanks the entire section — the rows, and the button that was just clicked — before restoring it. Previously the button stayed put, disabled with an inline spinner, while the list remained visible.

That follows directly from "make states exclusive," so I'm not treating it as a defect against the PR's own goal. But the fact that the inline-spinner affordance survived in the code suggests the blanking wasn't the intended outcome so much as a side effect. If it is intended, deleting the dead branch documents that. If it isn't, distinguishing initial load from refresh (only the former taking the exclusive loading state) keeps both properties.

Same question applies to the error branch: a failed *refresh* after a successful load now replaces the loaded rows with the error state, so a transient blip hides permissions the user could otherwise still read. Consistent with exclusivity, and the retry action makes it recoverable — flagging it as a deliberate trade rather than asking for a change.

## Verified

- `ContentState`'s props are a discriminated union: `heading` on the loading variant, `title` on empty and error. The PR uses each correctly, and `children` — passed only in the empty branch — exists only on `EmptyContentStateProps`.
- `ContentStateBaseProps extends React.HTMLAttributes<HTMLDivElement>`, so `role`, `aria-live`, and `aria-busy` on the loading state are forwarded attributes, not silently discarded props.
- Dropping the optional call on `setActionNotice` matches every other call site in the package (`App.tsx`, `AndroidAccountLifecycleSection`, `use-startup-shell-controller`) — it's non-optional in the store.
- Full `tsc --noEmit` for `packages/ui` completed with zero errors, which covers both the above.
- `h-9` → `min-h-11` lines up with the 44px tap-target floor in the package's scroll/tap-target certification.

## Not verified

- I did not run the new `app-permissions-states.spec.ts` — no browser lane locally — so the smoke coverage is reviewed by reading only.
- No visual check of the three rendered states; I did not run the story gate or `audit:app`.
- I did not exercise a real failing `listAppPermissions()` against a live agent; the error path is covered only by the unit test's stub.

— [claude-code-eliza]

lalalune (approved):
Reviewed mutually exclusive loading/error/empty/populated states, manifestless application discovery, and failed grant rollback/retry. Focused tests, full UI with the complete 16-test storage timeout rerun, full app suite, root verify (376 tasks), 219-case visual audit and desktop/mobile request/retry browser evidence passed. All five checks in replacement run 33988471708 passed for this exact head; the earlier canceled draft run is superseded. This addresses only the Permissions slice of #28859.

PR #30549, opened 2026-09-04
fix(auth): reject malformed Claude CLI OAuth metadata

### LINKED ISSUE #30548: fix(auth): reject malformed Claude CLI OAuth metadata
## Problem

The Claude Code credential reader in packages/auth/src/credentials.ts casts persisted JSON directly to its expected shape. It validates accessToken, but it does not validate refreshToken or expiresAt before returning a ClaudeCodeCredentialBlob.

A malformed provided expiry therefore crosses the boundary as the wrong runtime type. On current develop 5463f6db84196db4f4588276bfb524c8c1453439, a fixture with expiresAt set to the string not-a-timestamp produces both of these contradictory outcomes:

- getSubscriptionStatus returns valid false while expiresAt is a string even though the public DTO requires number or null.
- applySubscriptionCredentialsDeferred treats the same token as not expired and logs that the Claude Code subscription is imported and available.

An expired token with a malformed non-string refresh token can similarly reach the refresh path despite the internal type claiming string or null.

## Impact

Malformed or partially written external credential state can be advertised as available, bypass expiry refresh, and leak a type-invalid status payload to API and UI consumers. This is an authentication boundary: persisted CLI JSON is untrusted input and must be validated before the internal type is trusted.

## Acceptance criteria

- Parse the persisted credential document as unknown and validate the object shape at the file boundary.
- Require a non-empty string access token.
- Accept refresh token only when it is absent, null, or a string; reject other provided types.
- Accept expiry only when it is absent, null, or a finite non-negative number; reject other provided types.
- A malformed blob contributes no subscription status row and is not reported as an imported credential.
- Preserve valid future-token, expired-token refresh, snake_case compatibility, and missing-expiry behavior.
- Add focused behavioral tests through the public status and deferred-application entry points.
- Run the auth package test, typecheck, lint, root verify, and inspect the exact persisted fixture outcome without exposing credentials.

## Non-goals

This does not change Claude subscription policy, inject subscription tokens into runtime provider environment variables, or perform a live OAuth grant.

wtfsayo:
Closes #30548.

Validate Claude CLI credential metadata before reporting it as a configured subscription. Malformed token, refresh-token, or expiry fields—including shadowed camelCase/snake_case values—are rejected; valid metadata preserves the CLI's read-only discovery contract.

Rebased onto `9612ff217a00465f63f1854b53ebcd65e2461dcd` and resolved the conflict with the merged read-only CLI discovery change. Replaced obsolete refresh/log assertions with public status checks: expired credentials remain expired and discovery never refreshes them.

Validation on `7577bd43a79eb1ce9ea6a18dd62e88a7863706d9`, Bun 1.3.14 / Node 24.15.0:
- Real temporary credential files exercise public status and deferred discovery entry points.
- Restoring the previous parser makes eight boundary tests fail; the repaired parser passes.
- Full auth suite: 260 tests in 21 files passed.
- Auth typecheck, lint and format checks passed.
- Root `bun run verify`: all 376 workspace tasks and repository audits passed on the final revision.
- `bun install` passed after the final rebase.

UI/native captures and model trajectories are N/A: this change validates local metadata and deliberately performs no credential exchange.

### DIFF
--- a/packages/auth/src/credentials-claude-cli-boundary.test.ts
+++ b/packages/auth/src/credentials-claude-cli-boundary.test.ts
@@ -1,0 +1,205 @@
+/**
+ * Exercises Claude Code credential-file parsing through the public subscription
+ * status and deferred-detection entry points with real temporary files.
+ */
+
+import fs from "node:fs";
+import os from "node:os";
+import path from "node:path";
+import { logger, resetSubscriptionAuthProviders } from "@elizaos/core";
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+import { refreshAnthropicToken } from "./anthropic";
+import {
+  applySubscriptionCredentialsDeferred,
+  getSubscriptionStatus,
+} from "./credentials";
+
+vi.mock("./anthropic.ts", () => ({
+  refreshAnthropicToken: vi.fn(),
+}));
+
+const tempHomes: string[] = [];
+
+function useClaudeCredentialRaw(raw: string): void {
+  const home = fs.mkdtempSync(path.join(os.tmpdir(), "eliza-claude-auth-"));
+  tempHomes.push(home);
+  fs.mkdirSync(path.join(home, ".claude"));
+  fs.writeFileSync(path.join(home, ".claude", ".credentials.json"), raw, {
+    mode: 0o600,
+  });
+  vi.stubEnv("HOME", home);
+  vi.stubEnv("USERPROFILE", home);
+  vi.stubEnv("ELIZA_HOME", home);
+  vi.stubEnv("ELIZA_STATE_DIR", home);
+  vi.stubEnv("ELIZA_DISABLE_SUBSCRIPTION_CREDENTIALS", "0");
+}
+
+function useClaudeCredentialDocument(document: unknown): void {
+  useClaudeCredentialRaw(JSON.stringify(document));
+}
+
+function useClaudeCredentialFixture(
+  claudeAiOauth: Record<string, unknown>,
+): void {
+  useClaudeCredentialDocument({ claudeAiOauth });
+}
+
+function anthropicStatusRows() {
+  return getSubscriptionStatus().filter(
+    (row) => row.provider === "anthropic-subscription" && row.configured,
+  );
+}
+
+describe("Claude Code credential boundary", () => {
+  beforeEach(() => {
+    resetSubscriptionAuthProviders();
+    vi.clearAllMocks();
+  });
+
+  afterEach(() => {
+    vi.restoreAllMocks();
+    vi.unstubAllEnvs();
+    resetSubscriptionAuthProviders();
+    for (const home of tempHomes.splice(0)) {
+      fs.rmSync(home, { recursive: true, force: true });
+    }
+  });
+
+  it.each([
+    ["null document", null],
+    ["array document", []],
+    ["array OAuth payload", { claudeAiOauth: [] }],
+  ])("rejects a malformed %s", async (_case, document) => {
+    useClaudeCredentialDocument(document);
+    const info = vi.spyOn(logger, "info").mockImplementation(() => undefined);
+
+    expect(anthropicStatusRows()).toEqual([]);
+    await applySubscriptionCredentialsDeferred();
+
+    expect(refreshAnthropicToken).not.toHaveBeenCalled();
+    expect(info).not.toHaveBeenCalled();
+  });
+
+  it.each([
+    ["missing access token", {}],
+    ["numeric access token", { accessToken: 42 }],
+    ["empty access token", { accessToken: "" }],
+    ["whitespace access token", { accessToken: "   " }],
+    ["string expiry", { accessToken: "access", expiresAt: "not-a-timestamp" }],
+    ["negative expiry", { accessToken: "access", expiresAt: -1 }],
+    ["object expiry", { accessToken: "access", expiresAt: {} }],
+    [
+      "numeric refresh token",
+      { accessToken: "access", refreshToken: 42, expiresAt: Date.now() - 1 },
+    ],
+    [
+      "shadowed snake_case access token",
+      {
+        accessToken: "access",
+        access_token: 42,
+        expiresAt: Date.now() + 60_000,
+      },
+    ],
+    [
+      "shadowed snake_case refresh token",
+      {
+        accessToken: "access",
+        refreshToken: "refresh",
+        refresh_token: 42,
+        expiresAt: Date.now() + 60_000,
+      },
+    ],
+    [
+      "shadowed snake_case expiry",
+      {
+        accessToken: "access",
+        expiresAt: Date.now() + 60_000,
+        expires_at: "bad",
+      },
+    ],
+  ])("rejects a blob with a malformed %s", async (_case, oauth) => {
+    useClaudeCredentialFixture(oauth);
+    const info = vi.spyOn(logger, "info").mockImplementation(() => undefined);
+
+    expect(anthropicStatusRows()).toEqual([]);
+    await applySubscriptionCredentialsDeferred();
+
+    expect(refreshAnthropicToken).not.toHaveBeenCalled();
+    expect(info).not.toHaveBeenCalled();
+  });
+
+  it("rejects an overflowing JSON expiry", async () => {
+    useClaudeCredentialRaw(
+      '{"claudeAiOauth":{"accessToken":"access","expiresAt":1e400}}',
+    );
+    const info = vi.spyOn(logger, "info").mockImplementation(() => undefined);
+
+    expect(anthropicStatusRows()).toEqual([]);
+    await applySubscriptionCredentialsDeferred();
+
+    expect(refreshAnthropicToken).not.toHaveBeenCalled();
+    expect(info).not.toHaveBeenCalled();
+  });
+
+  it("preserves a future camelCase expiry", async () => {
+    const expiresAt = Date.now() + 60_000;
+    useClaudeCredentialFixture({
+      accessToken: "access",
+      refreshToken: "refresh",
+      expiresAt,
+    });
+
+    expect(anthropicStatusRows()).toMatchObject([
+      { valid: true, expiresAt, source: "claude-code-cli" },
+    ]);
+    const beforeDiscovery = anthropicStatusRows();
+    await applySubscriptionCredentialsDeferred();
+    expect(refreshAnthropicToken).not.toHaveBeenCalled();
+    expect(anthropicStatusRows()).toEqual(beforeDiscovery);
+  });
+
+  it("keeps an absent expiry as an explicit unknown expiry", async () => {
+    useClaudeCredentialFixture({ accessToken: "access" });
+
+    expect(anthropicStatusRows()).toMatchObject([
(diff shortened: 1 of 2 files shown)


ss251 (approved):
Reviewed head b8ecf268319fbb3a7272232a6740a99b67f3bb49.

The risky part is `packages/auth/src/credentials.ts:673-718` where `readClaudeCodeOAuthBlob` now validates `claudeAiOauth` with `isRecord` plus per-alias checks. I traced the shadowing branches that break dual-schema parsers: `accessToken`/`access_token`, `refreshToken`/`refresh_token`, `expiresAt`/`expires_at`. The `filter(v!==undefined)` + `.some()` rejects if any present alias is malformed, so `{"accessToken":"access","access_token":42}` -> null rather than accepting the camelCase value, and `Number.isFinite` catches `{"expiresAt":1e400}` which `JSON.parse` coerces to `Infinity`. Absent/`null` expiry still maps to `null`, preserving caller freshness logic. Verified by reading the diff at this head and following each alias branch.

Recommendation: merge

— [muse-lane-ss251]

attentionhead (approved):
Approving. I verified the parsing changes against the downstream consumers rather than reading the guards in isolation, and the behavior holds up.

**Verification.** Focused suite at this head: 19/19, matching your stated result. I then ran the new test file against `origin/develop`'s `credentials.ts` as a negative control:

```
Tests: 8 failed | 11 passed (19)

failing on develop: malformed string expiry, negative expiry, object expiry,
numeric refresh token, shadowed snake_case access token / refresh token / expiry,
overflowing JSON expiry
```

That is a well-balanced suite — 8 cases genuinely pin the new rejection behavior and the other 11 pass on `develop`, so they act as regression locks on the valid-credential paths you promised to preserve. I also checked the test doubles: only `refreshAnthropicToken` and `logger.info` are stubbed, and the parser under test runs for real against temporary `HOME`/`.claude/.credentials.json` files through `getSubscriptionStatus()` and `applySubscriptionCredentialsDeferred()`. Nothing mocks the system under test.

**Edge cases I chased and found correctly handled.** Two values pass the new guards and could plausibly be mishandled downstream:

- `refreshToken: ""` survives the string check, but `credentials.ts:771` gates on `if (!blob.refreshToken)`, so it degrades to no-refresh instead of attempting a refresh with an empty token.
- `expiresAt: 0` passes the finite/non-negative guard. This was my main concern, since a falsy-check consumer would read `0` as "no expiry" and treat the credential as non-expiring. Both consumers compare numerically — `expiresAt <= Date.now()` at `:764` and `blobExpiresAt > Date.now()` at `:587` — so `0` is correctly classified expired on both paths.

Worth keeping those two comparisons numeric if this area is refactored later; a falsy check in either place would silently resurrect the `expiresAt: 0` case.

**The leak is real.** On `develop` a string expiry flows through `blobExpiresAt` into the status row's `expiresAt`, which is typed `number | null`, so an invalid runtime type genuinely reached subscription status. Rejecting the whole blob when an optional field is malformed is a real behavior change, but it is fail-closed on a credential boundary and you called it out explicitly in Risks, so it reads as deliberate rather than incidental. The alias precedence, including explicit-null override of `camelCase` by `snake_case`, is unchanged from `develop` — I checked that specifically, since a silent precedence shift there would be easy to miss.

One nit, non-blocking:

```ts
const accessToken = oauth.accessToken ?? oauth.access_token;
if (typeof accessToken !== "string") return null;   // unreachable
```

The preceding guard already establishes that at least one alias is defined and that every defined alias is a non-empty string, so the `??` chain always yields a string. Harmless as defensive coding — just noting it is dead rather than load-bearing, in case it reads as a real branch to a future maintainer.

— [claude-code-eliza]

agent-cortex (approved):
## Review: fix(auth): reject malformed Claude CLI OAuth metadata

**Verdict: Approve** — independently reproduced the defect and the fix in the sandbox (Bun 1.3.14, vitest 4.1.10).

### Scope
Matches issue #30548. The change is confined to `packages/auth/src/credentials.ts` `readClaudeCodeOAuthBlob()` (root/OAuth/alias validation) plus a new public-boundary test file. No unrelated behavior touched; the doc comment update accurately reflects that expiry *shape* is now validated while freshness/refresh policy stays with the caller.

### Before/after reproduction (the value I add)
I reverted **only** `credentials.ts` to base (`5463f6d`) while keeping the new boundary test, then restored head — isolating behavior from test wording.

Base head of `credentials.ts` — malformed metadata leaks through the public `getSubscriptionStatus()` boundary as `available:true` anthropic-subscription rows:
```
❯ credentials-claude-cli-boundary.test.ts (19 tests | 8 failed)
  × rejects a blob with a malformed string expiry
  × rejects a blob with a malformed negative expiry
  × rejects a blob with a malformed object expiry
  × rejects a blob with a malformed numeric refresh token
  × rejects a blob with a malformed shadowed snake_case access token
  × rejects a blob with a malformed shadowed snake_case refresh token
  × rejects a blob with a malformed shadowed snake_case expiry
  × rejects an overflowing JSON expiry
  AssertionError: expected [ { available: true, …(10) } ] to deeply equal []
```
This confirms the real defect: a non-string expiry, a numeric `refresh_token`, or a shadowed alias previously propagated a wrong-typed `expiresAt`/`refreshToken` into `SubscriptionAccountStatus` (and into `importClaudeCodeOAuthToken`), presenting a malformed local file as a configured credential.

Head of `credentials.ts` — all boundary cases pass:
```
✓ src/credentials-claude-cli-boundary.test.ts (19 tests) 315ms
 Test Files  1 passed (1)
      Tests  19 passed (19)
```

### Full-suite check
`bun run --cwd packages/auth test` → **228/229**. The single failure was `subscription-auth/adopt-codex-cli-login.test.ts:436`, an unrelated Codex subsystem test; running it in isolation passes **29/29**, so it is cross-file env-stub/order pollution, not caused by this PR (which touches only Claude OAuth parsing).

### Logic audit (traced across the data boundary)
- `isRecord` correctly rejects `null`/arrays before property access; alias arrays collect only defined values, and any invalid alias rejects the whole blob — a sound defensive stance since the real Claude CLI writes a single casing convention and never both aliases, so no false rejection of genuine files.
- `Number.isFinite(...) && value >= 0` guard covers `1e400 → Infinity`, negatives, and non-numbers; null coalescing preserves explicit-null and snake_case selection correctly.
- Probed adjacent behavior: an empty-string `refreshToken` passes the shape guard and reaches the consumer, but `importClaudeCodeOAuthToken` treats `!blob.refreshToken` as "no refresh token available", so it degrades gracefully — no defect. The asymmetry (access token requires non-empty; refresh does not) is acceptable given refresh is optional.

### Notes (non-blocking)
- `typecheck` could not be run to completion in the sandbox — it errors only on unbuilt workspace subpath exports (`@elizaos/core/client-public`, `@elizaos/shared/contracts/first-run-options`) referenced by pre-existing imports outside this diff; that is an environment/build-graph limitation, not a defect in this change.

No correctness, security, or missing-test gaps found. The parsing now enforces the untrusted-input boundary per J3/J4 and the tests exercise it through the real public consumers with temp `.credentials.json` files.

---
— [eliza-swarm]

nothingxnowhere (changes requested):
Reviewed at head `b8ecf26831`. The validation change is right, but the branch no longer merges and four of its own tests fail against current `develop`, so this needs a rebase with the positive expectations re-derived rather than a plain conflict resolution.

**The parsing fix holds.** Against the merge-base (`5463f6db84`) the new suite is `8 fail / 11 pass` with develop's `credentials.ts` and `19 pass` at head — the eight failures are exactly the new rejections (string / negative / object / overflowing expiry, numeric refresh token, and the three shadowed snake_case aliases), which matches the control already posted above. I traced the alias logic and agree with it: `filter(v !== undefined)` + `some(malformed)` rejects a blob where *any* present alias is bad instead of quietly preferring the camelCase one, and `Number.isFinite` + `>= 0` closes the `1e400` and negative cases.

**What changed underneath it.** `develop` now carries `f0a54007dd` (`fix(auth): keep CLI credential diagnostics read-only`, merged `2026-09-05T00:47Z`, after both approvals on this PR at `20:57Z` / `21:03Z` the day before). It rewrote the same region of `credentials.ts` (`15 insertions, 84 deletions` against your merge-base), and GitHub now reports this PR `CONFLICTING`. Re-running your suite with the PR's test file against *current* develop's `credentials.ts` gives `12 fail / 7 pass`: the same eight rejections plus four positive cases —

- `preserves a future camelCase expiry`, `keeps an absent expiry as an explicit unknown expiry`, `accepts explicit null refresh and expiry fields`: `expected "info" to be called with arguments: [ StringContaining{…} ]` — develop removed the `[auth] Imported OAuth token from Claude Code …` and `… detected via Claude Code CLI` log lines those cases assert on.
- `refreshes an expired snake_case credential`: `expected "vi.fn()" to be called once, but got 0 times` — the deferred discovery path no longer calls `refreshAnthropicToken`, consistent with that commit's read-only intent.

So the positive half of the suite pins behaviour that develop has since deliberately removed. After the rebase those four cases need new assertions that describe what the read-only discovery path now does for a valid blob (status row shape is probably the durable thing to pin; the log-line and refresh-call expectations are not), and the RED/GREEN numbers in the body need regenerating against the new base. The rejection half should carry over unchanged.

Requesting changes for the rebase and re-pin only; the validation logic itself I would approve as-is.

```json
{
  "schemaVersion": "2",
  "headSha": "b8ecf26831",
  "recommendation": "revise",
  "reproduced": true,
  "risks": "branch conflicts with develop; 4 of 19 new tests fail against develop's credentials.ts after f0a54007dd",
  "effortBand": "medium",
  "complexity": "moderate",
  "impact": "medium",
  "reviewLoad": "standard",
  "recommendedTier": "small",
  "recommendedThirds": 3,
  "workUnitId": "wu_elizaos_eliza_pr_30549",
  "valueRationale": "Ran the new suite at head (19 pass), against the merge-base (8 fail / 11 pass, confirming the rejections) and against current develop (12 fail / 7 pass), and traced the four extra failures to log lines and a refresh call removed by f0a54007dd, which also makes the PR CONFLICTING.",
  "usefulArtifacts": ["three-way run of credentials-claude-cli-boundary.test.ts: head, merge-base source, current develop source"],
  "commands": ["bunx vitest run src/credentials-claude-cli-boundary.test.ts   # head: 19 pass; merge-base credentials.ts: 8 failed | 11 passed; origin/develop credentials.ts: 12 failed | 7 passed"],
  "evidenceUrls": [],
  "summary": "Claude CLI credential validation is correct, but develop's f0a54007dd rewrote the same file after the approvals: the PR is now conflicting and four positive tests assert removed log lines and a removed refresh call, so it needs a rebase with those expectations re-derived."
}
```

— [nxn-claude-code]

lalalune (approved):
Reviewed the credential parser, both naming conventions, status/deferred-discovery callers, and the complete regression suite. Rebased onto current develop, preserving read-only Claude CLI discovery and replacing obsolete refresh/log assertions. Real file-boundary tests fail against the previous parser and pass with the fix. Full auth suite: 260 passing tests; typecheck/lint/format, install, and root verify (376 tasks plus audits) passed on 7577bd43a79eb1ce9ea6a18dd62e88a7863706d9. Useful and ready to merge once exact-head hosted checks pass.

PR #30551, opened 2026-09-04
fix(app-core): reject malformed Anthropic usage response shapes

### LINKED ISSUE #30550: fix(app-core): reject malformed Anthropic usage response shapes
**Describe the bug**

`pollAnthropicUsage` casts the authenticated provider JSON directly to `AnthropicUsagePayload`. Explicitly malformed `five_hour`, `seven_day`, or `limits` values can therefore be treated like absent optional fields and produce a fresh, healthy-looking snapshot instead of a failed probe. `AccountPool.refreshUsage()` persists that snapshot, marks the account `ok`, and replaces previously known limit/reset data.

**To Reproduce**

From the repository root on current `develop`:

```bash
bun -e 'import { pollAnthropicUsage } from "./packages/app-core/src/services/account-usage.ts"; const payload={five_hour:7,seven_day:[],limits:{kind:"weekly_all"}}; const fetchImpl=async()=>new Response(JSON.stringify(payload),{status:200,headers:{"content-type":"application/json"}}); console.log(JSON.stringify(await pollAnthropicUsage("redacted",fetchImpl)));'\n```\n\nActual result:\n\n```json\n{"refreshedAt":1788557670038}\n```\n\n**Expected behavior**\n\nReject any present malformed root, window, limit-array, limit-entry, or nested scope/model shape before producing a `UsageSnapshot`. Valid legacy flat, current nested, and current `limits[]` response shapes must remain supported. A failed probe must not be represented as a successful empty snapshot.\n\n**Screenshots / recording of the wrong behavior (required for anything visible)**\n\n- N/A - server-side provider parsing defect with no direct rendered UI.\n\n**Evidence / reproduction proof**\n\n- N/A - no visual behavior.\n- Backend reproduction output and exact command are inline above.\n- N/A - provider response parsing does not involve an LLM trajectory.\n- Domain artifact: the returned `UsageSnapshot` above contains only a new `refreshedAt`, demonstrating the malformed response was accepted as healthy.\n\n**Acceptance criteria**\n\n- The parser validates the JSON root and every present structured field before reading it.\n- Malformed structured fields reject with an actionable typed error rather than returning a healthy empty snapshot.\n- Existing valid flat, nested, nullable reset, and `limits[]` payloads remain accepted.\n- Boundary tests prove rejection through `pollAnthropicUsage`, plus a consumer test proves `AccountPool.refreshUsage()` does not overwrite account health/usage when parsing fails.\n\n**Additional context**\n\nLive duplicate search found no issue or PR for malformed Anthropic usage response handling, and no open PR currently touches `packages/app-core/src/services/account-usage.ts`.

### PR DESCRIPTION
**Pull request by wtfsayo**
Anthropic account usage now rejects malformed response objects with an explicit unavailable result instead of presenting invalid data as valid quota information. Optional windows, limits, and model scopes still accept `null` as absent, preserving other valid buckets and the previously supported API behavior.

Fixes #30550. Targets `develop`; rebased and installed with Bun 1.3.14 and Node 24.15.0.

Review found that the stricter parser also rejected legitimate nullable optional fields. Four regressions failed on the original implementation and pass with the compatibility repair. Both malformed JSON and invalid response shapes reach the accounts route's explicit failure response.

Validation for `37aa4722c4ec7a71007715ce5fd9e12176d4417f`:

- Usage parser/account-pool focused tests: 81 passed. Accounts-route focused tests: 18 passed.
- Full app-core suite: 4,477 passed, 25 skipped; script suites: 74 Node and 21 Bun tests passed.
- Full agent runner completed all 757 files. Six files initially failed because workspace build artifacts or the test-only optimized-prompt integrity key were absent. After supplying those prerequisites, all six complete files passed: inbox mute, CORS preflight, Android plugins, deferred watchdog, provider vault boot, and wallet cache lifecycle. These files are unchanged from the base; no assertions or timeouts were weakened.
- Final app-core and agent typecheck/lint passed, with no lint rewrites.

Evidence is deterministic parser, service, and route coverage. No live Anthropic credential was used, and no end-to-end provider availability claim is made. Screenshots, video, and model trajectories are N/A: this changes account-usage validation and its error translation, not a rendered view, model request, or prompt.

### DIFF
--- a/packages/agent/src/api/accounts-routes.test.ts
+++ b/packages/agent/src/api/accounts-routes.test.ts
@@ -6,7 +6,7 @@ import { EventEmitter } from "node:events";
 import { mkdtempSync, rmSync } from "node:fs";
 import { tmpdir } from "node:os";
 import path from "node:path";
-import { LINKED_ACCOUNT_PROVIDER_IDS } from "@elizaos/core";
+import { ElizaError, LINKED_ACCOUNT_PROVIDER_IDS } from "@elizaos/core";
 import { codingProviderDescriptorForProvider } from "@elizaos/shared";
 import { beforeEach, describe, expect, it, vi } from "vitest";
 
@@ -736,6 +736,50 @@ describe("accounts routes", () => {
       },
     });
   });
+
+  it.each(["anthropic_usage.invalid_shape", "anthropic_usage.invalid_json"])(
+    "does not fall back to a healthy inline probe after malformed subscription usage (%s)",
+    async (code) => {
+      const existing = {
+        ...linkedAccount,
+        providerId: "anthropic-subscription",
+        health: "rate-limited",
+        healthDetail: { until: Date.now() + 60_000 },
+        usage: { refreshedAt: 100, sessionPct: 91, weeklyPct: 72 },
+      };
+      fakes.poolAccounts = [existing];
+      fakes.pool.refreshUsage.mockRejectedValueOnce(
+        new ElizaError('Anthropic usage response field "limits" was invalid', {
+          code,
+          severity: "fatal",
+        }),
+      );
+      const fetchMock = vi.fn(async () =>
+        Promise.resolve(new Response("{}", { status: 200 })),
+      );
+      vi.stubGlobal("fetch", fetchMock);
+      try {
+        const refreshed = makeContext(
+          "POST",
+          "/api/accounts/anthropic-subscription/account-1/refresh-usage",
+        );
+
+        await handleAccountsRoutes(refreshed.ctx);
+
+        expect(refreshed.errorCalls).toEqual([
+          {
+            message: 'Anthropic usage response field "limits" was invalid',
+            status: 502,
+          },
+        ]);
+        expect(fetchMock).not.toHaveBeenCalled();
+        expect(fakes.pool.upsert).not.toHaveBeenCalled();
+        expect(fakes.poolAccounts).toEqual([existing]);
+      } finally {
+        vi.unstubAllGlobals();
+      }
+    },
+  );
 
   it("starts and controls OAuth flows and validates the status stream", async () => {
     const started = makeContext(
--- a/packages/agent/src/api/accounts-routes.ts
+++ b/packages/agent/src/api/accounts-routes.ts
@@ -1656,8 +1656,9 @@ async function handleRefreshUsage(
   // Drive the canonical `pollAnthropicUsage` / `pollCodexUsage` through
   // the pool — same singleton used by the runtime, so health flips and
   // usage snapshots are consistent across UI and inference paths. Falls
-  // back to an inline 1-token probe only if the pool throws (network
-  // failure to the provider's usage endpoint, etc.).
+  // back to an inline 1-token probe for transport failures. Invalid provider
+  // data is terminal because a successful model probe cannot reconstruct the
+  // missing usage fields or safely replace the existing snapshot.
   try {
     await pool.refreshUsage(accountId, accessToken, {
       providerId,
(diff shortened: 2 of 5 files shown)

### CODE REVIEW

**Review by ss251 — APPROVED** [review]
Reviewed head 464f3e88c40d7259f080e190f322b647686abfc4.

The risky part is `packages/app-core/src/services/account-usage.ts:parseAnthropicUsageLimits` and `packages/agent/src/api/accounts-routes.ts:1677`. I followed the boundary where a 200 with `{"five_hour":7,"seven_day":[],"limits":{"kind":"weekly_all"}}` was previously cast to `AnthropicUsagePayload`, letting `utilizationToPct(undefined)` fabricate usage and overwrite a `rate-limited` snapshot. At this head numeric `five_hour`, array `seven_day`, and non-array `limits` all throw `invalid_shape`, string `scope` and null `model` are rejected while `scope:null` is preserved for optional chaining, and `res.json()` `SyntaxError` becomes typed `invalid_json` with cause. The route maps only those two codes to 502 and skips the 1-token probe.

Recommendation: merge

— [muse-lane-ss251]

**Review by attentionhead — CHANGES_REQUESTED** [review]
Requesting changes on one asymmetry. The boundary hardening is the right shape and the container/leaf split is well judged — but `five_hour: null` and `seven_day: null` now hard-fail where `develop` degraded gracefully, and I do not think that is intended, because you already handle exactly that case for `scope`.

**The inconsistency.** `scope` is parsed with `parseNullableRecord`, so an explicit `null` is accepted. The two windows go through `parseOptionalRecord`, where `null` falls past the `undefined` check into `isRecord(null) === false` and throws. Measured against `develop`:

```
                                     develop                     this PR
five_hour: null            OK   weeklyPct=50            THROW  field "five_hour" was invalid
seven_day: null            OK   sessionPct=50           THROW  field "seven_day" was invalid
scope: null                OK                           OK
valid nested   (control)   OK   12 / 9                  OK     12 / 9
valid flat     (control)   OK   12 / 9                  OK     12 / 9
```

Both controls are identical on either side, so the rest of the change is behaving. On `develop` a null window is simply absent — optional chaining yields `undefined` and the flat field still supplies the other window's number. Under this PR the whole refresh throws, which per your own description means an HTTP 502 that declines to overwrite prior usage and does not mark the account healthy. So a payload that currently produces a usable snapshot would leave the account looking stale or unhealthy.

That `scope` gets `parseNullableRecord` is what makes me read this as an oversight rather than a decision: it shows the provider is already known to send `null` for absent structures, and a window that has no data is at least as natural a null as an absent scope.

**Suggested fix**, two lines in `parseAnthropicUsageWindow`:

```ts
const record = parseNullableRecord(value, field);
if (record === undefined || record === null) return undefined;
```

I applied that locally and re-ran both sets:

```
five_hour: null   -> OK  weeklyPct=50      (matches develop)
seven_day: null   -> OK  sessionPct=50     (matches develop)
controls          -> OK  12 / 9            (unchanged)

five_hour: 42     -> still rejected
five_hour: []     -> still rejected
limits: {}        -> still rejected
root: []          -> still rejected
```

So it restores the graceful path for `null` without weakening any of the rejections this PR exists to add.

**Two things I checked that are fine**, noting them so they do not look unreviewed:

- Widening the leaf fields to `unknown` initially read like the validation stops short of the values, but it does not matter: `utilizationToPct` requires `typeof === "number" && Number.isFinite`, `normalizeResetTimestamp` gates on number/string, and `display_name` is guarded by a `typeof === "string"` check. Every leaf already degrades to `undefined` rather than propagating. Typing them `unknown` makes the declarations honest about untrusted input instead of asserting `number` over data you do not control — that is an improvement, not a gap.
- Strict on containers, lenient on leaves is a coherent policy: a wrong container shape means the contract changed and is worth failing on, whereas a wrong scalar can safely become "no reading". Worth stating that rule in the module header so the next person does not "tighten" the leaves and turn recoverable gaps into 502s.

Nice change otherwise — the error codes are actionable and keeping the transport fallback separate from typed parse failures is the right call.

— [claude-code-eliza]

lalalune (approved):
Reviewed the parser, account-pool consumers, and accounts-route failure translation. The shape checks are useful; repaired nullable optional windows, limits, and model scopes so stricter validation does not break previously valid responses. Four regressions were red before the repair, and focused parser/pool plus route coverage is green. Full app-core checks, all agent files with the documented prerequisite reruns, final package typecheck/lint, and root verification are complete. Approved pending exact-head hosted checks.

Issue #30552, opened 2026-09-04
fix(ci): restore production deployment Environment secret bindings


lalalune:
Production Cloud CF run 33711875349 passed the production Telegram identity authorization job, then failed in the reusable API deployment because ELIZA_APP_TELEGRAM_BOT_TOKEN resolved blank. Both required production Environment secrets existed before that run. The caller explicitly maps reusable secrets but omits these two names, so the protected Environment values are unavailable at the release boundary.

Acceptance:
- Preserve Environment-owned Telegram authority and fail-closed credential validation while making the two names available to the called deployment job.
- Confirm develop is the staging source, staging-approval is its policy-only admission checkpoint, and main is the production source; align live deployment branch policies.
- Validate the workflow contract, run the repository gates, and inspect successful hosted staging and production releases with matching live API and Pages source provenance.

CLAIMING: deployment environment recovery and production reusable-secret binding repair.
CLAIMING LEVER: Cloud CF staging/production deployment workflows and GitHub deployment environment branch policies. Existing runtime secrets and approval reviewers are preserved.

lalalune:
Deployment recovery progress:

- `staging-approval` now admits only branch `develop`; `production` admits only branch `main`. Runtime secrets remain in their existing environments. No reviewer or secret values were changed.
- #30557 merged production reusable-secret bindings and corrected the production database audit source to `main` after all hosted checks passed.
- #30562 merged the existing desktop-model installation opt-out into Cloud-only jobs after the current Develop Full setup failed on an unrelated embedding download HTTP 429. All exact-head hosted checks passed.
- #30561 merged unconditional operator receipt output and a read-only, identifier-free lifecycle report. All exact-head hosted checks passed.
- Staging deployment [33929636716](https://github.com/elizaOS/eliza/actions/runs/33929636716) is running at `1c710ac6e68785c15905fba1399e8430dd162773` with the deployed renderer proof enabled. Canonical source, staging admission, and protected Telegram authority have passed.
- Previous staging run [33922779397](https://github.com/elizaOS/eliza/actions/runs/33922779397) failed its real browser proof on all three attempts: Personal identity remained Shared while the existing Dedicated activation returned only pending cutover responses. API/Pages/routing success is not being treated as full acceptance.
- #30564 addresses the separate server-lane race where the prompts package removes its compiled output while concurrent agent tests import it. Package tests, 23 previously affected agent consumer tests, and the combined 376-task repository gate passed locally; hosted source validation is still running.

[...]

PR #30557, opened 2026-09-04
fix(ci): align deployment branches and production secret bindings

lalalune:
Production Cloud CF run [33711875349](https://github.com/elizaOS/eliza/actions/runs/33711875349) validated the production Telegram credentials in its authorization job, then failed because the reusable API deploy saw those same Environment-owned credentials as blank. The caller omitted both names from its explicit reusable-secret map.

Bind the two names to literal empty caller values so the selected Environment can supply them inside the called job. Repository and organization credentials cannot substitute through these entries; receipt authority, live identity attestation, missing-value rejection, and atomic Worker deployment remain enforced. Align both production database audit source checks with main, exercising the actual shell admission for main, develop, feature refs, tags, and empty input. Update the existing workflow boundary regression and document `develop` → staging admission/staging and certified `main` → production.

Validation at `98475f193b285d035d8ae1b78101c557c94a3ee1` on `develop` base `d04078cd890ea25dff7297434c7c447cdc399d38`:
- Pinned Bun 1.3.14 and Node 24.15.0; `bun install` passed.
- `bun run verify`: 376/376 tasks passed, exit 0.
- Deployment workflow, environment, secret, admission, and certification tests: 113 passed, 0 failed.
- Actionlint 1.7.12, Biome, Markdown link validation, and `git diff --check` passed.
- Live GitHub policy readback: `staging-approval` accepts branch `develop`; `production` accepts branch `main`. No runtime secret values or reviewer settings were changed.
- Hosted staging and production release/readback remain required after merge and will be recorded in the issue.

UI screenshots/video and model trajectories: N/A for this workflow-only change. Hosted release receipts, deployment logs, and API/Pages provenance are the applicable acceptance evidence.

Refs #30552.

### DIFF
--- a/.github/workflows/README.md
+++ b/.github/workflows/README.md
@@ -365,13 +365,20 @@ pair only when it matches the protected `staging` GitHub Environment secret
 `TELEGRAM_IDENTITY_AUTHORITY_SHA256`, and requires both components to differ
 from production. The receipt is the lowercase SHA-256 of the framed bytes
 `elizaOS/eliza\0staging\0telegram-public-identity\0v1\0<ID>\0<lowercase-username>\n`.
+`staging-approval` is the policy-only admission checkpoint for `develop`, not a
+third runtime or a Git branch. The `staging` Environment owns staging runtime
+configuration; `production` deploys certified `main` trees. Deployment branch
+policies must allow `develop` for staging approval and `main` for production.
+
 The protected staging entry job reads the receipt directly and verifies that the
 existing Worker has both Telegram binding names before any release mutation. It
 emits only a safe staging preserve-authority result; the reusable release
-receives that result and the already-admitted public pair, never the receipt,
-bot token, or webhook secret. This keeps a repository or organization secret
-from substituting for Environment authority or crossing the reusable-workflow
-boundary. The `deploy-api` job preserves a staging binding whose GitHub source
+receives that result and the already-admitted public pair, never the receipt.
+The caller maps both Telegram credential names to literal empty strings. This
+enables GitHub to resolve the selected Environment's secrets inside the reusable
+job while keeping repository and organization values out of that boundary.
+Omitting a name prevents its Environment secret from reaching the called job.
+The `deploy-api` job preserves a staging binding whose GitHub source
 is blank, then verifies its name again after the atomic Worker deploy. For
 production, the already-approved `production` authorization job verifies the
 live bot token against the canonical public identity and the deploy hard-fails
--- a/.github/workflows/cloud-cf-deploy.yml
+++ b/.github/workflows/cloud-cf-deploy.yml
@@ -1027,11 +1027,10 @@ jobs:
       cancel-in-progress: false
       queue: max
     uses: ./.github/workflows/cloud-cf-release.yml
-    # Explicitly forward every caller-scope secret consumed by the reusable
-    # release except Telegram authority and runtime credentials. The entry
-    # workflow validates the protected receipt and passes only public identity
-    # plus a safe runtime-authority result; `inherit` could admit a repository
-    # or organization secret with the same name.
+    # Telegram credentials have empty caller bindings so GitHub exposes their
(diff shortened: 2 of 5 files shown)


ss251 (approved):
Reviewed head 36d185500943d821970cbbcd98db8dd434874081.

The risky part is `.github/workflows/cloud-cf-deploy.yml:1052-1053`, where `ELIZA_APP_TELEGRAM_BOT_TOKEN` and `ELIZA_APP_TELEGRAM_WEBHOOK_SECRET` change from omitted to literal `""`. I checked the two boundaries that decide this: omission means the reusable `cloud-cf-release.yml` `workflow_call` name has no caller binding so the selected Environment secret never reaches the called job and production resolves blank, while `${{ secrets.X }}` would forward a caller repo/org value across the reusable boundary. The empty literal is the third state that exposes the name without forwarding a value, and `TELEGRAM_IDENTITY_AUTHORITY_SHA256` correctly stays unbound so the receipt never crosses.

`packages/scripts/__tests__/homepage-deploy-workflow.test.ts:536` pins exactly this contract. Verified by reading the diff at this head and following both secret maps.

Recommendation: merge

— [muse-lane-ss251]

attentionhead (approved):
Approving. Note your head moved from `36d1855` to `98475f1` while I was reviewing, so this covers the current head including the Railway audit change.

**Both workflow changes are genuinely test-locked.** Run under `bun test`, each against its own `develop` YAML as a control:

```
homepage-deploy-workflow    head: 19 pass, 0 fail   |  develop yaml: 18 pass, 1 fail
  -> fails on "keeps Telegram authority and runtime credentials in protected
     Environment jobs" (line 513), the new expect(callerSecretMap[name]).toBe("")

railway-database-authority  head:  9 pass, 0 fail   |  develop yaml:  3 pass, 6 fail
```

So neither assertion merely restates the YAML — both fail against the old files. `TELEGRAM_IDENTITY_AUTHORITY_SHA256` also stays excluded from the caller map and from `workflow_call.secrets`, so the receipt does not cross the reusable boundary.

The Railway change (`refs/heads/develop` → `refs/heads/main` for the production database authority audit, in both the source guard and `--canonical-ref`) is coherent with the branch policy this PR documents: auditing production from `develop` would be auditing an uncertified tree. Good that both the guard and the freshness check moved together — changing only one would have left a gap that still passed.

**One thing I would act on: this PR still carries an unrelated change that duplicates your own open #30489, divergently.** `app-verification.integration.test.ts` — replacing the `npx --yes -p typescript@5.6.2` fixture with the pinned compiler via `import.meta.resolve` — is the whole content of #30489 (*test(app-control): use pinned compiler fixture*), but the two have drifted:

- the comment wording differs;
- #30489 has an extra hunk at `@@ -118,9 +118,8 @@` that is absent here.

#30489 is already `DIRTY`, and merging this guarantees it stays that way, with whichever lands second needing rework. Either drop the file here so #30489 remains its owner, or close #30489 after confirming its extra hunk is not lost. The change itself is right — a deterministic fixture should not be network-bound — it just does not belong in a production-secrets PR you might want to revert on its own.

**An honest limit on what I could verify.** The mechanism claim — that an empty caller binding lets GitHub resolve the selected Environment's secret inside the called job, while omitting the name makes even an existing Environment secret resolve blank — is GitHub runtime behavior. The tests pin YAML shape, not resolution semantics, so nothing local can confirm it. Your summary already defers hosted staging and production readback to after merge, which is the right call; I am recording it explicitly so the fix for run 33711875349 is not treated as closed by a green CI. Worth noting the downside is bounded either way: if the semantics turn out inverted, the result is another blank-credential failure, not a credential leak, since no repository or organization value is forwarded in either arrangement.

The README additions are the most durable part here — stating the `develop` → staging-approval and `main` → production policies, and that omitting a name blocks its Environment secret, is exactly the thing the next person would otherwise rediscover the hard way.

— [claude-code-eliza]

PR #30561, opened 2026-09-04
fix(cloud): expose staging Dedicated lifecycle diagnostics

lalalune:
Staging release 33922779397 reaches the deployed app but repeatedly stalls while activating the selected Dedicated agent. The read-only selection-review run 33925510243 succeeded without emitting a result because its CLI receipt used an application logger that suppresses info output by default.

Write operator receipts and expected decision results to stdout. Before selection admission, inspect the authenticated smoke owner's primary database rows and emit lifecycle status, closed failure categories, and presence flags. This keeps failed or already-bound activation attempts diagnosable even when selection review cannot proceed. Raw errors, account and agent identifiers, addresses, credentials, job payloads, and message content remain outside the report. Receipt execution still requires the exact deployment, database identity, prior digest, and reviewed confirmation.

Validation:
- Pinned Bun 1.3.14 and Node 24.15.0; fresh `bun install` passed.
- Operator and workflow tests: 18 passed, including privacy rejection, missing-selection diagnostics, stale approval, source mismatch, and no-compute mutation checks.
- `bun run verify`, Biome, and `git diff --check` passed.
- Hosted preview readback is required after merging and deploying the exact source to staging; this change diagnoses the stalled activation and does not claim to repair its runtime cause.

UI captures and live-model trajectories: N/A for this operator CLI change. Hosted operator output is the applicable acceptance evidence.

Refs #30552.

### DIFF
--- a/packages/cloud/scripts/admin/personal-dedicated-rereview-staging.test.ts
+++ b/packages/cloud/scripts/admin/personal-dedicated-rereview-staging.test.ts
@@ -2,6 +2,7 @@
 
 import { describe, expect, test } from "bun:test";
 import {
+  diagnoseRereviewTarget,
   PersonalDedicatedRereviewOperatorError,
   previewDecisionEvidence,
   type RereviewOperatorConfig,
@@ -34,6 +35,13 @@ const snapshot = {
   agentDigest: "c".repeat(64),
   jobCount: 0,
   jobDigest: "d".repeat(64),
+  selectedTarget: diagnoseRereviewTarget({
+    status: "stopped",
+    database_status: "ready",
+    error_message: null,
+    sandbox_id: null,
+    bridge_url: null,
+  }),
 };
 
 function config(mode: "preview" | "execute"): RereviewOperatorConfig {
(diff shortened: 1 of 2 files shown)


ss251 (approved):
Reviewed head c5fe1f30a6fe1498a78372e91efc05933bfcf40a.

The risky part is `packages/cloud/scripts/admin/personal-dedicated-rereview-staging.ts:60` diagnoseRereviewTarget. I checked the leak boundary it must hold: it returns only `status`/`databaseStatus`, classified `provisionFailure`, and `hasContainer`/`hasBridge` booleans, never raw `sandbox_id`/`bridge_url`/`error_message`. The test at `personal-dedicated-rereview-staging.test.ts:96` pins it: input `error_message` containing `https://private-host.invalid/private-agent` + `sandbox_id=private-container-id` -> `provisionFailure=ingress_headscale_not_configured` with `JSON.stringify(result)` containing neither secret, plus invalid `status` -> `selected_target_status_invalid`.

Verified by reading the diff at this head and following both call sites in snapshot and lifecycle report.

Recommendation: merge

— [muse-lane-ss251]

attentionhead (approved):
Approving. For a change that writes database rows to stdout, the claim worth checking is the redaction one, so that is where I spent the review — and it holds structurally *and* under mutation.

**Nothing sensitive reaches the report.** `diagnoseRereviewTarget` receives `error_message`, `sandbox_id`, and `bridge_url`, and returns none of them:

```ts
interface RereviewTargetDiagnostic {
  status: string;
  databaseStatus: string;
  provisionFailure: ManagedDedicatedProvisionFailureCode;
  hasContainer: boolean;   // Boolean(target.sandbox_id)
  hasBridge: boolean;      // Boolean(target.bridge_url)
}
```

Locators collapse to presence flags, `error_message` goes through `classifyManagedDedicatedProvisionFailure` and comes back as a code rather than text, and the agent entries add only `updatedAt` plus three booleans — `agent.id` is used to correlate rows but never emitted. Job entries are the same shape: allowlisted `status`, a failure code, `attempts`, `retryableRequeues`, `updatedAt`.

**The status allowlists are the part I liked most.** Both `diagnoseRereviewTarget` and the job mapper reject unrecognised values by throwing rather than passing them through:

```ts
throw new PersonalDedicatedRereviewOperatorError("diagnostic_job_status_invalid");
```

That closes the obvious hole in a "we only emit vocabulary" design — a future migration introducing a new status, or a corrupted row, cannot smuggle arbitrary text into stdout as an unexpected enum value. Fail-closed is the right default for a report that ends up in CI logs.

**And the guard is enforced, not incidental.** Tests are 13/13 at this head. To check the privacy assertions actually have teeth rather than passing by construction, I mutated `diagnoseRereviewTarget` to leak both locators:

```ts
+ sandboxId: target.sandbox_id,
+ bridgeUrl: target.bridge_url,
```

Result: `12 pass, 1 fail`. The suite catches the regression, so the `not.toContain(privateLocator)` assertions are load-bearing and a future contributor cannot widen the diagnostic without tripping them. I reverted the mutation.

Two notes, neither blocking:

- `updatedAt` timestamps are emitted. They are not in your excluded list and are clearly needed to diagnose a stall, so this is fine — worth being aware that timing metadata is the one non-categorical field in the report.
- Your framing that this "diagnoses the stalled activation and does not claim to repair its runtime cause" is the right call, and I would keep that sentence in the merge commit. The hosted readback is still what closes #30552; this makes the next failure legible rather than silent.

One orientation note for other reviewers: this PR is stacked on #30557, so `git diff develop...` also shows that PR's workflow and secret-binding changes. Its own contribution is the two `personal-dedicated-rereview-staging` files. I reviewed #30557 separately.

Moving the receipt off the suppressed application logger and onto stdout is the actual fix for the silent run, and it is good that the decision result goes to the same place — a read-only review that succeeds without emitting anything is exactly the failure this started from.

— [claude-code-eliza]

PR #30562, opened 2026-09-04
fix(ci): avoid desktop model downloads in Cloud jobs

lalalune:
Cloud unit setup in Develop Full run 33926260898 failed before tests started because root postinstall downloaded the desktop embedding model and received HTTP 429 on every attempt. Cloud tests and the deployed staging renderer use fixture or remote inference, so that download is unrelated to their acceptance path.

Pass the existing explicit fused-inference setup opt-out during Cloud test dependency installation and deployed-renderer workspace setup. Dependency installation, linked workspace builds, browser tests, deployment authority, and desktop/device installer coverage retain their existing gates.

Validation: pinned Bun 1.3.14 and Node 24.15.0; `bun install` and `bun run verify` passed; installer behavior tests 8/8 and Cloud workflow tests 11/11 passed; Actionlint and `git diff --check` passed. Hosted Cloud setup and release verification remain required after merge.

UI screenshots and model trajectories: N/A for this CI setup change. Hosted setup logs are the applicable evidence.

Refs #30552 and #30440.

### DIFF
--- a/.github/actions/cloud-setup-test-env/action.yml
+++ b/.github/actions/cloud-setup-test-env/action.yml
@@ -94,6 +94,10 @@ runs:
     - name: Install dependencies
       shell: bash
       working-directory: packages/cloud/shared
+      env:
+        # Cloud tests use remote or fixture inference; desktop model setup is
+        # covered by the desktop/device lanes and must not gate this install.
+        ELIZA_SKIP_FUSED_INFERENCE_SETUP: "1"
       run: |
         for attempt in 1 2 3; do
           if [ "$attempt" -eq 3 ]; then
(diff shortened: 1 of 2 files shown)


ss251 (approved):
Reviewed head 72329bd5714971ed2d14b3d09b11ee7b8abcb669.

The risky part is `.github/actions/cloud-setup-test-env/action.yml:97` and `.github/workflows/cloud-cf-release.yml:3304`, where `ELIZA_SKIP_FUSED_INFERENCE_SETUP="1"` is injected to skip desktop model downloads. I checked the scoping that usually breaks this pattern: both `env:` blocks sit on the exact install step — the `packages/cloud/shared` retry loop and the `uses: setup-bun-workspace` browser-proof step — not at job level, so desktop/device lanes keep their model setup while these Cloud paths skip it. The comments bound intent to remote/fixture inference, so a future reader will not copy the skip elsewhere.

No blocking concerns; centralizing the flag in one shared action would be non-blocking polish.

Verified by reading the diff at this head against the two adjacent install steps.

Recommendation: merge

— [muse-lane-ss251]

PR #30564, opened 2026-09-04
fix(ci): serialize prompt artifact consumer tests

lalalune:
Develop Full run 33920498044 started the prompts package test at 21:45:45 while agent tests were still importing it. The package consumer tests remove the real workspace `dist/` directory to validate clean resolution, and the agent suites beginning during that interval failed to resolve `@elizaos/prompts`.

Mark the prompts test task serial through its existing `elizaos.scripts.testSerial` metadata and document the shared artifact constraint. The runner completes the concurrent pool before starting serialized tasks, so the package's complete build, clean-resolution, and published-consumer tests remain enabled without deleting artifacts beneath another suite.

Validation:
- Prompts package tests 25/25, typecheck, and lint passed.
- Previously affected agent consumer suites: 23/23 passed across three files.
- Task-pool tests 32/32 passed. The live server plan classifies the prompts task as serial with zero parallel tasks.
- Pinned Bun 1.3.14 and Node 24.15.0; `bun install`, root `bun run verify`, and `git diff --check` passed.
- Hosted server validation is required after merge.

UI captures and live-model trajectories: N/A. This changes test scheduling only; authored prompts and runtime code are unchanged.

Refs #30552 and #30440.

### DIFF
--- a/packages/prompts/README.md
+++ b/packages/prompts/README.md
@@ -56,6 +56,10 @@ The generated publish manifest rewrites every source-facing condition to
 compiled JavaScript and declarations in `dist/`, so the release tarball never
 publishes TypeScript source as runtime code.
 
+The repository runs this package's tests serially because they rebuild and
+temporarily remove `dist/` while checking consumer resolution. Concurrent
+workspace tests may still be importing that compiled package.
+
 ## Usage
 
 Runtime code imports the templates through `@elizaos/core`, which re-exports them and provides `composePrompt` to fill the `{{...}}` placeholders:
--- a/packages/prompts/package.json
+++ b/packages/prompts/package.json
@@ -70,6 +70,7 @@
         "order": 2,
         "script": "build:package"
       },
+      "testSerial": true,
       "testLanes": [
         "server"
       ]


ss251 (approved):
Reviewed head 6f59d6c3250381d7674052c1f5b54974f193aca2.

The risky part is `packages/prompts/package.json:73`, where `testSerial: true` is added alongside `testLanes: ["server"]`. I checked the failure window described in `packages/prompts/README.md:59-61`: these tests rebuild then temporarily remove `dist/` to verify consumer resolution, so any concurrent workspace job importing the compiled `@elizaos/prompts` package would hit a missing-file / unresolved-import error mid-run. Serializing just this package's tests closes that exact race without changing the `order: 2` build ordering or lane assignment, which the diff leaves untouched.

No blocking concerns; non-blocking nit would be naming the affected consumer-resolution test in the README note.

Verified by reading the two-hunk diff at this head and cross-checking the flag placement against the adjacent build/lane config.

Recommendation: merge

— [muse-lane-ss251]

attentionhead (approved):
Approving. I traced the scheduling claim end to end rather than trusting the metadata flag, and it holds.

**The ordering guarantee is real.** `run-all-tests.mjs:1706` partitions, then awaits the whole concurrent pool before touching the serial bucket:

```js
const poolResults = await runPool(parallel, ...);   // drains completely
for (const task of serial) { await runTask(task, { stream: true }); }
```

So a serialized task cannot start until every parallel task has finished. That is exactly the property this fix needs: the prompts consumer tests can only remove `dist/` after the suites importing `@elizaos/prompts` are done.

**And the flag lands where it needs to.** Partitioning the two relevant tasks directly:

```
lane=pr       parallel=[agent]  serial=[prompts]
lane=server   parallel=[]       serial=[prompts, agent]
lane=develop  parallel=[]       serial=[prompts, agent]
```

In the `pr` lane `prompts` moves to the serial bucket while `agent` stays parallel, which is precisely the interleaving that broke run 33920498044. Task-pool tests are 32/32 at this head.

I went looking for a way this could be a no-op, since `isParallelSafeTask` returns `false` whenever `lane !== "pr"` — outside that lane every task is already serial, so `testSerial` changes nothing there. That turns out to argue *for* the fix rather than against it: the failure requires two tasks running concurrently, concurrency exists only in the `pr` lane, so the `pr` lane is necessarily where the reported race occurred and is the only place the flag needs to bite. Worth knowing that the flag is inert in the other lanes, though — it is not a general "this package is dangerous" marker, and someone later adding parallelism to another lane would need to re-check that assumption.

Using the existing `elizaos.scripts.testSerial` metadata rather than re-adding a hardcoded package list is the right call, given `script-metadata.mjs:90` explicitly documents that field as the replacement for the old `SERIALIZE_PACKAGES` set. The README note is doing real work too — the constraint is non-obvious from the test code alone, and "these tests delete a shared build artifact" is exactly what a future contributor needs to know before flipping this back.

One orientation note for other reviewers: this PR is stacked on #30557, so `git diff develop...` shows that PR's workflow and secret-binding changes as well. Its own contribution is just the two prompts files (+5/-0). I reviewed #30557 separately.

Narrow, uses the mechanism already in place, and documents the constraint at the source rather than only in the commit message.

— [claude-code-eliza]

PR #30566, opened 2026-09-04
fix(auth): validate Anthropic OAuth token responses

### LINKED ISSUE #30563: fix(auth): reject malformed Anthropic OAuth token responses
## Describe the bug

The Anthropic authorization-code exchange casts every successful HTTP response directly to its expected token shape. A `200` response can therefore return credentials whose `access` and `refresh` values are `undefined`, or whose expiry is invalid. The refresh path performs only truthiness/type checks after the same unchecked cast, so it can also accept non-string tokens and non-finite or negative lifetimes.

These values cross the shared auth boundary and can reach subscription login, credential persistence, and later refresh logic as if the exchange succeeded.

## Reproduction

On current `origin/develop` (`f3017932d3501c1e13b78fa248c6f074dd5e5c88`), stub the successful token response and call the real exported exchange:

```bash
bun -e 'globalThis.fetch = async () => new Response(JSON.stringify({ expires_in: "3600" }), { status: 200, headers: { "content-type": "application/json" } }); const { exchangeAnthropicAuthorizationCode } = await import("./packages/auth/src/vendor/pi-oauth/anthropic-login.ts"); const credentials = await exchangeAnthropicAuthorizationCode("code#state"); console.log(JSON.stringify(credentials)); console.log(`access=${String(credentials.access)} refresh=${String(credentials.refresh)} expiresFinite=${Number.isFinite(credentials.expires)}`);'
```

Observed:

```text
{"expires":1788566348021}
access=undefined refresh=undefined expiresFinite=true
```

The string lifetime is silently coerced, while both required secrets are absent. Similar successful responses can admit numeric tokens, `Infinity`, `NaN`, zero/negative lifetimes, or whitespace-only secrets.

## Expected behavior

A successful HTTP status is not sufficient proof of a valid token exchange. Validate the unknown JSON response before constructing credentials:

- authorization-code exchange requires non-empty string access and refresh tokens plus a positive finite numeric lifetime;
- refresh requires a non-empty string access token and positive finite numeric lifetime, while preserving the existing refresh token only when the response genuinely omits its optional rotated token;
- malformed JSON, a non-object root, and invalid present fields fail explicitly without returning or persisting partial credentials;
- valid exchange and RFC 6749 refresh-rotation behavior remain unchanged.

## Scope / evidence

Non-UI auth boundary in `packages/auth/src/vendor/pi-oauth/anthropic-login.ts`, with direct consumers in the shared Anthropic flow and the public subscription exchange route. No schema, migration, deployment, payment, native, or model-call change.

Acceptance:

- Add adversarial response-shape coverage for both exchange and refresh.
- Prove malformed successful responses cannot produce credentials.
- Preserve callback state validation and valid token exchange behavior.
- Pass auth package tests, typecheck, lint, and root verification.

No matching open issue or pull request was found for Anthropic token-response shape validation.

wtfsayo:
Closes #30563.

Validate successful Anthropic OAuth responses before credentials can be persisted. Invalid responses produce typed errors while valid exchange and refresh behavior, including null/omitted refresh-token rotation, is preserved.

Reviewed both token exchange and refresh, their callers and persistence boundary, all added tests, and the earlier null-rotation finding. The refresh response keeps the existing token when rotation is omitted or null; exchange still requires a refresh token. Malformed tokens, non-positive lifetimes, numeric overflow, and invalid JSON reject before constructing credentials.

On `3640209cf0a42539c9999046ac82f49ad5b1b6ec`, rebased onto `10adfdb85226639480c544f6522954f0d44eb1f7`, Bun 1.3.14 / Node 24.15.0:
- `bun install`: passed.
- Full auth suite: 19 files, 234 tests passed.
- Auth typecheck, lint and format checks: passed.

Real Response bodies cover provider parsing; no live credential exchange was needed or claimed. UI/native captures and model trajectories are N/A for this parsing-only change. No further implementation changes were needed. Hosted checks must pass at this head before merge.

### DIFF
--- a/packages/auth/src/vendor/pi-oauth/anthropic-login.test.ts
+++ b/packages/auth/src/vendor/pi-oauth/anthropic-login.test.ts
@@ -59,6 +59,17 @@ describe("refreshAnthropicToken", () => {
     expect(creds.access).toBe("at-new");
   });
 
+  it("keeps the current refresh token when refresh_token is null", async () => {
+    mockTokenResponse({
+      access_token: "at-new",
+      refresh_token: null,
+      expires_in: 3600,
+    });
+    const creds = await refreshAnthropicToken("rt-old");
+    expect(creds.refresh).toBe("rt-old");
+    expect(creds.access).toBe("at-new");
+  });
+
   it("throws when the response lacks an access token instead of persisting a broken blob", async () => {
     mockTokenResponse({ refresh_token: "rt-new", expires_in: 3600 });
     await expect(refreshAnthropicToken("rt-old")).rejects.toThrow(
--- a/packages/auth/src/vendor/pi-oauth/anthropic-login.test.ts
+++ b/packages/auth/src/vendor/pi-oauth/anthropic-login.test.ts
@@ -1,5 +1,6 @@
-/** Verifies Anthropic OAuth exchange, refresh rotation, and callback state validation at the HTTP boundary. */
+/** Verifies Anthropic OAuth exchange, token-response validation, refresh rotation, and callback state validation at the HTTP boundary. */
 
+import { ElizaError } from "@elizaos/core";
 import { afterEach, describe, expect, it, vi } from "vitest";
 import {
   exchangeAnthropicAuthorizationCode,
@@ -10,12 +11,28 @@ import {
 function mockTokenResponse(body: unknown, ok = true): void {
   vi.stubGlobal(
     "fetch",
-    vi.fn(async () => ({
-      ok,
-      json: async () => body,
-      text: async () => JSON.stringify(body),
-    })),
+    vi.fn(
+      async () =>
+        new Response(JSON.stringify(body), {
+          status: ok ? 200 : 400,
+          headers: { "content-type": "application/json" },
+        }),
+    ),
   );
+}
+
+async function expectTokenError(
+  operation: Promise<unknown>,
+  code: string,
+): Promise<void> {
+  try {
+    await operation;
+    throw new Error("Expected Anthropic token operation to reject");
+  } catch (error) {
+    expect(error).toBeInstanceOf(ElizaError);
+    if (!(error instanceof ElizaError)) throw error;
+    expect(error.code).toBe(code);
+  }
 }
 
 describe("refreshAnthropicToken", () => {
@@ -46,6 +63,61 @@ describe("refreshAnthropicToken", () => {
     mockTokenResponse({ refresh_token: "rt-new", expires_in: 3600 });
     await expect(refreshAnthropicToken("rt-old")).rejects.toThrow(
       /missing access_token/,
+    );
+  });
+
+  it.each([
+    ["non-object root", null],
+    ["numeric access token", { access_token: 7, expires_in: 3600 }],
+    ["blank access token", { access_token: "   ", expires_in: 3600 }],
+    [
+      "invalid rotated refresh token",
+      { access_token: "at-new", refresh_token: 7, expires_in: 3600 },
+    ],
+    ["zero lifetime", { access_token: "at-new", expires_in: 0 }],
+    ["negative lifetime", { access_token: "at-new", expires_in: -1 }],
+    ["null lifetime", { access_token: "at-new", expires_in: null }],
+    ["overflowing lifetime", { access_token: "at-new", expires_in: 1e308 }],
+    ["string lifetime", { access_token: "at-new", expires_in: "3600" }],
+  ])("rejects a successful response with %s", async (_name, body) => {
+    mockTokenResponse(body);
+    await expectTokenError(
+      refreshAnthropicToken("rt-old"),
+      "anthropic_oauth.token_invalid_shape",
+    );
+  });
+
+  it("wraps malformed successful JSON as a typed token response error", async () => {
+    vi.stubGlobal(
+      "fetch",
+      vi.fn(
+        async () =>
+          new Response("{not-json", {
+            status: 200,
+            headers: { "content-type": "application/json" },
+          }),
+      ),
+    );
+    await expectTokenError(
+      refreshAnthropicToken("rt-old"),
+      "anthropic_oauth.token_invalid_json",
+    );
+  });
+
+  it("rejects a JSON lifetime that parses beyond the finite number range", async () => {
+    vi.stubGlobal(
+      "fetch",
+      vi.fn(
+        async () =>
+          new Response('{"access_token":"at-new","expires_in":1e400}', {
+            status: 200,
+            headers: { "content-type": "application/json" },
+          }),
+      ),
+    );
+    await expectTokenError(
+      refreshAnthropicToken("rt-old"),
+      "anthropic_oauth.token_invalid_shape",
     );
   });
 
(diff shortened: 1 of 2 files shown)


ss251 (approved):
Reviewed head 4deb3d1cdc6642bc94e24cff69f7324e1e83e82e.

The risky part is `packages/auth/src/vendor/pi-oauth/anthropic-login.ts` `parseAnthropicTokenResponse` expires_in handling. I followed the two finite checks that usually break this: `typeof expiresIn !== "number" || !Number.isFinite(expiresIn) || expiresIn <= 0` rejects string/-1/0/null, while `Date.now()+expiresIn*1000` with `Number.isFinite(expiresAt)` catches `1e308` which is itself finite but overflows when scaled to ms. The `1e400` JSON case never reaches the second check because `JSON.parse` yields `Infinity`, failing the first `isFinite`. Both map to `anthropic_oauth.token_invalid_shape`, matching the new `it.each` cases, while `{not-json` maps to `token_invalid_json` with cause preserved. Verified by reading the diff at this head and tracing the added overflow tests.

Recommendation: merge

— [muse-lane-ss251]

attentionhead (changes requested):
Requesting changes on one case. The validation is well built — 29/29 at this head, the exchange/refresh split is right, and the overflow guard is a genuinely good catch — but an explicit `refresh_token: null` in a refresh response now fails the refresh where `develop` handled it gracefully.

**Measured against `develop`:**

```
                                        develop              this PR
refresh_token OMITTED (RFC 6749 §6)     ok, keeps rt-old     ok, keeps rt-old
refresh_token: null                     ok, keeps rt-old     THROW token_invalid_shape
refresh_token rotated (control)         ok, rt-new           ok, rt-new
```

Only the `null` row changes. The guard is `refreshToken !== undefined && !isNonBlankString(refreshToken)`, so `null` clears the `undefined` check and then fails the string test.

**Why this one matters more than the equivalent in an ingest path.** You already treat *omission* as "the provider did not rotate; keep the current token" — which is the correct RFC 6749 §6 reading, and your test says so explicitly. A `null` carries exactly the same information: there is no rotated token here. But instead of keeping the working refresh token, this throws, and a failed refresh is user-visible — the session cannot be renewed and the user is pushed back through the OAuth flow. Rejecting a value that can never be a usable credential costs a re-authentication and buys nothing, because there is no scenario where `null` should have been persisted as a token.

**Suggested fix**, one line, and I verified it:

```ts
mode === "refresh" &&
refreshToken !== undefined &&
refreshToken !== null &&
!isNonBlankString(refreshToken)
```

With that applied:

```
refresh_token: null   -> ok, keeps rt-old   (matches develop and the omission path)
controls              -> unchanged

refresh_token: 7      -> still rejected
refresh_token: " "    -> still rejected
refresh_token: {}     -> still rejected
```

So it restores the graceful path without weakening a single rejection this PR adds. Worth a test case alongside the existing omission one, since `null` is currently the only refresh-mode shape with no coverage — the suite has `refresh_token: 7` but nothing for `null`.

This is the same shape as the null-window issue I raised on your #30551, so it may be worth a consistent rule across these boundary validators: an explicit `null` on an *optional* field means "absent", while a wrong *type* means "malformed". Applying that uniformly would have covered both cases up front.

**The rest is good, and two parts are better than they need to be:**

- Requiring `refresh_token` on `exchange` but not on `refresh` is exactly right, and encoding it in the overloads rather than a runtime flag means callers get the stricter type where it applies.
- The overflow guard is the standout. Validating `expires_in` as positive and finite, and then separately checking `Date.now() + expires_in * 1000` is still finite, catches the case where a plausible-looking large lifetime silently produces a non-finite absolute expiry. Your test at `expires_in: 1e308` pins it.
- Error context is `{ mode, reason }` with static reason strings, so no token material reaches the error. One thing to keep an eye on: the `token_invalid_json` path attaches the parse `cause`, and a JSON syntax error can quote a fragment of the offending body. That body is not a valid token response by definition, so the exposure is slight — just be careful that this `cause` never gets logged verbatim somewhere less careful than here.

— [claude-code-eliza]

wtfsayo:
Addressed the requested `refresh_token: null` compatibility case in `01654ee7e6f0e229c415d65119f965a5e1402a06`.

I reproduced it first with the repository Vitest runner: the new case was the sole failure (1 failed, 29 passed). Refresh mode now normalizes null exactly like omission and retains the current token; exchange mode remains strict, and numeric/blank/object values remain invalid.

Post-fix evidence:
- focused helper: 30/30
- helper + wrapper reviewer run: 32/32
- complete auth suite: 231/231
- auth typecheck, lint, format, build: pass
- two independent re-reviews: zero findings
- rebased onto current `origin/develop` with patch identity preserved by `range-diff`
- post-sync root `bun run verify`: pass

Please re-review the current head when convenient.

attentionhead (approved):
Re-reviewing at `01654ee7` after my `CHANGES_REQUESTED` at `4deb3d1c`. The null-rotation defect I reported is fixed correctly, and I verified the fix rather than taking the commit title for it.

**The fix holds.** `parseAnthropicTokenResponse` now carries `refreshToken !== null &&` in the refresh-mode guard, so a `refresh_token: null` payload falls through to `data.refreshToken ?? refreshToken` and preserves the stored token instead of throwing `token_invalid_shape` mid-refresh. The added test pins exactly that.

I mutation-checked it: deleting the new `refreshToken !== null &&` line and re-running gives `1 failed | 29 passed`, and the one failure is `keeps the current refresh token when refresh_token is null`. So the regression test genuinely covers the guard rather than passing incidentally. As shipped, `anthropic-login.test.ts` is 30/30, the whole `vendor/pi-oauth` directory is 41/41, and `tsc --noEmit` is clean.

One methodology note in case it saves someone time: this file must run under vitest, not `bun test`. Bun's `vi` shim has no `stubGlobal`, so `bun test` reports all 30 as failures with `TypeError: vi.stubGlobal is not a function` — a runner artifact, not a real failure. I hit exactly that before switching runners.

**On the rest of the change**, which I read fresh at this head rather than assuming it was unchanged:

The `exchange`/`refresh` overload pair asserts more than the implementation signature does — the `"exchange"` overload promises a required `refreshToken` while the implementation returns the optional-refresh shape. That is sound here only because the `mode === "exchange" && !isNonBlankString(refreshToken)` throw runs before the return, so the `isNonBlankString(refreshToken)` branch is the only reachable exit for exchange. Worth knowing the invariant is enforced by that guard and not by the types, since deleting the guard would leave the overload silently lying.

Tightening `expires_in` from `typeof !== "number"` to also reject non-finite and `<= 0` is a real behavior change worth stating plainly: a provider returning `expires_in: 0` now fails the refresh instead of minting an already-expired credential. That is the right call — a credential that expires at or before `Date.now()` is not usable, and failing loudly beats a silent refresh loop. The explicit `Number.isFinite(expiresAt)` check after the multiplication is a nice touch, since a huge `expires_in` overflows to `Infinity` and would otherwise persist as a never-expiring token.

**One follow-up, explicitly out of scope for this PR** — noting it so it isn't lost, not asking for it here. The sibling `openai-codex-login.ts` still uses the old shape: an unwrapped `response.json()` with an `as {...}` cast and truthiness-only validation, so a non-string `refresh_token` is persisted verbatim and `expires_in: 0` or a negative is accepted. Notably it does *not* have the null bug, because it uses `json.refresh_token ?? refreshToken` with no prior validation throw — which is precisely why the Anthropic regression only appeared once validation was added. Same hardening would apply cleanly there.

Approving. Scoped correctly, the defect is fixed, and the test proves it.

— [claude-code-eliza]

nothingxnowhere (approved):
Reviewed at head `01654ee7e6` against `develop` `1045297cca`. The null-rotation thread above is already closed with a mutant, so I spent the review on two things it did not cover: whether the base has drifted underneath this branch (that is what just sank a sibling auth PR), and whether the suite's split between new rejections and preserved behaviour is what the description claims.

**No base drift.** `merge-base` is `ff374d0f29`, and `git diff ff374d0f29..origin/develop -- packages/auth/src/vendor/pi-oauth/` is empty, so nothing on develop has moved either touched file; GitHub agrees (`MERGEABLE`). Worth stating because `packages/auth/src/credentials.ts` did move under #30549 after its approvals and that PR is now conflicting — this one is not exposed to that.

**The suite splits exactly as described.** At head: `30 pass`. With the PR's tests kept and `anthropic-login.ts` reverted to develop: `20 fail / 10 pass`. The 20 failures are precisely the new rejections — the nine refresh-mode shapes, the eight exchange-mode shapes, the two malformed-JSON wrappers, and the `1e400` lifetime — while all 10 that pass on develop are the behaviours that must not change: the rotated-token control, the omitted `refresh_token` (RFC 6749 §6), the explicit `refresh_token: null` case from the earlier round, the non-OK error-text path, and the exchange happy path. So on develop `null` and omission were already "keep `rt-old`", and at head they still are; the only delta is that a 200 with a garbage body now fails as a typed `anthropic_oauth.token_invalid_shape` / `token_invalid_json` instead of persisting `NaN` expiries or non-string tokens.

One reading note on the overload pair, agreeing with the earlier review: the `"exchange"` overload's required `refreshToken` is only sound because the `mode === "exchange" && !isNonBlankString(refreshToken)` throw runs before the shared return; the invariant is local to this function and the test `missing refresh token` pins it.

Approving.

```json
{
  "schemaVersion": "2",
  "headSha": "01654ee7e6",
  "recommendation": "accept",
  "reproduced": true,
  "risks": "none",
  "effortBand": "low",
  "complexity": "moderate",
  "impact": "medium",
  "reviewLoad": "standard",
  "recommendedTier": "small",
  "recommendedThirds": 3,
  "workUnitId": "wu_elizaos_eliza_pr_30566",
  "valueRationale": "Checked for develop-side drift on the touched files since the merge-base (none), ran the suite at head (30 pass) and against develop's anthropic-login.ts (20 fail / 10 pass), confirming the 20 failures are exactly the new rejections and the null/omitted refresh_token compatibility cases already held on develop.",
  "usefulArtifacts": ["RED/GREEN run of anthropic-login.test.ts against develop and head; merge-base drift check"],
  "commands": ["bunx vitest run src/vendor/pi-oauth/anthropic-login.test.ts   # head: 30 pass; develop source + PR tests: 20 failed | 10 passed"],
  "evidenceUrls": [],
  "summary": "Anthropic OAuth token responses are now validated before persistence; verified no base drift since the merge-base and that the control fails exactly the twenty new rejection cases while the refresh-token preservation cases already held on develop."
}
```

— [nxn-claude-code]

lalalune (approved):
Reviewed both token exchange and refresh, their callers and persistence boundary, all added tests, and the earlier null-rotation finding. The refresh response keeps the existing token when rotation is omitted or null; exchange still requires a refresh token. Malformed tokens, non-positive lifetimes, numeric overflow, and invalid JSON reject before constructing credentials.

On `3640209cf0a42539c9999046ac82f49ad5b1b6ec`, rebased onto `10adfdb85226639480c544f6522954f0d44eb1f7`, Bun 1.3.14 / Node 24.15.0:
- `bun install`: passed.
- Full auth suite: 19 files, 234 tests passed.
- Auth typecheck, lint and format checks: passed.

Real Response bodies cover provider parsing; no live credential exchange was needed or claimed. UI/native captures and model trajectories are N/A for this parsing-only change. No further implementation changes were needed. Hosted checks must pass at this head before merge.

Issue #30565, opened 2026-09-04
ops(staging): provision the Stripe subscription catalog bindings


lalalune:
## Problem

Live staging Worker tail on the exact deployed staging SHA repeatedly emits `SUBSCRIPTION_CATALOG_STRIPE_KEY_INVALID` while bootstrapping the full application. Read-only secret inventory confirms staging has none of the five required catalog bindings:

- `STRIPE_SECRET_KEY`
- `STRIPE_PLUS_MONTHLY_PRICE_ID`
- `STRIPE_PLUS_PRODUCT_ID`
- `STRIPE_PRO_MONTHLY_PRICE_ID`
- `STRIPE_PRO_PRODUCT_ID`

The thin-route repair in #30559 removes this cold bootstrap from gateway token exchange and proactive Discord greeting claims, but subscription catalog functionality itself remains unavailable on staging.

## Acceptance

- Install a Stripe test-mode secret key and matching test-mode Plus/Pro product and monthly price IDs in the staging Cloudflare Worker environment.
- Run `bun run --cwd packages/cloud/api preflight:subscription-catalog` against the staging authority without printing secret values.
- Redeploy staging and verify the exact deployed SHA.
- Prove a full-app request no longer emits `SUBSCRIPTION_CATALOG_STRIPE_KEY_INVALID`.
- Prove subscription quote/checkout surfaces use only the expected staging test-mode objects.
- Do not copy production Stripe credentials or live-mode object IDs into staging.

Issue #30559, opened 2026-09-04
fix(cloud): poll Dedicated provisioning job before personal cutover


lalalune:
## Problem

The signed-in Personal Dedicated activation client discards the job metadata returned by adopt-existing and immediately POSTs the full cutover endpoint every five seconds. A failed or stalled provision is therefore retried as generic readiness for the entire six-minute deadline. The current staging receipt recorded 66 cutover 4xx responses after one successful adoption confirmation.

The adoption route also starts its immediate provisioning nudge as an unregistered floating promise, triggerImmediate accepts non-2xx responses, cutover rejection branches lack structured diagnostics, and the privacy-safe browser artifact omits the actual cutover response code.

Live staging evidence also proved that the six-minute client deadline is shorter than the server's 390-second estimate and the supported cold pull-plus-health budget of roughly 11 minutes. Staging has warm-pool provisioning disabled, so this mismatch is exercised on the normal retained-target cold path.

## Acceptance

- Preserve and poll the adoption job until completed before attempting cutover.
- Preserve and poll the job returned by a fresh activation before cutover.
- Surface a sanitized terminal job failure immediately and make no cutover request for a failed job.
- Already-running adopted targets proceed directly to cutover.
- Retry cutover by explicit response code, not all 409/423/503 statuses.
- Attach supported background work to Cloudflare executionCtx.waitUntil and treat non-2xx trigger responses as failure.
- Log structured cutover rejection context without secrets or conversation content.
- Preserve privacy-safe cutover response code counts in deployed browser diagnostics.
- Make the default startup deadline cover the supported cold-boot budget and correct misleading under-a-minute progress copy.
- Add focused unit and contract tests for pending, completed, failed, retryable, terminal, and deadline paths.
- Prove the deployed staging browser flow and latency receipt on the exact merged SHA.

lalalune:
Follow-up diagnostics and correctness fix is in #30633 at exact head `9fdd94a6f425e494964ca1efddd7da695918db3d`.

Current measured staging evidence from run 33949516711: 44/44 paired warm calls succeeded on retry; gateway p50 167.15 ms and p95 959.95 ms, direct Cerebras p50 79.10 ms and p95 113.64 ms, overhead p50 89.73 ms and p95 846.31 ms. This is materially improved from the prior gateway p50/p95 737.61/4123.02 ms.

[...]

PR #30567, opened 2026-09-04
fix(cloud): await Dedicated jobs and trim cold paths

lalalune:
Dedicated activation now waits for its provisioning job before cutting over to the new backend. Fresh activation requires a nonempty job ID, adoption retains its own job identity, and terminal failures stop the transition with an actionable error. Pending work and response-body reads share an absolute deadline.

Worker provisioning nudges are retained through `executionCtx.waitUntil`; hosts without that context await the work. Shared history that is still warming returns a retryable `503 shared_history_unavailable`. Exact auth-token and pending-greetings bootstrap routes avoid unrelated initialization. Privacy-safe cutover diagnostics retain the latest machine code and status.

Relates to #30559. The independent staging Stripe catalog gap remains tracked in #30565.

## Verification

Reviewed head: `e18109441162fc68334f772cbb3d60e4153201a2`, rebased on `aba2fffd930f5ec256fe61cebc67457be9e69708` with a clean install. Final root verification passed all 376 tasks and all repository audits. [The exact-head hosted run passed all five checks](https://github.com/elizaOS/eliza/actions/runs/33942168176).

- Focused cutover API: 19 tests / 188 assertions; UI client: 64 tests; authorization/nudge: 54 tests / 181 assertions; adoption integration: 29 tests / 203 assertions; continuity: 31 tests. Removing the nudge-retention repair makes its regression fail.
- App scripts: 948 passed, 2 skipped. The shared host required a 120-second test harness limit for four unchanged timing-sensitive cases; production deadlines were not changed for that run.
- `audit:app`: 219 passed, 216 captures inspected by the audit, no computed broken or needs-work verdicts. Affected desktop/mobile rest and hover captures were reviewed.
- Startup renderer fixtures: all four viewport/entry cases passed. The final comparison rebuilt and ran the two mobile/desktop cases on both the base and final head; both revisions passed those cases. [Comparison receipt](https://github.com/elizaOS/eliza/releases/download/pr-evidence-11/30567-renderer-comparison.json).

## Reviewed renderer evidence

These captures use the real renderer with scripted HTTP and assistant fixtures. The visible JSON assistant response is fixture output. They prove the UI flow and preserve the existing layout; they do not establish deployed provisioning or live model success.

| View | Before (`aba2fffd`) | After (`e1810944`) | Final walkthrough |
| --- | --- | --- | --- |
| Mobile | Before mobile | After mobile | [MP4](https://github.com/elizaOS/eliza/releases/download/pr-evidence-11/30567-final-head-mobile-e1810944.mp4) |
| Desktop | Before desktop | After desktop | [MP4](https://github.com/elizaOS/eliza/releases/download/pr-evidence-11/30567-final-head-desktop-e1810944.mp4) |

[Final renderer console and network receipt](https://github.com/elizaOS/eliza/releases/download/pr-evidence-11/30567-final-renderer-console-network.json) retains every console event and API request method/path/status from the two final traces. Both viewports reached `ready`; neither recorded a console error. Query strings, request bodies, and headers are excluded from this public diagnostic receipt. Backend and live-model logs are not applicable to the scripted renderer run and remain required for deployed acceptance below.

## Deployed acceptance

The canonical Cloudflare staging workflow accepts `develop`, so deployed acceptance follows the merge. Deploy the exact merged revision with the coordinated worker rollout, run the renderer-enabled frontend receipt, inspect structured Cloudflare logs, and repeat direct-Cerebras versus gateway latency certification. Keep #30559 open until that proof is attached. The read-only personal Cloud check available during this review returned 401; no provisioning job was created from it. #30600 also repairs worker-before-renderer deployment ordering and must be coordinated with this staging run.

### DIFF
--- a/packages/app/test/cloud-live-continuity-contract.test.ts
+++ b/packages/app/test/cloud-live-continuity-contract.test.ts
@@ -238,6 +238,8 @@ describe("forbidden Cloud agent mutations", () => {
       completedDedicatedQuoteResponseBodyCount: 0,
       parsedDedicatedQuoteResponseBodyCount: 0,
       decodedDedicatedQuoteResponseCount: 0,
+      dedicatedCutoverResponseStatus: null,
+      dedicatedCutoverResponseCode: null,
       uninspectableDedicatedQuoteResponseBodyCount: 0,
       dedicatedActivationPostRequestCount: 0,
       successfulDedicatedActivationPostResponseCount: 0,
@@ -436,6 +438,8 @@ describe("forbidden Cloud agent mutations", () => {
       uninspectableDedicatedActivationResponseBodyCount: 0,
       dedicatedActivationResponseStatus: 202,
       dedicatedActivationResponseCode: null,
+      dedicatedCutoverResponseStatus: 200,
+      dedicatedCutoverResponseCode: null,
       dedicatedCutoverPostRequestCount: 2,
       successfulDedicatedCutoverPostResponseCount: 1,
       clientErrorDedicatedCutoverPostResponseCount: 1,
@@ -493,6 +497,28 @@ describe("forbidden Cloud agent mutations", () => {
       "dedicated_quote_changed",
     );
     expect(JSON.stringify(snapshot)).not.toMatch(/private user detail|quoteId/);
+  });
+
+  it("retains only the latest bounded Dedicated cutover status and error code", async () => {
+    const audit = createCloudLiveNetworkAudit();
+    const cutover =
+      "https://api.test/api/v1/eliza/agents/private-target/upgrade-tier/cutover";
+    audit.observeRequest("POST", cutover);
+    audit.observeResponse(
+      "POST",
+      cutover,
+      409,
+      boundedJsonBody({
+        success: false,
+        code: "dedicated_not_healthy",
+        error: "private operator detail",
+      }).responseBody,
+    );
+
+    const snapshot = await audit.snapshot();
+    expect(snapshot.dedicatedCutoverResponseStatus).toBe(409);
+    expect(snapshot.dedicatedCutoverResponseCode).toBe("dedicated_not_healthy");
+    expect(JSON.stringify(snapshot)).not.toMatch(/private operator detail/);
   });
 
   it("fails closed when an approved quote is followed by another agent target", async () => {
--- a/packages/app/test/cloud-live-continuity-contract.ts
+++ b/packages/app/test/cloud-live-continuity-contract.ts
@@ -97,6 +97,8 @@ export interface CloudLiveNetworkAuditSnapshot {
   uninspectableDedicatedActivationResponseBodyCount: number;
   dedicatedActivationResponseStatus: number | null;
   dedicatedActivationResponseCode: string | null;
+  dedicatedCutoverResponseStatus: number | null;
+  dedicatedCutoverResponseCode: string | null;
   dedicatedCutoverPostRequestCount: number;
   successfulDedicatedCutoverPostResponseCount: number;
   clientErrorDedicatedCutoverPostResponseCount: number;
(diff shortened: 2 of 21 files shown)


ss251 (approved):
Reviewed head 082afa99761521c416fac8482255a4637dfc71fc.

The risky part is `packages/ui/src/api/client-cloud.ts:4674` where cutover retry switches from `status 409/423/503` to `PERSONAL_DEDICATED_RETRYABLE_CUTOVER_CODES.has(code)`, checked at `:5300`. I cross-checked the nine allowlisted codes against the new `rejectCutover` codes in `packages/cloud/api/v1/eliza/agents/[agentId]/upgrade-tier/cutover/route.ts`: `dedicated_not_healthy`, `dedicated_not_reachable`, `dedicated_transport_unavailable`, `dedicated_history_import_failed` are in, while `dedicated_history_receipt_invalid`, `shared_history_identity_missing`, `invalid_dedicated_cutover` are out. So `409 + code dedicated_history_receipt_invalid` -> immediate throw after 1 attempt, not a deadline loop, which the new test asserts with `cutoverAttempts` toBe 1.

Verified by reading the diff at this head and matching route codes to the allowlist.

Recommendation: merge

— [muse-lane-ss251]

attentionhead (approved):
Approving, with an explicit note on how far I got: this is ten distinct changes across 19 files, so I verified the highest-risk claims rather than every line, and I would rather say that plainly than imply a line-by-line pass over +1159/-241.

**What I verified**

*The "privacy-safe" diagnostics claim.* `rejectCutover` spreads `...rejection.context` into a `logger.warn`, and the call sites put real identifiers there — `orgId: user.organization_id`, `userId: user.id`, `dedicatedAgentId: active.id`. That looked like a problem until I checked the sink and the convention. It is a server-side Worker log, and `packages/cloud/api/v1/eliza/agents/route.ts` already logs `orgId: user.organization_id` in roughly a dozen places, so id-correlated server logs are the established pattern here. The client-facing half is separate and safe: `rejection.error` values are static strings (`"Agent not found"`, `"A valid Dedicated target id is required."`), never upstream text. So the claim holds — worth knowing it rests on the sink, since the same fields in an operator report that lands in CI output would be a different matter, as your own #30561 was careful about.

*Continuity evidence.* The new `dedicatedCutoverResponseStatus`/`Code` fields and the `latestDedicatedCutoverResponse` lookup mirror the existing activation handling exactly, including the `.slice().reverse().find(...)` "latest with a settled status" shape. Generalising `lifecycleRequest` from `phase === "activation"` to `phase === dedicatedRequest` is the right way to add cutover without forking the path. The bounded-body reading it depends on is pre-existing infrastructure, not new code here.

*Cited evidence reproduces.* Continuity contract 31/31, matching your stated number exactly. `provisioning-jobs-trigger-immediate` 2/2.

**What I did not verify**, so it is not implied by this approval: the client cold-start budget change, the `waitUntil` nudge registration and rejection observation, the account-standing log de-duplication, and the bootstrap route re-pointing through the dependency-bounded shell. Those are four independent behaviors, each plausible from the diff, none of which I exercised.

**The one thing I would push back on is scope.** Ten changes in one PR, spanning cutover job polling, client cold-start copy, Worker `waitUntil` semantics, logging de-duplication, and frontend bootstrap routing, is hard to revert independently. If the cold-start budget turns out wrong in staging, backing it out means backing out the cutover fix that #30559 actually needs. The cutover/job-polling changes are the ones tied to the reported failure; the bootstrap-route trimming and the log de-duplication read as separable cleanups that would be easier to land and revert on their own. Not blocking, since the parts I checked are coherent and well tested, but worth weighing against how much staging risk you want coupled together.

Nice touches worth calling out: stopping immediately on terminal job failure and retrying cutover only for explicitly enumerated transient machine codes is the right asymmetry — blanket retry on an unknown failure code is how a terminal error becomes a hang. And preserving both adoption and fresh-activation job ids before cutover is what makes the polling honest rather than assuming a single provisioning path.

— [claude-code-eliza]

PR #30568, opened 2026-09-05
fix(ci): classify staging provisioning worker timeouts

lalalune:
The staging renderer reaches an activated Dedicated target whose provisioning jobs fail with a generic timeout. The existing canary diagnostic cannot distinguish that failure from image, SSH, runtime-health, and execution-watchdog failures elsewhere in the staging worker.

Extend the existing read-only staging diagnostic with an on-host journal classifier. It emits only closed signal names, numeric timeout durations, and allowlisted stack-function names. Raw journal messages, credentials, hostnames, agent identifiers, and user content remain on the host. No provisioning, cleanup, restart, or deployment action is added.

Validation: the actual embedded Python classifier passes mixed private-record, empty-journal, and malformed-input tests; focused diagnostics pass 8/8; Biome, actionlint, and diff checks pass; `bun install` and root `bun run verify` pass (376/376 tasks). The protected hosted diagnostic will provide the live timeout evidence after merge.

Refs #30552.

### DIFF
--- a/.github/workflows/live-smoke.yml
+++ b/.github/workflows/live-smoke.yml
@@ -535,7 +535,74 @@ jobs:
           bun-version: "1.3.14"
 
       - name: Validate provision diagnostic contract
-        run: bun test packages/cloud/scripts/admin/managed-dedicated-provision-diagnostic.test.ts
+        run: >-
+          bun test
+          packages/cloud/scripts/admin/managed-dedicated-provision-diagnostic.test.ts
+          packages/scripts/__tests__/staging-worker-journal-diagnostic.test.ts
+
+      - name: Classify recent provisioning worker failures
+        uses: appleboy/ssh-action@0ff4204d59e8e51228ff73bce53f80d53301dee2
+        with:
+          host: ${{ secrets.ELIZA_PROVISIONING_HOST }}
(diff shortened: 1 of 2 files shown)


ss251 (changes requested):
Reviewed head bf3c309aa33526d2c7db6d4b3b5e2cf80fe57869.

The risky hunk is `.github/workflows/live-smoke.yml:560` where `ssh_connect_timeout` is `(?:SSH|connection|connect).*timed out` (re.I). That also matches the `ssh_command_timeout` message: input `{"MESSAGE":"[docker-ssh] Command timed out after 10ms"}` increments both `ssh_command_timeout` and `ssh_connect_timeout` (via `ssh] Command timed out`), so `signals` double-counts one failure as two phases. The new test in `packages/scripts/__tests__/staging-worker-journal-diagnostic.test.ts:33` asserts `ssh_command_timeout==1` but never asserts `ssh_connect_timeout==0`, so it still passes with the overlap present. Non-blocking: `unclassified_timeout` overlaps by design.

Verified by reading the diff at this head and tracing both regexes against the test fixture; no live host reached.

Recommendation: fix

— [muse-lane-ss251]

PR #30580, opened 2026-09-05
fix(ci): report activated staging job timeout origin

lalalune:
The activated staging agent's provisioning failures are persisted in `jobs.error`, while the worker journal omits ordinary job failures. The existing coarse diagnostic reports only `timeout`, so the journal classifier alone cannot locate this failure.

Add a read-only query to the existing staging diagnostic. It resolves the exact canary owner, follows that owner's Dedicated activation receipt, and reads only the latest provisioning job belonging to the same organization and user. Its output contains counts, timestamps, attempts, storage presence, and allowlisted error terms and source module names. No raw error, identifier, address, credential, or user content is emitted. Missing or ambiguous ownership produces explicit zero target/job counts.

Validation: the actual SQL passes PostgreSQL/PGlite execution with competing organizations, competing users, older jobs, missing targets, duplicate canaries, and private error text. Focused diagnostics pass 9/9; Biome, actionlint, and diff checks pass; `bun install` and root `bun run verify` pass (376/376 tasks). Live output remains to be collected through the protected staging workflow after merge.

Refs #30552.

### DIFF
--- a/.github/workflows/live-smoke.yml
+++ b/.github/workflows/live-smoke.yml
@@ -863,6 +863,61 @@ jobs:
             "$CANARY_DIAGNOSTIC_RAW_PATH" \
             "$CANARY_DIAGNOSTIC_EVIDENCE_PATH"
 
+      - name: Diagnose activated target job origin
+        shell: bash
+        env:
+          DATABASE_URL: ${{ secrets.DATABASE_URL || secrets.RAILWAY_DATABASE_URL || secrets.NEON_DATABASE_URL }}
+          PGOPTIONS: -c default_transaction_read_only=on -c statement_timeout=20000
+        run: |
+          set -euo pipefail
+          PSQL_DATABASE_URL="$(bun packages/cloud/scripts/admin/libpq-connection-url.ts)"
+          echo "::add-mask::$PSQL_DATABASE_URL"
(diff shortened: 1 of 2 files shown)


ss251 (approved):
Reviewed head bc06d42c716a3a445da16f1630d1407d1c5a767b.

The risky part is `.github/workflows/live-smoke.yml:881` where `target`/`latest_job` resolve the activated staging job. I traced the guards that usually break tenant-scoped diagnostics: `target` requires `(SELECT count(*) FROM canary)=1` and `latest_job` requires `(SELECT count(*) FROM target)=1` with join on `agent_id/organization_id/user_id`. `packages/scripts/__tests__/staging-activation-job-origin.test.ts:34` covers it: competing `foreign-org`/`other-user` `agent_provision` rows and an older `Docker health timeout` are ignored, returning only the bound row with `attempts 3` and `errorTerms [connect,exceeded,pool,timeout,trying]`; duplicate-canary input collapses to `canaryCount 2/targetCount 0/jobCount 0`. Closed regex vocabs keep `PRIVATE_CREDENTIAL` out. Non-blocking: PGOPTIONS 20s duplicates SET LOCAL.

Verified by reading diff at this head against test fixtures.
Recommendation: merge

— [muse-lane-ss251]

PR #30584, opened 2026-09-05
fix(auth): validate OpenAI Codex OAuth token responses

### LINKED ISSUE #30581: fix(auth): validate OpenAI Codex OAuth token responses
## Problem

The OpenAI Codex OAuth token boundary currently asserts the decoded JSON shape. A successful provider response can therefore produce an invalid credential record when an optional token has the wrong runtime type or when `expires_in` is non-positive or overflows the absolute expiry calculation.

The authorization-code and refresh paths also validate the same wire contract separately, which makes their required-versus-optional behavior easy to drift.

## User impact

An invalid successful response can be returned as healthy credentials and then fail later during persistence or the next refresh. Zero or negative lifetimes can trigger immediate refresh loops, while numeric overflow can create a non-expiring credential timestamp.

## Acceptance criteria

- Parse successful token JSON as `unknown` at one shared boundary.
- Require nonblank string access and authorization-code refresh tokens.
- On refresh, preserve the current refresh token when the provider omits it or returns null; accept a rotated token only when it is a nonblank string.
- Require `expires_in` to be a positive finite number and reject a non-finite computed absolute expiry.
- Treat `id_token` as optional/nullish, but reject other present runtime types.
- Keep invalid-response diagnostics actionable and limited to field/reason metadata.
- Add real `Response`-backed regression tests for authorization-code and refresh success/error shapes, including numeric overflow.
- Pass the auth package tests, typecheck, lint/format checks, build, and root `bun run verify`.

## Scope

`packages/auth/src/vendor/pi-oauth/openai-codex-login.ts` and its focused tests. No UI, public API, or OAuth endpoint changes.

wtfsayo:
Closes #30581.

Validate OpenAI Codex OAuth token responses before constructing credentials. Preserve refresh-token rotation and compatible optional ID-token behavior, reject invalid JSON and malformed token fields, and keep token values out of diagnostics. Rejected HTTP responses are canceled before returning failure so repeated exchange and refresh attempts release their transport resources.

## Validation

Reviewed the token parser, its callers, response cleanup, and the existing findings. The validation is useful: malformed successful responses must not become persisted credentials, and diagnostics must exclude token values.

I repaired the response lifecycle regression from my earlier review. Both authorization exchange and refresh now cancel rejected response bodies before returning failure, without logging their contents. The new `packages/auth/src/openai-codex-transport.test.ts` runs real Node HTTP/Fetch exchanges with a finite 384 KiB response; both cases fail on the original implementation and pass after the repair. Each path performs two attempts and verifies that every response body is released.

Validation with Node 24.15.0 and Bun 1.3.14:

- Full auth suite: 20 files, 241 tests passed.
- Auth typecheck, lint, and format checks: passed.
- Root `bun run verify`: 376/376 workspace tasks and repository audits passed.
- `bun install` completed; working tree clean after committing the repair.

Reviewed head: `8ec579e9e51b0a48de98618cfabd2c8c92ca33a4`.
Rebase/validation base: `2fd687670a5` (the subsequent APNs-only base change does not touch this contract).

The HTTP fixture uses synthetic credentials and a real transport. This is transport and credential parsing evidence; no live provider login or user account mutation is claimed. No rendered UI or model behavior changes.

## Evidence

- [x] N/A - No rendered UI change.

- [x] N/A - No rendered UI change.

- [x] N/A - No interactive UI or native transport behavior changes.

- [x] Executed local HTTP fixture and command results are documented above.

- [x] N/A - No frontend path changes.

- [x] N/A - No model, prompt, provider inference, or action behavior changes.

- [x] Real temporary credential-response bodies were inspected by the HTTP tests; no user credentials or account state were changed.

- [x] N/A - No rendered UI change.

### DIFF
--- a/packages/auth/src/vendor/pi-oauth/openai-codex-login.ts
+++ b/packages/auth/src/vendor/pi-oauth/openai-codex-login.ts
@@ -46,16 +46,6 @@ export interface OAuthPrompt {
   placeholder?: string;
 }
 
-async function readOAuthFailureBody(response: Response): Promise<string> {
-  try {
-    return await response.text();
-  } catch (cause) {
-    // error-policy:J4 explicit diagnostic degrade — the non-2xx status remains
-    // authoritative when the optional response body cannot be read.
-    return `[response body unavailable: ${cause instanceof Error ? cause.message : String(cause)}]`;
-  }
-}
-
 async function createState(): Promise<string> {
   if (!isNodeLikeRuntime()) {
     throw new Error(
@@ -125,9 +115,126 @@ type TokenSuccess = {
    * account's CODEX_HOME to authenticate. */
   idToken?: string;
 };
-type TokenFailure = { type: "failed" };
+type TokenFailure = { type: "failed"; reason: string };
+type TokenResult = TokenSuccess | TokenFailure;
 
-async function exchangeAuthorizationCode(
+type TokenResponseContext =
+  | { mode: "exchange" }
+  | { mode: "refresh"; currentRefreshToken: string };
+
+function isRecord(value: unknown): value is Record<string, unknown> {
+  return value !== null && typeof value === "object" && !Array.isArray(value);
+}
+
+function isNonBlankString(value: unknown): value is string {
+  return typeof value === "string" && value.trim().length > 0;
+}
+
+function failedTokenResponse(reason: string): TokenFailure {
+  return { type: "failed", reason };
+}
+
+function parseTokenResponse(
+  payload: unknown,
+  context: TokenResponseContext,
+): TokenResult {
+  if (!isRecord(payload)) {
+    return failedTokenResponse("response root must be an object");
+  }
+
+  const rawAccessToken = payload.access_token;
+  const rawRefreshToken = payload.refresh_token;
+  const rawExpiresIn = payload.expires_in;
+  const rawIdToken = payload.id_token;
+
+  if (!isNonBlankString(rawAccessToken)) {
+    return failedTokenResponse(
+      "missing or invalid access_token; expected a non-blank string",
+    );
+  }
+  if (
+    typeof rawExpiresIn !== "number" ||
+    !Number.isFinite(rawExpiresIn) ||
+    rawExpiresIn <= 0
+  ) {
+    return failedTokenResponse(
+      "missing or invalid expires_in; expected a positive finite number",
+    );
+  }
+
+  const expires = Date.now() + rawExpiresIn * 1000;
+  if (!Number.isFinite(expires)) {
+    return failedTokenResponse(
+      "invalid expires_in; absolute expiry exceeds the supported numeric range",
+    );
+  }
+
+  let refresh: string;
+  if (context.mode === "exchange") {
+    if (!isNonBlankString(rawRefreshToken)) {
+      return failedTokenResponse(
+        "missing or invalid refresh_token; expected a non-blank string",
+      );
+    }
+    refresh = rawRefreshToken;
+  } else if (rawRefreshToken === undefined || rawRefreshToken === null) {
+    refresh = context.currentRefreshToken;
+  } else if (!isNonBlankString(rawRefreshToken)) {
+    return failedTokenResponse(
+      "invalid refresh_token; expected a non-blank string when present",
+    );
+  } else {
+    refresh = rawRefreshToken;
+  }
+
+  let idToken: string | undefined;
+  if (rawIdToken !== undefined && rawIdToken !== null) {
+    if (typeof rawIdToken !== "string") {
+      return failedTokenResponse(
+        "invalid id_token; expected a string when present",
+      );
+    }
+    idToken = rawIdToken || undefined;
+  }
+
+  return {
+    type: "success",
+    access: rawAccessToken,
+    refresh,
+    expires,
+    ...(idToken ? { idToken } : {}),
+  };
+}
+
+async function readTokenResponse(
+  response: Response,
+  context: TokenResponseContext,
+): Promise<TokenResult> {
+  let payload: unknown;
+  try {
+    payload = await response.json();
+  } catch {
+    // error-policy:J3 provider JSON is untrusted; malformed JSON becomes an
+    // explicit failed result and no partial credential record is constructed.
+    const failure = failedTokenResponse("response body was not valid JSON");
+    logger.error(
+      { mode: context.mode, reason: failure.reason },
+      "[openai-codex] Token response invalid",
+    );
+    return failure;
+  }
+
+  const result = parseTokenResponse(payload, context);
+  if (result.type === "failed") {
+    logger.error(
+      { mode: context.mode, reason: result.reason },
+      "[openai-codex] Token response invalid",
+    );
+  }
+  return result;
+}
+
+export async function exchangeAuthorizationCode(
   code: string,
   verifier: string,
   redirectUri: string = REDIRECT_URI,
@@ -144,35 +251,13 @@ async function exchangeAuthorizationCode(
     }),
   });
   if (!response.ok) {
-    const text = await readOAuthFailureBody(response);
     logger.error(
-      `[openai-codex] code->token failed: ${response.status} ${text}`,
+      { mode: "exchange", status: response.status },
+      "[openai-codex] Token request returned a non-success status",
     );
-    return { type: "failed" };
+    return failedTokenResponse("provider returned a non-success status");
   }
-  const json = (await response.json()) as {
-    access_token?: string;
-    refresh_token?: string;
-    expires_in?: number;
-    id_token?: string;
-  };
-  if (
-    !json.access_token ||
-    !json.refresh_token ||
-    typeof json.expires_in !== "number"
-  ) {
-    logger.error(
-      `[openai-codex] token response missing fields: ${JSON.stringify(json)}`,
-    );
-    return { type: "failed" };
-  }
-  return {
-    type: "success",
-    access: json.access_token,
-    refresh: json.refresh_token,
-    expires: Date.now() + json.expires_in * 1000,
-    ...(json.id_token ? { idToken: json.id_token } : {}),
-  };
+  return readTokenResponse(response, { mode: "exchange" });
 }
 
 async function refreshAccessToken(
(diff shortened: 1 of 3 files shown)


ss251 (approved):
Reviewed head 0dbea4ce040dafbf63d1543ce975e3c952d0a88e.

The risky part is `packages/auth/src/vendor/pi-oauth/openai-codex-login.ts:118` in new `parseTokenResponse`/`readTokenResponse`, which replaces raw `JSON.stringify(json)` and body-text error logging. I traced the two overflow boundaries: `expires_in: 1e308` passes `Number.isFinite` but fails the `Date.now()+expires_in*1000` finite-expiry check, while `1e400` becomes `Infinity` in `response.json()` and fails the first gate — both covered by mocked `Response` tests. The non-success branch now logs only `{mode,status}`, and the invalid-payload branch logs only `{mode,reason}`, so `access-secret-value`/`refresh-secret-value` never reach `logger.error`.

Verified by reading the diff at this head and following exchange/refresh callers through the shared helper.

Recommendation: merge

— [muse-lane-ss251]

attentionhead (approved):
Approving at `0dbea4ce`. I reviewed the Anthropic counterpart of this change (#30566) earlier, including the head where adding validation introduced a null-rotation regression, so I came to this one looking for that exact bug. It is not here, and the structure is better than the fix that closed it there.

**The null-rotation case is handled correctly, and first.**

```ts
} else if (rawRefreshToken === undefined || rawRefreshToken === null) {
    refresh = context.currentRefreshToken;
} else if (!isNonBlankString(rawRefreshToken)) {
    return failedTokenResponse("invalid refresh_token; ...");
}
```

This is the failure mode worth dwelling on, because it is what adding validation to the Anthropic file caused: `refresh_token: null` is a normal "keep your current token" response under RFC 6749 §6, and a validator that only special-cases `undefined` turns it into a hard failure that burns a still-valid stored token. Ordering the absent-value branch *before* the non-blank check makes that structurally impossible, rather than patching it with an extra `!== null &&` guard bolted into the rejection condition. The same shape applies to `id_token` just below it.

I confirmed the test actually pins this rather than passing incidentally. The regression case is parameterized over both shapes:

```ts
it.each([["omitted", ...], ["null", ...]])(
  "keeps the current refresh token when refresh_token is %s", ...)
```

and mutating the branch to `rawRefreshToken === undefined` — precisely the Anthropic defect — gives **1 failed, 28 passed**, with the failure being the `null` case. So the test is load-bearing against the exact regression this file was at risk of inheriting.

**Removing the failure-body copy is the right call, and the diagnostics survive it.** Deleting `readOAuthFailureBody` means provider error bodies are no longer interpolated into an error string that can travel. My first reaction was that this costs debuggability, since `"provider returned a non-success status"` alone cannot distinguish a 400 from a 429. It does not, because the status is preserved structurally on the other side of the boundary:

```ts
logger.error({ mode: "exchange", status: response.status },
  "[openai-codex] Token request returned a non-success status");
```

Generic reason on the returned value, precise status in the structured log, is the correct split — the same separation that makes the cutover diagnostics in #30567 safe.

**One follow-up, out of scope here.** This leaves the two OAuth vendors inconsistent: `anthropic-login.ts` on `develop` still does

```ts
const errText = await tokenResponse.text();
throw new Error(`Token exchange failed: ${errText}`);
```

at both its exchange and refresh sites. So after this merges, the Codex path stops copying provider bodies into errors while the Anthropic path continues to. Worth the same treatment, and worth doing as its own PR rather than widening this one.

**Verification:** `openai-codex-login.test.ts` 29/29, the whole `vendor/pi-oauth` directory 46/46, `tsc --noEmit` clean. Delta is 2 files, `+342/-77` — note that a naive range diff against a stale local `develop` shows 49 files of unrelated churn; against the actual merge base `9322e05c` it is the two files claimed.

The overload-free `TokenResult` union here reads better than the overloaded `parseAnthropicTokenResponse` in the sibling, since the exchange/refresh distinction is carried by `context.mode` and a returned failure value rather than by two overload signatures that assert more than the implementation. Returning `TokenFailure` instead of throwing also keeps the validation boundary honest about being an expected outcome.

Approving.

— [claude-code-eliza]

packages/auth/src/vendor/pi-oauth/openai-codex-login.ts
```diff
@@ -189,41 +274,21 @@ async function refreshAccessToken(
...
-    if (!json.access_token || typeof json.expires_in !== "number") {
       logger.error(
-        `[openai-codex] Token refresh response missing fields: ${JSON.stringify(
-          json,
-        )}`,
+        { mode: "refresh", status: response.status },
+        "[openai-codex] Token request returned a non-success status",
       );
-      return { type: "failed" };
+      return failedTokenResponse("provider returned a non-success status");
```
lalalune, line 297: [P2] Release the rejected response body before returning. This branch and the corresponding exchange branch at line 258 remove the previous body drain without cancellation. A finite 256 KiB error response retains the only connection in a bounded pool: a second refresh times out before reaching the server, while the exact base consumes both bodies and completes both requests. Cancel or discard the body without exposing its contents, and add a real HTTP lifecycle regression test.

lalalune (changes requested):
I reproduced a response-body lifecycle regression on this exact head using Node 24.15.0, real HTTP/fetch, a finite 256 KiB HTTP 400 body, and a one-connection Undici pool. The unchanged base handles two consecutive refresh requests with no timeouts; this head leaves the first response unread, and the second refresh times out before reaching the server. The 29 supplied tests and five hosted checks pass but do not exercise socket reuse. Please consume or cancel rejected response bodies without logging their contents before returning, and cover the transport lifecycle.

byt61 (approved):

- PR suite + PR source: 29/29 pass under the package vitest runner.
- Mutant check (PR tests vs pre-PR source): 20 fail — a mix of the newly-exported `exchangeAuthorizationCode` import and genuine validation gaps in the old cast-then-truthy parsing, so the suite is well-coupled to the fix.
- Read the source diff: all `{ type: \"failed\" }` constructions now flow through the single `failedTokenResponse` constructor, and the two in-file consumers narrow on `type` only, so the additive `reason` field is safe. The other repo-wide `exchangeAuthorizationCode` matches are unrelated per-plugin functions.
- The diagnostics change is a real security improvement: token-bearing `JSON.stringify(json)` and raw failure-body logging are replaced with structured `{ mode, reason/status }` records, and the secret-leak tests pin that down (including the non-success-body path).

lalalune (approved):
Reviewed the token parser, its callers, response cleanup, and the existing findings. The validation is useful: malformed successful responses must not become persisted credentials, and diagnostics must exclude token values.

I repaired the response lifecycle regression from my earlier review. Both authorization exchange and refresh now cancel rejected response bodies before returning failure, without logging their contents. The new `packages/auth/src/openai-codex-transport.test.ts` runs real Node HTTP/Fetch exchanges with a finite 384 KiB response; both cases fail on the original implementation and pass after the repair. Each path performs two attempts and verifies that every response body is released.

Validation with Node 24.15.0 and Bun 1.3.14:

- Full auth suite: 20 files, 241 tests passed.
- Auth typecheck, lint, and format checks: passed.
- Root `bun run verify`: 376/376 workspace tasks and repository audits passed.
- `bun install` completed; working tree clean after committing the repair.

Reviewed head: `8ec579e9e51b0a48de98618cfabd2c8c92ca33a4`.
Rebase/validation base: `2fd687670a5` (the subsequent APNs-only base change does not touch this contract).

The HTTP fixture uses synthetic credentials and a real transport. This is transport and credential parsing evidence; no live provider login or user account mutation is claimed. No rendered UI or model behavior changes.

lumix5 (approved):
## Review — APPROVE (exact head `8ec579e9e51b0a48de98618cfabd2c8c92ca33a4`)

The new head resolves the response-lifecycle blocker from the prior REQUEST_CHANGES, and the repair is proven failure-sensitive. Independent of review 5122043415, which landed while I was verifying; evidence below is from my own runs.

**The repair.** `releaseRejectedResponse` cancels the rejected body before both `!response.ok` returns (exchange and refresh), best-effort with a `// error-policy:J6` teardown tag and a warn that logs only the status — no body contents, no tokens. The base commit's validation is unchanged from the already-triple-approved shape: non-blank `access_token`, positive finite `expires_in` with absolute-expiry overflow guard, RFC 6749 §6 refresh-omission semantics preserved, `id_token` rejected when non-string. Compared to develop, the old `readOAuthFailureBody` text logging and `JSON.stringify(json)` missing-field logging are gone — provider/token material can no longer reach diagnostics.

Independent verification (disposable containers, network-disabled, files byte-verified against the head blobs):

- New `openai-codex-transport.test.ts` under Node 24 (the suite's CI runtime — `vitest` bin is node-shebanged and lanes run setup-node): **2 pass / 0 fail**. It is a real exchange, not a mock — local HTTP server returning 400 with a ~384 KiB body, `fetch` stubbed only for endpoint routing, two attempts per mode, asserting every rejected response is released.
- Mutation control (the two `releaseRejectedResponse` calls stripped): both tests **fail at `expect(response.bodyUsed).toBe(true)`** — the test kills exactly the bug lalalune reproduced. Restored, green again.
- PR-touched suites (`openai-codex-login.test.ts` + transport): **31 pass / 0 fail**.
- Full auth suite under Node 24: **232 pass / 6 fail**, all six environmental in my containers — three `account-storage-subprocess` tests require a bun runtime (absent in the node container; test names say "under bun"), three `adopt-codex-cli-login` permission-fixture tests false-fail as uid 0 (root bypasses the chmod-based EACCES simulation). Neither file is touched by this PR.
- Hosted PR Static Smoke green at head.

One portability note, non-blocking: under a pure bun runtime the new transport test's teardown throws — bun 1.3.14's `closeAllConnections()` stops the server, so the following `server.close(cb)` errors `ERR_SERVER_NOT_RUNNING` (reproduced with a bare six-line bun script, no test code). Under node it is green. The suite is wired for node-backed vitest, so no hosted lane is affected; if you ever expect contributors to run this under plain bun, guard the close error.

Final merge decision rests with the maintainers.

Compute receipt: 25270778 project-attributed tokens (exact; device-signed, locally reported)
AI provider/model: anthropic / glm-5.3-1m
— [eliza-review]

PR #30585, opened 2026-09-05
fix(auth): keep startup CLI credential discovery read-only

### LINKED ISSUE #30573: fix(auth): keep external CLI credential diagnostics read-only
Startup credential diagnostics currently refresh an expired credential owned by the Claude Code CLI only to log its availability. The returned rotated credential is discarded. Diagnostic reads should not perform a token exchange or change credential ownership.

Acceptance:
- Keep deferred startup probing local and read-only, including expired credentials and concurrent calls.
- Preserve the existing stored-account refresh/mutex/persistence path and the public startup APIs.
- Exercise the public diagnostic function against temporary credential files and prove no network exchange or file mutation.

Scope excludes the OAuth response validation and local credential metadata validation tracked in #30563 and #30548.

lalalune:
Startup's informational credential probe exchanged expired Claude Code tokens even though it discarded the result. The probe now reads local credential metadata without contacting the provider and leaves refresh to the CLI. Explicit account-store refresh and inference credential resolution retain their existing behavior.

Closes #30573.

Validation on scoped commit `f0a54007dd51d15456045161c11974105049e8ce`, based on `9322e05c91a2e645875c65bdc6d9df1e150d6b37`:

- All 213 auth tests passed with one worker. The initial parallel run exposed an unchanged two-process adoption test race; the complete serial run passed.
- The new regression uses real temporary credential files and verifies unchanged bytes and timestamps, with no refresh request across concurrent probes.
- Package tests, typecheck, lint, and build passed. Root `bun run verify` passed, including all 376 Turbo tasks and the final repository audits. Bun 1.3.14 and Node 24.15.0 were used.
- Independent runtime and cloud reviewers checked refresh ownership and existing OAuth consumers and found no remaining blocker.

Screenshots, video, native captures, and live-model outputs are N/A: this changes informational startup diagnostics, with no rendered view or inference behavior change. The credential-file regression uses synthetic test credentials; no real account credentials were changed.

## Review and risk

Low risk: removes an unused external token exchange from informational startup discovery. The public startup functions remain compatible. Updated module documentation describes the read-only behavior. No deployment or database migration is required.

## Evidence Gate

N/A - informational backend diagnostic; no rendered view changed.

N/A - informational backend diagnostic; no rendered view changed.

N/A - the observable contract is credential-file preservation and absence of network requests.

<details><summary>Terminal verification output</summary>

```text
Test Files  19 passed (19)
Tests       213 passed (213)
Package typecheck: exit 0
Package lint: exit 0
Package build: exit 0
Root verify: exit 0
[anti-larp] scanned 11360 test files — 0 focused, 0 orphaned skip(s)
```

</details>

N/A - no browser request or UI state changes.

N/A - no model, action, prompt, or inference behavior changed.

The real-file regression verified both expired and unexpired synthetic CLI credentials: four concurrent diagnostic calls made zero network requests, preserved file bytes and modification times, and created no sibling credential files. Disabled and absent credentials remained unavailable. These are observed test assertions against temporary filesystem state, not real user account data.
<details><summary>Real temporary-file diagnostic regression</summary>

```text

 RUN  v4.1.10 /Users/shawwalters/.codex/worktrees/code-quality-cleanup-20260904/packages/auth

stdout | src/credentials-cli-diagnostics.test.ts > external CLI startup diagnostics > leaves external credentials and network untouched at expiry 0
 Info       [auth] Detected local Claude Code CLI credentials. The CLI manages refresh. Add an API key or connect Eliza Cloud for the main agent.
 Info       [auth] Detected local Claude Code CLI credentials. The CLI manages refresh. Add an API key or connect Eliza Cloud for the main agent.
 Info       [auth] Detected local Claude Code CLI credentials. The CLI manages refresh. Add an API key or connect Eliza Cloud for the main agent.
 Info       [auth] Detected local Claude Code CLI credentials. The CLI manages refresh. Add an API key or connect Eliza Cloud for the main agent.

stdout | src/credentials-cli-diagnostics.test.ts > external CLI startup diagnostics > leaves external credentials and network untouched at expiry 1788574712462
 Info       [auth] Detected local Claude Code CLI credentials. The CLI manages refresh. Add an API key or connect Eliza Cloud for the main agent.
 Info       [auth] Detected local Claude Code CLI credentials. The CLI manages refresh. Add an API key or connect Eliza Cloud for the main agent.
 Info       [auth] Detected local Claude Code CLI credentials. The CLI manages refresh. Add an API key or connect Eliza Cloud for the main agent.
 Info       [auth] Detected local Claude Code CLI credentials. The CLI manages refresh. Add an API key or connect Eliza Cloud for the main agent.

 ✓ src/credentials-cli-diagnostics.test.ts > external CLI startup diagnostics > leaves external credentials and network untouched at expiry 0 397ms
 ✓ src/credentials-cli-diagnostics.test.ts > external CLI startup diagnostics > leaves external credentials and network untouched at expiry 1788574712462 186ms
 ✓ src/credentials-cli-diagnostics.test.ts > external CLI startup diagnostics > honors disabled probing and keeps an absent external login unavailable 66ms

 Test Files  1 passed (1)
      Tests  3 passed (3)
   Start at  21:18:13
   Duration  19.27s (transform 15.02s, setup 0ms, import 18.46s, tests 650ms, environment 0ms)

```

</details>

Known gaps: the initial parallel auth run hit the existing two-process credential-adoption test race. All 213 tests then passed serially; no test limits or assertions were relaxed. Hosted checks must pass on the current PR head before merge.

Maintainer CI exception: N/A - no exception requested.

### DIFF
--- a/packages/auth/src/credentials-cli-diagnostics.test.ts
+++ b/packages/auth/src/credentials-cli-diagnostics.test.ts
@@ -1,0 +1,82 @@
+/**
+ * Exercises deferred startup diagnostics against real temporary CLI credential
+ * files. The network boundary rejects unexpected refreshes; storage and the
+ * diagnostic path remain real so an expired external login cannot be rotated.
+ */
+import fs from "node:fs/promises";
+import os from "node:os";
+import path from "node:path";
+import { logger } from "@elizaos/core";
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+import { applySubscriptionCredentialsDeferred } from "./credentials.ts";
+
+let tempHome: string;
+let credentialsPath: string;
+let network: ReturnType<typeof vi.fn>;
+
+beforeEach(async () => {
+  tempHome = await fs.mkdtemp(path.join(os.tmpdir(), "eliza-cli-diagnostic-"));
+  for (const name of ["HOME", "USERPROFILE", "ELIZA_HOME", "ELIZA_STATE_DIR"]) {
+    vi.stubEnv(name, tempHome);
+  }
+  vi.stubEnv("ELIZA_DISABLE_SUBSCRIPTION_CREDENTIALS", "");
+  credentialsPath = path.join(tempHome, ".claude", ".credentials.json");
+  await fs.mkdir(path.dirname(credentialsPath));
+  network = vi.fn(async () => {
+    throw new Error("Diagnostics must not exchange external credentials");
+  });
+  vi.stubGlobal("fetch", network);
+});
+
+afterEach(async () => {
+  vi.restoreAllMocks();
+  vi.unstubAllEnvs();
+  vi.unstubAllGlobals();
+  await fs.rm(tempHome, { recursive: true, force: true });
+});
+
+describe("external CLI startup diagnostics", () => {
+  it.each([0, Date.now() + 3_600_000])(
+    "leaves external credentials and network untouched at expiry %s",
+    async (expiresAt) => {
+      const bytes = JSON.stringify({
+        claudeAiOauth: {
+          accessToken: "fixture-access",
+          refreshToken: "fixture-refresh",
+          expiresAt,
+        },
+      });
+      await fs.writeFile(credentialsPath, bytes);
+      const before = await fs.stat(credentialsPath);
+      await Promise.all(
+        Array.from({ length: 4 }, () => applySubscriptionCredentialsDeferred()),
+      );
+      expect(network).not.toHaveBeenCalled();
+      expect(await fs.readFile(credentialsPath, "utf8")).toBe(bytes);
+      expect((await fs.stat(credentialsPath)).mtimeMs).toBe(before.mtimeMs);
+      expect(await fs.readdir(path.dirname(credentialsPath))).toEqual([
+        ".credentials.json",
+      ]);
+    },
+  );
+
+  it("honors disabled probing and keeps an absent external login unavailable", async () => {
+    const info = vi.spyOn(logger, "info");
+    await applySubscriptionCredentialsDeferred();
+    expect(info).not.toHaveBeenCalled();
+    await fs.writeFile(
+      credentialsPath,
+      JSON.stringify({
+        claudeAiOauth: {
+          accessToken: "fixture-access",
+          refreshToken: "fixture-refresh",
+          expiresAt: 0,
+        },
+      }),
+    );
+    vi.stubEnv("ELIZA_DISABLE_SUBSCRIPTION_CREDENTIALS", "true");
+    await applySubscriptionCredentialsDeferred();
+    expect(info).not.toHaveBeenCalled();
+    expect(network).not.toHaveBeenCalled();
+  });
+});
--- a/packages/auth/src/credentials.ts
+++ b/packages/auth/src/credentials.ts
@@ -56,7 +56,6 @@ const DEFAULT_ACCOUNT_ID = "default";
 
 /** Buffer before expiry to trigger refresh (5 minutes) */
 const REFRESH_BUFFER_MS = 5 * 60 * 1000;
-const invalidClaudeCodeRefreshTokens = new Set<string>();
 
 /** Stable failure categories for callers that manage account-pool health. */
 export type AccessTokenFailureKind =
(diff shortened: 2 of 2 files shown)


ss251 (approved):
Reviewed head f0a54007dd51d15456045161c11974105049e8ce.

The risky part is `packages/auth/src/credentials.ts:865` where `applySubscriptionCredentialsDeferred` now only calls `readClaudeCodeOAuthBlob()` for presence logging instead of `importClaudeCodeOAuthToken()` which exchanged the refresh token via `refreshAnthropicToken` when `expiresAt <= Date.now()`. I followed the deleted branch: expired blob + `refreshToken` present -> network refresh, with `invalid_grant` cached in `invalidClaudeCodeRefreshTokens`. The new path performs no fetch, no file write, and no env/config mutation, keeping the early returns for disabled probing and existing `anthropic-subscription` accounts.

`packages/auth/src/credentials-cli-diagnostics.test.ts:47` pins the boundary that matters: `expiresAt` 0 and future, 4-way concurrent calls, asserting `fetch` never called, bytes and `mtimeMs` unchanged, and no sidecar files.

Recommendation: merge

— [muse-lane-ss251]

attentionhead (approved):
Approving at `f0a54007`. I verified the premise rather than the patch, because the whole change rests on one claim — that the probe's result was discarded — and if that were wrong this would be a behaviour change rather than a fix.

**The premise is exactly right.** On `develop`, `importClaudeCodeOAuthToken()` has a single caller, and the returned token is used for nothing but a conditional:

```ts
const claudeImported = await importClaudeCodeOAuthToken();
if (claudeImported) {
  logger.info("[auth] Anthropic subscription detected via Claude Code CLI — ...");
}
```

The token value itself never reaches config or `process.env`. So a network token exchange was being performed on the startup path solely to decide whether to print an informational line.

**That is worse than wasteful, and I think the description undersells it.** Anthropic rotates refresh tokens one-time-use — the same property that made the null-rotation handling in #30566 and #30584 matter. An exchange whose result is discarded consumes the stored refresh token's single use and throws the replacement away, which can leave the user's *actual* Claude Code CLI login broken, by a probe that exists only to log.

The strongest evidence that this was really happening is in the code being deleted: `invalidClaudeCodeRefreshTokens` and `isClaudeCodeInvalidGrantError` exist to remember refresh tokens that came back `invalid_grant` and stop retrying them. That machinery is an artifact of the breakage, not a guard against it — removing the exchange removes the need for both, which is why deleting them here is correct rather than a loss of safety.

**The replacement is genuinely read-only** — `if (readClaudeCodeOAuthBlob()) logger.info(...)`, no provider contact — and the explicit paths are untouched: `refreshAnthropicToken` is still called at `credentials.ts:415` for account-store refresh, which is where a refresh belongs because a caller there wants the new token.

**I mutation-tested the new regression rather than trusting its name.** It stubs global `fetch` and asserts `expect(network).not.toHaveBeenCalled()`, alongside unchanged file bytes, unchanged `mtimeMs`, and unchanged directory contents — that last one is a nice touch, since it catches a probe that writes a lock or temp file rather than only one that talks to the network. Adding a single `fetch` call back into the probe fails it immediately:

```
× leaves external credentials and network untouched at expiry 0
× leaves external credentials and network untouched at expiry 1788575257632
AssertionError: expected "vi.fn()" to not be called at all, but actually been called 4 times
```

Parameterising over both an expired and a live expiry is the right coverage, since the old code only reached the network on the expired branch.

**One semantic change worth stating explicitly**, since it is easy to miss in a diff this shape. The log previously fired only when a *usable* token was obtained — valid, or successfully refreshed. It now fires whenever credentials are merely present, including expired or revoked ones. That is the correct trade for a read-only probe, and the new wording is honest about it ("Detected local Claude Code CLI credentials. The CLI manages refresh.") rather than continuing to imply a working subscription. Worth knowing that presence no longer implies validity if anything downstream ever keys off that log line.

**Verification:** the new suite is 3/3, and the full auth package is **213/213 across 19 files**, matching your stated count exactly. Your note about needing one worker is accurate — the parallel run trips an unrelated two-process adoption race; `--no-file-parallelism` is green.

Approving. Small diff, real bug, and the deleted invalid-grant bookkeeping is the tell that it was biting users.

— [claude-code-eliza]

PR #30586, opened 2026-09-05
fix(ci): classify staging activation failures accurately

lalalune:
Staging provisioning diagnostics misreported inline errors as external and mixed exception messages with stack-file names. Correct the storage discriminator, classify known failure messages, and report the selected account's lifecycle state plus sandbox/job ownership and lease facts. Output remains limited to closed categories, booleans, timestamps, counts, and allowlisted source locations. Raw errors and tenant identifiers remain private; ambiguous ownership yields unavailable target details.

Validation: the actual diagnostic SQL executes against PostgreSQL/PGlite with competing tenants, ambiguous ownership, inline and external errors, account deletion state, and lease ownership. A stack referencing `with-timeout.ts` or the account-fence module cannot turn an unrelated exception into a timeout or account restriction. After rebasing on `89b5d92840dfa462653093651d429c7236912833`, the focused diagnostic and CI setup suites passed 11 tests and 43 assertions. Actionlint, Biome, `bun install`, and root `bun run verify` passed (376/376 tasks plus repository audits).

Live staging diagnostic output will be collected through the protected workflow after merge. No deployment or activation success is claimed by this change.

Refs #30552.

### DIFF
--- a/.github/workflows/live-smoke.yml
+++ b/.github/workflows/live-smoke.yml
@@ -924,7 +924,7 @@ jobs:
               WHEN error ~ 'canceling statement due to statement timeout' THEN 'database_statement_timeout'
               WHEN error ~ 'Reviewed backup authority changed' THEN 'reviewed_backup_changed'
               WHEN error ~ 'Reviewed fresh-boot authority changed' THEN 'reviewed_fresh_boot_changed'
-              WHEN error ~* 'timed out|timeout' THEN 'timeout'
+              WHEN split_part(error, E'\n', 1) ~* 'timed out|timeout' THEN 'timeout'
               ELSE 'unclassified' END FROM latest_job),
             'organizationActive', (SELECT is_active FROM organization),
             'accountLifecycleState', (SELECT account_lifecycle_state FROM organization),
--- a/.github/workflows/live-smoke.yml
+++ b/.github/workflows/live-smoke.yml
@@ -880,7 +880,8 @@ jobs:
             SELECT organization_id, user_id FROM agent_sandboxes
             WHERE agent_name = 'managed-dedicated-canary-' || :'suffix'
           ), target AS (
-            SELECT agent.id, agent.organization_id, agent.user_id
+            SELECT agent.id, agent.organization_id, agent.user_id, agent.status,
+              agent.lifecycle_job_id, agent.lifecycle_execution_generation, agent.deletion_attempt_id
             FROM canary
             JOIN personal_dedicated_upgrade_authorities AS authority
               ON authority.organization_id = canary.organization_id AND authority.user_id = canary.user_id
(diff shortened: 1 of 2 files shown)


ss251 (approved):
Reviewed head 9d73d68077962d11d72ef258426bb795ec390cb4.

The risky part is the `failureKind` CASE added in `.github/workflows/live-smoke.yml:913` plus the `errorTerms` narrowing to `split_part(error,E'\n',1)`. I traced the ordering that usually breaks this kind of classifier: the specific `Account lifecycle fenced provisioning job` / `generation is no longer current` / `conflicting .* job` branches all sit before the generic `~* 'timed out|timeout'` first-line catch-all, so a fenced error containing "timeout" later in the stack cannot be misclassified. The first-line split also keeps `errorTerms` closed-vocabulary — the test's `timeout exceeded when trying to connect PRIVATE_CREDENTIAL https://private.example\n at ...pg-pool/index.js` yields only `["connect","exceeded","timeout","trying"]` with `pool` correctly dropped and no `PRIVATE` leak, matching `packages/scripts/__tests__/staging-activation-job-origin.test.ts:58`.

Recommendation: merge

— [muse-lane-ss251]

attentionhead (approved):
Approving at `9d73d680`, with one piece of live information first: **this PR now conflicts with `develop`** (`mergeStateStatus: DIRTY`) at an unchanged head, so `develop` moved under it rather than anything being force-pushed. It needs a rebase before it can land. Everything below was verified at this head and should survive a clean rebase, since the conflict is not in the logic I checked.

**I independently reproduced your readiness diagnosis, and it is exactly right.**

I had looked at this same failure from the other end recently — the `Subscription authority PostgreSQL` job dying with `error: the database system is starting up` — and identified it as a service-container startup race without finding the mechanism. You found the mechanism, and it is more specific than "flaky": the official `postgres` entrypoint starts a *temporary* initialization server that listens on the Unix socket only, so `pg_isready -U postgres` (which defaults to the socket) reports ready during initdb, before the real server exists.

Rather than take that on faith I ran `postgres:16-alpine` locally and raced both probes from container start:

```
socket probe first passed at:  775ms
TCP probe    first passed at: 1010ms
socket-ready-before-TCP window: 235ms
```

And connecting for real inside that window:

```
psql -h 127.0.0.1 -U postgres -c 'select 1'
psql: error: connection to server at "127.0.0.1", port 5432 failed: Connection refused
```

So the old probe genuinely reports success during a window when nothing is listening on TCP. To be precise about what I did and did not reproduce: I observed `Connection refused`, whereas CI observed `the database system is starting up` — that is the adjacent state a moment later, once the real server binds TCP but is still starting. Both are the same root cause, and `pg_isready` over TCP returns non-zero for both (it reports a rejecting server, not a ready one), so the one-flag fix closes the whole window rather than just the half I happened to reproduce.

235 ms is comfortably wide relative to a `sleep 1` poll on a loaded runner, which is why this presented as intermittent rather than constant.

The inline comment carries the reasoning, which matters here — `-h 127.0.0.1` looks like a no-op to anyone who does not know the entrypoint detail, and is exactly the kind of thing a future cleanup deletes.

**On the diagnostics half**, the fix I care about most is the storage discriminator:

```diff
-'errorStoredExternally', (SELECT error_storage IS NOT NULL FROM latest_job),
+'errorStoredExternally', (SELECT error_storage = 'r2' FROM latest_job),
```

`IS NOT NULL` reports *any* recorded storage mode as external, so an inline error was being labelled external and its detail then withheld as if it lived in R2 — the failure mode where a diagnostic actively misleads the person reading it. Comparing against the actual value is correct.

Classifying `failureKind` through a closed set of regexes over known messages, rather than surfacing the raw error, keeps the output within the allowlisted-categories contract while still being useful. The `split_part(error, E'\n', 1)` fallback restricting the generic timeout match to the first line is the right instinct — matching `timeout` anywhere in a multi-line error would catch stack frames and misclassify.

**Tests:** `staging-activation-job-origin.test.ts` passes (1/1, 8 assertions) against PGlite at this head.

**What I did not verify:** the `live-smoke.yml` changes beyond reading them, and the new lifecycle/ownership fields against a real staging database — the PGlite test covers their shape, not their production values.

Approving on the verification above; please rebase onto current `develop` before merging.

— [claude-code-eliza]

PR #30589, opened 2026-09-05
fix(evidence): report newly retired invalid queue jobs as failures

### LINKED ISSUE #30576: fix(evidence): report invalid queued jobs in worker outcomes
`FileJobQueue.claim()` retires malformed pending files and writes failed result records internally. The worker never observes those outcomes: a queue containing invalid JSON reports zero failed jobs, and the one-shot CLI exits successfully despite writing a failed result.

Expose the newly retired invalid result to the claiming worker through an additive observation callback, count and emit its existing processed event exactly once, and return a nonzero CLI status for newly processed failures. Preserve atomic claiming, progress past invalid files, explicit skips, and the persisted schema. Do not count historical results or other workers' outcomes.

Acceptance: real filesystem/CLI tests cover all-invalid, mixed valid-invalid, historical failures and existing concurrency behavior; printed summary, exit status, events and persisted results agree. Evidence tests, typecheck and lint pass.

lalalune:
Malformed pending jobs were retired as failed on disk but omitted from the current worker totals, allowing the CLI to exit successfully. The worker now records and emits each newly retired invalid result once. The CLI exits with failure when the current run contains failed jobs; historical results do not affect a later run.

Closes #30576.

The change uses the existing queue state machine and durable retirement path. No new queue, scheduler, schema, or analyzer is introduced. Risk is low: only current-run outcome accounting and the corresponding CLI exit code change.

Validation: all 532 evidence tests passed, with two opt-in tests skipped; package typecheck and lint passed through root verification. The package has no standalone build. An independent reviewer reran the 22 focused real-filesystem queue tests and found no blocker. Root verification passed all 376 tasks again after rebasing onto develop at 89b5d92840dfa462653093651d429c7236912833 and running bun install. The rebased head is 1d5026b1b5f2fab680f833dca2f836d19dc49c69.

## Evidence Gate

N/A - filesystem queue and CLI behavior; no rendered UI changed.

N/A - filesystem queue and CLI behavior; no rendered UI changed.

N/A - the complete CLI outcome is captured in the transcript below.

<details><summary>Actual CLI drain followed by a second empty run</summary>

```text
[gpu-queue] worker draining /tmp/eliza-quality-review-20260904/queue-pr-proof at tier=gpu
[gpu-queue] failed 000-invalid — invalid queue job: not valid JSON
[gpu-queue] done: 0 completed, 1 failed, 0 skipped, 0 requeued
first exit: 1
[gpu-queue] worker draining /tmp/eliza-quality-review-20260904/queue-pr-proof at tier=gpu
[gpu-queue] done: 0 completed, 0 failed, 0 skipped, 0 requeued
second exit: 0
```

</details>

N/A - no browser or frontend request path changed.

N/A - no model, action, prompt, or provider behavior changed.

The malformed pending file was retired to done before the failure was counted. Its persisted result was inspected:

```json
{
  "schema": 1,
  "id": "000-invalid",
  "analyzerId": "unknown",
  "status": "failed",
  "completedAt": "2026-09-05T01:28:04.755Z",
  "reason": "invalid queue job: not valid JSON"
}

```

Known gaps: the two opt-in CLI test cases remain skipped in the standard package lane. The actual malformed-job CLI path was independently exercised above. Hosted checks must pass on the exact current head before merge.

Documentation: no public option or command changed. Maintainer CI exception: N/A - no exception requested.

### DIFF
--- a/packages/evidence/src/queue/cli.test.ts
+++ b/packages/evidence/src/queue/cli.test.ts
@@ -1,8 +1,4 @@
-// The evidence:gpu-queue CLI as a process boundary: argv parsing, usage/error
-// exit codes, `enqueue` writing a real pending job, and `worker --once` draining
-// to honest skip records when no vision service is reachable. Drives the real
-// runQueueCli with a captured CliIo (no spawned process); the library it calls
-// is the same one exercised elsewhere.
+/** Exercises queue CLI outcomes against actual files, including invalid jobs and unavailable analyzers. */
 
 import fs from "node:fs";
 import os from "node:os";
@@ -32,6 +28,36 @@ function captureIo() {
 }
 
 describe("runQueueCli", () => {
+  it("reports newly retired invalid jobs and exits nonzero without recounting historical failures", async () => {
+    const root = newRoot();
+    const queue = new FileJobQueue(root);
+    fs.writeFileSync(
+      path.join(root, "pending", "000-invalid.json"),
+      "{not JSON",
+    );
+    const first = captureIo();
+    expect(
+      await runQueueCli(["worker", "--root", root, "--once"], first.io),
+    ).toBe(1);
+    const result = queue.readResult("000-invalid");
+    expect(result).toMatchObject({ status: "failed", id: "000-invalid" });
+    expect(result?.reason).toBeTruthy();
+    expect(first.out.join("\n")).toContain(
+      "0 completed, 1 failed, 0 skipped, 0 requeued",
+    );
+    expect(fs.existsSync(path.join(root, "done", "000-invalid.json"))).toBe(
+      true,
+    );
+    const second = captureIo();
+    expect(
+      await runQueueCli(["worker", "--root", root, "--once"], second.io),
+    ).toBe(0);
+    expect(second.out.join("\n")).toContain(
+      "0 completed, 0 failed, 0 skipped, 0 requeued",
+    );
+    expect(queue.readResult("000-invalid")).toEqual(result);
+  });
+
   it("prints usage and exits 1 on an unknown command", async () => {
     const { io, err } = captureIo();
     expect(await runQueueCli(["bogus"], io)).toBe(1);
--- a/packages/evidence/src/queue/cli.ts
+++ b/packages/evidence/src/queue/cli.ts
@@ -126,7 +126,7 @@ async function runWorkerCommand(argv: string[], io: CliIo): Promise<number> {
   io.out(
     `[gpu-queue] done: ${counts.completed} completed, ${counts.failed} failed, ${counts.skipped} skipped, ${counts.requeued} requeued`,
   );
-  return 0;
+  return counts.failed > 0 ? 1 : 0;
 }
 
 function runEnqueueCommand(argv: string[], io: CliIo): number {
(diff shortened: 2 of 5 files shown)


ss251 (approved):
Reviewed head 1d5026b1b5f2fab680f833dca2f836d19dc49c69.

The risky part is `packages/evidence/src/queue/file-queue.ts:126` where `claim(onInvalid?)` now returns the `recordInvalid` result to the caller, wired at `packages/evidence/src/queue/worker.ts:255` to `counts.failed += 1` plus a `processed/failed` emit. I traced the two boundaries that usually break this accounting: the `continue` after `onInvalid?.(result)` keeps draining past poison jobs so the mixed-mode valid job still reaches `skipped`, and the callback fires only on the pending->processing->done transition, so the second `worker --once` with empty pending emits only `idle` with `0 failed` instead of recounting `done/000-invalid.json`. That matches the narrowed `cli.ts:129` `counts.failed>0 ? 1 : 0` exit branch.

No blocking concerns. Verified by reading the diff at this head and following the claim/retire/emit path.

Recommendation: merge

— [muse-lane-ss251]

attentionhead (approved):
Approving at `1d5026b1`. I reproduced the bug end-to-end against a real queue directory rather than reading the diff, because a verification tool that exits 0 while failing is the one bug class where the tests passing is not reassurance — the whole point is that the failure was invisible.

**The bug reproduces exactly as described.** Same malformed job file, same command, both heads:

```
develop:  [gpu-queue] done: 0 completed, 0 failed, 0 skipped, 0 requeued   exit 0
this PR:  [gpu-queue] done: 0 completed, 1 failed, 0 skipped, 0 requeued   exit 1
```

And on `develop` the failure *was* durably written — `results/badjob.json` contains `"status": "failed"` — while the run reported zero failures. That is the precise shape you describe: retired honestly on disk, omitted from the current totals, CLI exits green. A queue drain that poison-jobs its way through every enqueued item would have reported a clean run.

**I checked the history-isolation claim too**, since `counts.failed > 0` would be a trap if the count were ever derived from the results directory. Running the worker a second time over the same root, with the failed result from the previous run still present, gives `0 failed` and exit 0. So the exit code reflects the current run only, as claimed.

**The fix is broader than the title admits, and I think that is worth saying.** The title is about *newly retired invalid* jobs, but the operative line is:

```diff
-  return 0;
+  return counts.failed > 0 ? 1 : 0;
```

The old code returned `0` unconditionally, so a legitimately failing analyzer — nothing to do with malformed JSON — also exited green. This PR fixes that class as well. Anyone skimming the title to judge blast radius will underestimate it: this changes the exit code for *every* failure path, not just poison jobs. That is the right change, but it deserves to be visible in the title or at least called out first in the body, because it is the part most likely to surface a previously-hidden red in someone's pipeline.

**On the observer plumbing**, threading `onInvalid` through `claim()` rather than having the worker rescan is the right call: the invalid job is retired inside `claim` and never becomes a `ClaimedJob`, so there is no later point where the worker could observe it without re-reading the results directory — which is exactly the coupling that would have broken history isolation. Returning the `JobResult` from `recordInvalid` instead of reconstructing it in the worker keeps one source of truth for the reason string, so the emitted event and the durable record cannot drift.

The `onInvalid?.(result)` call sits after `recordInvalid` completes, so the observer only ever sees results that are already durable — matching the comment you added, and meaning a crash between retire and emit under-reports rather than reporting a failure that was never written.

**Tests:** the full `packages/evidence/src/queue/` suite is 53 tests, 0 failures at this head.

One process note: verifying this needed a fresh queue directory per run. My first attempt reused one, and the second invocation drained an already-empty queue and exited 0, which looks exactly like the bug not being fixed. Worth knowing if anyone else reproduces it by hand.

Approving.

— [claude-code-eliza]

PR #30591, opened 2026-09-05
fix(auth): validate Codex device-login credential files

### LINKED ISSUE #30588: fix(auth): validate Codex device-login credential files
## Problem

`startCodexDeviceLogin` parses the CLI-generated `auth.json` with a TypeScript assertion and then relies on truthiness for required tokens. The file is an external process boundary, so a successful Codex child process can currently return a credential record whose `refresh` or optional `idToken` value is not a string. Blank refresh tokens are also accepted, and a very large finite JWT `exp` can overflow the stored millisecond expiry to `Infinity`.

## User impact

The device flow can report success and persist malformed credentials that fail later during refresh or serialization. An infinite expiry can suppress required reauthentication indefinitely. These failures appear after the login that caused them, making recovery and diagnosis harder.

## Acceptance criteria

- Parse the device-login `auth.json` value as `unknown` and validate the root and `tokens` objects before constructing credentials.
- Require `access_token` and `refresh_token` to be nonblank strings.
- Treat `id_token` as optional/nullish and preserve the existing empty-string-as-absent behavior, while rejecting other present runtime types.
- Require the JWT `exp` claim and computed millisecond expiry to be positive finite numbers.
- Translate invalid files through the existing typed `codex_device.credentials_invalid` / `codex_device.invalid_access_token` terminal outcomes without exposing credential values.
- Add deterministic regression tests using the existing mocked-child-process and real-temporary-directory harness for malformed token types, blank required tokens, optional ID-token compatibility, and expiry overflow.
- Pass the auth package tests, typecheck, lint/format checks, build, and root `bun run verify`.

## Scope

`packages/auth/src/codex-device.ts` and its focused tests. No OAuth endpoint, UI, storage schema, or public credential shape changes.

## Existing work checked

Open PR #28606 is test-only and covers malformed JSON, missing tokens, invalid JWT claims, and ANSI prompt parsing. It does not validate wrong runtime types or change the production boundary, so this is a concrete production defect rather than a duplicate test-coverage request.

wtfsayo:
Closes #30588.

Reject malformed Codex device-login credential files before constructing a login result. Access and refresh tokens must be nonblank strings; supplied ID tokens must be strings; explicit expiry must be positive and finite, including conversion to milliseconds. Existing valid optional-ID-token behavior is preserved.

Independent review traced the credential file reader, device-flow completion, and persisted credential consumers. Real temporary-file tests exercise successful and rejected inputs through the device login boundary.

Validation after rebase onto `f567055f1c3656590369e6f9c416c253a773f283`, using Bun 1.3.14 and Node 24.15.0:
- `bun install`: passed.
- Full auth suite: 19 files, 226 tests passed.
- Auth typecheck, lint, and format checks: passed.
- Root `bun run verify`: 376/376 workspace tasks and all repository audits passed.

UI screenshots/video, frontend logs, native capture, and model trajectories are N/A: this change only validates a local credential file. Hosted checks must pass at the final head before merge.

### DIFF
--- a/packages/auth/src/codex-device.test.ts
+++ b/packages/auth/src/codex-device.test.ts
@@ -75,6 +75,18 @@ function accessToken(exp: number): string {
   return `header.${Buffer.from(JSON.stringify({ exp })).toString("base64url")}.signature`;
 }
 
+async function deviceLoginFromAuthJson(authJson: unknown) {
+  const child = mockChild();
+  const start = startCodexDeviceLogin();
+  const codexHome = spawnedCodexHome();
+  child.process.emit("spawn");
+  emitPrompt(child);
+  const flow = await start;
+  writeFileSync(path.join(codexHome, "auth.json"), JSON.stringify(authJson));
+  child.process.emit("close", 0, null);
+  return { credentials: flow.credentials, codexHome };
+}
+
 afterEach(() => {
   for (const temporaryHome of temporaryHomes) {
     if (existsSync(temporaryHome)) {
@@ -319,6 +331,110 @@ describe("startCodexDeviceLogin", () => {
     await expect(flow.credentials).rejects.toMatchObject({
       code: "codex_device.process_failed",
       context: { exitCode: null, signal: "SIGTERM" },
+    });
+    expect(existsSync(codexHome)).toBe(false);
+    expect(rmSyncMock).toHaveBeenCalledTimes(1);
+  });
+
+  it.each([
+    ["non-object root", null],
+    ["non-object token block", { tokens: [] }],
+    [
+      "numeric access token",
+      { tokens: { access_token: 7, refresh_token: "refresh" } },
+    ],
+    [
+      "blank access token",
+      { tokens: { access_token: " ", refresh_token: "refresh" } },
+    ],
+    [
+      "numeric refresh token",
+      {
+        tokens: {
+          access_token: accessToken(2_000_000_000),
+          refresh_token: 7,
+        },
+      },
+    ],
+    [
+      "blank refresh token",
+      {
+        tokens: {
+          access_token: accessToken(2_000_000_000),
+          refresh_token: " ",
+        },
+      },
+    ],
+    [
+      "numeric ID token",
+      {
+        tokens: {
+          access_token: accessToken(2_000_000_000),
+          refresh_token: "refresh",
+          id_token: 7,
+        },
+      },
+    ],
+  ])("rejects auth.json with a %s", async (_name, authJson) => {
+    const { credentials, codexHome } = await deviceLoginFromAuthJson(authJson);
+    await expect(credentials).rejects.toMatchObject({
+      code: "codex_device.credentials_invalid",
+    });
+    expect(existsSync(codexHome)).toBe(false);
+    expect(rmSyncMock).toHaveBeenCalledTimes(1);
+  });
+
+  it.each([
+    ["zero", 0],
+    ["negative", -1],
+    ["overflowing", 1e308],
+  ])("rejects an access token with a %s exp claim", async (_name, exp) => {
+    const { credentials, codexHome } = await deviceLoginFromAuthJson({
+      tokens: {
+        access_token: accessToken(exp),
+        refresh_token: "refresh",
+      },
+    });
+    await expect(credentials).rejects.toMatchObject({
+      code: "codex_device.invalid_access_token",
+    });
+    expect(existsSync(codexHome)).toBe(false);
+    expect(rmSyncMock).toHaveBeenCalledTimes(1);
+  });
+
+  it.each([
+    ["null", null],
+    ["empty", ""],
+  ])("omits an %s optional ID token", async (_name, idToken) => {
+    const { credentials, codexHome } = await deviceLoginFromAuthJson({
+      tokens: {
+        access_token: accessToken(2_000_000_000),
+        refresh_token: "refresh",
+        id_token: idToken,
+      },
+    });
+    await expect(credentials).resolves.toEqual({
+      access: accessToken(2_000_000_000),
+      refresh: "refresh",
+      expires: 2_000_000_000_000,
+    });
+    expect(existsSync(codexHome)).toBe(false);
+    expect(rmSyncMock).toHaveBeenCalledTimes(1);
+  });
+
+  it("preserves a whitespace-only ID token for compatibility", async () => {
+    const { credentials, codexHome } = await deviceLoginFromAuthJson({
+      tokens: {
+        access_token: accessToken(2_000_000_000),
+        refresh_token: "refresh",
+        id_token: " ",
+      },
+    });
+    await expect(credentials).resolves.toEqual({
+      access: accessToken(2_000_000_000),
+      refresh: "refresh",
+      expires: 2_000_000_000_000,
+      idToken: " ",
     });
     expect(existsSync(codexHome)).toBe(false);
     expect(rmSyncMock).toHaveBeenCalledTimes(1);
--- a/packages/auth/src/codex-device.ts
+++ b/packages/auth/src/codex-device.ts
@@ -65,15 +65,31 @@ export function codexDeviceLoginUsesShell(
   return platform === "win32";
 }
 
+function isRecord(value: unknown): value is Record<string, unknown> {
+  return value !== null && typeof value === "object" && !Array.isArray(value);
+}
+
+function isNonBlankString(value: unknown): value is string {
+  return typeof value === "string" && value.trim().length > 0;
+}
+
 function expiryFromJwt(token: string): number {
   try {
-    const payload = JSON.parse(
+    const payload: unknown = JSON.parse(
       Buffer.from(token.split(".")[1] ?? "", "base64url").toString("utf8"),
-    ) as { exp?: unknown };
-    if (typeof payload.exp !== "number" || !Number.isFinite(payload.exp)) {
-      throw new Error("Codex access token has no finite exp claim");
+    );
+    if (!isRecord(payload)) {
+      throw new Error("Codex access token payload must be an object");
     }
-    return payload.exp * 1000;
+    const exp = payload.exp;
+    if (typeof exp !== "number" || !Number.isFinite(exp) || exp <= 0) {
+      throw new Error("Codex access token has no positive finite exp claim");
+    }
+    const expires = exp * 1000;
+    if (!Number.isFinite(expires) || expires <= 0) {
+      throw new Error("Codex access token expiry exceeds the numeric range");
+    }
+    return expires;
   } catch (cause) {
     // error-policy:J2 context-adding rethrow — fabricating an expiry can make an
     // invalid token look healthy and delay required reauthentication.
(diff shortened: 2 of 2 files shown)


ss251 (approved):
Reviewed head 67bca7e514cc62f878c231f7195e1c1a1437a1f8.

The risky hunk is `packages/auth/src/codex-device.ts:268-286`, where untrusted `auth.json` is now gated by `isRecord`/`isNonBlankString` before `expiryFromJwt`. I traced the boundaries the old `if (!access || !refresh)` missed: numeric `access_token: 7` and `" "` are now rejected since `isNonBlankString` trims, array `tokens: []` fails `isRecord`, and `id_token: 7` throws while `null`/`""` collapse via `rawIdToken || undefined` but `" "` is deliberately preserved for compatibility. The `expiryFromJwt` tightening to `exp > 0` plus the `exp * 1000` finiteness check correctly rejects the `0`/`-1`/`1e308` cases the new tests cover.

No blocking concerns; whitespace-ID-token asymmetry is intentional, non-blocking.

Recommendation: merge

— [muse-lane-ss251]

attentionhead (approved):
Approving at `67bca7e5`. This is the third PR in this validation series I have reviewed, so I checked the two things that actually distinguish good tightening from bad: whether it reintroduces the family's known regression, and whether it still accepts credential files the real Codex CLI writes.

**No regression of the null-rotation defect.** The `id_token` branch treats both absent forms as absent:

```ts
if (rawIdToken !== undefined && rawIdToken !== null && typeof rawIdToken !== "string") {
  throw new Error("Codex device login returned an invalid ID token");
}
```

That is the lesson from #30566 — where adding validation turned a legitimate `null` into a hard failure — applied correctly here rather than relearned. I mutation-tested it: deleting `rawIdToken !== null &&` gives **1 failed, 21 passed**, and the failure is `omits an null optional ID token`. So the guard is pinned by a test, not merely present. 22/22 as shipped.

**I verified the tightening against a real Codex credential file rather than only fixtures.** This is validation at an external boundary, so the failure mode that matters is rejecting something the CLI legitimately produces — and fixtures written alongside the validator cannot catch that, because they encode the same assumptions. Checking a live `~/.codex/auth.json` on this machine structurally (key names and types only, no values):

```
top-level keys: auth_mode, OPENAI_API_KEY, tokens, last_refresh
tokens isRecord: true
  id_token: string   access_token: string   refresh_token: string   account_id: string
access / refresh pass isNonBlankString: true
id_token accepted by new guard: true
```

and decoding its access token's payload, the `exp` claim is a positive finite number and `exp * 1000` stays in range, so the new `exp <= 0` and overflow checks accept it too.

Two things that matters beyond "it works". First, the real file carries `account_id`, `auth_mode`, `OPENAI_API_KEY` and `last_refresh` — fields the validator never mentions — and `isRecord` plus per-field checks correctly ignore them. A stricter shape check that rejected unknown keys would have broken every real login, and it is worth being explicit that this deliberately does not. Second, `auth.json` holds an `OPENAI_API_KEY` alongside the OAuth tokens, which is a good reason the error paths here stay generic rather than echoing file contents.

**On the rest of the tightening.** Rejecting `exp <= 0` is right — an expiry at or before the epoch cannot produce a usable credential, and the old code accepted it, then produced a token that looked healthy and was already expired. The separate `Number.isFinite(expires)` check after multiplying catches a large `exp` overflowing to `Infinity`, which would otherwise persist as a never-expiring token. Both match the reasoning in the sibling PRs.

Behaviour for `access`/`refresh` is essentially unchanged in practice: the old `!access || !refresh` already rejected empty strings, and `isNonBlankString` additionally rejects whitespace-only. `isRecord(parsed.tokens)` rejecting `null` produces the same outcome the old optional chaining did, just with a clearer message.

**One suggestion, not a request.** `isRecord` is now defined in both `codex-usage.ts` and `codex-device.ts`, and your two in-flight siblings (#30566, #30584) each add their own copies of `isRecord` and `isNonBlankString` to `vendor/pi-oauth`. Once all three land there will be four independent definitions of the same two predicates in one package. They are trivial, but they are also the exact predicates that encode this series' security decisions — the `null`-versus-`undefined` distinction being the one that already bit once. A single shared module would make the next change to that rule apply everywhere at once instead of three-quarters of the way.

Approving.

— [claude-code-eliza]

agent-cortex (approved):
**Approve** — independently reproduced the defect and the fix in a sandbox (network-denied), inspected the boundary logic, and confirmed the added test coverage is meaningful.

## Scope
Diff is confined to the two files the PR claims (`packages/auth/src/codex-device.ts`, `codex-device.test.ts`); no unrelated surface, no new exports, no dependency/manifest/CI changes. It tightens the untrusted `auth.json` credential-file boundary produced by the Codex CLI, matching issue #30588.

## What I verified (exact head `67bca7e5`, base merge-base `36181c29`)

Setup: `bun install --frozen-lockfile --ignore-scripts`, then `node packages/shared/scripts/generate-keywords.mjs` and `bun run --cwd packages/prompts build:package` (the auth vitest config aliases core/shared/logger/vault to source but resolves `@elizaos/prompts` from its dist).

1. **Head is green** — `vitest run src/codex-device.test.ts` → `22 passed (22)`.

2. **Red-before-fix** — reverting *only* `codex-device.ts` to the base version (`git show 36181c29:…/codex-device.ts`) while keeping the new tests → **8 fail / 14 pass**:
   ```
   × rejects auth.json with a numeric access token
   × rejects auth.json with a blank access token
   × rejects auth.json with a numeric refresh token
   × rejects auth.json with a blank refresh token
   × rejects auth.json with a numeric ID token
   × rejects an access token with a zero exp claim
   × rejects an access token with a negative exp claim
   × rejects an access token with a overflowing exp claim
   ```
   The failure payloads confirm the *real-world* consequence, not just a shape mismatch: on base an overflowing `exp` resolves to a credential with `expires: Infinity`, and a negative `exp` resolves to `expires: -1000` — i.e. the old `if (!access || !refresh)` gate let a malformed token be accepted as a healthy credential, which per the J2 comment "can make an invalid token look healthy and delay required reauthentication." The head correctly rejects all eight with the typed `codex_device.invalid_access_token` / `credentials_invalid` terminal codes.

   Note the two `non-object root` / `non-object token block` cases pass on base as well — the base gate already caught fully-missing tokens; the new value proven here is precisely the type/blank/expiry-range hardening.

3. **Cleanup contract holds** — every rejecting case still asserts `existsSync(codexHome) === false` and `rmSyncMock` called once; the temp `CODEX_HOME` is removed on the malformed path, so no malformed credential artifact is persisted.

## Boundary/logic review
- `expiryFromJwt` now guards object-payload, positive-finite `exp`, and the `exp*1000` overflow to `Infinity` — the overflow case (`1e308`) is the one the base silently accepted; verified rejected.
- ID-token handling is compatibility-preserving as documented: `null`/`""` omit `idToken`, non-string present values reject, whitespace-only (`" "`) is retained (truthy). Test coverage matches each branch.

## Notes (non-blocking)
- In-sandbox `bun run --cwd packages/auth typecheck` fails only on pre-existing unbuilt-workspace declaration resolution (`@elizaos/core/client-public`, `@elizaos/cloud-routing`, `@elizaos/shared/contracts/first-run-options`); none reference `codex-device.ts`, so this is environmental (needs a full `bun run verify`/dist build), not a regression from this diff.
- This is one of a series of sibling validation-hardening PRs (Anthropic/OpenAI OAuth). The same untrusted-JSON pattern in the sibling token responders is out of scope here and addressed in the companion PRs.

No correctness, security, test, or required-evidence gap found. The evidence gate's N/A rows are appropriate (no rendered surface; the file/child-process boundary is exercised by the real parser + real temp dirs with a mocked CLI, which is the correct harness).

---
— [eliza-swarm]

nothingxnowhere (approved):
Reviewed at head `67bca7e514` against `develop`. Ran the suite rather than reading the guards: at head `22 pass`; with the PR's tests kept and `codex-device.ts` reverted to develop, `8 fail / 14 pass`, and the split is the useful part.

The 8 that fail on develop are exactly the new rejections — numeric or blank `access_token` / `refresh_token`, a numeric `id_token`, and `exp` of `0`, `-1` or `1e308`. The other 14, including the null/empty `id_token` omissions and the whitespace-only `id_token` pass-through, already hold on develop. So the PR does not move the accepted shape of a real `auth.json` (the Codex CLI writes string tokens and a positive integer `exp`); it only closes what develop let through: `if (!access || !refresh)` accepted `7` and `" "`, and `expiryFromJwt` accepted `exp: 0` — a token that reads as expired-at-epoch rather than invalid — and `1e308`, which multiplies to `Infinity` and never expires. `expiryFromJwt` has one caller on develop (`:250`), so the tightened `exp` contract has no second consumer to break.

One relabel worth knowing rather than objecting to: a numeric `access_token` used to get past the truthiness check, reach `expiryFromJwt`, throw on `.split`, and surface as the JWT error — the outer catch passes an `ElizaError` cause through unchanged (`:255-263` on develop). It now fails earlier as `codex_device.credentials_invalid`, which is the right bucket for a malformed file. `CODEX_HOME` cleanup still runs exactly once on every rejection path per the `rmSyncMock` assertions.

Approving.

```json
{
  "schemaVersion": "2",
  "headSha": "67bca7e514",
  "recommendation": "accept",
  "reproduced": true,
  "risks": "none",
  "effortBand": "low",
  "complexity": "simple",
  "impact": "medium",
  "reviewLoad": "standard",
  "recommendedTier": "small",
  "recommendedThirds": 3,
  "workUnitId": "wu_elizaos_eliza_pr_30591",
  "valueRationale": "Ran the suite at head (22 pass) and against develop's codex-device.ts (8 fail / 14 pass), confirming only the new rejections are new and the accepted shape of a real auth.json is unchanged; traced the numeric-token error-code relabel through develop's outer catch.",
  "usefulArtifacts": ["RED/GREEN run of codex-device.test.ts against develop and head"],
  "commands": ["bunx vitest run src/codex-device.test.ts   # head: 22 pass; develop source + PR tests: 8 failed | 14 passed"],
  "evidenceUrls": [],
  "summary": "Codex device-login auth.json now requires non-blank string tokens, a string-or-absent id_token, and a positive finite exp; the control shows exactly the eight new rejections fail on develop while the compatibility cases already held."
}
```

— [nxn-claude-code]

byt61 (approved):

- PR suite + PR source: 22/22 pass under the package vitest runner.
- Mutant check (PR tests vs pre-PR source): 8 fail, confirming the new rejection cases genuinely guard the validation rather than passing vacuously. The remaining new cases are behavior-preserving (null/empty/whitespace id_token), which correctly pass on both.
- Error-code mapping checked by reading `codex-device.ts:255-269`: new plain-Error throws translate to `codex_device.credentials_invalid`, and `expiryFromJwt` ElizaErrors pass through as `codex_device.invalid_access_token`, matching the asserted codes. The exp overflow guard (`1e308 * 1000`) and the non-positive-exp rejection close real holes in the old `payload.exp * 1000` path.

lalalune (approved):
Reviewed the credential-file parser, device-login lifecycle, OAuth-flow caller, and regression tests. This is a useful boundary fix: malformed auth.json data now produces a typed terminal failure and the temporary credential directory is still removed exactly once. Valid credentials and the established optional ID-token behavior remain compatible. No implementation repair was needed.

Rebased and validated head `bf1cae56b96d2295d4cd5e45c60e4ba0e84ff348`, based on `f567055f1c3656590369e6f9c416c253a773f283`, with Node 24.15.0 and Bun 1.3.14:

- Full auth suite: 19 files, 226 tests passed.
- Auth typecheck, lint:check, and format:check passed.
- `bun install` passed.
- Root `bun run verify`: all 376 workspace tasks and all repository audits passed.

The tests exercise real isolated credential files and cleanup with a mocked child-process lifecycle. No live provider login or user-account mutation is claimed. No UI, model, or prompt behavior changes.

PR #30592, opened 2026-09-05
fix(agent): preserve project files through failed snapshot restoration

### LINKED ISSUE #30578: fix(agent): preserve VFS state across snapshot and restore failures
Workbench VFS snapshot listing hides filesystem read failures as empty results and accepts unchecked persisted metadata. Restore deletes the live directory before copying the saved tree, so a filesystem failure can destroy the current workspace. Snapshot, restore, Git, and shell mutations also bypass the existing write serialization.

Acceptance:
- Distinguish absent snapshots from unreadable or invalid metadata with typed errors visible at the workbench API.
- Validate complete saved state before touching the live workspace; publish snapshots only after their files and metadata are complete.
- Stage restores and preserve/recover the prior live tree across copy, swap, metadata-write failures and interrupted operations.
- Share project serialization across VFS operations and its Git/shell writers.
- Prove real filesystem faults, invalid-input non-mutation, cross-instance concurrency, and recovery using focused service and route tests.

Keep the existing per-project on-disk VFS and snapshot formats; this is a consistency repair, not a replacement filesystem architecture.

lalalune:
Snapshot restoration previously removed live project data before a fallible copy, and unreadable snapshot metadata could appear as an empty list. Snapshot creation and restoration now validate and stage filesystem state before publication, serialize through the existing project queue, and recover interrupted restores using a durable commit marker. Corrupt recovery evidence produces an explicit error; a missing snapshot retains its 404 contract.

Closes #30578.

The same queue now covers VFS reads and mutations, Git operations, and builtin shell directory creation. Public snapshot APIs remain compatible. The main risk is restore ordering; real filesystem tests cover all four interruption phases, failed publication, corrupt metadata, invalid/symlink trees, and concurrent service instances.

Validation: 70 focused VFS/Git/shell/workbench tests passed on this scoped head. The full agent lane executed all 758 files in 48 batches; its three failures were an unbuilt dependency and two timing/teardown failures under shared-host load. After building the dependency, all three entire failing files passed (27 tests). An independent reviewer examined the staged data, backup, commit marker, and recovery ordering and found no remaining blocker. After rebasing onto develop at d00c610abca5d716754b9d96ff7d30b99cf7a668 and running bun install, scoped root verification passed all 376 tasks again.

## Evidence Gate

N/A - host filesystem and HTTP storage contract; no rendered view changed.

N/A - host filesystem and HTTP storage contract; no rendered view changed.

N/A - the actual filesystem sequence is captured in the transcript below.

<details><summary>Real temporary project: corruption, failed restore, reopen, and successful retry</summary>

```jsonl
{"case":"unreadable snapshot","error":{"code":"VFS_STORAGE_FAILED","cause":"EISDIR"},"live":"current bytes"}
{"case":"commit failure","error":"ROLLBACK_FAILED","reopenedLive":"current bytes"}
{"case":"successful retry","live":"saved bytes","marker":{"snapshotId":"2026-09-05T01-43-45-366Z-6490e877","projectId":"review","rolledBackAt":"2026-09-05T01:43:45.388Z","previousSnapshotId":"2026-09-05T01-43-45-387Z-fff62b13"},"root":"/var/folders/1g/77s889gx10n7mtl6z1nfrxzm0000gn/T/eliza-vfs-final-K9dqGw"}
```

</details>

N/A - no browser request behavior changed. Real workbench route tests prove malformed metadata becomes HTTP 500 and missing snapshots remain HTTP 404.

N/A - no model, prompt, provider, or action result changed.

Actual file contents and durable marker after reopening and retrying:

```jsonl
{"case":"unreadable snapshot","error":{"code":"VFS_STORAGE_FAILED","cause":"EISDIR"},"live":"current bytes"}
{"case":"commit failure","error":"ROLLBACK_FAILED","reopenedLive":"current bytes"}
{"case":"successful retry","live":"saved bytes","marker":{"snapshotId":"2026-09-05T01-43-45-366Z-6490e877","projectId":"review","rolledBackAt":"2026-09-05T01:43:45.388Z","previousSnapshotId":"2026-09-05T01-43-45-387Z-fff62b13"},"root":"/var/folders/1g/77s889gx10n7mtl6z1nfrxzm0000gn/T/eliza-vfs-final-K9dqGw"}
```

The complete persistence tests also inspect saved and live directory contents after each interrupted restore phase.

Documentation: no new public command or option. Known gaps: the failure injection uses real temporary filesystems; it does not simulate machine power loss. Hosted checks must finish on the current head before merge. Maintainer CI exception: N/A.

### DIFF
--- a/packages/agent/src/api/workbench-vfs-routes.test.ts
+++ b/packages/agent/src/api/workbench-vfs-routes.test.ts
@@ -132,6 +132,52 @@ describe("workbench VFS routes", () => {
     });
   });
 
+  it("returns storage failures for unreadable or invalid snapshots instead of an empty list", async () => {
+    await callRoute("POST", "/api/workbench/vfs/projects", {
+      projectId: "broken-snapshot",
+    });
+    const saved = await callRoute<SnapshotResponse>(
+      "POST",
+      "/api/workbench/vfs/projects/broken-snapshot/snapshots",
+      {},
+    );
+    const metadata = path.join(
+      tmpDir,
+      "agent-vfs",
+      "projects",
+      "broken-snapshot",
+      "snapshots",
+      saved.body.snapshot.id,
+      "snapshot.json",
+    );
+    await fsp.rm(metadata);
+    await fsp.mkdir(metadata);
+    const unreadable = await callRoute<ErrorResponse>(
+      "GET",
+      "/api/workbench/vfs/projects/broken-snapshot/snapshots",
+    );
+    expect(unreadable.status).toBe(500);
+    expect(unreadable.body.error).toMatch(/read VFS metadata/);
+    expect(unreadable.body).not.toHaveProperty("snapshots");
+    await fsp.rm(metadata, { recursive: true });
+    await fsp.writeFile(
+      metadata,
+      JSON.stringify({ id: saved.body.snapshot.id }),
+    );
+    const invalid = await callRoute<ErrorResponse>(
+      "POST",
+      "/api/workbench/vfs/projects/broken-snapshot/rollback",
+      { snapshotId: saved.body.snapshot.id },
+    );
+    expect(invalid.status).toBe(500);
+    expect(invalid.body.error).toMatch(/Invalid saved snapshot metadata/);
+    const missing = await callRoute<ErrorResponse>(
+      "GET",
+      "/api/workbench/vfs/projects/broken-snapshot/snapshots/missing",
+    );
+    expect(missing.status).toBe(404);
+  });
+
   it("rejects a non-enum GET file encoding before reading bytes", async () => {
     await callRoute("POST", "/api/workbench/vfs/projects", {
       projectId: "encoding-vfs",
--- a/packages/agent/src/api/workbench-vfs-routes.ts
+++ b/packages/agent/src/api/workbench-vfs-routes.ts
@@ -512,7 +512,12 @@ function sendVfsError(ctx: WorkbenchRouteContext, err: unknown): void {
         ? 404
         : err.code === "QUOTA_EXCEEDED"
           ? 413
-          : 400;
+          : err.code === "INVALID_SNAPSHOT" ||
+              err.code === "VFS_STORAGE_FAILED" ||
+              err.code === "ROLLBACK_FAILED" ||
+              err.code === "ROLLBACK_RECOVERY_FAILED"
+            ? 500
+            : 400;
     ctx.error(ctx.res, err.message, status);
     return;
   }
--- a/packages/agent/src/services/vfs-builtin-shell.ts
+++ b/packages/agent/src/services/vfs-builtin-shell.ts
@@ -10,7 +10,6 @@
  * grep/rg evaluate caller patterns in a time-bounded worker because native
  * JavaScript regular-expression matching is synchronous.
  */
-import fsp from "node:fs/promises";
 import path from "node:path";
 import { runVfsSearchPattern } from "./vfs-search-pattern.ts";
 import { createVirtualFilesystemService } from "./virtual-filesystem.ts";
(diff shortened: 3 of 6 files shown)


ss251 (approved):
Reviewed head faed396653a0d68573e5a0d822ccafae4f2b69bc.

The risky part is `packages/agent/src/services/virtual-filesystem.ts:367` where `rollback()` copies the snapshot to `.rollback-stage-*`, journals `.pending-rollback.json`, then renames `files` to `.rollback-backup-*` before installing the stage, with `last-rollback.json` as the commit marker. I followed the two branches that usually break this: pre-commit interruption restores `backup` over any partial install via `recoverPendingRollback(false)`, while committed recovery only deletes `backup`, and `validateSnapshotTree` byte-count check rejects missing/changed/symlink trees before any live mutation, so `state.txt` stays `current bytes`.

No blocking concerns (non-blocking: shell `mkdir` symlink path now relies on `ensureSafeParentDirectory`).

Verified by reading the diff at this head and tracing stage/backup/commit branches; no live run.

Recommendation: merge

— [muse-lane-ss251]

attentionhead (approved):
Approving at `faed3966`. I reproduced the data loss on `develop` before reading the fix, because "removed live project data before a fallible copy" is a claim that should be demonstrable rather than argued — and it is.

**The bug is real, and I have a before/after.** On `develop`, `rollback` is:

```ts
const previous = await this.createSnapshot(`pre-rollback:${snapshotId}`);
await fsp.rm(this.filesRoot, { recursive: true, force: true });   // destroy
await fsp.mkdir(this.filesRoot, { recursive: true, mode: 0o700 });
await fsp.cp(snapshotRoot, this.filesRoot, { ... });              // fallible
```

I wrote a file, snapshotted, wrote newer content, then made the restore copy fail (stripping read permission from the snapshot tree) and called `rollback`. Same probe, both heads:

```
develop:  restore threw: true | live files after failed restore: []
this PR:  passes — important.txt still present
```

So on `develop` a failed restore leaves the project tree empty. The `pre-rollback` snapshot means the data is recoverable in principle, but the live tree is destroyed and recovery is a manual step the user has to know about — for an operation they expected to either succeed or leave things alone.

**The new ordering is the right shape.** Copy into a `.rollback-stage-<uuid>`, validate the staged tree, persist the commit marker, then two `rename` calls to swing the old tree aside and the staged tree in. Every fallible step now happens before anything destructive, and the destructive steps are atomic renames rather than a recursive `rm` plus a copy that can die halfway. Staging inside `projectRoot` rather than the OS temp dir is deliberate and correct — a cross-device rename would not be atomic.

**The second bug is subtler and I think the more dangerous of the two.** On `develop`:

```ts
const raw = await fsp.readFile(metadataPath, "utf-8").catch(() => null);
return raw ? (JSON.parse(raw) as VirtualFilesystemSnapshot) : null;
```

That `.catch(() => null)` swallows *every* read failure — EACCES, EIO, anything — and the lister then skips the entry. A user whose snapshot metadata became unreadable would be told they have no snapshots at all. Combined with the first bug, the worst case is ugly: a destructive restore whose safety net is invisible. Worth noting the old code was also inconsistent about it — read errors were swallowed while `JSON.parse` threw a raw `SyntaxError`.

**The fix avoids the obvious over-correction**, which is what I checked next: turning swallowed errors into hard failures often breaks a legitimate 404. It does not here. `readSnapshot` now separates a missing directory (`SNAPSHOT_NOT_FOUND`, preserving the documented 404 contract) from a non-directory or symlink (`INVALID_SNAPSHOT`), and then validates that the parsed record's `id` and `projectId` actually match what was requested — which also closes cross-project snapshot confusion, beyond what the description claims.

**Tests:** `virtual-filesystem-persistence.test.ts` and `workbench-vfs-routes.test.ts` are 32/32 at this head.

**What I did not verify:** the four interruption phases and crash recovery through the commit marker, the concurrent-service-instance cases, and the queue serialization now covering VFS reads, Git operations and builtin shell `mkdir`. I read them and their tests pass; I only independently exercised the failed-restore path and the metadata contract.

One thing worth stating for the release notes rather than the code: this changes `rollback` from "may destroy the tree on failure" to "leaves the tree untouched on failure". Anyone who built a recovery habit around the `pre-rollback:` snapshots will find they are still created, but no longer load-bearing.

Approving — this is a genuine data-loss fix and the reproduction is straightforward for anyone who wants to confirm it.

— [claude-code-eliza]

Issue #30577, opened 2026-09-05
fix(cloud): consolidate bounded fetch and preserve decoded response contracts


lalalune:
Cloud shared owned-bounded-fetch, bounded-provider-fetch and gateway-webhook bounded-fetch independently implement the same fetch deadline, bounded body buffering and cleanup loop. Their behavior has diverged: the older provider/gateway copies retain content-encoding and wire content-length after fetch has decoded the bytes, and may replace the selected timeout reason with a transport abort error.

Move the runtime-neutral transport loop into cloud services _common, its documented owner, with explicit subpath export. Keep the three public adapters and their error classes/codes, provider context, caller bounds and fetch injection compatibility. Normalize rebuilt decoded response headers and preserve selected abort provenance, deadlines through body reads, cleanup and late-response cancellation. Add the explicit cloud-shared workspace dependency.

Acceptance: real localhost HTTP/stream tests exercise each adapter for compressed response bytes/headers, normal/empty responses, bounds, cancellation and hanging headers/body. Existing adversarial stream and message-router/gateway tests stay green; affected package tests, typecheck, lint and build pass. No provider calls or deployment are required.

lalalune:
Completed and merged into develop in PR #30597 (a59939119e0565549f7dad8124e88a5101c7e1d9).

All three REST adapters use the canonical bounded helper. Real workerd cases passed gzip decoding, bodyless response, size rejection and body deadline; actual Bun probes confirmed all three bodyless statuses normalize to null bodies. Root verification passed all 376 tasks, and all five hosted checks passed on the reviewed PR head. The PR contains the detailed validation and evidence.

Issue #30575, opened 2026-09-05
fix(cloud-sdk): enforce the total job polling deadline


lalalune:
`ElizaCloudClient.pollJob()` checks its timeout only between requests. `getJob()` receives no request timeout, so a server that never sends headers or finishes its response body can keep polling pending forever. A polling interval longer than the remaining budget also exceeds the advertised total timeout.

Pass the remaining total budget through the SDK's existing HTTP deadline and cap each interval by that budget. Add optional request timeout/signal options to getJob without changing existing callers. Validate timer-safe polling options; preserve terminal completed/failed results and immediate expiration for zero timeout.

Acceptance: real localhost HTTP tests for hanging headers/body, pending-to-completed and failed outcomes, and intervals exceeding remaining budget; invalid options fail clearly. SDK tests, typecheck, lint and build pass.

The same defect affects waitForCliLogin/pollCliLogin, including cancellation during a request or interval. Include both SDK pollers under one lifecycle helper, preserving the login cancellation message and authenticated/expired/error outcomes.

lalalune:
Completed and merged into develop in PR #30597 (a59939119e0565549f7dad8124e88a5101c7e1d9).

Real HTTP deadline cases cover headers, response bodies, waits, and cancellation for job and CLI-login polling. The unchanged polling suite passed all 25 cases in its final serial run. Root verification passed all 376 tasks, and all five hosted checks passed on the reviewed PR head. The PR contains the detailed validation and evidence.

PR #30597, opened 2026-09-05
fix(cloud): enforce response contracts and total request deadlines

### LINKED ISSUE #30574: fix(cloud-sdk): reject fabricated success for non-JSON responses
Parsed SDK requests currently replace any successful text or empty body with `{success:true}`. A public `getUser()` call receiving an HTML maintenance response therefore resolves with invented account data. JSON strings are also replaced. Two tests encode the fabricated success behavior.

Accept valid JSON media types (including structured +json) and preserve every JSON value. Reject successful non-JSON or unexpectedly empty data responses with a typed CloudApiError. Explicit bodyless HTTP responses (HEAD, 204, 205) must return undefined; raw text remains available through requestRaw. Preserve non-2xx mapping and request deadlines.

Acceptance: real localhost HTTP tests through public SDK methods exercise HTML, missing/wrong media types, empty/malformed JSON, JSON primitives, structured media types, bodyless protocols and raw text; owning package tests, typecheck, lint and build pass. Document the deliberate correction for callers previously relying on invented success.

lalalune:
The Cloud SDK could cast a successful non-JSON response to the requested object type, and polling deadlines stopped at request dispatch instead of covering headers, body reads and the next interval. Parsed SDK calls now reject invalid payloads, preserve valid JSON primitives, and enforce a total monotonic deadline through every polling operation. Raw-response calls retain raw access; HEAD, 204 and 205 remain bodyless.

Three cloud REST adapters now use one bounded Fetch lifecycle for deadlines, body reads, cancellation and decoded response framing. Each adapter retains its error identity, context and existing limits. This removes duplicated lifecycle code without introducing a blanket chunk limit.

Closes #30574, #30575 and #30577.

Compatibility: successful parsed calls returning non-JSON or malformed JSON now reject instead of masquerading as the caller's requested data. JSON vendor media types are accepted. The SDK README documents this correction. The dependency and package export for the shared transport are explicit in the manifest and lockfile.

Validation:

- Scoped suites: SDK 165 passed, common 59, gateway 496, shared transport 27; relevant package typecheck, lint and builds passed. The consuming plugin passed 631 Vitest cases, 21 Bun contracts and five built-package imports.
- Full cloud-shared lane: 342 batches across 1,168 files, with 12,391 tests passed, 201 skipped and none failed. Earlier attempts encountered host contention or missing build output; the final complete run passed with stable prerequisites.
- Real boundaries: 68 Node HTTP cases and four installed-workerd cases passed. An independent reviewer also ran 40 SDK HTTP and 28 adapter HTTP cases with no blocker.
- Rebase verification: all changed cloud files are identical after rebase. On the final head, 67 of 68 HTTP cases passed concurrently; an 80 ms deadline expired before the server observed dispatch in one case. Its complete 25-test polling file passed on serial rerun without changing deadlines or assertions. All four Worker cases passed again.
- Root gate: the first scoped run passed all 376 tasks. The final rebased run also passed all 376 tasks and every repository audit, including the scan of 11,367 tests (zero focused tests or untracked skips).

Complete [validation record](https://github.com/elizaOS/eliza/releases/download/pr-evidence-11/30597-cloud-validation-5ac21104.json).

Review follow-up: the two cloud-shared wrapper files run in the normal Bun lane; their five cases passed in the complete package run. The supplemental Vitest config is for suites requiring Vitest-only APIs. The null-body status branch is reachable in Bun 1.3.14: its native Fetch exposes a stream for 204/205/304, whereas Node 24 and workerd return null. The canonical helper normalizes those Bun responses to null bodies; the existing real HTTP wrapper tests cover 204. An additional real-server check produced:

```json
{"runtime":"Bun 1.3.14","status":204,"rawBodyIsNull":false,"normalizedBodyIsNull":true,"passed":true}
{"runtime":"Bun 1.3.14","status":205,"rawBodyIsNull":false,"normalizedBodyIsNull":true,"passed":true}
{"runtime":"Bun 1.3.14","status":304,"rawBodyIsNull":false,"normalizedBodyIsNull":true,"passed":true}
```

## Evidence Gate

N/A - SDK and server HTTP contracts; no rendered UI changed.

N/A - SDK and server HTTP contracts; no rendered UI changed.

N/A - the real HTTP/readback contract is captured below.

[Final-head polling rerun](https://github.com/elizaOS/eliza/releases/download/pr-evidence-11/30597-cloud-poll-5ac21104.log).

```text
Command: ['node', 'node_modules/vitest/vitest.mjs', 'run', 'packages/cloud/services/gateway-webhook/src/adapters/bounded-fetch-http.test.ts', 'packages/cloud/sdk/src/http-server.test.ts', 'packages/cloud/sdk/src/poll-server.test.ts']

 RUN  v4.1.10 /Users/shawwalters/.codex/worktrees/code-quality-cleanup-20260904


 Test Files  3 passed (3)
      Tests  68 passed (68)
   Start at  21:59:28
   Duration  1.33s (transform 842ms, setup 0ms, import 1.00s, tests 1.53s, environment 0ms)


Command exit: 0
```

N/A - no browser page flow changed. Node HTTP tests use real servers for late headers, stalled/chunked bodies, caller cancellation and JSON contracts.

N/A - no model inference, prompt, provider or action-result construction changed.

[Final-head Worker readback](https://github.com/elizaOS/eliza/releases/download/pr-evidence-11/30597-cloud-worker-5ac21104.json).

Actual installed workerd, using the cloud API compatibility date, handled gzip decoding, a bodyless response, an oversize rejection and a stalled body deadline. The API Worker production dry-run also bundled successfully. No production deployment was performed.

```json
{
  "runtime": "workerd (installed Miniflare)",
  "decoded": {
    "status": 200,
    "body": "{\"complete\":\"worker-response worker-response worker-response worker-response worker-response worker-response worker-response worker-response worker-response worker-response worker-response worker-response worker-response worker-response worker-response worker-response worker-response worker-response worker-response worker-response worker-response worker-response worker-response worker-response worker-response worker-response worker-response worker-response worker-response worker-response worker-response worker-response worker-response worker-response worker-response worker-response worker-response worker-response worker-response worker-response worker-response worker-response worker-response worker-response worker-response worker-response worker-response worker-response worker-response worker-response worker-response worker-response worker-response worker-response worker-response worker-response worker-response worker-response worker-response worker-response worker-response worker-response worker-response worker-response worker-response worker-response worker-response worker-response worker-response worker-response worker-response worker-response worker-response worker-response worker-response worker-response worker-response worker-response worker-response worker-response worker-response worker-response worker-response worker-response worker-response worker-response worker-response worker-response worker-response worker-response worker-response worker-response worker-response worker-response worker-response worker-response worker-response worker-response worker-response worker-response \"}",
    "encoding": null,
    "length": "1615"
  },
  "empty": {
    "status": 204,
    "body": "",
    "encoding": null,
    "length": null
  },
  "oversize": {
    "error": "too-large",
    "name": "Error"
  },
  "stalled": {
    "error": "Worker deadline expired",
    "name": "TimeoutError"
  }
}
```

Known gaps: real-time deadline tests are sensitive to heavy host contention, as recorded above. Opt-in SDK integration tests requiring external credentials remain skipped. All changed HTTP contracts were exercised locally with real servers and the installed Worker runtime. Hosted checks must pass on the current head before merge. Maintainer CI exception: N/A.

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -728,6 +728,7 @@
         "@electric-sql/pglite-socket": "0.1.6",
         "@elevenlabs/elevenlabs-js": "^2.45.0",
         "@elizaos/cloud-sdk": "workspace:*",
+        "@elizaos/cloud-services-common": "workspace:*",
         "@elizaos/core": "workspace:*",
         "@elizaos/plugin-cloud-apps": "workspace:*",
         "@elizaos/plugin-doordash": "workspace:*",
--- a/packages/cloud/sdk/README.md
+++ b/packages/cloud/sdk/README.md
@@ -145,3 +145,9 @@ throw `CloudApiError`; they never become an invented `{ success: true }` DTO.
 HEAD and HTTP 204/205 responses resolve to `undefined`, as these protocols have
 no response body. Use `requestRaw` (or a generated `Raw` method) for text,
 binary, streaming, or status-sensitive responses.
+
+`pollJob` and `waitForCliLogin` enforce a total timeout through every request,
+response-body read and polling interval. Their timeout and interval options
+accept integers from 0 through 2,147,483,647 milliseconds; zero timeout expires
+immediately. Login cancellation also interrupts an in-flight request or wait.
+Direct `getJob` and `pollCliLogin` calls accept optional `timeoutMs` and `signal`.
--- a/packages/cloud/sdk/README.md
+++ b/packages/cloud/sdk/README.md
@@ -135,3 +135,13 @@ Build and publish:
 bun run build
 npm publish --access public
 ```
+
+## Parsed responses
+
+`request` and typed endpoint methods accept `application/json` and application
+media types ending in `+json`, preserving objects, arrays and JSON primitives.
+Successful HTML/text, malformed JSON and unexpectedly empty JSON responses
+throw `CloudApiError`; they never become an invented `{ success: true }` DTO.
+HEAD and HTTP 204/205 responses resolve to `undefined`, as these protocols have
+no response body. Use `requestRaw` (or a generated `Raw` method) for text,
+binary, streaming, or status-sensitive responses.
--- a/packages/cloud/sdk/src/client.ts
+++ b/packages/cloud/sdk/src/client.ts
@@ -11,6 +11,7 @@
 
 import { isCliLoginSessionId } from "./cli-login.js";
 import { CloudApiClient, CloudApiError, ElizaCloudHttpClient } from "./http.js";
+import { pollUntil } from "./poll.js";
 import { ElizaCloudPublicRoutesClient } from "./public-routes.js";
 import type {
   CreateRedemptionRequest,
@@ -411,11 +412,14 @@ export class ElizaCloudClient {
     });
   }
 
-  pollCliLogin(sessionId: string): Promise<CliLoginPollResponse> {
+  pollCliLogin(
+    sessionId: string,
+    options?: Pick<CloudRequestOptions, "signal" | "timeoutMs">,
+  ): Promise<CliLoginPollResponse> {
     return this.request<CliLoginPollResponse>(
       "GET",
       `/api/auth/cli-session/${encodePathParam(sessionId)}`,
-      { skipAuth: true },
+      { ...options, skipAuth: true },
     );
   }
 
@@ -441,25 +445,22 @@ export class ElizaCloudClient {
       signal?: AbortSignal;
     } = {},
   ): Promise<CliLoginPollResponse> {
-    const timeoutMs = options.timeoutMs ?? 300_000;
-    const intervalMs = options.intervalMs ?? 2_000;
-    const deadline = Date.now() + timeoutMs;
-
-    while (Date.now() < deadline) {
-      if (options.signal?.aborted) {
-        throw new Error("Eliza Cloud sign-in was cancelled");
-      }
-      const result = await this.pollCliLogin(sessionId);
-      if (result.status === "authenticated") {
-        return result;
-      }
-      if (result.status === "expired" || result.status === "error") {
-        throw new Error(result.error ?? `Eliza Cloud sign-in ${result.status}`);
-      }
-      await new Promise((resolve) => setTimeout(resolve, intervalMs));
-    }
-
-    throw new Error("Timed out waiting for Eliza Cloud sign-in");
+    return pollUntil({
+      read: (requestOptions) => this.pollCliLogin(sessionId, requestOptions),
+      isComplete: (result) => {
+        if (result.status === "expired" || result.status === "error") {
+          throw new Error(
+            result.error ?? `Eliza Cloud sign-in ${result.status}`,
+          );
+        }
+        return result.status === "authenticated";
+      },
+      timeoutMs: options.timeoutMs ?? 300_000,
+      intervalMs: options.intervalMs ?? 2_000,
+      signal: options.signal,
+      timeoutMessage: "Timed out waiting for Eliza Cloud sign-in",
+      cancellationMessage: "Eliza Cloud sign-in was cancelled",
+    });
   }
 
   pairWithToken(token: string, origin: string): Promise<AuthPairResponse> {
@@ -1518,27 +1519,29 @@ export class ElizaCloudClient {
     );
   }
 
-  getJob(jobId: string): Promise<JobStatus> {
-    return this.request("GET", `/api/v1/jobs/${encodePathParam(jobId)}`);
+  getJob(
+    jobId: string,
+    options?: Pick<CloudRequestOptions, "signal" | "timeoutMs">,
+  ): Promise<JobStatus> {
+    return this.request(
+      "GET",
+      `/api/v1/jobs/${encodePathParam(jobId)}`,
+      options,
+    );
   }
 
   async pollJob(
     jobId: string,
     options: { timeoutMs?: number; intervalMs?: number } = {},
-  ) {
-    const timeoutMs = options.timeoutMs ?? 120_000;
-    const intervalMs = options.intervalMs ?? 2_000;
-    const deadline = Date.now() + timeoutMs;
-
-    while (Date.now() < deadline) {
-      const job = await this.getJob(jobId);
-      if (job.status === "completed" || job.status === "failed") {
-        return job;
-      }
-      await new Promise((resolve) => setTimeout(resolve, intervalMs));
-    }
-
-    throw new Error(`Timed out waiting for Eliza Cloud job ${jobId}`);
+  ): Promise<JobStatus> {
+    return pollUntil({
+      read: (requestOptions) => this.getJob(jobId, requestOptions),
+      isComplete: (job) =>
+        job.status === "completed" || job.status === "failed",
+      timeoutMs: options.timeoutMs ?? 120_000,
+      intervalMs: options.intervalMs ?? 2_000,
+      timeoutMessage: `Timed out waiting for Eliza Cloud job ${jobId}`,
+    });
   }
 
   getUser(): Promise<UserProfileResponse> {
--- a/packages/cloud/sdk/src/http-server.test.ts
+++ b/packages/cloud/sdk/src/http-server.test.ts
@@ -1,0 +1,96 @@
+/**
+ * Exercises parsed SDK response contracts over real localhost HTTP, including
+ * public account access and raw/bodyless protocol behavior without fetch mocks.
+ */
+import { createServer } from "node:http";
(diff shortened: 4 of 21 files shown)


ss251 (approved):
Reviewed head 5ac211041636cce843332474c652b6ba198df68d.

The risky part is `packages/cloud/sdk/src/http.ts:694` where parsed `request` now rejects non-JSON 2xx instead of inventing `{ success: true }`. I followed the three guards in order: HEAD/204/205 returns `undefined` before any content-type check, then `hasJsonContentType` throws `unexpected_response_content_type` for missing or `text/plain` types, then empty body throws `empty_response_body` and malformed JSON throws `malformed_json_response`. The suffix test `/^application\/[\w.+-]+\+json$/` runs on the lowercased token split at `;`, so `Application/Problem+JSON; charset=utf-8` from `http-server.test.ts:26` still parses while `{"id":"real"}` labeled `text/plain` correctly rejects. JSON primitives, arrays and objects fall through as `body as TResponse`.
No blocking concerns; text/binary stays available via `requestRaw`.
Verified by reading the diff at this head and tracing the real-HTTP contracts in `http-server.test.ts`.

Recommendation: merge

— [muse-lane-ss251]

ss251 (approved):
Reviewed head 5ac211041636cce843332474c652b6ba198df68d.

The risky part is `packages/cloud/sdk/src/http.ts:694` where the `request` success path replaces the invented `{ success: true }` fallback with a strict contract. I traced the three gates in order: HEAD/204/205 return `undefined` before any content check, `hasJsonContentType` at :702 rejects missing/`text/plain`/HTML with `unexpected_response_content_type`, and empty vs malformed JSON throw distinct `empty_response_body`/`malformed_json_response` codes. The suffix test `/^application\/[\w.+-]+\+json$/` correctly still accepts the `Application/Problem+JSON; charset=utf-8` fixture in `http-server.test.ts` while rejecting `text/plain`, and JSON primitives, arrays and objects pass through unmodified. `requestRaw` remains the text/binary escape hatch.

No blocking concerns; non-blocking: 304 bodyless handling lives only in `bounded-fetch.ts`, not this gate.

Recommendation: merge

— [muse-lane-ss251]

attentionhead (approved):
Approving at `5ac211041636cce843332474c652b6ba198df68d`.

Two separable pieces here — a three-way transport consolidation and a deliberate SDK response-contract break. I checked them separately.

## The consolidation

`gateway-webhook/adapters/bounded-fetch.ts`, `cloud/shared/utils/bounded-provider-fetch.ts` and `cloud/shared/utils/owned-bounded-fetch.ts` collapse into `_common/src/bounded-fetch.ts`, each keeping a thin wrapper. The thing that matters in a consolidation like this is whether the wrappers still fail the way their callers expect, and they do: `INVALID_GATEWAY_TIMEOUT` / `GATEWAY_RESPONSE_TOO_LARGE`, `INVALID_PROVIDER_REQUEST_BOUNDS` / `PROVIDER_RESPONSE_TOO_LARGE`, and `INVALID_CLOUD_REST_BOUNDS` / `CLOUD_REST_RESPONSE_TOO_LARGE` with `severity: "ephemeral"` all match the deleted implementations verbatim, and `maxResponseChunks` stays optional so gateway and provider keep their previous uncapped-chunk behavior.

The shared version is not a pure union, though — it applies `owned`'s header rebuild to all three callers, and that is an observable fix. Against a local origin returning gzipped JSON:

```
develop:   status=200 ce=gzip cl=37 body="{\"hello\":\"world\"}"
PR #30597: status=200 ce=null  cl=17 body="{\"hello\":\"world\"}"
```

`fetch` hands back decoded bytes, so develop's gateway/provider responses advertise `content-encoding: gzip` and a `content-length` of 37 while carrying 17 decoded bytes. The new code strips the encoding headers and restates the length. I traced the blast radius before weighting this: the three gateway adapters (`twilio`, `blooio`, `whatsapp`) parse internally rather than forwarding the `Response` downstream, so this is a latent correctness issue rather than a live one — but it was wrong, and it is now right.

One hypothesis of mine did **not** hold, and I am recording it so nobody re-derives it: I expected the new `bodyless` guard for 204/205/304 to be fixing a real `TypeError`, since `new Response(new Uint8Array(0).buffer, {status: 204})` does throw. It isn't. Both old implementations return early at `if (!response.body)`, and undici gives a 204 a null body, so that path was already safe. The guard is defensive and, as far as I can tell, unreachable.

## The SDK contract change — this one is breaking

`ElizaCloudHttpClient` used to answer a 2xx that carried a non-JSON content-type or an empty body with a fabricated `{ success: true }`. It now throws `unexpected_response_content_type` or `empty_response_body`, with `HEAD`/204/205 resolving to `undefined`. That is a genuine break for downstream consumers and worth calling out as such in the release notes if it isn't already.

It is handled well: the README documents it, `requestRaw` is the documented escape hatch and is exposed on the client and on generated `Raw` methods, and the contract test that asserted the old behavior is updated in place and renamed (`returns {success: true} for non-JSON 200 response` → `rejects non-JSON 200 responses instead of inventing a DTO`) rather than quietly deleted.

I swept in-repo routes for 2xx-with-empty-body handlers that a typed call would now reject. The 204 sites are all exempted. The one empty-**200** I found is `adminStatusResponse` in `api/v1/admin/moderation/route.ts:235,245`, which carries its whole answer in `X-Is-Admin` / `X-Admin-Role` headers — it isn't in `public-routes.ts`, so it isn't reachable through the typed client, and a typed JSON call against it could never have worked anyway. Under the old behavior it would have returned `{success: true}` and silently discarded the headers, which is roughly the failure mode this PR exists to remove.

## Total deadlines

The polling rewrite is the substantive half. `waitForCliLogin` and `pollJob` previously ran `while (Date.now() < deadline)` loops that never passed a timeout or signal into the individual request, so a single hung request could outlive the whole budget, and cancellation could not interrupt an in-flight `setTimeout` wait. `pollUntil` threads the remaining budget into each `read` and makes the interval wait abortable.

I confirmed that is load-bearing rather than cosmetic by deleting just the threading:

```
result = await options.read({ signal: options.signal });   // was: { timeoutMs: remaining, signal }
```

```
× bounds headers stalls and long intervals by the total timeout 120003ms
× bounds body stalls and long intervals by the total timeout 120002ms   (×2 more)
Tests  4 failed | 36 passed (40)
```

The four failures are hangs that run out the 120 s vitest timeout — exactly the escape the change closes — while the other 36 still pass, so the suite discriminates.

## Worth a follow-up, not blocking

`packages/cloud/shared/src/lib/utils/owned-bounded-fetch.test.ts` and `bounded-provider-fetch.test.ts` import from `vitest` but are absent from the `include` allowlist in `packages/cloud/shared/vitest.config.ts`, so that lane does not run them. That predates this PR and the config is untouched here, but it bites harder now: those are the tests covering the two modules this PR rewrote, and the config's own comment warns about precisely this failure mode. I ran them directly against the new delegating implementation and both pass (5 tests), so there is no drift hiding behind the gap — but five tests is thin cover for two rewritten transports, and adding them to `include` would be cheap.

## Verification

- `_common/tests/bounded-fetch.test.ts` — 4 pass (bun)
- `gateway-webhook` — 496 pass / 0 fail across 37 files (bun); 35 pass across the 3 vitest files including the new `bounded-fetch-http.test.ts`
- `cloud/sdk` — 148 pass (vitest) + 17 pass (bun); new `http-server.test.ts` and `poll-server.test.ts` included
- `plugin-elizacloud/__tests__/cloud-api-client-contract.test.ts` — 21 pass
- typecheck after `bun run build:core`: `cloud/sdk`, `_common`, `gateway-webhook` all 0 errors

A note on that last line, since it cost me time on the previous review: this repo's static lane runs `build:core` before `run-turbo.mjs run typecheck --affected`, and typechecking these packages from a stale worktree without that step produces resolution errors that look like real breaks. Anyone re-verifying should build first.

## Not verified

I did not exhaustively cross-reference every generated route in `public-routes.ts` against its handler for empty-2xx responses; my sweep was over route source rather than the generated surface, so a typed route returning an empty 200 could still exist. I did not run the full repository suite, and `Source static smoke` had not reported when I reviewed.

— [claude-code-eliza]

lalalune:
Independent review completed against merged develop `e6cd97a9e31401fc3d23e4d04786d22589aec529`. Read all 21 diffs and the SDK/transport callers; no further defect found. Independently passed SDK 165 tests, common transport 59, real gateway HTTP 28, provider transport four, and owned transport one (257 passed; the SDK retains 19 explicit live opt-in skips).

The exercised boundaries include malformed/empty/non-JSON responses, structured JSON media types and primitive values, HEAD/204/205, response-body deadlines, cancellation, gzip, late transport cleanup, and the total polling deadline. No HTML or text response is reported as fabricated parsed success.

Issue #18081, opened 2026-08-08
Post-merge #18079: Unicode sanitizer misses request fields and JSON keys


standujar:
## Post-merge regression

#18079 was merged to develop as 6160885f561f2f38023d37d73315ef5929c5a4e3. Re-auditing that exact merged tree confirms the remaining blockers from its exact head b607317fcb7e726e7df29e429ba0b5f7bc83ff14. The five relevant files are byte-identical between those commits.

## Deterministic evidence at 6160885

1. JSON object keys are not sanitized.

packages/core/src/utils/well-formed.ts:142-147 recursively sanitizes each entry value, then reuses the original key. The exact merged helper produces:

    deepToWellFormedUnicode({ ["bad\uD83D"]: "ok" })
    JSON.stringify(...) === '{"bad\\ud83d":"ok"}'

2. An own __proto__ key is corrupted when any sibling changes.

The clone is a normal object and writes next[key] = sanitized. With:

    JSON.parse('{"__proto__":{"marker":"kept"},"bad":"\\ud83d"}')

the input has an own __proto__ property, but the sanitized result does not; serialization becomes {"bad":"�"}. The assignment mutates the clone prototype instead of preserving the JSON member.

3. The provider boundary still does not sanitize the whole request.

- plugins/plugin-openai/models/text.ts:2029-2041 sanitizes prompt/messages and system, then appends raw tools, toolChoice, output, and providerOptions.
- plugins/plugin-anthropic/models/text.ts:1329-1350 does the same and also appends raw stopSequences.

A lone surrogate in an Anthropic stop sequence, or in an OpenAI/Anthropic tool description/schema, therefore still reaches captured request JSON as a \ud8xx escape. This contradicts the merged comment that every outgoing string is forced well-formed.

4. Surrogate-safe truncation reports the wrong dropped count.

At packages/core/src/runtime/planner-rendering.ts:79-90 and packages/core/src/runtime/planner-loop.ts:2279-2281, the marker count is computed from requested head/tail lengths before the helpers back off at surrogate boundaries. Exact repro:

    input = "💀".repeat(100) // 200 UTF-16 code units
    truncateToolResultText(input, 30)
    => "💀💀 [193 chars truncated] 💀"

The output preserves 6 code units, so 6 + 193 = 199, not 200.

## Release impact

#18028 currently points at the same merged commit 6160885 and therefore contains these defects. It should not be promoted until this issue is fixed and the release gates are green.

## Required fix/coverage

[...]

PR #30599, opened 2026-09-05
fix(core): keep own __proto__ entries in sanitized trajectory JSON

andreolf:
# Relates to

No linked issue. Found during a read-only audit of `develop` and reproduced locally before any change.

      evidence attached below.

# Sync with develop

# Risks

Low. Two accumulators inside one module-private walker change from a `{}` literal
to a null-prototype staging object with `Object.defineProperty`, restored to
`Object.prototype` before returning. Behaviour is identical for every input that
has no own `__proto__` key — the added test `still normalizes ordinary records
unchanged` pins that. No exported signature, schema, or serialized shape changes;
`JSON.stringify` output is unchanged for ordinary records.

# Background

## What does this PR do?

`sanitizeTrajectoryJsonValueInternal` rebuilt both the plain-object and `Map`
branches with `output[key] = sanitized` on a `{}` accumulator. `JSON.parse`
creates `__proto__` as an **own data property**, so `Object.entries` surfaces it,
the loop sanitizes it — and then the assignment invokes the `__proto__` *setter*
instead of defining an own entry.

Two things go wrong at once:

```
input own keys : [ 'path', '__proto__' ]
input JSON     : {"path":"/etc/passwd","__proto__":{"approved":true}}
output own keys: [ 'path' ]
output JSON    : {"path":"/etc/passwd"}
prototype replaced?: true
inherited leak -> out.approved: true
```

The key is **silently dropped** from the recorded value, and the sanitized
object's prototype is **replaced by the caller's own data**, so `out.approved`
reads `true` on a value the sanitizer had just declared safe. (Global
`Object.prototype` is not affected — the pollution is per-returned-object.)

This runs over model-generated planner tool arguments.
`boundStepsForPersistence` in `features/trajectories/TrajectoriesService.ts`
sanitizes `step.action`, `step.observation`, every `step.llmCalls` entry and
`step.metadata` before `stringifyTrajectoryJsonForSql` writes the row, and
`runtime/trajectory-recorder.ts` records tool stage I/O on the planner path. So a
persisted trajectory could misrepresent the arguments the model actually emitted
— including in the training-data export surface.

The fix stages both accumulators on `Object.create(null)`, defines own
properties, and restores `Object.prototype` before returning. That is the
pattern already used by `defineOwn` in `database/connector-json.ts` and the
staging object in `utils/well-formed.ts`, which solved this same class in #18081.

# Testing

`defineOwnEntry` and the two accumulators in
`packages/core/src/services/trajectory-json.ts`, then the
`trajectory JSON own-key preservation` block appended to
`packages/core/src/services/trajectory-json.test.ts`.

## Detailed testing steps

```bash
bun run --cwd packages/core test -- src/services/trajectory-json.test.ts
```

Four new cases through the exported `sanitizeTrajectoryJsonValue` /
`sanitizeTrajectoryJsonObject`: an own `__proto__` key is preserved rather than
dropped; the returned object is not re-parented and exposes no inherited member;
the same holds nested inside a recorded object; and the same holds for an entry
recorded through a `Map`. A fifth pins that ordinary records are unchanged.

Verified red before green — against unpatched source **4 fail**:

```
 × keeps an own __proto__ key instead of dropping it
 × does not re-parent the sanitized object to caller data
 × preserves an own __proto__ key nested inside a recorded object
 × keeps an own __proto__ entry recorded through a Map
 Tests  4 failed | 2 passed (6)
```

With the fix, `Tests 6 passed (6)`.

# Evidence Gate

- [x] Before full-page screenshots: `N/A - no UI surface; this is a module-private JSON sanitizer with no rendered component.`

- [x] After full-page screenshots: `N/A - no UI surface; see above.`

- [x] Video walkthrough: `N/A - no user-facing flow; the observable behaviour is the sanitizer's return value, asserted directly by the unit tests above.`

- [x] Backend logs: the before/after sanitizer output and the red/green test runs are pasted inline above and below.

- [x] Frontend console/network logs: `N/A - no frontend code path is touched.`

- [x] Real-LLM trajectory: `N/A - the changed code performs no model call. It normalizes an already-captured value; a live run would exercise nothing this diff changes, and the recorded shape is asserted deterministically by the tests.`

- [x] Domain artifacts: the sanitized record is the artifact, asserted inline (`{"path":"/etc/passwd","__proto__":{"approved":true}}` round-trips intact after the fix).

# Evidence Details

## Backend logs

Against the real module before the fix:

<details>
<summary>own __proto__ dropped, prototype replaced</summary>

```
input own keys : [ "path", "__proto__" ]
input JSON     : {"path":"/etc/passwd","__proto__":{"approved":true}}
output own keys: [ "path" ]
output JSON    : {"path":"/etc/passwd"}
key dropped?   : true
prototype replaced?: true
inherited leak -> out.approved: true
global Object.prototype polluted?: false
```

</details>

After the fix, same input:

<details>
<summary>key preserved, prototype intact</summary>

```
input own keys : [ "path", "__proto__" ]
output own keys: [ "path", "__proto__" ]
output JSON    : {"path":"/etc/passwd","__proto__":{"approved":true}}
key dropped?   : false
prototype replaced?: false
inherited leak -> out.approved: undefined
global Object.prototype polluted?: false
```

</details>

Sibling suites that consume this module, post-rebase:

```
bun run --cwd packages/core test -- src/services/trajectory-export.test.ts \
  src/runtime/__tests__/trajectory-recorder.test.ts
 Test Files  2 passed (2)
      Tests  58 passed (58)
```

Biome on both changed files: `Checked 2 files. No fixes applied.`

`N/A - no UI surface is touched by this diff.`

## Audio / voice walkthrough

`N/A - no voice, transcript, TTS, or STT path is touched.`

## Known gaps / failures

- **`bun run verify` was not run in full.** The owning package's focused lane
  (`trajectory-json`, plus the `trajectory-export` and `trajectory-recorder`
  suites that consume this module — 64 tests total) and the Biome gate were run
  after the final rebase. The full repository verify is a long multi-package lane
  and was not completed in this environment. The diff touches one core module and
  its colocated test; no export, schema, or serialized shape changed. Happy to
  attach a full run if a maintainer wants one before merge.
- Every `N/A` evidence row above is a non-UI, non-model code path, with the
  reason recorded inline on the row rather than left blank.

### DIFF
--- a/packages/core/src/services/trajectory-json.test.ts
+++ b/packages/core/src/services/trajectory-json.test.ts
@@ -1,7 +1,10 @@
 /** Exercises lossless trajectory JSON normalization for large prompt evidence. */
 
 import { describe, expect, it } from "vitest";
-import { sanitizeTrajectoryJsonObject } from "./trajectory-json";
+import {
+	sanitizeTrajectoryJsonObject,
+	sanitizeTrajectoryJsonValue,
+} from "./trajectory-json";
 
 describe("trajectory JSON normalization", () => {
 	it("preserves strings and collections beyond the former capture limits", () => {
@@ -13,5 +16,62 @@ describe("trajectory JSON normalization", () => {
 		const sanitized = sanitizeTrajectoryJsonObject({ prompt, messages });
 		expect(sanitized?.prompt).toBe(prompt);
 		expect(sanitized?.messages).toEqual(messages);
+	});
+});
+
+describe("trajectory JSON own-key preservation", () => {
+	// `JSON.parse` creates `__proto__` as an own data property. Rebuilding the
+	// record with `output[key] = …` invokes the prototype setter instead, which
+	// both drops the argument from the persisted trajectory and re-parents the
+	// sanitized object to caller-controlled data.
+	const rawToolCall = String.raw`{"path":"/etc/passwd","__proto__":{"approved":true}}`;
+
+	it("keeps an own __proto__ key instead of dropping it", () => {
+		const sanitized = sanitizeTrajectoryJsonValue(JSON.parse(rawToolCall)) as
+			| Record<string, unknown>
+			| undefined;
+
+		expect(Object.keys(sanitized ?? {})).toEqual(["path", "__proto__"]);
+		expect(JSON.stringify(sanitized)).toBe(rawToolCall);
+	});
+
+	it("does not re-parent the sanitized object to caller data", () => {
+		const sanitized = sanitizeTrajectoryJsonValue(JSON.parse(rawToolCall)) as
+			| Record<string, unknown>
+			| undefined;
+
+		expect(Object.getPrototypeOf(sanitized)).toBe(Object.prototype);
+		expect((sanitized as { approved?: unknown }).approved).toBeUndefined();
+		expect(({} as { approved?: unknown }).approved).toBeUndefined();
+	});
+
+	it("preserves an own __proto__ key nested inside a recorded object", () => {
+		const sanitized = sanitizeTrajectoryJsonObject(
+			JSON.parse(String.raw`{"action":{"__proto__":{"approved":true}}}`),
+		);
+
+		expect(JSON.stringify(sanitized)).toBe(
+			String.raw`{"action":{"__proto__":{"approved":true}}}`,
+		);
+	});
+
+	it("still normalizes ordinary records unchanged", () => {
+		const sanitized = sanitizeTrajectoryJsonObject({ a: 1, b: ["x"] });
+
+		expect(sanitized).toEqual({ a: 1, b: ["x"] });
+		expect(Object.getPrototypeOf(sanitized)).toBe(Object.prototype);
+	});
+
+	it("keeps an own __proto__ entry recorded through a Map", () => {
+		const sanitized = sanitizeTrajectoryJsonValue(
+			new Map<string, unknown>([
+				["path", "/etc/passwd"],
+				["__proto__", { approved: true }],
+			]),
+		) as Record<string, unknown> | undefined;
+
+		expect(Object.keys(sanitized ?? {})).toEqual(["path", "__proto__"]);
+		expect(Object.getPrototypeOf(sanitized)).toBe(Object.prototype);
+		expect((sanitized as { approved?: unknown }).approved).toBeUndefined();
 	});
 });
--- a/packages/core/src/services/trajectory-json.ts
+++ b/packages/core/src/services/trajectory-json.ts
@@ -50,6 +50,25 @@ function reserveContainer(state: SanitizationState): JsonValue | undefined {
 	return undefined;
 }
 
+/**
+ * Records one own entry on a null-prototype accumulator. A plain
+ * `output[key] = value` would invoke the `__proto__` setter for a key
+ * `JSON.parse` produced as an own data property, dropping it from the recorded
+ * trajectory and re-parenting the sanitized object to caller-controlled data.
+ */
+function defineOwnEntry(
+	target: Record<string, JsonValue>,
+	key: string,
+	value: JsonValue,
+): void {
+	Object.defineProperty(target, key, {
+		configurable: true,
+		enumerable: true,
+		value,
+		writable: true,
+	});
+}
+
 function reserveObjectKey(state: SanitizationState, key: string): boolean {
 	return reserveBytes(
 		state,
@@ -118,7 +137,7 @@ function sanitizeTrajectoryJsonValueInternal(
 		const containerMarker = reserveContainer(state);
 		if (containerMarker !== undefined) return containerMarker;
 		state.seen.add(value);
-		const output: Record<string, JsonValue> = {};
+		const output = Object.create(null) as Record<string, JsonValue>;
 		for (const [keyValue, entry] of value.entries()) {
 			const key = String(keyValue);
 			reserveObjectKey(state, key);
@@ -127,9 +146,10 @@ function sanitizeTrajectoryJsonValueInternal(
 				state,
 				depth + 1,
 			);
-			if (sanitized !== undefined) output[key] = sanitized;
+			if (sanitized !== undefined) defineOwnEntry(output, key, sanitized);
 		}
 		state.seen.delete(value);
+		Object.setPrototypeOf(output, Object.prototype);
 		return output;
 	}
 	if (value instanceof Set) {
@@ -180,7 +200,7 @@ function sanitizeTrajectoryJsonValueInternal(
 				? {}
 				: normalizedScalar(String(value), state);
 		}
-		const output: Record<string, JsonValue> = {};
+		const output = Object.create(null) as Record<string, JsonValue>;
 		for (const [key, entry] of entries) {
 			reserveObjectKey(state, key);
 			const sanitized = sanitizeTrajectoryJsonValueInternal(
@@ -188,9 +208,10 @@ function sanitizeTrajectoryJsonValueInternal(
 				state,
 				depth + 1,
 			);
-			if (sanitized !== undefined) output[key] = sanitized;
+			if (sanitized !== undefined) defineOwnEntry(output, key, sanitized);
 		}
 		state.seen.delete(value);
+		Object.setPrototypeOf(output, Object.prototype);
 		return output;
 	}
 	return normalizedScalar(String(value), state);


ss251 (approved):
07bc8da890b4fdd694cc858a152da7e32ce49fd9

Fix is real, narrowly scoped, and addresses a material data-loss / prototype-safety bug in trajectory sanitization.

Input -> wrong output before: `JSON.parse('{"__proto__":{"approved":true}}')` produces `__proto__` as an own data property, but `packages/core/src/services/trajectory-json.ts:149` did `output[key] = sanitized` on a plain `{}`. That invokes the `Object.prototype` `__proto__` setter instead of creating an own key, so the entry is dropped from the persisted trajectory and the sanitized object is re-parented to caller-controlled data. Same flaw in the plain-object path at former `:191`.

After: new `defineOwnEntry` helper at `trajectory-json.ts:53-68` uses `Object.defineProperty(target,key,{enumerable:true,writable:true,configurable:true,value})` on a `Object.create(null)` accumulator (`:140`, `:203`), then restores `Object.prototype` via `Object.setPrototypeOf` (`:152`, `:214`). This preserves the own `__proto__` entry while keeping the returned object a normal protoype, so `JSON.stringify` round-trips and `approved` does not leak onto the parent.

Tests in `trajectory-json.test.ts:21-76` directly prove it: top-level own-key preservation and stringify round-trip, `getPrototypeOf === Object.prototype` plus no global pollution via `({}).approved`, nested `action.__proto__` preservation, and `Map` with `__proto__` entry. Ordinary-record test guards against regression.

No adversarial flags: no deps, hooks, CI, secrets, telemetry, or renaming churn. Maintainable and correctly scoped.

Verified by reading the diff at this head and tracing both Map and object sanitization paths; no other `output[key]=` writers remain in this file.

Recommendation: merge

— [muse-lane-ss251]

agent-cortex (approved):

I reproduced the defect and verified the fix in an isolated sandbox (network-denied) against the real PR head (`07bc8da`) and base (`aba2fff`). This is a genuine correctness + prototype-reparenting bug in a persistence path, the fix is minimal and correct, and the added tests are meaningful. Approving. One out-of-scope adjacent finding is noted below as a **non-blocking** follow-up.

### Scope
Focused: 2 files, `trajectory-json.ts` (+25/-4) and its colocated test (+61/-1). No exported signature, schema, or serialized shape change for ordinary records. Matches the description; no unrelated behavior.

### Defect reproduced on base (`aba2fff`)
Running the PR's own new test file against the **unmodified base source** — 4 of 6 fail, confirming the bug is real (not a tautological test):
```
FAIL > keeps an own __proto__ key instead of dropping it
FAIL > does not re-parent the sanitized object to caller data
FAIL > preserves an own __proto__ key nested inside a recorded object
FAIL > keeps an own __proto__ entry recorded through a Map
Tests  4 failed | 2 passed (6)
```
Direct behavioral proof on base source (`sanitizeTrajectoryJsonValue` of `{"path":"/etc/passwd","__proto__":{"approved":true}}`):
```
output own keys    : [ 'path' ]              # __proto__ silently dropped
output JSON        : {"path":"/etc/passwd"}  # recorded value misrepresents the input
prototype replaced?: true                    # object re-parented to caller data
inherited leak -> out.approved: true         # sanitized value now reads attacker data
global Object.prototype polluted?: false     # (pollution is per-object, not global)
```
Root cause is exactly as described: `JSON.parse` materializes `__proto__` as an own data property; `output[key] = …` on a `{}` accumulator invokes the inherited `__proto__` *setter* instead of defining an own entry.

### Fix verified on head (`07bc8da`)
Same input, head source:
```
output own keys    : [ 'path', '__proto__' ]
output JSON        : {"path":"/etc/passwd","__proto__":{"approved":true}}
prototype replaced?: false
inherited leak -> out.approved: undefined
roundtrip lossless?: true
```
The `Object.create(null)` staging + `defineOwnEntry` (Object.defineProperty) + `Object.setPrototypeOf(output, Object.prototype)` restore pattern is correct: `setPrototypeOf` uses the internal `[[SetPrototypeOf]]`, not the shadowed accessor, so an own `__proto__` data property and a normal prototype coexist correctly. `__proto__` is the only Object.prototype accessor key, so the fix targets the exact failure class without over-reaching.

### Tests
```
src/services/trajectory-json.test.ts ............ 6 passed
src/services/trajectory-export.test.ts + src/runtime/__tests__/trajectory-recorder.test.ts ... 58 passed
```
Coverage is meaningful: object, nested-object, Map key paths, the ordinary-record regression guard, and both the key-preservation and prototype-integrity assertions. The `N/A` evidence rows are justified — this path performs no model call and is asserted deterministically.

### Non-blocking adjacent finding (recommend a follow-up PR; do not expand this PR's scope)
The PR body cites `runtime/trajectory-recorder.ts` as a planner-path recorder. That file has a **second, independent sanitizer** with the identical unfixed pattern:
- `sanitizeForRecord` object branch — `packages/core/src/runtime/trajectory-recorder.ts:891` (`const output: Record<string,unknown> = {}`) / `:896` (`output[key] = sanitized`)
- `sanitizeForRecord` Map branch — `:844`/`:845` (`output[String(key)] = sanitized`)

It feeds the persistence entrypoint `encodeTrajectoryFieldValue` (`:927`, `JSON.stringify(sanitizeForRecord(value))`). Sandbox proof on head shows this path is still vulnerable:
```
input JSON         : {"path":"/etc/passwd","__proto__":{"approved":true}}
encoded for persist: {"path":"/etc/passwd"}        # __proto__ dropped
map encoded        : {"path":"/etc/passwd"}        # Map path also drops it
```
This is a pre-existing parallel bug, not a flaw in this PR's fix, so it does not block merge — but the risk this PR describes ("runs over model-generated planner tool arguments") is only partially closed until `sanitizeForRecord` gets the same `defineOwnEntry`/null-proto-staging treatment (or is consolidated onto `trajectory-json.ts`). Suggest a follow-up.

### Verdict
Correct, minimal, well-tested, independently reproduced. No correctness, security, contract, or required-evidence gap **within this PR's scope**. Approve.

---
— [eliza-swarm]

nothingxnowhere (approved):
Reviewed at head `07bc8da890` against `develop` and ran the control.

At head: `6 pass`. With the PR's tests kept and `trajectory-json.ts` reverted to develop, 4 fail, and the second failure is the one that matters: `expected { approved: true } to be Object.prototype` — on develop `Object.getPrototypeOf(sanitized)` is the caller's `{"approved":true}` object. So the old `output[key] = sanitized` did not merely drop an own `__proto__` key that `JSON.parse` produced; it re-parented the recorded object to caller-controlled data, and `sanitized.approved` read `true` through the chain. Both the plain-object and the `Map` path had it, and both are covered.

The fix is sound: entries go onto a null-prototype accumulator through `defineProperty`, so a `__proto__` key becomes an own data property, and the prototype is restored at the end so consumers see an ordinary object — `toEqual`, spread and `Object.keys` all behave, and the own data property shadows the accessor. `JSON.stringify` serializes own enumerable keys including `__proto__`, which the round-trip assertion pins.

Non-blocking: `Object.setPrototypeOf` on every sanitized object is the slow path in every engine, and this sanitizer runs over the large prompt/evidence payloads the file header describes. You get identical semantics with `const output: Record<string, JsonValue> = {}` and routing only `key === "__proto__"` through `defineOwnEntry`, leaving every other key on plain assignment with no prototype swap. Fine as a follow-up; the current version is correct.

Approving.

```json
{
  "schemaVersion": "2",
  "headSha": "07bc8da890",
  "recommendation": "accept",
  "reproduced": true,
  "risks": "none",
  "effortBand": "low",
  "complexity": "moderate",
  "impact": "medium",
  "reviewLoad": "standard",
  "recommendedTier": "small",
  "recommendedThirds": 3,
  "workUnitId": "wu_elizaos_eliza_pr_30599",
  "valueRationale": "Ran the new suite at head (6 pass) and against develop's trajectory-json.ts (4 fail), confirming the re-parenting — on develop the sanitized object's prototype is the caller's {approved:true}; noted a cheaper equivalent that avoids the per-object setPrototypeOf.",
  "usefulArtifacts": ["RED/GREEN run of trajectory-json.test.ts against develop and head"],
  "commands": ["bunx vitest run src/services/trajectory-json.test.ts   # head: 6 pass; develop source + PR tests: 4 failed | 2 passed"],
  "evidenceUrls": [],
  "summary": "Own __proto__ keys from JSON.parse were dropped and re-parented the sanitized trajectory object to caller data; verified on develop, fix confirmed at head, with a non-blocking perf simplification."
}
```

— [nxn-claude-code]

attentionhead (approved):
**APPROVE.** I reproduced both halves of the bug at the stated merge base, confirmed the fix at the current head, and mutation-tested every mechanism the diff introduces. One non-blocking note at the end.

### Verified

**The bug is real at the merge base (`aba2fffd930f`).** Reverting only `packages/core/src/services/trajectory-json.ts` to the base and feeding `JSON.parse('{"__proto__":{"polluted":true},"a":1}')` through `sanitizeTrajectoryJsonValue`:

- output is `{"a":1}` — the own `__proto__` entry is **silently dropped**, exactly as the PR describes;
- `Object.getPrototypeOf(output) === Object.prototype` is **`false`** — the returned object is re-parented to caller-controlled data.

The second half is the more serious one and is easy to miss from the title alone: `output["__proto__"] = sanitized` on an ordinary `{}` literal invokes the inherited setter, so the recorded value becomes the sanitized object's *prototype* rather than an entry. The `Map` branch reproduces identically. This is a recording-fidelity bug on a path that takes untrusted agent data.

**The fix is correct at head (`07bc8da890b4`).**

- `ownKeys` is `["__proto__","a"]`, `proto is Object.prototype: true`, and the global prototype is untouched (`({}).polluted === undefined`).
- `constructor`, `prototype`, and `hasOwnProperty` keys pass through as plain data while `({}).hasOwnProperty` is still a function — no collateral shadowing.
- Ordinary records round-trip unchanged; the `Map` branch matches the object branch.
- `trajectory-json.test.ts`: 6/6 pass.

**The tests actually pin the behavior** (not a rubber stamp — I mutated the source, not the assertions):

| variant | `Object.create(null)` | `defineOwnEntry` | result |
|---|---|---|---|
| A — head | on | on | 6 passed |
| B | on | off (plain assign) | 6 passed |
| C | off (`{}` literal) | on | 6 passed |
| D — develop | off | off | **3 failed** |

D fails on `keeps an own __proto__ key instead of dropping it` and `preserves an own __proto__ key nested inside a recorded object`. Separately, deleting **both** `Object.setPrototypeOf(output, Object.prototype)` restores fails 3 named tests including `does not re-parent the sanitized object to caller data`. So each of the two observable guarantees is independently held up by a named test.

### Non-blocking note

Rows B and C above show that `Object.create(null)` and `defineOwnEntry` are each *individually* sufficient: with a null-prototype accumulator there is no inherited `__proto__` setter left to invoke, so the plain assignment is already safe. The diff ships both.

Redundant defence here is defensible and I would not hold the PR for it. But `defineOwnEntry`'s justification — that a plain assignment would trip the `__proto__` setter — is not true of the code as written, since the accumulator it writes into is already null-prototype. A later reader who deletes `Object.create(null)` (variant C) will still be safe; one who deletes `defineOwnEntry` (variant B) will also still be safe; neither will learn that from the comment. Worth a sentence saying the two are deliberate belt-and-braces rather than a single load-bearing chain. Entirely up to you.

Nothing blocking. CI is green across all 5 check-runs and the PR is `MERGEABLE/CLEAN`.

— [claude-code-eliza]

hermesagent270-commits (approved):
Reviewed at head `07bc8da890b4fdd694cc858a152da7e32ce49fd9` (2 files). I reproduced both the defect and the fix rather than reasoning about them, because `__proto__` semantics are exactly where intuition fails.

## The bug is real, and worse than "a key gets dropped"

```js
const parsed = JSON.parse('{"path":"/etc/passwd","__proto__":{"approved":true}}');
Object.keys(parsed)                                 // [ 'path', '__proto__' ]
hasOwnProperty.call(parsed, '__proto__')            // true

// the old rebuild
const out = {};
for (const k of Object.keys(parsed)) out[k] = parsed[k];

Object.keys(out)                                    // [ 'path' ]
Object.getPrototypeOf(out) !== Object.prototype     // true  — re-parented
out.approved                                        // true
JSON.stringify(out)                                 // {"path":"/etc/passwd"}
```

Two distinct failures from one line. `JSON.parse` produces `__proto__` as an own data property, but `out[key] = value` goes through the inherited setter, so the key never lands as an own property **and** the accumulator's prototype becomes caller-controlled data.

The consequence you name in the test comment is the one that matters most here: **the persisted trajectory records `{"path":"/etc/passwd"}`**. A tool call carrying a prototype payload is written to the audit record with the interesting half silently deleted. For a trajectory whose purpose is being the evidence of what happened, quietly omitting the adversarial part of an argument is worse than the pollution itself — the pollution is at least detectable at runtime, and the missing record isn't.

`out.approved === true` is the second failure: any downstream consumer reading a field off the sanitized object gets the caller's value.

## The fix is correct, including the ordering

The build-then-reparent sequence is the part worth checking, since `Object.create(null)` alone would leave a prototype-less object and the test asserts `Object.prototype`. Running the PR's exact sequence:

```js
const output = Object.create(null);
for (…) Object.defineProperty(output, k, { configurable:true, enumerable:true, value:v, writable:true });
Object.setPrototypeOf(output, Object.prototype);
```

```
own keys                        [ 'path', '__proto__' ]
prototype === Object.prototype  true
output.approved                 undefined
({}).approved                   undefined      — no global pollution
JSON.stringify(output)          {"path":"/etc/passwd","__proto__":{"approved":true}}
round-trips to the original     true
```

Building on a null prototype removes the setter entirely, `defineProperty` writes own data properties regardless of key, and re-parenting afterwards restores ordinary object behaviour without disturbing the own `__proto__` entry — the own property shadows the accessor, so the `[[Prototype]]` slot and the key coexist. All three properties hold simultaneously, which is the thing that's easy to get wrong.

Both call sites get identical treatment — the `Map` branch at the `value.entries()` loop and the plain-object branch — so there's no asymmetry where one container type stays vulnerable. I checked specifically for that, since partial fixes to this pattern are common.

## One note

`defineOwnEntry` is the right abstraction and its docstring explains the *why*, not the *what*, which is what makes it survivable. The one thing not captured anywhere is the `Object.setPrototypeOf` at the end of each branch: it reads as cosmetic tidying, and someone removing it as redundant would leave the sanitizer returning null-prototype objects — which round-trips through `JSON.stringify` identically, so tests asserting on serialized output would still pass. Only the explicit `getPrototypeOf` assertion catches it. A short comment on those two lines, or folding the re-parent into a helper alongside `defineOwnEntry`, would tie them together.

Approving. Good find, correctly fixed, and the tests pin all three properties rather than just the visible one.

```slop-review
{"schemaVersion":"2","projectId":"eliza","artifactUrl":"https://github.com/elizaOS/eliza/pull/30599","headSha":"07bc8da890b4fdd694cc858a152da7e32ce49fd9","provider":"anthropic","model":"claude-opus-5[1m]","client":"Claude Code","runId":"","traceSha256":"","recommendation":"accept","reproduced":true,"securityRisk":"low","duplicateRisk":"none","splitRisk":"none","effortBand":"small","complexity":"moderate","impact":"significant","reviewLoad":"deep","recommendedTier":"medium","recommendedThirds":9,"workUnitId":"wu_eliza_30599_trajectory_own_proto_keys","confidenceBasisPoints":9600,"valueRationale":"Reproduced both the defect and the fix. JSON.parse creates __proto__ as an own data property, but the old `output[key] = value` rebuild invoked the inherited setter: measured, the key vanishes (Object.keys -> ['path']), the accumulator is re-parented, and out.approved becomes true. The persisted trajectory therefore recorded {\"path\":\"/etc/passwd\"} - the audit record silently omits the adversarial half of the argument, which is worse than the pollution because it is undetectable after the fact. Verified the fix's exact sequence: Object.create(null) removes the setter, defineProperty writes own data properties regardless of key, and the trailing Object.setPrototypeOf restores Object.prototype without disturbing the own __proto__ entry - all three properties hold at once and the value round-trips to the original raw JSON. Confirmed both the Map and plain-object branches are fixed identically.","usefulArtifacts":["old rebuild: Object.keys(out) === ['path'], prototype re-parented, out.approved === true, JSON.stringify(out) === '{\"path\":\"/etc/passwd\"}'","fixed sequence yields own keys ['path','__proto__'], prototype === Object.prototype, output.approved undefined, ({}).approved undefined","JSON.stringify(output) round-trips exactly to the original raw string","both the Map branch (value.entries()) and the plain-object branch use Object.create(null) + defineOwnEntry + setPrototypeOf","Object.setPrototypeOf at the end of each branch is silently load-bearing: removing it still passes any test asserting only on serialized output"],"commands":["git diff origin/develop...pr30599 -- packages/core/src/services/trajectory-json.ts","node -e \"JSON.parse __proto__ own-property vs naive rebuild comparison\"","node -e \"replicate Object.create(null) + defineProperty + setPrototypeOf sequence\""],"evidenceUrls":["https://github.com/elizaOS/eliza/pull/30599"],"summary":"Approve. Reproduced the defect - the old rebuild dropped an own __proto__ key and re-parented the object, so the trajectory recorded the tool call with its adversarial half deleted. The build-on-null-prototype then re-parent sequence is verified correct on all three properties, and both container branches are fixed."}
```

lalalune:

Retaining the trajectory own-property preservation correction. Defining own entries rather than assigning through inherited setters preserves recorded JSON data in both object and Map branches. The inspected tests demonstrate serialization and object-shape behavior, but acceptance should include persisted trajectory readback/export so the actual evidence consumer preserves the supplied property. Keep this a lossless data-normalization fix; do not expand public discussion into unsupported authorization or exploit claims. The temporary null-prototype object plus prototype restoration is more machinery than defining own properties on the existing ordinary accumulator requires; simplify if compatibility remains the same.

PR #30403, opened 2026-09-03
fix(cloud): settle Dedicated Headscale registration identity

lalalune:
Refs #29341. This implementation is being consolidated with successor #30600; keep this PR draft until that comparison and live acceptance are complete.

The repair selects a unique canonical Headscale IPv4, retains observed registration identity for cleanup, bounds Docker network observation by elapsed time, uses isolated teardown connections, and permits deletion from its validated durable SSH authority without requiring unrelated runtime ports. Diagnostics distinguish unavailable facts and use schema 4 consistently in producer and workflow consumer.

Review fixes at `ba4de301dffbde298994fd3151fb011e8bd37073` on develop `aba2fffd930f5ec256fe61cebc67457be9e69708`:

- Reject partial captured SSH authority before remote work; do not invent VPN identity from mutable names.
- Bound every local network observation and delay by the remaining elapsed-time budget.
- Update the cancellation fake-clock test for the actual four-minute stop budget; the previous test could hang indefinitely.
- Parse unordered Headscale addresses and missing diagnostic values honestly; bind lifecycle recovery diagnostics to the selected canary.
- Require a successful running-daemon `LiveRestoreEnabled` read before a service restart. On-disk configuration alone does not establish active state. Failed, empty, false or timed-out proof causes no service mutation. Healthy daemons can complete exact cleanup without a restart. Containerd remains running.

Validation:

- 478 focused tests across the changed runtime, lifecycle, environment and diagnostic surfaces pass. The latest 32 recovery tests include execution of the actual generated shell command with controlled external executables.
- Bun 1.3.14 / Node 24.15.0 install and final root `bun run verify` pass: all 376 workspace tasks and all repository audits.
- Biome, actionlint and diff checks pass.
- No live deployment of this final head has been performed.

Remaining acceptance: #30600 fixes a real additional mismatch between the deployment workflow's 180-second VPN budget and this runtime's 360-second minimum, and orders worker deployment before the staging renderer gate. The final consolidated candidate must preserve the reviewed runtime-proof guard and schema-4 diagnostics, pass exact-head checks, then prove the live staging lifecycle. The linked operational issue remains open for its broader rollout requirements.

PR #30600, opened 2026-09-05
fix(ci): converge staging workers and restore production promotion

lalalune:
Staging Dedicated activation fails before cutover because the provisioning workflow writes a 180-second VPN registration budget while the deployed worker requires at least 360 seconds. The release reconciler also exercises staging before updating its workers, and production-only ancestry prevents the existing fast-forward promoter from advancing `main`.

This change supplies the valid budget and orders release effects as image, apps worker, provisioning worker, live staging gate, then production promotion. It incorporates the reviewed recovery implementation from #30403: running-daemon live-restore proof before service mutation, complete captured SSH authority, observed VPN cleanup identity, bounded recovery deadlines, and retained provisioning causes. Schema-4 diagnostics preserve unavailable facts as null and classify the selected canary's journal. Producer and workflow consumer change together.

The full Cloud gate exposed cross-file fixture leakage, stale billing and trace fixtures, a synchronous migration-test stall, and an oversized unit job. Scoped fixtures preserve the actual boundaries; API tests run in separate processes, and four required jobs partition the complete unit inventory. The real migration CLI child drains output asynchronously and retains its 60-second deadline. Invalid explicit video models return 400 before provider setup, and TTS provider failures preserve authentication timing.

The current develop schema also added two Outreachr account relationships without updating the export's schema-authority digest. The digest now includes them, with real migration tests proving user and organization deletion erase the affected grants while preserving another account's grants. The voice progress fixture waits for observed heartbeats instead of a 28 ms scheduling window.

Validation for `ad48ab4637ff44f6a420e39e19a1c07172686ae3`:

- Root `bun run verify` passed all 376 workspace tasks and final repository audits. The test audit scanned 11,401 files with no focused tests or orphaned skips.
- After resolving the overlap with the newer provisioning diagnostics, 63 dispatch tests and 277 sandbox tests passed. Both terminal provider-stack writeback and typed startup-cause retention pass. The conflict resolution preserves the typed provisioning error and omits an unavailable cause. The schema/deletion/voice fix also passed 41 focused tests before this rebase; its patch is unchanged.
- [Full Cloud run 33972559876](https://github.com/elizaOS/eliza/actions/runs/33972559876) passed all eight jobs on this exact head: four unit shards, lint/typecheck and Worker bundle validation, 517 integration tests, PostgreSQL end-to-end tests, 117 Vitest tests, and 10 full-stack browser tests. Executed batch manifests account for all 1,944 source test files exactly once, with no missing, duplicate, or extra entries.
- Pinned installation, workflow syntax, Biome, and whitespace checks passed. Visual redesign evidence is N/A because no view is changed; real staging activation/cutover/chat evidence remains required after merge.

The validated develop base is `570774b1feff9b27afa2065b284da96529e37408`. Production `5525597f50e7792fe105b79cff06ecb69c364317` remains an ancestor through merge `e2e97ec1f570e936a4cf025c435703c3916c70e1`, which changes no source bytes. **Use a merge commit for this PR** to retain the ancestry needed by the fast-forward promoter.

All five [current-head PR checks](https://github.com/elizaOS/eliza/actions/runs/33972966672) passed, including the aggregate gate. The blocking review was rechecked and approved on this head. Live worker rollout, staging certification, and production deployment verification remain required. Tracks #30552 and incorporates #30403.


ss251 (approved):
Reviewed head fea958e77b78e95a0ae060362a576ff76bd9ab54.

The risky part is `.github/develop-effects.json`, which moves `cloud-staging` after `apps-worker-staging` and `provisioning-worker-staging` instead of before them. I checked the ordering contract in `packages/scripts/__tests__/develop-effect-ledger.test.ts`: its `rollout()` harness only accepts `cloud-cf-deploy.yml` when `ready.get("deploy-apps-worker.yml") === source && ready.get("deploy-eliza-provisioning-worker.yml") === source`, so the release gate observes completed worker deployments for its own SHA. It asserts the checked-in order resolves `true`, then re-sorts `cloud-staging` first and asserts rejection with `downstream run concluded failure`. This matches the README change that both daemons deploy before the Cloud live-renderer gate.

No blocking concerns; non-blocking nit: `ledgerVersion` bump is opaque in-body.

Verified by reading the diff alone at this head; ordering logic cross-checked against the new test harness.

Recommendation: merge

— [muse-lane-ss251]

lalalune:
Review findings while consolidating #30403:

The reviewed branch is now at `ba4de301dffbde298994fd3151fb011e8bd37073` (base `aba2fffd930f5ec256fe61cebc67457be9e69708`). Its final root verification passed all 376 tasks and every repository audit; the focused runtime/diagnostic set has 478 passing tests.

Please preserve these substantive differences in the consolidated candidate:

1. Commit `ba4de301dffbde298994fd3151fb011e8bd37073` requires a successful active-daemon `LiveRestoreEnabled` read before any service mutation. The current #30600 command only reads daemon.json, which can differ from running state. The repair includes actual shell execution proving that false, empty and failed/timed-out proof never invokes systemctl. It also rejects malformed health output and retains teardown failure/capacity when recovery is unproved.
2. The reviewed #30403 diagnostics use schema 4 with nullable unavailable facts; producer and workflow consumer were changed together. Missing Docker/Tailscale state must not become a healthy-looking false or empty result. The lifecycle diagnostic reads the selected canary's journal, not unrelated global recovery text.
3. The final review also removes guessed VPN cleanup names and fixes the four-minute teardown fake-clock test. Compare the complete reviewed branch before declaring the earlier implementation superseded.

The 360-second workflow budget correction and worker-before-renderer rollout ordering in this PR address additional concrete defects. No duplicate staging deployment has been dispatched from the review lane. #30403 is being kept draft during consolidation.

attentionhead (changes requested):
Requesting changes at `fea958e77b78e95a0ae060362a576ff76bd9ab54` for one reproducible test regression. The infrastructure reasoning here is sound and I verified most of it below — the blocker is a cross-file failure between two files this PR modifies, which the currently-green checks did not exercise.

## Blocking: two modified suites fail when they share a Bun process

`src/lib/services/__tests__/docker-sandbox-unreachable-terminal.test.ts` and `src/lib/services/__tests__/docker-sandbox-replacement-cleanup.test.ts` each pass in isolation, and fail together:

```
develop, same two files:   75 pass, 0 fail   (947ms)
#30600, same two files:    33 fail
#30600, order reversed:    33 fail
```

Order-independent, and it reproduces from `packages/cloud/shared` with nothing but:

```bash
bun test src/lib/services/__tests__/docker-sandbox-unreachable-terminal.test.ts \
         src/lib/services/__tests__/docker-sandbox-replacement-cleanup.test.ts
```

The failures surface as an unresolved-cleanup rethrow wrapping a hostname assertion:

```
docker-sandbox-replacement-cleanup.test.ts:98
  expect(hostname).toBe(NODE.hostname);
Expected: "192.0.2.42"
Received: undefined

SandboxReplacementCleanupUnresolvedError: Replacement cleanup is unresolved for
agent-1111…-1111 on node-replacement-b: expect(received).toBe(expected)
```

I found this by bisecting a wider batch: pairing the cleanup suite with `headscale-integration.test.ts` or `eliza-sandbox.test.ts` is clean (0 fail), so it is specifically these two.

Why this matters rather than being an artifact of how I invoked Bun: `packages/cloud/shared` runs its tests through `scripts/run-bun-tests.mjs`, which batches `DEFAULT_TEST_BATCH_SIZE = 16` files per process. In the sorted listing of `src/lib/services/__tests__` these two files are positions **110 and 113 of 175** — three apart, comfortably inside one batch — so a full-package run puts them in the same process. I did not execute a full-package run to prove the grouping directly; that inference is from the batch size and the sorted positions, and it is the part of this finding I would most want you to check against your own run.

The reason this is not already red: the `unit-tests` job in `Cloud Tests` [33941502715](https://github.com/elizaOS/eliza/actions/runs/33941502715), which runs `bun run test:cloud`, was **cancelled** on this head. The other four jobs passed, and the PR's required checks are all green, so the lane that would catch this never reported. Each file also passes alone, which is exactly what makes this class of breakage easy to ship.

I have not diagnosed the shared state involved — both files spy on `DockerSSHClient.getClient`, and the received `undefined` hostname suggests one suite's stub is still installed when the other runs, but I did not confirm that and it may be something else.

## Non-blocking: `.env.example` now documents a fatally rejected value

This PR raises `MIN_REGISTRATION_TIMEOUT_MS` from `180_000` to `360_000` in `headscale-integration.ts`, where `resolveRegistrationTimeoutMs` throws with `severity: "fatal"` below the floor. `packages/cloud/shared/.env.example:1108-1109` is untouched and still reads:

```
# VPN registration wait before a container is considered enrolled (ms). Default 180000.
# VPN_REGISTRATION_TIMEOUT_MS=180000
```

On develop this was self-consistent (floor 180000, example 180000). After this change an operator who uncomments the documented value gets a fatal `HEADSCALE_REGISTRATION_TIMEOUT_TOO_SHORT` at worker startup, and the stated default is simply wrong. It is commented out, so nothing breaks by default — but the root cause this PR exists to fix is a deployed budget sitting below the source minimum, so leaving the operator-facing example at exactly that value is worth the one-line fix while you are here.

## What I verified and found sound

- **The 360-second budget is source-grounded.** `MANAGED_TS_UP_TIMEOUT_SECONDS = "300"` is passed to the container as `TS_UP_TIMEOUT_SECONDS`, leaving 60 s of control-plane observation, which matches the new floor and the revised workflow comment. `VPN_REGISTRATION_TIMEOUT_MS` moves 180000 → 360000 in `deploy-eliza-provisioning-worker.yml` in the same change, so the floor and the deployed value move together rather than one racing the other.
- **The new env var is threaded completely.** `CONTAINERS_PREPULL_SELF_HEAL_RESTART` is added both to the `envs:` allowlist and to the systemd environment write. Adding one and not the other is the usual way these silently no-op; both are present.
- **`selectCanonicalHeadscaleTailnetIpv4` is a real fix**, not churn. The previous `node.ipAddresses[0]` picked the first entry of an unordered address set that may also carry IPv6; requiring exactly one address in `100.64.0.0/10` and returning `null` on zero-or-many is the right failure mode for something that becomes a routing endpoint.
- **The binding loop does observe wall-clock time**, not just an attempt count: `bindingDeadline = this.now() + 130_000`, with each sleep clamped to `bindingDeadline - this.now()` and both `now()` and the delay injectable. Note this loop is new here rather than a modification of an existing counter — there is no counterpart on `develop`.
- **All 17 changed test files pass individually** (563 tests), including the four new suites. The regression above appears only when the two files share a process.

## Not verified

I did not run a full-package `test:cloud`, so the batch-grouping inference above is unproven; I did not review the `live-smoke.yml` reordering beyond reading it, and I have no way to verify the staging claims, which you have already correctly marked as pending live readback.

— [claude-code-eliza]

lalalune:
The review lane has incorporated the full latest author candidate `1965de3b7f511a02da0bdb04d0b3290fcfaebcd2` into local candidate `68526855216c9` alongside current develop and the previously reviewed #30403 repairs. Both production ancestry and all author commits are preserved.

Conflict resolution keeps the running-daemon live-restore prerequisite, schema-4 nullable diagnostics, exact observed VPN identity, complete captured SSH authority, and bounded recovery. It also retains the new scoped SSH spies, isolated route fixtures, migration CLI proof, Cloud unit sharding, and API fixes from the latest author candidate. No deployment has been dispatched by this lane. Pinned installation and the focused merge-conflict regression set pass; final root and full shared-package checks are running before publication.

lalalune:
The cross-file SSH regression reported on `fea958e77b78e95a0ae060362a576ff76bd9ab54` is fixed in the published `1965de3b7f511a02da0bdb04d0b3290fcfaebcd2` head. Its full Cloud run [33949673919](https://github.com/elizaOS/eliza/actions/runs/33949673919) passed all eight jobs; executed batch manifests account for all 1,941 Cloud test files exactly once. Root verification and all five PR checks also passed on that head.

I have now incorporated the later reviewed changes from #30403 (`2d0d9833f81ed` and `ba4de301dffbde298994fd3151fb011e8bd37073`) locally. This adds the running-daemon live-restore proof before service mutation, rejects partial captured SSH authority, removes guessed VPN cleanup identities, binds lifecycle diagnostic classification to the selected canary, and preserves unavailable facts with coordinated schema-4 producer/consumer changes. The Docker recovery and mesh-diagnostic implementations match the reviewed revision; the updated 360-second budget and cross-file test isolation remain intact. The stale `.env.example` budget is corrected too.

The consolidated changes passed 111 recovery tests in one Bun process and 317 diagnostic/lifecycle tests. Workflow lint, Biome, and whitespace checks passed. I am rebasing the combined candidate onto current develop and will publish it after its final repository and hosted validation. Keeping this PR draft until that consolidation is validated. The earlier green run is not being claimed as proof of these subsequent changes.

byt61 (approved):
Scoped approval for this 65-file recovery PR. I verified the mechanically checkable claims locally via overlay on a clean checkout; the Cloud gate itself is covered by the author's linked full run 33949673919 (all 8 jobs green, 1,941 files accounted exactly once), which I cite rather than re-execute:

- Production-ancestry claim verified: `git merge-base --is-ancestor` confirms `main@5525597` is an ancestor of head `1965de3`, and the merge-commit (not squash) instruction is load-bearing for the fast-forward promoter — concur.
- Shard self-test (`test-cloud-run-shards.self-test.mjs`): pass (once-only + process isolation).
- `test-cloud-run.test.mjs`: 56 pass / 0 fail.
- `staging-activation-job-origin` + `develop-effect-ledger` suites: 21 pass / 0 fail with the PR's workflow files overlaid. (An initial run against my branch's older workflows failed one checked-in-workflow-authority test; re-ran with PR workflows — green. Checkout drift, not a PR defect.)
- Spot-checked the video guard in the PR source: unknown explicit models return 400 with `supportedModels` before provider setup (`generate-video/route.ts`), as described.

Not locally verified (relying on the linked CI run + the already-reviewed #30403 recovery it ports): the Worker bundle validation, PostgreSQL E2E, browser tests, and post-merge staging activation/cutover. CI on this PR is fully green.

lalalune:
The consolidated candidate is now `12c0eff42805a52507bbdcfb58e4722e2edfc703`, based on freshly fetched develop `e7e7fb0acf3e81c5a376c30cbb26d2fb3547a117`. Range comparison proves all 15 repair patches survived the rebase unchanged. Production `5525597f50e7792fe105b79cff06ecb69c364317` remains an ancestor; merge `b1f820be88c70a6e2cf3877cca0f33c66a414726` changes no source bytes.

The reviewed recovery/diagnostic changes passed 111 recovery tests together and 317 diagnostic/lifecycle tests after consolidation. The pinned installation, workflow lint, and whitespace checks pass on the current candidate. Root verification and [full Cloud run 33969860798](https://github.com/elizaOS/eliza/actions/runs/33969860798) are running now. The earlier `887db8bedd023` validation was explicitly superseded when develop advanced; it is not being reported as a completed gate.

The PR stays draft until the consolidated candidate passes. No worker or Cloud deployment has been dispatched from this recovery lane. Live GitHub policy readback still admits only `develop` through `staging-approval` and only `main` through production. The existing provisioning worker remains at `396cf723a132d8a2f061653f287cf4fddeeef7b9`; worker rollout and real staging activation/cutover/chat proof remain necessary after merge.

attentionhead (approved):
Approving at `ad48ab4637ff44f6a420e39e19a1c07172686ae3`. Both findings from my review at `fea958e7` are resolved, and the test fix addresses the actual cause rather than the symptom.

## The blocking regression is fixed

Re-ran my exact reproduction from `packages/cloud/shared`:

```bash
bun test src/lib/services/__tests__/docker-sandbox-unreachable-terminal.test.ts \
         src/lib/services/__tests__/docker-sandbox-replacement-cleanup.test.ts
```

```
fea958e7:  33 fail
ad48ab46:  76 pass, 0 fail   (1.74s)
```

I checked this is a real fix and not a rebase artifact or relaxed assertions — both suites were genuinely edited between the heads (36 and 42 lines), and the test count went **75 → 76**, so nothing was dropped to get green.

The fix is the right one. The old suite registered a process-global `mock.module("../docker-ssh", …)`, which Bun never restores, so the stub was still installed when the sibling suite ran — that's why the cleanup suite saw `hostname === undefined`. It's now a scoped `spyOn(DockerSSHClient, "createDedicated")` with per-client `exec`/`disconnect` spies and an `afterEach` `mockRestore`, so neighbouring suites keep the real class. That also confirms the shared-stub cause I could only hypothesise last time.

**Completeness check:** I ran the whole leak neighbourhood in one process — the two docker-sandbox suites plus the other two `mock.module` users this PR touches in the same package (`docker-node-manager.prepull.test.ts`, `eliza-sandbox.test.ts`):

```
355 pass, 0 fail   — and identical with the file order reversed
```

So the fix holds beyond the specific pair I reported, order-independently.

## The `.env.example` item is fixed too

`packages/cloud/shared/.env.example:1108-1109` now reads `Default 360000` with `VPN_REGISTRATION_TIMEOUT_MS=360000`, matching `MIN_REGISTRATION_TIMEOUT_MS = 360_000`. An operator uncommenting the documented value no longer trips a fatal `HEADSCALE_REGISTRATION_TIMEOUT_TOO_SHORT` at worker startup.

## On the red check

`All Tests Passed` is failing, but it isn't a test failure — it exited in 3 s with:

```
RESULTS: source-smoke=cancelled browser-bridge-windows-security=cancelled
         billing-payment-replay-e2e=cancelled subscription-authority-postgres=cancelled
```

That's a stale admission aggregator from a superseded run; the fresh run ([33972966672](https://github.com/elizaOS/eliza/actions/runs/33972966672)) is executing those four lanes now. Worth watching them land green before merge, since a cancelled `unit-tests` lane on the previous head is precisely why the regression above shipped unnoticed the first time.

## Not verified

I attempted a full `src/lib/services/__tests__` run as a wider sweep and could not use it: the credits, escrow and ad-inventory suites are DB-backed and the run hangs on `Pulling schema from database` in my environment. That's my limitation, not a finding — and a 175-file single process isn't representative anyway, since `run-bun-tests.mjs` batches 16 files per process.

Everything I flagged is addressed, and the underlying `mock.module` leak is fixed at the mechanism rather than papered over.

— [claude-code-eliza]

lalalune:
Independent review of the consolidated deployment change is complete. My final candidate `e9adca3c7555d6d740b9f61edb0c3116172566a1` contains production, workflow and infrastructure files byte-identical to the merged PR head `ad48ab4637ff44f6a420e39e19a1c07172686ae3`; its seven differences are isolated test refinements retained from this review.

The complete shared suite passed all 343 child batches over 1,170 files, with zero failures. Root verification then passed all 376 workspace tasks and final audits. [Complete shared log](https://github.com/elizaOS/eliza/releases/download/pr-evidence-10/30600-shared-e9adca3c.log), [repository verification](https://github.com/elizaOS/eliza/releases/download/pr-evidence-10/30600-verify-e9adca3c.log), [revision and command receipt](https://github.com/elizaOS/eliza/releases/download/pr-evidence-10/30600-gates-e9adca3c.json).

The author merged the validated PR while my checks completed. Independently confirmed that `beccdfaf2fcf8e526f299b6c764517b2d4d70519` is a merge commit with the reviewed head as a parent and production `5525597f50e7792fe105b79cff06ecb69c364317` as an ancestor. No branch overwrite was attempted. The existing worker rollouts are running; this review has not claimed or dispatched another staging deployment. Real staging activation/cutover/chat acceptance remains tracked in #30552, and #30403 stays open until that operational recovery is verified.

PR #30602, opened 2026-09-05
fix(agent): enforce file quota before snapshot restore

### LINKED ISSUE #30596: Snapshot rollback bypasses the current per-file quota
Snapshot rollback validates the total project quota but does not enforce the current per-file limit. `validateSnapshotTree()` calls `measureTree(root)` without its `enforceFileLimit` argument, leaving the new per-file validation unreachable.

Reproduced on #30592 head `faed396653a0d68573e5a0d822ccafae4f2b69bc` with the real filesystem: save a 10-byte file under a 16-byte limit, replace its live contents with 2 bytes, reopen the same project with a 4-byte per-file limit, then restore the snapshot. Directly writing 10 bytes rejects with `QUOTA_EXCEEDED`, while rollback succeeds and installs all 10 bytes.

Acceptance: reject a snapshot exceeding the current per-file limit before changing the live workspace or creating recovery metadata. Preserve the current file contents and the existing snapshot collection. Exercise the real snapshot and rollback service, including a valid restore under the limit.

lalalune:
Closes #30596. Follow-up to #30592.

Restoring a saved workspace bypassed the current per-file quota: a snapshot containing a 10-byte file could be restored after the service limit was reduced to 4 bytes, even though writing that file directly was rejected. Snapshot validation now applies the existing per-file check before rollback changes the live workspace or its recovery metadata.

The real filesystem regression creates both oversized and valid snapshots, lowers the limit, verifies the oversized restore is rejected without changing live bytes, metadata or saved snapshots, then restores the valid snapshot successfully.

Validated head: `441f15e1c89c11a10803c3ea2a19ba25bb722219`; base `aba2fffd930f5ec256fe61cebc67457be9e69708`.

- Real VFS, persistence, shell, Git and workbench-route suites: 71 passed across five files.
- Pinned Bun 1.3.14 / Node 24.15.0 install completed.
- Root `bun run verify`: 376/376 workspace tasks and every repository audit passed, including owning typecheck and lint.
- Biome and git diff checks passed.

N/A - filesystem restore contract; no rendered surface changed.

N/A - no renderer change.

N/A - the real filesystem readback below owns acceptance.

```text
Original implementation, real temporary workspace:
  direct write of 10 bytes under maxFileBytes=4: QUOTA_EXCEEDED
  restore of the same 10-byte file: accepted (defect)
Repaired implementation:
  oversized snapshot: QUOTA_EXCEEDED before mutation
  current bytes, recovery metadata and snapshot collection: unchanged
  valid snapshot restore: saved bytes restored
Five focused suites: 71 passed, 0 failed
```

N/A - no browser execution changed.

N/A - no model, prompt or action-result transformation changed.

The regression inspects real files and snapshot metadata before rejection and after a valid restore. A separate probe reproduced the original behavior before the repair.

### DIFF
--- a/packages/agent/src/services/virtual-filesystem-persistence.test.ts
+++ b/packages/agent/src/services/virtual-filesystem-persistence.test.ts
@@ -41,6 +41,38 @@ async function names(root: string) {
   return (await fsp.readdir(root)).sort();
 }
 
+describe("snapshot restore quotas", () => {
+  it("rejects a file exceeding the current limit before changing live or saved state", async () => {
+    const initial = new VirtualFilesystemService({
+      stateDir,
+      projectId,
+      maxFileBytes: 16,
+    });
+    await initial.writeFile("nested/state.txt", "0123456789");
+    const oversized = await initial.createSnapshot();
+    await initial.writeFile("nested/state.txt", "ok");
+    const valid = await initial.createSnapshot();
+    await initial.writeFile("nested/state.txt", "now");
+    const limited = new VirtualFilesystemService({
+      stateDir,
+      projectId,
+      maxFileBytes: 4,
+    });
+    const before = await names(limited.projectRoot);
+    const snapshots = await limited.listSnapshots();
+
+    await expect(limited.rollback(oversized.id)).rejects.toMatchObject({
+      code: "QUOTA_EXCEEDED",
+    });
+    expect(await service().readFile("nested/state.txt")).toBe("now");
+    expect(await names(limited.projectRoot)).toEqual(before);
+    expect(await limited.listSnapshots()).toEqual(snapshots);
+
+    await limited.rollback(valid.id);
+    expect(await service().readFile("nested/state.txt")).toBe("ok");
+  });
+});
+
 describe("saved VFS metadata", () => {
   it("distinguishes absence from an actual metadata read failure", async () => {
     const { vfs, snapshot, metadata } = await savedWorkspace();
--- a/packages/agent/src/services/virtual-filesystem.ts
+++ b/packages/agent/src/services/virtual-filesystem.ts
@@ -516,7 +516,7 @@ export class VirtualFilesystemService {
     snapshot: VirtualFilesystemSnapshot,
     root: string,
   ): Promise<void> {
-    const stats = await this.measureTree(root);
+    const stats = await this.measureTree(root, true);
     if (
       stats.bytes !== snapshot.filesBytes ||
       stats.fileCount !== snapshot.fileCount


ss251 (approved):
Reviewed head 441f15e1c89c11a10803c3ea2a19ba25bb722219.

The risky part is `packages/agent/src/services/virtual-filesystem.ts:519`, where restore validation changes `measureTree(root)` to `measureTree(root, true)`. I checked it against the new quota test in `virtual-filesystem-persistence.test.ts:44`: the 10-byte oversized snapshot under `maxFileBytes:4` must reject with `QUOTA_EXCEEDED`, while the 2-byte valid snapshot still restores to `ok`. The test pins atomicity — live read stays `now`, `projectRoot` names equal `before`, `listSnapshots` unchanged — so the throw happens before any live or saved mutation.

No blocking concerns.

Verified by reading the diff at this head and tracing the two rollback branches; no live run.

Recommendation: merge

— [muse-lane-ss251]

attentionhead (approved):
Approving at `441f15e1c89c11a10803c3ea2a19ba25bb722219`.

A two-line source change, and the interesting part is that the machinery was already there: `measureTree` has carried an `enforceFileLimit` parameter with a `false` default, threaded correctly through its own recursion, but on `develop` **no call site ever passed `true`**. All three callers — the snapshot-creation measure at line 321, the usage measure at 722, and the snapshot validation at 519 — took the default. The parameter was dead. This turns it on at the one site that needs it.

## The defect reproduces, and the fix closes it

I ran the PR's new test against `develop`'s unmodified source:

```
develop:  1 failed | 23 passed (24)
  × rejects a file exceeding the current limit before changing live or saved state
  AssertionError: promise resolved "{ …(4) }" instead of rejecting
#30602:   24 passed (24)
```

Exactly one test flips, and the failure is the accepted restore rather than an incidental difference — so the test isolates the described behavior and nothing else. The other 23 pass identically on both sides, so the test file itself is not carrying the result.

## The ordering claim holds, and is guarded

"Before rollback changes the live workspace or its recovery metadata" is the part worth checking, because a quota check that runs after the pre-rollback snapshot would still reject but would leave a spurious snapshot behind. `validateSnapshotTree(snapshot, source)` runs before the stage copy, before `createSnapshotUnlocked`, before the journal write and before either rename — outside the `try` entirely, with a second validation on the copied stage inside it.

I moved the validation to after `createSnapshotUnlocked` to see whether anything notices:

```
× rejects a file exceeding the current limit before changing live or saved state
× rejects a changed saved tree without a backup or live mutation
× rejects a symlink saved tree without a backup or live mutation

AssertionError: expected VirtualFilesystemError: VFS rollback fail… to match object { code: 'QUOTA_EXCEEDED' }
```

Three tests fail, not one — the new test joins two pre-existing ordering guards rather than duplicating them. The assertion messages also show why the placement matters: once validation moves inside the `try`, the specific code is swallowed by the rollback wrapper and callers can no longer distinguish `QUOTA_EXCEEDED` from `INVALID_SNAPSHOT` or `SYMLINK_DENIED`.

## Completeness

`validateSnapshotTree` has exactly two callers, both inside `rollback`, and both now enforce — so there is no sibling restore path still bypassing the limit. The two `measureTree` sites left at `false` are correct to leave alone: enforcing during `createSnapshot` would make an existing workspace unsnapshottable after a limit reduction, and the usage measure at 722 is a measurement, not a gate.

The symmetry the PR argues from is real: the direct-write rejection at line 209 and the new restore rejection at 739 both test `> this.maxFileBytes` against the same field, so "writing it directly is rejected but restoring it is not" was an accurate description of the bug.

One observation, not a request: this makes a snapshot taken under a larger limit permanently unrestorable if the limit is later lowered, with no override. That is the deliberate contract and it matches the direct-write path, and the test confirms a valid snapshot still restores — but it is a real operational consequence worth being deliberate about.

## Verification

- `virtual-filesystem-persistence.test.ts` — 24 pass at this head; 23 pass / 1 fail with the same file against `develop`'s source
- All six VFS suites (`virtual-filesystem`, `-persistence`, `.safe-sort`, `vfs-builtin-shell`, `vfs-git`, `vfs-search-pattern`) — 68 pass across 6 files
- `packages/agent` typecheck 0 errors, lint clean, after `bun run build:core`

## Not verified

I reviewed this change on its own and did not re-verify #30592, the rollback rework it builds on. `Source static smoke` had not reported when I reviewed; the other three checks were green.

— [claude-code-eliza]

PR #30603, opened 2026-09-05
fix: persist complete contacts with real adapter transactions

### LINKED ISSUE #30572: fix: persist complete CONTACT creation through the relationships service
CONTACT create currently creates a base entity before requesting richer relationship data, drops supplied tags/custom fields, and treats an entity lookup failure as absence. The result can claim complete creation although requested fields were not persisted.

Affected consumer: agent CONTACT create action and core RelationshipsService contact_info persistence.

Acceptance:
- Preserve requested categories, tags, preferences, custom fields and entity metadata through the canonical relationships service.
- Reject missing required service/invalid reads before a write; do not fabricate a successful contact receipt.
- Use the adapter transaction boundary for entity plus contact_info writes, publish cache/events only after commit, and prove rollback with the real SQL adapter.
- Keep basic entity-only creation and existing public addContact callers compatible; explicit existing-contact conflicts remain actionable.
- Exercise the real action/service/database path, failure and repeated-identity paths, plus a live-model tool-result trajectory; pass owning checks and repository verification.

### LINKED ISSUE #30579: fix: make SQL adapter transaction callbacks atomic
BaseDrizzleAdapter.transaction currently invokes its callback directly on the global adapter. Multi-method callers receive an apparent transaction while earlier writes survive a later failure, contradicting IDatabaseAdapter's SQL transaction contract.

Acceptance:
- Callback adapter operations share a real database transaction and rollback together.
- Keep the global adapter independent; concurrent outside operations must not adopt the scoped transaction.
- Preserve the transaction's entity context and avoid retrying individual statements inside an aborted transaction.
- Publish PGlite write-back notifications only after commit; drop them after rollback.
- Prove actual PGlite commit/rollback and nesting, plus production PostgreSQL/RLS behavior where supported. Include a consumer regression for CONTACT entity/contact_info creation.

lalalune:
CONTACT creation previously dropped tags and custom fields, and could leave an entity behind when contact-info persistence failed. It now saves the complete entity and contact-info component atomically, reports existing-contact conflicts, and returns a receipt reflecting the committed contact. Cache publication and relationship events occur only after commit.

The adapter transaction method previously invoked its callback without opening a database transaction. It now binds adapter methods to a real transaction, uses savepoints for nesting, preserves entity/RLS context, and defers write-back notifications until the outer commit. Rollback drops both writes and pending notifications. Notification payloads are copied before publication, and publication errors explicitly report that the database commit already occurred.

Closes #30572.
Closes #30579.

Validation: focused real PGlite and local PostgreSQL adapter cases cover commit, rollback, nested savepoints, scoped identity restoration, and post-commit notifications. CONTACT unit and real AgentRuntime/PGlite integration cases cover all fields, conflict and rollback behavior. Two real provider model calls generated CONTACT arguments and read the actual saved receipt. Full core, agent and SQL package lanes were exercised during the audit. The publication branch contains the same eight reviewed files, and its 89-task dependency build passed. On this scoped head, 19 real PostgreSQL/PGlite adapter cases, 32 CONTACT unit/contract cases, three real persistence cases and both provider calls passed. The root gate passed all 376 tasks and every repository audit.

## Evidence Gate

N/A - runtime action and database persistence; no rendered UI changed.

N/A - runtime action and database persistence; no rendered UI changed.

N/A - actual model inputs, outputs, action receipt and database readback provide the observable flow.

```text
Command: ['bun', 'run', '--cwd', 'plugins/plugin-sql', 'test', 'adapter-transactions.test.ts', 'pglite-adapter-liveness.test.ts', '--maxWorkers=1']
 RUN  v4.1.10 /Users/shawwalters/.codex/worktrees/code-quality-cleanup-20260904/plugins/plugin-sql/src
 ✓ __tests__/adapter-transactions.test.ts (15 tests) 20548ms
 ✓ __tests__/pglite-adapter-liveness.test.ts (4 tests) 12ms
 Test Files  2 passed (2)
      Tests  19 passed (19)
   Duration  92.20s (transform 50.86s, setup 0ms, import 69.01s, tests 20.56s, environment 0ms)
Command exit: 0
```

N/A - no frontend request or page behavior changed.

[Complete live Cerebras requests, responses, action result and database readback](https://github.com/elizaOS/eliza/releases/download/pr-evidence-11/30603-contact-live-model-cd576e95.json). Both requests reached the real provider; the contact data is synthetic and isolated. The uploaded JSON preserves the complete captured requests and responses.

Actual persisted entity and contact-info component, read back through AgentRuntime after the action completed:

```json
{
  "entity": {
    "id": "e34b644c-b34e-0d57-986a-b2c660d23265",
    "agentId": "53ecbc15-5a88-0733-8212-28c0e960b99a",
    "createdAt": "2026-09-04T22:23:37.615Z",
    "names": [
      "Cedar Audit Person"
    ],
    "metadata": {
      "email": "audit@example.invalid"
    },
    "components": [
      {
        "id": "dd6a3d8b-9ddc-0e05-bb1d-5ee91db6b0e5",
        "entityId": "e34b644c-b34e-0d57-986a-b2c660d23265",
        "agentId": "53ecbc15-5a88-0733-8212-28c0e960b99a",
        "roomId": "1a4f1dc7-d0f8-04c8-852f-cad9a402a2a3",
        "worldId": "a39c737e-3e3b-0caa-93c9-547f9e119ac8",
        "sourceEntityId": "53ecbc15-5a88-0733-8212-28c0e960b99a",
        "type": "contact_info",
        "data": {
          "tags": [
            "studio",
            "weekly"
          ],
          "handles": [],
          "entityId": "e34b644c-b34e-0d57-986a-b2c660d23265",
          "categories": [
            "friend"
          ],
          "preferences": {
            "timezone": "UTC"
          },
          "customFields": {
            "project": "CEDAR-b373b3a8-4be6-48b4-9bbf-c52575f29855",
            "displayName": "Cedar Audit Person"
          },
          "interactions": [],
          "lastModified": "2026-09-05T03:23:37.613Z",
          "privacyLevel": "private",
          "relationshipStatus": "active"
        },
        "createdAt": "2026-09-05T03:23:37.617Z"
      }
    ]
  },
  "component": {
    "id": "dd6a3d8b-9ddc-0e05-bb1d-5ee91db6b0e5",
    "entityId": "e34b644c-b34e-0d57-986a-b2c660d23265",
    "agentId": "53ecbc15-5a88-0733-8212-28c0e960b99a",
    "roomId": "1a4f1dc7-d0f8-04c8-852f-cad9a402a2a3",
    "worldId": "a39c737e-3e3b-0caa-93c9-547f9e119ac8",
    "sourceEntityId": "53ecbc15-5a88-0733-8212-28c0e960b99a",
    "type": "contact_info",
    "data": {
      "tags": [
        "studio",
        "weekly"
      ],
      "handles": [],
      "entityId": "e34b644c-b34e-0d57-986a-b2c660d23265",
      "categories": [
        "friend"
      ],
      "preferences": {
        "timezone": "UTC"
      },
      "customFields": {
        "project": "CEDAR-b373b3a8-4be6-48b4-9bbf-c52575f29855",
        "displayName": "Cedar Audit Person"
      },
      "interactions": [],
      "lastModified": "2026-09-05T03:23:37.613Z",
      "privacyLevel": "private",
      "relationshipStatus": "active"
    },
    "createdAt": 1788578617617
  }
}
```

Risk: transaction callbacks now have real atomicity and nesting semantics. External side effects inside callbacks still cannot be rolled back; the adapter-owned notification path waits for commit. No schema migration is introduced. Known gaps: live behavior was verified against a disposable local database, with synthetic contacts only. Hosted checks must finish successfully on the current head before merge. Maintainer CI exception: N/A.

Final review: both independent reviewers approved the current head. One reproduced the old delegate-only transaction implementation under the new tests and observed eight of nine cases fail. All five hosted checks passed in [run 33943823963](https://github.com/elizaOS/eliza/actions/runs/33943823963). Merged with a commit guard as `e6cd97a9e31401fc3d23e4d04786d22589aec529` at 2026-09-05T04:47:27Z.

An additional real AgentRuntime/PGlite probe removed the relationships service, invoked CONTACT with a tag, and read back the database: the result was `RELATIONSHIPS_UNAVAILABLE`, `success:false`, and `persistedEntity:null`. This supplementary probe ran in implementation worktree commit `e76928e11e68964e46956987361ec5e0d88ed3ea`; its CONTACT source was verified byte-for-byte equal to this PR head. The earlier full persistence and provider proofs above ran directly on the publication head. Adapter caches created within a transaction remain scoped to that transaction.

### DIFF
--- a/packages/agent/src/actions/contact.persistence.integration.test.ts
+++ b/packages/agent/src/actions/contact.persistence.integration.test.ts
@@ -1,0 +1,121 @@
+/**
+ * Exercises CONTACT creation through AgentRuntime, RelationshipsService and
+ * PGlite, including complete field persistence and a database write fault.
+ */
+import { type Memory, stringToUuid, type UUID } from "@elizaos/core";
+import type { RelationshipsService } from "@elizaos/core/services/relationships";
+import {
+  createTestRuntime,
+  type TestRuntimeResult,
+} from "@elizaos/core/testing";
+import { sql } from "drizzle-orm";
+import { afterAll, beforeAll, describe, expect, it } from "vitest";
+import { contactAction } from "./contact.ts";
+
+describe("CONTACT persisted creation", () => {
+  let harness: TestRuntimeResult;
+  let service: RelationshipsService;
+  beforeAll(async () => {
+    harness = await createTestRuntime({
+      characterName: "ContactPersistenceProof",
+    });
+    service = (await harness.runtime.getServiceLoadPromise(
+      "relationships",
+    )) as RelationshipsService;
+  }, 180_000);
+  afterAll(async () => {
+    await harness?.cleanup();
+  });
+
+  async function create(id: UUID, overrides: Record<string, unknown> = {}) {
+    const result = await contactAction.handler(
+      harness.runtime,
+      {
+        id: stringToUuid(`message-${id}`),
+        entityId: harness.runtime.agentId,
+        agentId: harness.runtime.agentId,
+        roomId: stringToUuid("contact-proof-room"),
+        content: { text: "Create synthetic audit contact", source: "test" },
+      } as Memory,
+      undefined,
+      {
+        parameters: {
+          action: "create",
+          entityId: id,
+          name: "Audit Person",
+          email: "audit@example.invalid",
+          categories: ["friend"],
+          tags: ["studio", "weekly"],
+          preferences: { timezone: "UTC" },
+          customFields: { project: "Cedar", source: "audit" },
+          ...overrides,
+        },
+      } as never,
+    );
+    if (!result) throw new Error("CONTACT returned no result");
+    return result;
+  }
+
+  it("persists every requested field and returns the stored contact receipt", async () => {
+    const id = stringToUuid("contact-persisted-complete");
+    const result = await create(id);
+    expect(result.success).toBe(true);
+    const entity = await harness.runtime.getEntityById(id);
+    expect(entity?.metadata).toMatchObject({ email: "audit@example.invalid" });
+    const component = await harness.runtime.getComponent(id, "contact_info");
+    expect(component?.data).toMatchObject({
+      categories: ["friend"],
+      tags: ["studio", "weekly"],
+      preferences: { timezone: "UTC" },
+      customFields: {
+        project: "Cedar",
+        source: "audit",
+        displayName: "Audit Person",
+      },
+    });
+    if (!component?.data)
+      throw new Error("Contact component was not persisted");
+    expect(result.data?.contact).toMatchObject(component.data);
+    expect((await service.getContact(id))?.tags).toEqual(["studio", "weekly"]);
+  });
+
+  it("rolls back a new entity when the real database rejects contact fields", async () => {
+    const id = stringToUuid("contact-persisted-rollback");
+    const db = harness.runtime.adapter.db;
+    if (!db || !("execute" in db) || typeof db.execute !== "function")
+      throw new Error("PGlite SQL unavailable");
+    await db.execute(
+      sql.raw(
+        `ALTER TABLE components ADD CONSTRAINT audit_contact_write_fault CHECK (entity_id <> '${id}'::uuid)`,
+      ),
+    );
+    try {
+      const result = await create(id);
+      expect(result.success).toBe(false);
+      expect(await harness.runtime.getEntityById(id)).toBeNull();
+      expect(await harness.runtime.getComponent(id, "contact_info")).toBeNull();
+      expect(await service.getContact(id)).toBeNull();
+    } finally {
+      await db.execute(
+        sql.raw(
+          "ALTER TABLE components DROP CONSTRAINT audit_contact_write_fault",
+        ),
+      );
+    }
+  });
+
+  it("does not replace existing contact fields when creation is repeated", async () => {
+    const id = stringToUuid("contact-persisted-repeat");
+    expect((await create(id)).success).toBe(true);
+    const before = await harness.runtime.getComponent(id, "contact_info");
+    const result = await create(id, {
+      tags: ["overwrite"],
+      customFields: { project: "Changed" },
+    });
+    expect(result.success).toBe(false);
+    expect(result.values).toMatchObject({ error: "CONTACT_ALREADY_EXISTS" });
+    expect(await harness.runtime.getComponent(id, "contact_info")).toEqual(
+      before,
+    );
+  });
+});
--- a/packages/agent/src/actions/contact.test.ts
+++ b/packages/agent/src/actions/contact.test.ts
@@ -215,9 +215,13 @@ describe("CONTACT exports and dispatch", () => {
 describe("CONTACT create", () => {
   it("creates an entity with sanitized metadata and promotes rich contact data", async () => {
     const entityId = stringToUuid("created-contact");
-    const addContact = vi.fn(async () => true);
+    const createContact = vi.fn(async (entity, fields) => ({
+      entity,
+      contact: { ...fields, entityId: entity.id },
+      created: true,
+    }));
     const { runtime, createEntity } = makeRuntime({
-      relationships: makeGraph({ addContact }),
+      relationships: makeGraph({ createContact }),
     });
 
     const result = await invoke(runtime, {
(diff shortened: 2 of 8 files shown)


ss251 (approved):
Reviewed head cd576e95cfe6d47ddb09744b01034121e6dff338.

The risky part is `plugins/plugin-sql/src/base.ts:6646` where `transaction()` builds a scoped facade over drizzle `db.transaction` and defers `publishCommittedWrite`, paired with `plugins/plugin-sql/src/pglite/adapter.ts:36` snapshotting each row via `structuredClone`. I checked the two paths that break this pattern: a savepoint abort must not leak `notifyWrite`, and caller mutation of `metadata` after `createEntities` must not poison write-back. The queue lives only on the outer `writes` array so throw discards it, and the clone at `notifyWrite` time freezes the persisted value. The new savepoint test asserts `notify` silent inside tx and exactly one `entities/insert` for the kept id.

No blocking concerns, non-blocking nit: `WRITE_BACK_ROW_UNAVAILABLE` now throws where lookup once only logged.
Verified by reading the diff at this head; no suite run applies here.
Recommendation: merge

— [muse-lane-ss251]

attentionhead (approved):
Approving at `cd576e95cfe6d47ddb09744b01034121e6dff338`.

A note on the checks first, since it looks worse than it is: the `All Tests Passed: FAILURE` on this PR belongs to run `33943729505`, which was cancelled and superseded. The live run `33943823963` has three green and `Source static smoke` still in progress.

## `transaction()` was not a transaction

This is the part worth dwelling on. On `develop`:

```ts
async transaction<T>(callback, _options): Promise<T> {
  // Delegate to the callback with this adapter as the transaction context.
  // True DB-level transactions are handled by drizzle's this.db.transaction() in individual methods.
  return callback(this as IDatabaseAdapter<DrizzleDatabase>);
}
```

No `BEGIN`, no rollback — the callback just receives the adapter. The comment is candid about it, but the consequence is that every caller which reached for `adapter.transaction()` for atomicity got none, silently. A partial failure left whatever had already been written.

The replacement opens a real transaction through `withEntityContext`, hands the callback a `Object.create(this)` facade with `db` bound to the transaction handle, routes nested `withEntityContext` calls through `db.transaction` for savepoints, and defers write-back notifications into a `transactionWrites` list published only after the outer commit.

I checked that the new suite is actually load-bearing rather than merely present, by restoring `develop`'s delegate-only body underneath it:

```
Tests  8 failed | 1 passed (9)

× rolls back a successful child when its parent fails
× isolates a caught child failure in a savepoint while committing its parent
× inherits nested entity context and rejects switching identities
× uses the transaction connection for participant, relationship and deleted-memory write-back IDs
× publishes the persisted metadata even if the caller mutates its input before commit
× reports committed publication failures and still publishes later writes
× publishes only committed writes after the outermost commit
```

Eight of nine die. The test names map one-to-one onto the claims in the description, including the two subtle ones — payloads copied before publication, and a post-commit publication failure reported as committed with an explicit "do not replay".

## The CONTACT half depends on the first half

`createContact` wraps the entity insert and the `contact_info` component in `runtime.transaction(...)`, which is exactly the call that did nothing before. So the "could leave an entity behind" fix is not independent of the adapter change — it would have been inert without it. Treating these as one PR is right.

On the dropped fields, I verified the specifics rather than taking the summary:

- **`tags` really were dropped, twice over.** `develop`'s `addContact` takes no `tags` parameter at all and hardcodes `tags: []` in the `ContactInfo` it builds, and the action never passed them.
- **`customFields` really were dropped.** The action called `addContact(entityId, addCategories, addPrefs, { displayName: name })`. The real signature's fourth parameter is `customFields`, so the user's custom fields were not merely ignored — they were replaced by `{ displayName }`. The action's local `RelationshipsServiceLike` interface declared that slot as `extra?: { displayName?: string }`, which is what let the mismatch typecheck.
- **`preferences` were not dropped.** They were passed as the third argument and stored. The description lists preferences alongside tags and custom fields; that part is an overstatement, harmless but worth correcting if it ends up in a changelog.

`createContact` now takes all four as an explicit `Pick<ContactInfo, ...>`, so the shape mismatch that caused this can't recur silently.

## Two things worth your attention, neither blocking

**A new uncovered failure path.** When the promoted branch is taken and the relationships service is missing, the action now returns `RELATIONSHIPS_UNAVAILABLE`. Previously that same situation created a plain entity, dropped the extra fields, and reported success. Failing loudly is the better behavior — silently discarding user-supplied tags is what this PR exists to stop — but it is a user-visible change, and `RELATIONSHIPS_UNAVAILABLE` appears only in `contact.ts` with no test exercising it. Everything else new here is covered; this one path isn't.

**Scoped-facade caveat.** `Object.create(this)` means writes to `this.<field>` inside a transaction land on the throwaway facade rather than the adapter. That is deliberate and mostly desirable, and `_connectorAccountStore` is explicitly reset for it — but any future adapter method that caches on `this` will silently lose that cache inside a transaction. Worth a comment at the facade if you agree.

## Verification

- `adapter-transactions.test.ts` — 9 pass at this head; 1 pass / 8 fail with `develop`'s `transaction()` restored
- `contact.test.ts` — 22 pass
- `contact.persistence.integration.test.ts` — 3 pass via `test:integration`. It is excluded from the default `packages/agent` lane by the `**/*.integration.test.*` exclude, which briefly looked like a coverage gap to me; it is not, the dedicated lane runs it.
- `pglite-adapter-liveness.test.ts` — 4 pass; `packages/core` relationships — 78 pass across 11 files
- typecheck 0 errors in `packages/core`, `packages/agent`, `plugins/plugin-sql`; lint clean; all after `bun run build:core`

## Not verified

The PostgreSQL-only cases in the adapter suite did not execute here — I ran against PGlite, so `keeps PostgreSQL entity context through nested savepoints` and the RLS `set_config` save/restore path are unexercised by my run, and that path is the most intricate part of the change. Your real-PostgreSQL evidence covers it; I could not independently confirm it. I also did not audit the other `adapter.transaction()` callers across the repo to see which were silently non-atomic before this fix.

— [claude-code-eliza]

lalalune:
Independent review of `cd576e95cfe6d47ddb09744b01034121e6dff338`, subsequently merged as `e6cd97a9e31401fc3d23e4d04786d22589aec529` by another task:

- Read all eight changed files, their callers, and the complete two-call Cerebras request/output/persisted-contact artifact.
- Independently passed 19 real PostgreSQL/PGlite transaction tests, 22 CONTACT unit tests, and three real AgentRuntime/PGlite contact-persistence integration tests.
- Reviewed nested transactions, savepoint rollback, RLS identity restoration, post-commit notification failure, duplicate identity handling, and contact metadata readback. No additional defect found.
- The independent full SQL suite passed 185 tests with three existing opt-in skips, but seven live-query cases exceeded their explicit 10-second test timeout during PGlite setup. A separate run of both affected files failed all nine cases with the same timeout on the clean pre-PR checkout `f67b9bc98ba858e8179e6d96d7e73c8adff11e5a`; the current host is heavily loaded. This is a reproduced baseline limitation, not a green full-suite result.

The final PR head had all five required hosted checks successful in run 33943823963. The develop run on the merge commit was cancelled by a newer develop run.

PR #30604, opened 2026-09-05
fix(types): align agent plugin exports and video adapters

lalalune:
Agent plugin imports need canonical subpath mappings, and video consumers need declarations matching the installed fetch and fluent-ffmpeg APIs. This change adds eleven agent path mappings, types the callable fetch/stream adapter, and adds the fluent-ffmpeg methods used by video consumers.

The remaining diff is three files of TypeScript configuration and type adaptations. The Android identity changes from the earlier revision have already reached develop.

Validation at `7770fd01caf2a05cac4b3d005860e44fc1f74166`, synchronized with develop `9bd31de1ac22e18b13acb39dad6d217ac435f9ac`:

- Normal `bun install` passed. The documented fused-inference setup skip preserved the native checkout.
- `RUN_TURBO_CONCURRENCY=1 bun run verify` passed all 373 uncached workspace tasks in 6m14.097s and every final repository audit. The source worktree remained clean.
- All three owned files are byte-identical to the independently reviewed predecessor. Focused validation on those same file contents passed 43 video tests and 10 agent tests, plus owning typechecks; four opt-in real-network tests were skipped.
- Independent source review checked all eleven mappings and the installed fluent-ffmpeg APIs with no blocking finding.
- Hosted checks on this newly published head remain a merge gate. The predecessor's green CI does not certify this head.

## Evidence Gate

N/A - compile-time contracts have no rendered UI surface.

N/A - compile-time contracts have no rendered UI surface.

N/A - compilation and source/API inspection demonstrate this type-only change.

[Normal install](https://github.com/elizaOS/eliza/releases/download/pr-evidence-10/30604-install-7770fd01.log) and [complete root verification](https://github.com/elizaOS/eliza/releases/download/pr-evidence-10/30604-root-7770fd01.log).

N/A - no frontend runtime behavior changes.

N/A - no model or prompt behavior changes.

[Exact-head validation and source-hash receipt](https://github.com/elizaOS/eliza/releases/download/pr-evidence-10/30604-7770fd01-receipt.json).


lumix5 (approved):
Independent review at exact head `9acd3e4fd2569559d4463a09a339b7968a898c8b`. Recommendation: **merge** — the premise is a real Android platform constraint, the boundary design preserves the descriptor-backed trust model where it applies, and every new failure mode fails closed with a typed error. Final merge decision rests with the maintainers.

- **The premise.** Android SELinux permits `stat` on platform-owned ancestors (`/`, `/data`, `/data/user`) but denies opening directory descriptors for them, so the existing validator — which opens every ancestor with `O_DIRECTORY|O_NOFOLLOW` for descriptor-backed revalidation — cannot create an installation identity on device. The fix adds an explicit `ELIZA_ANDROID_APP_DATA_DIR` boundary (bound in `ElizaAgentService.java` from `getDataDir()` with an accurate comment): ancestors strictly *above* the boundary are trusted by `lstat` + real-directory/not-symlink assertions + `sameIdentity` revalidation (the platform-owned set is immutable per SELinux); the boundary itself and everything below keep full descriptor-backed validation. A post-`realpath` containment re-check rejects any resolved escape from the app-data boundary.
- **Boundary validation fails closed**, pinned by tests I ran: a missing or relative `ELIZA_ANDROID_APP_DATA_DIR` throws (`requires an absolute …`); a state directory outside the boundary throws (`must remain inside …`); a `0o777` boundary is rejected by the existing replaceable-by-another-user mode check without creating the state directory (`ENOENT` asserted).
- **Red/green in a disposable container** (oven/bun:1.3.14, frozen-lockfile install, keywords, `@elizaos/prompts` + `@elizaos/core` built for workspace resolution): the two new android tests and the SELinux-open race test fail on develop's implementation and pass at head; at head the combined suites are 23 pass / 1 skip in this environment. Two *pre-existing* race tests ("cleans both directory inodes after each publication substitution phase", "reports ambiguous recovery only after post-link cleanup…") fail **identically on develop's implementation under both bun and node here** — they are container artifacts (the inode-substitution/rename tricks depend on filesystem semantics overlayfs does not preserve), not PR-caused: hosted CI is green at this head on all four PR lanes, and the author's local run reports the complete root verify at 376/376 workspace tasks.
- **The type-resolution half.** All 11 new `packages/agent/tsconfig.json` path mappings point at files that exist at this head (checked each). The affected-closure typecheck+build ran hosted in Source static smoke (27m44s, green), which is what exercises the mappings; the `plugin-video` changes are type-level only (a `FetchImpl` alias and a double-cast for `Readable.fromWeb`). The Java change is one environment binding, pinned by the `android-periodic-wake-reconciliation` source-contract test in the diff.

## Notes (non-blocking)

- PR-surface fact worth a maintainer's awareness: `test.yml` (the Server/Client/Plugins/… unit lanes) has no `pull_request` trigger — it runs only via develop-full — so this PR's new security tests' first **hosted** execution is post-merge; on the PR the suite evidence is the author's local verify plus this container proof. That is current repo policy, not a defect of this PR.
- `ELIZA_PLATFORM` is read through `readAliasedEnv` while `ELIZA_ANDROID_APP_DATA_DIR` is read from `process.env` directly — if the alias indirection is meant to cover all `ELIZA_*` inputs, the boundary variable should follow; if the path is deliberately machine-local, a one-line comment would lock that in.
- The two overlayfs-flaky race tests are worth knowing about for anyone else verifying in Docker default storage.

## Commands run

- Full diff read (runtime-installation-id + both test files, tsconfig, Java, reconciliation test, plugin-video, fluent-ffmpeg.d.ts); `git cat-file -e` existence check on all 11 mapping targets.
- Container: `bun install --frozen-lockfile --ignore-scripts`; keywords; builds of `@elizaos/prompts` and `@elizaos/core`; `vitest run` of both suites at head (bun and node) and with develop's implementation swapped in (red).
- Hosted: all four PR lanes green at head; log grep confirming where the suite does and does not run; `test.yml`/`develop-full.yml` caller analysis.

## Remaining checks

Maintainer judgment on the stat-trust window above the app-data boundary (sound per the SELinux contract, but it is the security-relevant delta), and the first post-merge develop-full run carrying these suites.


Compute receipt: 12134575 project-attributed tokens (exact; device-signed, locally reported)
— [eliza-review]

ss251 (approved):
9acd3e4fd2569559d4463a09a339b7968a898c8b

Android identity boundary fix is real and scoped correctly. `packages/agent/src/runtime/runtime-installation-id.ts` previously attempted `O_DIRECTORY|O_NOFOLLOW` opens on every ancestor, which SELinux denies above app-data on Android. New `resolveAndroidStateBoundary` + `isAndroidPlatformAncestor` keeps descriptor-backed validation from `ELIZA_ANDROID_APP_DATA_DIR` downward while lstat-only validating platform ancestors above it, and `openTrustedStateDirectory` enforces `isDescriptorParent(parent)` so the immediate parent can never be platform-only. Resolved-path escape check after `realpath` closes symlink escape out of the boundary.

`packages/app-core/platforms/android/app/src/main/java/ai/elizaos/app/ElizaAgentService.java` binds `ELIZA_ANDROID_APP_DATA_DIR=getDataDir()` alongside `ELIZA_PLATFORM=android`, matching validator input. Tests cover the claim: race test faults `beforeOpen` outside the boundary to prove no forbidden open, boundary test proves missing/escaped/0o777 rejects with no artifact left.

Secondary scope is type-only and additive: `packages/agent/tsconfig.json` adds ~10 path aliases, `plugins/plugin-video/src/services/binaries.ts` loosens `fetchImpl` to `FetchImpl` and `Readable.fromWeb` cast, `plugins/plugin-video/src/types/fluent-ffmpeg.d.ts` adds `toFormat/ffprobe`. No runtime behavior change there.

Residual: platform ancestors are TOCTOU lstat-only by necessity; accepted given SELinux constraint.

Verified by reading the diff at this head and tracing ancestor/open/revalidate paths and env plumbing; no suite applies to Java env wiring from this host.

Recommendation: merge

— [muse-lane-ss251]

byt61 (changes requested):
The core design is sound — I traced the validation state machine: strict ancestors above the boundary get lstat-only real-dir checks, the boundary dir itself keeps descriptor validation, the immediate parent is pinned with `isDescriptorParent`, and the post-realpath escape check closes symlink redirection. The non-Android path is behavior-identical (new branches only add explicit tags). The tsconfig/plugin-video/Java portions also read clean.

One blocking, verified issue: the 3 new Android tests fail on macOS. `os.tmpdir()` returns unresolved `/var/folders/...` and `/var` is a symlink to `/private/var`, so the strict-ancestor assertion rejects the fixture with `Android app-data ancestor must be a real directory` before any production logic runs. CI is green only because Linux `/tmp` has no symlinked ancestors. I reproduced all 3 failures locally via overlay (22 pass + 3 fail), then proved the fix: wrapping the fixture root in `fs.realpath` —

```ts
const root = await fs.realpath(
  await fs.mkdtemp(path.join(os.tmpdir(), "runtime-owner-android-")),
);
```

— takes the suites to 25 passed / 1 skipped (the skip is pre-existing). Please apply that to the two fixtures in `runtime-installation-id.test.ts` and the one in `runtime-installation-id.race.test.ts` so macOS checkouts stay green.

lalalune:
Independent review of `9acd3e4fd2569559d4463a09a339b7968a898c8b` found another Android startup incompatibility: `openTrustedLexicalChain` rejects the platform-owned `/data/user/0` symlink. Current [AOSP Zygote](https://android.googlesource.com/platform/frameworks/base/+/refs/heads/main/core/jni/com_android_internal_os_Zygote.cpp#1318) creates that alias in the app mount namespace. A real-filesystem regression recreating this layout fails before identity creation on the original PR.

The repair preserves trusted symlink owner/inode validation and compares the resolved state path with the resolved app-data boundary. The app-owned boundary itself remains a real directory. Independent tests now cover both direct and aliased Android layouts under simulated ancestor-open denial, aliased boundary rejection, escaping state redirects, and alias replacement before identity creation. All 29 executed identity/race tests and the native environment suite pass locally; the existing opt-in identity test stays skipped. The branch has been integrated onto develop `5c9ae7f173e2b099da143bf673efa43b56e9fa14`; final root verification is running before a SHA-leased update. Physical Android health/ASR acceptance remains pending, so this stays draft.

dal-mnr (changes requested):
@C:\Users\MUNIR-~1\AppData\Local\Temp\opencode\review-30604.md

NubsCarson (approved):
Reviewed exact head 7770fd01caf2a05cac4b3d005860e44fc1f74166 against develop 9bd31de1ac22e18b13acb39dad6d217ac435f9ac. No blocking source finding in the remaining three-file type-resolution change.

The eleven agent aliases resolve to canonical workspace exports, the fetch adapter remains callable with the existing stream conversion, and the fluent-ffmpeg declarations match the installed methods used by consumers. An independent reviewer confirmed all three files are byte-identical to the previously reviewed revision.

Normal install and the full root verification passed at this head: 373 uncached tasks plus every final audit, with a clean source tree. Attached exact-head logs and the receipt passed the evidence gate. Hosted CI 34072199756 is still running and must pass before merge. This approval covers the current compile-time change; earlier Android behavior has already landed in develop.

PR #30605, opened 2026-09-05
fix(shared): consolidate domain SQL and time helpers

### LINKED ISSUE #30582: fix(shared): consolidate SQL helpers without hiding invalid results
The duplicated raw-SQL adapters in scheduling, personal-assistant, finances, and finance migration accept malformed database results as empty rows and silently drop malformed row entries. This makes unavailable storage resemble an empty scheduled-item or finance collection. The finance transaction helper also executes multi-record operations inline when the adapter lacks transaction support.

Acceptance:
- Consolidate the shared result parsing/query primitives below domain plugins while retaining their existing import paths and runtime-resolution policy.
- Accept complete row arrays and the supported database result envelopes; preserve valid empty SELECT and DDL results. Reject malformed envelopes or any malformed row instead of returning partial/empty data.
- Reject missing finance transaction support before invoking a mutation callback.
- Exercise real database queries, statements, rollback, and consumer-visible invalid-result failures.

Related consolidation tracking: #30125. The identical personal-assistant/health time helpers will also move into the existing shared normalization module without changing their behavior.

lalalune:
Personal-assistant, scheduling and finance carried separate raw SQL adapters; personal-assistant and health also carried identical time helpers. They now use canonical shared helpers while retaining the existing package import paths. Raw SQL execution validates every returned row, including sparse entries, and rejects malformed results instead of silently turning them into empty or partial data. The finance migration path requires real transaction support before invoking a transaction callback.

Closes #30582. Related consolidation work: #30125.

The shared SQL module remains runtime-neutral and uses Drizzle execution while retaining the domain SQL value encoders. Existing domain wrappers retain their public API. No schema or data migration is introduced. Time helper behavior is unchanged.

Validation: shared 2,847, personal-assistant 2,492, scheduling 638, health 199 and finance 198 tests passed, with existing opt-in skips documented in their logs. Real PGlite contract tests exercise SQL value encoding, result validation and rollback through the finance consumer of the canonical executor. Existing personal-assistant and scheduling tests cover their compatibility wrappers. On the publication branch, the 69-task build, all 198 finance tests, 17 scheduling persistence cases, 27 personal-assistant recalibration cases, 28 real approval-queue integration cases, 199 health tests and 22 shared normalization cases passed. The health lane retains its four existing skips. Root verification passed all 376 tasks and every repository audit; the test-integrity scan found no focused tests or untracked skips across 11,373 test files.

## Evidence Gate

N/A - shared storage and time helpers; no rendered UI changed.

N/A - shared storage and time helpers; no rendered UI changed.

N/A - observable SQL behavior is recorded in the test transcript below.

```text
Command: ['bun', 'run', '--cwd', 'plugins/plugin-finances', 'test', '--maxWorkers=1']
$ bunx vitest run --config ./vitest.config.ts "--maxWorkers=1"

 RUN  v4.1.10 /Users/shawwalters/.codex/worktrees/code-quality-cleanup-20260904/plugins/plugin-finances


 Test Files  15 passed (15)
      Tests  198 passed (198)
   Start at  00:22:17
   Duration  211.99s (transform 52.08s, setup 1.36s, import 120.31s, tests 81.77s, environment 2.44s)


Command exit: 0
```

N/A - no frontend request behavior changed.

N/A - no prompt, model inference or provider output construction changed.

Actual PGlite readback through the finance consumer; the invalid-result case substitutes only the adapter result envelope.

```json
{
  "head": "e5223fe8ea5560144ebd4c27e5ac53373406742b",
  "saved": [
    {
      "id": 1,
      "label": "O'Reilly studio",
      "details": {
        "amount": 0,
        "marker": "532fa46f-b94f-4021-9835-4b3249397aff",
        "enabled": false
      }
    }
  ],
  "rollbackError": "Failed query: INSERT INTO proof_rows VALUES (1, 'duplicate', '{}')\nparams: ",
  "afterRollback": [
    {
      "id": 1,
      "label": "O'Reilly studio",
      "details": {
        "amount": 0,
        "marker": "532fa46f-b94f-4021-9835-4b3249397aff",
        "enabled": false
      }
    }
  ],
  "missingTransaction": {
    "callbackRan": false,
    "code": "FINANCES_TRANSACTION_REQUIRED"
  },
  "invalidResult": {
    "injectedBoundary": "adapter result envelope",
    "code": "SQL_RESULT_INVALID"
  },
  "passed": true
}
```

Risk: a malformed database adapter result now fails explicitly. Public compatibility wrappers remain available, and valid array/rows results retain their existing values. Hosted checks must pass on the current head before merge. Maintainer CI exception: N/A.


ss251 (approved):
Reviewed head e5223fe8ea5560144ebd4c27e5ac53373406742b.

The risky hunk is `packages/shared/src/db/raw-sql.ts:115` `extractRows`, where consolidation tightens semantics: the old finance/PA/scheduling copies mapped rows through `asObject` + `filter(row !== null)` and returned `[]` on malformed envelopes, while the shared version throws `RawSqlError(SQL_RESULT_INVALID)` with `rowIndex` on the first invalid row. I checked the two boundaries that break this kind of change: `plugins/plugin-finances/src/db/sql-contract.test.ts:87` explicitly asserts `null`, `{}`, `{rows:null}`, `{rows:[null]}`, `[{id:1},false]` all reject with that code instead of returning an empty finance list, and the array-envelope branch still accepts DDL/empty results (`[]`) plus raw row-arrays without dropping values. Migration's local silent `extractRows` is deleted in favor of this stricter path.

Verified by reading the diff at this head and tracing extractRows callers in finances-repository and migration paths; no live DB run.
Recommendation: merge

— [muse-lane-ss251]

attentionhead (approved):
Approving at `e5223fe8ea5560144ebd4c27e5ac53373406742b`.

Three separable claims here — a byte-identical time-helper move, a deliberate strictness change in row extraction, and a fail-closed transaction guard. I checked each rather than treating this as one consolidation.

## "Time helper behavior is unchanged" — verified literally

`plugin-health/src/util/time-util.ts` and `plugin-personal-assistant/src/lifeops/time-util.ts` were byte-identical to each other on `develop`, and diffing their exported bodies against the new `packages/shared/src/lifeops-normalize/time-util.ts` reports no difference — only the header comment changed. Both plugin files are now four-line re-export shims, so the existing import paths still resolve. This is the one part of the PR where "unchanged" can be taken at face value, and it is.

## Row extraction: a real behavior change, well guarded

`develop`'s `extractRows` silently discarded anything it did not like:

```ts
if (!object) return [];
if (!Array.isArray(rows)) return [];
return rows.map(asObject).filter((row) => row !== null);
```

A malformed envelope became an empty list and a sparse or non-object row simply vanished from the result — a partial read indistinguishable from a genuinely short one. The shared version throws `SQL_RESULT_INVALID` with the offending `rowIndex`. That is the right call for a persistence boundary, and it is properly tested: reverting just `extractRows` to the lenient form fails 6 of the 10 new contract cases.

```
× rejects malformed result 3 instead of an empty or partial finance list
× rejects malformed result 4 …
× rejects malformed result 5 …
Tests  6 failed | 4 passed (10)
```

The four that still pass are the encoder cases, so the suite discriminates rather than failing wholesale.

I also checked the encoders for drift, since that is where a consolidation usually goes wrong. `sqlQuote` is byte-identical to `develop`'s (`'${value.replace(/'/g, "''")}'`), as are `sqlText`/`sqlInteger`/`sqlNumber`/`sqlBoolean`/`sqlJson` in behavior — the escaping strategy and its `sql.raw` execution path are carried over unchanged, not newly introduced by this PR.

## The transaction guard closes a silent-atomicity hole

`develop`'s `withTransaction` ended with:

```
// Adapter does not support transactions (likely a test fake). Run inline.
```

so a multi-record finance mutation against an adapter without `.transaction` ran with no atomicity at all and could partially commit payment source identities, transactions and aggregate counts. It now throws `FINANCES_TRANSACTION_REQUIRED` with `severity: "fatal"`. This is the same defect class as the adapter `transaction()` that was `return callback(this)` in #30603 — a fallback that quietly degrades a durability guarantee — and failing closed is right. Worth knowing that any consumer relying on a non-transactional fake will now throw rather than silently proceed; that is the point, but it is a behavior change.

A fourth duplicate `extractRows` inside `plugin-finances/src/services/migration.ts` is removed too, which is easy to miss when consolidating.

## Context for the follow-up, not a blocker

The same duplicated pair is still defined in roughly fifteen other places. I spot-checked three and they all carry the **lenient** version this PR just replaced:

- `plugins/plugin-goals/src/db/sql.ts:88`
- `plugins/plugin-inbox/src/db/sql.ts:85`
- `plugins/plugin-calendar/src/internal/sql.ts:92`

all with the identical `if (!Array.isArray(rows)) return []` and `.filter(row => row !== null)`. So the silent partial-read this PR fixes for finance, personal-assistant and scheduling is still live in those domain stores, plus copies in `plugin-reminders`, `packages/agent/src/services/{approval,knowledge-graph,pendant-session}` and others. The description scopes this to three domains and points at #30125 as related work, so this is not a gap in the PR — but it does mean the fix is a third of the way through the defect class, and those three files are the concrete next targets.

## Verification

Every count in the description that I could reproduce matched exactly:

- `plugin-finances` — 198 pass (15 files); new `sql-contract.test.ts` — 10 pass
- `plugin-scheduling` — 638 pass; `plugin-health` — 199 pass, 4 skipped
- `plugin-personal-assistant` — 2492 pass, 6 skipped (290 files)
- `packages/shared` — 2844 pass, 0 failed
- typecheck 0 errors in `packages/shared`, `plugin-finances`, `plugin-scheduling`, `plugin-health`, after `bun run build:core`

Two results that look like problems and are not, recorded so nobody re-derives them:

- `plugin-personal-assistant` typecheck reports **173 errors — identical on `develop`**. They are cross-plugin `TS2307` module-resolution failures needing sibling plugin dists; pre-existing, not this PR.
- `packages/shared` shows 1 failed test *file*, `src/terminal/theme.test.ts`, which cannot load because `esbuild` is absent from my environment. The PR does not touch that file, and no test in it failed on assertions.

## Not verified

I did not exercise the real-PGlite rollback path the description cites; my mutation testing covered row-validation strictness rather than transaction rollback against a live database. I also did not audit every consumer of the three rewritten domain wrappers for reliance on the previously lenient row filtering — the three plugin suites passing is good evidence, but it is not the same as an exhaustive call-site review.

— [claude-code-eliza]

PR #30607, opened 2026-09-05
feat(cloud): add scoped Outreachr identity, Google, and billing integration

### LINKED ISSUE #30598: feat: scoped Eliza Cloud integration for Outreachr
Outreachr is adding a hosted CRM using Eliza login, managed Gmail, inference, and Stripe. Existing app authorization exchanges only identity; it does not provide narrow, revocable access to the signed-in user's managed Gmail or product checkout.

Add an Outreachr registration-bound delegation exchange and scoped connector/billing routes that reuse the existing Cloud services. Keep existing Cloud subscriptions and user organization membership unchanged. Gmail sending must preserve an application-supplied RFC Message-ID and return the provider receipt for reconciliation.

Acceptance: wrong-app and replayed authorization codes fail; grants expire and revoke; primary user and mailbox ownership are checked on every operation; no Google secrets leave Cloud; invalid or ambiguous provider results are not reported as delivered; product checkout uses exact separate USD monthly prices ($49 Sol, $200 Astra), existing Stripe customer authority, and stable idempotency keys. Include behavioral tests and live integration evidence before deployment is declared complete.

lalalune:
# Relates to

Closes #30598. Adds the Eliza Cloud boundary required by Outreachr's hosted CRM.

- [x] Targets `develop`; rebased onto `7bb7938d1e397` without conflicts.
- [x] Post-sync `bun install --frozen-lockfile` passed; `bun run verify` passed.
- [x] Reproducible behavior and validation limits are recorded below.

# Sync with develop

- [x] Rebased onto the latest fetched `origin/develop`; zero conflicts.
- [x] Final post-sync verification passed at `8384d56f5c41d45f91eae3c1f08d79c96e981be7`.

# Risks

Medium: introduces a credential-bearing API boundary and a separate product's Stripe operations. Routes require a registered app and hashed client secret. User delegation is short-lived, revocable, and checked against primary account authority; Google ownership remains enforced by the managed connector service. Billing validates app/workspace/customer/price metadata and verifies Stripe webhook signatures. All new routes remain unavailable until explicit Outreachr bindings are configured.

# Background

## What does this PR do?

Outreachr can exchange its existing app-login code for scoped identity and managed Google access without copying Google credentials into the CRM. The primary database retains an authorization-code digest so replay stays blocked even after grant revocation or a stale cache read. Google transport permits only the Gmail/calendar operations used by the CRM.

The billing boundary reuses Cloud's Stripe service for separate Outreachr workspace customers and monthly seat prices. Stable idempotency keys protect customer and checkout creation; current subscriptions are read from Stripe and webhooks are signature-verified. Existing Cloud products and subscriptions are not changed.

Feature; additive backend routes and an append-only delegation migration.

Deployment bindings and migration order are documented in the companion Outreachr cloud README and implementation plan. Apply `0376_outreachr_delegations.sql` before enabling the routes; configure the exact app ID/origin/client-secret digest plus Sol/Astra price IDs and a dedicated webhook secret. Use the standard staging certification and production approval workflow.

# Testing

`outreachr-delegation.test.ts`, `outreachr-billing.test.ts`, and `outreachr-delegations-migration.pglite.test.ts` under `packages/cloud/shared/src`.

## Detailed testing steps

- Run the three focused tests with Bun. Eleven tests passed at the final rebased head: app/client binding, verified active identity, expired/revoked grants, replay rejection, constrained Google transport, migration replay, Stripe customer/price ownership, exact seat pricing, and real Stripe SDK webhook signatures/tampering.
- `bun run verify` passed at the final rebased head: all 376 typecheck/lint tasks, Cloud API router/Worker bundle checks, and final repository audits. The final audit scanned 11,379 tests with zero focused tests and zero orphaned skips. Hosted run 33949880548 passed every check, including All Tests Passed.
- Cloud API typecheck includes router-contract checks and the production Worker dry run. The dry run bundled the new routes successfully.
- Companion Outreachr tests passed through real PostgreSQL and the real CRM command/connector/vault path using controlled provider responses, including approval, provider receipt persistence and duplicate-send rejection.

# Evidence Gate

- [x] N/A - this diff has no rendered UI surface.

- [x] N/A - this diff has no rendered UI surface.

- [x] N/A - backend-only integration boundary.

- [ ] Focused backend test results are below; live route readback remains pending deployment.

- [x] N/A - this diff does not change frontend code.

- [x] N/A - this diff does not implement or change inference, models, prompts, or agent actions.

- [x] PGlite migration/replay test and real Stripe SDK signature tests exercise the changed artifacts.

# Evidence Details

## Real LLM-call trajectory

N/A - no inference or agent behavior changes in this companion backend PR.

## Backend + frontend logs

Focused test results at head `8384d56f5c41d45f91eae3c1f08d79c96e981be7` (focused tests and full verification completed successfully at this head):

```text
Delegation and PGlite migration: 8 pass, 0 fail
Stripe SDK billing boundary: 3 pass, 0 fail
Cloud API TypeScript: exit 0
Worker dry run: exit 0
Root typecheck/lint: 376 successful, 376 total
```

Frontend logs: N/A - no frontend changes in this PR. Live backend acceptance remains pending.

### Before
N/A - no rendered UI changes.
### After
N/A - no rendered UI changes.
### Walkthrough video
N/A - backend-only boundary.

## Audio / voice walkthrough

## Known gaps / failures

Post-sync verification, hosted checks, registration and staging/live acceptance are pending. The routes remain disabled until explicitly configured. Shaw's real Google login reached Cloud's join page, but dedicated-agent provisioning reported insufficient credits. No production app delegation, Stripe checkout, charge, or real email has been claimed as verified. Initial local failures were corrected: the generated migration snapshot needed its table entry/formatting, and the Stripe test signing helper needed its asynchronous variant.

N/A - no exception requested.


ss251 (approved):
Reviewed head 3fb54e5eadd8b5861c8a165ac64b37afbe272d2a.

The risky part is `packages/cloud/api/src/middleware/auth.ts:31`, which adds `/api/v1/outreachr` to `publicPathPrefixes`, opting the entire new shard out of global auth. I traced every handler in `packages/cloud/api/v1/outreachr/route.ts` to see if any rides that bypass open: `POST /token` goes through `exchange()` with `X-Outreachr-Client` plus an `eac_`-prefixed code, `GET /identity`, `GET /google/connections` and `POST /google/request` each call `authorize()` with the client secret plus `Authorization: Bearer`, and `POST /billing` calls `requireClient()`. No handler returns without one of those three registration-bound gates, so the public prefix delegates rather than exposes.

Non-blocking: `/billing` passes `?? ""` for the Stripe price IDs, so a missing env surfaces deeper instead of failing fast.

Verified by reading the diff at this head and following all six handlers to their auth calls.
Recommendation: merge

— [muse-lane-ss251]

attentionhead (approved):
Approving at `3fb54e5eadd8b5861c8a165ac64b37afbe272d2a`.

On the checks first: `All Tests Passed: FAILURE` belongs to run `33947480368`, whose `Source static smoke` was cancelled. The live run `33947851895` is three-for-three green with smoke still in progress.

Of the +25,398 lines, 24,115 are the generated `0359_snapshot.json`, so the reviewable surface is about 1,280 lines. I spent it on the security boundary, since the change that matters most here is one line of middleware.

## The middleware exemption is the load-bearing change

`/api/v1/outreachr` is added to `publicPathPrefixes`, so the entire prefix bypasses the platform session resolver and the route becomes solely responsible for its own authentication. I checked that claim rather than trusting the comment, and it holds:

- **Every handler authenticates.** `/token` and `/billing` call `requireClient`; `/revoke`, `/identity`, `/google/connections` and `/google/request` call `authorize`, which itself begins with `requireClient`. There is no handler reachable without the registered BFF secret.
- **The prefix match is boundary-aware.** `pathname === p || pathname.startsWith(\`${p}/\`)`, so a future sibling like `/api/v1/outreachr-admin` would not silently inherit the exemption. This was the specific trap I went looking for.
- **Configuration fails closed.** `outreachrRegistration` requires a UUID app id, a 64-hex secret digest, and an exact HTTPS origin with no embedded credentials, throwing 503 otherwise — a missing environment variable cannot degrade into an open endpoint.
- **The secret comparison is timing-safe** (`timingSafeEqual` over the digests, with a length floor).

## The delegation model

The grant record binds `appId`, `registrationDigest`, `userId`, `organizationId` and an expiry, and `authorize` re-checks all of them plus the principal on every request. Two consequences worth naming because they are easy to omit and are present here: rotating the client secret invalidates every outstanding grant, and a user who loses email verification, is deactivated, or changes organization stops being authorized immediately rather than at token expiry.

Tokens are 256 bits from `randomBytes`, stored only as `sha256` digests, and the format regex matches the 43-char base64url output exactly.

**The replay fence is real, not just asserted.** The adapter comment claims a unique code hash prevents two exchanges from minting credentials, and `saveGrant` relies on `onConflictDoNothing().returning()` returning zero rows. That only works if the uniqueness is actually declared, so I checked both sides: `.unique()` on `authorization_code_hash` in the Drizzle schema, and `CONSTRAINT "outreachr_delegations_authorization_code_hash_unique" UNIQUE(...)` in `0376_outreachr_delegations.sql`. Had that constraint been missing, `onConflictDoNothing` would only have collided on the primary key — always unique for a fresh random token — and the defense would have been inert.

## Delegated Google transport

This is the other place a mistake would be costly, since the caller supplies a URL that then gets a user's Google token attached. The allowlist compares `url.origin` for exact equality rather than testing the hostname for a substring, rejects embedded credentials and fragments, anchors every path pattern, blocks encoded separators, and strips no-op auth overrides via `access_token`/`oauth_token`/`key`. Crucially the route forwards `operation.url` — the validated value — rather than the raw `input.url`.

There is no IDOR on `connectionId`: `getGoogleAccessToken` resolves the grant only from `getScopedGoogleConnections({organizationId, userId, side})`, and both ids come from `authorize()` rather than the request body, so a connection belonging to another org simply is not in the candidate list and 404s.

Billing is fenced separately — Stripe objects must carry both `outreachr_app_id` matching the registration and a matching `outreachr_workspace_id`, and the webhook path uses `constructEventAsync` with a configured secret or refuses. `/billing` deliberately authenticates with the client secret alone and takes `workspaceId` on trust, which is correct here because `workspaceId` is Outreachr's own identifier and never resolves against an Eliza resource.

## One non-blocking defect

`deleteGrant` in the adapter returns `true` unconditionally:

```ts
await dbWrite.update(outreachrDelegations).set({ revoked_at: new Date() }).where(...);
return true;
```

The service reads that boolean and raises `OUTREACHR_REVOCATION_UNCONFIRMED` (503) when it is false — but false is unreachable, so that guard is dead code and `/revoke` answers `{ success: true }` for a token that was never stored or was already revoked. The write itself is fine; what is lost is the confirmation the service is explicitly trying to make. Adding `.returning({ tokenHash: ... })` and returning `rows.length === 1` would make the existing error path mean what it says. Not worth holding the PR for.

## Verification

- `outreachr-delegation.test.ts`, `outreachr-billing.test.ts`, `outreachr-delegations-migration.pglite.test.ts` — 11 pass across 3 files
- `packages/cloud/shared` and `packages/cloud/api` typecheck 0 errors, after `bun run build:core`
- Route registration is consistent: 708 → 709 routes, 127 → 128 shards, with the new shard key present

## Not verified

I did not exercise the live flow end to end — no real authorization-code exchange, Google call, or Stripe webhook — so my confidence is in the code paths and the schema, not in an observed request. I read the generated `0359_snapshot.json` only as a header; I did not diff 24,115 lines of it against the schema, so I am trusting codegen there. I also did not review `consumeAppAuthCode`, which is pre-existing and carries the single-use guarantee this flow depends on.

— [claude-code-eliza]

lalalune:
Independent review found a billing defect at `8384d56f5c41d45f91eae3c1f08d79c96e981be7` (the billing/delegation source is byte-identical to the preceding reviewed head).

`priceFor` accepts a $49 monthly price with `transform_quantity: { divide_by: 10, round: "up" }`. Through the real Stripe SDK with controlled provider responses, requesting three seats reached `checkout.sessions.create` instead of being rejected. Stripe applies that transformation before computing the billed amount, so checking only `unit_amount` does not enforce the advertised per-seat price. The subscription-read validator has the same gap. Reference: [Stripe Price object](https://docs.stripe.com/api/prices/object).

I am preparing a focused repair and regression cases in an isolated review worktree: require untransformed licensed per-unit prices at each relevant boundary, and reject invalid quantities rather than returning an ambiguous seat count. No Stripe account was contacted and no checkout or charge was created during this reproduction. Please keep this PR unmerged while the repair and exact-head checks finish.

lalalune:
The PR merged at `8384d56f5c41d45f91eae3c1f08d79c96e981be7` before the pricing repair described in my review comment was included. I am moving the completed two-file repair into a follow-up PR on the merged revision. It passed 17 focused tests and all 376 root tasks plus audits at `20e5eed7291c45cab36d04eb6eff9ef80b883092`; the new publication head will be recorded separately. Live route/provider acceptance remains unverified.

PR #30609, opened 2026-09-05
fix(mobile): preserve runtime payloads and resolve partial builds

### LINKED ISSUE #30606: fix(android): preserve provenance-bound JNI runtime bytes during release packaging
## Contribution provenance

- AI assistance: yes
- AI provider/model: OpenAI / gpt-6-astra
- Client / agent tooling: Codex desktop
- Contribution skill revision: elizaOS/eliza@9dc84f447e6511d0499ebec7e253f2ccff937ff7:packages/skills/skills/contribute-to-eliza

## Bug

The Android system APK packages different bytes for `lib/arm64-v8a/libeliza_bun.so` than the runtime provenance generated before Gradle assembly. The fail-closed OS bundle collector correctly rejects the candidate after the full AOSP build.

Reproduced on Linux with app revision `9dc84f447e6511d0499ebec7e253f2ccff937ff7` and OS revision `804e3a1d6a08d39f4114702bc1c43eee79b837fc`:

- Staged JNI Bun SHA-256: `6b7ff62dd7a3b2c8281256952fde278fa51dc545b3de2182ebf74326b1608b02`.
- Unsigned Gradle APK, OS-staged APK, and platform-signed APK Bun SHA-256: `8c7c46fa103d7d868093a0de0e95b095f658c585a32ec3ae72b21124e65445e6`.
- Collector error: `privileged APK runtime file does not match provenance: lib/arm64-v8a/libeliza_bun.so`.

This establishes that the mutation precedes AOSP signing. The current Gradle JNI packaging policy preserves debug symbols for a camera dependency but not the provenance-bound executable/runtime payloads. Trace the stripping intermediates, prevent runtime-byte mutation at the packaging boundary, and retain the existing provenance rejection.

## Acceptance

- Preserve exact bytes for the staged, provenance-bound JNI runtime payloads through Gradle release assembly and AOSP signing.
- Apply the policy to the maintained Android project and generated mobile packaging configuration, idempotently and without removing unrelated packaging settings.
- Add regression tests for existing and generated packaging blocks.
- Verify a real rebuilt APK against every recorded runtime size and hash, including deliberate mismatch rejection; then rerun the unchanged OS bundle collector.
- Do not alter provenance records to conceal packaging drift, weaken verification, or promote hardware eligibility.

## Evidence

- Generated artifacts and logs: exact before/after hashes and collector failure above; full build logs retained privately on the build host.
- Screenshots / MP4: N/A - packaging-only failure; the failed candidate was not flashed.
- Frontend/backend logs: N/A - no application behavior was executed.
- Live-model trajectory: N/A - no model behavior changed.
- Physical boot and runtime acceptance: pending a verified image, not claimed by a successful compilation.

— [android-runtime-packaging]

lalalune:
Fixes #30606.

Android packaging now preserves provenance-bound runtime libraries instead of stripping their payloads. Mobile builds resolve the actual platform export before accepting a workspace as built, including browser root/subpath exports when only partial output exists. Both native targets stage the SQL trigram archive and the complete bundled skills corpus.

Installed-target review found additional startup failures. Android now provisions the installation UUID with a cross-process file lock and atomic move before JavaScript validates it; the app sandbox denies the hard links previously required for first boot. Its watchdog uses the existing authenticated request builder. iOS receives a canonical Foundation Application Support boundary before either engine starts, retaining ownership, permissions, containment, symlink, and inode validation within that boundary. Android remains compatible with minSdk 26.

Owning tests run under their correct Node/Bun runners. Existing app-core fixtures now use explicit configured models, the current embedding artifact, and an isolated login home. A metadata-only desktop-window test was removed under the repository test-quality policy.

## Source and validation

Published candidate: `cf4f03f95e4f64af06006ab2c2f41a2df9af2753`, rebased onto develop `95fbe00bf690baab9b459c7c3eced902f6e6bd19`. Pinned Bun 1.3.14 and Node 24.15.0.

- Final normal installation and `bun run verify` passed: **373 tasks**, 3m58.847s, with final audits clean. The documented fused-inference setup escape was used; the pinned llama.cpp submodule was initialized separately. Host Llama inference was not exercised.
- Core/provider integration replay on predecessor `1a7b907e` (those source trees are identical in this candidate): **99 core tool/extractor/planner/pipeline cases**, **46 provider wire/fast-failure cases**, **9 real mobile Bun build/execute cases**, and **31 identity filesystem cases** passed; one platform-specific identity case skipped.
- Current Settings reconciliation: **43 tests across four files** and Biome passed. The previously reviewed fixture repair removes obsolete header metadata assertions while retaining real section navigation, deep links, and error-boundary recovery.
- Full agent sweep on the frozen pre-rebase repair tree `6fb29072`: **773/773 files passed**, 1714.8s. Full app-core Vitest on the preceding fixture repair tree passed **364 files / 4,561 tests**, with 1 file / 26 tests skipped. Its initial script phase required the missing pinned native submodule; after initializing that prerequisite, the final owning script suites passed **99 Node and 21 Bun tests**.
- Android identity verification includes **3 real Java filesystem tests** and **16 independent process contenders** returning one stable identity.

The [current source reconciliation](https://github.com/elizaOS/eliza/releases/download/pr-evidence-4/30609-cf4-source-reconciliation.json) records **11 unchanged runtime/native/core/provider trees** relative to `1a7b907e`. The intervening Settings production change and its fixture were reconciled above. Both current native artifacts were rebuilt, installed, and rechecked: **agent bundle bytes remain identical on both platforms**, all 21 Android DEX files and all 38 iOS native binaries are unchanged, and the current renderer files match their installed build. Only the generated plugin manifest changes within the runtime payload inventory. The earlier full agent sweep and watchdog observation are retained as predecessor evidence; they were not silently rerun or relabeled. [Current validation/build/runtime logs](https://github.com/elizaOS/eliza/releases/download/pr-evidence-4/30609-cf4-validation-logs.txt.gz) · [Earlier core/provider reconciliation and replay](https://github.com/elizaOS/eliza/releases/download/pr-evidence-4/30609-1a7-source-reconciliation.json) · [Earlier integration logs](https://github.com/elizaOS/eliza/releases/download/pr-evidence-4/30609-1a7-validation-logs.txt.gz) · [Full agent sweep receipt](https://github.com/elizaOS/eliza/releases/download/pr-evidence-4/30609-1a7-agent-full-predecessor.log).

## Installed native evidence

All runs used separate QA identity `ai.elizaos.qa.pr30609` on an owned Android 35 arm64 emulator and iOS 26.4 arm64 simulator. Existing applications and user state were not modified.

**Android:** the installed APK and extracted agent bundle match the final build. All **101 inventoried payload entries** verify: 98 retain exact bytes; three gzip archives become complete, matching tar bytes through aapt2. Authenticated native health, status, and skills return 200, with 35 bundled skills exposed. The installation UUID survives restart and the final upgrade. The watchdog was observed over two normal 600-second intervals plus the probe timeout without a spurious restart; all 21 final DEX files are identical to that observed implementation, and the refreshed backend passes its authenticated health probe.

[Final Android payload hashes](https://github.com/elizaOS/eliza/releases/download/pr-evidence-4/30609-cf4-android-payload.json) · [Installed/extracted API and identity receipt](https://github.com/elizaOS/eliza/releases/download/pr-evidence-4/30609-cf4-android-acceptance.json) · [Watchdog observation](https://github.com/elizaOS/eliza/releases/download/pr-evidence-4/30609-1a7-watchdog-predecessor.json).

**iOS:** all **78 installed agent files and 38 native binaries** match the final build. Native health reports runtime/database ready. A synthetic transcript was created through the real SQL-backed memory adapter (201), read (200), retained with identical complete JSON across native process termination/relaunch, and deleted with subsequent 404 verification. The installation UUID survives both upgrade and restart.

[Final iOS payload hashes](https://github.com/elizaOS/eliza/releases/download/pr-evidence-4/30609-cf4-ios-payload.json) · [Native API receipt](https://github.com/elizaOS/eliza/releases/download/pr-evidence-4/30609-cf4-ios-api.json) · [Transcript before restart](https://github.com/elizaOS/eliza/releases/download/pr-evidence-4/30609-cf4-ios-transcript-before.json) · [After restart and cleanup](https://github.com/elizaOS/eliza/releases/download/pr-evidence-4/30609-cf4-ios-transcript-after.json) · [Identity before](https://github.com/elizaOS/eliza/releases/download/pr-evidence-4/30609-cf4-ios-identity-before.json) / [after](https://github.com/elizaOS/eliza/releases/download/pr-evidence-4/30609-cf4-ios-identity-after.json).

### Reviewed current captures

These show the actual QA boundaries: Android Cloud sign-in and iOS notification/model-unavailable states. Native startup and persistence evidence is recorded separately above.

Android QA application

iOS QA application

[Android MP4](https://github.com/elizaOS/eliza/releases/download/pr-evidence-4/30609-cf4-android.mp4) · [iOS MP4](https://github.com/elizaOS/eliza/releases/download/pr-evidence-4/30609-cf4-ios.mp4) · [Artifact inventory and hashes](https://github.com/elizaOS/eliza/releases/download/pr-evidence-4/30609-cf4-evidence-inventory.json).

## Explicit acceptance boundaries

- No text model was configured in these QA apps. Live-model trajectories are N/A for this packaging/identity change; live inference, complete chat, notifications, coordinator behavior, hardware devices, App Store/no-JIT execution, and local Llama inference are not established.
- iOS skills are verified as staged/installed payloads. Its existing native `/api/skills` route returns 404; only Android's skills route was accepted. Existing iOS conversation routes use an in-memory map, so the persistence proof uses the real transcript adapter instead of conversation CRUD.
- Android local Notes retains `NOTES_STORE_WRITE_FAILED`: the unchanged develop `NotesStore` requires `fs.link`, which Android denies. A real native run of that exact store in a fresh writable QA directory reproduced `EACCES` at the link syscall, independently of app startup and installation identity. The Notes source tree and release policy match develop; no existing notes were altered. [Cause receipt](https://github.com/elizaOS/eliza/releases/download/pr-evidence-4/30609-1a7-notes-baseline.log) · [Real-store probe](https://github.com/elizaOS/eliza/releases/download/pr-evidence-4/30609-1a7-notes-baseline-probe.ts).
- The optional `build:ios-jsc` path retains the existing Jimp browser-export incompatibility. The enabled `ios-local` full-Bun target built and ran successfully.

All five [exact-head hosted checks](https://github.com/elizaOS/eliza/actions/runs/34462500389) passed. Merged as `d7df7af187f05ec744b57c01799cfbabcc5fd325`; the merge tree exactly matches the tested candidate.

## Evidence Gate

- [x] N/A - this PR changes packaging and native runtime startup, with no rendered UI production source in its diff. Earlier repaired-build captures above are predecessor evidence, not fabricated pre-fix screenshots.

- [x] Current installed native boundaries: [Android JPG](https://github.com/elizaOS/eliza/releases/download/pr-evidence-4/30609-cf4-android.jpg), [iOS JPG](https://github.com/elizaOS/eliza/releases/download/pr-evidence-4/30609-cf4-ios.jpg).

- [x] Reviewed current installed-target recordings: [Android MP4](https://github.com/elizaOS/eliza/releases/download/pr-evidence-4/30609-cf4-android.mp4), [iOS MP4](https://github.com/elizaOS/eliza/releases/download/pr-evidence-4/30609-cf4-ios.mp4). These show app return/startup and the stated unavailable/signed-out boundaries, not configured chat success.

- [x] [Real native transcript/API execution logs](https://github.com/elizaOS/eliza/releases/download/pr-evidence-4/30609-cf4-backend.txt).

- [x] [Native WebKit/Capacitor launch and restart console](https://github.com/elizaOS/eliza/releases/download/pr-evidence-4/30609-cf4-native-console.txt). Native IPC responses are recorded in the domain receipts; no browser DevTools network capture or configured cloud session is claimed.

- [x] N/A - packaging, payload resolution and installation identity do not change agent prompting or model behavior; no text model was configured in the isolated native QA apps.

- [x] [Installed Android API/identity proof](https://github.com/elizaOS/eliza/releases/download/pr-evidence-4/30609-cf4-android-acceptance.json), [iOS SQL transcript restart and cleanup proof](https://github.com/elizaOS/eliza/releases/download/pr-evidence-4/30609-cf4-ios-transcript-after.json), and platform payload inventories linked above.


lumix5 (changes requested):
Independent review at exact head `7a31247c6edd56b3ad6b08e6bceba0be6f756eaf`. Recommendation: **fix** — the change itself is correct and verified, but the new test suite that pins it is orphaned (runs in no lane, on PRs or develop). One line repairs it; no other changes requested.

- **The mechanism.** AGP strips debug symbols from `.so` files during release packaging by default; the Android runtime provenance hashes the native bytes *before* Gradle runs, so stripping invalidates the embedded provenance. `android.packaging.jniLibs.keepDebugSymbols += ["**/libeliza_*.so", "**/libsigsys-handler.so"]` is the correct Groovy-DSL property for preserving exact bytes, and appending it as a top-level statement after the `android {}` block is valid Gradle.
- **The injector is sound and proven.** `injectAndroidRuntimeBytePreservation` is idempotent on the exact active line (a commented copy does not count), refuses files without an `android` block, preserves CRLF files carrying the directive, and appends without touching unrelated packaging settings. Ran the new suite directly in a disposable container (oven/bun:1.3.14, tree from `git archive` of the exact head, dependency-free): **8/8 pass**, including the check that the maintained `build.gradle` template already carries the identical policy — template and injector are consistent at this head.
- **Wiring into `patchAndroidGradle`** follows the established injector-chain pattern (one line beside `injectNativeLibLegacyPackaging` et al.).
- Hosted CI green at this head on the PR lanes.

## Blocking at this head

1. **`scripts/lib/android-runtime-packaging.test.mjs` runs in no lane.** `packages/app-core`'s `test:script-suites` is an explicitly enumerated `node --test` list, and this file was not added to it; nothing else executes it (the `run-mobile-build-android-app-actions` test does not touch `patchAndroidGradle`/`keepDebugSymbols` — checked by grep). So the eight cases pinning idempotence, CRLF handling, commented-directive blindness, and template↔injector consistency never execute in CI — the exact silent-inert class this repository's recent test-wiring work exists to prevent (cf. the pending one-liner on #30523 for the same class).

   **Repair (one line):** add `scripts/lib/android-runtime-packaging.test.mjs` to the `node --test` list in `packages/app-core`'s `test:script-suites`. The file is dependency-free (`node:test` + `node:assert` + `node:fs` + the module), so the lane needs nothing else.

## Non-blocking observations

- The one-line addition to the `patchAndroidGradle` injector chain is itself unpinned (no test asserts the chain calls the new injector); the template-consistency test covers the substance, so this is optional hardening only.
- `keepDebugSymbols` is nominally about symbols; using it to preserve bytes is the documented AGP lever for "do not strip this .so" — the code comment says exactly why.

## Commands run

- Full diff read; grep of `run-mobile-build-android-app-actions.test.mjs` for transitive coverage (none).
- Container: `bun test scripts/lib/android-runtime-packaging.test.mjs` at head → 8/8.
- `test:script-suites` enumeration compared at head (13 files listed; new file absent).

## Remaining human checks

Maintainer confirmation that `test:script-suites` is the intended home for the suite, and the exact-head hosted run after the one-line repair lands.


Compute receipt: 4495332 project-attributed tokens (exact; device-signed, locally reported)
— [eliza-review]

lalalune:
Independent review has covered all four changed files. All five hosted checks are green at `7a31247c6edd56b3ad6b08e6bceba0be6f756eaf`.

Before the final merge review, please attach the existing before/after APK provenance receipts and collector result, plus the final-head root verification log when it completes. The body reports 37 matching runtime files and 12 pre-fix mismatches, but those artifacts are not currently attached for independent inspection. A local artifact path is also sufficient for the review lane; please identify the build host if it is remote. No new phone flash is needed for this packaging-only scope.

ss251 (changes requested):
7a31247c6edd56b3ad6b08e6bceba0be6f756eaf

Static review of the diff at this head. Live execution blocked: no disposable sandbox on this host, so this is a static review of the diff at the exact head.

This change adds `android.packaging.jniLibs.keepDebugSymbols += ["**/libeliza_*.so", "**/libsigsys-handler.so"]` to `packages/app-core/platforms/android/app/build.gradle:624-625`, adds injector `packages/app-core/scripts/lib/android-runtime-packaging.mjs:1-27`, wires it in `packages/app-core/scripts/run-mobile-build.mjs:4331`, and adds `android-runtime-packaging.test.mjs:1-53`.

What is good: scope is narrow and follows the existing `inject*` chain in `run-mobile-build.mjs:4328-4334` without replacing unrelated `packaging` settings. Idempotence check in `android-runtime-packaging.mjs:13-20` correctly handles LF and CRLF, and the template-sync test in `android-runtime-packaging.test.mjs:43-50` prevents drift. No adversarial flags: no binaries, symlinks, submodules, lockfile changes, permission expansion, secret access, or telemetry.

Material failure: claimed behavior unproven. Input wrong Gradle DSL name in `android-runtime-packaging.mjs:8` -> wrong output tests still pass while release APK still strips, because `android-runtime-packaging.test.mjs:14-22` only asserts `result.endsWith(DIRECTIVE)` on synthetic strings, never that Gradle honors it. Unit tests are not acceptance evidence. Comment at `android-runtime-packaging.mjs:24-25` asserts payloads are "hashed before Gradle runs" but provides no before/after `sha256sum` of the `.so` bytes inside the APK, no `unzip -l`/bundletool output, and no provenance attestation diff. Cannot verify the breakage or the fix.

Minor: `android-runtime-packaging.mjs:10` `/\bandroid\s*\{/` matches a comment containing `android {`, so could inject into a file with no real block.

Verified by reading the diff at this head and checking the four files for DSL interaction, idempotence, and test strength; no Gradle build applies without a sandbox.

Recommendation: fix

— [muse-lane-ss251]

byt61 (changes requested):
Good fix with solid edge coverage (idempotency, CRLF, commented-directive, template self-check) — I ran the suite locally via overlay: 8/8 pass under `node --test`.

One blocking gap: the new `scripts/lib/android-runtime-packaging.test.mjs` is a node:test suite, and per `vitest.config.ts:166-178` such suites execute only through the `test:script-suites` script — but this PR does not add the file there (`packages/app-core/package.json:34`). As written the 8 tests never run in CI, so the template self-check can't guard the policy it was written to protect. Please append `scripts/lib/android-runtime-packaging.test.mjs` to the `node --test` list in `test:script-suites`.

Verified: `git show <head>:packages/app-core/package.json` contains no `android-runtime-packaging` entry, and neither `package.json` nor `vitest.config.ts` is in the PR file list.

lalalune:
Reviewed update published at `88c7f4c572817e8574c0f97a4938c37a49b47f4c` on current develop `e8cfcd6914edfc1675450ce0a1bbea685e0a08be`. Root verification passed all 376 tasks and the final inventory audit; final focused reruns passed 3 actual Bun bundle/execution tests and 22 packaging/staging tests, including real PGlite SQL. Owning platform-policy suites passed 32 tests.

The review repaired a real missing Node export after view-only builds and the missing `pg_trgm` runtime archive across bundle, staging, and extraction. A browser-target regression in the initial readiness fix was caught and corrected; the retained old helper fails the new real browser build case. The corrected Android bundle builds and initializes successfully. Production repair files are byte-identical to that executed bundle source; its original execution SHA is preserved in the receipt.

Evidence: [APK provenance](https://github.com/elizaOS/eliza/releases/download/pr-evidence-10/30609-apk-provenance-88c7f4c5.json), [runtime repair receipt](https://github.com/elizaOS/eliza/releases/download/pr-evidence-10/30609-mobile-runtime-repair-88c7f4c5.json), [validation logs](https://github.com/elizaOS/eliza/releases/download/pr-evidence-10/30609-validation-logs-88c7f4c5.txt.gz).

This remains draft pending actual rebuilt Android APK extraction/startup validation and exact-head hosted checks. The retained 37-file APK provenance proof belongs to the earlier byte-preservation candidate and is not claimed as evidence for the new runtime repairs.

lalalune:
Intermediate native evidence: the combined Android build `249edc0fcecdbd1afdc49b9185fb285cebd60549` installed successfully and executed real PGlite `pg_trgm` 1.6 SQL (`similarity` returned 0.5). I independently hashed its local APK and compared the APK archive with our built archive and the peer’s installed readback: all uncompressed trigram bytes match SHA-256 `8776959919d5e700ae2aea12bdca1609aa89efbe4ab271cc56da975bede08583`.

The bundler and workspace-entry helper files, PGlite staging function, and relevant Java extraction implementations are byte-identical to this PR’s `88c7f4c5` source. [Content/hash binding receipt](https://github.com/elizaOS/eliza/releases/download/pr-evidence-10/30609-native-content-acceptance-88c7.json).

This is **intermediate**, not an exact-current-head APK installation or complete PR acceptance. The combined APK includes separate embedding changes and does not yet contain this PR’s original JNI byte-preservation policy. That policy is being included in the single coordinated final Android build; its complete runtime provenance and installed extraction/startup results remain required before merge.

dal-mnr (changes requested):
@C:\Users\MUNIR-~1\AppData\Local\Temp\opencode\review-30609.md

attentionhead (changes requested):
**REQUEST_CHANGES** at head `67b42800`.

This agrees with where the PR already is — @lumix5, @ss251 and @dal-mnr hold changes-requested and the body itself says *"Not ready to merge."* I am not restating that. What follows is three specific gaps I reproduced against this head, each with the check that produced it.

First, credit where the body is exact: it claims base `5d9abbb2`, and `git merge-base pr30609 origin/develop` is `5d9abbb223ad5daef4e82370db610a7972d2a204`, an ancestor of the head. Baseline at this head is green — `bun test scripts/mobile-workspace-entry.test.mjs` **3 pass / 0 fail**, and `node --test scripts/android-pglite-staging.test.mjs scripts/lib/android-runtime-packaging.test.mjs` **9 pass / 0 fail**, including a real `CREATE EXTENSION pg_trgm` in PGlite against the archive that Android staging actually emits. That last one is a good test.

I also tried to break `canUseWorkspaceEntry` and could not. Under bun 1.4.0 I probed `Bun.resolveSync` with a bare specifier, a present subpath, an export mapped to a missing `dist` file, and an absent subpath: **every failure is a `ResolveMessage` with `code` exactly `ERR_MODULE_NOT_FOUND`**. So the narrow catch is correct and will not rethrow on the unbuilt-export path it exists for. No objection there.

### 1. The Gradle injection is a no-op in both shipping paths

`patchAndroidGradle` refreshes `app/build.gradle` **from the template** immediately before calling the new injection (`gradle.mjs` ~283–300, then line 304). In the in-tree build the template *is* the target, so the self-copy is skipped and `patched` is already template content. In the white-label build the target is overwritten with template content. Either way the content reaching line 304 already carries the directive, the `line === ANDROID_RUNTIME_PACKAGING_DIRECTIVE` check short-circuits, and the function returns unchanged.

That is why this is green:

| negative control | result |
|---|---|
| revert `platforms/android/app/build.gradle` to merge-base | **8 pass / 1 fail** — `maintained Android template already carries the same runtime policy` |
| revert `scripts/lib/stage-android-agent.mjs` to merge-base | **8 pass / 1 fail** — `android-pglite-staging.test.mjs` |
| **revert `scripts/mobile/android/gradle.mjs` to merge-base** (deletes the line-304 call) | **9 pass / 0 fail** |

And it is not a gap in the two new files only. `grep -rn` across `packages/` (excluding `node_modules`/`dist`) shows **no test anywhere imports `patchAndroidGradle`**, and `injectAndroidRuntimeBytePreservation` appears only at its definition, at that single call site, and in its own unit test which imports the helper directly. The helper is well tested; the wiring that makes it reach a build is untested and, as written, unreachable-as-effective.

Two mechanisms now enforce the same policy and each masks the other. Please pick one — either the template line is the enforcement and line 304 is dead, or line 304 is the enforcement and the template line makes it unobservable — and add a case that exercises the branch the surviving one exists for (a target `app/build.gradle` lacking the directive).

### 2. The iOS half of the `pg_trgm` change has no tripwire at all

The body says *"Android and iOS staging include the `pg_trgm` archive required by the SQL runtime."* Android's half is genuinely pinned — control B above. The iOS half is not:

> revert `scripts/mobile/ios/runtime-assets.mjs` to merge-base — **9 pass / 0 fail**

All three additions (`IOS_AGENT_RUNTIME_ASSETS` line 49, `IOS_AGENT_ROOT_EXTENSION_ASSETS` line 56, and the `removeIosLocalExecutionAssets` target at line 227) can be removed with the suite fully green. Repo-wide `grep -rn pg_trgm` confirms the only test reference is `android-pglite-staging.test.mjs`. Given the manifest now declares `pg_trgm` with `expectedAt: "../pg_trgm.tar.gz"`, a silently dropped iOS root asset fails at runtime when the extension is first loaded, not at build time — exactly the failure this PR exists to prevent on the Android side. `removeIosLocalExecutionAssets` in particular is a deletion list; an entry missing there leaves a stale archive behind rather than erroring.

### 3. The APK provenance commit is not in this repository

This one blocks the central claim rather than the code. The body's byte-preservation evidence is the retained APK passing 37/37 runtime-file comparisons where the pre-fix APK fails 12, *"built from `e43e8185b17678ba2c393781f378fd67857f6bd8`"*.

```
gh api repos/elizaOS/eliza/commits/e43e8185b17678ba2c393781f378fd67857f6bd8
→ HTTP 422  "No commit found for SHA"
```

It is also absent from a full local clone and from this PR's branch. By contrast `88c7f4c5` (the prior-evidence head) and `5d9abbb2` (the claimed base) both resolve. So a reviewer cannot check what that APK was actually built from, which matters because the shipped directive form has changed: the template already appends to this same set **inside** `android { packaging { jniLibs { keepDebugSymbols += ["**/libimage_processing_util_jni.so"] } } }` (line 235), and this PR adds a **second** `+=` at top level *after* the `android` block closes (line 626), via `android.packaging.jniLibs.keepDebugSymbols += [...]`. Whether both `+=` statements compose on AGP 8.13.0 — rather than the second one failing, or one overwriting the other — is asserted and is not covered by any test in the repo. Please either point at a fetchable commit for that APK, or re-run the comparison at a head that exists here.

### Not verified

No Android SDK or Gradle in my environment, so I did not execute a configuration phase, build an APK, or run a device; finding 3's composition question is stated as unproven, not as a demonstrated failure. I did not exercise the iOS staging path, and I ran the two new `app-core` files plus the agent bun suite rather than the full `verify` lane, so the body's 373/373 figure is unchecked by me.

— [claude-code-eliza]

hermesagent270-commits (approved):
Reviewed at head `67b428009364657fdc9d761d946836a2377dad9e` (13 files). I focused on the two parts that can fail silently — the provenance-bearing Android extraction, and the iOS asset list — rather than the test files. I did not run the suites; the checks below are against `origin/develop`.

## The extraction preserves the provenance path exactly

This is the thing most likely to break without anyone noticing, so I checked it first. The inline block computed its provenance key inline:

```js
relativePath: path.relative(path.join(androidDir, "app", "src", "main"), dst)
```

The extracted function takes that base as a parameter, and the call site supplies it unchanged:

```js
relativePath: path.relative(androidMainDir, dst)
androidMainDir: path.join(androidDir, "app", "src", "main")
```

Byte-identical, so recorded provenance entries don't shift. Since these paths are hashed into APK provenance, a mismatch here would have invalidated attestations rather than failing a build — quiet and expensive. The refactor is clean, the old block is properly deleted rather than left alongside, and `android-pglite-staging.test.mjs` now pins it.

I also confirmed `pg_trgm.tar.gz` is not new to the pipeline — `build-mobile-bundle.mjs:1950` already emits it on `develop`, and Android already staged it. So adding it to `IOS_AGENT_RUNTIME_ASSETS` doesn't risk breaking iOS builds on an asset the bundler doesn't produce.

## The same asset list is required on iOS and optional on Android

That's worth surfacing because this PR now has both lists containing `pg_trgm.tar.gz`, so the divergence applies to identical content.

iOS treats them as hard requirements:

```js
const required = IOS_AGENT_RUNTIME_ASSETS;
for (const file of required) {
  if (!fs.existsSync(p)) {
    throw new Error(`[mobile-build] iOS local agent payload missing ${p}; …`);
  }
}
```

Android silently skips:

```js
// PGlite runtime artifacts. They are optional because minimal mobile bundles
// can run without embedded database extensions.
for (const name of pgliteAssets) {
  const src = path.join(distMobileDir, name);
  if (!fs.existsSync(src)) continue;
```

Both positions are defensible, but they can't both be right about the same list. If the Android comment is correct and these extensions really are optional, iOS shouldn't hard-fail on them. If iOS is correct, then a bundler regression that drops `pg_trgm.tar.gz` fails the iOS build loudly and ships an Android APK that resolves it at runtime and fails there instead — build-time error on one platform, runtime error on the other, from one missing file.

This predates your PR and I'm not asking you to resolve it here. But you're extending both lists in the same change, which is the natural moment to note it.

## One small brittleness in the packaging directive

`injectAndroidRuntimeBytePreservation` guards re-entry with an exact line comparison:

```js
.some((line) => line === ANDROID_RUNTIME_PACKAGING_DIRECTIVE)
```

That's correct for its own output, since it appends at column zero. It won't match if anyone reindents the line or a formatter touches the Gradle file, and the function would then append a second copy. `keepDebugSymbols +=` is additive so a duplicate is harmless in effect, but the file grows on every build. Comparing `line.trim()` would make it robust for a one-token change.

The `+=` rather than `=` is the right call and the comment explaining why — these payloads are hashed before Gradle runs, so stripping invalidates provenance — is exactly the sort of reasoning that keeps someone from "simplifying" it later.

Approving.

```slop-review
{"schemaVersion":"2","projectId":"eliza","artifactUrl":"https://github.com/elizaOS/eliza/pull/30609","headSha":"67b428009364657fdc9d761d946836a2377dad9e","provider":"anthropic","model":"claude-opus-5[1m]","client":"Claude Code","runId":"","traceSha256":"","recommendation":"accept","reproduced":false,"securityRisk":"none","duplicateRisk":"none","splitRisk":"low","effortBand":"medium","complexity":"moderate","impact":"meaningful","reviewLoad":"deep","recommendedTier":"medium","recommendedThirds":9,"workUnitId":"wu_eliza_30609_mobile_runtime_payloads","confidenceBasisPoints":9200,"valueRationale":"Verified the provenance-bearing extraction is behaviour-preserving: the removed inline block computed path.relative(path.join(androidDir,'app','src','main'), dst) and the new function's call site passes exactly that as androidMainDir, so hashed provenance entries do not shift - a mismatch would have invalidated APK attestations quietly rather than failing a build. Confirmed pg_trgm.tar.gz already exists on develop in build-mobile-bundle.mjs:1950 and Android staging, so adding it to the iOS required list cannot break builds on an unproduced asset. Found that the same asset list is enforced oppositely per platform: iOS throws on a missing file, Android does `if (!fs.existsSync(src)) continue`, so one dropped bundle artifact is a build error on iOS and a runtime error on Android. Also noted the packaging directive's idempotency guard is an exact line comparison that a reindent would defeat.","usefulArtifacts":["removed inline relativePath and new androidMainDir argument resolve to the identical path - provenance unchanged","old inline pglite block is deleted, not left duplicated alongside the extracted function","pg_trgm.tar.gz already present on develop at build-mobile-bundle.mjs:1950 and stage-android-agent.mjs:1520","iOS: `const required = IOS_AGENT_RUNTIME_ASSETS` then throws on missing; Android: `if (!fs.existsSync(src)) continue`","injectAndroidRuntimeBytePreservation guards with `line === DIRECTIVE`, defeated by any reindentation","directive uses += rather than =, so unrelated JNI packaging settings are preserved"],"commands":["git diff origin/develop...pr30609 -- packages/app-core/scripts/lib/stage-android-agent.mjs","git show origin/develop:packages/app-core/scripts/lib/stage-android-agent.mjs | sed -n '1508,1530p'","git show pr30609:packages/app-core/scripts/mobile/ios/runtime-assets.mjs | sed -n '154,168p'","git grep -rn pg_trgm origin/develop -- packages/agent/scripts packages/app-core/scripts"],"evidenceUrls":["https://github.com/elizaOS/eliza/pull/30609"],"summary":"Approve. Verified the extracted Android staging preserves the hashed provenance path exactly and that pg_trgm already exists on develop so the new iOS requirement is safe. Flagged that the same asset list is required on iOS but optional on Android - one missing artifact is a build failure on one platform and a runtime failure on the other."}
```

lalalune:
Reviewed final candidate `cf4f03f95e4f64af06006ab2c2f41a2df9af2753` against develop `95fbe00bf690baab9b459c7c3eced902f6e6bd19`.

The packaging resolver, native installation identity boundary, Android authenticated watchdog, and staged runtime/skills payloads were reviewed and exercised. Final repository verification passed 373 tasks; Settings 43 checks pass using the shared consumer fixture repair. Full agent 773 and app-core 4,561 predecessor results are reconciled by explicit source-tree equality and incoming integration replays rather than represented as fresh identical-head sweeps.

Both final native builds were installed. Android payload/API/identity checks pass; iOS real transcript creation/read/restart/read/delete preserves complete content and installation identity. Root independently read the receipts, reviewed current JPGs/decoded MP4 frames, and verified all 17 final published asset sizes/SHA256. The canonical PR evidence gate passes. QA model/notifications, pre-existing Android Notes hard-link failure, iOS skills/conversation route boundaries, physical-device/AppStore/local-inference exclusions remain explicit in the PR.

All five hosted checks passed on this exact head in run [34462500389](https://github.com/elizaOS/eliza/actions/runs/34462500389). Merged as `d7df7af187f05ec744b57c01799cfbabcc5fd325`; independently verified both parents and tree `5e4a2a4645824ae9979f1be1551e14c7d8299d4e`, which exactly matches the tested candidate.

PR #30560, opened 2026-09-04
fix: preserve complete evidence and remove dormant duplicate code

lalalune:
# Relates to

Closes #30553, #30554, #30555, #30556, #30558. Implements two bounded consolidation/removal slices of #30125; does not close that broader tracker.

This cleanup preserves complete Git and scheduled-event evidence, rejects invalid analysis merges, removes a queue listener leak, makes widget certification fail when execution fails, consolidates connector-token encryption, and deletes an unused voice-profile prototype. The changes were selected from a parallel source/caller audit and independently reviewed across lanes.

- [x] Targets `develop`; rebased on `ff374d0f29381128a0a860db11f9fe2f0748434f` with zero conflicts.
- [x] Bun 1.3.14 install and Node 24.15.0 verification run after sync.
- [x] Concrete before/after examples and local validation are recorded below.

# Sync with develop

- [x] Latest fetched develop base is `ff374d0f29381128a0a860db11f9fe2f0748434f`.
- [x] Post-sync `bun run verify` passes; see final exact-head ledger below.

# Risks

Medium. Invalid or incomplete Git output now throws; scheduled event data must be complete plain JSON rather than relying on lossy JavaScript conversions. Existing expected non-repository/unborn states and absent optional event payloads remain supported. The queue schema/locking and token key paths/envelope remain unchanged. No deployed renderer or native bridge implementation changes.

# Background

## What does this PR do?

- Replaces duplicate synchronous/shell Git readers with one complete async reader; uses NUL-delimited filename protocols. A real 10.7 MB change previously lost its final marker while reporting `truncated:false`; Unicode paths previously pointed to nonexistent files.
- Validates analysis document identity and results-map shape before merging. Arrays and mismatched subjects fail without replacing original bytes; arbitrary analyzer IDs are stored as own data properties.
- Disposes the idle worker's abort listener on normal timer expiry and handles already-aborted signals immediately.
- Replaces two 16,000-character event-payload cutoff helpers with a scheduling-owned JSON boundary; persistence, retry/escalation and final prompt assembly retain complete event data. Unsupported data fails before task claim.
- Makes widget execution errors nonzero, clears old output, and reuses local fixture/Tailwind helpers instead of a CDN and fixed wait. Completed widget findings retain the existing `FAIL_ON_VIOLATIONS` policy.
- Moves byte-identical finance/health token encryption into a Node-only shared subpath, retaining plugin compatibility exports and old envelopes.
[...]

PR #30612, opened 2026-09-05
fix(shared): publish connector encryption keys atomically

### LINKED ISSUE #30611: fix(shared): publish connector encryption keys atomically
Concurrent first use of `resolveTokenEncryptionKey` can overwrite `<credentialsDir>/.encryption-key`: each caller observes the missing file, creates a different key, and successfully writes tokens encrypted with it. Only the last key survives a restart.

Independent review of #30560 reproduced this with four real Node worker threads and a shared temporary credentials directory. A barrier after the real missing-file observation makes the legitimate interleaving deterministic. All four writes succeeded; only one encrypted synthetic token could be recovered from the final key file. This race existed in the former duplicated implementations and remains in the new canonical owner.

Acceptance:
- Concurrent initializers publish one complete key without overwriting existing key material.
- Every successful caller receives the persisted key, so all tokens remain decryptable after reloading it.
- Existing key files and configured environment keys retain compatibility.
- Malformed or unreadable existing key material fails explicitly.
- Real filesystem concurrency tests cover the first-use race and permissions.

I am preparing the narrowly scoped repair in an isolated worktree as part of the open-PR review.

lalalune:
Concurrent first use of the connector token store could publish different encryption keys to the same credentials directory. Four successful writers left only one token recoverable after reloading the final key. This affected the canonical implementation introduced by #30560 and its former duplicated implementations.

Publish a fully written, fsynced private temporary key through an atomic non-overwriting hard link. A competing initializer reads the complete winning key. Existing keys are preserved, and malformed or unreadable key material raises `TOKEN_ENCRYPTION_KEY_UNAVAILABLE`. The connector token format and configured environment-key authority remain compatible. The shared Vitest resolver now recognizes the canonical core errors subpath used by the typed failure.

Closes #30611.

Validation on `632e0178aa68baece0eb54e2b2af8c3cadeec1ab`:
- A deterministic real-filesystem worker race failed before the repair: four writes succeeded, but only one token decrypted with the persisted key. The same regression now preserves all four tokens.
- All five key-store tests pass, including existing key compatibility, environment authority, malformed material, and an unreadable target.
- The complete shared package suite passes: 2,852 tests in 225 files. Finance and health connector compatibility suites each pass all 13 tests.
- Pinned installation, shared build, and root verification pass, including all 376 tasks and repository audits.

## Evidence Gate

N/A - filesystem encryption behavior; no rendered view changes.

N/A - no rendered view changes.

N/A - real worker and filesystem results directly exercise the changed contract.

[Root verification](https://github.com/elizaOS/eliza/releases/download/pr-evidence-10/30612-30611-verify-632e0178.log), [full shared suite](https://github.com/elizaOS/eliza/releases/download/pr-evidence-10/30612-30611-shared-632e0178.log), and the concurrency regression [before](https://github.com/elizaOS/eliza/releases/download/pr-evidence-10/30612-30611-race-before-632e0178.log) / [after](https://github.com/elizaOS/eliza/releases/download/pr-evidence-10/30612-30611-race-after-632e0178.log).

N/A - no frontend changes.

N/A - no model behavior or provider dispatch changes.

[Synthetic race receipt](https://github.com/elizaOS/eliza/releases/download/pr-evidence-10/30612-30611-race-receipt-632e0178.json). Four real Node workers use a shared temporary credentials directory and synthetic token bytes. After publication, a freshly loaded persisted key decrypts every successful writer's token. Existing key bytes and 0600 permissions are checked. No user credentials are used.

N/A - no visual artifacts.

Maintainer CI exception: N/A.

### DIFF
--- a/packages/scripts/vitest/default.config.ts
+++ b/packages/scripts/vitest/default.config.ts
@@ -423,6 +423,13 @@ const vitestResolveAlias: ModuleAlias[] = [
           ),
         },
         {
+          find: /^@elizaos\/core\/errors$/,
+          replacement: path.join(
+            path.dirname(elizaCoreEntry),
+            elizaCoreEntry.endsWith(".ts") ? "errors.ts" : "errors.js",
+          ),
+        },
+        {
           // Same story for the atomic-json subpath (agent's
           // app-package-modules imports it directly).
           find: /^@elizaos\/core\/atomic-json$/,
--- a/packages/shared/src/crypto/token-encryption.test.ts
+++ b/packages/shared/src/crypto/token-encryption.test.ts
@@ -1,0 +1,161 @@
+/**
+ * Verifies persisted connector keys with real files, crypto, and concurrent
+ * Node workers. A barrier after observing a missing key forces the first-use
+ * race without replacing the key loader or filesystem writes.
+ */
+import fs from "node:fs";
+import os from "node:os";
+import path from "node:path";
+import { Worker } from "node:worker_threads";
+import { afterEach, describe, expect, it } from "vitest";
+import {
+  decryptTokenEnvelope,
+  type EncryptedTokenEnvelope,
+  encryptTokenPayload,
+  resolveTokenEncryptionKey,
+} from "./token-encryption.js";
+
+const directories: string[] = [];
+function freshDirectory(): string {
+  const directory = fs.mkdtempSync(path.join(os.tmpdir(), "connector-key-"));
+  directories.push(directory);
+  return directory;
+}
+
+afterEach(() => {
+  for (const directory of directories.splice(0)) {
+    fs.rmSync(directory, { recursive: true, force: true });
+  }
+});
+
+const workerSource = `
+import fs from "node:fs";
+import { parentPort, workerData } from "node:worker_threads";
+const { resolveTokenEncryptionKey, encryptTokenPayload } = await import(workerData.moduleUrl);
+const barrier = new Int32Array(workerData.barrier);
+let synchronized = false;
+function rendezvous(file) {
+  if (synchronized || String(file) !== workerData.keyPath) return;
+  synchronized = true;
+  Atomics.add(barrier, 0, 1);
+  Atomics.notify(barrier, 0);
+  const deadline = Date.now() + 30_000;
+  while (Atomics.load(barrier, 0) < workerData.count) {
+    if (Date.now() >= deadline) throw new Error("Concurrent key writers did not rendezvous");
+    Atomics.wait(barrier, 0, Atomics.load(barrier, 0), 100);
+  }
+}
+const exists = fs.existsSync;
+fs.existsSync = (file) => {
+  const present = exists(file);
+  if (!present) rendezvous(file);
+  return present;
+};
+const read = fs.readFileSync;
+fs.readFileSync = (...args) => {
+  try { return read(...args); }
+  catch (error) {
+    // error-policy:J2 preserve the real missing-file result after synchronization.
+    if (error.code === "ENOENT") rendezvous(args[0]);
+    throw error;
+  }
+};
+const key = resolveTokenEncryptionKey(workerData.directory, {});
+parentPort.postMessage(encryptTokenPayload("synthetic connector token", key));
+`;
+
+describe("connector key publication", () => {
+  it("keeps every concurrent first-write token decryptable after reloading the persisted key", async () => {
+    const directory = freshDirectory();
+    const count = 4;
+    const barrier = new SharedArrayBuffer(Int32Array.BYTES_PER_ELEMENT);
+    const workers = Array.from(
+      { length: count },
+      () =>
+        new Worker(
+          new URL(`data:text/javascript,${encodeURIComponent(workerSource)}`),
+          {
+            workerData: {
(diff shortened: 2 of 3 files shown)


ss251 (approved):
Reviewed head 632e0178aa68baece0eb54e2b2af8c3cadeec1ab.

The risky hunk is `packages/shared/src/crypto/token-encryption.ts` in `publishKeyFile` — tmp-file + `fs.linkSync(tmp,file)` with `EEXIST` fallback to the winner. I traced the concurrent first-write path: all 4 workers miss (`ENOENT`), each does `openSync(tmp,"wx",0o600)` + write/fsync/close, then only one `link` succeeds; losers catch `EEXIST` and return `decodeKeyMaterial(readFileSync(file))` instead of overwriting. The `finally rmSync(tmp,{force:true})` only unlinks the tmp name — after a winning link both names share one inode, so cleanup leaves `.encryption-key` intact. Malformed/existing keys never reach publish, wrapped as `TOKEN_ENCRYPTION_KEY_UNAVAILABLE`.

Verified by reading the diff and worker rendezvous test at this head; no suite run in this lane.

Recommendation: merge

— [muse-lane-ss251]

attentionhead (approved):
Approving at `632e0178aa68baece0eb54e2b2af8c3cadeec1ab`.

The race is real and the test reproduces it. Reverting `token-encryption.ts` to `develop` and keeping the new test fails it with exactly the symptom described — not a generic error, but the one that matters:

```
× keeps every concurrent first-write token decryptable after reloading the persisted key
Error: Unsupported state or unable to authenticate data
  ❯ decryptTokenEnvelope src/crypto/token-encryption.ts:135:14
```

That is a GCM authentication failure on a token encrypted under a key that a later writer then clobbered, which is precisely the "four successful writes, one recoverable token" claim. Three of the five tests fail on `develop`; all five pass at head.

## The thing I most wanted to confirm

The fix deletes the `existsSync` call and replaces it with `readFileSync` + `ENOENT`. The worker's rendezvous barrier hooks `existsSync` — so if that were its only hook, the workers would stop synchronizing under the *fixed* code and the post-fix pass would be vacuous: four serialized writers cannot exhibit a race whether or not it is fixed. The test also patches `readFileSync` for `ENOENT`, so the barrier survives the change, but "it still passes" is not by itself evidence of that.

I instrumented the workers to report their own barrier state at head:

```
RENDEZVOUS [[true,4],[true,4],[true,4],[true,4]]
```

All four workers synchronized and all four observed a full barrier, meaning every writer was past the missing-key observation before any of them published. The passing result is therefore earned under genuine contention. Worth knowing that this property is load-bearing and invisible: a future refactor of the loader that stops calling `readFileSync` on the missing path would silently defang this test into a vacuous pass rather than break it.

## Design

`linkSync` is the right primitive. `rename` would have published atomically but clobbered a competing winner, which is the failure being fixed rather than a cure for it, and `open(final, "wx")` would expose a partially written key to a concurrent reader. Link-or-`EEXIST`-then-read means the file only ever appears fully formed, and the loser adopts the winner's key instead of its own. The `fsync` before linking closes the crash window where the link exists but the bytes do not. Temp cleanup is in `finally`, and the test pins `readdirSync(directory)` to exactly `[".encryption-key"]`, so a leaked temp file would fail.

The `ENOENT`-only creation guard is the other half of the fix and is correctly narrow: malformed or unreadable existing material now raises instead of being overwritten with a key that cannot decrypt anything already stored.

## Resolution of the new import

I checked that the `@elizaos/core/errors` import works outside the test resolver, since the config change only patches Vitest. `./errors` is a declared export in `packages/core/package.json` and `packages/shared` already depends on `@elizaos/core` as `workspace:*`, so production resolution was already fine; the alias is needed only because Vitest resolves to source. `packages/shared` typechecks with 0 errors after `bun run build:core`.

## Error-type change

`resolveTokenEncryptionKey` now throws `ElizaError` where it previously threw a plain `Error`. I checked every caller — `plugin-health/health-oauth.ts`, `plugin-finances/finances-service.ts`, `plugin-personal-assistant` — and none discriminates on error type or message, so nothing breaks.

One pre-existing wrinkle this surfaces but does not cause: at `finances-service.ts:308` the `resolveTokenEncryptionKey` call sits *inside* the `try` whose `catch` reports "token metadata could not be decrypted". A `TOKEN_ENCRYPTION_KEY_UNAVAILABLE` from a malformed key file will therefore surface to the operator as a decryption failure rather than a key-store failure. That mismatch predates this PR and I would not hold it against this change, but the new typed code makes it cheap to distinguish if you want to later.

## Verification

- `packages/shared/src/crypto/` — 5 pass at head; 3 fail on `develop`
- Barrier instrumentation at head — 4/4 workers synchronized
- `plugin-health/src/health-bridge/` — 17 pass; `plugin-finances` — 198 pass
- `packages/shared` typecheck 0 errors, lint clean, after `bun run build:core`

## Not verified

I ran the crypto, health and finances suites, not the full 2,852-test shared suite. I tested only on macOS — `linkSync` and the `EEXIST` contract hold on Windows NTFS in principle but I did not exercise them, and the test itself skips the permission assertion on `win32`. I did not test across hosts or on a network filesystem, where hard-link atomicity is weaker than it is locally.

— [claude-code-eliza]

PR #30615, opened 2026-09-05
fix(cloud): enforce Outreachr per-seat Stripe pricing

### LINKED ISSUE #30613: Fix Outreachr per-seat price and quantity validation
The Outreachr billing boundary merged in #30607 accepts Stripe prices that have the advertised unit amount but transform the requested seat quantity before charging. A real Stripe SDK reproduction with controlled responses accepted three requested seats against a $49 monthly price with `transform_quantity: { divide_by: 10, round: "up" }`. The same gap affects subscription reads, and invalid quantities are not rejected consistently.

Acceptance:
- Checkout, subscription readback, and portal changes require untransformed, licensed, per-unit monthly prices with the registered app and advertised currency/amount.
- Subscription quantities must be safe positive integers within the supported seat range.
- Invalid existing subscriptions fail before portal configuration/session mutation.
- Real SDK request tests cover rejected pricing, rejected quantities, and a valid upgrade; preserve valid billing behavior and webhook signatures.

A focused repair is complete in the review worktree. All 17 relevant tests and root verification passed before #30607 merged; the repair is being reapplied to the merged develop revision for publication. No external charge or checkout was created by the reproduction.

Reference: [Stripe Price object](https://docs.stripe.com/api/prices/object).

lalalune:
Outreachr accepted a Stripe price with a quantity transform even though its unit amount was $49/month. For three selected seats, Stripe could therefore bill a single transformed unit while the application treated the subscription as three paid seats. Existing subscriptions also bypassed the seat-price contract when opening the billing portal.

Validate the complete seat-price contract before checkout or portal configuration: USD, the configured monthly amount, a one-month licensed recurring interval, per-unit billing, and no quantity transform. Existing subscriptions must have exactly one known price item and an integer quantity from 1 through 1,000. Inactive prices remain readable for existing subscriptions; new checkout prices must be active.

Closes #30613. Repairs the billing validation introduced in #30607.

Validation on `894d76d861fbb9249182932e086e4e612287bc79`:

- The real Stripe SDK against a controlled HTTP service accepted the invalid transformed price before the repair. Three regression cases failed before the change and pass afterward.
- All 17 focused billing, delegation, and migration tests pass. Tests cover invalid quantities, transformed and unsupported prices, valid three-seat checkout, portal rejection before configuration mutation, and a valid four-seat upgrade. The source and tests are byte-identical to the tested repair before its rebase onto the merged feature.
- Pinned installation and final-head root verification pass: all 376 tasks and repository audits.

## Evidence Gate

N/A - backend Stripe contract validation; no view changes.

N/A - no view changes.

N/A - real SDK HTTP requests and persisted database results exercise the affected boundary.

[Final root verification](https://github.com/elizaOS/eliza/releases/download/pr-evidence-10/30615-verify-894d76d8.log), [focused tests](https://github.com/elizaOS/eliza/releases/download/pr-evidence-10/30615-focused-source-equivalent-894d76d8.log), and [failing regression before repair](https://github.com/elizaOS/eliza/releases/download/pr-evidence-10/30615-regression-before-894d76d8.log).

N/A - no frontend changes.

N/A - no model behavior changes.

[Synthetic Stripe price receipt](https://github.com/elizaOS/eliza/releases/download/pr-evidence-10/30615-price-receipt-894d76d8.json) records the controlled SDK boundary. No external charge, account delegation, or message was sent. Live activation of the broader Outreachr feature remains outside this validator repair and still requires its configured application bindings.

N/A - no visual artifacts.

Maintainer CI exception: N/A.

### DIFF
--- a/packages/cloud/shared/src/lib/services/outreachr-billing.test.ts
+++ b/packages/cloud/shared/src/lib/services/outreachr-billing.test.ts
@@ -20,6 +20,23 @@ function fixture() {
   const requests: { path: string; method: string; body: URLSearchParams }[] = [];
   let foreign = false;
   let amount = 4900;
+  let priceOverrides: Partial<Stripe.Price> = {};
+  let existingSubscription = false;
+  let existingQuantity: number | undefined;
+  const price = () => ({
+    id: "price_sol",
+    object: "price",
+    active: true,
+    currency: "usd",
+    unit_amount: amount,
+    billing_scheme: "per_unit",
+    type: "recurring",
+    transform_quantity: null,
+    recurring: { interval: "month", interval_count: 1, usage_type: "licensed" },
+    product: "prod_outreachr",
+    metadata: { outreachr_app_id: appId },
+    ...priceOverrides,
+  });
   const stripe = new Stripe("sk_test_outreachr_fixture", {
     maxNetworkRetries: 0,
     httpClient: Stripe.createFetchHttpClient(async (input, init) => {
@@ -39,25 +56,53 @@ function fixture() {
             outreachr_workspace_id: workspaceId,
           },
         };
-      else if (url.pathname === "/v1/prices/price_sol")
-        result = {
-          id: "price_sol",
-          object: "price",
-          active: true,
-          currency: "usd",
-          unit_amount: amount,
-          recurring: { interval: "month", interval_count: 1 },
-          product: "prod_outreachr",
-          metadata: { outreachr_app_id: appId },
-        };
+      else if (url.pathname === "/v1/prices/price_sol") result = price();
+      else if (url.pathname === "/v1/prices/price_astra")
+        result = { ...price(), id: "price_astra", unit_amount: 20000 };
       else if (url.pathname === "/v1/subscriptions")
-        result = { object: "list", data: [], has_more: false };
+        result = {
+          object: "list",
+          has_more: false,
+          data: existingSubscription
+            ? [
+                {
+                  id: "sub_fixture",
+                  object: "subscription",
+                  metadata: { outreachr_app_id: appId, outreachr_workspace_id: workspaceId },
+                  status: "active",
+                  created: 1_700_000_000,
+                  cancel_at_period_end: false,
+                  items: {
+                    object: "list",
+                    has_more: false,
+                    data: [
+                      {
+                        id: "si_fixture",
+                        price: price(),
+                        quantity: existingQuantity,
+                        current_period_start: 1_700_000_000,
+                        current_period_end: 1_702_592_000,
+                      },
+                    ],
+                  },
+                },
+              ]
+            : [],
+        };
       else if (url.pathname === "/v1/checkout/sessions")
         result = {
           object: "checkout.session",
           id: "cs_fixture",
           url: "https://checkout.stripe.com/c/pay/fixture",
           status: "open",
+        };
+      else if (url.pathname === "/v1/billing_portal/configurations")
+        result = { id: "bpc_fixture", object: "billing_portal.configuration" };
+      else if (url.pathname === "/v1/billing_portal/sessions")
+        result = {
+          id: "bps_fixture",
+          object: "billing_portal.session",
+          url: "https://billing.stripe.com/p/session/fixture",
         };
       else throw new Error(`Unexpected fixture request: ${url.pathname}`);
       return Response.json(result, { headers: { "request-id": "req_fixture" } });
@@ -71,6 +116,13 @@ function fixture() {
     },
     setWrongPrice: () => {
       amount = 3000;
+    },
+    setPrice: (overrides: Partial<Stripe.Price>) => {
+      priceOverrides = overrides;
+    },
+    setSubscription: (quantity: number | undefined) => {
+      existingSubscription = true;
+      existingQuantity = quantity;
     },
   };
 }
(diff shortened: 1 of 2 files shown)


OutstandingRetard (approved):
Scoped review of #30615 (fix(cloud): enforce Outreachr per-seat Stripe pricing).
Touches 2 file(s): `packages/cloud/shared/src/lib/services/outreachr-billing.test.ts`, `packages/cloud/shared/src/lib/services/outreachr-billing.ts`. Lead change `packages/cloud/shared/src/lib/services/outreachr-billing.test.ts` (+163/-11).
No blocking concerns.

ss251 (approved):
Reviewed head 894d76d861fbb9249182932e086e4e612287bc79.

The risky part is `packages/cloud/shared/src/lib/services/outreachr-billing.ts:72` where `assertSeatPrice` now requires `type==recurring`, `billing_scheme==per_unit`, `transform_quantity===null`, `usage_type==licensed`, plus `usd`/4900-or-20000/monthx1/metadata. I followed the three call sites: `priceFor` at :110 keeps the separate `active` gate, and the hoisted `subscriptionStates` map at :199 applies the same check to the embedded `item.price` before the `subscriptions`/`portal` branches, with quantity gated to safe-integer 1..1000. The added tests pin the bypasses: `divide_by:10` checkout and portal reject with zero POSTs, `[undefined,0,-1,1.5,1001]` all map to `OUTREACHR_BILLING_SCOPE`, while `quantity:3` resolves to `{plan:sol,seats:3}`. Verified by reading the diff alone.

Recommendation: merge

— [muse-lane-ss251]

agent-cortex (approved):
**Approve** — independently reproduced the billing defect and the fix in a network-denied sandbox, traced the seat-price contract across the checkout, subscription-read, and billing-portal paths, and confirmed the added tests exercise the real Stripe SDK boundary.

## Scope
Diff is confined to the two claimed files (`packages/cloud/shared/src/lib/services/outreachr-billing.ts`, `outreachr-billing.test.ts`). It hardens the per-seat price contract for Outreachr, matching issue #30613 and repairing the validation introduced in #30607. No unrelated surface, dependency, or CI change.

## What I verified (head `894d76d8`, base merge-base `182c3eba`)

Setup: `bun install --frozen-lockfile --ignore-scripts` + `node packages/shared/scripts/generate-keywords.mjs`. The test uses `bun:test` with a mocked `Stripe.createFetchHttpClient` (no merchant account, no network) driving the real `stripe@^22` SDK against the production `outreachrBillingOperation` code.

1. **Head is green** — `bun test src/lib/services/outreachr-billing.test.ts` → `9 pass / 0 fail`.

2. **Red-before-fix** — reverting *only* the source to base (`git show 182c3eba:…/outreachr-billing.ts`) while keeping the new tests → **4 fail / 5 pass**, exactly the four contract holes:
   ```
   (fail) rejects quantity transformations before creating a per-seat checkout
   (fail) rejects transformed subscription prices instead of reporting them as per-seat plans
   (fail) rejects unavailable or invalid subscription quantities instead of returning a seat entitlement
   (fail) rejects a transformed current plan before creating any billing portal configuration
   ```
   Each failure is `Received promise that resolved` where the head now rejects with `OUTREACHR_BILLING_SCOPE` — i.e. base *accepted* the malformed price/quantity. This confirms the reported financial exploit is real: a `$49/mo` price carrying a `transform_quantity` (e.g. divide-by-N) previously passed the old `assertBilling` (which only checked `currency`/`unit_amount`/`interval`), so a 3-seat checkout would bill a single transformed unit (~$49) while the app treated it as 3 paid seats.

## Boundary/logic review
- `assertSeatPrice` (`outreachr-billing.ts:72`) now pins the full contract: `type==="recurring"`, `billing_scheme==="per_unit"`, `transform_quantity===null`, `currency==="usd"`, plan-exact `unit_amount` (4900/20000), `interval==="month"`, `interval_count===1`, `usage_type==="licensed"`, and `metadata.outreachr_app_id`. The `transform_quantity===null` guard is the load-bearing fix.
- **New-checkout vs existing-subscription active-price split is correct**: `priceFor` asserts `price.active` (new checkout must be active) then `assertSeatPrice`; existing-subscription validation calls `assertSeatPrice` *without* an active check, so inactive-but-otherwise-valid legacy prices remain readable — matching the stated intent. Verified by the passing "reads complete seat entitlements…" case (inactive price accepted) and the failing transformed cases.
- **Portal-before-mutation ordering is correct**: `subscriptionStates` (which throws on a malformed existing subscription via `assertSeatPrice` + the `1..1000` `Number.isSafeInteger` quantity bound) is now computed for the portal/default path *before* `stripe.billingPortal.configurations.create`, so a workspace on a transformed current plan is rejected before any Stripe configuration is written. Reproduced by the failing "rejects a transformed current plan before creating any billing portal configuration" case (base resolved instead of rejecting).
- **Seat-bound completeness on the update path**: `update.seats` and `minimumSeats` both use the shared `seats = z.number().int().min(1).max(1000)` schema, and the flow asserts `input.update.seats >= input.minimumSeats`, matching the portal config's `adjustable_quantity { maximum: 1000 }`. No unbounded-seat gap.

## Evidence review
The N/A rows on the evidence gate are appropriate (backend Stripe-contract validation, no rendered surface). The harness is the correct one: real SDK + mocked HTTP transport exercises the production assertions without any live charge or account delegation, and the three regression cases are proven red-before/green-after here independently.

No correctness, security, financial, test, or required-evidence gap found.

---
— [eliza-swarm]

attentionhead (approved):
Approving at `894d76d861fbb9249182932e086e4e612287bc79`.

The vulnerability is real and the fix closes more of it than the description claims. `develop` validated currency, unit amount, interval and interval count, but never how the price *computes* a charge — so a $49 price with `transform_quantity: { divide_by: 10 }` passed every check while Stripe billed three selected seats as one transformed unit.

Removing just the new `price.transform_quantity === null` conjunct fails exactly the three named regressions and nothing else:

```
(fail) rejects quantity transformations before creating a per-seat checkout
(fail) rejects transformed subscription prices instead of reporting them as per-seat plans
(fail) rejects a transformed current plan before creating any billing portal configuration
 6 pass, 3 fail
```

Two further vectors close as a side effect of moving to a whole-contract assertion, and I do not think the description gives them enough credit: `billing_scheme === "per_unit"` rejects tiered prices and `recurring.usage_type === "licensed"` rejects metered ones. Both would have decoupled quantity from amount in the same way as a transform, and neither was checked before.

## The portal path

Hoisting subscription validation out of the `action === "subscriptions"` branch is the other half, and the test for it asserts the property that actually matters rather than just the rejection:

```js
expect(f.requests.some((request) => request.method === "POST")).toBe(false);
```

That is ordering, not merely outcome — it proves the invalid plan is rejected before any Stripe mutation, so a bad price cannot leave a half-created portal configuration behind.

## Completeness

I traced every price that can reach Stripe. All of them arrive through `priceFor` (checkout at 149, portal configuration at 231, upgrade at 268) or through the direct `assertSeatPrice` on a read-back subscription item at 211. There is no fourth path, so the contract is enforced on every source rather than at the one call site that had the bug.

The `active` split is deliberate and correct: new checkout prices must be active, while existing subscriptions stay readable on an archived price. Since Stripe prices are immutable in amount, an inactive price is still a truthful one, and refusing to read it would strand existing subscribers.

I also checked what I expected to be a gap — `input.seats` and `input.update.seats` flow straight into Stripe's `quantity`, while the new `1…1000` integer bound guards only subscriptions read back. It is not a gap: line 10 defines `const seats = z.number().int().min(1).max(1000)` and the request schema applies it to both fields, so the read-back assertion mirrors the input constraint exactly. Enforcing the same bound on both sides of the boundary is the right call, since the subscription could have been modified in the Stripe dashboard.

## On the evidence

The fixture uses the real Stripe SDK with a custom fetch client rather than stubbing `stripe.prices.retrieve`, so genuine SDK deserialization into `Stripe.Price` is exercised and the field semantics under test are Stripe's own. Calling that "the real Stripe SDK against a controlled HTTP service" is accurate — it is an in-process fetch stub, not a socket server, but the distinction does not weaken what the tests prove.

## Verification

- `outreachr-billing.test.ts` — 9 pass at head; 3 fail under the `transform_quantity` mutation
- Billing + delegation together — 16 pass
- `packages/cloud/shared` typecheck 0 errors after `bun run build:core`; Biome clean on 2,605 files
- Note for anyone re-running: this suite imports `bun:test`, so it needs `bun test`; `vitest` reports "No test files found" and exits 1

## Not verified

I did not run the full `packages/cloud/shared` suite or the migration tests, and no request reached real Stripe, so this confirms the contract the code enforces against the SDK's parsed shapes rather than Stripe's live behavior. I also did not verify the webhook `event` path, which returns before this validation and was not changed here.

— [claude-code-eliza]

nothingxnowhere (approved):
Reviewed at head `894d76d861` against `develop`, with the lockfile's `stripe@22.5.0`.

At head the file is `9 pass / 0 fail`. With the PR's tests kept and `outreachr-billing.ts` reverted to develop, `4 fail / 5 pass`: the four failures are exactly the rejection cases (transformed checkout price, transformed subscription price, the `[undefined, 0, -1, 1.5, 1001]` quantity sweep, and a transformed current plan before any portal configuration), while both positive cases — the licensed per-unit subscription reading `seats: 3`, and the four-seat `astra` portal upgrade retaining `adjustable_quantity.minimum = 3` and the confirm-flow items — already pass on develop. So the happy path is unchanged and the delta is purely the newly refused shapes, which is what a contract-tightening should look like.

The vulnerability reads correctly from develop: the old assertion checked currency, amount, interval and interval count but never `billing_scheme`, `transform_quantity`, `type` or `usage_type`, so a `$49` price with `transform_quantity: { divide_by: 10 }` cleared every check while Stripe billed three selected seats as one transformed unit. `assertSeatPrice` closes that, and splitting `active` out of it is right: an existing subscription on a since-retired price stays readable while a new checkout must use an active one.

One widening the body does not mention, which I think is correct but should be a known choice: `subscriptionStates` is now computed before the action dispatch, so every action — `checkout` included, not only `portal` — refuses a customer whose existing subscription fails the seat contract or carries a non-integer or out-of-range quantity. On develop `checkout` only checked `matches(sub.metadata, workspaceId)` on existing subscriptions. That is fail-closed in the direction billing wants, and a customer with no subscriptions is unaffected (empty map), but it does mean a malformed legacy subscription now blocks new checkout as well.

Approving.

```json
{
  "schemaVersion": "2",
  "headSha": "894d76d861",
  "recommendation": "accept",
  "reproduced": true,
  "risks": "none",
  "effortBand": "medium",
  "complexity": "moderate",
  "impact": "high",
  "reviewLoad": "standard",
  "recommendedTier": "small",
  "recommendedThirds": 3,
  "workUnitId": "wu_elizaos_eliza_pr_30615",
  "valueRationale": "Ran the suite with the pinned stripe SDK at head (9 pass) and against develop's outreachr-billing.ts (4 fail / 5 pass), showing the positive paths were already green and only the four refusals are new; noted the checkout-path widening from hoisting subscriptionStates.",
  "usefulArtifacts": ["RED/GREEN run of outreachr-billing.test.ts with stripe@22.5.0"],
  "commands": ["bun test --conditions=eliza-source src/lib/services/outreachr-billing.test.ts   # head: 9 pass; develop source + PR tests: 4 fail / 5 pass"],
  "evidenceUrls": [],
  "summary": "Per-seat Stripe price contract now checks billing scheme, quantity transform, type and usage type, plus integer seat quantities; verified only the four refusal cases are new and the positive paths already held, and flagged that checkout now also validates existing subscriptions."
}
```

— [nxn-claude-code]

byt61 (approved):
Approved. Verified by executing at PR head in a detached worktree (module postdates my checkout, so overlay was impossible):

- Full suite: 9/9 pass under `bun test`.
- Mutant check (PR tests vs pre-PR source): exactly the 4 new rejection tests fail (transform_quantity on checkout/subscriptions/portal, invalid quantities); the 2 new positive tests pass on both, as expected for behavior-preserving cases.
- Read the source diff: `assertSeatPrice` correctly short-circuits on `recurring == null` before touching `interval_count`; `!sub.items.has_more` closes the multi-page hole; error translation stays fail-closed with `OUTREACHR_BILLING_SCOPE`.

One behavior note (not blocking): subscription-state validation now runs for every action, so a legacy non-conforming subscription will also fail checkout/portal, not just the subscriptions read. That matches the fail-closed billing invariant, but flagging in case grandfathering was intended.

PR #30616, opened 2026-09-05
fix(agent): preserve complete coding evidence and contacts availability

### LINKED ISSUE #30571: fix(contacts): report unsupported web address book instead of empty success
The off-Android ContactsWeb fallback reports granted contacts permission and returns an empty address book despite having no supported contacts store. When plugin-contacts loads on web/Node, its actual androidContacts provider therefore injects a valid-looking count-zero address book instead of the existing explicit contactsError/unavailable result.

Affected boundary: plugins/plugin-native-contacts/src/web.ts -> plugins/plugin-contacts/src/providers/contacts.ts. The consumer guide says listContacts should throw off Android, while the native guide/README and consumer README document the fabricated empty success.

Scope: preserve input validation and Android behavior; make unsupported off-Android list and permission operations truthful and align documentation/platform metadata. No contact writes, Android Kotlin changes or additional contacts collection are required.

Acceptance: the actual registered Capacitor web implementation rejects unsupported reads and never reports a usable grant; the real provider emits unavailable/error planning context and reports the failure. A legitimate successful empty Android list remains distinct. Both package suites/typechecks/lints, guide parity and root verify pass; real web/Node boundary evidence is inspected.

### LINKED ISSUE #30583: fix(orchestrator): retain exact and complete completion-evidence paths
ACP completion evidence loses explicit write paths after the first 500, and baseline subtraction can remove a different whitespace-named file while retaining a baseline file's patch.

The affected production path is AcpService tool events -> getChangedPaths -> captureChangeSet -> OrchestratorTaskService completion evidence. recordEditedPaths caps the session set at 500 and trims filesystem path bytes. subtractChangeSetBaseline trims baseline paths and splits Git headers with a whitespace-sensitive regex; Git quotes tabs, newlines and Unicode, and spaces are legitimate path bytes.

Acceptance:
- Retain all explicitly written session paths, including paths after entry 500, with exact whitespace and Unicode.
- Subtract only exact baseline paths, keeping the complete remaining file list and patches consistent across Git quoting, additions, deletions, renames and binary files.
- Keep persisted change-set compatibility deliberately; never silently produce partial evidence when a patch cannot be attributed.
- Prove the capture/completion contract with real temporary Git repositories and a real filesystem consumer of more than 500 tool paths; run owning package checks.

Scope is the ACP completion evidence path. No subprocess, scheduler or lock redesign.

lalalune:
Agent context now distinguishes an unavailable address book from a valid empty result and preserves complete coding-completion evidence.

Unsupported device-contact platforms previously returned an empty address book and granted permission, while writes threw generic errors. Their registered Capacitor fallback now rejects valid permission, read and mutation calls with UNAVAILABLE. The contacts provider reports the unavailable capability explicitly. Android's implemented native address-book path remains unchanged.

ACP stopped collecting explicit write paths at 500 and trimmed valid filenames. It now retains every exact path, including whitespace and Unicode, and carries per-file patch ownership through capture and baseline filtering. Completion counts reflect the retained files. Legacy patch ownership uses complete Git paths and explicit rename/copy destinations. Incomplete saved inventories require recapture instead of attributing a longer path to a suffix filename. Existing import paths and aggregate patches remain compatible.

Closes #30571.
Closes #30583.

Validation: the implementation audit ran both complete contact package lanes and the full orchestrator lane, followed by focused and serial reruns for the identified build/timing failures. The scoped build and both contacts suites pass. The six focused orchestrator files initially passed 160/162 cases; the corrected router file then passed all 100 cases, and the large scaffold case passed in isolation within its unchanged explicit 20-second deadline. Root verification passed all 376 tasks and all repository audits at eb68bbd1802594a8a0f78c0a41dbe06c59a7ca4c. The branch then integrated develop at 182c3ebad2b90276e132c0e01f65984bc19eb992 with all 13 changed files byte-identical. Both real-provider proofs passed again on d6c8a90b6dda5d6f70209c984f3f23f48ff27a1b; final root verification passed all 376 tasks and every repository audit on the integrated head. Positive completion tests now await actual delivery instead of fixed sleeps; negative observation windows remain intact. Under a measured host load near 255 on 16 cores, the local filesystem correctness rerun uses a 60-second runner deadline; repository and production deadlines are unchanged, and the original timeout/failure log is retained. The address-book proof uses actual Capacitor registration and the real provider; the coding proof passes 505 actual ACP write events and complete filesystem evidence to a live model, which must return the final five exact paths and contents.

A follow-up to review reproduced a suffix collision with actual Git output: an incomplete saved inventory could map `nested b/bar` to `bar`. The correction matches complete ordinary headers and explicit rename/copy destination records. All 48 workspace-diff cases pass, including persisted-inventory, literal pattern characters and directory ownership regressions, plus quoted, whitespace, Unicode, binary, deleted and renamed paths. Per-file selection excludes descendants, so a directory never owns a child's patch. The displayed summary is now derived from that same retained inventory, replacing the second Git shortstat query; full patch content remains complete. The internal Git subprocess clears inherited literal/glob/noglob/icase pathspec modes so they cannot reinterpret capture selectors. The final provider proof explicitly verifies complete file bytes including line endings; an earlier attempt returned trailing newlines while its ambiguous harness expected marker-only values, and that failed attempt is retained. The branch is based on develop at dbeba62bfb7d98b75a52f5f6202d9d81eae4b09d. Root verification passed all 376 tasks and all repository audits again on 04400bf6243a8821e03f4688347687faeb670a2f; actual contacts and ACP provider proofs plus persisted Git ownership readback also passed on this revision.

## Evidence Gate

N/A - no rendered component or visual styling changed.

N/A - no rendered component or visual styling changed.

N/A - provider trajectories and filesystem/provider readback show the changed behavior.

[Final correction tests, real provider and full root verification logs](https://github.com/elizaOS/eliza/releases/download/pr-evidence-10/30616-agent-context-final-validation-04400bf6.log). [Retained intermediate provider assertion and build interruption](https://github.com/elizaOS/eliza/releases/download/pr-evidence-10/30616-agent-context-final-history-04400bf6.log). Historical validation and retained failed attempts: [previous complete checks](https://github.com/elizaOS/eliza/releases/download/pr-evidence-10/30616-agent-context-validation-d6c8a90b.log), [earlier attempts](https://github.com/elizaOS/eliza/releases/download/pr-evidence-10/30616-agent-context-validation-history-d6c8a90b.log).

N/A - no frontend request or page flow changed.

[Actual contacts request and response](https://github.com/elizaOS/eliza/releases/download/pr-evidence-10/30616-agent-context-contacts-04400bf6.json). [Actual ACP request and response with complete context](https://github.com/elizaOS/eliza/releases/download/pr-evidence-10/30616-agent-context-acp-04400bf6.json). Both use Cerebras gemma-4-31b on 04400bf6243a8821e03f4688347687faeb670a2f.

[Complete 505-file filesystem capture](https://github.com/elizaOS/eliza/releases/download/pr-evidence-10/30616-agent-context-filesystem-04400bf6.json). [Real persisted Git ownership readback and invalid-inventory rejection](https://github.com/elizaOS/eliza/releases/download/pr-evidence-10/30616-agent-context-ownership-04400bf6.json). [Rebase source identity and the subsequent ownership corrections](https://github.com/elizaOS/eliza/releases/download/pr-evidence-10/30616-agent-context-source-04400bf6.json).

N/A - no rendered component or visual styling changed.

Risk: unsupported contacts and ambiguous legacy coding captures now report explicit errors where callers previously received misleading context. Evidence uses isolated files and actual provider calls; no real address book is mutated. Hosted checks must pass on the current head before merge. Maintainer CI exception: N/A.

### DIFF
--- a/plugins/plugin-agent-orchestrator/__tests__/unit/sub-agent-router.test.ts
+++ b/plugins/plugin-agent-orchestrator/__tests__/unit/sub-agent-router.test.ts
@@ -1720,10 +1720,11 @@ describe("SubAgentRouter", () => {
         acp.emit(SESSION_ID, "task_complete", {
           response: `${localStyle}\n${publicScript}`,
         });
-        await new Promise((r) => setTimeout(r, 200));
+        await vi.waitFor(() => {
+          expect(handleMessage).toHaveBeenCalledTimes(1);
+        });
 
         expect(spawnSession).not.toHaveBeenCalled();
-        expect(handleMessage).toHaveBeenCalledTimes(1);
         const posted = handleMessage.mock.calls[0]?.[1];
         expect(posted?.content?.text).not.toContain(localPage);
         expect(posted?.content?.text).toContain(publicPage);
@@ -1940,11 +1941,12 @@ describe("SubAgentRouter", () => {
         acp.emit(SESSION_ID, "task_complete", {
           response: `Wrote files under apps/random-tweet-generator/. Public URL ${appUrl}`,
         });
-        await new Promise((r) => setTimeout(r, 200));
+        await vi.waitFor(() => {
+          expect(handleMessage).toHaveBeenCalledTimes(1);
+        });
 
         // No spurious verify-retry, and the completion turn is posted as-is.
         expect(spawnSession).not.toHaveBeenCalled();
-        expect(handleMessage).toHaveBeenCalledTimes(1);
         const posted = handleMessage.mock.calls[0]?.[1];
         expect(posted?.content?.text).not.toContain(
           "not updated during this session",
@@ -2001,10 +2003,11 @@ describe("SubAgentRouter", () => {
         acp.emit(SESSION_ID, "task_complete", {
           response: `Done — live at ${appUrl}`,
         });
-        await new Promise((r) => setTimeout(r, 200));
+        await vi.waitFor(() => {
+          expect(handleMessage).toHaveBeenCalledTimes(1);
+        });
 
         expect(spawnSession).not.toHaveBeenCalled();
-        expect(handleMessage).toHaveBeenCalledTimes(1);
         const posted = handleMessage.mock.calls[0]?.[1];
         expect(posted?.content?.text).not.toContain("[verification:");
       } finally {
@@ -2194,13 +2197,14 @@ describe("SubAgentRouter", () => {
         acp.emit(SESSION_ID, "task_complete", {
           response: localUrl,
         });
-        await new Promise((r) => setTimeout(r, 200));
+        await vi.waitFor(() => {
+          expect(handleMessage).toHaveBeenCalledTimes(1);
+        });
 
         const fetched = fetchMock.mock.calls.map(([url]) => String(url));
         expect(fetched).toContain(localUrl);
         expect(fetched).toContain(publicUrl);
         expect(spawnSession).not.toHaveBeenCalled();
-        expect(handleMessage).toHaveBeenCalledTimes(1);
         const posted = handleMessage.mock.calls[0]?.[1];
         expect(posted?.content?.metadata?.subAgentVerifiedUrls).toEqual([
           publicUrl,
@@ -2379,12 +2383,13 @@ describe("SubAgentRouter", () => {
       acp.emit(SESSION_ID, "task_complete", {
         response: `The cached URL ${staleUrl} is stale; the app now uses ${freshUrl}`,
       });
-      await new Promise((r) => setTimeout(r, 200));
+      await vi.waitFor(() => {
+        expect(handleMessage).toHaveBeenCalledTimes(1);
+      });
 
       const fetched = fetchMock.mock.calls.map(([url]) => String(url));
       expect(fetched).toEqual([freshUrl]);
       expect(spawnSession).not.toHaveBeenCalled();
-      expect(handleMessage).toHaveBeenCalledTimes(1);
       const posted = handleMessage.mock.calls[0]?.[1];
       expect(posted?.content?.text).not.toContain("[verification:");
     });
--- a/plugins/plugin-agent-orchestrator/__tests__/unit/sub-agent-router.test.ts
+++ b/plugins/plugin-agent-orchestrator/__tests__/unit/sub-agent-router.test.ts
@@ -2792,7 +2792,9 @@ describe("SubAgentRouter — change-set narration (GAP C)", () => {
         response:
           "[tool output: Read index.html]\n<h1>placeholder</h1>\n[/tool output]\nSearch for the image element.",
       });
-      await vi.waitFor(() => expect(handleMessage).toHaveBeenCalled());
+      await vi.waitFor(() => expect(handleMessage).toHaveBeenCalled(), {
+        timeout: 3_000,
+      });
 
       const posted = handleMessage.mock.calls[0]?.[1];
       const text = String(posted?.content?.text ?? "");
--- a/plugins/plugin-agent-orchestrator/__tests__/unit/sub-agent-router.test.ts
+++ b/plugins/plugin-agent-orchestrator/__tests__/unit/sub-agent-router.test.ts
@@ -1314,9 +1314,10 @@ describe("SubAgentRouter", () => {
       acp.emit(SESSION_ID, "task_complete", {
         response: `Done — the app is live at ${DEAD_URL}`,
       });
-      await new Promise((r) => setTimeout(r, 1000));
 
-      expect(spawnSession).toHaveBeenCalledTimes(1);
+      await vi.waitFor(() => {
+        expect(spawnSession).toHaveBeenCalledTimes(1);
+      });
       const arg = spawnSession.mock.calls[0]?.[0] as {
         initialTask?: string;
         metadata?: Record<string, unknown>;
@@ -1342,10 +1343,13 @@ describe("SubAgentRouter", () => {
       acp.emit(SESSION_ID, "task_complete", {
         response: `Done — local check failed at ${DEAD_URL}`,
       });
-      await new Promise((r) => setTimeout(r, 200));
+
+      await vi.waitFor(() => {
+        expect(handleMessage).toHaveBeenCalledTimes(1);
+      });
 
       expect(spawnSession).not.toHaveBeenCalled();
-      expect(handleMessage).toHaveBeenCalledTimes(1);
+
       const posted = handleMessage.mock.calls[0]?.[1];
       expect(posted?.content?.text).not.toContain("127.0.0.1");
       expect(posted?.content?.text).not.toContain("localhost");
@@ -1363,12 +1367,14 @@ describe("SubAgentRouter", () => {
       acp.emit(SESSION_ID, "task_complete", {
         response: `Done — the app is live at ${DEAD_URL}`,
       });
-      await new Promise((r) => setTimeout(r, 1000));
+      await vi.waitFor(() => {
+        expect(spawnSession).toHaveBeenCalledTimes(1);
+      });
       acp.emit(SESSION_ID, "error", {
         message: '"Method not found": session/cancel',
       });
-      await new Promise((r) => setTimeout(r, 200));
 
+      await new Promise((r) => setTimeout(r, 200));
       expect(spawnSession).toHaveBeenCalledTimes(1);
       expect(handleMessage).not.toHaveBeenCalled();
     });
@@ -1421,9 +1427,10 @@ describe("SubAgentRouter", () => {
       acp.emit(SESSION_ID, "error", {
         message: "build failed: GET /api returned 405 method not found",
       });
-      await new Promise((r) => setTimeout(r, 200));
 
-      expect(handleMessage).toHaveBeenCalled();
+      await vi.waitFor(() => {
+        expect(handleMessage).toHaveBeenCalled();
+      });
     });
 
     it("does surface a build error mentioning a non-ACP path like fs/promises", async () => {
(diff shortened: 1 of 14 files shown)


ss251 (approved):
d6c8a90b6dda5d6f70209c984f3f23f48ff27a1b
Static review of the diff at this head — fixes two real evidence bugs, not cosmetic.

Input ` spaced.txt ` / `tab\tname` via tool_call -> wrong output dropped/misattributed evidence. `plugins/plugin-agent-orchestrator/src/services/acp-service.ts` removed `.trim()` and the `set.size>=500 break` cap, now stores exact `value`/`location.path` bytes unbounded. This preserves complete coding evidence: the new `completion-path-evidence.test.ts` writes 505 artifacts including ` spaced.txt `, `tab\tname`, `line\nname`, `snow☃`, ` ` and asserts `changedFiles` equals exact names and rendered body contains each path plus `+CONTENT_END`.

Input shared workdir dirty files + special filenames -> wrong output contradictory changeset. `plugins/plugin-agent-orchestrator/src/services/workspace-diff.ts` adds `fileDiffs:{path,diff}` ownership, `joinFileDiffs`, `quotedGitPath`, and `readFileDiffs` that fail-closed with `WORKSPACE_CHANGESET_PATCH_OWNERSHIP_INVALID` instead of regex `diff --git a/(\S+)` suffix guessing that broke on spaces/quotes/unicode. `subtractChangeSetBaseline` now subtracts exact baseline paths and their patches only; legacy `fileDiffs===undefined` Git text is matched against authoritative inventory with both quotePath variants, throwing on ambiguous/unowned patches. Neighbor test ` x ` vs `x` stays distinct.

Input web contacts read -> wrong output successful empty address book. `plugins/plugin-native-contacts/src/web.ts` now `throw this.unavailable(...)` for `listContacts/create/import/check/requestPermissions` with code `UNAVAILABLE`, and `plugins/plugin-contacts` provider/view propagate error instead of `contactsAvailable=false` with no error. Docs updated consistently. `sub-agent-router.test.ts` sleep->`vi.waitFor` conversions reduce flake, not weakening; assertions unchanged.

Adversarial: no binaries, symlinks, lockfiles, CI perms, secrets, telemetry. Minor note: changed-path set is now unbounded and `diffStat` becomes JSON list — acceptable for correctness.
Verified by reading the diff at this head and tracing path capture through render/subtract and web fallback through provider; no sandbox applies to docs/test-only plus TS logic change.
Recommendation: merge

— [muse-lane-ss251]

attentionhead (approved):
Approving at `d6c8a90b6dda5d6f70209c984f3f23f48ff27a1b`, with one coverage gap worth closing that I do not think should block.

Both halves fix the same class of bug — code that answered confidently when it should have said "I don't know" — and in both cases the old behavior was worse than an error because it was indistinguishable from a legitimate result.

## Coding evidence

The two `acp-service.ts` changes are each load-bearing:

```
restore `if (set.size >= 500) break`   → × renders every tool write after the first 500, with exact path bytes
restore `.trim()` on captured paths    → × (same test)
```

The 500 cap is the more serious of the two, because nothing recorded that truncation had happened — the completion evidence simply stopped at 500 paths and read as complete. Dropping `.trim()` matters for a narrower but real case: a filename with leading or trailing whitespace is a valid path, and trimming it produced a path that does not exist.

The `fileDiffs` restructuring is the right fix rather than a patch on the symptom. `develop` reconstructed per-file ownership by re-parsing its own concatenated output with `/^diff --git a\/(\S+) b\//`, and `\S+` stops at the first space, so any baseline path containing whitespace failed to match — whereupon `if (!match) return true` **kept** the hunk. A baseline file's patch therefore leaked into the session's completion evidence, which is exactly the failure the docstring describes chasing. Carrying ownership structurally at capture time removes the parse entirely for new captures, and the `joinFileDiffs(entries) !== changeSet.diff` equality check is a good consistency guard against a persisted capture whose two representations disagree.

## The gap: the ambiguity branch is untested

In the legacy path, `if (paths.length !== 1) throw invalidPatchOwnership()` is what implements "require recapture instead of guessing". I weakened it to `paths.length === 0`, so an ambiguous header silently takes the first candidate — the precise behavior the change exists to prevent — and **all 104 tests still pass**, including the pre-existing `changeset-baseline.test.ts` that this PR does not touch.

I want to be fair about severity. I could not construct a real collision: the candidates are matched with `endsWith(\` ${encoded}\`)` against three encodings, and for two distinct repo-relative paths to both match one header you need a contrived shape. So this reads as genuinely defensive code, not a live hole, which is why I am not blocking. But it is the one branch where the new fail-closed contract could regress to guessing without any test noticing, so a case with two paths that both suffix-match a single legacy header would be worth adding.

The zero-match half of that condition is also worth a case: an unowned patch is the more likely real-world shape, since it only takes a persisted capture whose `changedFiles` no longer covers its own diff.

## Contacts

Replacing `checkPermissions` returning `{ contacts: "granted" }` on web with an explicit `unavailable()` is correct, and reporting a granted permission for a store that does not exist was the worst part of the old behavior — it invited callers into a read that could only ever return nothing.

I checked the callers, since this converts "silently empty" into "throws". Both view call sites already do `requestPermissions().catch(() => null)` and deliberately fall through, with a comment saying `listContacts` is expected to surface the real error. And `providers/contacts.ts` — untouched by this PR — already wraps the read in a catch that reports through `contactsError` and documents "never a fabricated empty contact list". That existing handler is precisely why no provider change was needed: the web path now reaches the branch that was already written for it, instead of returning `count: 0` as though the address book were genuinely empty.

## Verification

- `completion-path-evidence.test.ts` 4 pass; 1 fail under each of the two `acp-service` mutations
- `sub-agent-router.test.ts` 100 pass; pre-existing `changeset-baseline.test.ts` 3 pass (no regression)
- `plugin-contacts` 41 pass, `plugin-native-contacts` 9 pass
- Typecheck 0 errors across all three plugins after `bun run build:core`; Biome clean on 451 files (1 info)

## Not verified

I did not run the full orchestrator lane, the live-model coding proof, or the real-Capacitor address-book proof described in the evidence — my contacts result is the unit and contract suites only. I also did not exercise a genuinely persisted legacy change set from disk; my reading of the legacy path is from the code and the tests that construct change sets in memory.

— [claude-code-eliza]

byt61 (approved):
Approved. Verified locally via overlay on a clean checkout (package vitest runners):

- `plugin-native-contacts/src/web.test.ts`: 9/9 pass.
- `plugin-contacts` contract + new `contacts.web.test.ts`: 5/5 pass, running against the overlaid new bridge through the src alias — a true end-to-end of the contacts half.
- New `completion-path-evidence.test.ts`: 4/4 pass (500+ exact-path writes incl. whitespace/unicode names, baseline subtraction with per-file ownership).
- Existing workspace-diff suites: 34/34 pass at PR head. (An initial run mixed PR source with my checkout's newer test files and showed 2 `parseLsFiles` failures; I confirmed by diffing that my branch carries a newer drop-tail implementation the PR predates, then re-ran with PR-head tests — all green. Not a PR regression.)
- Static checks: `throw this.unavailable(...)` is correct usage (`WebPlugin.unavailable` in the installed @capacitor/core *returns* a `Capacitor.Exception` with code `UNAVAILABLE`); no behavioral capacitor-contacts consumers exist outside the two updated plugins; `AGENTS.md`/`CLAUDE.md` byte-identical; `subtractChangeSetBaseline` only validates patch ownership when subtraction actually removes files, and the legacy path throws rather than silently dropping unattributable hunks — fail-closed throughout.

lalalune:
Independent review of `d6c8a90b6dda5d6f70209c984f3f23f48ff27a1b` is complete. The changes fix two concrete consumer failures: unsupported contacts appeared empty, and ACP evidence lost paths after 500 or changed valid filename bytes. Reviewed all 13 changed files, baseline subtraction callers, and exact per-file patch ownership/legacy rejection. The router test changes wait for actual delivery and retain negative observation windows.

After a pinned dependency install, independent runs passed [Contacts: 41 tests](https://github.com/elizaOS/eliza/releases/download/pr-evidence-10/30616-review-contacts-d6c8a90b.log), [native fallback: 9 tests](https://github.com/elizaOS/eliza/releases/download/pr-evidence-10/30616-review-native-contacts-d6c8a90b.log), and [orchestrator: 129 tests](https://github.com/elizaOS/eliza/releases/download/pr-evidence-10/30616-review-orchestrator-d6c8a90b.log). Inspected the attached final-head root log: all 376 tasks and repository audits passed. The latest hosted run 33953518008 has all five checks successful; the earlier run was cancelled during ready transition.

Independently validated the attached provider records against the actual complete 505-file capture: every early file and all five late exact paths/content are present in the dispatched prompt; the recorded successful Cerebras response matches the parsed expected result. The real Capacitor/provider contacts record yields unavailable with a null count. These prove the changed unsupported-platform and context paths; they do not claim an Android device session. No remaining defect found in this PR's scope. Merging the exact reviewed head into develop.

lalalune:
The exact-head merge was refused because a newly merged develop change overlaps the router test. Rebased the complete PR onto `e7e7fb0acf3e81c5a376c30cbb26d2fb3547a117`, preserving the outcome-based wait. Local candidate is `5bea56c94f3c2cd2df756702b5f02e1e735ae350`; all reviewed PR files, including the final router test, are byte-identical to `d6c8a90`. Pinned install, full orchestrator suite and root verification are running before a SHA-leased branch update. No merge has occurred.

lalalune:
Independent review now covers the final ownership correction at `04400bf6243a8821e03f4688347687faeb670a2f`. All changed production paths were reviewed; the latest real-Git ownership suites passed 36 cases locally. The earlier byte-identical core of this PR also passed my full 3,138-test orchestrator run. I downloaded and checked the complete new 505-file filesystem capture, actual HTTP request/response correspondence, exact late-file bytes including line endings, unsupported contacts result, and persisted ownership rejection. The attached final root log records all 376 tasks and repository audits passing.

[Independent provider/artifact review receipt](https://github.com/elizaOS/eliza/releases/download/pr-evidence-10/30616-independent-provider-review-04400bf6.json) · [Final ownership tests](https://github.com/elizaOS/eliza/releases/download/pr-evidence-10/30616-review-ownership-04400bf6.log). The old candidate was not pushed: the exact lease prevented overwriting this newer correction. Final hosted checks are still required before merge.

PR #30617, opened 2026-09-05
fix(cloud-sdk): harden successful response contracts

wtfsayo:
Related to #30574; follow-up to #30597.

Harden the cloud SDK success-response contract: validate JSON media types and UTF-8, expose bodyless responses as `T | undefined`, and provide `requestData` for callers that require a DTO. Generated routes distinguish JSON, optional JSON and raw responses; plugin consumers use the appropriate contract.

Review found that generic mutation helpers incorrectly rejected successful 204/205 responses. Repaired `post`, `put`, `patch`, `delete` and `postUnauthenticated` to preserve the generic contract, while migrating data-required high-level and plugin callers to `requestData`. Real HTTP regressions fail on the original implementation and pass with the repair. Updated generated routes against the rebased API inventory.

Validation on `298bf89036cb2123baba8393a930fa97ac1e6c59`, based on `9612ff217a00465f63f1854b53ebcd65e2461dcd`, Bun 1.3.14 / Node 24.15.0:
- SDK: 191 passed, 19 expected credential/live skips; generated route audit passed.
- Plugin: 631 passed, 5 skipped; packaged distribution checks: 5 passed.
- Real HTTP mutation regression: 2 failures before repair; complete HTTP server suite: 17 passed after repair.
- Live public API: model listing and CLI login bootstrap/poll both passed. No paid generation or resource provisioning was performed.
- SDK and plugin typecheck, lint and builds passed.
- `bun install` and root `bun run verify` passed: all 376 workspace tasks and repository audits.

Risk: public generic response types now accurately include `undefined`; consumers requiring a body must use the strict helper. Raw-response APIs and the bound legacy custom-transport fallback remain supported.

UI screenshots, video, native captures and model trajectories are N/A: this is a transport/response-contract change with real HTTP and live public API evidence. Credentialed SDK cases remain explicitly skipped.

### DIFF
--- a/packages/cloud/sdk/README.md
+++ b/packages/cloud/sdk/README.md
@@ -103,6 +103,28 @@ binary, and text routes return `Response` from the primary generated method;
 mixed routes such as chat completions keep the JSON method and use `Raw` when
 the request asks for streaming.
 
+Parsed `request()` calls require an explicit JSON media type
+(`application/json` or a structured `+json` type) and preserve every JSON
+value, including primitives and `null`. Successful text, missing media types,
+malformed JSON, and unexpectedly empty data responses throw `CloudApiError`;
+use `requestRaw()` for text or binary bodies. `HEAD`, `204`, and `205`
+responses are the deliberate bodyless exceptions and resolve to `undefined`.
+The generic `get`, `post`, `put`, `patch`, `delete`, and
+`postUnauthenticated` helpers preserve this behavior. Use `requestData` when
+a caller requires JSON data even on a successful status.
+Generated endpoints that deliberately return either JSON or `204` expose the
+same `T | undefined` result, while data-required helpers reject a bodyless
+response instead of hiding it behind their DTO type.
+Older SDK releases could replace successful text or empty bodies with an
+invented `{ success: true }` object; callers relying on that fallback must use
+an explicit bodyless status or return a JSON response instead.
+
+`pollJob` and `waitForCliLogin` enforce a total timeout through every request,
+response-body read, and polling interval. Their timeout and interval options
+accept integers from 0 through 2,147,483,647 milliseconds; zero timeout expires
+immediately. Login cancellation also interrupts an in-flight request or wait.
+Direct `getJob` and `pollCliLogin` calls accept optional `timeoutMs` and `signal`.
+
 Refresh and verify route coverage after adding or changing API routes:
 
 ```bash
@@ -135,19 +157,3 @@ Build and publish:
 bun run build
 npm publish --access public
 ```
-
-## Parsed responses
-
-`request` and typed endpoint methods accept `application/json` and application
-media types ending in `+json`, preserving objects, arrays and JSON primitives.
-Successful HTML/text, malformed JSON and unexpectedly empty JSON responses
-throw `CloudApiError`; they never become an invented `{ success: true }` DTO.
-HEAD and HTTP 204/205 responses resolve to `undefined`, as these protocols have
-no response body. Use `requestRaw` (or a generated `Raw` method) for text,
-binary, streaming, or status-sensitive responses.
-
-`pollJob` and `waitForCliLogin` enforce a total timeout through every request,
-response-body read and polling interval. Their timeout and interval options
-accept integers from 0 through 2,147,483,647 milliseconds; zero timeout expires
-immediately. Login cancellation also interrupts an in-flight request or wait.
-Direct `getJob` and `pollCliLogin` calls accept optional `timeoutMs` and `signal`.
--- a/packages/cloud/sdk/scripts/generate-public-routes.mjs
+++ b/packages/cloud/sdk/scripts/generate-public-routes.mjs
@@ -77,6 +77,14 @@ function pathParamTypeLine(endpoint) {
   return `  ${quote(endpoint.key)}: { ${fields} };`;
 }
 
+const JSON_OR_EMPTY_ROUTES = new Set([
+  "DELETE /api/v1/apis/storage/objects/_",
+  "GET /api/v1/advertising/conversions/track",
+  "GET /api/v1/marketing/inventory/serve",
+  "GET /api/v1/remote/sessions/{id}/commands",
+  "POST /api/v1/twilio/voice/status",
+]);
+
 function responseModeFor(method, route, source) {
   if (
     route === "/api/v1/apis/storage/objects/_" &&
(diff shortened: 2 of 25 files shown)


ss251 (approved):
Reviewed head 6ae02f77a98ef9b356460cef8146d1286601c91f.

The risky part is `packages/cloud/sdk/src/http.ts:106` `isJsonMediaType` plus the `requestParsed` bodyless branch. I followed the media-type boundaries the new integration test pins: `*/problem+json` and `application/*+json` fail `MEDIA_TYPE_RESTRICTED_NAME` so become `kind:text` -> `unexpected_response_content_type`, while `application/problem+json; charset=utf-8` is split on `;`, trimmed/lowercased, and matches the `+json` suffix so decodes as JSON. I also traced the `HEAD/204/205` early return: it runs before `parseResponseBody`, cancels the detached body, returns `undefined` when `allowBodyless` is true (`request`) but throws `empty_response_body` for `requestData`, which is what makes the `CloudResponse<T>` vs data-required split sound.

No blocking concerns; non-blocking nit: legacy-transport fallback in `call()` throws plain `Error`.
Recommendation: merge

— [muse-lane-ss251]

attentionhead (approved):
Approving at `6ae02f77a98ef9b356460cef8146d1286601c91f`.

The central move here is making an existing runtime behavior honest rather than changing behavior, and that distinction is what makes a 23-file contract change safe to land. On `develop`, `request()` already returned `undefined` for HEAD/204/205 — it just did so as `return undefined as TResponse`, lying to every caller. This PR keeps that runtime result and stops lying about it.

## The breaking-type question

`request<T>` goes from `Promise<T>` to `Promise<CloudResponse<T>>`, and `get<T>` widens with it while `post`/`put`/`patch`/`delete` stay `Promise<T>`. That asymmetry is right — GET is the verb that can legitimately answer 204 — but it is a breaking signature change for a package with a wide consumer surface, and this PR only updates `plugin-elizacloud`. `packages/ui`, `plugin-cloud-apps`, `packages/cloud/shared` and `packages/cloud/test-mocks` all depend on `@elizaos/cloud-sdk` and are untouched.

Typechecking those consumers at this head:

```
plugins/plugin-cloud-apps   0 errors
packages/ui                 0 errors
plugins/plugin-elizacloud   0 errors
```

I did not take that at face value, because a 0-error result would be worthless if TypeScript were resolving `@elizaos/cloud-sdk` to stale `dist/*.d.ts` from before the change — that is the usual way this kind of check lies. The emitted declarations carry the new signature:

```
dist/http.d.ts:37: request<TResponse>(...): Promise<CloudResponse<TResponse>>;
```

so the consumers really were checked against the widened contract and really do survive it.

## Strict-path rejections

Mutating the newest of the four rejection behaviors back to `develop`'s lenient form — `new TextDecoder("utf-8", { fatal: true })` → `new TextDecoder()` — kills exactly the intended test and nothing else:

```
× rejects malformed UTF-8 instead of substituting JSON string data
Tests  1 failed | 22 passed (23)
```

That is the right assertion to have written: a lenient decoder substitutes U+FFFD and hands back a plausible-looking string, which is precisely the "fabricated success" this PR is trying to eliminate.

## Two things I checked that look like regressions but are not

**The dropped `isMalformedJsonBody` branch in `normalizeErrorBody`.** Removing it looks like it would lose the raw response text from error messages. It does not: `parsedBodyValue` maps `malformed-json` to `body.text`, and the existing string fallback produces `HTTP ${status}: ${body}` — byte-identical to what the deleted branch produced. Pure simplification.

**The `isJsonMediaType` rewrite.** The new parser is stricter about media-type syntax but slightly *wider* on structured suffixes: `develop` matched `^application\/[\w.+-]+\+json$`, so only `application/*+json`, while this accepts any type with a `+json` subtype. That is more correct per RFC 6839 and I could not construct a content type that `develop` treated as JSON and this rejects.

## Test reachability

The new 295-line integration test imports from `vitest`, but this package's declared command is `bun test src`, and mixed runners in this repo usually mean a file silently does not execute. I checked instead of assuming — both runners execute it in full:

```
bun test src/http-success-response.integration.test.ts   23 pass, 0 fail
bun test src                                             208 tests, 0 fail (19 pre-existing skips)
bunx vitest run (3 changed suites)                        59 pass
```

## Verification

- SDK: 208 tests under `bun test src`, 59 under vitest across the three changed suites
- `plugin-elizacloud`: 57 pass across all five changed suites
- `node scripts/audit-api-routes.mjs` — exit 0
- Consumer typechecks 0 errors, confirmed against non-stale declarations
- Biome clean on 48 files

## Not verified

I did not run `packages/ui` or `plugin-cloud-apps` *tests*, only their typechecks — so I am claiming those consumers still compile against the new contract, not that their runtime behavior is unchanged. I did not exercise the legacy custom-transport fallback or the live e2e lane, and I did not re-run the generator to confirm `public-routes.ts` is reproducible from `generate-public-routes.mjs`.

— [claude-code-eliza]

byt61 (approved):

- SDK suites under their own runners: 36 pass / 0 fail across `http.test.ts` + `public-routes.test.ts` (`bun test`).
- The new 23-test localhost integration suite cannot bind sockets on this machine (`listen EPERM` on 127.0.0.1 is machine-wide — a bare node http server fails identically), so I executed its core strict paths without sockets instead, via a throwaway injected-fetch probe against the overlaid PR source: wildcard type/subtype rejection, invalid-UTF-8 rejection, 204 `request→undefined` + `requestData` throw, `+json` suffix acceptance, malformed-JSON code — 6/6 pass on PR source, 0/6 on pre-PR source. Probe kept at `/tmp/30617-probe.ts` (not committed).
- Read the `http.ts` diff: `isJsonMediaType` correctly takes the pre-`;` token, lowercases, requires exactly one `/`, and validates both halves against the restricted-name pattern (wildcards rejected); the "+json" suffix rule requires a non-empty prefix, matching the old regex's strictness; error-path normalization is preserved (`parsedBodyValue` feeds the same string branch the old malformed marker used); bodyless 2xx now cancels the body explicitly instead of leaking it. Caller migrations are mechanical `get→requestData` for data-required paths, and CI typecheck covers the `T | undefined` widening.

One environment note: also confirmed `bun test` is the wrong runner for the integration file (it imports vitest and mis-executes); the package's `bun test src` picks it up — worth confirming CI runs it under a socket-capable lane, since it passed there.

lalalune (approved):
Reviewed the complete SDK and plugin diff and callers. Repaired generic mutation helpers rejecting valid bodyless successes, verified the regression against a real HTTP server, regenerated routes against the current API, and validated the full SDK/plugin suites, builds, live public API and root verification. Approved at 298bf89036cb2123baba8393a930fa97ac1e6c59; merge remains conditional on exact-head hosted checks.

PR #30618, opened 2026-09-05
fix(app-core): keep source asset manifests stable after normal builds

### LINKED ISSUE #30601: fix(app-core): keep source asset manifests stable after normal builds
The normal app build runs `packages/app/scripts/sync-homepage-assets.mjs`, producing eight ignored marketing/font/installer files in `packages/app/public`. `validateStaticAssetManifest()` inventories every non-hidden file on disk, so this supported build makes the checked-in source/CDN manifest stale. Both the app-core contract test and the release/CDN validation callers then reject the checkout. Regenerating the manifest after the build would include ignored paths that are absent from the published Git ref.

Observed in a real native build checkout: `eliza-logo.webp`, `elizawallpaper.webp`, both Geist font files, `grain.webp`, `install.ps1`, `install.sh` and `tbg.webp` are reported as unexpected. The unchanged source manifest passes against a pristine archive (97 app assets and 71 homepage assets).

Plan: use Git's tracked and non-ignored untracked public-file inventory when validating a checkout, intersected with files actually present on disk. Preserve existing filesystem inventory for standalone archive/fixture roots. Reuse the same canonical builder for generation and validation; do not add a second marketing asset list or change deployed bytes.

Acceptance:

- The real homepage-copy producer leaves the source manifest unchanged and valid.
- New non-ignored assets and missing tracked assets still make the manifest stale.
- Tracked assets remain included even if an ignore rule matches them.
- Git metadata/command failures fail explicitly, rather than becoming an empty valid inventory.
- Standalone archive trees retain their current behavior; relevant manifest, CDN and asset-copy tests pass.

Resolution: adopted the independently merged asset-manifest implementation in #30618. Its integrated contract checks passed all nine asset tests and root verification passed all 376 tasks at `89e62a3b35711e9834d17e814eb350d50afb0cef`. The duplicate implementation was removed from #30643.

wtfsayo:
Closes #30601.

Normal homepage synchronization copies ignored files into `packages/app/public`. The source-manifest generator previously included those build outputs, so normal builds made an otherwise valid source manifest appear stale.

Use Git's tracked and unignored inventory in repository checkouts, while retaining the filesystem walk for source archives. Deleted source files are excluded; invalid repository metadata produces an explicit error. The canonical manifest is `scripts/generated/static-asset-manifest.json`.

Reviewed the Git inventory boundary, filesystem archive fallback, symlink/error behavior, callers, and the actual asset-sync producer. This is a useful fix: normal ignored build output must not alter the source manifest.

Validated head `98023bb6a4638dff119d54ba49f08e34b9bee6b8` after rebase onto `f567055f1c3656590369e6f9c416c253a773f283` with Bun 1.3.14 / Node 24.15.0:
- `bun install` passed.
- Actual homepage asset sync: the old builder counted 105 app assets; this builder correctly counted 97, excluding eight ignored copies. The 71 homepage assets and canonical `scripts/generated/static-asset-manifest.json` stayed unchanged; manifest validation passed.
- Focused static-asset contract: 8/8 passed.
- Full app-core Vitest run: 4,467 passed, 25 skipped, one 120-second timeout in an unchanged coding-account refresh test under machine load. An unchanged rerun of that entire file passed all 27 tests. No timeout was raised or test disabled.
- Package script suites: 74 Node tests and 21 Bun tests passed.
- Changed-file Biome and package format check passed. Root `bun run verify` passed all 376 workspace tasks and every repository audit, including package lint/typecheck.

No implementation changes were needed beyond rebasing. This script-only change has no rendered UI or model behavior; screenshots, native captures, and model trajectories are N/A. Hosted checks must pass on this head before merge.

### DIFF
--- a/packages/app-core/scripts/__tests__/static-asset-contract.test.ts
+++ b/packages/app-core/scripts/__tests__/static-asset-contract.test.ts
@@ -1,17 +1,31 @@
 /**
- * Contract tests for the archive-overlay retirement (#16290): the checked-in
- * static asset manifest must match a pristine checkout (no implicit artifact
- * sync may be needed to make it green), and the overlay-only assets retired
- * with the eliza-archive sync must not resurface — neither as files nor as
- * references from production source. Runs against the real repository tree
- * and the real generator; no mocks.
+ * Contract tests for source-owned static asset inventory. The suite exercises
+ * the real repository and homepage copy producer, temporary Git checkouts, and
+ * standalone filesystem roots so ignored build output cannot alter a source
+ * manifest while real additions and deletions remain visible.
  */
 import { execFileSync } from "node:child_process";
-import { existsSync } from "node:fs";
+import {
+  existsSync,
+  mkdirSync,
+  mkdtempSync,
+  rmSync,
+  unlinkSync,
+  writeFileSync,
+} from "node:fs";
+import os from "node:os";
 import path from "node:path";
 import { fileURLToPath } from "node:url";
 import { describe, expect, it } from "vitest";
-import { validateStaticAssetManifest } from "../lib/static-asset-manifest.mjs";
+import {
+  HOMEPAGE_PUBLIC_ASSETS,
+  syncHomepageAssets,
+} from "../../../app/scripts/sync-homepage-assets.mjs";
+import {
+  buildStaticAssetManifest,
+  validateStaticAssetManifest,
+  writeStaticAssetManifest,
+} from "../lib/static-asset-manifest.mjs";
 
 const REPO_ROOT = path.resolve(
   path.dirname(fileURLToPath(import.meta.url)),
@@ -20,6 +34,22 @@ const REPO_ROOT = path.resolve(
   "..",
   "..",
 );
+
+function writeFixtureFile(rootDir: string, relativePath: string): void {
+  const absolutePath = path.join(rootDir, relativePath);
+  mkdirSync(path.dirname(absolutePath), { recursive: true });
+  writeFileSync(absolutePath, `${relativePath}\n`);
+}
+
+function createGitFixture(): string {
+  const rootDir = mkdtempSync(path.join(os.tmpdir(), "eliza-static-assets-"));
+  execFileSync("git", ["init", "--quiet", rootDir]);
+  return rootDir;
+}
+
+function stageFixturePaths(rootDir: string, ...relativePaths: string[]): void {
+  execFileSync("git", ["-C", rootDir, "add", "--", ...relativePaths]);
+}
 
 /**
  * Assets that existed only in the retired eliza-archive overlay. They have no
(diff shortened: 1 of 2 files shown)


ss251 (approved):
Reviewed head d08920b7e03666ae1548b4040bb2bc52c8675ac7.

The risky part is `packages/app-core/scripts/lib/static-asset-manifest.mjs:95` where `buildStaticAssetManifest` switches to `git ls-files --cached --others --exclude-standard -z` when `.git` exists. I checked the boundary that decides this fix: ignored untracked build output under `packages/app/public/**` is excluded by `--exclude-standard`, so the real `syncHomepageAssets` copy in the new `stays valid after the real homepage asset producer` test leaves `validateStaticAssetManifest(rootDir).ok` true; while a tracked path matching `.gitignore` is still returned because `--cached` includes tracked files despite ignore rules, confirmed by the `includes present tracked assets even when an ignore rule matches` test via `isIncludedStaticAsset` plus `existsSync` filtering.

Recommendation: merge

— [muse-lane-ss251]

ss251 (approved):
Reviewed head d08920b7e03666ae1548b4040bb2bc52c8675ac7.

The risky part is `packages/app-core/scripts/lib/static-asset-manifest.mjs:143` where `buildStaticAssetManifest` branches on `.git`: checkouts use `git ls-files --cached --others --exclude-standard -z` (:105), archives keep `listPublicFiles`. I traced the boundaries: `--cached` keeps a tracked file even when `.gitignore` matches it, so the tracked-ignored fixture stays included; `--others --exclude-standard` drops ignored `packages/app/public/**` output from `syncHomepageAssets`, so the homepage-copy test stays valid; the `existsSync` filter strips index entries deleted from disk, so missing-tracked still validates false. The `startsWith(prefix+"/")` split preserves the `packages/...` paths the contract asserts.

No blocking concerns.
Verified by reading the diff at this head and walking each new fixture through the two branches.

Recommendation: merge

— [muse-lane-ss251]

byt61 (approved):
Approved. Verified locally via overlay on a clean checkout (vitest, app-core):

- PR lib + PR tests: 7/8 pass; all 5 new tests pass (homepage-producer ignored-output stability, addition/deletion detection, tracked-beats-ignored, explicit Git-metadata failure, standalone-archive filesystem fallback).
- The 1 failure (`retired overlay assets do not exist in the checkout`) is pre-existing checkout state, not the PR: a baseline run with fully-tracked files fails the same test (my lane branch carries 19 retired-overlay offenders plus extra on-disk files). Notably, the PR version *fixes* the other baseline failure — `checked-in manifest matches a pristine checkout` passes with the git-inventory lib where the old filesystem walk fails on a checkout containing ignored build output. That is direct evidence the change does what the title claims.
- Read the source diff: `git ls-files --cached --others --exclude-standard -z` plus the `existsSync` filter correctly handles staged-new, unstaged-deleted, and ignored-but-tracked cases; the J2 throw on Git failure keeps validation fail-closed; `fs.existsSync(.git)` also covers linked-worktree `.git` files.

attentionhead (approved):
Approving at `d08920b7e03666ae1548b4040bb2bc52c8675ac7`.

Switching source-asset discovery from a filesystem walk to `git ls-files --cached --others --exclude-standard` is the right fix rather than a heuristic: the manifest is supposed to describe the *source* contract, so asking Git what is source — instead of inferring it from whatever happens to be on disk — makes normal build output structurally incapable of changing it.

## Both halves are load-bearing

```
drop --exclude-standard  → × stays valid after the real homepage asset producer copies ignored output
drop the existsSync filter → × detects non-ignored additions and missing tracked assets
```

The first test is stronger than a synthetic fixture would have been, since it runs the real producer and reproduces the actual reported failure. The second guards a subtlety that is easy to miss: `--cached` lists tracked paths even after they are deleted from the working tree, so without the `fs.existsSync` filter a removed asset would linger in the manifest — the mirror image of the bug being fixed.

## Verified end to end on a real checkout

I built a throwaway repository under the app public prefix and ran `buildStaticAssetManifest` against it directly:

```
"packages/app/public/normal.txt"
"packages/app/public/sub/new\nline.txt"
"packages/app/public/with space.txt"
```

`.DS_Store`, a `.hidden/` directory, and a gitignored `dist-ignored.txt` are all correctly absent. The embedded-newline case is the one worth calling out: `-z` is doing real work there, because a line-based parse of `git ls-files` would have split that single path into two bogus entries, and the default (non-`-z`) output would have arrived octal-quoted. Choosing NUL delimiting up front avoids a whole class of path-quoting bugs.

Renaming `.git` away on the same fixture and re-running produces the identical list plus `dist-ignored.txt`, which confirms the divergence between the two discovery paths is exactly the intended one and nothing more.

## Parity between the paths

I checked that the two inventories agree on filtering, since a checkout and a standalone archive disagreeing would be a silent contract drift. `listFilesRecursive` skips any entry whose name starts with `.` or is in `IGNORED_STATIC_ASSET_BASENAMES`, and `isIncludedStaticAsset` applies the same two rules to every segment of the repo-relative path. Equivalent, and the probe above confirms it behaviourally.

Failing explicitly when `git` cannot inventory the checkout is the correct call for a validation input. A fallback to the filesystem walk would silently reintroduce the very build output this change exists to exclude, so a partial manifest is the worse outcome.

## Verification

- `static-asset-contract.test.ts` — 8 pass; 1 fail under each of the two mutations
- `validate-cdn-assets-retry-env.test.ts` — the other consumer of `buildStaticAssetManifest`, untouched by this PR, still passes
- Real-repository probe of spaces, embedded newlines, dot-directories, `.DS_Store`, and ignored output, in both checkout and archive modes
- Biome clean on both changed files

## Not verified

I did not run the full `packages/app-core` suite or an actual CDN-asset validation against a real build, and I did not exercise a Git worktree (where `.git` is a file rather than a directory) — `fs.existsSync` covers both and `git -C` works in a worktree, but I confirmed that by inspection rather than by running it.

— [claude-code-eliza]

nothingxnowhere (approved):
Reviewed at head `d08920b7e0` against `develop` `1045297cca`. Two things I wanted to know that the fixtures do not answer on their own: does the Git inventory produce the *same* manifest as the walk it replaces on a real checkout, and does it work when `.git` is a file rather than a directory.

**Parity on a real checkout.** I imported both implementations side by side — head's `buildStaticAssetManifest` and develop's — and ran them against the same repository root:

```
head:          app=97 homepage=71 sha=a6c33d052a4a
develop-walk:  app=97 homepage=71 sha=a6c33d052a4a
identical: true   onlyHead: []   onlyDev: []
```

So the switch from `listFilesRecursive` to `git ls-files --cached --others --exclude-standard` changes nothing in the checked-in contract. That includes the two tracked `.well-known` files under `packages/homepage/public` (73 tracked, 71 in the manifest): the walk skipped dot-entries and `isIncludedStaticAsset` skips dot-segments, so both exclude them identically. Worth knowing that this is a pre-existing exclusion and not something this PR introduces or fixes.

**Worktree checkouts.** My checkout is a `git worktree`, where `.git` is a pointer file, not a directory. `fs.existsSync` is true for it, the Git branch is taken, `git -C <root> ls-files` resolves through the pointer, and the full contract suite is `8 pass` there. The fixtures only ever `git init` a directory, so this is the configuration they do not cover, and it holds.

**Control.** With the PR's tests kept and `static-asset-manifest.mjs` reverted to develop: `2 fail / 6 pass`. The two are the cases that need the new path — the real producer writing ignored copies (on develop the walk lists all eight and the manifest is reported stale, which is the bug) and the explicit failure on unreadable checkout metadata. The other three new fixtures (non-ignored additions and deleted tracked assets, tracked-but-ignored inclusion, standalone root) pass on both, so they pin behaviour the walk already had; that is the right shape for a change whose promise is "same answer, minus generated output".

One small correction for the body: the checked-in manifest is `scripts/generated/static-asset-manifest.json`, not `packages/app-core/src/static-asset-manifest.ts`, which does not exist on develop. Approving.

```json
{
  "schemaVersion": "2",
  "headSha": "d08920b7e0",
  "recommendation": "accept",
  "reproduced": true,
  "risks": "none",
  "effortBand": "medium",
  "complexity": "moderate",
  "impact": "medium",
  "reviewLoad": "standard",
  "recommendedTier": "small",
  "recommendedThirds": 3,
  "workUnitId": "wu_elizaos_eliza_pr_30618",
  "valueRationale": "Imported head's and develop's manifest builders side by side on a real checkout (identical 97/71, same digest, including the shared dot-path exclusion), ran the suite in a git worktree where .git is a file (8 pass), and confirmed the control fails exactly the two cases that need the Git inventory.",
  "usefulArtifacts": ["side-by-side parity probe of buildStaticAssetManifest; contract suite in a worktree checkout; RED run against develop's module"],
  "commands": ["bunx vitest run --config packages/app-core/vitest.config.ts packages/app-core/scripts/__tests__/static-asset-contract.test.ts   # head: 8 pass; develop mjs + PR tests: 2 failed | 6 passed"],
  "evidenceUrls": [],
  "summary": "Static-asset manifest inventory moves from a filesystem walk to Git's tracked/non-ignored view; verified byte-identical output on a real checkout, worktree compatibility, and that only the two new-path cases fail on develop."
}
```

— [nxn-claude-code]

lalalune (approved):
Reviewed the Git inventory boundary, filesystem archive fallback, symlink/error behavior, callers, and the actual asset-sync producer. This is a useful fix: normal ignored build output must not alter the source manifest.

Validated head `98023bb6a4638dff119d54ba49f08e34b9bee6b8` after rebase onto `f567055f1c3656590369e6f9c416c253a773f283` with Bun 1.3.14 / Node 24.15.0:
- `bun install` passed.
- Actual homepage asset sync: the old builder counted 105 app assets; this builder correctly counted 97, excluding eight ignored copies. The 71 homepage assets and canonical `scripts/generated/static-asset-manifest.json` stayed unchanged; manifest validation passed.
- Focused static-asset contract: 8/8 passed.
- Full app-core Vitest run: 4,467 passed, 25 skipped, one 120-second timeout in an unchanged coding-account refresh test under machine load. An unchanged rerun of that entire file passed all 27 tests. No timeout was raised or test disabled.
- Package script suites: 74 Node tests and 21 Bun tests passed.
- Changed-file Biome and package format check passed. Root `bun run verify` passed all 376 workspace tasks and every repository audit, including package lint/typecheck.

No implementation changes were needed beyond rebasing. This script-only change has no rendered UI or model behavior; screenshots, native captures, and model trajectories are N/A. Hosted checks must pass on this head before merge.

PR #30621, opened 2026-09-05
fix(registry): isolate generation CLI from public imports

### LINKED ISSUE #30587: Bundled registry imports unexpectedly run the generation CLI and break iOS startup
Importing the public community-registry package from a Bun-bundled agent runs its registry-generation CLI. The barrel re-exports generate.ts, whose direct-execution check becomes true after bundling because its import.meta.url and the consumer entrypoint both identify the bundle.

Observed impact: the current iOS local-agent build fails during startup when the imported generator attempts to write generated-registry.json into the app bundle. This happens before the native Fetch acceptance checks. A separate minimal Bun bundle importing only isValidRegistryPackageName also creates an empty generated-registry.json without any explicit generation request.

Acceptance criteria:

- Importing and using the public registry API in an executed Bun bundle performs no registry generation or filesystem writes.
- The existing package generate command still explicitly regenerates the same wire registry.
- Public programmatic generator exports remain compatible.
- A real bundler/subprocess regression covers the import boundary, and the installed iOS build progresses past this startup failure.

Plan: separate the executable generator entrypoint from the importable conversion module, preserve the package command, and verify both paths. Related native transport cleanup: #30569.

Resolved by merged PR #30621. The overlapping implementation in #30643 was removed during integration so the repository retains one import-safe generator and one explicit CLI. At combined validation head f90d1589a38d99f9242ab09bb27800a423379e88, all 57 registry tests passed, including real bundled import and explicit package generation against the retained upstream implementation.

wtfsayo:
Closes #30587.

Separate the registry filesystem writer into an unexported CLI entrypoint. Importing the public conversion API from an executed Bun bundle now has no generation side effect; the explicit package generate command keeps byte-identical output.

## Validation

Reviewed the complete diff and public registry consumers. This fixes a real import side effect: with current develop's generator, a minimal executed Bun bundle importing `isValidRegistryPackageName` writes an unintended empty generated-registry.json. I reproduced that behavior independently. The separated CLI prevents the write, while explicit generation remains byte-compatible.

Rebased head `6f17ed026f379727ff8c17b5757db02ddf08bc93` onto `f567055f1c3656590369e6f9c416c253a773f283` without conflicts. No implementation changes were necessary.

Independent validation with Bun 1.3.14 / Node 24.15.0:

- Registry tests: 7 files, 57 tests passed, including actual Bun build/subprocess import and CLI generation.
- Registry typecheck, lint:check, format:check, validate, generate:first-party:check: passed.
- Explicit `generate` left the canonical JSON byte-unchanged.
- Guide parity: passed.
- `bun install`: passed.
- Root `bun run verify`: 376/376 workspace tasks and all repository audits passed.

No rendered UI or model-facing behavior changed. The real executed bundle and generated-file inspection cover the changed package boundary; the author's separately documented native simulator startup is additional evidence, not a run I performed.

## Evidence

- [x] N/A - No rendered UI change.

- [x] N/A - No rendered UI change.

- [x] N/A - No interactive UI or native transport behavior changes.

- [x] Executed Bun bundle and generator command results are documented above.

- [x] N/A - No frontend path changes.

- [x] N/A - No model, prompt, provider inference, or action behavior changes.

- [x] The explicit generator output was byte-compared against the canonical JSON; import-only execution creates no file.

- [x] N/A - No rendered UI change.

### DIFF
--- a/packages/registry/AGENTS.md
+++ b/packages/registry/AGENTS.md
@@ -36,7 +36,8 @@ src/
   types.ts        RegistryEntry (source) + GeneratedRegistry (wire) types
   schema.ts       validateRegistryEntry / assertRegistryEntry (dependency-free)
   loader.ts       loadThirdPartyEntries — read + validate entries/third-party
-  generate.ts     generateRegistry / toGeneratedEntry — entries → wire format
+  generate.ts     import-safe generateRegistry / toGeneratedEntry conversion API
+  generate-cli.ts explicit filesystem writer used only by the package command
   validate-cli.ts `bun run validate`
   index.ts        public barrel (typed loader for programmatic consumers)
   first-party/    @elizaos/registry/first-party — curated bundled registry
@@ -102,8 +103,10 @@ bun run --cwd packages/registry format:check
 
 - The `@elizaos/*` scope is reserved for first-party packages — the validator
   rejects it in source entries.
-- `generate.ts` and `validate-cli.ts` are run with `bun` (TypeScript directly);
-  there is no `dist` build step beyond regenerating the JSON.
+- `generate-cli.ts` and `validate-cli.ts` are run with `bun` (TypeScript
+  directly); there is no `dist` build step beyond regenerating the JSON.
+- Keep executable entrypoints out of `src/index.ts`. Public imports are bundled
+  into native agents and must not perform generation or filesystem writes.
 - `zod` supports the first-party catalog. Do not make community-registry
   validation depend on first-party catalog types or schemas.
 
--- a/packages/registry/CLAUDE.md
+++ b/packages/registry/CLAUDE.md
@@ -36,7 +36,8 @@ src/
   types.ts        RegistryEntry (source) + GeneratedRegistry (wire) types
   schema.ts       validateRegistryEntry / assertRegistryEntry (dependency-free)
   loader.ts       loadThirdPartyEntries — read + validate entries/third-party
-  generate.ts     generateRegistry / toGeneratedEntry — entries → wire format
+  generate.ts     import-safe generateRegistry / toGeneratedEntry conversion API
+  generate-cli.ts explicit filesystem writer used only by the package command
   validate-cli.ts `bun run validate`
   index.ts        public barrel (typed loader for programmatic consumers)
   first-party/    @elizaos/registry/first-party — curated bundled registry
@@ -102,8 +103,10 @@ bun run --cwd packages/registry format:check
 
 - The `@elizaos/*` scope is reserved for first-party packages — the validator
   rejects it in source entries.
-- `generate.ts` and `validate-cli.ts` are run with `bun` (TypeScript directly);
-  there is no `dist` build step beyond regenerating the JSON.
+- `generate-cli.ts` and `validate-cli.ts` are run with `bun` (TypeScript
+  directly); there is no `dist` build step beyond regenerating the JSON.
+- Keep executable entrypoints out of `src/index.ts`. Public imports are bundled
+  into native agents and must not perform generation or filesystem writes.
 - `zod` supports the first-party catalog. Do not make community-registry
   validation depend on first-party catalog types or schemas.
 
--- a/packages/registry/README.md
+++ b/packages/registry/README.md
@@ -19,7 +19,8 @@ packages/registry/
     ├── types.ts                RegistryEntry + GeneratedRegistry types
     ├── schema.ts               dependency-free entry validation
     ├── loader.ts               read + validate entries/third-party/*.json
-    ├── generate.ts             entries → generated-registry.json
+    ├── generate.ts             import-safe entries → wire conversion API
+    ├── generate-cli.ts         explicit generated-registry.json writer
     ├── validate-cli.ts         `bun run validate`
     └── index.ts                typed loader for programmatic consumers
 ```
@@ -30,8 +31,8 @@ Two formats live here:
    `elizaos plugins submit --dry-run` author. One file per package.
 2. **Generated wire registry** (`generated-registry.json`) — the
    `{ registry: { "<package>": { … } } }` shape the runtime consumes from
-   `plugins.eliza.app/generated-registry.json`. Produced by `generate.ts`;
-   never hand-edited.
+   `plugins.eliza.app/generated-registry.json`. Produced by `bun run generate`
+   through `generate-cli.ts`; never hand-edited.
 
 ## Commands
 
--- a/packages/registry/package.json
+++ b/packages/registry/package.json
@@ -44,7 +44,7 @@
   },
   "scripts": {
     "validate": "bun run src/validate-cli.ts",
-    "generate": "bun run src/generate.ts",
+    "generate": "bun run src/generate-cli.ts",
     "generate:first-party": "bun run src/first-party/generate.ts",
     "generate:first-party:check": "bun run src/first-party/generate.ts --check",
     "typecheck": "tsc --noEmit -p tsconfig.json",
--- a/packages/registry/src/bundled-import.test.ts
+++ b/packages/registry/src/bundled-import.test.ts
@@ -1,0 +1,105 @@
+/**
+ * Proves the public registry API is inert in a real Bun bundle while the explicit
+ * package command still generates the canonical wire artifact in a filesystem fixture.
+ */
+
+import { execFileSync } from "node:child_process";
+import {
+  cpSync,
+  existsSync,
+  mkdirSync,
+  mkdtempSync,
+  readFileSync,
+  rmSync,
+  writeFileSync,
+} from "node:fs";
+import { tmpdir } from "node:os";
+import path from "node:path";
+import { fileURLToPath } from "node:url";
+import { afterEach, describe, expect, it } from "vitest";
+
+const packageRoot = path.resolve(
+  path.dirname(fileURLToPath(import.meta.url)),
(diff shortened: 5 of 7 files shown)


ss251 (approved):
Reviewed head d665656677115f44157ade8ef4ce097ee36ee2cf.

Risk is `packages/registry/src/generate.ts` deleting `isDirectExecution()`+`main()` auto-write. Old path wrote `generated-registry.json` when `import.meta.url` matched `argv[1]`; its comment admits in mobile bundles both identify the bundle, so only `ELIZA_DISABLE_DIRECT_RUN`/`__ELIZA_MOBILE_BUNDLE__` stopped a boot-time rewrite of a read-only asset. New `generate.ts` imports only `loader.ts`+types, no `node:fs/path/url`, so inert; write is now explicit in `generate-cli.ts:15` wired by `package.json:47` to `bun run src/generate-cli.ts`.

`bundled-import.test.ts:42` pins the boundary: `bun build --target bun` of an entry importing `isValidRegistryPackageName` from `src/index.ts` must print `validator-ok` with `existsSync(generatedPath)` false; second test asserts fixture output byte-equals canonical.

Verified by reading diff at this head and tracing write edges.

Recommendation: merge

— [muse-lane-ss251]

byt61 (approved):
Approved. Verified locally via overlay on a clean checkout (package vitest runner):

- New boundary suite: 2/2 pass (bundled-import inertness + generate byte-compatibility).
- Mutant check (new tests vs pre-PR `generate.ts`): the bundle-inertness test fails — the old `isDirectExecution` check fires inside the Bun bundle and writes `generated-registry.json` — while byte-compatibility passes on both. Exactly the right coupling: the test reproduces the mobile-boot hazard the split eliminates, and proves the wire output is unchanged.
- Static checks at PR head: `AGENTS.md`/`CLAUDE.md` are byte-identical; nothing outside `packages/registry` references `generate-cli`; the barrel still exports the now side-effect-free `generateRegistry`, which is safe because generation-on-import is gone rather than guarded; the mobile-bundle markers removed from this module still have live consumers in `packages/agent` (`eliza.ts`), so no orphaned mechanism.

attentionhead (approved):
Approving at `d665656677115f44157ade8ef4ce097ee36ee2cf`.

Deleting the direct-run mechanism outright is the right call rather than hardening it further. The `import.meta.url === pathToFileURL(process.argv[1]).href` check is unreliable precisely where it matters: inside a bundle both sides identify the bundle, so the module regenerated a packaged read-only asset on every agent boot. The previous fix bolted two bundler-defined markers onto that check; this one removes the reason to ask the question at all by moving the side effect into an entrypoint that is never importable.

## The regression test is real, and it is the right one

I confirmed the test is not vacuous by restoring `develop`'s `main()`/`isDirectExecution()` into `generate.ts`:

```
× does not generate files when the public API runs from a Bun bundle
Tests  1 failed | 1 passed (2)
```

It builds an actual Bun bundle of `src/index.ts`, executes it, and asserts no artifact appears — reproducing the reported failure mode end to end rather than asserting against a mock. The paired case, that `bun run generate` still writes the canonical artifact into a filesystem fixture, keeps the removal honest: it would be easy to make the first assertion pass by breaking generation entirely.

## Isolation holds

`generate-cli.ts` is reachable only through the `generate` package script. It is absent from `exports` — which lists just `.`, `./first-party`, and the JSON artifacts — and is not re-exported from `src/index.ts`, so no consumer can pull the filesystem write back in transitively. `generate.ts` keeps exporting `generateRegistry` and is now genuinely import-safe, having lost its `node:fs`/`node:path`/`node:url` imports along with the side effect.

I also checked the docs actually moved with the code: no remaining instruction anywhere in the package's markdown or `package.json` tells a reader to run `src/generate.ts` directly.

## The removed markers are not orphaned

`ELIZA_DISABLE_DIRECT_RUN` and `__ELIZA_MOBILE_BUNDLE__` are still defined by `build-mobile-bundle.mjs` and still consumed by `packages/agent/src/runtime/eliza.ts`, so dropping them here leaves no dead configuration and correctly does not disturb the other consumer.

## One observation, explicitly not blocking

The sibling generator `src/first-party/generate.ts` still ends in a module-scope `if (import.meta.main) { main(); }` guarding `writeFileSync` calls. That is the same shape of hazard this PR is retiring, and `import.meta.main` is true for a bundle entry too. It is not a live bug — the module is not in `exports` and nothing imports it, so it is never bundled — but the invariant this PR establishes ("an executable entrypoint must never be exported by the importable package API") is only structurally enforced for one of the two generators. If that file ever becomes importable, the same failure returns. Worth the same split at some point; nothing to do here.

## Verification

- `bundled-import.test.ts` 2 pass; 1 fail with the direct-run mechanism restored
- Full `packages/registry` suite — 57 pass
- Typecheck 0 errors; Biome clean on 56 files
- `exports` and `src/index.ts` inspected for reachability of the CLI

## Not verified

I did not build or boot the mobile bundle itself, so the iOS runtime evidence in the description stands on your testing rather than mine — my bundle check is the package-level reproduction, not the shipped artifact. I also did not run the root `verify` lane, so I cannot speak to the unrelated stale-UI-inventory blocker you recorded.

— [claude-code-eliza]

nothingxnowhere (approved):
Reviewed at head `d665656677` against `develop` `1045297cca`. The question a reviewer has to answer here is whether anything still reaches for the old executable path, and whether the new bundle test actually distinguishes the two states. Both check out.

**Nothing else invokes the old entrypoint.** On develop there are zero references to `registry/src/generate.ts` outside the package, no workflow runs the registry generator, and the `ELIZA_DISABLE_DIRECT_RUN` / `__ELIZA_MOBILE_BUNDLE__` defines the deleted guard consumed are still consumed by `packages/agent/src/runtime/eliza.ts` (`:243`, `:6785-6790`), so `build-mobile-bundle.mjs`'s defines are not orphaned by this change. The public barrel does import `generateRegistry` from `./generate.ts` (`index.ts:12`), which is why the executable tail lived in every bundle that touched the registry API.

**The bundle regression is real.** At head `bundled-import.test.ts` is `2 pass`. With only `src/generate.ts` reverted to develop, the first case fails with `expected true to be false` — the develop module, bundled from `src/index.ts` and run as `bun entry.js`, sees `import.meta.url === pathToFileURL(process.argv[1])` and writes `generated-registry.json` next to the bundle. That is precisely the boot-time regeneration the mobile guard existed to suppress, reproduced without any guard marker set. The second case (fixture copy of `src`, `entries` and `package.json`; `bun run generate`; byte-compare to the canonical artifact) passes both ways, as it should — the writer's output is unchanged, only its location.

**One thing for the author.** The two parameterised `does not regenerate when the … marker is set` cases in `registry.test.ts` still exist and still pass at head, but their parameterisation no longer means anything: `generate.ts` has no execution tail, so they now assert only that importing the module prints nothing, regardless of marker. Worth collapsing into a single "importing generate.ts has no side effects" case, or deleting in favour of the new bundle test. (In my checkout those two fail identically on develop and at head because `vitest` here runs under Node 20, which cannot import `.ts`; the repo pins Node 24, so that is my environment, not the PR.)

Approving.

```json
{
  "schemaVersion": "2",
  "headSha": "d665656677",
  "recommendation": "accept",
  "reproduced": true,
  "risks": "none",
  "effortBand": "medium",
  "complexity": "simple",
  "impact": "medium",
  "reviewLoad": "standard",
  "recommendedTier": "small",
  "recommendedThirds": 3,
  "workUnitId": "wu_elizaos_eliza_pr_30621",
  "valueRationale": "Checked on develop that nothing outside the package invokes the old generate.ts and that the mobile-bundle defines still have a consumer, then ran the bundle regression at head (2 pass) and with develop's generate.ts (the bundle writes generated-registry.json, 1 fail), and flagged the now-vacuous marker tests.",
  "usefulArtifacts": ["RED/GREEN run of bundled-import.test.ts; caller and define-consumer greps on develop"],
  "commands": ["bunx vitest run src/bundled-import.test.ts   # head: 2 pass; develop generate.ts: 1 failed | 1 passed (bundle wrote the registry file)"],
  "evidenceUrls": [],
  "summary": "Moves the registry writer out of the publicly imported generate.ts into an unexported CLI; verified no external callers, the bundle regression reproduces the boot-time write on develop, and the byte-compat generate path is unchanged."
}
```

— [nxn-claude-code]

lalalune (approved):
Reviewed the complete diff and public registry consumers. This fixes a real import side effect: with current develop's generator, a minimal executed Bun bundle importing `isValidRegistryPackageName` writes an unintended empty generated-registry.json. I reproduced that behavior independently. The separated CLI prevents the write, while explicit generation remains byte-compatible.

Rebased head `6f17ed026f379727ff8c17b5757db02ddf08bc93` onto `f567055f1c3656590369e6f9c416c253a773f283` without conflicts. No implementation changes were necessary.

Independent validation with Bun 1.3.14 / Node 24.15.0:

- Registry tests: 7 files, 57 tests passed, including actual Bun build/subprocess import and CLI generation.
- Registry typecheck, lint:check, format:check, validate, generate:first-party:check: passed.
- Explicit `generate` left the canonical JSON byte-unchanged.
- Guide parity: passed.
- `bun install`: passed.
- Root `bun run verify`: 376/376 workspace tasks and all repository audits passed.

No rendered UI or model-facing behavior changed. The real executed bundle and generated-file inspection cover the changed package boundary; the author's separately documented native simulator startup is additional evidence, not a run I performed.

PR #30166, opened 2026-08-31
feat(cloud): reserve subscription allowance before credits

wtfsayo:
## Summary

- reserve allowance-eligible usage from the current subscription period before purchased credits
- keep cash-only operations entirely on purchased credits
- lock the organization and both funding sources in one transaction
- persist the exact allowance and purchased-credit allocation through #30154's reviewed authority tables
- settle under-estimates to the original sources, recording expired allowance refunds without minting cash
- route paid inference through the subscription funding path while free organizations retain the existing credit flow

Stacked on #30165, which is stacked on #30154. This is the reservation and bounded settlement slice of #23092. Overage collection and remaining debit-boundary migrations follow in the next stacked PR.

## Verification

- Cloud shared TypeScript check
- allowance-first and cash-only split tests
- paid/free inference funding selection tests
- funding policy and production debit inventory tests
- Biome on changed TypeScript files
- `git diff --check`

PR #30623, opened 2026-09-05
fix(cloud): repair cloud test contracts dropped with Outreachr FKs and admission reads

lumix5:
## Summary

Two deterministic `Cloud tests` failures at the develop head (`1045297ccaffff37519a6a4b2a0203c273443a84`, Develop Full run 33953591123, job 101273052031):

1. **`account-deletion-foreign-key-policy.test.ts`** — pinned a 248-descriptor FK inventory, got 250. `182c3ebad2` (#30607) added `outreachr_delegations` with `user_id`→users and `organization_id`→organizations cascade FKs without classifying them. The delegation client exposes an external revoke (`outreachr-delegation.ts` `deleteGrant`), and every other connection-style table (`discord_connections`, `telegram_chats`, `vendor_connections`, …) is `reconcile_external_resource` — a bare cascade would leave a live external identity grant behind after account deletion. This classifies the table as an external resource and re-pins the inventory (250, external 75) with the recomputed snapshot digest.
2. **`chat-completions-explicit-max-tokens-route.test.ts`** — 6 failures turning the 402-translation and forwarding cases into 500s. `a445c705a` (#30166) moved the authoritative subscription-entitlement read (`isSubscriptionFundedOrganization`) ahead of credit reservation in `admitOrganizationInference`; this suite's spread-actuals `ai-billing` mock let the real read hit the schemaless test PGlite (`relation "organization_entitlements" does not exist`). The same commit updated its sibling suites (`messages-billing-offpath.test.ts`, `generative-route-auth.test.ts`) but missed this one. The fix pins the boundary to credits-funded with the sibling-suite idiom.

## Evidence

| Row | Result |
| --- | --- |
| Evidence head | `da8092db990d5433c3308a5ffd382938230c9989` |
| Reproduction (red) at develop head | disposable container (oven/bun:1.3.14, `bun install --frozen-lockfile --ignore-scripts`, archive of `1045297cca`, md5-verified): FK policy 2 fail/6, max-tokens 6 fail/10 — matches hosted job 101273052031 |
| Fix (green) at fix head, same container | FK policy 6/6 pass, max-tokens 10/10 pass |
| Regression sweep | `account-deletion-provider-adapters.test.ts` 13/13, `account-deletion-export.test.ts` 8/8 (isolated per the run-unit-isolated contract) |
| Hygiene | `biome check` clean on all three changed files; `tsc --noEmit` clean in `packages/cloud/shared` and `packages/cloud/api` |
| Hosted checks | fork PR — awaiting maintainer workflow approval (native `action_required` banner); no hosted lane runs on fork heads until then |
| Desktop/mobile screenshots | N/A - no UI surface changed (cloud test/policy files only) |
| Walkthrough video | N/A - no user-visible behavior change |
| Live-model trajectory | N/A - no agent/model-path change |

## Reviewer notes

- The digest constant change (`40df404d…` → `188ac838…`) is the recomputed SHA-256 over the widened 250-descriptor serialization; the recompute script and per-class counts (`reconcile_external_resource` 73→75 — the table contributes **two** FKs) are in the container log.
- The remaining `Smoke lanes` failures at this head are NOT addressed here and are separately owned or unclassified: cartesia-synthesis (open #30529), route-provider-selection Server-Timing (open #30395), the `object-store-immutable-upload` 15ms HEAD deadline race and the `agent-bridge-runtime-routing` sentinel-traceId assertion (both flaky-suspect, no repo cause pinned), and the Smoke (6) browser-workspace assertion.

— [eliza-review]

### DIFF
--- a/packages/cloud/api/__tests__/chat-completions-explicit-max-tokens-route.test.ts
+++ b/packages/cloud/api/__tests__/chat-completions-explicit-max-tokens-route.test.ts
@@ -304,6 +304,10 @@ mock.module("@/lib/services/ai-billing", () => ({
   })),
   recordUsageAnalytics: mock(async () => ({ id: "usage-record-1" })),
   InsufficientCreditsError: TestInsufficientCreditsError,
+  // Admission pays its authoritative subscription-entitlement read before credit
+  // reservation (organization-inference-admission); this harness has no
+  // subscription schema, so treat every org as credits-funded.
+  isSubscriptionFundedOrganization: mock(async () => false),
 }));
 
 mock.module("@/lib/services/ai-billing-records", () => ({
--- a/packages/cloud/shared/src/db/account-deletion-foreign-key-policy.test.ts
+++ b/packages/cloud/shared/src/db/account-deletion-foreign-key-policy.test.ts
@@ -27,7 +27,7 @@ describe("account deletion full-schema foreign-key policy", () => {
       .update(descriptors.map(serializeDescriptor).join("\n"))
       .digest("hex");
 
-    expect(descriptors).toHaveLength(248);
+    expect(descriptors).toHaveLength(250);
     expect(digest).toBe(ACCOUNT_DELETION_FOREIGN_KEY_SNAPSHOT_SHA256);
   });
 
@@ -37,10 +37,10 @@ describe("account deletion full-schema foreign-key policy", () => {
       action: classifyAccountDeletionForeignKey(descriptor),
     }));
 
-    expect(classified).toHaveLength(248);
+    expect(classified).toHaveLength(250);
     expect(classified.every(({ action }) => Boolean(action))).toBe(true);
     expect(classified.filter(({ action }) => action === "reconcile_external_resource").length).toBe(
-      73,
+      75,
     );
     expect(classified.filter(({ action }) => action === "transfer_shared_resource").length).toBe(
       11,
--- a/packages/cloud/shared/src/db/account-deletion-foreign-key-policy.ts
+++ b/packages/cloud/shared/src/db/account-deletion-foreign-key-policy.ts
@@ -23,9 +23,9 @@ export interface AccountDeletionForeignKeyDescriptor {
   targetColumns: string;
   onDelete: string;
 }
-/** SHA-256 of the 248 sorted direct user/organization FK descriptors. */
+/** SHA-256 of the 250 sorted direct user/organization FK descriptors. */
 export const ACCOUNT_DELETION_FOREIGN_KEY_SNAPSHOT_SHA256 =
-  "40df404dcdd7a24d2ff9dae7e9ae308eb4f422db2a94db8ca0d912d5eb1a9608";
+  "188ac83877c2538f847e72cb055a5a684d72fdf9d84c29437d522ce07ab9eb75";
 
 function serializeDescriptor(descriptor: AccountDeletionForeignKeyDescriptor): string {
   return [
@@ -110,6 +110,7 @@ const EXTERNAL_RESOURCE_TABLES = new Set([
   "org_storage_put_operations",
   "org_storage_read_operations",
   "organization_encryption_keys",
+  "outreachr_delegations",
   "platform_credential_sessions",
   "platform_credentials",
   "pooled_credentials",


ss251 (approved):
da8092db990d5433c3308a5ffd382938230c9989

This head repairs two contracts broken by upstream changes: new Outreachr FKs missing from the account-deletion policy, and a new admission entitlement read missing from the chat-completions test harness.

FINDINGS:

- `packages/cloud/shared/src/db/account-deletion-foreign-key-policy.ts:26-31` hash `40df40...` -> `188ac8...` with comment 248 -> 250, plus `:113` adding `outreachr_delegations` to `EXTERNAL_RESOURCE_TABLES`. Input: user/org with outreachr delegation rows deleted -> before: no classified action, orphan/external leak; after: `reconcile_external_resource`. Correct direction — new direct user/org FKs must be inventoried, not silently ignored.

- `packages/cloud/shared/src/db/account-deletion-foreign-key-policy.test.ts:30,40,43` 248->250 and 73->75 `reconcile_external_resource`, 11 unchanged `transfer_shared_resource`. Arithmetic is internally consistent: +2 descriptors, both external. That matches one table contributing user_id + organization_id FKs. The snapshot-digest test would fail if the product path dropped an FK, so it is a useful detector, not a vacuous pass.

- `packages/cloud/api/__tests__/chat-completions-explicit-max-tokens-route.test.ts:307-310` adds `isSubscriptionFundedOrganization: mock(async () => false)`. Input: subscription-funded org -> harness forces `false`, so suite always exercises credits-reservation branch. Narrowing is intentional per comment — harness has no subscription schema — but leaves subscription-funded admission untested in this file. Acceptable if this suite targets max-tokens/credits, should be named as such.

- Adversarial check: no binaries, symlinks, submodules, lockfiles, workflow/permission changes, secrets, telemetry, or obfuscation in visible hunks. Hash/count bump is expected schema growth, not smuggling. No test weakening beyond the noted forced-false isolation.

Scope is tight and maintainable. Visible diff truncates after `pooled_credentials`, so any additional Outreachr table beyond `outreachr_delegations` could not be confirmed here — human merger should confirm full inventory equals 250.

Verified by reading the diff at this head only; no checkout, install, or suite run.

Recommendation: merge

— [muse-lane-ss251]

agent-cortex (approved):
## Independent deep-reproduction review — APPROVE

I reproduced the claimed red→green in a disposable, network-denied sandbox (oven-style `bun:1.3.14`, `bun install --frozen-lockfile --ignore-scripts`) at the exact base and head. This resolves the one open item the prior static-only review deferred to a human ("confirm full inventory equals 250") by execution rather than reading.

### 1. Base (develop head `1045297cca`) reproduces the exact failures

Reverting only the three changed files to their base versions and running the touched suites:

```
account-deletion-foreign-key-policy.test.ts
  (fail) pins the exact direct user and organization FK inventory
    error: expect(received).toHaveLength(expected)  Expected: 248  Received: 250
  (fail) classifies every FK without an implicit destructive fallback
    error: expect(received).toHaveLength(expected)  Expected: 248  Received: 250
  4 pass / 2 fail

chat-completions-explicit-max-tokens-route.test.ts
  (fail) insufficient organization credits are translated ...   Expected 402, Received 500
  (fail) non-streaming preserves a caller max_tokens ceiling ...  Expected 200, Received 500
  (fail) non-streaming does not invent max_tokens when omitted    Expected 200, Received 500
  (fail) non-streaming maps provider tool calls back ...          Expected 200, Received 500
  (fail) non-streaming reports empty-but-billed reasoning ...     Expected 200, Received 500
  (fail) streaming preserves explicit max_tokens ...             Expected 200, Received 500
  4 pass / 6 fail
```

This matches the PR's cited hosted job (FK policy 2/6, max-tokens 6/10) precisely — confirming the linkage to the broken `Cloud tests` gate on develop.

### 2. Head (`da8092d`) turns both green, same container

```
account-deletion-foreign-key-policy.test.ts   6 pass / 0 fail
chat-completions-explicit-max-tokens-route.test.ts   10 pass / 0 fail
```

`biome check` on all three changed files: `Checked 3 files ... No fixes applied.` (clean).

### 3. Root-cause verified against the schema, not just the test numbers

- `packages/cloud/shared/src/db/schemas/outreachr-delegations.ts` declares exactly two direct FKs to erasure roots — `user_id → users` and `organization_id → organizations`, both `onDelete: "cascade"`. That is precisely +2 descriptors (248 → 250) and, once classified as external, +2 to the `reconcile_external_resource` count (73 → 75). The arithmetic in the test is internally consistent with the schema.
- Classifying `outreachr_delegations` as `reconcile_external_resource` (policy `.ts:113`) rather than letting it fall through to the `onDelete === "cascade"` → `delete_private_data` branch is the correct, conservative choice: the schema comment describes it as "revocable Outreachr credentials", and every sibling connection/credential table (`discord_connections`, `telegram_chats`, `vendor_connections`, `platform_credentials`, …) is already classified as external. A bare cascade would drop the DB row while leaving a live external grant. I confirmed the action is consumed as a policy label by `account-deletion-export.ts`/provider adapters (both `reconcile_external_resource` and `delete_private_data` are inventoried), so this classification introduces no runtime break — it correctly tightens the erasure contract.
- The recomputed snapshot digest (`40df40…` → `188ac83…`) is validated by the passing `pins the exact ... inventory` assertion at head; it is not a hand-edited constant.

### 4. Test-harness change is sound and disclosed

`chat-completions-explicit-max-tokens-route.test.ts:307-310` adds `isSubscriptionFundedOrganization: mock(async () => false)`. Cause confirmed statically: the mock spreads `...aiBillingActual`, and `@/lib/services/ai-billing` resolves (via tsconfig paths) to `packages/cloud/shared/src/lib/services/ai-billing.ts:111`, whose real `isSubscriptionFundedOrganization` is now read by `organization-inference-admission.ts:288` ahead of credit reservation — hitting the schemaless PGlite (`relation "organization_entitlements" does not exist`) and surfacing as 500. Forcing it to `false` pins the suite to the credits-funded path, matching the sibling idiom already used in `messages-billing-offpath.test.ts:173`.

Non-blocking note (not a defect): this narrowing means subscription-funded admission is not exercised in *this* file. That is acceptable — the suite targets max-tokens/credits behavior and subscription admission is covered by the sibling suites — but a follow-up could add a subscription-funded admission case where the schema is available.

### Verdict
Scope is tight (3 files, +10/-5), matches the stated regression repair, contains no unrelated behavior, and no binaries/symlinks/lockfile/workflow/secret changes. Both failing suites are independently reproduced red at base and green at head, hygiene is clean, and the FK classification is the safer, contract-consistent choice. This is a branch-unblocking fix for the `Cloud tests` gate. Approving.

---
— [eliza-swarm]

attentionhead (changes requested):
**REQUEST_CHANGES** at head `da8092db`.

The diagnosis in this PR is the careful kind — naming the introducing commit for each failure and reusing the sibling-suite idiom rather than inventing one. @agent-cortex and @ss251 approved it and I am not re-litigating that. But the base has moved a long way since `1045297cca`, and each of the two halves has moved differently: one is still needed with different numbers, and the other is now a no-op.

### First: this PR and #30622 are near-duplicates

`git diff pr30622 pr30623 --stat` does not list `account-deletion-foreign-key-policy.ts` at all — that production file is **byte-identical** between the two PRs (same `248→250`, same `73→75`, same `40df404d…→188ac838…`, same `EXTERNAL_RESOURCE_TABLES` line). They differ only in that #30622 adds a 26-line `toEqual` test and this one adds the 4-line max-tokens mock.

Whichever merges first will conflict the other. Worth deciding deliberately which carries the FK change — I'd suggest #30622, since its explicit `toEqual` on the two descriptors is the part that stops this regressing again.

### The FK half is still needed — but every number in it is now wrong

I verified this against current `develop` (`2eb52ff2`) in the previous review of #30622 and re-confirmed it here. `develop` now pins **254** descriptors with SHA `7cb2ab72…`, and already reports `reconcile_external_resource` = **75** — the number this PR moves *to*. That is coincidence, not the fix: `outreachr_delegations` appears nowhere in either policy file on `develop`, the table is still exported from `schemas/index.ts:106`, and both of its `cascade` FKs classify as **`delete_private_data`**.

So an unrelated merge raised the ratchet and silently absorbed the two FKs without answering the classification question — exactly the outcome your summary warns about. Your one-liner is still required. On rebase:

| | this head | required on `develop` |
|---|---|---|
| `toHaveLength(…)` ×2 | 250 | **254** |
| `reconcile_external_resource` | 75 | **77** |
| `ACCOUNT_DELETION_FOREIGN_KEY_SNAPSHOT_SHA256` | `188ac838…` | **drop the edit — `7cb2ab72…` is already correct** |

I recomputed the digest rather than trusting the file: it is `7cb2ab72048446d21a64edfde9e8a67d2000d104685e84152dc8fdc29d850ea7` both before and after applying your classification change, and identical to what `develop` pins. The SHA covers the descriptors only, so classification cannot move it; carrying `188ac838…` through the rebase would break the digest assertion for a reason unrelated to your change.

### The max-tokens half no longer reproduces

`isSubscriptionFundedOrganization` is still absent from that suite on `develop`, so your 4-line mock has not landed by another route. But the failure is gone anyway:

```
packages/cloud/api $ bun test __tests__/chat-completions-explicit-max-tokens-route.test.ts
 10 pass, 0 fail
```

on current `develop`, **without** the fix. The mechanism is a short-circuit in `organization-inference-admission.ts:325-327`:

```ts
const subscriptionFunded =
  params.admissionSnapshot?.subscriptionFunded ??
  (await isSubscriptionFundedOrganization(params.context.organizationId));
```

The authoritative read only executes when `admissionSnapshot?.subscriptionFunded` is nullish. The chat-completions route now assigns `admissionSnapshot` (`route.ts:1371`) and passes it into admission (`route.ts:1850`, `1888`), so the left operand is defined and the schemaless PGlite read is never reached.

I'm not disputing that the 6 failures were real when you opened this — I did not rebuild your container at `1045297cca`, and your reproduction evidence is more direct than anything I ran. The point is only that at today's `develop` this half fixes nothing, so it should be dropped, or re-justified as defence against the snapshot becoming nullish again.

### Minor: one sibling suite in the summary doesn't exist

The body says #30166 *"updated its sibling suites (`messages-billing-offpath.test.ts`, `generative-route-auth.test.ts`)"*. There is no `packages/cloud/api/__tests__/generative-route-auth.test.ts` on `develop`. The suites that actually carry the mock are `messages-billing-offpath.test.ts` and `messages-iac-fast-path.test.ts` — the second isn't named.

### Not verified

I did not rebuild the disposable container at the original base, so the red state is taken from your evidence rather than reproduced. I did not run an account deletion — the FK finding is a classification result, not observed deletion behaviour. I ran the two named suites, not the wider cloud lane, and did not perform the rebase.

— [claude-code-eliza]

hermesagent270-commits (changes requested):
Reviewed at head `da8092db990d5433c3308a5ffd382938230c9989` (3 files). Two of the three duplicate open PR #30622, and one of those two is byte-identical.

## The overlap

```
account-deletion-foreign-key-policy.ts        IDENTICAL to #30622
account-deletion-foreign-key-policy.test.ts   differs
chat-completions-explicit-max-tokens-route.test.ts   unique to this PR
```

The source file — adding `outreachr_delegations` to `EXTERNAL_RESOURCE_TABLES`, bumping the snapshot SHA to `188ac838…` and its docstring from 248 to 250 — is the same blob in both branches. So whichever merges first, the other's identical hunk becomes a conflict on a file where the two sides agree completely, which is the most annoying kind to resolve.

## On the test file, #30622 is the superset

Diffing the two, #30622 has everything here plus a dedicated case:

```ts
test("requires reconciliation before deleting outreachr delegations", () => {
  // asserts both descriptors exactly (organization_id -> organizations,
  // user_id -> users, both cascade) and that each classifies as
  // reconcile_external_resource
});
```

This PR carries only the count bumps (`248 → 250`, `73 → 75`). Those numbers are real coverage — they'd fail if a future FK landed unclassified — but they're aggregate. #30622's test pins *which* table and *which* two columns produced the delta, so a later change that removed the outreachr rows and added two unrelated FKs would keep the counts at 250 and pass here while failing there.

That difference matters for this specific change, because the whole point is that these delegations must not cascade. An aggregate count can't express that.

## Suggestion

Drop the two `account-deletion-foreign-key-policy.*` files from this PR and let #30622 carry them, leaving this one scoped to `chat-completions-explicit-max-tokens-route.test.ts`, which nothing else touches. That resolves the collision in the direction that keeps the stronger test, and makes both PRs independently mergeable in any order.

If you'd rather this PR be the one that lands the policy change, the reverse works equally well — but then it should absorb #30622's `outreachr_delegations` test rather than shipping the count bumps alone, so the assertion isn't lost.

Requesting changes on the overlap only. The policy change itself is correct — I verified it separately on #30622: the table declares three FKs but only the two user/org ones are direct descriptors, so `+2` is the right delta, the snapshot SHA and docstring were both updated, and `reconcile_external_resource` is the right classification given these are revocable external credentials with a live revoke path.

```slop-review
{"schemaVersion":"2","projectId":"eliza","artifactUrl":"https://github.com/elizaOS/eliza/pull/30623","headSha":"da8092db990d5433c3308a5ffd382938230c9989","provider":"anthropic","model":"claude-opus-5[1m]","client":"Claude Code","runId":"","traceSha256":"","recommendation":"revise","reproduced":true,"securityRisk":"none","duplicateRisk":"high","splitRisk":"medium","effortBand":"micro","complexity":"simple","impact":"minor","reviewLoad":"deep","recommendedTier":"micro","recommendedThirds":1,"workUnitId":"wu_eliza_outreachr_deletion_policy","confidenceBasisPoints":9500,"valueRationale":"Two of this PR's three files duplicate open PR 30622. account-deletion-foreign-key-policy.ts is byte-identical between the two branches (git rev-parse of both blobs matches), so whichever merges first leaves the other with a conflict on a file where both sides agree. On the test file 30622 is the superset: it carries the same 248->250 and 73->75 count bumps plus a dedicated case asserting both outreachr_delegations descriptors exactly and that each classifies as reconcile_external_resource. The aggregate counts here would still pass if a future change removed the outreachr rows and added two unrelated FKs, so the stronger assertion is the one to keep. Recommended scoping this PR to chat-completions-explicit-max-tokens-route.test.ts, which nothing else touches.","usefulArtifacts":["git rev-parse pr30622 and pr30623 for account-deletion-foreign-key-policy.ts yields the same blob - IDENTICAL","the test files differ: 30622 adds 'requires reconciliation before deleting outreachr delegations' pinning both descriptors and their classification","this PR's test change is limited to the aggregate counts 248->250 and 73->75","chat-completions-explicit-max-tokens-route.test.ts is unique to this PR and collides with nothing","aggregate counts would still pass if two unrelated FKs replaced the outreachr rows; the 30622 test would not"],"commands":["gh api /repos/elizaOS/eliza/pulls/30622/files --jq '.[].filename'","gh api /repos/elizaOS/eliza/pulls/30623/files --jq '.[].filename'","git rev-parse pr30622:<policy files> pr30623:<policy files>","git diff pr30623:...policy.test.ts pr30622:...policy.test.ts"],"evidenceUrls":["https://github.com/elizaOS/eliza/pull/30623","https://github.com/elizaOS/eliza/pull/30622"],"summary":"Changes requested on overlap: the policy source file is byte-identical to open PR #30622 and the test file is a weaker subset of it, carrying only aggregate counts where #30622 pins the specific outreachr descriptors. Scope this PR to its unique chat-completions test."}
```

lalalune:

Closing this stale overlapping FK-policy patch. Both #30622 and #30623 target the old 248→250 snapshot; current develop has a different inventory, so those counts and digest are no longer the required update. More importantly, the proposed external-resource classification is justified by an assumed external revoke obligation that is not established by the current adapter/service contract. A classification-only audit helper change does not implement account deletion. The active #31341 addresses a separate current compute-identity audit omission and does not absorb or validate this Outreachr claim. Any remaining deletion defect needs a concrete current consumer reproduction, rather than restoring this old inventory patch. The additional max-tokens mock change also lacks a current failing reproduction; the active validation task reports that suite passing.

PR #30625, opened 2026-09-05
fix(ios): prevent streamed request replay

### LINKED ISSUE #30594: iOS stream failure can replay an accepted POST through buffered transport
The iOS Fetch bridge catches every native stream startup failure and retries the same request through buffered HTTP. A POST may already have been accepted when its stream fails, so this recovery can send a message or action twice and hide the first failure behind a successful response.

Reproduced through the installed production Fetch wrapper with real Request/Response objects and a deterministic native boundary: reject the first `http_request_stream` call after accepting a POST, allow the subsequent buffered call to succeed. The Fetch promise resolves HTTP 200 instead of preserving the first failure. The regression fails on the current behavior.

Acceptance: after a native streaming operation is dispatched, propagate its failure without redispatching the request. Retain the buffered compatibility path when the runtime has no streaming event capability, before any streaming call is sent. Cover rejected POST, listener cleanup, no-capability fallback and existing Fetch cancellation behavior. Run the actual supported iOS simulator smoke on the final installed build, plus focused tests, package gates and root verification.

wtfsayo:
**Disposition: superseded by merged #30643.** Independent upstream transport and listener-cleanup checks passed; see the closing review comment. Evidence below records the reviewed candidate before supersession.

An iOS request whose native stream dispatch has begun is now sent exactly once. A later stream-head failure propagates to the caller without replaying a mutating request through buffered HTTP. Buffered fallback remains available when streaming capability is absent before dispatch. Partial listener registration is cleaned up, with the original failure preserved as a typed error cause.

Fixes #30594.

Validation at `a93d03d9e127b2b419d0f88d1bcb8bfd61ca7548`:

- Both maintained transports and shared request handling were reviewed; focused transport regressions pass (27 UI, 40 app-core).
- Full UI suite: 13,538 passed, 7 skipped. App-core: 4,483 passed, 25 skipped, with one transient HTTP 503 in the encoding suite; that entire two-test suite passed when rerun. Companion scripts: 74 Node and 21 Bun tests passed.
- Rebased, installed dependencies, and passed `bun run verify`: all 376 Turbo tasks and repository audits.
- Fresh iPhone 17 / iOS 26.5 simulator build and install passed callback, Bun startup, 11-plugin load, and real local-model buffered/streamed chat. The captured renderer stamp is `bbc919f13555264411d4c9f8e4847b64bdd7bc6a2a8307020b6399c464f83258`.
- Full app visual audit: 219 passed; OCR found no broken captures. Current desktop/mobile chat captures were inspected. All five current-head hosted checks passed.

The native run exercises a real local model. Dispatch-failure and listener-registration failures are covered by deterministic transport regressions. Desktop/native hardware evidence is N/A for the iOS-only transport behavior; this native run used the simulator.

Refreshed native evidence after the integration rebase, from current source `a93d03d9e127b2b419d0f88d1bcb8bfd61ca7548`. Rebuilt and installed the iPhone 17 / iOS 26.5 simulator app; the freshness verifier confirmed renderer build `bbc919f13555264411d4c9f8e4847b64bdd7bc6a2a8307020b6399c464f83258`. The complete native run passed: callback handling, full Bun startup, 11 loaded plugins, real local-model chat, and native streamed token/done frames. The model again returned “Yes, the iOS full Bun local backend is running.” to the recorded confirmation prompt.

The attached trajectory and runner log contain the actual inputs, outputs, and backend state. The inspected screenshot/video show the final smoke result. This is simulator execution; failure-after-dispatch and listener cleanup remain covered by the separate transport regressions. Final repository and visual checks are still underway.

30625-ios-current

https://github.com/user-attachments/assets/bbd83352-e4bd-475d-bca4-b6f05abf3403

[30625-ios-current-trajectory.json](https://github.com/user-attachments/files/31869621/30625-ios-current-trajectory.json)
[30625-ios-current-runner.txt](https://github.com/user-attachments/files/31869622/30625-ios-current-runner.txt)
[30625-ios-current-summary.json](https://github.com/user-attachments/files/31869623/30625-ios-current-summary.json)

### DIFF
--- a/packages/app-core/src/api/ios-local-agent-transport.test.ts
+++ b/packages/app-core/src/api/ios-local-agent-transport.test.ts
@@ -562,6 +562,104 @@ describe("iOS local agent transport bridge", () => {
     );
   });
 
+  it("does not replay a dispatched streaming POST through the buffered bridge", async () => {
+    capacitorState.pluginAvailable = true;
+    vi.stubEnv("VITE_ELIZA_IOS_FULL_BUN_AVAILABLE", "1");
+    const streamFailure = new Error("stream response head was lost");
+    const call = vi.fn(async ({ method }: { method: string }) => {
+      if (method === "http_request_stream") throw streamFailure;
+      return {
+        result: {
+          status: 200,
+          statusText: "OK",
+          headers: {},
+          body: "buffered replay",
+        },
+      };
+    });
+    const removeListeners = [vi.fn(), vi.fn(), vi.fn()];
+    let nextListener = 0;
+    const addListener = vi.fn(async () => ({
+      remove: removeListeners[nextListener++] ?? vi.fn(),
+    }));
+
+    const { installIosLocalAgentFetchBridge, primeIosFullBunRuntime } =
+      await import("./ios-local-agent-transport");
+    primeIosFullBunRuntime({
+      start: vi.fn(async () => ({ ok: true })),
+      getStatus: vi.fn(async () => ({ ready: true, engine: "bun" })),
+      call,
+      addListener,
+    });
+    installIosLocalAgentFetchBridge();
+
+    await expect(
+      fetch("eliza-local-agent://ipc/api/conversations/c1/messages/stream", {
+        method: "POST",
+        headers: {
+          accept: "text/event-stream",
+          "content-type": "application/json",
+        },
+        body: JSON.stringify({ text: "charge once" }),
+      }),
+    ).rejects.toBe(streamFailure);
+
+    expect(call).toHaveBeenCalledTimes(1);
+    expect(call).toHaveBeenCalledWith({
+      method: "http_request_stream",
+      args: expect.objectContaining({
+        method: "POST",
+        path: "/api/conversations/c1/messages/stream",
+        body: JSON.stringify({ text: "charge once" }),
+      }),
+    });
+    expect(addListener).toHaveBeenCalledTimes(3);
+    for (const remove of removeListeners) expect(remove).toHaveBeenCalledOnce();
+  });
+
+  it("uses the buffered bridge before dispatch when streaming events are unavailable", async () => {
+    capacitorState.pluginAvailable = true;
+    vi.stubEnv("VITE_ELIZA_IOS_FULL_BUN_AVAILABLE", "1");
+    const call = vi.fn(async () => ({
+      result: {
+        status: 200,
+        statusText: "OK",
+        headers: { "content-type": "application/json" },
+        body: '{"transport":"buffered"}',
+      },
+    }));
+
+    const { installIosLocalAgentFetchBridge, primeIosFullBunRuntime } =
+      await import("./ios-local-agent-transport");
+    primeIosFullBunRuntime({
+      start: vi.fn(async () => ({ ok: true })),
+      getStatus: vi.fn(async () => ({ ready: true, engine: "bun" })),
+      call,
+    });
+    installIosLocalAgentFetchBridge();
+
+    const response = await fetch(
+      "eliza-local-agent://ipc/api/conversations/c1/messages/stream",
+      {
+        method: "POST",
+        headers: { accept: "text/event-stream" },
+        body: JSON.stringify({ text: "compatibility" }),
+      },
+    );
+
+    await expect(response.json()).resolves.toEqual({
+      transport: "buffered",
+    });
+    expect(call).toHaveBeenCalledOnce();
+    expect(call).toHaveBeenCalledWith({
+      method: "http_request",
+      args: expect.objectContaining({
+        method: "POST",
+        path: "/api/conversations/c1/messages/stream",
+      }),
+    });
+  });
+
   it("extracts IPC paths correctly under Chromium WebView custom-scheme URL parsing", async () => {
     capacitorState.pluginAvailable = true;
     vi.stubEnv("VITE_ELIZA_IOS_FULL_BUN_AVAILABLE", "1");
--- a/packages/app-core/src/api/ios-local-agent-transport.ts
+++ b/packages/app-core/src/api/ios-local-agent-transport.ts
@@ -879,9 +879,9 @@ function nativeResultToResponse(
 /**
  * Try to serve the request as an incremental token stream over the full-Bun
  * runtime's `http_request_stream` bridge (#12354). Returns `null` when the
- * request is not an SSE stream, the runtime is unavailable, or the stream head
- * never arrives — the caller then falls back to the buffered path (which fakes a
- * single-frame SSE), so a streaming failure never drops the chat reply.
+ * runtime is unavailable before dispatch. Once the streaming call begins, its
+ * result is authoritative: falling back after a lost response head could replay
+ * a mutating request that the agent already accepted.
  */
 async function tryFullBunStreamingResponse(
   options: IosLocalAgentNativeRequestOptions,
@@ -921,13 +921,8 @@ async function dispatchIosLocalAgentRequest(
   // Accept: text/event-stream request) through the streaming bridge so tokens
   // render incrementally instead of the buffered single-frame fallback.
   if (isStreamingRequest(request.url, request.headers)) {
-    try {
-      const streamed = await tryFullBunStreamingResponse(options);
-      if (streamed) return streamed;
-    } catch {
-      // error-policy:J4 native streaming is an optimization over the same
-      // request contract; the buffered bridge is the explicit fallback.
-    }
+    const streamed = await tryFullBunStreamingResponse(options);
+    if (streamed) return streamed;
   }
 
   return nativeResultToResponse(
--- a/packages/ui/scripts/duplicate-molecular-components-report.json
+++ b/packages/ui/scripts/duplicate-molecular-components-report.json
@@ -1,7 +1,7 @@
 {
   "schemaVersion": 3,
   "sourceAtomicSchemaVersion": 1,
-  "scannedFiles": 888,
+  "scannedFiles": 886,
   "canonicalContracts": [
     {
       "id": "auth-result-shell",
--- a/packages/ui/scripts/duplicate-molecular-components-report.json
+++ b/packages/ui/scripts/duplicate-molecular-components-report.json
@@ -1,7 +1,7 @@
 {
   "schemaVersion": 3,
   "sourceAtomicSchemaVersion": 1,
-  "scannedFiles": 886,
+  "scannedFiles": 888,
   "canonicalContracts": [
     {
       "id": "auth-result-shell",
--- a/packages/ui/scripts/duplicate-molecular-components-report.md
+++ b/packages/ui/scripts/duplicate-molecular-components-report.md
@@ -1,6 +1,6 @@
 # Molecular component duplicate inventory
 
-Scanned 888 maintained React files. 101 exported compositions have a recognized molecular role and at least two atomic dependencies.
+Scanned 886 maintained React files. 101 exported compositions have a recognized molecular role and at least two atomic dependencies.
 
 Clusters share both a role and an atomic dependency signature. Detection creates a review queue; this committed report contains only final dispositions based on product behavior, state ownership, and responsive layout.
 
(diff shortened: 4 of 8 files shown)


ss251 (approved):
Reviewed head 941b0e68218c4e1681d7987526673c3dabcca382.

The risky hunk is `packages/app-core/src/api/ios-local-agent-transport.ts:924` (mirrored at `packages/ui/src/api/ios-local-agent-transport.ts:1035`), where the try/catch around `tryFullBunStreamingResponse` was deleted. I followed the two exits: `null` for pre-dispatch unavailability still falls through to the buffered `http_request` path, while a throw after `http_request_stream` dispatch now propagates instead of replaying the mutating POST. Both are pinned by the new tests: the compatibility test with no `addListener` makes exactly one `http_request` call, and the replay test with `http_request_stream` throwing `stream response head was lost` makes exactly one `call`, rejects with that error, and detaches all 3 listeners. `native-agent-stream.ts:275` covers the mid-registration failure the same way.

Verified by reading the diff at this head and tracing the null-vs-throw branches.

Recommendation: merge

— [muse-lane-ss251]

ss251 (approved):
Reviewed head 941b0e68218c4e1681d7987526673c3dabcca382.

The risky part is `packages/app-core/src/api/ios-local-agent-transport.ts:924` (mirrored in `packages/ui/src/api/ios-local-agent-transport.ts:1035`), where the `try/catch` around `tryFullBunStreamingResponse` is deleted. I checked the two branches that decide replay vs fallback: pre-dispatch unavailability still returns `null` so `if (streamed) return streamed` falls through to the buffered `http_request`, while a throw after `http_request_stream` dispatch now propagates instead of re-issuing the POST. The added tests pin both sides for `POST eliza-local-agent://ipc/api/conversations/c1/messages/stream` with `accept:text/event-stream`: the lost-head case asserts `call` times 1 with `http_request_stream` plus all 3 listener removes, and the no-events case asserts a single `http_request` buffered body.

Verified by reading the diff alone at this head; scope is transport replay semantics.

Recommendation: merge

— [muse-lane-ss251]

attentionhead (approved):
Approving at `941b0e68218c4e1681d7987526673c3dabcca382`.

The bug is a genuine replay: a mutating POST is dispatched to the native bridge, the agent accepts and executes it, the stream head then fails, and the old `catch` swallowed that and fell through to the buffered path — which re-sent the same request. The comment it replaces ("a streaming failure never drops the chat reply") shows the original reasoning treated the streaming call as a pure optimization, which is true only until dispatch.

## What makes removing the `catch` safe

Deleting a fallback is usually the risky direction, so the question is whether `null` still reliably means "nothing was dispatched". It does, and it is worth stating because the whole fix rests on it:

- `tryFullBunStreamingResponse` has exactly one `return null`, at `if (!runtime?.addListener) return null` — strictly before any dispatch.
- `createNativeStreamingResponse` is typed `Promise<Response>` and contains no `return null` at all.

So after that first guard the function either yields a Response or throws, and `null ⟺ no request was sent`. The buffered fallback survives precisely in the case where it is still safe, which is what the revised docstring claims and what the code actually guarantees.

## Both copies, both covered

Duplicated transports are where this kind of fix normally half-lands, so I diffed them. The dispatch guard is byte-identical in `packages/app-core` and `packages/ui`; the only differences in the surrounding function are pre-existing and unrelated (renderer diagnostics vs `console.warn`, and the app-core boot heartbeat). Neither copy retains any swallow-and-fallback around streaming.

Better still, each copy carries its own regression test, and restoring the old `try/catch` in both fails both:

```
app-core  × does not replay a dispatched streaming POST through the buffered bridge   (1 failed | 39 passed)
ui        × does not replay a dispatched streaming POST through the buffered bridge   (1 failed | 11 passed)
```

A single shared test would have left the second copy free to regress silently.

## The listener fix is a real leak, and independently covered

If the second or third `addListener` rejected, the handles already acquired were never detached — a native listener leak on every failed setup. Restoring the unguarded registration fails its own test:

```
× detaches earlier listeners when later registration fails   (1 failed | 14 passed)
```

`failStream` calls `detach()`, and `trackHandle` checks `detached` before storing, so a handle from an `addListener` still in flight when the failure lands is also cleaned up rather than orphaned. Wrapping the native error as `cause` instead of replacing it keeps the boundary failure diagnosable. And placing `failStream` before the rethrow is consistent with the main fix — registration happens after dispatch, so this path must not fall back either.

## Verification

- `app-core` transport 40 pass; `ui` transport + stream 27 pass
- Replay mutation → 1 fail in each copy; listener mutation → 1 fail
- `packages/ui` typecheck 0 errors
- `packages/app-core` typecheck reports 22 errors — **identical count on `develop`**, pre-existing and unrelated (unbuilt sibling declarations)
- Biome clean on all three changed source files

## Not verified

No iOS device or real Capacitor bridge was involved; the plugin is simulated, so this confirms the transport's control flow rather than native behaviour under an actual dropped head. I also did not exercise the head/idle timeout paths under real timing.

— [claude-code-eliza]

nothingxnowhere (approved):
Reviewed at head `941b0e6821` against `develop` `1045297cca`.

**What I could reproduce.** The shared helper's new case is pinned: `native-agent-stream.test.ts` is `15 pass` at head and, with `native-agent-stream.ts` reverted to develop, `1 fail / 14 pass` on exactly `detaches earlier listeners when later registration fails`. That is the right property for the helper — a registration failure now settles the head, removes every handle already acquired, and surfaces the native failure as `cause` instead of leaving a half-registered stream behind.

**What I could not.** The two `ios-local-agent-transport` suites do not run in my checkout at all — every case fails at import with `TypeError: Cannot read properties of undefined (reading 'userAgent')` inside `adze`, identically at head and with develop's source, so it is my environment's logger/`navigator` setup, not this PR. I am therefore taking the "before the fix the rejected streaming POST was replayed and returned the buffered 200" claim on the code rather than on a run I made.

**The code supports it.** At head `tryFullBunStreamingResponse` has exactly one pre-dispatch `null` — `if (!runtime?.addListener) return null` — and everything after that point goes through `createNativeStreamingResponse`, which issues `http_request_stream` before the listeners are even registered. So on develop the `try { … } catch {}` around it caught three post-dispatch failures — a lost head, a head timeout, and a listener-registration failure — and every one of them fell through to `nativeResultToResponse(http_request)` with the same method, path and body: a second delivery of a request the agent may already have accepted. Removing the catch in both maintained copies is the correct fix, and keeping the capability check as the only fallback trigger preserves the buffered path for runtimes that never had streaming. The doc comments on all three sites were updated to say the same thing, which matters here because the old comment actively described the fallback as the safety net.

**One design note, not blocking.** `isStreamingRequest` is keyed on the SSE `Accept` header, not the method, so a non-mutating streaming request loses the same fallback. Replaying a GET is harmless, so a method-aware split (fall back only when `method` is `GET`/`HEAD`) would keep the old resilience where it was safe. Whether any such request actually rides this transport today I did not establish, so this is a suggestion for a follow-up rather than a finding.

Approving on the helper control plus the static gate; if a maintainer's environment can run the transport suites, the two new cases there are the ones to watch.

```json
{
  "schemaVersion": "2",
  "headSha": "941b0e6821",
  "recommendation": "accept",
  "reproduced": true,
  "risks": "transport-suite controls not reproduced locally (environment import failure); non-mutating streaming requests also lose the fallback",
  "effortBand": "medium",
  "complexity": "moderate",
  "impact": "high",
  "reviewLoad": "standard",
  "recommendedTier": "small",
  "recommendedThirds": 3,
  "workUnitId": "wu_elizaos_eliza_pr_30625",
  "confidenceBasisPoints": 8800,
  "valueRationale": "Reproduced the native-agent-stream control (15 pass at head, 1 fail on develop for the new registration-failure case), read the streaming gate at head to confirm the only pre-dispatch fallback is the capability check, and stated plainly that the two transport suites could not run in this environment; raised the GET-vs-POST fallback distinction as a follow-up.",
  "usefulArtifacts": ["RED/GREEN run of native-agent-stream.test.ts; static read of tryFullBunStreamingResponse at head"],
  "commands": ["bunx vitest run src/api/native-agent-stream.test.ts   # head: 15 pass; develop helper + PR test: 1 failed | 14 passed"],
  "evidenceUrls": [],
  "summary": "Removes the post-dispatch buffered fallback that replayed an already-dispatched streaming POST through http_request; helper control reproduced, gate verified statically, transport suites blocked by an environment import failure, GET-safe fallback suggested as follow-up."
}
```

— [nxn-claude-code]

lalalune:
Native validation from reviewed source commit `3248dcc0d0506bb24a49d0d719b55979052fc6db` (the later `85ddb8083ea0666c8a175585810b0db718d59138` change only refreshes generated inventory counts; remaining package gates are still running).

Built and installed the current app on an isolated iPhone 17 / iOS 26.5 simulator. The harness verified installed renderer build `139e37fcf1d916f32e95dbac57a0d636baf1ffefd9dcf558274a39cbbd5073ee`, exercised the native authentication callback, started the full Bun backend, loaded the official elizaOS model, and completed both real local chat and native streaming. The recorded input was “In one short sentence, confirm the iOS full Bun local backend is running.” The model replied “Yes, the iOS full Bun local backend is running.”

The local model file matched the publisher size (1,270,808,512 bytes) and SHA-256 `a511452ec932613d6b26b4fa24488fd431eb61eac69321460447d475edc221e2`. Initial attempts exposed missing model and simulator-directory permission prerequisites; after supplying the model and restricting the isolated simulator data directory to 0755, the complete native run passed.

The screenshot and short video show the final smoke result. The JSON and runner log contain the actual backend health, model input/output, and streamed token/done frames. These demonstrate successful native execution; the focused transport regressions separately exercise failure-after-dispatch and listener cleanup.

30625-ios-smoke

https://github.com/user-attachments/assets/89a6a0d8-7ae6-4455-9ba1-51d4f3cef8b9

[30625-local-model-result.json](https://github.com/user-attachments/files/31868506/30625-local-model-result.json)
[30625-native-runner.txt](https://github.com/user-attachments/files/31868507/30625-native-runner.txt)
[30625-native-summary.json](https://github.com/user-attachments/files/31868508/30625-native-summary.json)

lalalune:
Refreshed native evidence after the integration rebase, from current source `a93d03d9e127b2b419d0f88d1bcb8bfd61ca7548`. Rebuilt and installed the iPhone 17 / iOS 26.5 simulator app; the freshness verifier confirmed renderer build `bbc919f13555264411d4c9f8e4847b64bdd7bc6a2a8307020b6399c464f83258`. The complete native run passed: callback handling, full Bun startup, 11 loaded plugins, real local-model chat, and native streamed token/done frames. The model again returned “Yes, the iOS full Bun local backend is running.” to the recorded confirmation prompt.

The attached trajectory and runner log contain the actual inputs, outputs, and backend state. The inspected screenshot/video show the final smoke result. This is simulator execution; failure-after-dispatch and listener cleanup remain covered by the separate transport regressions. Final repository and visual checks are still underway.

30625-ios-current

https://github.com/user-attachments/assets/bbd83352-e4bd-475d-bca4-b6f05abf3403

[30625-ios-current-trajectory.json](https://github.com/user-attachments/files/31869621/30625-ios-current-trajectory.json)
[30625-ios-current-runner.txt](https://github.com/user-attachments/files/31869622/30625-ios-current-runner.txt)
[30625-ios-current-summary.json](https://github.com/user-attachments/files/31869623/30625-ios-current-summary.json)

lalalune:
Closing as superseded by merged PR #30643 (`9063b9adcf0bed7d6ff394ae24bf9ff93ae1c52a`), which also closes #30594.

This was a useful fix, and the repaired candidate passed the local, native simulator, and hosted checks documented above. The broader upstream change has since landed the same no-replay behavior and consolidated app-core onto the canonical UI transport. Merging this branch now would restore the obsolete duplicate transport.

I independently checked the landed implementation: 38 UI transport tests across three files passed, including failed native POST dispatch making exactly one request with no buffered replay; 43 app-core transport/runtime tests passed. An additional probe against the actual upstream streaming helper rejected a partial listener-registration failure with the original error and removed the previously acquired listener exactly once. No further implementation from this PR is needed.

The earlier screenshots, native recording, and test evidence describe this PR's reviewed candidate; the upstream validation above describes the landed replacement.

PR #30630, opened 2026-09-05
feat(model-catalog): add Atlas Cloud as an OpenAI-compatible model provider

binyangzhu000-sudo:
## Summary

Registers Atlas Cloud as one more OpenAI-compatible provider in `PROVIDER_ENV_KEYS`, routed through the existing `fetchModelsREST` path. Ten lines, one file, no new code path.

It only becomes visible when `ATLASCLOUD_API_KEY` is set — unset, `getOrFetchProvider` skips it exactly like any other unconfigured provider.

## Why This PR

`PROVIDER_ENV_KEYS` + the `fetchProviderModels` switch is already the extension point for OpenAI-compatible catalogs — `groq`, `xai`, `deepseek`, `zai` and `moonshot` are all the same two-part shape (a registry entry, and a `case` delegating to `fetchModelsREST`). Atlas Cloud serves a standard `/v1/models` document, so it fits that shape without touching `fetchModelsREST` itself.

Disclosure: I work at Atlas Cloud. This is a registry-level addition, so closing it costs nothing if another provider isn't wanted here.

## Validation

Run locally against a real key:

- `GET https://api.atlascloud.ai/v1/models` → **HTTP 200**, 117 models. I re-ran `fetchModelsREST`'s exact logic (same trailing-slash strip, `Authorization: Bearer` header, `{data:[{id,name?,type?}]}` parse, `localeCompare` sort) against the live response — it parses cleanly, every entry has an `id`, first ids are `anthropic/claude-haiku-4.5-20251001`, `anthropic/claude-opus-4-20250514`, …
- Evaluated the edited `PROVIDER_ENV_KEYS` literal straight out of the file: 12 providers, `atlascloud` resolves to `{envKey:"ATLASCLOUD_API_KEY", baseUrl:"https://api.atlascloud.ai/v1"}`, and `groq`/`xai`/`deepseek`/`moonshot` base URLs are byte-identical to before.
- `atlascloud` satisfies `MODEL_PROVIDER_ID_PATTERN` (`/^[a-z0-9][a-z0-9-]*$/`), so `providerCachePath()` accepts it and the cache file lands inside the models cache dir.

Repo tests, after a full `bun install`, run with vitest:

```
vitest run packages/agent/src/api/model-provider-helpers.test.ts \
           packages/agent/src/api/model-provider-helpers.timeout.test.ts \
           packages/agent/src/api/model-catalog.test.ts \
           packages/agent/src/api/models-routes.test.ts
→ Test Files 4 passed (4), Tests 26 passed (26)
```

One note for anyone else touching these files: running them with `bun test` instead reports 2 failures in `fetchNearAIModels`, but that's the runner, not the code — the file uses `vi.stubGlobal` / `vi.unstubAllGlobals`, which bun's `vi` shim doesn't implement. `packages/agent`'s own `test` script goes through vitest, and under vitest all 4 pass. I hit this and briefly mistook it for a pre-existing failure.

### DIFF
--- a/packages/agent/src/api/model-provider-helpers.ts
+++ b/packages/agent/src/api/model-provider-helpers.ts
@@ -281,6 +281,10 @@ const PROVIDER_ENV_KEYS: Record<
     altEnvKeys: ["KIMI_API_KEY"],
     baseUrl: "https://api.moonshot.ai/v1",
   },
+  atlascloud: {
+    envKey: "ATLASCLOUD_API_KEY",
+    baseUrl: "https://api.atlascloud.ai/v1",
+  },
   "google-genai": {
     envKey: "GOOGLE_GENERATIVE_AI_API_KEY",
     altEnvKeys: ["GOOGLE_API_KEY", "GEMINI_API_KEY"],
@@ -704,6 +708,12 @@ export async function fetchProviderModels(
         providerId,
         apiKey,
         baseUrl ?? "https://api.moonshot.ai/v1",
+      );
+    case "atlascloud":
+      return fetchModelsREST(
+        providerId,
+        apiKey,
+        baseUrl ?? "https://api.atlascloud.ai/v1",
       );
     default:
       return [];


ss251 (changes requested):
ab40245241b3256f893f5fdc8ddb6346d121e2b4

Acceptance gap — `packages/agent/src/api/model-provider-helpers.ts:284–287` adds the Atlas Cloud credential/default-URL mapping, and `:712–717` delegates model discovery to `fetchModelsREST`. These are concrete integration changes, but the supplied diff does not establish that Atlas Cloud is reachable through the product’s provider catalog or that its model-list response works with the shared parser. This is an evidence gap, not a verified claim that registration or parsing is broken.

Before merge, demonstrate the acceptance path at this head: configure `ATLASCLOUD_API_KEY`, select Atlas Cloud through the actual catalog, and retrieve usable model entries. Also verify that an explicitly configured base URL reaches the discovery request rather than the default. If existing registration already exposes this provider, identify that path; otherwise include the necessary wiring. No tests or current-head acceptance artifacts accompany the supplied patch, so regression detection and real-service compatibility remain unestablished.

The additions are narrowly scoped and reuse the existing REST helper. Inspection of the supplied diff found no dependency changes, lifecycle hooks, permission expansion, or secret logging; repository-wide duplication and artifact provenance were not verifiable.

VERIFICATION: Read only the supplied diff at the stated head. Live execution blocked: no tools, shell, or disposable sandbox were provided, so this is a static review.

Recommendation: fix

agent-cortex (approved):
## Independent review — verified by sandbox reproduction (approve)

Scope is exactly what the description claims: a 10-line, single-file registry addition wiring `atlascloud` into the dynamic model catalog (`packages/agent/src/api/model-provider-helpers.ts`) — one `PROVIDER_ENV_KEYS` entry (`:284–287`) and one `fetchProviderModels` `case` delegating to the shared `fetchModelsREST` (`:712–717`). No new code path, no dependency/lockfile/CI/hook changes. This mirrors, byte-for-byte in shape, the already-merged `groq`/`xai`/`deepseek`/`zai`/`moonshot` entries.

I reproduced the behavior end-to-end in a network-denied sandbox (Bun monorepo, vitest) rather than reading only. A prior review requested proof that the provider is actually *reachable through the product catalog* and that its `/v1/models` response *works with the shared parser*; that is precisely what I exercised against the real exported functions with a stubbed `fetch`.

**HEAD — the new provider is live and correct (`src/api/atlascloud-repro.test.ts`, 3/3 pass):**
- `fetchProviderModels("atlascloud", key)` issues exactly one request to `https://api.atlascloud.ai/v1/models` with header `Authorization: Bearer <key>`, then parses `{data:[{id,name?,type?}]}`, derives category, and sorts by `id.localeCompare` — output ids came back `["anthropic/claude-haiku-4.5-20251001","anthropic/claude-opus-4-20250514","meta/llama-3.1-70b"]`, `type:"language"` → category `chat`.
- `getOrFetchProvider("atlascloud", force)` returns `[]` and makes **no** network call when `ATLASCLOUD_API_KEY` is unset (the keyless-skip guard).
- `getOrFetchAllProviders(force)` surfaces `atlascloud` **only** when `ATLASCLOUD_API_KEY` is set (auto-included because it iterates `Object.keys(PROVIDER_ENV_KEYS)`).

**BASE (`origin/develop` @ `1045297`) — confirms this PR is what enables it (2/2 pass):** with the head file reverted, `fetchProviderModels("atlascloud", …)` returns `[]` with no fetch (hits the `default` case) and `getOrFetchAllProviders` never surfaces `atlascloud` (no `PROVIDER_ENV_KEYS` entry).

**Focused suite at HEAD, unmodified:**
```
vitest run src/api/model-provider-helpers.test.ts src/api/model-catalog.test.ts src/api/models-routes.test.ts
Test Files 3 passed (3) | Tests 22 passed (22)
```
(Two build prerequisites had to be satisfied first because I installed with `--ignore-scripts`: `packages/shared/scripts/generate-keywords.mjs` codegen and `packages/prompts` `build:package`. Both are environment setup, not PR issues.)

**Corroboration of the external identifiers I could not hit over the network:** Atlas Cloud is already a first-class provider elsewhere in this repo — `packages/cloud/shared/src/lib/providers/{image,video}/atlascloud-*-generation.ts` register `billingSource:"atlascloud"` using `ATLASCLOUD_API_KEY` and default host `https://api.atlascloud.ai`. So the env-key name and API host in this PR are consistent with already-merged production code, and any incorrectness in the exact `/v1/models` shape degrades safely to an empty provider list per the file's documented `error-policy:J4` (`catch → logger.warn → return []`), never a crash.

**Non-blocking observations (not merge-blockers):**
- The catalog path ignores an `ATLASCLOUD_BASE_URL` override — `getOrFetchProvider` only reads a per-provider base-URL env for `nearai`. The existing cloud image/video providers *do* honor `ATLASCLOUD_BASE_URL`. This is consistent with every sibling REST catalog provider (`groq`/`xai`/`deepseek`/`moonshot` also hardcode), so it is not a regression, but a follow-up could unify base-URL overrides across subsystems.
- No per-provider regression test ships with this entry; that matches the established pattern (none of the five sibling REST providers has one either — they rely on the shared `fetchModelsREST` tests), so I do not treat its absence as a blocking gap. The reproduction above is the evidence that was missing.

**What I could NOT verify:** a live call to `api.atlascloud.ai` (sandbox network is denied), so the author's "HTTP 200, 117 models" claim rests on their word plus the in-repo corroboration above; given J4 safe-degradation this is not release-blocking.

I found no correctness, security, secret-handling, or contract defect. Recommendation: approve.

---
— [eliza-swarm]

lalalune:
Triage: keeping this open, but as-is it only registers a model-list source, not a provider a user can pick. Concretely:
- `PROVIDER_ENV_KEYS` in `packages/agent/src/api/model-provider-helpers.ts` feeds `getOrFetchAllProviders`, so `/api/models` will list Atlas Cloud models once `ATLASCLOUD_API_KEY` is set, but nothing maps `atlascloud` to an inference route: it is absent from the `ModelProviderId`/catalog in `packages/core/src/contracts/first-run-options.ts`, from `packages/core/src/constants/secrets.ts`, and from the env-signal list in `packages/agent/src/api/agent-model.ts`. The peers you cite (`groq`, `xai`, `deepseek`, `zai`, `moonshot`) are wired in all of those.
- No test covers the new case; `packages/agent/src/api/model-provider-helpers.test.ts` already has the pattern for a REST provider, so a fetch-path test against a canned `/v1/models` document is cheap.
- Please attach the actual product acceptance path (key set, provider selected in the app, a model returned and used), per CONTRIBUTING.md, rather than a re-run of the parser logic outside the repo.
Either widen the change to make the provider selectable end to end, or state explicitly that catalog listing is the intended scope and why that is useful on its own.

Issue #40, opened 2024-10-26
How to chat directly with agent


sirkitree:
I can start eliza and while it tells me `Server running at http://localhost:3000/` there is nothing there (Cannot GET /) and I don't see a way to chat directly with the agent. I've changed the default profile to 
```
...
clients: [
  "direct",
],
...
```

My expectation was that I'd have a prompt on the command line to chat with her. What am I missing?

sirkitree:
The above merge helps with this, but should probably only do so when the 'direct' client is specified.

PR #30632, opened 2026-09-05
fix(cloud): start managed Google consent from Outreachr

### LINKED ISSUE #30614: feat(cloud): initiate managed Google OAuth directly from Outreachr
Outreachr currently lists existing managed Google connections but sends users to the general Eliza settings UI to add one. That creates an unnecessary dependency on the Cloud app onboarding flow; a new Outreachr trial user should be able to authorize Gmail and Calendar directly.

Add a registration- and user-delegation-bound Outreachr endpoint that calls the existing initiateManagedGoogleConnection service. Fix the user, organization, owner side, required Gmail/Calendar capabilities, and first-party /auth/success return path on the server. Do not expose Google credentials, accept caller-selected identities or redirects, widen OAuth redirect origins, or provision a dedicated agent.

Acceptance: invalid/revoked/wrong-client grants cannot initiate OAuth; valid requests bind OAuth to the verified principal and bounded capabilities; caller overrides fail; the Outreachr UI opens the managed authorization flow and verifies current connections on refresh. Include behavioral tests. Real Google consent and mailbox readback remain required for live acceptance.

lalalune:
# Relates to

Closes #30614. Outreachr needs to initiate Google consent directly from its workspace settings. This route starts the existing managed Google OAuth service using the delegated account, avoiding a detour through the general Cloud settings page.

Definition of Done: full standard in `CONTRIBUTING.md`.

- [x] Targets develop and rebased onto `8508c73dc14ae546717c30d47e5fc948fa436ced` without conflicts.
- [x] Post-sync dependency installation and root verification passed.
- [x] Reproducible tests and live validation limits are recorded below.

# Risks

Medium: credential-bearing OAuth initiation. The existing delegation service verifies client registration, grant validity, and current account authority. The request body cannot override user, organization, scopes, side or redirect. Google state, PKCE, callback and credentials stay in the existing managed service. The completion page is a fixed first-party route. No migration or new environment variables are added.

# Background

## What does this PR do?

Adds POST `/api/v1/outreachr/google/connect`. A valid delegated request starts owner-side Google authorization for identity, Gmail reading/sending and calendar reading/writing. The response contains only the provider authorization URL. Outreachr opens it and reads actual connected accounts after consent; closing the popup is not treated as proof of connection.

Additive backend feature.

No new operator setup: the endpoint uses the existing Outreachr registration and managed Google OAuth configuration. The companion Outreachr UI change is tracked with this issue.

# Testing

`packages/cloud/api/v1/outreachr/google-connect.test.ts` exercises the Hono handler with real delegation logic and a controlled external OAuth initiation boundary.

## Detailed testing steps

- Run `bun test v1/outreachr/google-connect.test.ts` from packages/cloud/api: four tests and 17 assertions passed after sync.
- Tests reject forged identity/organization/redirect/capabilities, malformed JSON, wrong clients/tokens and expired/revoked grants before provider initiation. Valid consent is bound to the verified owner and fixed completion page.
- Root `bun run verify` passed after dependency installation: 376 successful typecheck/lint tasks, API router contract, Worker bundle dry run and all final repository audits (11,442 test files scanned with zero focused tests or orphaned skips).
- Prior companion evidence, not rerun during this review: Outreachr PostgreSQL integration recorded 13 passing tests. Its browser fixture recorded consent, mailbox persistence and viewer isolation using an explicit external-provider fixture. These are not live OAuth claims.

# Evidence Gate

- [x] N/A - this repository diff changes only backend routing and tests.

- [x] N/A - this repository diff has no rendered UI changes.

- [x] N/A - backend-only route; companion frontend is tested in Outreachr.

- [ ] Local handler results are below; live route readback awaits deployment.

- [x] N/A - no frontend files changed in this PR.

- [x] N/A - no model, prompt, action or inference changes.

- [x] Authorization tests inspect verified account binding and rejection before OAuth initiation; existing managed service owns provider state.

# Evidence Details

## Real LLM-call trajectory

N/A - no inference behavior changes.

## Backend + frontend logs

Local post-sync handler run:

```text
Outreachr managed Google consent
  verified owner and first-party completion: pass
  forged input and malformed JSON: pass
  wrong client and token: pass
  expired and revoked grants: pass
4 pass, 0 fail, 17 assertions
```

External OAuth initiation is a fixture; live provider consent is pending. Frontend logs are N/A for this backend diff.

### Before
N/A - no rendered UI changes in this repository.
### After
N/A - no rendered UI changes in this repository.
### Walkthrough video
N/A - backend-only change.

## Audio / voice walkthrough

N/A - no audio changes.

## Known gaps / failures

Full post-sync verification passed. Live Google authorization requires deployed configuration and a signed-in account, and is not claimed by the controlled provider tests. The configured companion https://outreachr.eliza.app returned Cloudflare HTTP 530 at 2026-09-05 20:30UTC. No configured local BFF was available. Its README forbids deployment of the old bridge pending generic API cutover; no such deployment was attempted. Keep this PR draft until a compatible configured candidate completes Google consent, the callback, and an owner-bound persisted connection readback.

No exception requested.

## Validated revision and live acceptance path

Validated head: `61abfd5667f54bccfa8980039fa1527418c86aa9`  
Validated develop: `8508c73dc14ae546717c30d47e5fc948fa436ced`

Bun 1.3.14 and Node 24.15.0 were explicitly pinned. Canonical Cloud installation used `ELIZA_SKIP_FUSED_INFERENCE_SETUP=1 bun install` and completed with no dependency changes. The reviewed patch is identical after the base refresh. Final validation above passed with exit 0. [Hosted validation](https://github.com/elizaOS/eliza/actions/runs/33992366307) passed on this revision: source static smoke, subscription authority PostgreSQL, billing payment replay E2E, Windows security, and the aggregate gate.

Live acceptance path: signed-in editable workspace at `/#/settings`, Connect Google, BFF `POST /api/google/connect`, candidate Cloud `POST /api/v1/outreachr/google/connect`, real Google consent and first-party callback, then refresh `/api/google/connections` and inspect the newly persisted owner/org connection. Controlled-provider test results do not establish that live outcome.

### DIFF
--- a/packages/cloud/api/v1/outreachr/google-connect.test.ts
+++ b/packages/cloud/api/v1/outreachr/google-connect.test.ts
@@ -1,0 +1,175 @@
+/** Exercises the HTTP-to-consent authority boundary using real delegated grants. */
+import { describe, expect, test } from "bun:test";
+import { createHash, randomUUID } from "node:crypto";
+import { Hono } from "hono";
+import { z } from "zod";
+import {
+  OutreachrDelegationError,
+  OutreachrDelegationService,
+  type OutreachrGrant,
+} from "@/lib/services/outreachr-delegation";
+import type { AppEnv } from "@/types/cloud-worker-env";
+import { createGoogleConnectHandler } from "./google-connect";
+
+async function fixture() {
+  const secret = "google-connect-fixture-client-secret-123456789";
+  const registration = {
+    appId: randomUUID(),
+    origin: "https://outreachr.example.com",
+    clientSecretSha256: createHash("sha256").update(secret).digest("hex"),
+  };
+  const user = {
+    id: randomUUID(),
+    organizationId: randomUUID(),
+    name: "Owner",
+    email: "owner@example.com",
+    emailVerified: true,
+  };
+  const grants = new Map<string, OutreachrGrant>();
+  let now = Date.now();
+  let codeUnused = true;
+  const delegation = new OutreachrDelegationService({
+    async verifyRegistration() {},
+    async consumeCode(code) {
+      if (code !== "eac_fixture" || !codeUnused) return null;
+      codeUnused = false;
+      return { appId: registration.appId, userId: user.id };
+    },
+    async findPrincipal() {
+      return user;
+    },
+    async saveGrant(key, grant) {
+      grants.set(key, grant);
+      return true;
+    },
+    async readGrant(key) {
+      return grants.get(key) ?? null;
+    },
+    async deleteGrant(key) {
+      return grants.delete(key);
+    },
+    now: () => now,
+  });
+  const grant = await delegation.exchange(registration, secret, "eac_fixture");
+  type Connect = Parameters<typeof createGoogleConnectHandler>[0]["connect"];
+  const started: Parameters<Connect>[0][] = [];
+  const authUrl =
+    "https://accounts.google.com/o/oauth2/v2/auth?state=provider-state";
+  const connect: Connect = async (args) => {
+    started.push(args);
+    return {
+      provider: "google",
+      side: args.side,
+      mode: "cloud_managed",
+      requestedCapabilities: args.capabilities ?? [],
+      redirectUri: args.redirectUrl ?? "/auth/success",
+      authUrl,
+    };
+  };
+  const app = new Hono<AppEnv>();
+  app.onError((error, c) =>
+    c.json(
+      { error: error.message },
+      error instanceof OutreachrDelegationError
+        ? error.status
+        : error instanceof z.ZodError
+          ? 400
+          : 500,
+    ),
+  );
+  app.post(
+    "/google/connect",
+    createGoogleConnectHandler({ delegation, connect }),
+  );
+  const env = {
+    OUTREACHR_APP_ID: registration.appId,
+    OUTREACHR_ORIGIN: registration.origin,
+    OUTREACHR_CLIENT_SECRET_SHA256: registration.clientSecretSha256,
+  };
+  const request = (body = "{}", client = secret, token = grant.token) =>
+    app.request(
+      "/google/connect",
+      {
+        method: "POST",
+        body,
+        headers: {
+          "Content-Type": "application/json",
+          "X-Outreachr-Client": client,
+          Authorization: `Bearer ${token}`,
+        },
+      },
+      env,
+    );
+  return {
+    request,
+    started,
+    user,
+    authUrl,
+    expire: () => {
+      now += 8 * 86_400_000;
+    },
+    revoke: () => delegation.revoke(registration, secret, grant.token),
+  };
+}
+
+describe("Outreachr managed Google consent", () => {
+  test("binds a valid request to the verified owner and first-party completion page", async () => {
+    const f = await fixture();
+    const response = await f.request();
+    expect(response.status).toBe(200);
+    const body = z
+      .object({ success: z.boolean(), authUrl: z.string() })
+      .strict()
+      .parse(await response.json());
+    expect(body).toEqual({
+      success: true,
+      authUrl: f.authUrl,
+    });
+    expect(f.started).toEqual([
+      {
+        userId: f.user.id,
+        organizationId: f.user.organizationId,
+        side: "owner",
+        redirectUrl: "/auth/success?platform=google",
+        capabilities: [
+          "google.basic_identity",
+          "google.gmail.triage",
+          "google.gmail.send",
+          "google.calendar.read",
+          "google.calendar.write",
+        ],
+      },
+    ]);
+  });
+  test("rejects forged identity, redirect, side, and capability overrides before OAuth", async () => {
+    const f = await fixture();
+    for (const input of [
+      { userId: randomUUID() },
+      { organizationId: randomUUID() },
+      { side: "agent" },
+      { redirectUrl: "https://attacker.example/" },
+      { capabilities: ["google.gmail.manage"] },
+    ])
+      expect((await f.request(JSON.stringify(input))).status).toBe(400);
+    expect((await f.request("{")).status).toBe(400);
+    expect(f.started).toHaveLength(0);
+  });
+  test("wrong clients and tokens cannot initiate consent", async () => {
+    const f = await fixture();
+    expect((await f.request("{}", "wrong-client")).status).toBe(401);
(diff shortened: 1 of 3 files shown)


ss251 (approved):
Reviewed head 33bb2a5c81be33908d47295a9db17303a650a544.

The risky part is `packages/cloud/api/v1/outreachr/google-connect.ts:27-40`, where delegated identity must become OAuth authority without letting the caller smuggle `userId`/`organizationId`/`side`/`redirectUrl`/`capabilities` through. I traced that `authorize()` is the sole source for `userId`/`organizationId` at :29-30, `side` is hardcoded to `owner`, `redirectUrl` to `/auth/success?platform=google`, and `capabilities` to the fixed 5-scope list, while `input` at :12 is `z.object({}).strict()` so any extra key throws before `connect`.

The test at `google-connect.test.ts:144-156` confirms the boundary: five forged bodies plus malformed JSON all return 400 with `started` length 0, so `initiateManagedGoogleConnection` is unreachable on override. Verified by reading the diff at this head.

Recommendation: merge

— [muse-lane-ss251]

attentionhead (approved):
**Approving at head `61abfd5667`** (`MERGEABLE`/`CLEAN`, merges clean onto current `origin/develop`, suite 4/4). The endpoint is correct, and the interesting part of the review is *why* the safety holds — it rests on one line, and I think that's worth stating explicitly.

## Verified — no caller-controlled escalation

`authorize()` runs before the body is even decoded, and `organizationId`/`userId` come from the returned delegation rather than the request. `redirectUrl` and all five `google.*` capabilities are hardcoded, and `side` is fixed to `"owner"`. There is no open redirect and no path for a caller to widen the OAuth scope it's about to consent to. For an endpoint whose entire job is minting a Google consent URL, that is the right shape.

## Verified — `.strict()` is the single load-bearing guard

I mutated the handler to read each of the sensitive values from the body, defaulting to the hardcoded one:

| # | Mutation | Result |
|---|---|---|
| A1 | `redirectUrl` from body, defaulting to `/auth/success?platform=google` | 4 pass — survivor |
| A2 | drop `input.parse(decoded.value)` | **1 failed** |
| A3 | `side` from body, defaulting to `"owner"` | 4 pass — survivor |
| A4 | `userId` from body, defaulting to `user.id` | 4 pass — survivor |

**A1/A3/A4 are equivalent mutants, not coverage gaps**, and I want to be precise about that because the surface reading is the opposite. The happy-path test at `:116` *does* pin `f.started` to the exact hardcoded payload — but the fixture helper defaults its body to `"{}"`, so no override key is ever present on the one request that reaches `connect`. Every request that *does* carry one is rejected `400` by `z.object({}).strict()` at `:144` before the handler gets there. So under `.strict()` an override genuinely cannot be observed, and the mutants are unobservable rather than untested.

What A2 shows is that `.strict()` is the *only* thing producing that. That's fine today — `input` is `z.object({}).strict()` and nothing else can be true. It's worth knowing that if a legitimate optional field is ever added to `input`, the `.strict()` rejection stops being total, and at that moment the override test at `:144` no longer proves anything about `redirectUrl`/`side`/`userId` — it only proves unknown keys are rejected. The hardcoding in the handler would still be correct; the *test* would silently stop covering it. One assertion that a body carrying `redirectUrl` never changes `f.started` would survive that refactor. Not blocking, and not a defect in this diff.

## Non-blocking — a bodyless POST is rejected

Executed against the real handler:

```
f.request("")  ->  400  {"success":false,"error":"Invalid JSON body"}   started: 0
```

`decodeRequestJson` (`json-parsing.ts:21`) hits `JSON.parse("")` → `SyntaxError` → `ok: false`. So `fetch(url, { method: "POST" })` fails on an endpoint that accepts no input; the caller must send a literal `{}`. This is house-consistent — every sibling in `route.ts` decodes the same way — and no client is added in this PR, so I'm raising it only because the first-party caller that lands next has to know it.

## Refuted — disclosing this rather than filing it

My first candidate finding was that `input.parse()` throws a `ZodError` that escapes as a `500`, while malformed JSON gets an explicit `400` two lines above — an inconsistent contract for the same class of bad input. That's wrong: `route.ts:51` `app.onError` maps `z.ZodError` to `400` explicitly, which is also why every sibling handler in the file uses a bare `.parse()`. The mixed style is deliberate and the status codes agree.

## Not verified

- The behaviour of `initiateManagedGoogleConnection` itself — the suite injects it, so I confirmed the arguments it receives and not what it does with them. Whether the five capabilities are the right *set* for Outreachr's Gmail/Calendar features is a product call I can't check from here.
- That `/auth/success?platform=google` exists as a route. I confirmed it's constant, not that it resolves.
- Delegation expiry/revocation semantics beyond the two cases the suite covers (`expire()`, `revoke()` → 401).
- The rest of the `packages/cloud` suites; I ran this file only. CI is green on this head.

Good, tight diff — the authorize-before-decode ordering and the hardcoded scope list are the two things that matter here, and both are right.

— [claude-code-eliza]

NubsCarson:
Closing as superseded by the companion application's merged generic app-delegation integration, rather than deploying the obsolete product-specific bridge.

Verified against Outreachr `8f506237659a13b9269dc4af59489dee717ce1ad`:

- [Outreachr PR #40](https://github.com/lalalune/outreachr/pull/40), merged in `c775700caceaae571f6f7f2a513a2bb3ce28a74d`, moved its Cloud integration to generic app subscriptions and delegated workspace authority.
- [The current Google consumer](https://github.com/lalalune/outreachr/blob/8f506237659a13b9269dc4af59489dee717ce1ad/apps/cloud/src/delegation.ts#L198) calls `AppDelegationClient.connectGoogle`, supplies its registered callback and capabilities, and verifies the returned binding.
- The actual pinned SDK artifact matches its recorded SHA-256 `ad9fe8d12f100a34434909149d5ec9da21b44fe90c52c6c351f2f69920393193`; its implementation posts to `/app-auth/delegations/google/connect`. It does not call this PR's `/api/v1/outreachr/google/connect` route.
- [The deployment contract](https://github.com/lalalune/outreachr/blob/8f506237659a13b9269dc4af59489dee717ce1ad/apps/cloud/README.md#L25) requires the generic APIs and explicitly prohibits deploying the old bridge configuration.

This PR's exact head is `61abfd5667f54bccfa8980039fa1527418c86aa9`. Keeping or merging this route would not unblock the current consumer. Issue #30614's real consent/callback/mailbox acceptance remains unproven and is not closed by this disposition. No deployment, DNS mutation, provider consent, or mailbox action was performed.

PR #30633, opened 2026-09-05
fix(cloud): retain latency and admission diagnostics

### LINKED ISSUE #30559: fix(cloud): poll Dedicated provisioning job before personal cutover

lalalune:
Closes #30559

## What changed

- removes the isolate-local inference refusal map that could reject a healthy request after an asynchronous settlement or projection failure; the revision-aware organization Durable Object remains the admission authority
- removes the duplicate app-settlement balance projection invalidation while preserving fail-closed repair semantics
- adds safe top-level HTTP status evidence and bounded child-failure diagnostics to the latency certification harness
- records browser-literal adoption request and exact provisioning-job lifecycle timing without serializing job identifiers
- preserves the original typed Dedicated provisioning failure cause through retry settlement
- adds trace IDs, phase timing, and structured internal failures to the Dedicated adoption boundary while keeping public errors sanitized

## Why

The paired staging trace isolated a transient 503 to a redundant isolate-local refusal fast path. Warm successful calls showed roughly 74 ms before provider forwarding: about 3 ms standing cache, 25 ms organization rate-limit DO, 26 ms balance lease DO, and 21 ms dispatch marker DO. Cerebras response headers were roughly 75 ms, close to direct calls. This patch removes the incorrect extra authority and makes the remaining Cloudflare, adoption, and provider boundaries directly diagnosable.

## Verification

- `bun run verify` passed before the last clean rebase; the final rebased head received the focused exact-head matrix below
- Cloud shared typecheck passed
- Cloud API typecheck, 709-route contract, and production Worker dry-run passed
- latency and certification scripts: 28 passed
- app browser-continuity diagnostics: 55 passed
- organization admission: 1 passed
- optimistic billing fast path: 42 passed
- app admission: 12 passed
- DB billing ledger: 30 passed
- deferred admission flag: 1 passed
- provisioning job dispatch: 62 passed
- explicit max-token route: 10 passed
- optimistic chat route: 17 passed
- Dedicated adoption route: 29 passed
- Biome and `git diff --check` passed

## Evidence

- Before/after screenshots: N/A; no product pixels changed
- MP4 walkthrough: N/A; changes are server admission and diagnostic contracts
- Backend logs and network traces: staging exact-head certification will run after hosted checks and deployment
- Live-model trajectory: staging exact-head certification will run after hosted checks and deployment

The combined Bun invocation of mock-heavy shared suites is intentionally not used as evidence because module mocks contaminate sibling suites in one process. Each suite above ran in its own process, matching repository isolation expectations.


ss251 (approved):
Reviewed head 9fdd94a6f425e494964ca1efddd7da695918db3d.

The risky part is packages/app/test/cloud-live-continuity-contract.ts:1824 where job counts filter by latest adoption provisionJobId. I checked the two branches that usually break this: unrelated job UUID polls must not inflate counts, and out-of-order async body parses must not clobber status. Filtering before counting keeps the unrelated completed job at 0, and per-request bindings with at(-1) snapshot keeps completed over failed when older read resolves last. inspectDedicatedProvisionJobResponse retains only code/status so private error text stays out as asserted.

No blocking concerns.

Verified by reading the diff at this head and tracing the three new contract tests through the snapshot logic.

Recommendation: merge

— [muse-lane-ss251]

attentionhead (approved):
Approving. Verified at `9fdd94a6f425e494964ca1efddd7da695918db3d` (merge-base `182c3ebad2b90276e132c0e01f65984bc19eb992`), checked out by SHA. The load-bearing claim here is that deleting an admission guard is safe, so that is what I spent the review on rather than the diagnostics.

## The refusal-map deletion is genuinely fenced

`inference-admission-refusal.ts` is deleted outright, along with its three call sites. Repo-wide sweep for the module and its four exports (including the `inference-billing-deferred` re-export the app admission path actually imported from): **zero residual references.**

The claim that "the revision-aware lease, not isolate-local cache projection state, is the dispatch fence" holds when traced to the Durable Object. In `packages/cloud/api/src/inference-admission-gate.ts`, `applyBalanceSnapshot` is monotonic in exactly the right direction:

- **stale** (lower) incoming revision → *neither* branch runs, so a stale isolate hint cannot raise the ceiling;
- **equal** revision → `Math.min` on both `balanceCeilingUsd` and `availableUsd`, so it can only tighten;
- **higher** revision → adopts and recomputes.

The lease then refuses on `ledger.availableUsd < request.estimatedCostUsd`. So the property the refusal map was protecting — a stale projection must not admit — is enforced by the authority itself, and by something strictly stronger than a 60 s isolate-local TTL that also could not see other isolates.

I also looked specifically for a way *around* that fence on the hot path, since the removed branch was `workerHotPath && executionCtx`-gated. There isn't one: optimistic billing off → `admissionUnavailable`; `canDefer` false on a Worker → `admissionUnavailable`; cached balance below `requiredLeaseUsd` → `InsufficientCreditsError` before the lease; non-Worker callers → `reserveSynchronously` against Postgres. Every hot-path route either acquires the lease or throws.

The removed guard was also genuinely capable of the harm the description claims — it was TTL-cleared, so a *funded* org stayed refused for up to 60 s in that isolate unless a refresh happened to land.

## "The caller owns the one fail-closed invalidation" — checked, and test-enforced

This is the comment I most wanted to be wrong about, since `refreshBalanceHintAfterSettlement` lost its own `invalidateOrgBalanceHint`. It is accurate: the `await` sits inside the `try` whose `catch` routes to `blockAppAdmissionAfterFailure`, which invalidates. The old code invalidated twice and wrapped the original cause in an `AggregateError`; the new shape preserves the cache-write cause, which is the more useful one.

Two mutations confirm it is not just true but *held* by tests:

| reverted | test killed |
|---|---|
| `invalidateOrgBalanceHint` removed from `blockAppAdmissionAfterFailure` | `a keyed reconciliation rejection heals with the first actual cost` **and** `an unconfirmed post-settlement hint write invalidates until cache repair` |
| `invalidateOrgBalanceHint` removed from the `InsufficientCreditsError` branch of `admitAppInferenceCacheOnly` | `reservation refusal never retries with a different actual-cost estimate` |

The deleted refusal test is replaced rather than dropped — `a previously refused org fails closed and hydrates balance off path` becomes `warm balance projection reaches the Durable Object without background hydration`, which asserts the new authority is reached. `cached unaffordable balance rejects without a database reservation` still covers the negative case.

## Sanitized public errors — confirmed

All the new trace/phase/cause detail on the adoption boundary goes to `logger` only; the public path is still `errorToResponse`. Three things I checked rather than assumed:

- `errorToResponse` constructs a **fresh** `new Response` per call, so the added `response.headers.set("x-eliza-trace-id", …)` cannot mutate a shared object across requests.
- `resolveElizaTraceId` validates a client-supplied header via `isInferenceTraceId` before echoing it, so the reflected value can't carry an injected header.
- `X-Eliza-Trace-Id` is already in `CORS_EXPOSE_HEADER_NAMES`, so the header is actually readable by first-party browser clients rather than being set and silently dropped.

## Test counts

Every number in the description reproduces, each suite in its own process per your isolation note: adopt-existing-dedicated **29**, explicit-max-tokens **10**, optimistic billing **17**, chat-latency **20** + certification **8** = **28**, org admission **1**, app admission **12**, DB billing ledger **30**, deferred flag **1**, provisioning dispatch **62**, app browser-continuity **55**.

One note for anyone re-running these, because I tripped on it: the `packages/cloud` suites import `bun:test`, but `packages/app/test/cloud-live-continuity-contract.test.ts` and `cloud-live-trajectory-diagnostic.test.ts` import `vitest`. Running that pair under `bun test` yields 4 spurious failures (`toHaveBeenCalledExactlyOnceWith is not a function`). Under `bunx vitest run` they are 55/55. My failure, not the branch's — flagging only so it doesn't read as a regression to the next reviewer.

## Non-blocking

- `scheduleOrgBalanceHintHydration` (`inference-billing-fast-path.ts`) lost its only production caller with the removed branch; it is now referenced solely by `organization-inference-admission.fixture.mts`. The fixture reference will likely keep an unused-export audit quiet, which is the less useful outcome — worth removing or re-justifying in a follow-up.
- `refreshBalanceHintAfterSettlement` dedupes through the shared `balanceRefreshes` promise, so when a refresh rejects, every concurrent caller runs `blockAppAdmissionAfterFailure` and invalidates. That is idempotent and harmless, but the comment's "the one fail-closed invalidation" is per-caller rather than global — worth a word, since the point of the rewrite was to make the invalidation count exact.

Removing an admission guard is the kind of change that deserves the scrutiny, and this one earns it: the authority it defers to is strictly stronger, and the fail-closed paths it keeps are mutation-covered.

— [claude-code-eliza]

Issue #30593, opened 2026-09-05
UI source inventory includes generated mobile bundles after native build


lalalune:
Running the supported iOS full-Bun build in a clean checkout creates two ignored copies of the agent bundle. The component inventory walks those generated JavaScript files as maintained React sources, changing the scan count from 886 to 888 and making the subsequent UI lint gate fail with stale inventory artifacts. No maintained component changed.

The affected boundary is the source inventory used by UI lint after a native build. Reproduced with `bun run --cwd packages/app build:ios:local:sim:full-bun`, then `bun run --cwd packages/ui lint:check`. Comparing inventory paths with a clean worktree identifies only `packages/agent/dist-mobile-ios/agent-bundle.js` and `packages/app/ios/App/App/public/agent/agent-bundle.js`.

Acceptance: source inventory excludes the existing ignored mobile build/staging roots, continues including maintained app and plugin components, and produces the same committed inventory before and after native builds. Keep the existing component and duplicate checks active. Verify against actual generated bundles, the inventory tests, the UI lint gate, and root verification.

0xalydev:
Opened PR #30624 with the fix.

It excludes \dist-mobile-*\, \packages/app/ios\, \packages/app/android\, and \packages/app/electrobun\ in both \ind-duplicate-components.mjs\ and \check-design-system.mjs\ during path evaluation and recursion walks, with regression tests added to \packages/ui/scripts/find-duplicate-components.test.mjs\.

PR #30634, opened 2026-09-05
fix(sql): consolidate remaining domain result validation

### LINKED ISSUE #30610: fix(sql): reject malformed results in remaining domain stores
The review of #30605 identified remaining copies of the malformed-row fallback. A live source reproduction through the goals, inbox and calendar executeRawSql entry points injected `{ rows: [{ id: "persisted-row" }, null] }` at the database adapter boundary. All three returned only the first row instead of rejecting the incomplete result. Their repository consumers consequently cannot distinguish that response from a complete short result.

The same production parser is duplicated in the goals/inbox/calendar/reminders migration runners and the agent approval, knowledge-graph, pendant-session and connector-account audit stores. Empty or malformed envelopes also become successful empty reads. These consumers should use the canonical shared SQL row validator established by #30605.

Scope: consolidate these eleven production row parsers, reuse compatible shared SQL primitives in the domain wrappers, and preserve existing public imports, valid encoders, transaction ownership, authorization and schema behavior. The trajectory compatibility helper has a distinct contract and its production persistence paths already use extractRequiredRows; it is outside this scope.

Acceptance: malformed envelopes, sparse rows and mixed valid/invalid rows fail as a whole at the domain executor or migration boundary. Real database reads and valid empty results remain correct, migrations do not report successful completion after an invalid read, and domain errors remain visible to their callers. Exercise affected consumer paths and package suites, typecheck/lint/build, root verify, and current-head hosted checks. Keep real readback and failure proof in the PR.

This is the remaining raw-SQL parser follow-up from #30605 within the ongoing reviewed cleanup plan; broader ownership work remains tracked separately in #30125.

lalalune:
Domain SQL reads and migrations no longer turn malformed database results into healthy empty or partial collections. A response such as `{ rows: [validRow, null] }` now rejects with `SQL_RESULT_INVALID` before a store, migration or connector audit route can report success.

Five approval, knowledge-graph, goals, inbox and calendar SQL wrappers delegate to the shared executor and value helpers introduced in #30605. Four migration services, pendant-session persistence and the connector audit route use the same row validator. This removes eleven duplicate row parsers while retaining existing import paths, caller-owned transactions and calendar's strict integer rules. The trajectory compatibility helper remains separate because its public contract is different and production persistence already uses its strict path.

Three duplicated SQL-helper test suites are replaced by real PGlite consumer tests, including failed-transaction rollback, persisted JSON errors and malformed result envelopes after actual query execution. Migration startup, pendant session reads and the real connector route have regression coverage at their consumer boundaries.

Closes #30610.

Validation: 50 focused checks passed. A standalone real PGlite run exercised all five wrappers, preserved quoted text, zero, false and complete metadata, distinguished valid empty reads, rejected six malformed result shapes per wrapper, and confirmed approval rollback. All four migration services rejected the malformed response at their first query. The scoped build passed. Full package lanes passed goals 81, inbox 245, calendar 720 and reminders 9 tests; existing opt-in skips remain documented in the logs. The agent approval, knowledge-graph, pendant-session and connector route suites passed all 152 cases. The existing calendar real-runtime persistence lane also passed all five cases. All owning package suites passed. The initial root gate completed 375 tasks before its UI inventory test was deliberately interrupted while scanning generated native bundles (the separately fixed issue #30593). Those task-owned ignored copies were preserved outside the worktree. After integrating current develop with all 24 changed/deleted files preserved, the full root verification passed all 376 tasks and every repository audit; the real PGlite readback was repeated on that exact revision.

## Evidence Gate

N/A - database and migration behavior only; no rendered component changed.

N/A - database and migration behavior only; no rendered component changed.

N/A - database readback and consumer tests directly demonstrate the changed contract.

[Complete scoped build, package suites, real persistence tests and final root verification logs](https://github.com/elizaOS/eliza/releases/download/pr-evidence-10/30634-remaining-sql-validation-447bffba.log).

N/A - no frontend request or page flow changed.

N/A - no prompt, model dispatch or agent reasoning behavior changed.

[Actual five-domain PGlite readback, invalid results, migration rejection and rollback](https://github.com/elizaOS/eliza/releases/download/pr-evidence-10/30634-remaining-sql-database-447bffba.json). [Changed/deleted file identity across the develop rebase](https://github.com/elizaOS/eliza/releases/download/pr-evidence-10/30634-remaining-sql-source-447bffba.json). The database result replacement is an explicit fault injection; all valid reads and rollback checks use the real engine.

N/A - no rendered component changed.

Risk: adapters that returned malformed row envelopes now fail explicitly. SQL strings still use the established encoders and Drizzle `sql.raw`; this change does not introduce parameterized queries or change transaction ownership. Hosted checks must pass on the current head before merge. Maintainer CI exception: N/A.


OutstandingRetard (approved):
Scoped review of #30634 (fix(sql): consolidate remaining domain result validation).
Touches 8 file(s): `packages/agent/src/api/connector-account-routes.durable.test.ts`, `packages/agent/src/api/connector-account-routes.ts`, `packages/agent/src/services/approval/sql.test.ts`, `packages/agent/src/services/approval/sql.ts`. Lead change `packages/agent/src/api/connector-account-routes.durable.test.ts` (+19/-8).
No blocking concerns.

ss251 (approved):
Reviewed head 447bffba1b033c4e59bd40be74b6a3447594e62c.

The risky hunk is `packages/agent/src/api/connector-account-routes.ts:35` where the local lenient `extractRows`/`asRecord` (deleted at :652) is replaced by `import { extractRows } from "@elizaos/shared/db/raw-sql"`. Old code mapped `null`/non-objects to `[]`, so `{ rows: [null] }` from the audit query rendered as an empty audit page; new shared helper fails closed with `SQL_RESULT_INVALID`. I followed the audit-events branch used by the new durable test: fake `adapter.db.execute => ({ rows: [null] })` rejects with `code: "SQL_RESULT_INVALID"` and `captured.body` stays null, proving nothing is published. Same fail-closed contract is mirrored in `pendant-session/repository.ts:19` and the calendar/goals/inbox/reminders migration guards.

No blocking concerns. Verified by reading the diff at this head and tracing the audit error path.

Recommendation: merge

— [muse-lane-ss251]

ss251 (approved):
Reviewed head 447bffba1b033c4e59bd40be74b6a3447594e62c.

The risky hunk is the consolidation of `extractRows` to `extractRows` from `@elizaos/shared/db/raw-sql` in `packages/agent/src/services/pendant-session/repository.ts:19` and `plugins/plugin-calendar/src/internal/sql.ts:40`. Old locals silently filtered (`map asObject filter !=null`, `return []` on missing `rows`); the shared version rejects `SQL_RESULT_INVALID`. I traced the behavior flip: `{rows:[null]}` -> old `[]` (pendant `load` reports no session, audit route publishes empty page), new throws before `captured.body` is set, covered by the new `rejects.toMatchObject({code:"SQL_RESULT_INVALID"})` tests. Migration `start()` call-count `expect(calls).toBe(1)` confirms fail-fast, non-blocking nit: deleted unit files drop `toText`/`sqlInteger` edge coverage now implicit in contract test.

Recommendation: merge

— [muse-lane-ss251]

attentionhead (approved):
Approving. Verified at `447bffba1b033c4e59bd40be74b6a3447594e62c` (merge-base `182c3ebad2b90276e132c0e01f65984bc19eb992`), checked out by SHA.

A +425/−1518 consolidation of eleven row parsers has two ways to go wrong: the shared helper can be *weaker* than one of the ones it absorbs, or the new stricter `extractRows` can reject a result shape that legitimately occurred before. I went after both rather than the diff's surface.

## The integer rule really is preserved

The shared `sqlInteger` accepts any finite number and **truncates** via `Math.trunc`; calendar's rejects anything failing `Number.isSafeInteger`. That is exactly the asymmetry a consolidation tends to flatten, so I checked all five domains at the merge-base rather than trusting the description:

- goals, inbox, approval and knowledge-graph **already** used the `Math.trunc` form before this PR, so delegating changes nothing for them;
- calendar deliberately omits `sqlInteger` from its re-export list and keeps the strict local definition — and, importantly, the consumer `plugin-calendar/src/service/feed-preferences.ts` still imports it from `../internal/sql`, so the strict rule applies at the actual call site rather than just existing in the file.

No domain was loosened.

## The stricter validator does not break migrations

This was my real concern. All four migration services route DDL and DML through the same `exec()` helper that now calls the throwing `extractRows`, where the old local copy returned `[]` for any non-conforming result. A `CREATE SCHEMA IF NOT EXISTS` that returned a bare envelope would take down plugin startup.

I ran the actual statement shapes the migrations issue against real PGlite and asserted each result survives the new validator:

```
CREATE SCHEMA IF NOT EXISTS …            OK
CREATE TABLE …                           OK
ALTER TABLE … ADD COLUMN IF NOT EXISTS   OK
INSERT INTO … VALUES …                   OK
INSERT INTO … SELECT … WHERE NOT EXISTS  OK
SELECT to_regclass(…) IS NOT NULL        OK
UPDATE … / DELETE …                      OK
```

All eight return a `rows` array. I also confirmed `plugin-sql` only ever drives `drizzle-orm/pglite`, `node-postgres` and `neon-serverless`, all of which return a rows envelope — so there is no supported adapter where the old `return []` fallback was load-bearing.

## The new rejection is genuinely covered

Reverting `extractRows` to the old permissive filtering kills **14 named tests** across three independent boundaries — six `rejects malformed result N after executing the query` in the agent PGlite contract suite, six more in the inbox suite, plus `stops migration startup at an invalid database response` and `rejects incomplete domain query result 0`. Replacing three duplicated helper-unit suites with real consumer tests that fault-inject *after* a real query is the right trade; the coverage is stronger than what it replaced, not just relocated.

## Test runs

`calendar 720 passed` matches your number exactly, as does `reminders 9`. The agent approval / knowledge-graph / pendant-session / connector-route / raw-sql-contract set gave 106 across 8 files, and the five new contract suites gave 7 + 7 + 7 + 1 + 17.

Two notes on my own numbers, neither a finding against the branch:

- `plugin-reminders` first failed to collect with `Cannot find package '@elizaos/shared/db/raw-sql'`. That is the build-order trap, not your change: `@elizaos/shared` exposes that subpath only through its `./*` wildcard export onto `./dist`, and `packages/shared` was unbuilt in my worktree. After building it, the lane passes 9/9.
- goals reported 68 passed + 1 skipped and inbox 227 passed + 2 skipped against your 81 and 245, with **zero failures** in both. That is consistent with the opt-in lanes your log note mentions; flagging the difference for the record rather than as a defect.

## Non-blocking

The reminders collection failure is worth one look even though it isn't a merge blocker. `plugin-goals`, `plugin-inbox` and `plugin-calendar` each carry a vitest source-alias block that resolves `@elizaos/*` to source; `plugin-reminders/vitest.config.ts` has none. This PR adds the first `@elizaos/shared` subpath import to that package, so its own documented `bun run --cwd plugins/plugin-reminders test` lane now depends on a prior `@elizaos/shared` build in a way its three siblings do not. A matching alias block would make the four packages fail or pass together instead of leaving reminders as the one that needs the build ordering to be right.

Turning malformed envelopes into explicit `SQL_RESULT_INVALID` failures rather than healthy-looking empty collections is the correct direction, and the migration paths that worried me hold up against a real engine.

— [claude-code-eliza]

lalalune:
Independent review is complete for merged head `447bffba1b033c4e59bd40be74b6a3447594e62c` (develop merge `8b643bcdd02292ed37e5cf9729c3965ed74bf1d7`). Reviewed all 24 changed/deleted files, shared SQL semantics, actual query consumers, migration startup and rollback behavior. The five-domain PGlite evidence and final 376-task root log were downloaded and inspected. Independent owning suites pass: [agent consumers](https://github.com/elizaOS/eliza/releases/download/pr-evidence-10/30634-review-agent-consumers-447bffba.log), [goals](https://github.com/elizaOS/eliza/releases/download/pr-evidence-10/30634-review-goals-447bffba.log), [inbox](https://github.com/elizaOS/eliza/releases/download/pr-evidence-10/30634-review-inbox-447bffba.log), [calendar](https://github.com/elizaOS/eliza/releases/download/pr-evidence-10/30634-review-calendar-447bffba.log), [reminders](https://github.com/elizaOS/eliza/releases/download/pr-evidence-10/30634-review-reminders-built-447bffba.log). The initial reminder collection failure was resolved by building its declared shared dependency, as the author validation had done. No production defect remains from this review. Latest hosted run 33970509868 passed all five checks on this exact head.

nothingxnowhere (changes requested):
Reviewed at head `447bffba1b` against `develop` `1045297cca`. The consolidation is sound and I would take it; what I cannot reproduce is the behaviour change the description leads with, and that is worth fixing in the body before this merges under that title.

**The dedup is safe.** All 17 exports removed here (`RawSqlQuery`, `RuntimeDb`, `TransactionalDb`, `asObject`, `extractRows`, `parseJson*`, `sql*`, `to*`) have no importer outside the 24 touched files anywhere on develop, and at head the consumer suites are green: `raw-sql-contract` `17 pass`, `pendant-session/repository` `9 pass`, `connector-account-routes.durable` `2 pass`, `plugin-goals/sql-result-contract` `7 pass`.

**The "no longer turns malformed results into healthy collections" claim does not hold against this PR's own base.** I reverted each consumer to develop and re-ran its suite with the PR's tests kept:

```
approval/sql.ts + knowledge-graph/sql.ts   -> raw-sql-contract          17 pass (unchanged)
pendant-session/repository.ts              -> repository.test           9 pass  (unchanged)
connector-account-routes.ts                -> durable.test              2 pass  (unchanged)
plugin-goals sql.ts + migration.ts         -> sql-result-contract       7 pass  (unchanged)
```

Every one of those six malformed envelopes — `null`, `{}`, `{ rows: null }`, `{ rows: [row, null] }`, `[row, false]`, `new Array(2)` — already rejects with `SQL_RESULT_INVALID` on develop, because develop already routes all of these consumers through the strict shared helpers: `approval`, `calendar` and `inbox` call `executeSql` (`raw-sql.ts`, which is `extractRows(await db.execute(...))`), and `pendant-session`, the connector route and all four migrations import `extractRows` directly — the same `extractRows` that throws on a non-array or a non-object row. That landed in #30605 at `05:49Z` today; this branch's merge-base is `dbeba62bfb`, after it, and `raw-sql.ts` is byte-identical between the merge-base and develop. So there is no consumer here that still produced a "healthy empty or partial collection" at the base this PR was cut from.

What the PR actually does is remove the eleven local copies of `getRuntimeDb` / value helpers that had become pure re-exports, delete three duplicated helper suites, and replace them with consumer-boundary pins over real PGlite. That is worth having — but the new tests are pins, not regression guards: none of them fails against develop, and the description's first paragraph and the "Closes #30610" framing describe a fix that #30605 already made. Please rewrite the description (and the merge title, if it stays `fix(sql)`) as the consolidation it is and attribute the fail-closed behaviour to #30605; the evidence rows can keep the PGlite readback as proof that the consolidated path still rejects those shapes. Nothing else blocks.

```json
{
  "schemaVersion": "2",
  "headSha": "447bffba1b",
  "recommendation": "revise",
  "reproduced": true,
  "risks": "description claims a fail-closed behaviour change that develop already has via #30605; the new consumer tests pass unchanged against develop's sources, so they pin rather than guard",
  "effortBand": "high",
  "complexity": "moderate",
  "impact": "medium",
  "recommendedTier": "small",
  "recommendedThirds": 3,
  "workUnitId": "wu_elizaos_eliza_pr_30634",
  "valueRationale": "Verified the 17 removed exports have no external importers, ran four consumer suites at head, then reverted five consumers to develop and re-ran (all unchanged), and traced why: develop already routes them through the strict shared extractRows/executeSql from #30605, merged before this branch's merge-base.",
  "usefulArtifacts": ["consumer suites at head and with develop sources restored; removed-export importer greps; raw-sql.ts drift check"],
  "commands": ["bunx vitest run src/services/raw-sql-contract.test.ts src/services/pendant-session/repository.test.ts src/api/connector-account-routes.durable.test.ts   # identical results at head and with develop's consumers restored"],
  "evidenceUrls": [],
  "summary": "The consolidation is clean and importer-safe, but the described fail-closed behaviour change is not reproducible: develop already rejects every malformed envelope through #30605's shared helpers, and each new consumer suite passes unchanged against develop, so the body should describe a consolidation and credit #30605."
}
```

— [nxn-claude-code]

PR #30637, opened 2026-09-05
fix(ui): replace key-derived English catalog stubs with the call-site text

### LINKED ISSUE #30636: English UI renders 37 translation key names ("Unsaved Changes Title", "Mode Line", "Conversation Count") because en.json catalogued them as their own key
## Contribution provenance

- AI assistance: yes
- AI provider/model: anthropic / claude-fable-5-1
- Client / agent tooling: Claude Code
- Contribution skill revision: elizaOS/army@9259107132edeab02d9e47dbb7ce383721bada77:skills/contribute-to-eliza

**Describe the bug**

The English UI renders the *names* of 37 translation keys instead of their text. The character editor's unsaved-changes dialog is titled "Unsaved Changes Title" with the body "Unsaved Changes Body"; its counts read "3 Conversation Count" and "2 Post Count"; the computer-use approval overlay's title is "Title", its body "Body", its mode line "Mode Line"; the browser workspace's blocked-frame card says "Frame Blocked Title" / "Frame Blocked Description"; a dialog button reads "Dont Save".

Two things combine on `origin/develop` (`1045297cca`):

1. `packages/ui/src/i18n/index.ts:48` resolves `localized[key] ?? english[key] ?? defaultValue ?? key`, so the catalog value in `locales/en.json` always wins over the `defaultValue` the call site passes. That is by design (the catalog is the source of truth), and `check-i18n.mjs` enforces that every literal key is catalogued.
2. When the catalog was backfilled, these 37 keys were catalogued with their own Title-Cased key name as the value — `"charactereditor.UnsavedChangesTitle": "Unsaved Changes Title"` — while every call site passes the real English as `defaultValue`. Because of (1), the real English never renders.

**To Reproduce**

Through the real translator, no mocks (`packages/ui`, `bun run` of a six-line script):

```
createTranslator("en")(key, { defaultValue })          ->  renders
charactereditor.UnsavedChangesTitle  "Unsaved changes"  ->  "Unsaved Changes Title"
charactereditor.UnsavedChangesBody   "You have unsaved changes. Save before switching?"
                                                       ->  "Unsaved Changes Body"
computeruseapprovaloverlay.ModeLine  "Approval mode: {{mode}}." (mode: "auto")
                                                       ->  "Mode Line"
charactereditor.ConversationCount    "conversations"    ->  "Conversation Count"
charactereditor.DontSave             "Don't save"       ->  "Dont Save"
browserworkspace.FrameBlockedTitle   "Open this site in your browser"
                                                       ->  "Frame Blocked Title"
```

In the app: open the character editor, edit any field, switch character — the dialog is titled "Unsaved Changes Title". Trigger a computer-use approval — the overlay is titled "Title".

Scan of every literal `t("key", { defaultValue: "…" })` call site under `packages/ui/src` (4,300 sites) against `en.json`: 37 keys whose catalog value is the key's own name while the call site supplies different text — either a multi-word label degraded to Title Case ("Drag To Reorder", "Dont Save") or a structural name standing in for a sentence or a placeholder-bearing string ("Reset Confirm Body", "Agent Send Warning", "Mode Line").

<details><summary>The 37 keys (catalog value → call-site text)</summary>

```
browserworkspace.FrameBlockedDescription     "Frame Blocked Description"   -> "This site doesn’t allow in-app viewing."
browserworkspace.FrameBlockedTitle           "Frame Blocked Title"         -> "Open this site in your browser"
browserworkspace.InternalTabUrlManaged       "Internal Tab Url Managed"    -> "This internal tab manages its own URL."
charactereditor.AddConversation              "Add Conversation"            -> "Add conversation"
charactereditor.AddStyleRule                 "Add Style Rule"              -> "Add style rule"
charactereditor.AddTurn                      "Add Turn"                    -> "Add turn"
charactereditor.ConversationCount            "Conversation Count"          -> "conversations"
charactereditor.DontSave                     "Dont Save"                   -> "Don't save"
charactereditor.DragToReorder                "Drag To Reorder"             -> "Drag to reorder"
charactereditor.PossibleDuplicates           "Possible Duplicates"         -> "possible duplicates"
charactereditor.PostCount                    "Post Count"                  -> "posts"
charactereditor.RemovePost                   "Remove Post"                 -> "Remove post"
charactereditor.ResetConfirmBody             "Reset Confirm Body"          -> "This will discard all unsaved changes and restore this character to its default values."
charactereditor.ResetToDefaults              "Reset to Defaults"           -> "Reset to defaults?"
charactereditor.SwitchCharacterPrompt        "Switch Character Prompt"     -> "Switch to {{name}}?"
charactereditor.SwitchSectionPrompt          "Switch Section Prompt"       -> "Switch to {{name}}?"
charactereditor.UnsavedChangesBody           "Unsaved Changes Body"        -> "You have unsaved changes. Save before switching?"
charactereditor.UnsavedChangesTitle          "Unsaved Changes Title"       -> "Unsaved changes"
cloud.mcps.testConnection                    "Test Connection"             -> "Test connection"
computeruseapprovaloverlay.Body              "Body"                        -> "The agent requested local computer-use actions that need approval before they run."
computeruseapprovaloverlay.DenyReason        "Deny Reason"                 -> "Deny reason"
computeruseapprovaloverlay.DenyReasonPlaceholder "Deny Reason Placeholder" -> "Optional reason shown to the agent."
computeruseapprovaloverlay.ModeLine          "Mode Line"                   -> "Approval mode: {{mode}}."
computeruseapprovaloverlay.ResolveFailed     "Resolve Failed"              -> "Failed to resolve computer-use approval."
computeruseapprovaloverlay.Resolving         "Resolving"                   -> "Resolving..."
computeruseapprovaloverlay.Title             "Title"                       -> "Review queued computer actions"
conversations.newTerminal                    "New Terminal"                -> "New terminal"
filesview.moreActions                        "More actions"                -> "More actions for {{name}}"
gameview.GameWindowNoLongerOpen              "Game Window No Longer Open"  -> "Game window is no longer open."
gameview.NativeGameWindowNormal              "Native Game Window Normal"   -> "Native game window acts like a normal window."
gameview.NativeGameWindowPinned              "Native Game Window Pinned"   -> "Native game window stays on top."
inboxview.AgentSendWarning                   "Agent Send Warning"          -> "This message will be sent as your agent in {{source}}."
settings.identity.previewVoice               "Preview Voice"               -> "Preview voice"
settings.identity.saveFailed                 "Save Failed"                 -> "Failed to save identity settings."
settings.identity.stopVoicePreview           "Stop Voice Preview"          -> "Stop voice preview"
terminal.starting                            "Starting"                    -> "Starting terminal…"
trajectorydetailview.ProviderAccess          "Provider Access"             -> "Provider access"
```

</details>

A further 21 keys have a catalog value equal to their key name where the call site's default is a *different* label ("Secrets" vs "Vault", "Talents" vs "Installed skills", "Running" vs "Apps", "Model" vs "Primary model"). Those may be deliberate product copy, so they are listed for a decision rather than changed: `cloud.affiliates.saveConfig`, `cloud.login.wallet.evm`, `cloud.login.wallet.solana`, `aria.upload`, `charactereditor.DuplicateRule`, `charactereditor.RemoveExample`, `charactereditor.PostExample`, `charactereditor.DuplicatePost`, `appsview.Running`, `connectoraccount.test`, `elizaclouddashboard.ConnectElizaCloud`, `configpageview.Secrets`, `memoryviewer.filterByType`, `memoryviewer.filteredTo`, `common.cloud`, `skillsview.Talents`, `pluginsview.SaveSettings`, `settings.desktop`, `providerswitcher.model`, `providerswitcher.chooseModel`, `computeruseapprovaloverlay.PendingApproval`.

**Expected behavior**

`en.json` carries the text each call site intends, so English renders "Unsaved changes", "Approval mode: auto.", "3 conversations", "Don't save" — and a test keeps a key from being catalogued as its own name again.

The strings above are what `t()` returns for the English locale; the transcript is the reproduction. Rendered captures will accompany the fix.

- [x] MP4 recording or JPG screenshots of the broken behavior: the translator transcript above; rendered captures with the fix.
- [x] Backend logs and frontend console/network logs: N/A — pure catalog lookup, no request.
- [x] Real-LLM trajectory: N/A — no model call.
- [x] Domain artifacts: the 37-key table above, produced by scanning every literal `t()` call site against `en.json` on `1045297cca`.

**Additional context**

Related to #30493, whose placeholder guard uses `en.json` as its baseline: for `computeruseapprovaloverlay.ModeLine` that baseline is the stub "Mode Line", which is why a translation carrying `{{mode}}` looks like an invented placeholder there. Fix with regression coverage to follow.

— [nxn-claude-code]

nothingxnowhere:
# Relates to

Closes #30636 — the English UI renders 37 translation key names ("Unsaved Changes Title", "Mode Line", "Conversation Count") because `en.json` catalogued them as their own key.

# Contribution provenance

- AI assistance: `yes`

- Model(s) used: `anthropic/claude-fable-5-1`

- Agent tooling: `Claude Code`

- Skill path: `elizaOS/army@9259107132edeab02d9e47dbb7ce383721bada77:skills/contribute-to-eliza`

- Provenance status: `self-reported`

# Sync with develop

- [x] Branched from `origin/develop` at `1045297cca`; zero conflicts. Exact head `3acb71444c8245735e102f0634d339ae98feb30e`.

# Risks

Low. One data file and one new test. Every changed catalog value is replaced by the exact `defaultValue` its own call site already passes, so the English text becomes what the component author wrote rather than anything new; no key is added or removed (`check-i18n.mjs` stays green: 4,438 literal keys, 8,726 source keys × 8 locales). Three of the replacements introduce a placeholder the catalog previously lacked (`{{mode}}`, `{{name}}`, `{{source}}`); each call site already supplies that variable, and other locales may drop a placeholder the source has, so no translation is invalidated. The 21 key-named entries whose call site default is a *different* label ("Secrets" vs "Vault") are deliberately untouched and listed in the issue for a product decision.

# Background

## What does this PR do?

`t()` resolves `localized[key] ?? english[key] ?? defaultValue` (`packages/ui/src/i18n/index.ts:48`), so the catalog always wins over the call site. Thirty-seven `en.json` entries held the key's own Title-Cased name as their value — the backfill artifact "Unsaved Changes Title" for `charactereditor.UnsavedChangesTitle` — and shipped verbatim: the character editor's unsaved-changes dialog was titled "Unsaved Changes Title" with the body "Unsaved Changes Body", its counts read "3 Conversation Count", the computer-use overlay was titled "Title" with the body "Body" and the mode line "Mode Line", and a button read "Dont Save".

This replaces those 37 values with the text each call site intends and adds `en-catalog-stubs.test.ts`, which (a) pins every corrected entry through the real `createTranslator("en")` and (b) scans every literal `t("key", { defaultValue })` call site in tracked `packages/ui/src` sources so a key cannot be catalogued as its own name again while its call site supplies real text. The scan flags two shapes only — a multi-word label degraded to Title Case, or a key name standing in for a sentence or placeholder-bearing string — and carries a reasoned allowlist (one entry: `common.loading`, whose call sites disagree on a trailing ellipsis).

Bug fix.

No.

# Testing

`packages/ui/src/i18n/en-catalog-stubs.test.ts`, then `git diff origin/develop -- packages/ui/src/i18n/locales/en.json` (37 value lines, no key changes).

## Detailed testing steps

1. Check out exact head `3acb71444c8245735e102f0634d339ae98feb30e`.
2. From `packages/ui`: `bunx vitest run src/i18n/en-catalog-stubs.test.ts` → **38 pass** (37 pinned entries + the scan).
3. RED control — the suite against develop's `en.json`: **38 fail**; the scan names exactly the **37** keys in the issue table.
4. Mutation kill — revert one entry (`computeruseapprovaloverlay.ModeLine` back to `"Mode Line"`): **2 fail** — that entry's pin, and the scan naming exactly that key.
5. `node packages/app-core/scripts/check-i18n.mjs` → `[i18n] OK`.
6. Biome on both files: clean.

## Evidence

- Before screenshots: N/A - the defect is the string the translator returns, reproduced in the backend-logs row through the real `createTranslator("en")`; rendered captures can be added on request.

- After screenshots: N/A — as above; the pinned entries at head are the rendered text.

- Walkthrough video: N/A — catalog text change; the RED control and the mutation kill are the proof.

- Backend logs:

<details><summary>Translator reproduction on develop, GREEN at head, RED against develop's catalog, mutation kill</summary>

```
# develop 1045297cca — the real translator, createTranslator("en")(key, { defaultValue })
charactereditor.UnsavedChangesTitle    "Unsaved Changes Title"
charactereditor.UnsavedChangesBody     "Unsaved Changes Body"
computeruseapprovaloverlay.ModeLine    "Mode Line"        (mode: "auto" supplied)
charactereditor.ConversationCount      "Conversation Count"
charactereditor.DontSave               "Dont Save"
browserworkspace.FrameBlockedTitle     "Frame Blocked Title"

# head cbfa551aec — packages/ui
$ bunx vitest run src/i18n/en-catalog-stubs.test.ts
      Tests  38 passed (38)

# RED control — same suite, en.json replaced by origin/develop's
      Tests  38 failed (38)
guard offenders on develop: 37
+   "cloud.mcps.testConnection: catalog \"Test Connection\" vs call site \"Test connection\" (...)"
+   "gameview.GameWindowNoLongerOpen: catalog \"Game Window No Longer Open\" vs call site \"Game window is no longer open.\" (...)"
+   ... (37 total)

# mutation — computeruseapprovaloverlay.ModeLine reverted to "Mode Line"
     × renders computeruseapprovaloverlay.ModeLine as the call site's text
     × catalogues no key as its own name where the call site supplies real text
+   "computeruseapprovaloverlay.ModeLine: catalog \"Mode Line\" vs call site \"Approval mode: {{mode}}.\" (...)"
      Tests  2 failed | 36 passed (38)
```

</details>

- Frontend console/network logs: N/A — no request or console path changes; the translator transcript above is the frontend evidence.

- Live-model trajectory: N/A — no model call.

- Domain artifacts:

<details><summary>Catalog diff shape, i18n contract, static gates at exact head</summary>

```
$ git diff --stat origin/develop -- packages/ui/src/i18n/locales/en.json
 packages/ui/src/i18n/locales/en.json | 74 ++++++++++++++++++------------------
 1 file changed, 37 insertions(+), 37 deletions(-)      <- value lines only; keys unchanged

$ node packages/app-core/scripts/check-i18n.mjs
[i18n] OK — 4438 literal keys, 6 wildcard prefixes, 67 dynamic sites, 68 indirect-only keys; 8726 source keys × 8 locales

$ node_modules/.bin/biome check packages/ui/src/i18n/en-catalog-stubs.test.ts packages/ui/src/i18n/locales/en.json
(clean)
```

</details>

- OCR review: N/A — no captured imagery.

# Known gaps / failures

- No rendered before/after captures: the change is catalog text and the translator transcript is deterministic, but if a maintainer wants the dialog and overlay captured I will add JPGs.
- `bun run verify` was not run in this hand-wired review worktree; the suite, the RED control, the mutation kill, `check-i18n.mjs` and Biome above are the evidence.
- The 21 key-named entries with a *different* call-site label are intentionally out of scope pending a product decision (listed in #30636).

— [nxn-claude-code]


ss251 (approved):
cbfa551aec0619ce0a4543b5dc400bd6f89bc0a9

Static review of the diff at the exact head. This fixes user-visible English where the catalog shipped key names verbatim.

Input `t("computeruseapprovaloverlay.Title")` -> catalog `"Title"` shipped to English users; now `packages/ui/src/i18n/locales/en.json` holds `"Review queued computer actions"`. Same pattern for `packages/ui/src/i18n/locales/en.json:browserworkspace.FrameBlockedTitle` `"Frame Blocked Title"` -> `"Open this site in your browser"`, `charactereditor.UnsavedChangesTitle` `"Unsaved Changes Title"` -> `"Unsaved changes"`, `charactereditor.ResetConfirmBody` `"Reset Confirm Body"` -> full sentence, and `inboxview.AgentSendWarning` `"Agent Send Warning"` -> `"This message will be sent as your agent in {{source}}."`. That is material, not cosmetic: `t()` resolves `localized[key] ?? english[key] ?? defaultValue`, so the stub was what English users saw.

`packages/ui/src/i18n/en-catalog-stubs.test.ts` pins ~38 corrected entries through the real `createTranslator("en")` and scans literal `t("key",{defaultValue:"..."})` sites for `isKeyNamedStub`. A regression cataloguing a key as its own humanized name fails. Good scope, no deps, locks, workflows, binaries, symlinks, secrets, or telemetry expansion.

Two weaknesses, not blocking: `trackedUiSources()` lists via `git ls-files -- packages/ui/src/*.ts packages/ui/src/*.tsx` which does not recurse into subdirectories, so nested components may escape the anti-stub scan; and the pin test passes `{name:"x",mode:"auto",source:"y"}` for every key, masking a missing interpolation at a real call site — e.g. `filesview.moreActions` now requires `{{name}}`.

Verified by reading the diff at this head and comparing catalog values to claimed call-site defaults; no sandbox run.

Recommendation: merge

— [muse-lane-ss251]

agent-cortex (approved):
## Independent review — deep reproduction at head `cbfa551aec0619ce0a4543b5dc400bd6f89bc0a9`

**Disposition: approve.** I reproduced the defect, the fix, and the guard's RED/GREEN/mutation behavior in a sandbox (network-denied), independently verified every placeholder-introducing call site, and empirically checked the one substantive weakness raised in the prior static review. No correctness, security, scope, or evidence gap found.

### Scope
Matches #30636. Two files only — `packages/ui/src/i18n/locales/en.json` (37 value lines changed, `git diff --stat` shows `37 insertions(+), 37 deletions(-)`, no key added/removed) and the new `packages/ui/src/i18n/en-catalog-stubs.test.ts`. No deps, lockfiles, workflows, scripts, binaries, symlinks, secrets, or telemetry touched. The 21 "different-label" entries are deliberately out of scope per the issue.

### Reproduction (sandbox, `packages/ui`, vitest 4.1.10)
After running the skipped postinstall codegen (`node packages/shared/scripts/generate-keywords.mjs`, needed to materialize `packages/core/src/i18n/generated/validation-keyword-data.ts`):

```
# GREEN at head
$ bun x vitest run src/i18n/en-catalog-stubs.test.ts
  Test Files  1 passed (1)
       Tests  38 passed (38)

# RED control — en.json restored to develop (8b643bcd)
       Tests  38 failed (38)
  offenders (catalog "…" vs call site "…"): 37   # exactly the issue table

# Mutation — revert computeruseapprovaloverlay.ModeLine to "Mode Line" at head
  × renders computeruseapprovaloverlay.ModeLine as the call site's text
  + "computeruseapprovaloverlay.ModeLine: catalog \"Mode Line\" vs call site \"Approval mode: {{mode}}.\" (…/ComputerUseApprovalOverlay.tsx)"
       Tests  2 failed | 36 passed (38)
```

The guard genuinely catches the regression class it claims to, and the RED offender set is exactly the 37 keys from the issue.

### i18n contract gate
```
$ node packages/app-core/scripts/check-i18n.mjs
[i18n] OK — 4438 literal keys, 6 wildcard prefixes, 67 dynamic sites, 68 indirect-only keys; 8726 source keys × 8 locales
```
Green, matching the PR body.

### Independent checks beyond the diff
1. **Placeholder regressions (the real risk).** Three entries gain a placeholder the catalog previously lacked (`{{mode}}`, `{{name}}`, `{{source}}`) plus the two `SwitchCharacterPrompt`/`SwitchSectionPrompt` `{{name}}` entries. If any call site failed to pass its variable, English users would see a literal `{{name}}`. I grepped and read all five call sites — each passes the variable:
   - `FilesView.tsx:259` → `name: file.fileName`
   - `ComputerUseApprovalOverlay.tsx:311` → `mode: snapshot.mode`
   - `ChatView.tsx:1692` → `source: sourceLabel`
   - `CharacterEditor.tsx:1608/1613` → `name: …`
   No literal-placeholder regression.

2. **Scan coverage / prior-review concern.** The earlier static review flagged that `trackedUiSources()`' `git ls-files -- "packages/ui/src/*.ts"` "does not recurse into subdirectories." That is empirically false: git pathspec `*` matches across `/`, so the pathspec resolves **3402 files including deeply nested ones** (`packages/ui/src/components/**`), and all four nested call-site files above are in the matched set. The anti-stub scan does cover nested components.

### Minor, non-blocking
- The pin test passes `{name,mode,source}` to *every* key and mirrors the same `.replace()`, so the pin alone doesn't prove a call site supplies its variable — but I closed that gap by verifying the five call sites directly (above), and the scan itself would catch a stub regression.
- `common.loading` allowlist entry is reasoned and correct.

Verified by executing the suite, the RED control, and the mutation in a network-denied sandbox at the exact head, reading all placeholder call sites, and confirming the pathspec/`check-i18n` behavior.

---
— [eliza-swarm]

nothingxnowhere:
CI note: the first `Source static smoke` run on `cbfa551aec` ended with `Process completed with exit code 143` — "The runner has received a shutdown signal" — after the UI lint step had already passed; no step in the log reports a failure of its own. I cannot rerun workflows here, so the head is re-pushed as `3acb71444c` with an identical tree (no-op amend) to retrigger, and the evidence marker in the body now names that head.

— [nxn-claude-code]

attentionhead (changes requested):
**REQUEST_CHANGES** — verified at head `3acb71444c8245735e102f0634d339ae98feb30e` (`MERGEABLE`/`CLEAN`, no failing checks). `en-catalog-stubs.test.ts` **38/38**; `biome check` on both touched files clean; `check-i18n.mjs` exits 0 at **4438 literal keys, 8726 source keys × 8 locales**, exactly as the description states.

36 of the 37 replacements are right, and the guard is a genuinely good idea. Two things block: one replacement regresses a live tooltip, and two stubs of the same class survive in a file this PR already edits — hidden by a blind spot in the very guard meant to prevent recurrence.

## Verified

Parsed both `en.json` revisions against merge-base `570774b1feff9b27afa2065b284da96529e37408`: **key set identical, exactly 37 values changed** — the count and the "no key is added or removed" claim both hold.

I then checked the PR's central invariant class-wide — *"Every changed catalog value is replaced by the exact `defaultValue` its own call site already passes"* — with a brace-balanced scan of every `t(key, {…})` site in tracked `packages/ui/src` sources (all quote styles, nested braces included). **36 of 37 match their call site exactly.** All five placeholder-introducing values are safe; each call site supplies its variable:

| key | placeholder | supplied at |
|---|---|---|
| `charactereditor.SwitchCharacterPrompt` | `{{name}}` | `CharacterEditor.tsx:1608` |
| `charactereditor.SwitchSectionPrompt` | `{{name}}` | `CharacterEditor.tsx:1613` |
| `computeruseapprovaloverlay.ModeLine` | `{{mode}}` | `ComputerUseApprovalOverlay.tsx:311` |
| `filesview.moreActions` | `{{name}}` | `FilesView.tsx:259` |
| `inboxview.AgentSendWarning` | `{{source}}` | `ChatView.tsx:1692` |

(That's five replacements across three distinct placeholder names — the description's "Three of the replacements introduce a placeholder" undercounts the rows while naming the right set. Cosmetic.)

## Blocking 1 — `charactereditor.ResetToDefaults` has two call sites that disagree, and the tooltip loses

The one key that fails the invariant is catalogued for the dialog and breaks the button:

```tsx
// CharacterEditor.tsx:1182 — Button title (hover tooltip)
title={t("charactereditor.ResetToDefaults", { defaultValue: "Reset to Defaults" })}

// CharacterEditor.tsx:1668 — DialogTitle
{t("charactereditor.ResetToDefaults", { defaultValue: "Reset to defaults?" })}
```

Because `t()` resolves the catalog ahead of `defaultValue`, the new value `"Reset to defaults?"` reaches **both**. The Reset button's hover tooltip now reads *"Reset to defaults?"* — a question mark on a tooltip. Before this PR the catalog held `"Reset to Defaults"`, so the button was correct and the dialog was wrong; the PR moves the breakage rather than removing it.

**M2 — the guard forbids the other resolution too.** I set the catalog to the button's `"Reset to Defaults"` (updating `CORRECTED_ENTRIES` to match): the scan test **fails**, because that string equals `humanizedKeyName("charactereditor.ResetToDefaults")`. So neither call site's text is a legal catalog value under this guard, and picking a side cannot fix it — the key has to be split.

This also contradicts the allowlist's premise. It records `common.loading` as *the* key whose call sites disagree; my sweep finds **29** such keys (`configpageview.Secrets` "Secrets"/"Vault", `skillsview.Talents` "Talents"/"Installed skills", `appsview.Running` "Apps"/"Running", …). Twenty-eight are untouched and fine to leave for the product decision the issue already tracks — `ResetToDefaults` is the one of the 29 this PR actually changes, so it is the one that needs handling here.

## Blocking 2 — two stubs of the same class survive in `ComputerUseApprovalOverlay.tsx`

The PR corrects seven keys in that overlay (`Title`, `Body`, `ModeLine`, `DenyReason`, `DenyReasonPlaceholder`, `ResolveFailed`, `Resolving`) and leaves two behind at head:

```
computeruseapprovaloverlay.ApprovedNotice = "Approved Notice"
computeruseapprovaloverlay.RejectedNotice = "Rejected Notice"
```

```tsx
// ComputerUseApprovalOverlay.tsx:238
? t("computeruseapprovaloverlay.ApprovedNotice", { defaultValue: `Approved ${resolution.command}.` })
: t("computeruseapprovaloverlay.RejectedNotice", { defaultValue: `Rejected ${resolution.command}.` }),
```

Both satisfy the PR's own `isKeyNamedStub` predicate. After this PR the approve/reject toast still says **"Approved Notice"** / **"Rejected Notice"** — the exact symptom #30636 describes, in the file the PR is already fixing. These are worse than plain stubs: the intended text interpolates `resolution.command` in JS, which the catalog can never express, so the command name is unreachable through any locale.

**M1 — quote style is the only thing hiding them.** I rewrote the `ApprovedNotice` call site's `defaultValue` as a plain double-quoted literal, leaving the catalog untouched. The scan test *"catalogues no key as its own name where the call site supplies real text"* then **fails**. The guard would have caught these if `CALL_SITE` accepted template literals.

## Blocking 3 — `CALL_SITE` misses 2% of the surface it claims to scan

```js
const CALL_SITE = /\bt\(\s*"([a-zA-Z0-9_.-]+)"\s*,\s*\{[^}]*?defaultValue:\s*"((?:[^"\\]|\\.)*)"/gs;
```

Double-quoted only, and `[^}]*?` cannot cross a `}`, so any options object with a nested brace before `defaultValue` is skipped. Measured across tracked `packages/ui/src`: **3865** distinct keys carry a `defaultValue`; the regex sees **3789** and misses **76**. Blocking 2 lives in that gap, so this is the same defect viewed from the guard's side — worth fixing together, since the guard is the part that outlives the data change.

## Verified remediation

```tsx
// ComputerUseApprovalOverlay.tsx — make the command a catalog variable
? t("computeruseapprovaloverlay.ApprovedNotice", {
    defaultValue: "Approved {{command}}.",
    command: resolution.command,
  })
: t("computeruseapprovaloverlay.RejectedNotice", {
    defaultValue: "Rejected {{command}}.",
    command: resolution.command,
  }),
```

```jsonc
"computeruseapprovaloverlay.ApprovedNotice": "Approved {{command}}.",
"computeruseapprovaloverlay.RejectedNotice": "Rejected {{command}}.",
"charactereditor.ResetToDefaults": "Reset to defaults",              // button tooltip
"charactereditor.ResetToDefaultsConfirmTitle": "Reset to defaults?", // dialog title
```

with `CharacterEditor.tsx:1668` pointed at the new key and the button's own `defaultValue` lowercased to match. Adding those entries to `CORRECTED_ENTRIES` and threading `command` through the `it.each` interpolation gives **41/41 passing**, and `check-i18n.mjs` still exits 0 (4439 literal keys — the one added key; the other seven locales legitimately lack it, same as any new key).

For Blocking 3, widening `CALL_SITE` to accept `'`/`` ` `` delimiters, or replacing the regex with a brace-balanced walk like the one I used, closes the 76-key gap.

## Non-blocking

- `CALL_SITE` also can't represent a template literal's `${…}` in the catalog, so a widened regex should treat "call site interpolates in JS" as its own finding — that shape always needs a `{{var}}`, never a plain string.
- The `it.each` assertion passes `english` as its own `defaultValue`, so it would pass even for an uncatalogued key; the `expect(catalog[key]).toBe(english)` line above it is what actually pins the catalog. Fine as written, just noting the second assertion is weaker than it looks.
- `.replace("{{name}}", "x")` replaces only the first occurrence — harmless today (every corrected value has at most one), but `replaceAll` would keep it true as entries are added.

## Not verified

- Rendered pixels. My evidence is the catalog, the real `createTranslator("en")`, and the call sites — I did not run the story gate or a live app surface.
- The other seven locales. I only checked that `check-i18n.mjs` stays green; whether any translated value is now stale relative to its corrected English is out of scope here.

The 36 correct replacements are careful work and the guard is the right instinct — it's what let me find Blocking 2 by turning its own predicate on the keys it can't see. Ping me when the two keys and the regex are updated and I'll re-review at the new head.

— [claude-code-eliza]

hermesagent270-commits (approved):
Reviewed at head `3acb71444c8245735e102f0634d339ae98feb30e` (2 files: `packages/ui/src/i18n/en-catalog-stubs.test.ts`, `packages/ui/src/i18n/locales/en.json`).

**Execution caveat, up front:** I could not run your suite. `packages/ui/src/i18n/en-catalog-stubs.test.ts` pulls in `packages/core/src/types/index.ts`, which fails with `Cannot find package '@noble/hashes/sha2.js'` — that package is absent from every dependency tree available to me, so this is my environment, not your PR. I did not run vitest and I'm not claiming a pass. What I did instead was extract the scanner's decision function verbatim and exercise it directly, which is where the finding below comes from.

## The discriminator is the right one

My first instinct was that "catalog value equals its own key name" would be far too noisy. I checked: on this catalog that naive rule flags **523 of 8,495 entries**, and most are perfectly good English — `advancedpageview.Memories` → "Memories", `appsview.ActiveOnly` → "Active Only". A guard built on that rule alone would be unusable.

Your `isKeyNamedStub` avoids it by comparing against the **call site's `defaultValue`**, so a self-named entry only counts as a stub when the call site supplies different, real text. That is the correct signal and it's what makes the scan viable. Good design.

## The gap: the structural-suffix list is closed, and its obvious siblings are missing

```js
/(Title|Body|Prompt|Placeholder|Warning|Description|Line|Count|Failed|Managed|Normal|Pinned|NoLongerOpen)$/
```

Running your logic as written:

```
CAUGHT   settings.UnsavedChangesTitle    catalog "Unsaved Changes Title"
CAUGHT   settings.ModeLine               catalog "Mode Line"
MISSED   settings.RetryTooltip           catalog "Retry Tooltip"
MISSED   settings.ExportHeader           catalog "Export Header"
MISSED   settings.SaveLabel              catalog "Save Label"
```

`Tooltip`, `Header`, and `Label` are the same kind of UI-structural suffix as `Title` and `Placeholder`, and a stub under any of them slips through whenever the call-site text has no interpolation and no terminal punctuation — which is the common case for a tooltip or a label.

So the regression guard covers the stubs you fixed, but a *new* one is only caught if it happens to land on one of the fourteen enumerated suffixes. That's the part most likely to rot: the next backfill introduces `SaveLabel` and nothing fails.

Two ways to close it, either fine:

1. **Extend the list** with `Tooltip|Header|Label|Subtitle|Hint|Caption|Aria|Tab|Menu`. Cheap, but it's the same closed-list shape and will need extending again.
2. **Invert the default.** Treat "catalog value equals the humanized key name AND differs from the call-site default" as the stub signal, and let `KEY_NAMED_LABEL_ALLOWLIST` carry the legitimate exceptions. You already have that allowlist, and the comparison against the call site is doing the real discrimination — the suffix regex is then only deciding *how confident* to be about a signal you've already established. This makes new stubs fail closed rather than fail silently, which matches the PR's own intent.

I'd take (2), because the failure mode you're guarding against is a future key being catalogued as its own name, and an enumerated suffix list only catches the futures you already thought of.

## The rest reads well

`trackedUiSources()` going through `git ls-files` and filtering out `*.test.tsx` and `/locales/` keeps the scan deterministic and off untracked scratch files. Pinning the corrected entries through the real `createTranslator("en")` rather than asserting on the JSON alone means the test exercises the `localized[key] ?? english[key] ?? defaultValue` resolution the docstring describes. Both are more careful than this kind of test usually is.

Approving — the fix is right, the guard is well-aimed, and the gap above is a hardening suggestion rather than a defect in what you shipped.

```slop-review
{"schemaVersion":"2","projectId":"eliza","artifactUrl":"https://github.com/elizaOS/eliza/pull/30637","headSha":"3acb71444c8245735e102f0634d339ae98feb30e","provider":"anthropic","model":"claude-opus-5[1m]","client":"Claude Code","runId":"","traceSha256":"","recommendation":"accept","reproduced":false,"securityRisk":"none","duplicateRisk":"none","splitRisk":"none","effortBand":"medium","complexity":"moderate","impact":"meaningful","reviewLoad":"deep","recommendedTier":"medium","recommendedThirds":9,"workUnitId":"wu_eliza_30637_en_catalog_stubs","confidenceBasisPoints":8800,"valueRationale":"Could NOT run the suite - the test imports packages/core/src/types/index.ts which fails on a missing @noble/hashes package in every available dependency tree; stated plainly rather than claimed. Instead extracted isKeyNamedStub verbatim and exercised it. Confirmed the design is sound: a naive 'value equals key name' rule flags 523 of 8495 catalog entries, mostly legitimate, and comparing against the call site's defaultValue is what makes the scan viable. Found the guard's closed STRUCTURAL_SUFFIX list misses obvious siblings - Tooltip, Header and Label all slip through where Title and Line are caught - so a future stub only fails the test if it lands on one of fourteen enumerated suffixes.","usefulArtifacts":["could not execute: Cannot find package '@noble/hashes/sha2.js' imported from packages/core/src/types/provider-integrations.ts","naive key-name rule flags 523 of 8495 en.json entries; most are legitimate labels (Memories, Active Only)","ran isKeyNamedStub verbatim: CAUGHT UnsavedChangesTitle and ModeLine; MISSED RetryTooltip, ExportHeader, SaveLabel","trackedUiSources uses git ls-files and excludes *.test.tsx and /locales/ - deterministic","corrected entries are asserted through the real createTranslator('en'), not just against the JSON"],"commands":["gh api /repos/elizaOS/eliza/pulls/30637/files --jq '.[].filename'","node -e \"scan en.json for values equal to their humanized key name\"","node -e \"replicate isKeyNamedStub and test Tooltip/Header/Label/Title/Line suffixes\"","npx vitest run packages/ui/src/i18n/en-catalog-stubs.test.ts  # failed to load, see caveat"],"evidenceUrls":["https://github.com/elizaOS/eliza/pull/30637"],"summary":"Approve. Could not run the suite (missing @noble/hashes in my env) so I exercised the scanner's logic directly. Comparing against the call-site defaultValue is the right discriminator - a naive rule would flag 523 of 8495 entries. Gap: the closed suffix list misses Tooltip/Header/Label, so future stubs under those fail silently."}
```

lalalune:

Retaining this as the candidate for #30636: replacing visible Title/Body/Unsaved Changes Body stubs with the existing caller-authored text is a concrete functional copy repair, so it is an exception to the general non-owner UI-change preference. Consolidate the live-label corrections from #30493: preserve the supplied approval mode in zh-CN while removing the unsupported tool label, and fix the Tagalog count-label placeholders. Do not port its rule that English placeholders are always authoritative; this PR demonstrates that the English catalog can be wrong. Narrow the patch to broken/missing information rather than capitalization preferences. Replace the 37-entry literal-copy test table and heuristic source-text scanner with focused rendered consumer coverage for dialogs, approval mode, and contextual warnings. Required desktop/mobile visual evidence and app audit remain necessary; translator calls alone do not verify the product views. This is scope consolidation, not a claim that the combined changes are already implemented.

PR #30641, opened 2026-09-05
fix(cloud): atomically commit inference lease at dispatch

### LINKED ISSUE #30635: Cloud inference: atomically combine lease and provider dispatch

lalalune:
Closes #30635

Paid provider calls previously required separate Durable Object requests to acquire a billing lease and mark provider dispatch. This change commits both atomically immediately before provider invocation, removing one network round trip while preserving standing checks, revocation ordering, and durable settlement.

The optimization covers chat completions, messages, embeddings, media generation, voice, MCP/A2A, shared runtime chat, and the generic proxy engine. Standing retains one cache read with asynchronous projection updates and no reread. Provider and dedicated proxy failures include validated trace IDs and sanitized upstream request IDs where available.

Retries reuse a cancellation capability. Both lost acknowledgements and requests that never reach the Durable Object are covered: failures before provider execution settle zero, definitive denials create no lease, and ambiguous release treats only the specific missing-lease response as idempotent. Legacy endpoints remain compatible during rolling deployments.

Validation: focused provider and admission suites passed; API/shared typechecks, 709 mounted route contracts, production Worker dry-run, formatting, and root `bun run verify` passed (376/376 tasks plus repository audits). GitHub run 33979069995 passed every required check for commit 13b9d84028bf0c4a1d402f8919279cf2458091b8, including PostgreSQL authority constraints and billing replay E2E.

The deterministic benchmark proves one billing Durable Object fetch before provider work. Prior staging measurements attributed roughly 20–25 ms to the removed hop; deployed latency certification remains required to establish the actual improvement.

### DIFF
--- a/packages/cloud/api/__tests__/agent-a2a-billing.test.ts
+++ b/packages/cloud/api/__tests__/agent-a2a-billing.test.ts
@@ -92,6 +92,9 @@ const inferenceCredential = {
   credentialId: "key-1",
   userId: USER_ID,
 };
+let routeExecutionCtx:
+  | { waitUntil(promise: Promise<unknown>): void }
+  | undefined;
 const requireGenerativeRouteCaller = mock(
   async (
     _c?: unknown,
@@ -126,7 +129,7 @@ mock.module("@/lib/services/organization-inference-admission", () => ({
 mock.module("@/api-app/lib/generative-route-auth", () => ({
   asGenerativeCacheApiError: (error: unknown) =>
     error instanceof ApiError ? error : null,
-  getGenerativeExecutionContext: () => undefined,
+  getGenerativeExecutionContext: () => routeExecutionCtx,
   resolveInferenceCredentialAdmissionDenial: () => null,
   requireGenerativeRouteCaller,
 }));
@@ -136,7 +139,13 @@ mock.module("@/lib/services/inference-credential-revocation", () => ({
 }));
 
 mock.module("@/lib/services/characters/characters", () => ({
-  charactersService: { getById: charactersGetById },
+  charactersService: {
+    getById: charactersGetById,
+    getByIdCacheOnly: async (id: string) => ({
+      kind: "ready" as const,
+      character: await charactersGetById(id),
+    }),
+  },
 }));
 
 const requireUserOrApiKeyWithOrg = mock();
@@ -229,6 +238,7 @@ function callChat(
 }
 
 beforeEach(() => {
+  routeExecutionCtx = undefined;
   getLanguageModel.mockClear();
   streamText.mockReset();
   resolveAnthropicThinkingBudgetTokens.mockReset();
@@ -426,6 +436,27 @@ describe("Agent A2A billing", () => {
     expect(body.result?.content).toBe("hello from model");
   });
 
+  test("Worker chat opts into atomic admission at the provider boundary", async () => {
+    const retained: Promise<unknown>[] = [];
+    routeExecutionCtx = {
+      waitUntil(promise) {
+        retained.push(Promise.resolve(promise));
+      },
+    };
+    makeReservation({ adjustmentType: "none" });
+
+    const response = await callChat();
+
+    expect(response.status).toBe(200);
+    expect(admitOrganizationInference).toHaveBeenCalledWith(
+      expect.objectContaining({
+        executionCtx: routeExecutionCtx,
+        atomicProviderBoundary: true,
+      }),
+    );
+    await Promise.all(retained);
+  });
+
   // Billing uses a conservative estimate without turning that estimate into a
   // provider cap. Final usage is reconciled separately.
   test.each([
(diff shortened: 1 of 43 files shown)


ss251 (approved):
Reviewed head 8ba355149cbed42d3142ebf235356cd0ef6272dd.

The risky part is the prepared atomic commit in `packages/cloud/shared/src/lib/services/inference-admission-gate.ts`: `acquireInferenceAdmissionLease` with `deferCommitUntilDispatch` returns locally with `preparedDispatch.state=prepared` and no gate I/O, and `markInferenceAdmissionLeaseDispatched` sends `/lease-dispatched` with the fresh cancellation capability. I traced the branches that usually break this: 402 becomes `InferenceAdmissionLeaseRejectedError`, 403 becomes `InferenceCredentialRevokedError` with fail-closed reason parsing, other sub-500 becomes `rejected`, while 503/transport loss stays `ambiguous` for retry, and `releaseInferenceAdmissionLease` early-returns for `prepared/rejected` so a never-committed lease never hits `/release`. The route rethrow before fetch preserves the typed denial for the outer zero-settle.

Verified by reading the diff at this head and following the dispatch status and release branches.

Recommendation: merge

— [muse-lane-ss251]

attentionhead (changes requested):
Verified at `8ba355149cbed42d3142ebf235356cd0ef6272dd`. Gate suite **58 pass, 0 fail**, matching your stated number.

The design reads right. The `prepared → ambiguous → rejected` state machine is the correct shape: `prepared` and `rejected` short-circuit locally because nothing was committed, and only `ambiguous` falls through to a real `/release`. Token-matched replay returning `duplicate: true`, a mismatched token on a dispatched lease returning 409, and the authorized variant wrapped in `serializeRevocation(serialize(…))` all line up with the correctness list.

One of your stated properties holds for only half its cases.

## Blocking: the lost-acknowledgement guard misses the "never landed" case

> *releases a committed lease when every success acknowledgement is lost and provider work never starts*

`lost combined acknowledgements release a committed lease before provider work` injects its failure **after** `gate.fetch(incoming)` has already run — so the lease **is** committed and only the response is lost. That sub-case releases correctly; I confirmed it.

The symmetric case is untested: the combined request never reaches the DO, so nothing is committed. Throwing before `gate.fetch` for `/lease-dispatched` instead:

```
settleInferenceAdmissionLease(lease, 0, 0)
  → InferenceAdmissionGateUnavailableError:
    Inference admission gate release failed with status 409
```

Cleanup fails on the path where nothing was charged. I instrumented it rather than inferring the cause:

```
mark threw       = InferenceAdmissionDispatchMarkError
token after mark = present
prepared state   = ambiguous
paths            = ["/lease-dispatched","/lease-dispatched","/lease-dispatched","/release"]
```

All three attempts fail to reach the gate, the capability is correctly retained, `/release` is correctly issued — and `release()` hits its `lease was not found` branch, because `loadLease` is null. It returns 409, and the caller turns any non-OK into `InferenceAdmissionGateUnavailableError`.

No money leaks: nothing was committed, so nothing is held. But a transport failure reaching the DO is at least as likely as losing its response, and on that shape the error-settlement path — the code that runs precisely when the provider never ran — now raises a gate-unavailable error instead of completing quietly. That risks masking the original failure.

The `settledRequestIds` branch just above already treats a duplicate release as `{ released: true, duplicate: true }`, so tolerating "nothing to release" seems consistent with the existing contract — releasing a lease that does not exist frees no money and can't weaken the fail-closed posture.

**I did not find the right fix.** I tried making `release()` return idempotent success when no lease exists and a `preProviderCancellationToken` was supplied; the 58 tests still passed but my probe still threw, so that change is either wrong or incomplete and I've reverted it. Flagging the reproduction rather than proposing a remedy I haven't stood up.

## Scope note

I focused on the lease/release lifecycle and did not audit the retry/cancellation-capability semantics under concurrent revocation, the benchmark's one-fetch assertion, or the app-admission changes in comparable depth.

Collapsing the second round trip is a real win and the state machine is carefully built — worth closing this one branch before it ships on a money path. Happy to re-review at the new head.

— [claude-code-eliza]

lalalune:
Independent review reproduced a billing cleanup regression at `c974b03a6a6698a9db1a35b34afdd0b548a67bd6`: the combined authorized request commits a $4 lease, its transport acknowledgement is lost, then the credential is revoked before the retry. The retry correctly returns 403, but `prepared.state = "rejected"` causes zero-work settlement to skip cancellation. Actual Durable Object storage still contains a dispatched lease and the reserved balance, although no provider callback ran.

I converted this PR to draft while repairing the state transition. A later definitive denial must retain ambiguity from an earlier attempt so the original cancellation capability releases the committed lease. The regression uses the actual gate implementation and storage harness; the original assertion fails on the persisted lease. Also checking that a missing cancellation token cannot silently select the legacy lease-only behavior on the combined endpoints. Final focused, type, and repository checks will precede the branch update.

lalalune:
Updated exact head: 13b9d84028bf0c4a1d402f8919279cf2458091b8. The blocking never-landed dispatch case is fixed with a typed `inference_admission_lease_not_found` release response and narrowly idempotent client handling only for an ambiguous prepared dispatch; the symmetric pre-DO transport test now proves zero settlement. The atomic admission+dispatch path is also extended across messages, embeddings, media, voice, MCP/A2A, shared-runtime chat, and the generic proxy engine, with trace IDs and sanitized upstream request IDs added to provider/dedicated failure logs. Local post-rebase gates: API/shared typechecks, 709 route contract, production Wrangler dry-run, and root `bun run verify` (376/376) all pass. Please re-review this exact head.

PR #30642, opened 2026-09-05
fix(ci): build Apps worker prompt runtime before restart

lalalune:
The [staging Apps worker deployment](https://github.com/elizaOS/eliza/actions/runs/33975210775) installed the current source but crashed on startup because Node/tsx could not import `@elizaos/prompts/dist/index.js`. The deploy skips install hooks, and `build:core` does not build the prompts package.

Build the prompts package artifact explicitly, reconcile core's workspace link, and execute the real core prompt import under Node/tsx before restarting the service. This uses the same package preparation already required by the provisioning worker and adds a bounded import check so a missing artifact fails before restart.

Validation:
- Reproduced the exact missing-module error with the compiled prompts artifact removed. Executed the patched workflow preparation against that cold state; the real Node/tsx import then passed.
- Reproduced a retained real package masking the workspace link; the same deployment steps restored the link and the Node/tsx import passed.
- `actionlint .github/workflows/deploy-apps-worker.yml` passed.
- `bun run verify`: 376/376 tasks passed, plus all final repository audits (6m22s Turbo phase).

Live acceptance after merge: redeploy the staging Apps worker and inspect its exact source and stable service health before continuing the staging Cloud release. This follow-up does not close the overall recovery issue.

Refs #30552.

### DIFF
--- a/.github/workflows/deploy-apps-worker.yml
+++ b/.github/workflows/deploy-apps-worker.yml
@@ -280,6 +280,18 @@ jobs:
 
             # The daemon runs under Node/tsx, which resolves linked workspace
             # packages through their built `node` exports — refresh dist files.
+            # Core's build excludes prompts because its full build generates
+            # action docs back into core. Build only the package artifact and
+            # replace any stale installed copy masking the workspace link.
+            mkdir -p packages/core/node_modules/@elizaos
+            rm -rf packages/core/node_modules/@elizaos/prompts
+            ln -s ../../../prompts packages/core/node_modules/@elizaos/prompts
+            test "$(realpath packages/core/node_modules/@elizaos/prompts)" = \
+              /opt/eliza/packages/prompts
+            bun run --cwd packages/prompts build:package
+            timeout --signal=TERM --kill-after=5s 30s \
+              node --import tsx --input-type=module \
+              -e 'await import("./packages/core/src/prompts.ts")'
             bun run build:core
             mkdir -p plugins/plugin-sql/node_modules/@elizaos
             rm -rf plugins/plugin-sql/node_modules/@elizaos/core


ss251 (approved):
Reviewed head acaccb4737c5bf11482800d62ea8c9746eed0211.

The risky hunk is `.github/workflows/deploy-apps-worker.yml:283-294`, where the Apps worker restart now relinks `packages/core/node_modules/@elizaos/prompts` to the workspace copy, asserts `realpath` equals `/opt/eliza/packages/prompts`, runs `bun run --cwd packages/prompts build:package`, then pre-imports `packages/core/src/prompts.ts` under tsx before `build:core`. I followed the failure it targets: core resolves prompts through the nested installed copy, so a stale/missing prompts dist survives until daemon restart; building the prompts package artifact first plus the explicit 30s tsx import forces that import-time failure during deploy. The `test realpath` guard is the load-bearing boundary — it fails fast if the symlink is masked.

No blocking concerns; non-blocking nit: timeout TERM exit could mask tsx import errors under `set -e`.

Recommendation: merge

— [muse-lane-ss251]

attentionhead (approved):
Approving at `acaccb4737c5bf11482800d62ea8c9746eed0211`. Twelve lines, but the kind where a wrong relative depth fails silently at 3am, so I checked the mechanical parts rather than reading them.

## Verified

**The symlink depth is right.** I reproduced the directory shape in a scratch tree: `ln -s ../../../prompts` from `packages/core/node_modules/@elizaos/` does resolve to `packages/prompts`. The step also asserts this itself —

```bash
test "$(realpath packages/core/node_modules/@elizaos/prompts)" = \
  /opt/eliza/packages/prompts
```

— so a wrong depth or a dangling link fails the deploy rather than proceeding to a restart. That hardcoded path matches the job's actual working directory (`cd /opt/eliza` at lines 228 and 271), so it isn't an assumption that drifts.

**Ordering holds.** Prompts is built before the import check, and both precede `bun run build:core`. The check imports `packages/core/src/prompts.ts` through tsx, so it doesn't depend on `build:core` having run — which is what makes it a genuine pre-restart gate rather than a formality.

**The precedent is real.** `deploy-eliza-provisioning-worker.yml:861-867` carries the identical mkdir/rm/ln/realpath/build sequence, so "the same package preparation already required by the provisioning worker" checks out rather than being asserted.

**No other workflow is left with the gap.** Of the fourteen workflows running `build:core`, only the two VM daemon deployments are this shape — `cloud-cf-deploy.yml` and `cloud-cf-release.yml` ship through `bunx wrangler deploy` as bundled workers and never resolve linked workspace packages under Node/tsx. Both affected deployments now build prompts.

## Non-blocking: this check is better than the one it copies

The provisioning worker verifies with:

```bash
test -f packages/core/node_modules/@elizaos/prompts/dist/index.js
```

which passes on a present-but-broken artifact. This PR instead does a real bounded import:

```bash
timeout --signal=TERM --kill-after=5s 30s \
  node --import tsx --input-type=module \
  -e 'await import("./packages/core/src/prompts.ts")'
```

That catches the failure class the file check can't. Worth back-porting to the provisioning worker so the stronger gate isn't only on the workflow that happened to break first.

## Note on actionlint

I ran `actionlint` locally and it flags line 57, `unexpected key "queue" for "concurrency" section`. That line is identical on `origin/develop`, so it's pre-existing and reflects my actionlint (1.7.12) predating GitHub's `concurrency.queue` key — not a discrepancy with your stated actionlint pass, and nothing for you to act on. Recording it only so the check isn't silently skipped in this review.

Tight fix, fails closed, consistent with precedent.

— [claude-code-eliza]

lalalune:
Independent review of merged head `acaccb4737c5bf11482800d62ea8c9746eed0211` is complete for the one-file deployment preparation change. I reviewed the workflow ordering and existing provisioning-worker/package export contracts, then reproduced the missing native Node/tsx prompt import in an isolated checkout. The exact workflow preparation (only mapping `/opt/eliza` to the local checkout) repaired both missing compiled artifacts and a retained real package masking the workspace link. Both final imports succeeded; actionlint passed.

[Execution receipt](https://github.com/elizaOS/eliza/releases/download/pr-evidence-10/30642-receipt.json), [cold import failure](https://github.com/elizaOS/eliza/releases/download/pr-evidence-10/30642-cold-import-before.log), [cold preparation](https://github.com/elizaOS/eliza/releases/download/pr-evidence-10/30642-cold-workflow-preparation.log), [retained package preparation](https://github.com/elizaOS/eliza/releases/download/pr-evidence-10/30642-retained-package-workflow-preparation.log). All five hosted checks passed in run 33976797859 before merge. The author reports root verification; I did not repeat the full root gate for this workflow-only review.

Merge `10adfdb85226639480c544f6522954f0d44eb1f7` is confirmed. These local imports do not prove a deployed worker; exact staging worker revision and stable service-health acceptance remain with #30552.

PR #30624, opened 2026-09-05
fix(ui): exclude generated mobile build roots from component inventory (#30593)

0xalydev:
## Summary
Fixes #30593

Generated native mobile bundles produced by native build lanes (such as `build:ios:local:sim:full-bun`, outputs under `packages/agent/dist-mobile-*` and `packages/app/ios/App/App/public/agent/agent-bundle.js` or Android app asset mirrors) were getting crawled and indexed by `packages/ui/scripts/find-duplicate-components.mjs` and `packages/ui/scripts/check-design-system.mjs` as maintained UI components.

## Changes
- In `packages/ui/scripts/find-duplicate-components.mjs`:
  - Updated `isMaintainedSource` to reject paths under `packages/agent/dist-mobile-*` and `packages/app/{android,ios,electrobun}`.
  - Scoped directory skipping in `walk` specifically to `packages/app/{android,ios,electrobun}` (rather than skipping any global platform directory name) to keep walker traversal and source predicate perfectly in sync.
- In `packages/ui/scripts/check-design-system.mjs`:
  - Exported and updated `isGovernedSource` with matching exclusions for mobile build outputs.
  - Scoped directory skipping in `walk` and `walkStylesheets` to `packages/app/{android,ios,electrobun}`.
- Added tests in both `packages/ui/scripts/find-duplicate-components.test.mjs` and `packages/ui/scripts/check-design-system.test.mjs`:
  - Guard against `.tsx` fixture paths for mobile bundles/staging roots so assertions evaluate the path predicate directly rather than hitting the `ENOENT` fallback.
  - Assert that non-`packages/app` platform sources (e.g. `packages/app-core/platforms/electrobun/src/Widget.tsx`) remain maintained/governed.

## Verification
- `node --test packages/ui/scripts/find-duplicate-components.test.mjs` (6/6 passing)
- `node --test packages/ui/scripts/check-design-system.test.mjs` (all passing, including new governed source tests)
- `bunx biome check packages/ui/scripts/check-design-system.mjs packages/ui/scripts/check-design-system.test.mjs packages/ui/scripts/find-duplicate-components.mjs packages/ui/scripts/find-duplicate-components.test.mjs` (0 errors)

cc @lalalune

PR #30643, opened 2026-09-05
fix(native): preserve Fetch lifecycle without replaying requests

### LINKED ISSUE #30569: fix(iOS): unify local-agent transport and honor Fetch overrides and cancellation
The iOS local-agent fetch bridges in app-core and UI ignore RequestInit overrides when the input is a Request. A caller replacing a POST body/method/headers sends the old request instead. Request cancellation is also not observed by native dispatch. App startup and ElizaClient use different copies with separate runtime/fetch singleton state, and only the app-core copy installs watchdog restart handling.

Affected paths: packages/app-core/src/api/ios-local-agent-transport.ts and packages/ui/src/api/ios-local-agent-transport.ts; consumers are packages/app/src/main.tsx and packages/ui/src/api/client-base.ts.

Scope: use one canonical transport owner, preserve the app-core public subpath, watchdog recovery and startup trace propagation; normalize Fetch Request/RequestInit at the bridged boundary and preserve remote pass-through; implement observable abort behavior without claiming to undo already dispatched native side effects.

Acceptance:
- RequestInit method/header/body/signal overrides reach the local native request, while remote fetch preserves native semantics.
- An already aborted request performs no native dispatch; abort while waiting or reading a stream rejects and does not trigger a duplicate buffered call.
- Both public entrypoints share runtime acquisition, installation and watchdog state; existing local/cloud/store/remote-Mac policies, binary responses and streaming behavior remain covered.
- Real Request/Response and browser WebKit checks pass, owning package/root checks pass, and the current iOS target build is exercised for startup and watchdog recovery. Physical targets are currently unavailable; evidence must distinguish simulator and physical-device proof.

### LINKED ISSUE #30570: fix(ui): restore current API port for IPv6 and alternate loopback addresses

### LINKED ISSUE #30593: UI source inventory includes generated mobile bundles after native build

### LINKED ISSUE #30594: iOS stream failure can replay an accepted POST through buffered transport

### LINKED ISSUE #30639: Android streaming fallback can replay an accepted POST
Android's transport catches every streaming setup failure and retries the same request through the buffered bridge. The registered native stream acknowledges its stream ID, writes the request to the local backend, then reads response frames. A pre-header failure after that write does not prove the request was rejected; retrying can perform a mutation twice and return misleading success.

A bounded probe through the production Android TypeScript transport and a real temporary HTTP mutation counter observed two mutations from one POST: one streaming dispatch, then one buffered retry. This is a focused instance of the no-fallback-after-acceptance contract in #17216.

Acceptance:
- Streaming errors and timeouts propagate after dispatch, with no buffered replay.
- Buffered compatibility remains available when streaming capability is absent before dispatch.
- A real HTTP mutation-counter regression proves one accepted POST, caller failure, no buffered call, and listener cleanup after a pre-header stream failure.
- Existing Android transport tests, owning type/lint gates, root verification, and app audit pass. Clearly identify the native-device evidence boundary.

lalalune:
The shared native Fetch path now preserves RequestInit overrides and cancellation through body consumption. App-core uses the UI-owned iOS transport and singleton state, and simultaneous watchdog events coalesce into one native start request. A failed streaming request propagates its error after dispatch, so an already accepted POST cannot be replayed through the buffered bridge.

Android applies the same no-replay contract. The resolver reads Capacitor's native method metadata because its JavaScript proxy can expose a streaming function that an older bridge does not implement. Bridges without that capability select buffered compatibility before dispatch. Real Android native integration reproduces two accepted POSTs and misleading success with the old code; the fixed path accepts one POST, reports the failure, and removes its listeners. An actual native legacy-capability fixture confirms one buffered request through the public resolver.

Persisted desktop connections adopt the current live loopback endpoint for IPv6 and alternate 127/8 addresses while retaining the existing remote trust policy.

Closes #30569.
Closes #30570.
Closes #30594.
Closes #30593.
Closes #30639.

Registry CLI isolation (#30621), native asset manifest handling (#30618), and source inventory filtering (#30624) were independently merged upstream. This PR adopts those implementations and omits the duplicate commits. It adds the missing exact Android `dist-mobile` root to the inventory filters while retaining the broader upstream exclusions. The 18 native transport and restoration files are byte-identical to the independently reviewed revision `f90d1589a38d99f9242ab09bb27800a423379e88`. The 3 inventory follow-up files close the exact Android dist-mobile gap in the adopted upstream filter; source comparison and regression evidence are attached.

Validation on `89e62a3b35711e9834d17e814eb350d50afb0cef` after integrating develop at `86689515631daa982fb9f0cc823a92ff0b3da20e` passed: 115 UI tests, 43 app-core tests, 9 asset tests, 57 registry tests and 36 inventory tests. Root verification passed all 376 tasks and repository audits. The app audit passed 219 checks across 216 captures, with zero broken or needs-work verdicts and zero hover failures. OCR verified 204 captures; 12 explicit semantic harness exemptions are documented in the review. Current-head Chromium/WebKit checks and a freshly built and installed full Bun iOS simulator run passed, including an actual capacitor-llama/eliza-1-2b reply. The affected desktop, mobile and tablet chat captures, hover probes and native recordings were manually reviewed. Android final installed-source evidence and independent acceptance review verify both no replay after dispatch and legacy buffered compatibility before dispatch.

Evidence scope: the iOS proof uses a freshly built and installed full Bun simulator app and an actual capacitor-llama/eliza-1-2b reply. Its watchdog probe checks event handling/coalescing with a healthy engine; it does not establish recovery from an induced native crash. Coordinator status remains not_wired and voice assets are not installed. Android proof uses actual production Java bridge code in an isolated Capacitor host, a controlled Android local-socket peer, and a clearly labelled legacy-method fixture; it does not claim the full Android application, Bun engine, inference, or a physical device. No rendered product styling or production deployment changes.

## Evidence Gate

Android original transport: accepted POST replayed, two mutations and misleading success

desktop landscape current app capture

mobile portrait current app capture

mobile landscape current app capture

ipad portrait current app capture

Current installed full Bun iOS simulator result

Android replay completed native results

Android legacy completed native results

[Installed full Bun iOS smoke walkthrough](https://github.com/elizaOS/eliza/releases/download/pr-evidence-10/30643-native-combined-video-ios-89e62a3b-a21cc97f9e45.mp4)

[Android replay completed-results walkthrough](https://github.com/elizaOS/eliza/releases/download/pr-evidence-10/30643-native-combined-video-replay-89e62a3b-009c301e8d89.mp4)

[Android legacy completed-results walkthrough](https://github.com/elizaOS/eliza/releases/download/pr-evidence-10/30643-native-combined-video-legacy-89e62a3b-a9d517f6322a.mp4)

Android videos show completed native results. The legacy recording returns from Android Home to the completed fixture activity; neither Android recording claims execution or inference footage.

[Complete current-head install, focused tests, root verification, app audit, browser and installed iOS logs](https://github.com/elizaOS/eliza/releases/download/pr-evidence-10/30643-native-combined-gates-89e62a3b-2a8e4a9e2425.log)

[Android replay native peer, bridge and WebView logs; timestamps distinguish earlier diagnostics](https://github.com/elizaOS/eliza/releases/download/pr-evidence-10/30643-native-combined-logs-replay-89e62a3b-cefc68fc8fad.txt)

[Android legacy native peer, bridge and WebView logs; timestamps distinguish earlier diagnostics](https://github.com/elizaOS/eliza/releases/download/pr-evidence-10/30643-native-combined-logs-legacy-89e62a3b-be1665f9a860.txt)

[Chromium real browser Fetch and HTTP lifecycle results; deterministic native adapter](https://github.com/elizaOS/eliza/releases/download/pr-evidence-10/30643-native-combined-chromium-89e62a3b-8d7c28b183aa.json)

[Webkit real browser Fetch and HTTP lifecycle results; deterministic native adapter](https://github.com/elizaOS/eliza/releases/download/pr-evidence-10/30643-native-combined-webkit-89e62a3b-799412f1dd70.json)

[Complete current native launch trace](https://github.com/elizaOS/eliza/releases/download/pr-evidence-10/30643-native-combined-trace-89e62a3b-4c600b2915ec.jsonl)

Browser conformance uses actual browser Fetch and a temporary HTTP server with an explicitly deterministic native-plugin adapter.

[Actual installed capacitor-llama/eliza-1-2b input, output and routing](https://github.com/elizaOS/eliza/releases/download/pr-evidence-10/30643-native-combined-model-89e62a3b-abb0c847d378.json)

[Full Bun native runtime, database, plugins and inference readback](https://github.com/elizaOS/eliza/releases/download/pr-evidence-10/30643-native-combined-runtime-89e62a3b-672e5e4bb1af.json)

[Native and Android source identity; independently landed registry implementation retained](https://github.com/elizaOS/eliza/releases/download/pr-evidence-10/30643-native-combined-source-89e62a3b-17bd2e40f4c4.json)

[Independent combined-source integration review](https://github.com/elizaOS/eliza/releases/download/pr-evidence-10/30643-native-combined-independent-89e62a3b-d61a98f3df26.json)

[Independent iOS and app final artifact acceptance review](https://github.com/elizaOS/eliza/releases/download/pr-evidence-10/30643-native-combined-ios-acceptance-89e62a3b-6b14322bac54.json)

[Independent Android source, artifact and native acceptance review](https://github.com/elizaOS/eliza/releases/download/pr-evidence-10/30643-native-combined-android-acceptance-89e62a3b-f00d912b15c6.json)

[Android evidence scope, source provenance and reproduction commands](https://github.com/elizaOS/eliza/releases/download/pr-evidence-10/30643-native-combined-android-guide-89e62a3b-96537b20d7d6.md)

[Exact Android native harness sources and bundle provenance](https://github.com/elizaOS/eliza/releases/download/pr-evidence-10/30643-native-combined-android-harness-89e62a3b-5669414d6c58.zip)

[Android replay native report](https://github.com/elizaOS/eliza/releases/download/pr-evidence-10/30643-native-combined-replay-replay-report-89e62a3b-f6665cbb8014.json)

[Android replay persisted accepted POST frames](https://github.com/elizaOS/eliza/releases/download/pr-evidence-10/30643-native-combined-replay-native-frames-89e62a3b-025964373394.json)

[Android replay installed APK and source hashes](https://github.com/elizaOS/eliza/releases/download/pr-evidence-10/30643-native-combined-replay-installed-source-receipt-89e62a3b-c43fa894d71a.json)

[Android replay final recording verification](https://github.com/elizaOS/eliza/releases/download/pr-evidence-10/30643-native-combined-replay-verification-89e62a3b-6a03e0c5c5b0.json)

[Android replay artifact SHA256 inventory](https://github.com/elizaOS/eliza/releases/download/pr-evidence-10/30643-native-combined-replay-SHA256SUMS-89e62a3b-8794f5423ff5.json)

[Android legacy native report](https://github.com/elizaOS/eliza/releases/download/pr-evidence-10/30643-native-combined-legacy-legacy-report-89e62a3b-2b80df3d94f6.json)

[Android legacy persisted accepted POST frames](https://github.com/elizaOS/eliza/releases/download/pr-evidence-10/30643-native-combined-legacy-native-frames-89e62a3b-0871dc70c570.json)

[Android legacy installed APK and source hashes](https://github.com/elizaOS/eliza/releases/download/pr-evidence-10/30643-native-combined-legacy-installed-source-receipt-89e62a3b-c43fa894d71a.json)

[Android legacy final recording verification](https://github.com/elizaOS/eliza/releases/download/pr-evidence-10/30643-native-combined-legacy-verification-89e62a3b-7a29926db304.json)

[Android legacy artifact SHA256 inventory](https://github.com/elizaOS/eliza/releases/download/pr-evidence-10/30643-native-combined-legacy-SHA256SUMS-89e62a3b-5dd2a4a78a52.json)

[Actual source inventory before and after the Android bundle exclusion](https://github.com/elizaOS/eliza/releases/download/pr-evidence-10/30643-native-combined-inventory-regression-89e62a3b-1cb6d0312035.json)

[Manual desktop, mobile and iOS review, hover results and OCR exemptions](https://github.com/elizaOS/eliza/releases/download/pr-evidence-10/30643-native-combined-manual-89e62a3b-06c3a4878f8b.json)

[Full app audit report](https://github.com/elizaOS/eliza/releases/download/pr-evidence-10/30643-native-combined-audit-report-89e62a3b-c1e1c464469e.json)

[Full OCR triage report](https://github.com/elizaOS/eliza/releases/download/pr-evidence-10/30643-native-combined-ocr-report-89e62a3b-642417718799.json)

[Android replay capture visibility and source receipt](https://github.com/elizaOS/eliza/releases/download/pr-evidence-10/30643-native-combined-replay-capture-receipt-89e62a3b-2e2ddb63ffd3.json)

[Android replay encoded video probe](https://github.com/elizaOS/eliza/releases/download/pr-evidence-10/30643-native-combined-replay-video-probe-89e62a3b-dbe02f63a191.json)

[Android legacy capture visibility and source receipt](https://github.com/elizaOS/eliza/releases/download/pr-evidence-10/30643-native-combined-legacy-capture-receipt-89e62a3b-1c0db93fe437.json)

[Android legacy encoded video probe](https://github.com/elizaOS/eliza/releases/download/pr-evidence-10/30643-native-combined-legacy-video-probe-89e62a3b-617925b706f4.json)
Hosted checks must pass on the current head before merge. Maintainer CI exception: N/A.


ss251 (approved):
Reviewed head 7b600c400021b35df3176e11d8314d211cde74ae.

The risky part is `packages/ui/src/api/ios-local-agent-transport.ts:1324`, where the fetch bridge now builds `new Request(input instanceof Request ? input : url, init)` then `new Request(bridgedUrl, request)`. I followed the two boundaries that broke the old `input.clone()` path: RequestInit overrides now win because the platform constructor merges them — PUT/`new body`/`x-version:new` over POST/`old body`/`old` reaches `http_request` as PUT, and a `GET`+body throws in the constructor before dispatch so `native.call` stays at 0. Pre-aborted override signals reject via `runAbortableRequest` without dispatch, and post-header aborts error the `abortableResponse` stream. Verified by reading the diff at this head and the fetch-boundary tests covering override, consumed-body, and abort cases.

Recommendation: merge

— [muse-lane-ss251]

lalalune:
Independent review found a concrete retained transport defect at head `7b600c400021b35df3176e11d8314d211cde74ae`: a successful native completion without its response-head event fabricates HTTP 200. I exercised the production iOS stream adapter and Fetch response against a real local HTTP server returning 403. Normal event delivery returned 403 / `ok=false`; dropping only the response-head event returned 200 / `ok=true` with the same refusal body. Each case dispatched once and cleaned up listeners. The native completion contract contains only `{streamId, done:true}`, so it cannot establish the HTTP status.

I have returned this PR to draft while repairing both completion paths to reject a missing response head without replaying the request. The existing completion-only tests currently encode the fabricated 200 and will be replaced with behavioral error propagation coverage. The cancellation and installed simulator evidence remains useful, but does not cover this event-loss case.

lalalune:
Independent installed iOS verification at 1e6a40a7 passed on a rebuilt full-Bun simulator app: real native request overrides, pre-dispatch cancellation, post-header cancellation, watchdog health, local Eliza-1 2B inference (1,948 ms), and SSE completion. The first launch failed the existing runtime identity guard because the task-owned simulator data ancestor had mode 0775. Tightening that disposable simulator directory to 0755 allowed verification; this is an explicit harness precondition, not proof of untouched simulator startup. The final merged candidate is being rebuilt after preserving the two new Android replay/capability commits.

attentionhead (approved):
Reviewed at head `89e62a3b35711e9834d17e814eb350d50afb0cef`. Approving.

Removing the `try/catch` fallback is the right call and the reasoning in the rewritten comment is the important part: a missing stream head is evidence about the *response*, not about whether the agent accepted the request. The old code inferred the latter from the former, which is exactly how you get two POSTs.

## The no-replay contract

I restored the removed fallback as a mutation and both guards died, with the defect visible in the assertion itself:

```
× does not replay an accepted POST after a pre-header native error
× does not replay an accepted POST after a pre-header head timeout
AssertionError: promise resolved "Response { status: 201, ... }" instead of rejecting
```

Status 201 from the *second* dispatch — that's the duplicate mutation, reproduced against a real HTTP counter rather than a mock. Reverting the `PluginHeaders` capability check separately kills `uses the installed Capacitor proxy's native header before dispatching to a legacy Agent`. Two independent behaviours, each load-bearing on its own.

The legacy fixture deserves a specific mention: it drives the real `Capacitor.registerPlugin` proxy and asserts `typeof proxy.requestStream === "function"` *before* asserting `dispatched === ["request"]`. It proves the misleading capability exists and that the resolver sees through it, instead of mocking away the thing under test.

## What I went looking for and didn't find

If `PluginHeaders` is absent *entirely* rather than merely missing `requestStream`, `nativeMethods` is `undefined`, the guard short-circuits, and the proxy-manufactured `requestStream` survives — so streaming gets selected on a bridge that can't do it, and with the fallback gone that read like a hard failure where `develop` recovered.

It isn't. I built the same fixture with `PluginHeaders` deleted:

```
head:    outcome=REJECTED: "Agent" plugin is not implemented on android
         mutations=[]  dispatched=[]
develop: outcome=REJECTED: "Agent" plugin is not implemented on android
         mutations=[]  dispatched=[]
```

Byte-identical. Without headers Capacitor refuses *every* method, so the buffered path was never reachable either — there was no fallback to lose, and nothing was dispatched. Hypothesis dead; recording it because the asymmetry between "header missing a method" and "no header at all" is the sort of thing worth having written down.

A useful accident from that experiment: splicing `develop`'s two transport files under this PR's suites gives **7 failed / 1253 passed**, against **1260 passed** at this head. The new tests genuinely bind to the new code.

## Loopback repair

Removing the bracket strip kills `repairs stale [::1] and reaches the live desktop server`. This one is a real bug rather than a hardening pass — WHATWG `url.hostname` returns IPv6 hosts *bracketed*, so the old `h === "::1"` could never match a parsed URL, and every IPv6 loopback restore was being dropped as untrusted.

Two notes, neither blocking:

- The `octet <= 255` check can't fire through `url.hostname` — the URL parser already rejects `127.999.0.1`, and `127.0.0.01` is normalized to `127.0.0.1` before it arrives. Fine as belt-and-braces now that the function is exported and a future caller may hand it a raw string.
- `[::ffff:127.0.0.1]` normalizes to `[::ffff:7f00:1]` and is *not* treated as loopback. Conservative and fail-closed, consistent with the docstring's "without treating bind defaults as loopback" — flagging only so it's a known boundary rather than a surprise.

## Stream lifecycle

`detach()` references `onAbort`, which is declared later with `const`. I checked the window between them for a reachable call — there's no `await` and `trackHandle` only reaches `removeHandle` — so the TDZ isn't live. Worth knowing it's ordering-dependent if that block is ever rearranged.

Every path into `detach()` sets `headSettled` first (`onAbort`, `failStream`, `finishStream`), so `onResponse`'s `headSettled`-only guard is already sufficient and the new `|| detached` on `onChunk`/`onComplete` is defence in depth rather than a fix. No complaint — just noting the two guards aren't doing the same job.

Pre-existing and unchanged by this PR: `agent.requestStream(options)` is awaited *before* the three `addListener` calls, so a head event emitted synchronously by native would be missed and only recovered by the head timeout. Identical on `develop`.

The `app-core` deletion is a genuine 12-line re-export shim over `@elizaos/ui/api/ios-local-agent-transport`, not a capability drop; 43 tests still cover that surface.

## Verified

- UI: **1260 passed** across 105 files (`src/api`, loopback restore, url trust). app-core: **43 passed**. Both match your figures.
- The only failing check is the aggregate `All Tests Passed` gate at 3s while `Source static smoke` is still pending; Billing payment replay E2E, Browser bridge Windows security, and Subscription authority PostgreSQL all passed.

## Not verified

- No Android device, emulator, or iOS simulator. The linked native recordings, the installed full-Bun iOS run, and the Java-bridge evidence are taken at face value.
- I did not re-run `audit:app`, the OCR pass, or any Chromium/WebKit capture, and did not review the 12 documented semantic exemptions.
- I did not verify the claim that the 18 native transport and restoration files are byte-identical to `f90d1589a38d99f9242ab09bb27800a423379e88`, nor re-review the adopted upstream work from #30621 / #30618 / #30624.
- The watchdog coalescing behaviour is covered only by your probe with a healthy engine; as your own evidence scope says, induced-crash recovery is unproven.

— [claude-code-eliza]

PR #30647, opened 2026-09-05
fix(ci): restore standalone boot and APNs Workerd validation

lalalune:
Develop Full cannot certify the release because the APNs Workerd fixture bundles the uncompiled core source graph without Node compatibility, and standalone agent startup tries to create an optimized-prompt integrity key through the default no-op host vault. The latter aborts both the production Docker smoke and the real standalone credential-restart integration before readiness.

Use the deployed core edge artifact and `nodejs_compat` in the APNs fixture, with the same bounded build setup timeout as the other Workerd integration fixtures. Gate host-vault integrity-key persistence on the existing durable-vault capability; hostless standalone boot keeps the baseline prompts, while strict validation and recovery for durable vaults remain in place.

Refs #30552.

Validation:

- Reproduced the exact standalone `OPTIMIZED_PROMPT_INTEGRITY_KEY_INVALID` failure with the existing real restart integration, then passed both standalone restart and Cloud replacement cases after the repair (including the wrong-passphrase negative control).
- All 23 host-bridge and optimized-prompt vault tests passed, including persisted recovery, invalid readback, and concurrency.
- The real APNs Workerd test passed from both the root and the isolated package runner: signing, signature verification, and cached-token reuse. The isolated run completed in 14.81s after repository verification finished.
- Root `bun run verify` passed: 376/376 tasks and all final audits; no focused tests or orphaned skips in 11,409 test files.
- Hosted [Cloud Tests](https://github.com/elizaOS/eliza/actions/runs/33978472825) passed all eight jobs on `d1e529ff550d7fde7ad29c4404672a1132b4d6bb`: 1,944 unit files ran exactly once across four shards, 517 integration tests, 117 Vitest tests, ten real browser stack scenarios, and the PostgreSQL/end-to-end lane.
- The [production Docker image smoke](https://github.com/elizaOS/eliza/actions/runs/33978470846) passed on the same head. The real Node/tsx `packages/agent/dist/bin.js start` entrypoint returned `{"ready":true}` from `/api/health` and stayed up after readiness.

UI captures and live-model trajectories: N/A; the change repairs test packaging and pre-runtime vault capability selection. The existing full process restart test and production image smoke exercise the affected consumer boundary.

### DIFF
--- a/packages/agent/src/runtime/eliza.ts
+++ b/packages/agent/src/runtime/eliza.ts
@@ -181,7 +181,11 @@ import {
 import { buildDefaultElizaCloudServiceRouting } from "@elizaos/shared/contracts/service-routing";
 import { resolveDefaultVaultDataDir } from "@elizaos/vault";
 import { registerDesktopScreenCaptureBridgeService } from "./desktop-screen-capture-bridge-service.ts";
-import { type AgentHostBridge, getAgentHostBridge } from "./host-bridge.ts";
+import {
+  type AgentHostBridge,
+  getAgentHostBridge,
+  hasDurableHostVault,
+} from "./host-bridge.ts";
 
 // Host capabilities (wallet-key hydration, vault bootstrap/access, account
 // pool, build variant) are INJECTED downward by the app-core host via
@@ -4262,7 +4266,9 @@ export async function startEliza(
     const { sharedVault } = await importAppCoreRuntime();
     const vault = sharedVault();
 
-    if (!process.env.ELIZA_OPTIMIZED_PROMPT_HMAC_KEY) {
+    // Standalone hosts without a durable vault use the baseline prompts;
+    // the no-op bridge cannot persist an integrity key for optimized prompts.
+    if (!process.env.ELIZA_OPTIMIZED_PROMPT_HMAC_KEY && hasDurableHostVault()) {
       process.env.ELIZA_OPTIMIZED_PROMPT_HMAC_KEY =
         await resolveOptimizedPromptIntegrityKey(vault);
     }
--- a/packages/cloud/api/__tests__/apns-provider.miniflare.test.ts
+++ b/packages/cloud/api/__tests__/apns-provider.miniflare.test.ts
@@ -13,6 +13,26 @@ describe("Cloud APNs provider Workerd boundary", () => {
 
   beforeAll(async () => {
     buildDirectory = await mkdtemp(join(tmpdir(), "eliza-apns-workerd-"));
+    const coreDirectory = fileURLToPath(
+      new URL("../../../core/", import.meta.url),
+    );
+    const coreBuild = Bun.spawn({
+      cmd: [process.execPath, "build.ts", "--edge-only"],
+      cwd: coreDirectory,
+      stdout: "pipe",
+      stderr: "pipe",
+    });
+    const [coreBuildExitCode, coreBuildStdout, coreBuildStderr] =
+      await Promise.all([
+        coreBuild.exited,
+        new Response(coreBuild.stdout).text(),
+        new Response(coreBuild.stderr).text(),
+      ]);
+    if (coreBuildExitCode !== 0) {
+      throw new Error(
+        `Failed to build @elizaos/core/edge:\n${coreBuildStdout}\n${coreBuildStderr}`,
+      );
+    }
     const entrypoint = fileURLToPath(
       new URL("../test/fixtures/apns-provider-worker.ts", import.meta.url),
     );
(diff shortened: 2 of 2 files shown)


ss251 (approved):
Reviewed head d1e529ff550d7fde7ad29c4404672a1132b4d6bb.

The risky part is `packages/agent/src/runtime/eliza.ts:4272`, where the optimized-prompt HMAC bootstrap is now gated on `hasDurableHostVault()`. I followed the standalone-boot path: `startEliza` imports the app-core runtime on hosts without a durable vault, so `sharedVault()` returns the no-op bridge and unconditional `resolveOptimizedPromptIntegrityKey(vault)` tried to load/persist an integrity key that has nowhere to persist, breaking boot instead of falling back to baseline prompts. The added `&&` keeps the `ELIZA_OPTIMIZED_PROMPT_HMAC_KEY` env fast-path first and only skips vault-derived keygen when durable storage is absent, leaving managed hosts on the existing path.

No blocking concerns; a false-negative from `hasDurableHostVault()` would only hold a managed host on baseline prompts, non-blocking.

Verified by reading the diff at this head against the adjacent vault-bootstrap branch.

Recommendation: merge

— [muse-lane-ss251]

attentionhead (approved):
Reviewed at head `d1e529ff550d7fde7ad29c4404672a1132b4d6bb`. Both halves verified against negative controls. Approving.

## The Workerd test was genuinely broken, and this genuinely fixes it

Ran the suite on `origin/develop` first, so the "restore" claim isn't taken on faith:

```
service core:user:: Uncaught ReferenceError: process is not defined
  at worker.mjs:646:58255
 code: "ERR_RUNTIME_FAILURE"
 0 pass | 2 fail
```

At this head, under `bun test` (the file imports `bun:test`):

```
 1 pass | 0 fail   [11.57s]
```

The diagnosis embedded in the plugin comment is the right one — the bundle was following the workspace TypeScript alias for `@elizaos/core/edge` into core's unbundled source graph, which still references `process`.

**The `onResolve` target is the real artifact, not a guess.** `packages/core/package.json` maps `./edge` to `dist/edge/index.edge.js` for both `import` and `default` — exactly the path the plugin points at — and `packages/core/build.ts:1738` does support `--edge-only`. So the harness now bundles what actually ships.

**`nodejs_compat` doesn't weaken the boundary this test exists to police.** That was my main concern: a Workerd fidelity test that quietly grants itself a more permissive runtime than production stops being a fidelity test. It checks out — `packages/cloud/api/wrangler.toml` declares `compatibility_date = "2026-04-01"` and `compatibility_flags = ["nodejs_compat"]`, matching the harness exactly on both. Fidelity to the deployed Worker goes up here, not down.

## The standalone-boot guard is correct

I confirmed the failure it removes rather than inferring it. With the default no-op bridge:

```
hasDurableHostVault(): false
THREW: OPTIMIZED_PROMPT_INTEGRITY_KEY_INVALID
       optimized-prompt integrity key has an invalid encoded shape
```

That follows from `noopVault`: `setIfAbsent` returns `false` so the freshly generated key is never stored, then `readOptimizedPromptIntegrityKeyOrRecover` reads back `""` and fails the 32-byte assertion. Boot dies on a host that was never going to be able to persist the key. The guard is the right shape and is scoped to exactly that condition.

I also checked a mismatch that would have made the guard wrong: it queries `getAgentHostBridge().sharedVault()` while `resolveOptimizedPromptIntegrityKey` is handed the vault from `importAppCoreRuntime()`. Those are the same object — `importAppCoreRuntime` (eliza.ts:198) is a direct `return getAgentHostBridge()`. No divergence.

## Non-blocking: the fallback comment is slightly stronger than the code

"Standalone hosts without a durable vault use the baseline prompts" holds for a *fresh* standalone host, and that's the case that matters here: writes at `optimized-prompt.ts:721` need the key, so no artifacts can exist, and the discovery path (`listCompleteVersionNumbers`) wraps `verifyArtifactMac` in a `try`/`catch` that treats an unverifiable version as simply unavailable.

But `loadArtifactFromPath` calls `verifyArtifactMac` at line 1191 **outside** any `try`, and `resolveHmacKey()` throws `OPTIMIZED_PROMPT_INTEGRITY_KEY_UNAVAILABLE` at `severity: "fatal"` when the env var is unset. So a host that reaches that path with an artifact *and* its `.mac` already on disk — one that previously ran with a durable vault and later lost it, or an image that ships prebuilt artifacts — would get a fatal throw rather than the baseline fallback the comment promises.

I did not reproduce this, and it may well be unreachable in every configuration you ship; it's narrow either way and shouldn't hold up a CI restore. Flagging it only because the guard makes "key legitimately absent" a supported steady state for the first time, which is what turns that unguarded call into a live edge rather than a theoretical one.

## Not verified

- No standalone host booted end-to-end; the guard evidence is the bridge/vault path exercised directly, not a full `startEliza`.
- The four CI checks on this PR were still pending when I reviewed — my runs are local.
- I did not exercise the APNs provider against real Apple endpoints; the miniflare test's own assertions are the boundary I checked.

— [claude-code-eliza]

PR #30662, opened 2026-09-05
fix(cloud): cancel ambiguous billing leases after credential revocation

lalalune:
A billing Durable Object can commit a lease and lose its acknowledgement. If the credential is revoked before the identical retry, the retry returns 403. The client previously discarded the earlier ambiguity, so zero settlement left the committed charge and active lease behind even though no provider request ran.

Preserve possible commitment across subsequent definitive denials so zero settlement uses the stable cancellation capability. Reject combined admission requests without a nonempty cancellation token before committing; those requests must not fall through to legacy lease behavior.

Refs #30635. Follow-up to #30641.

Validation at `899117c195fff4b537d88ca054a16b7b33d0a4b1`, based on develop `9e738793d1584`: pinned installation, all 376 root verification tasks and final audits passed. The admission gate suite passed 63 cases; app admission/recovery passed 24 cases; APNs and logger cases passed. All 595 isolated Cloud API test files also passed. API typecheck, 709-route contract validation, and production Worker bundle dry-run passed. The real SQLite Durable Object harness runs the production gate and client, injects lost transport, malformed acknowledgement, and 503 acknowledgement failures, then revokes the credential before retry. All three cases invoke no provider, restore the balance, remove the active lease, and clear its alarm. The successful control uses one billing request. External billing persistence and provider calls are explicitly outside this harness.

## Evidence Gate

N/A - backend billing admission; no rendered surface changes.

N/A - backend billing admission; no rendered surface changes.

N/A - the affected boundary is a SQLite Durable Object lease, exercised in the linked fault-injection result.

[Before regression, passing gate and recovery tests, API validation, and full root verification](https://github.com/elizaOS/eliza/releases/download/pr-evidence-10/30662-lease-cancellation-validation-899117c1.log).
[Full isolated Cloud API suite: 595 files passed](https://github.com/elizaOS/eliza/releases/download/pr-evidence-10/30662-full-api-595-files-899117c1.log).

N/A - no browser or renderer changes.

N/A - all three fault cases must stop before invoking a provider; no model prompt or behavior changes.

[Real SQLite Durable Object state, calls, balance, lease, and alarm results](https://github.com/elizaOS/eliza/releases/download/pr-evidence-10/30662-lease-cancellation-workerd-899117c1.json). [Exact-head gate receipt](https://github.com/elizaOS/eliza/releases/download/pr-evidence-10/30662-lease-cancellation-gates-899117c1.json).

### DIFF
--- a/packages/cloud/api/__tests__/inference-admission-gate.test.ts
+++ b/packages/cloud/api/__tests__/inference-admission-gate.test.ts
@@ -2012,6 +2012,36 @@ describe("InferenceAdmissionGate", () => {
     expect(recoverExpiredLease).not.toHaveBeenCalled();
   });
 
+  test.each(["/lease-dispatched", "/lease-dispatched-authorized"])(
+    "%s refuses a missing cancellation capability without reserving money",
+    async (path) => {
+      const storage = new TestStorage();
+      const gate = createGate(storage);
+      await hydrateGate(gate, 10);
+      const response = await post(gate, path, {
+        organizationId: "org-a",
+        requestId: "missing-combined-capability",
+        balanceUsd: 10,
+        balanceRevision: "1",
+        estimatedCostUsd: 4,
+        recovery: organizationRecovery("missing-combined-capability"),
+        credential: {
+          kind: "api_key",
+          credentialId: "key-a",
+          userId: "user-a",
+        },
+      });
+      expect(response.status).toBe(400);
+      expect(
+        storage.read(storedLeaseKey("missing-combined-capability")),
+      ).toBeUndefined();
+      expect(storage.read("ledger")).toMatchObject({
+        availableUsd: 10,
+        activeLeaseCount: 0,
+      });
+    },
+  );
+
   test("prepared admission performs exactly one combined gate call before provider invocation", async () => {
     const storage = new TestStorage();
     const gate = createGate(storage);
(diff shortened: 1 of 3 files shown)


OutstandingRetard (approved):
Scoped review of #30662 (fix(cloud): cancel ambiguous billing leases after credential revocation).
Touches 3 file(s): `packages/cloud/api/__tests__/inference-admission-gate.test.ts`, `packages/cloud/api/src/inference-admission-gate.ts`, `packages/cloud/shared/src/lib/services/inference-admission-gate.ts`. Lead change `packages/cloud/api/__tests__/inference-admission-gate.test.ts` (+103/-0).
No blocking concerns.

ss251 (approved):
Reviewed head 899117c195fff4b537d88ca054a16b7b33d0a4b1.

The risky part is `packages/cloud/shared/src/lib/services/inference-admission-gate.ts:751` where `mayHaveCommitted` seeds from `prepared.state==="ambiguous"` and latches true on transport throw (:771), failed combined status (:805), 5xx (:816) and unreadable 2xx body (:829), guarding the `prepared.state="rejected"` writes at :775, :789 and :806.

I traced the injected revocation path: first `/lease-dispatched-authorized` commits 200, harness revokes key-a, then faults throw/malformed-200/503; retry gets 403 with flag true, so it throws `InferenceCredentialRevokedError` without clearing ambiguity, letting `settle` cancel. Assertions check lease undefined, ledger 10/0 and no alarm.

Verified by reading the diff at this head and following both retry branches.

Recommendation: merge

— [muse-lane-ss251]

ss251 (approved):
899117c195fff4b537d88ca054a16b7b33d0a4b1

Static review of the diff at this head for cancel-ambiguous-leases after credential revocation.

Findings: the change fixes two real billing-correctness paths. Input `POST /lease-dispatched` with `preProviderCancellationToken` missing/blank -> old output proceeded to `this.lease()` and could reserve money without a cancellation capability; new output at `packages/cloud/api/src/inference-admission-gate.ts:1480` and `:1492` returns `400 Invalid inference admission dispatch capability` via `validTrimmedId` before any serialize path. The new `inference-admission-gate.test.ts:2015` test proves it: no stored lease key and ledger stays `{availableUsd:10, activeLeaseCount:0}`.

Second, input revoked credential after a committed-but-unacknowledged combined dispatch -> old output set `prepared.state="rejected"` on the later 403/402/4xx in `packages/cloud/shared/src/lib/services/inference-admission-gate.ts:785`, erasing ambiguity so `settleInferenceAdmissionLease` never cancelled the committed lease and funds stayed held. New output tracks `mayHaveCommitted` initialized from `prepared.state==="ambiguous"` at `:751`, set on transport throw at `:771`, on unexpected status at `:805`, on 5xx at `:815`, and on unreadable 2xx body at `:829`, and only marks rejected when `!mayHaveCommitted`. Replay with the same capability then surfaces `InferenceCredentialRevokedError` and settle cancels.

Tests at `:2140` cover transport/malformed/unavailable faults, asserting 2 combined calls, lease undefined, ledger restored, and no alarm — a test that would fail if the old unconditional-rejected logic returned. Scope is tight, no workflow/permission/lockfile changes, no binaries, symlinks, secrets, or telemetry expansion.

Verified by reading the diff at this head and tracing lease reserve/cancel paths and added tests; no suite applies without a sandbox.

Recommendation: merge

— [muse-lane-ss251]

attentionhead (changes requested):
Reviewed at head `899117c195fff4b537d88ca054a16b7b33d0a4b1`.

The `mayHaveCommitted` logic is correct and well covered — that half I'd merge as-is. Requesting changes on one line in a test fixture, because **one of the two new money-path guards has no binding coverage**: it can be deleted entirely and the suite stays green.

## The core fix is solid

I audited every exit of the retry loop: the fetch throw, the combined-dispatch failure, the general 5xx branch, and the unreadable-2xx-body branch all set `mayHaveCommitted` before `continue`/`break`, and the three definitive-rejection sites are the guarded ones. `mayHaveCommitted` is also initialised from `prepared.state` *before* the optimistic `= "ambiguous"` assignment, so it correctly distinguishes a prior invocation's ambiguity from this one's.

Reverting all three guards to unconditional assignment fails exactly the three new tests (60 pass / 3 fail — transport, malformed, unavailable). Suite at this head: **63 pass** under `bun test` (this file imports `bun:test`, not `vitest`).

## Blocking: `/lease-dispatched-authorized` guard is never exercised

Removing **only** the guard in the `/lease-dispatched-authorized` branch leaves all **63 tests passing**. Its case in your `test.each` is satisfied by a different rejection entirely.

The cause is in the fixture. `authorizedLease` rejects first:

```ts
if (!request.credential ||
    request.credential.organizationId !== request.organizationId) {
  return jsonError("Invalid authorized inference admission lease", 400);
}
```

and the new test supplies `credential: { kind, credentialId, userId }` with **no `organizationId`**. So `undefined !== "org-a"` returns 400 before the capability check is ever reached. The assertions — status 400, no stored lease, ledger untouched — hold identically for both causes, so nothing distinguishes them.

**The guard itself is necessary, not redundant** — I checked, because "delete it" would be the wrong conclusion. `lease()` only validates the token when it is present:

```ts
(preProviderCancellationToken !== undefined && !validTrimmedId(preProviderCancellationToken))
```

A *missing* token passes that check, which is precisely the hole you're closing: a committed lease with no cancellation capability is uncancellable. That applies to the authorized path too, once a well-formed credential is supplied.

### Verified fix (one line)

Add the org to the fixture's credential:

```ts
credential: {
  kind: "api_key",
  credentialId: "key-a",
  userId: "user-a",
  organizationId: "org-a",
},
```

With that change:
- guards intact → **63 pass**
- authorized guard removed → **`/lease-dispatched-authorized refuses a missing cancellation capability without reserving money` fails**

The coverage becomes real, and the `test.each` then tests both paths as its name promises.

I'm blocking on this rather than filing it as a note for consistency with the standard I applied to #28977 in this same series: a guard on a money path that no test can catch regressing is the specific thing worth stopping for. It's one field, and everything else here is ready.

## Not verified

- No live Durable Object or Workers runtime; all evidence is the in-repo `TestStorage` harness the suite already uses.
- I did not audit whether other `/lease-*` routes accept an absent `preProviderCancellationToken` — I checked the two this PR touches plus `lease()`'s shared validation.
- The four CI checks were still pending when I reviewed; my runs are local.

— [claude-code-eliza]

Issue #30671, opened 2026-09-05
Measure and reduce cross-region inference admission round trips


lalalune:
Paired staging evidence identifies a remaining latency concentration: run33983112711 attempt2 at9e738793, SJC warm p50 auth cache4ms, rate-only admission94.5ms, atomic billing+dispatch95ms, direct Cerebras120.93ms and gateway327.53ms. Prior IAD run33949516711 warm DO phases totaled72ms versus189.5ms in SJC. Sources/SHAs differ, so this is not yet a controlled before/after code comparison. Existing rate-only object identities deliberately isolate quota progress from billing storage stalls; preserve that and durable credential/balance authority. Proposed experiment: rate-only DO commits idempotent rate decision, releases rate serialization, then delegates to billing DO and returns the combined result. Before rollout measure actual rate-to-billing latency and same-SHA multi-region controls; prove stalled billing does not block later rate decisions, 429 skips billing, billing denial never calls provider, ambiguous retry does not double count or double lease, and legacy endpoints remain compatible. Do not migrate existing object identities or apply blanket Worker placement. Existing Cloudflare objects cannot be relocated with later hints. Acceptance is a real canary with lower paired overhead and unchanged authorization/billing/concurrency semantics. Evidence: https://github.com/elizaOS/eliza/actions/runs/33983112711/artifacts/9974924143 .

lalalune:
Same-SHA geographic replication obtained: run33983112711 attempt3 again served9e738793 but used ORD instead of SJC. All44 paired calls succeeded. ORD overall headers p50 direct80.95ms/gateway205.78ms; warm direct80.95/gateway204.60; warm paired overhead114.27ms. Warm phase p50: KV2ms, rate58ms, atomic billing49ms, reserve0ms, upstream72ms. Compare same-SHA SJC warm phases4/94.5/95/0/117.5ms. Thus geography is a material contributor without relying on the old different-SHA IAD comparison. Auth certification still fails on the known undeployed telemetry fix (#30668). Artifact: https://github.com/elizaOS/eliza/actions/runs/33983112711/artifacts/9975296817 . Next missing evidence remains actual rate-to-billing subrequest timing and delayed-billing concurrency proof before transport redesign.

lalalune:
[...]

Issue #30668, opened 2026-09-05
Preserve inference auth timings through cache-miss hydration


lalalune:
Live staging certification run 33983112711 attempt 2 at 9e738793d1584882f3fa0765216c9e1d42c559e1 completed all 44 paired requests, then failed its auth probe with Missing required authorized_origin timing. The cacheOnly hydration continuation in inference-auth-context.ts sets authorized_origin on the outer trace without copying the child authoritative key/user/moderation timings or deferred cache-write status. The inner and outer resolution can also emit duplicate InferenceAuth records for one trace ID. This prevents strict live proof of the one-read standing contract. Acceptance: preserve child authoritative timings and async-write state on the response; emit one canonical auth trace per request; do not add cache reads or weaken probe assertions; cover cache-miss/controlled-probe/cache-hit and failure paths; rerun deployed auth certification.

Issue #30663, opened 2026-09-05
Avoid full-history fetch for ordinary deployment freshness checks


lalalune:
Cloud staging release 33982816539 spent several minutes in each deployment freshness check. The CLI currently unshallows all origin branches before checking whether the candidate or served commit is older, even for ordinary same-branch releases whose ancestry can be proven from a bounded history.

Optimize the guard to positively prove ancestry in both directions, progressively fetch only the candidate and served histories when needed, and retain full-history fallback when neither direction can be established. A failed ancestry probe in a shallow clone must remain indeterminate.

Acceptance: real temporary Git repositories prove stale and fresh releases without unshallowing, divergent and deep histories retain correct fallback behavior, missing commits preserve the existing indeterminate policy, and existing freshness tests pass. This affects deployment duration, not inference latency.

Issue #30375, opened 2026-09-03
[staging][QA] Deployed renderer certification stalls before Personal identity


standujar:
Parent epic: #29918

## Reproduced failure

The exact-SHA deployed staging renderer certification stalls before Personal Eliza resolution. In [Cloud CF Deploy run 33646345712](https://github.com/elizaOS/eliza/actions/runs/33646345712), API, Pages deployment, routing, and public pre/postflight checks all passed at served SHA `bc5ced13a279722befaa4ab536d338ad8af8c8fb`; only the [deployed renderer trajectory](https://github.com/elizaOS/eliza/actions/runs/33646345712/job/100308358594) failed.

The privacy-safe diagnostic records:

- phase `personal-identity`, elapsed 426,820 ms;
- runtime Cloud action: 1 attempt, 0 successes, 0 action timeouts;
- no runtime recovery or Personal identity retry visible;
- zero Personal identity GET requests/responses/failures/pending;
- zero Dedicated quote/activation/cutover/adoption activity and zero consent actions.

## Current exact-SHA reproduction

The same product failure reproduces read-only on the current served renderer:

- `origin/develop`, the Pages alias manifest, and `staging.eliza.app/eliza-renderer-build.json` all resolve to `cb721535b6fec5a67323cff0a2d5abc3697441df`, build `96d88462e69df6c34dc960a8ece6bd04ad146b380cf55cb1c008b8f7ccd57bad`;
- a fresh 1440x900 browser on `https://develop.eliza-app.pages.dev/`, with blank active-server/completion state and a non-secret syntactically usable placeholder Steward token, renders `Pairing Required`;
- the only relevant calls are `GET /api/first-run/status -> 401` and `GET /api/auth/me -> 401`;
- there is no `GET /api/v1/eliza/personal`, no WebSocket, and no non-GET/HEAD/OPTIONS request.

This reproduces outside the certification harness, so a harness-only change is not an acceptable fix.

## Root cause

The hosted non-dev web build resolves `branding.cloudOnly=true`, and the `eliza-app` Pages project is a first-party Cloud-hosted surface. Two startup boundaries nevertheless recognize only canonical/legacy Cloud hostnames:

1. `shouldProbeExistingLocalInstall` skips the local-install probe only for recognized Cloud control-plane hosts. The Pages deployment alias is unrecognized, so the cloud-only renderer probes same-origin `/api/first-run/status`.
[...]

PR #30672, opened 2026-09-05
fix(cloud): preserve latency certification telemetry and deadlines

lalalune:
## Summary

Make the deployed latency certification preserve the runtime evidence it is intended to verify.

- Preserve authoritative auth timings and deferred projection state through cache-miss hydration. Emit one canonical auth record per request and correlate shared projection completion to each consumer without another cache read or physical write. Keep probe validation strict and name missing timing fields in failures.
- Give the provision-capable browser identity phase 15m30s, matching the worker's 15-minute cold-provision ceiling plus commit margin. Keep the fresh-context continuity join at 2 minutes; no product/client timeout changes.
- Resolve deployment ancestry with positive bidirectional shallow-history proofs and targeted deepening before the existing full-history fallback. Do not interpret shallow negative ancestry as a conclusive result.

Refs #30668, #30375, #30663. Remaining cross-region admission work is tracked separately in #30671.

## Live reproduction

- Staging run [33983112711 attempt 2](https://github.com/elizaOS/eliza/actions/runs/33983112711/artifacts/9974924143) completed all 44 paired requests, then rejected incomplete `authorized_origin` timing evidence.
- [Attempt 3](https://github.com/elizaOS/eliza/actions/runs/33983112711/artifacts/9975296817) reproduced the same auth failure on the same deployed revision. All 44 paired calls again succeeded. ORD warm median direct/gateway headers were 80.95/204.60 ms; SJC was 120.93/327.53 ms. This is geographic evidence, not a post-fix performance claim.
- Deployed renderer runs [33982816539](https://github.com/elizaOS/eliza/actions/runs/33982816539) and [33983051052](https://github.com/elizaOS/eliza/actions/runs/33983051052) reached accepted Dedicated adoption but hit the approximately 390-second identity deadline while the job remained `in_progress`. These results do not prove terminal provisioning failure; a protected read-only job diagnostic remains required.

## Verification

Validated base: `1755a9423a226784331a85338e554ec880e82891`.
PR head: `dea404c0d7c6a50c4e332bc875f4b1240a310c65`.

- Bun 1.3.14 / Node 24.15.0; `bun install`: passed after integration/rebase.
- Auth resolver: 53 passing; HTTP telemetry: 21 passing; probe/certification: 22 passing; shared typecheck: passed.
- Real temporary-Git freshness suite: 25 passing, including both shallow directions, divergent/deep fallback, and missing commits.
- Browser trajectory diagnostics: 21 passing; deployed Playwright spec discovery/load passed.
- Scoped Biome and diff checks passed.
- Independent combined-diff review found no blocker in trace ownership, shared projection, or shallow-history safety.
- Full `bun run verify`: passed (exit 0), including all typecheck/lint tasks and 11,432-test focused/skip audit.
- Required `bun run --cwd packages/app audit:app`: passed (219 tests; zero broken/needs-work findings). OCR: 216 views, 204 verified, zero broken, 12 needs-eyeball. Desktop/mobile chat captures inspected; no rendered UI source changed.

## Evidence applicability

- [x] N/A - no rendered UI source changed; browser test-harness timing only.

- [x] N/A - no rendered UI source changed; required repository app audit passed.

- [x] N/A - no rendered UI or interactive behavior changed.

- [x] Live pre-fix failure and paired-call trace evidence: [sanitized certification artifact](https://github.com/elizaOS/eliza/actions/runs/33983112711/artifacts/9975296817). Post-deploy corrected trace proof remains required.

```text
deployment commit: 9e738793d1584882f3fa0765216c9e1d42c559e1; gateway/direct pairs: 22; all 44 calls ok=true.
ORD warm gateway phase medians: auth_cache_read=2ms, gateway_middle=58ms, gateway_reserve=0ms, gateway_setup=49ms.
[cloud-latency-certification] Inference auth probe failed with exit 1: [inference-auth-latency] Missing required authorized_origin timing
```

- [x] Real deployed browser network progression and identity deadline: [privacy-safe browser diagnostics](https://github.com/elizaOS/eliza/actions/runs/33982816539/artifacts/9974712905). Post-deploy browser proof remains required.

```json
{"phase":"personal-identity","elapsedMs":399171}
{"dedicatedAdoptionConfirmationResponseStatus":202,"dedicatedAdoptionConfirmationElapsedMs":1105}
{"dedicatedProvisionJobGetResponseCount":76,"dedicatedProvisionJobResponseStatus":200,"dedicatedProvisionJobStatus":"in_progress"}
```

- [x] N/A - no prompt, model, provider dispatch, or agent behavior changes; paired live provider calls are linked as diagnostic context above.

- [x] Real temporary-Git fixtures exercise shallow ancestry, fallback, and missing commits; resolver tests exercise deferred shared projection and response timing.

```text
Freshness tests: 25 passed, 0 failed; both shallow directions retained shallow history.
Auth resolver: 53 passed; HTTP telemetry: 21 passed; probe/certification: 22 passed.
Full bun run verify exited 0; 11432 test files scanned with no focused tests or orphaned skips.
```

- [x] N/A - no rendered visual surface changed. Required app audit/OCR completed with zero broken views; 12 unchanged captures retain needs-eyeball classification.

## Release validation

After merge, pin the staging release, latency certification, and read-only Personal Dedicated preview to the same develop revision. Inspect the terminal job state, rerun strict auth certification, and exercise real browser chat/history. Do not treat this PR's local tests as deployed end-to-end or production proof.


ss251 (approved):
Reviewed head dea404c0d7c6a50c4e332bc875f4b1240a310c65.

The risky part is `packages/cloud/scripts/deploy-freshness-guard-cli.mjs:126` where `probeLinearAncestry` checks both directions before any fetch. I traced the boundary that usually breaks shallow guards: in a `--depth=1` clone `merge-base --is-ancestor` returns exit 1 for missing history, indistinguishable from divergence. The fix does not treat a single exit 1 as `false` — `false`+`false` falls through to `null` — only an exit 0 in either direction is conclusive (`run→served` true, `served→run` false). The escalation loop at `:139` then fetches `--depth=50/200/1000` for both SHAs and re-probes retained history before `hydrateHistoryForAncestry`, which matches the new tests asserting `is-shallow=true` for linear 300-commit cases and `false` only for diverged/1005-commit cases.

No blocking concerns; depth-ceiling naming is non-blocking.

Recommendation: merge

— [muse-lane-ss251]

attentionhead (approved):
Reviewed at head `dea404c0d7c6a50c4e332bc875f4b1240a310c65`. Approving — all three claims hold under mutation, and the concurrency contract is pinned rather than assumed.

## Ancestry resolution

Suite: **25 pass, 46 expect() calls**, matching your figure. Both new behaviours are load-bearing, tested one at a time:

- Deleting only the reverse-direction line from `probeLinearAncestry` → `(fail) proves a newer run through reverse ancestry while remaining shallow`.
- Forcing `deepenRelevantHistoryForAncestry` to return `null` immediately → `(fail) proves a stale run through targeted history while remaining shallow` **and** the reverse-ancestry test.

The asymmetry is the right one: a positive `merge-base --is-ancestor` is conclusive even in a shallow clone because the walk actually found the path, while a negative is only "not in retained objects." Checking both directions before spending a fetch is what makes the common same-branch and stale-run cases free. Identical SHAs fall out correctly — a commit is its own ancestor, so the first check answers.

One note, not a defect: `git fetch --depth=N` is absolute, not additive, so the first iteration at `50` can *shorten* history that a deeper checkout had retained. It's harmless here because `deepenRelevantHistoryForAncestry` only runs after `probeLinearAncestry` already returned `null` — if the answer had been in retained objects, the probe would have found it — and 200/1000 re-anchor from the fetched tips. Worth a comment if anyone later moves that call earlier.

## Auth telemetry

Suite: **53 pass, 209 expect() calls**, matching your figure. Removing the three-line `trace.timings` propagation kills 4 named tests, headed by `controlled cold hydration emits complete HTTP timings and one trace per concurrent caller`.

The part I went looking for a bug in and didn't find: `getOrCreateApiKeyHydration` is shared, so two concurrent callers read the *same* `authoritativeTelemetry()` and stamp one key lookup's timings into two traces. That test settles it deliberately — `authBoundaryCalls === [KEY]`, two cache reads, one cache write, and each `traceId` asserted to receive exactly two canonical records (its own `authorized_origin` trace plus the correlated `kind: "cache_write"`). Attributing shared work to each consumer is the contract, and re-stamping `traceId` in `observeHydrationProjection` is what keeps the correlation honest. Good test.

I also traced the `INFERENCE_AUTH_HYDRATION_TELEMETRY_MISSING` throw, since a telemetry gap turning a successful auth into a 500 would be a bad trade. It's unreachable: the nested resolve sets `authoritativeTelemetry` in its `finally`, which runs on both the return and the 401/403 throw path that `.catch` converts to a `rejected` decision; and the deadline arm of the `Promise.race` resolves `undefined`, so `continued` is falsy there. Defensive, correctly placed.

Splitting `[InferenceAuth] hydration cache write` / `hydration trace` out of the canonical `[InferenceAuth] trace` scope is the right call — the certification was reading hydration-internal records as request records.

## Timeout budget

The arithmetic is exact, not approximate: `930_000` (15m + 30s commit margin) + `120_000` = `1_050_000`, precisely half of the 2_100_000 test timeout, which is what the rewritten comment claims.

Flagging one thing about the shape of the guard: `expect(BUDGET).toBeLessThanOrEqual(TRAJECTORY - BUDGET)` reduces to `2 × BUDGET ≤ TRAJECTORY`, and it currently passes at exact equality. That's a ratchet with zero slack — any future bump to either identity constant fails it, which I read as intended, but it's worth knowing the test is sitting on its boundary rather than under it.

Trajectory diagnostics: **21 passed**, matching.

## One number I couldn't reproduce

You report "probe/certification: 22 passing". `packages/cloud/scripts/inference-auth-latency.test.mjs` is a `node:test` file and `node --test` gives me **14 pass, 0 fail** on it. I assume your 22 spans another file I didn't identify — flagging only so the count in the PR body isn't taken as this file's. The one-line probe change is properly covered: the suite asserts `Missing required authorized_origin timing: auth_key_lookup`, so the field name in the message is pinned, which is exactly what made the staging failure unreadable before.

## Not verified

- Nothing deployed. I did not re-run the staging certification, the Playwright lane, or `audit:app`; the linked artifacts are taken at face value.
- The 15m30s ceiling matching the worker's real cold-provision bound is a claim about infrastructure I can't observe — I checked only that the constant equals what the PR body says and that the arithmetic closes.
- I did not evaluate whether the remaining half of the trajectory budget (two 240s history proofs, 180s chat, two renderer boots) is comfortable at the new split.
- Cross-region admission work is out of scope per #30671.

— [claude-code-eliza]

PR #30674, opened 2026-09-05
fix(ci): restore staging browser gates and read-only diagnosis

lalalune:
# Relates to

Continues #30552: restore deploy validation and protected staging diagnosis.

# Sync with develop

Rebased on current `origin/develop` (`c884a2f8bd97b1ffee8542b1df828edb3b392af4`). Exact head `1e5ade1429a2e11788d3cc0237c3a77ab46a6741`: `bun install` passed, root `bun run verify` passed 376/376 tasks, all three recorded Chromium tests passed, and 54 operator/workflow/database-identity tests passed. Required app audit passed 219 tests with zero broken or needs-work views; affected desktop/mobile captures were inspected.

# Risks

Low. Two deterministic browser fixtures change, plus the admission check for a read-only diagnostic. Execution still requires the exact deployed/workflow commit, prior approval digest, reviewed reason, and confirmation. The diagnostic continues to verify the served staging commit and protected database identity.

# Background

## What does this PR do?

Restores the deferred Cloud model-gate fixture's required hardware response. Stubs the desktop sign-in prewarm in the walkthrough and waits for its response, preventing a pending production request from racing the final diagnostics assertion. Allows read-only staging preview after develop advances, while rejecting unknown modes and retaining all execution checks.

Bug fixes for release validation and operational diagnosis.

No product documentation changes. The workflow explains the distinction between preview and execution.

# Testing

The failing hosted browser jobs and the before/after boundary results below. The workflow test executes its real shell admission command, including rejected mutations and accepted read-only preview.

## Detailed testing steps

- Run the deferred-model and walkthrough Chromium specs together with recording.
- Run the database-identity, re-review operator, and workflow tests.
- Run `bun run verify` after dependency installation and sync.

# Evidence Gate

- [x] Before screenshots: [desktop failure](https://github.com/elizaOS/eliza/releases/download/pr-evidence-10/30674-pre-fix-failure-1.png), [mobile failure](https://github.com/elizaOS/eliza/releases/download/pr-evidence-10/30674-pre-fix-failure-2.png).

- [x] After screenshots: [desktop ready](https://github.com/elizaOS/eliza/releases/download/pr-evidence-10/30674-1e5-desktop-after.png), [mobile ready](https://github.com/elizaOS/eliza/releases/download/pr-evidence-10/30674-1e5-mobile-after.png).

- [x] Walkthrough video: [desktop model gate](https://github.com/elizaOS/eliza/releases/download/pr-evidence-10/30674-1e5-browser-flow-1.mp4), [mobile model gate](https://github.com/elizaOS/eliza/releases/download/pr-evidence-10/30674-1e5-browser-flow-2.mp4), [onboarding and chat](https://github.com/elizaOS/eliza/releases/download/pr-evidence-10/30674-1e5-browser-flow-3.mp4).

- [x] N/A - no backend implementation changes; browser requests use deterministic fixtures and the changed workflow admission runs as a shell subprocess.

- [x] Frontend logs: [browser execution](https://github.com/elizaOS/eliza/releases/download/pr-evidence-10/30674-1e5-browser.log), [trace network and console report](https://github.com/elizaOS/eliza/releases/download/pr-evidence-10/30674-1e5-frontend-network.json).

- [x] N/A - no agent, model, prompt, or provider behavior changes; these tests do not dispatch a live model.

- [x] Admission results: [54 operator contracts](https://github.com/elizaOS/eliza/releases/download/pr-evidence-10/30674-1e5-operator-contracts.log), [376 root verification tasks](https://github.com/elizaOS/eliza/releases/download/pr-evidence-10/30674-1e5-root-verify.log).

- [x] App audit/OCR: [219 passing tests and visual/OCR review](https://github.com/elizaOS/eliza/releases/download/pr-evidence-10/30674-app-audit.log).

# Evidence Details

## Real LLM-call trajectory

N/A - deterministic browser fixtures and read-only workflow admission. The separate live staging trajectory is tracked in #30552 and has not passed.

## Backend + frontend logs

Before: Develop Full job 101352261662 failed both deferred-model cases because the strict parser rejected their missing hardware response. Job 101352261641 failed the walkthrough on an unmocked production `/api/auth/cli-session` CORS error. A local pre-fix trace also captured that pending request even when timing allowed the test to pass.

After: all three browser tests pass, with zero console errors and only HTTP 200/201 or WebSocket 101 responses. The walkthrough trace records the awaited fixture response as HTTP 200.

The previous diagnostic guard rejects a valid older served-commit preview; the fixed guard accepts it while rejecting mismatched execution, non-develop refs, malformed commits, absent approval fields, and unknown modes.

### Before

Hosted failures: [desktop/mobile model gate](https://github.com/elizaOS/eliza/actions/runs/33982927203/job/101352261662) and [walkthrough production prewarm](https://github.com/elizaOS/eliza/actions/runs/33982927203/job/101352261641). Reproduced failure screenshots are attached above.

### After

Initial gate: [desktop](https://github.com/elizaOS/eliza/releases/download/pr-evidence-10/30674-1e5-desktop-before.png) and [mobile](https://github.com/elizaOS/eliza/releases/download/pr-evidence-10/30674-1e5-mobile-before.png). After Cloud registration without reload: [desktop](https://github.com/elizaOS/eliza/releases/download/pr-evidence-10/30674-1e5-desktop-after.png) and [mobile](https://github.com/elizaOS/eliza/releases/download/pr-evidence-10/30674-1e5-mobile-after.png). All four captures were inspected.

### Walkthrough video

The three exact-head MP4 recordings are linked in the evidence gate above.

## Audio / voice walkthrough

## Known gaps / failures

The live staging trajectory still times out while Dedicated provisioning remains in progress. This PR enables read-only diagnosis of that separate release failure; it does not claim production readiness. An earlier local verification overlapped a logger artifact restore, causing a transient module-resolution failure; the unchanged focused typecheck passed afterward. A concurrent cold dependency probe exceeded its default five-second budget; the unchanged 54-test suite passed alone in 1.85 seconds. One recording preparation stalled in a native Maps bundler before tests started; that owned process was stopped and the subsequent build/test passed.

N/A - no exception requested.

# Deploy Notes

## Database changes

None. Preview is read-only, and existing protected database identity checks remain required.

## Deployment instructions

After merge, dispatch `Personal Dedicated selection review` from develop in preview mode with the current served staging commit and the reviewed staging database identity receipts. Keep execute mode unchanged. Continue the deployment recovery in #30552.

## Discord username

N/A - GitHub coordination in #30552.

### DIFF
--- a/.github/workflows/personal-dedicated-rereview-staging.yml
+++ b/.github/workflows/personal-dedicated-rereview-staging.yml
@@ -76,7 +76,7 @@ jobs:
           install-native-deps: "false"
           run-postinstall: "false"
 
-      - name: Require exact develop deployment authority
+      - name: Require protected staging diagnostic authority
         env:
           EXPECTED_COMMIT: ${{ inputs.expected_cloud_commit }}
           CHECKED_OUT_COMMIT: ${{ github.sha }}
@@ -89,8 +89,12 @@ jobs:
           bun -e '
             const expected = process.env.EXPECTED_COMMIT ?? "";
             if (process.env.REF_NAME !== "refs/heads/develop") process.exit(1);
-            if (!/^[0-9a-f]{40}$/.test(expected) || expected !== process.env.CHECKED_OUT_COMMIT) process.exit(1);
+            if (!/^[0-9a-f]{40}$/.test(expected)) process.exit(1);
+            if (!["preview", "execute"].includes(process.env.MODE ?? "")) process.exit(1);
+            // Read-only preview diagnoses the served commit after develop advances.
+            // The operator still verifies its exact API and database identities.
             if (process.env.MODE === "execute") {
+              if (expected !== process.env.CHECKED_OUT_COMMIT) process.exit(1);
               if (!/^[0-9a-f]{64}$/.test(process.env.APPROVAL_DIGEST ?? "")) process.exit(1);
               const rereview = process.env.REVIEWED_REASON === "retain_current_receipt_target_after_duplicate_inventory_review" && process.env.CONFIRMATION === "REREVIEW_STALE_SELECTION_WITHOUT_COMPUTE_MUTATION";
               const bootstrap = process.env.REVIEWED_REASON === "select_unique_verified_backup_after_duplicate_inventory_review" && process.env.CONFIRMATION === "SELECT_UNIQUE_VERIFIED_BACKUP_WITHOUT_COMPUTE_MUTATION";
--- a/packages/app/test/ui-smoke/deferred-cloud-model-gate.spec.ts
+++ b/packages/app/test/ui-smoke/deferred-cloud-model-gate.spec.ts
@@ -12,6 +12,7 @@ import {
   installPageDiagnosticsGuard,
   openAppPath,
   seedAppStorage,
+  UI_SMOKE_CPU_ONLY_HARDWARE,
 } from "./helpers";
 
 test.describe("deferred Eliza Cloud model registration", () => {
@@ -82,6 +83,7 @@ test.describe("deferred Eliza Cloud model registration", () => {
           status: 200,
           contentType: "application/json",
           body: JSON.stringify({
+            hardware: UI_SMOKE_CPU_ONLY_HARDWARE,
             textReadiness: {
               slots: {
                 TEXT_SMALL: {
--- a/packages/app/test/ui-smoke/walkthrough/walkthrough-capture-smoke.spec.ts
+++ b/packages/app/test/ui-smoke/walkthrough/walkthrough-capture-smoke.spec.ts
@@ -258,10 +258,27 @@ test.describe("walkthrough capture smoke", () => {
     await seedAppStorage(page, { "eliza:first-run-complete": "" });
     await injectFullCapabilityHost(page);
     await installDefaultAppRoutes(page);
+    // Desktop prewarms sign-in before a click; capture must not contact Cloud.
+    await page.route("**/api/auth/cli-session", async (route) => {
+      if (route.request().method() !== "POST") {
+        await route.fallback();
+        return;
+      }
+      await fulfillJson(route, 200, {
+        sessionId: "11111111-1111-4111-8111-111111111111",
+      });
+    });
     const firstRun = await installMutableFirstRunStatus(page);
     await installWalkthroughConversationStore(page);
 
+    const desktopLoginWarmup = page.waitForResponse(
+      (response) =>
+        new URL(response.url()).pathname === "/api/auth/cli-session" &&
+        response.request().method() === "POST" &&
+        response.status() === 200,
+    );
     await page.goto("/", { waitUntil: "domcontentloaded" });
+    await desktopLoginWarmup;
     const onboarding = page.getByTestId("chat-overlay");
     await expect(onboarding).toBeVisible({ timeout: 20_000 });
     await expect(page.getByText(FIRST_RUN_SIGN_IN_PROMPT)).toBeVisible({
--- a/packages/cloud/scripts/admin/personal-dedicated-rereview-staging-workflow.test.ts
+++ b/packages/cloud/scripts/admin/personal-dedicated-rereview-staging-workflow.test.ts
@@ -50,26 +50,61 @@ describe("personal Dedicated staging re-review workflow", () => {
     });
   });
 
-  test("requires exact develop SHA, prior digest, reviewed reason, and confirmation", () => {
-    const guard = step("Require exact develop deployment authority");
-    expect(guard.run).toContain(
-      'process.env.REF_NAME !== "refs/heads/develop"',
-    );
-    expect(guard.run).toContain("expected !== process.env.CHECKED_OUT_COMMIT");
-    expect(guard.run).toContain("APPROVAL_DIGEST");
-    expect(guard.run).toContain(
-      "retain_current_receipt_target_after_duplicate_inventory_review",
-    );
-    expect(guard.run).toContain(
-      "REREVIEW_STALE_SELECTION_WITHOUT_COMPUTE_MUTATION",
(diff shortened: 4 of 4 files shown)


ss251 (approved):
Reviewed head 1e5ade1429a2e11788d3cc0237c3a77ab46a6741.

The risky part is `.github/workflows/personal-dedicated-rereview-staging.yml:92` where preview no longer requires `EXPECTED_COMMIT === CHECKED_OUT_COMMIT`. I followed both branches: `MODE=preview` with `EXPECTED_COMMIT=b repeated 40` against `CHECKED_OUT_COMMIT=a repeated 40` now exits 0, matching the new `preview of the currently served older commit` case, while the same mismatch under `MODE=execute` exits 1 via the nested equality check at :97 plus the 64-hex `APPROVAL_DIGEST` and the two exact `REVIEWED_REASON`/`CONFIRMATION` pairs. `REF_NAME` must still be `refs/heads/develop` and `MODE` in `preview,execute`, so main-branch or unknown-mode still fails.

No blocking concerns; preview drift is scoped as read-only diagnosis after develop advances. Verified by reading the diff at this head.

Recommendation: merge

— [muse-lane-ss251]

ss251 (approved):
Reviewed head 1e5ade1429a2e11788d3cc0237c3a77ab46a6741.

The risky part is `.github/workflows/personal-dedicated-rereview-staging.yml:92-98`, where preview mode no longer requires `EXPECTED_COMMIT === CHECKED_OUT_COMMIT`. I traced the new branch: preview still exits 1 unless `REF_NAME` is exactly `refs/heads/develop`, `expected` matches `^[0-9a-f]{40}$`, and `MODE` is in `preview|execute`; the equality check plus 64-hex `APPROVAL_DIGEST` plus the exact `REVIEWED_REASON`/`CONFIRMATION` pair match now gate only `execute`. The replacement test in `packages/cloud/scripts/admin/personal-dedicated-rereview-staging-workflow.test.ts:53` actually executes the guard via `Bun.spawnSync`, covering preview-of-older-commit exit 0 versus execution-against-older-commit exit 1, plus main-ref, malformed commit, unknown mode, and missing digest paths.

Verified by reading the diff at this head and following both MODE branches.
Recommendation: merge

— [muse-lane-ss251]

attentionhead (approved):
Reviewed at head `1e5ade1429a2e11788d3cc0237c3a77ab46a6741`. Approving.

The part worth scrutiny is the admission relaxation: moving `expected !== CHECKED_OUT_COMMIT` inside the `execute` branch lets a `preview` run when the operator's declared commit no longer equals the checked-out develop HEAD. That is a loosened gate on a workflow holding `DATABASE_URL` and `ELIZAOS_CLOUD_API_KEY`, so I chased what still holds it closed rather than taking "read-only" on the label.

Both properties survive:

- **Preview cannot mutate.** In `personal-dedicated-rereview-staging.ts`, `dependencies.execute(...)` is reachable only inside `if (config.mode === "execute")`, and `mode` is independently re-validated to `preview | execute` at line 293. The shell guard is not the only thing standing between a preview and a write.
- **The served commit is still verified.** `dependencies.verifyDeployment(config.expectedCloudCommit)` runs at line 374, ahead of both branches. So what the change drops is repo-HEAD equality, not the check that staging actually serves the commit being diagnosed — which is precisely the distinction that makes diagnosing an older deployment coherent.

## The guards are executable now, and that matters more than it looks

Both new clauses are load-bearing, tested one at a time:

- Deleting the `["preview","execute"].includes(MODE)` line → `(fail) enforces shell admission for unknown operation`.
- Deleting the execute-branch `expected !== CHECKED_OUT_COMMIT` line → `(fail) enforces shell admission for execution against an older commit`.

The more interesting result is what the *old* test could not do. I changed `if (process.env.MODE === "execute")` to `if (false && process.env.MODE === "execute")`, leaving every guard string literally present in the YAML. The deleted test asserted only `expect(guard.run).toContain(...)` on those strings, so it would have passed against an entirely dead admission gate. The new table fails four ways:

```
(fail) execution against an older commit
(fail) execution without a prior digest
(fail) execution without confirmation
(fail) execution without a reviewed reason
```

Replacing a source-text assertion with `Bun.spawnSync` against the real `guard.run` is a strict upgrade, and it's the difference between pinning that a string is present and pinning that the gate works. Good change to make in the same PR that loosens the gate.

Suite at head: **14 pass, 28 expect() calls**.

## Fixtures

`UI_SMOKE_CPU_ONLY_HARDWARE` is a real export (`helpers.ts:280`) and `fulfillJson` is already local to the walkthrough spec, so neither addition is a dangling reference.

The `cli-session` route is registered *after* `installDefaultAppRoutes`, which is correct — Playwright matches in reverse registration order, so it takes precedence — and the non-POST `route.fallback()` correctly delegates rather than swallowing. `installDefaultAppRoutes` doesn't handle that path itself, so this is filling a real hole, not shadowing. The `waitForResponse` is created before `goto` and awaited after, which is the ordering that actually closes the race rather than narrowing it.

## Two small things, neither blocking

- The table covers `MODE: "unknown"` but not `MODE: ""`. `process.env.MODE ?? ""` handles it correctly (empty isn't in the allowlist), but empty is the value a misconfigured dispatch actually produces, and it's one more row.
- `await desktopLoginWarmup` makes the walkthrough hard-depend on the prewarm firing. If that request is ever removed the spec times out at the Playwright default rather than failing with a useful message. Worth a short explicit timeout on that wait.

## Not verified

- I did not run either Chromium spec — no browser lane locally — so the two fixture repairs are reviewed by reading, not by execution. The claim that they fix the hosted job failures is taken from the linked runs.
- I did not run `bun run verify`, the app audit, or the operator/database-identity suites.
- I did not exercise the workflow itself against staging; the admission analysis is of `guard.run` and the script's branch structure.

— [claude-code-eliza]

Issue #30692, opened 2026-09-05
fix(android): provision and route fused local embeddings through the bionic host


lalalune:
A fresh Android local-runtime install cannot compute local embeddings. Device testing on the canonical sideload APK found no `gte-small_fp16.gguf` under the runtime models directory. `BionicHostLoader` also has no `embed` method, so `ensureLocalInferenceHandler` selects the desktop fused FFI handler; the musl agent cannot load Android bionic native libraries through that path. The Java inference server has an embedding operation, but currently uses the resident chat context and last-token pooling.

Acceptance: normal Android build/setup packages or synchronizes the pinned, hash-verified embedding artifact; the Android loader routes embeddings through the bionic host with a dedicated compatible embedding context and correct pooling; real device reads return finite nonzero vectors with stable dimensions, preserve the text-generation context, and fail explicitly if the model/native path is unavailable. Include actual artifact hashes, native input/output, repeated calls, and a text turn before/after embedding. Desktop setup verification is tracked in #30673.

PR #30698, opened 2026-09-05
fix(setup): verify complete fused libraries and live embeddings on install

### LINKED ISSUE #30673: fix(inference): verify complete fused library readiness during repository setup
A normal `bun install` already initializes the pinned fused inference source and stages the host library and embedding model. Its freshness gate hashes only the main library, however: a missing or changed GGML/llama companion can leave `--check` successful and cause the runtime loader (including local embeddings) to fail. The stamp also does not qualify host platform/architecture, and its dirty-source fingerprint hashes status filenames rather than changed bytes.

Acceptance: install/ensure must reject incomplete or altered library sets, changes to native source bytes, and mismatched host targets; rebuild and stage a complete matching set; and demonstrate an actual local embedding through the installed library. Existing explicit setup skips remain explicit, not the normal development path. Review clone/install and subsequent startup reuse so library readiness is not inferred from file presence alone.

lalalune:
Normal `bun install` now verifies the complete fused inference library set and computes a local embedding before reporting setup as ready. A missing GGML/llama companion previously passed the freshness check while the real loader failed. Reuse now checks all staged library hashes, the host architecture, and native source bytes, including repeated edits to already-dirty files. An explicit backend change invalidates reuse.

The installer and native stage command share state-directory resolution, so relative and home-relative paths match the runtime. Native source must be initialized, and the existing pinned submodule and hash-verified embedding-model setup remain automatic. An explicit setup skip does not establish readiness.

Closes #30673.

Validation: 19 focused setup tests pass. A real Metal build produced the complete library set. Actual native embeddings returned three finite, nonzero 384-dimensional vectors; related sentences had cosine similarity 0.8954 versus 0.6785 for unrelated sentences. Removing `libllama.0.dylib` from a copied otherwise-current set makes the new check fail, while the old check reported success and the actual loader failed. Normal install, 19 focused setup tests, and the full root verification gate pass on 69ed4e1a20bf5e50fe58e4d651fa3b1ce76a9647. Hosted checks must pass before merge. Native execution evidence is macOS arm64; Android packaging/routing defects found during separate device testing are tracked in #30692.

Additional regression readbacks: [Old false freshness result](https://github.com/elizaOS/eliza/releases/download/pr-evidence-10/30698-30673-missing-companion-before-check.log); [New complete-set success](https://github.com/elizaOS/eliza/releases/download/pr-evidence-10/30698-30673-complete-set-fresh.log).

## Evidence Gate

N/A — repository setup and native loader changes; no rendered product surface changes.

N/A — repository setup and native loader changes; complete CLI and native artifacts are provided.

N/A — no interactive UI flow changes; before/after native loader and embedding execution is recorded in logs.

[Normal install and actual embedding probe](https://github.com/elizaOS/eliza/releases/download/pr-evidence-10/30698-30673-install-69ed4e1a.log); [376 tasks and final audits](https://github.com/elizaOS/eliza/releases/download/pr-evidence-10/30698-30673-verify-69ed4e1a.log).

N/A — no frontend changes or browser requests.

N/A — no generative agent or LLM behavior changes. Real embedding model inputs and outputs are supplied as native domain artifacts below.

[Final source hashes and proof provenance](https://github.com/elizaOS/eliza/releases/download/pr-evidence-10/30698-30673-review-receipt-69ed4e1a.json); [Real Metal build](https://github.com/elizaOS/eliza/releases/download/pr-evidence-10/30698-30673-native-ensure-first.log); [Native library inventory](https://github.com/elizaOS/eliza/releases/download/pr-evidence-10/30698-30673-companion-regression.json); [Actual old loader failure](https://github.com/elizaOS/eliza/releases/download/pr-evidence-10/30698-30673-missing-companion-before-load.log); [New missing-companion rejection](https://github.com/elizaOS/eliza/releases/download/pr-evidence-10/30698-30673-complete-set-missing.log).
[19 setup tests](https://github.com/elizaOS/eliza/releases/download/pr-evidence-10/30698-30673-setup-tests-69ed4e1a.log); [Native embedding inputs and complete vectors before rebuild](https://github.com/elizaOS/eliza/releases/download/pr-evidence-10/30698-30673-live-embedding-before.log); [Native embedding inputs and complete vectors after rebuild](https://github.com/elizaOS/eliza/releases/download/pr-evidence-10/30698-30673-live-embedding-after.log). Native before/after executions predate the final rebase; the final install log records the current product embedding probe.

N/A — no screenshots or rendered text changed.

### DIFF
--- a/packages/app-core/README.md
+++ b/packages/app-core/README.md
@@ -40,3 +40,19 @@ bun run --cwd packages/app-core lint         # Biome
 ```
 
 This package is consumed by `@elizaos/agent`, `@elizaos/ui`, `@elizaos/shared`, the `packages/app` shell, and most `plugins/*` app plugins. It targets Node `>=24`, with `react`/`react-dom`/`three` as peer dependencies and the `@elizaos/capacitor-*` mobile bridges as optional dependencies.
+
+## Native inference setup
+
+A normal root `bun install` initializes the pinned fused inference submodule,
+ensures the host library and its companion libraries are current, and provisions
+the hash-verified default embedding model. The runtime discovers these artifacts
+under the same state directory without additional environment configuration.
+Relative `ELIZA_STATE_DIR` paths resolve from the current working directory;
+`~` expands to the user's home directory.
+
+Reuse checks cover every staged native library, host architecture, and native
+source changes. Missing or altered companions trigger a rebuild on install;
+setup failures fail the install instead of reporting inference as ready.
+`ELIZA_SKIP_FUSED_INFERENCE_SETUP=1` is an explicit escape hatch and does not
+establish runtime readiness. Android and iOS packaging use their platform build
+lanes to produce the corresponding NDK or Apple artifacts.
--- a/packages/app-core/README.md
+++ b/packages/app-core/README.md
@@ -45,7 +45,8 @@ This package is consumed by `@elizaos/agent`, `@elizaos/ui`, `@elizaos/shared`, 
 
 A normal root `bun install` initializes the pinned fused inference submodule,
 ensures the host library and its companion libraries are current, and provisions
-the hash-verified default embedding model. The runtime discovers these artifacts
+the hash-verified default embedding model. Setup then loads the native library
+and computes a local embedding before reporting readiness. The runtime discovers these artifacts
 under the same state directory without additional environment configuration.
 Relative `ELIZA_STATE_DIR` paths resolve from the current working directory;
 `~` expands to the user's home directory.
--- a/packages/app-core/scripts/ensure-fused-inference-install.mjs
+++ b/packages/app-core/scripts/ensure-fused-inference-install.mjs
@@ -16,9 +16,9 @@ import {
   rmSync,
   writeFileSync,
 } from "node:fs";
-import os from "node:os";
 import path from "node:path";
 import { fileURLToPath } from "node:url";
+import { resolveSetupStateDir } from "./lib/setup-state-dir.mjs";
 
 const scriptDir = path.dirname(fileURLToPath(import.meta.url));
 const defaultRepoRoot = path.resolve(scriptDir, "..", "..", "..");
@@ -61,23 +61,6 @@ function runChecked(command, args, options = {}) {
   }
 }
 
-function resolveStateDir(env) {
-  const explicit = env.ELIZA_STATE_DIR?.trim();
-  if (explicit) {
-    return path.isAbsolute(explicit)
-      ? explicit
-      : path.join(os.homedir(), explicit);
-  }
-  const namespace = env.ELIZA_NAMESPACE?.trim() || "eliza";
-  const xdg = env.XDG_STATE_HOME?.trim();
-  if (xdg) {
-    return path.isAbsolute(xdg)
-      ? path.join(xdg, namespace)
-      : path.join(os.homedir(), xdg, namespace);
-  }
-  return path.join(os.homedir(), ".local", "state", namespace);
-}
-
 export function resolveEmbeddingArtifactPath({
   env = process.env,
   repoRoot = defaultRepoRoot,
@@ -85,7 +68,7 @@ export function resolveEmbeddingArtifactPath({
   const configured = env.MODELS_DIR?.trim();
   const modelsDir = configured
     ? path.resolve(repoRoot, configured)
-    : path.join(resolveStateDir(env), "models");
+    : path.join(resolveSetupStateDir(env), "models");
   return path.join(modelsDir, FUSED_EMBEDDING_ARTIFACT.filename);
 }
 
--- a/packages/app-core/scripts/ensure-fused-inference-install.mjs
+++ b/packages/app-core/scripts/ensure-fused-inference-install.mjs
@@ -250,6 +250,18 @@ export async function ensureFusedInferenceInstall({
   );
   run(bunExecutable, [stageScript, "--ensure"], { cwd: repoRoot, env });
   const embedding = await ensureEmbedding({ env, repoRoot, fetchImpl });
+  run(
+    bunExecutable,
+    [
+      "--conditions=eliza-source",
+      path.join(
+        repoRoot,
+        "packages/app-core/scripts/verify-fused-embedding.mjs",
+      ),
+      embedding.path,
+    ],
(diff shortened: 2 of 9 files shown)


ss251 (approved):
Reviewed head 69ed4e1a20bf5e50fe58e4d651fa3b1ce76a9647.

The risky part is `packages/app-core/scripts/lib/fused-artifact-integrity.mjs:58`, where reuse compares the union of stamped vs on-disk native libraries. I traced the three cases that would silently ship broken inference: deleted `libggml.so.0` leaves `actual[name]` undefined so `actual!==expected` pushes companion-missing; rewritten bytes change the sha256 inventory so the same branch fires; extra `libggml-cuda.so` appears only in actual but the `new Set([...expected,...actual])` still visits it. The guard at :38 rejects legacy stamps without `libraries` before I/O, and inventory read failure returns stale fail-closed.

Verified by reading the diff at this head against its fixture test; live native build not exercised.
Recommendation: merge

— [muse-lane-ss251]

attentionhead (changes requested):
Reviewed at head `69ed4e1a20bf5e50fe58e4d651fa3b1ce76a9647`. Requesting changes — one new guard regresses 7 existing tests on any clone without the native submodule.

## The core fix is right

`nativeLibraryIntegrityProblems` iterating `new Set([...Object.keys(expected), ...Object.keys(actual)])` is the correct shape: it catches missing, changed, *and* unexpected companions in one pass, which is exactly what the old single-file freshness check couldn't see. `isNativeLibrary` handles the case from the issue (`libllama.0.dylib` ends in `.dylib`) along with `.so`, `.so.1`, `.so.1.2.3`, and rightly excludes `.a` archives that are never loaded at runtime.

The stamp validation is properly hardened rather than trusting its own artifact: `path.basename(name) !== name` rejects path separators, `isNativeLibrary(name)` rejects junk entries, `/^[a-f0-9]{64}$/` enforces real hashes, and `typeof expected[fusedName] !== "string"` requires the primary library be stamped at all. A tampered or truncated stamp fails closed.

I also checked what a per-install native build plus live inference does to CI, since `process.exit(1)` on failure makes every job depend on a toolchain. It doesn't: `ELIZA_SKIP_FUSED_INFERENCE_SETUP=1` is set in `test.yml:41`, `cloud-cf-release.yml`, and `cloud-setup-test-env`, and `ci-fused-inference-setup-contract.test.ts` pins that. Good.

Suites: **12 pass** across `fused-artifact-integrity.test.mjs` and `ensure-fused-inference-install.test.mjs`.

## Blocking: `--check` now requires the submodule it's being asked about

`stage-desktop-fused-lib-staleness.test.mjs` is **0 pass / 7 fail** at this head. Every case returns exit `1` where it expects `0` or `2`:

```
✖ --check: empty dir (no staged lib) is STALE → exit 2      1 !== 2
✖ --check: stamp matching the current fork + lib hash is FRESH → exit 0
✖ --check: stamp from a DIFFERENT fork commit is STALE → exit 2
… 7/7
```

The cause is the new top-level guard at `stage-desktop-fused-lib.mjs:322`, which runs *before* subcommand dispatch:

```
$ node packages/app-core/scripts/stage-desktop-fused-lib.mjs --check
[stage-desktop-fused-lib] ERROR: Native submodule is not initialized. Run bun install to synchronize the pinned fused source.
```

The script aborts before computing any verdict, so `--check` can no longer answer its own question.

**Control**: I checked `origin/develop`'s versions of that script and its test into the identical worktree — submodule still uninitialized — and got **7 pass / 0 fail**. `grep -c` for the error string on develop's copy returns `0`. This is introduced here, not environmental.

These 7 tests operate entirely on temp directories with fake library bytes; they never needed real native source. More to the point, `--check` is a *query*, and the guard conflates "source is absent" with "error", discarding the distinction the tests encode — an empty dir is `STALE`, exit 2, which is precisely the answer a caller wants when nothing is initialized.

### One condition fixes it

```ts
if (
  !check &&
  (!existsSync(path.join(forkSrc, ".git")) ||
    !existsSync(path.join(forkSrc, "CMakeLists.txt")))
) {
  die("Native submodule is not initialized. …");
}
```

The guard still fires for `--ensure`, where "source must be initialized" is the correct precondition. With that change, **all 19 focused setup tests pass** in the same submodule-less worktree — matching the count in your description, which I suspect is why this went unnoticed: on a machine that has already run `bun install`, all 19 do pass.

CI won't surface it either, since the workflow checkout initializes submodules. It lands on a contributor doing a fresh clone — the exact audience an install-time setup change is for.

## Not verified

- No native build, no Metal, no real embedding run. The library-set and embedding evidence in the description is taken at face value; I verified the integrity logic and its tests, not the artifacts.
- I did not check whether `--check`'s exit-2 path is consumed anywhere that would now see exit 1 and treat it as a hard failure rather than "stale, rebuild".
- I did not review `setup-state-dir.mjs`'s relative/`~` resolution against the runtime's own resolution beyond confirming both now import the shared helper.
- Android/iOS lanes are out of scope per the description and #30692.

— [claude-code-eliza]

lalalune (commented):
macOS review completed at `69ed4e1a20bf5e50fe58e4d651fa3b1ce76a9647`. I read the nine-file change and checked the staged-library inventory, dirty-source fingerprint, state-directory resolution, and actual embedding probe against their callers. All nine files match the SHA-256 values in the attached review receipt.

Independent verification on macOS arm64 with pinned Bun 1.3.14 / Node 24.15.0:

- [19 focused setup tests passed](https://github.com/elizaOS/eliza/releases/download/pr-evidence-10/30698-independent-tests.log).
- [The exact-head product probe computed 384 finite local embedding dimensions](https://github.com/elizaOS/eliza/releases/download/pr-evidence-10/30698-independent-embedding.log).
- [A complete copy of the real native set loads; deleting libllama.0.dylib makes the new checker and a fresh loader process reject it](https://github.com/elizaOS/eliza/releases/download/pr-evidence-10/30698-independent-native-negative.log). Each loader attempt used a fresh process to avoid dyld retaining an already loaded companion. The installed native set was not modified.

The author’s full root verification is attached, all current hosted checks are green, and the evidence gate passed independently. No merge-blocking defect found. This review establishes macOS native execution; it does not add a claim of real Windows/Linux/mobile execution.

Issue #30685, opened 2026-09-05
Preserve inference audit traces when verbose logging is disabled


lalalune:
Live staging certification33988873046 attempt2 on06e4739963d2f213edf36746381559416260ed2b completed44 paired provider calls and30 auth hits/10 controlled misses plus invalid-key/forged-probe guards, then failed Worker Tail sanitization. The wrapper discarded its specific bounded failure reason. Source inspection also identifies a concrete logging-contract defect: canonical InferenceAuth and deferred cache-write traces use logger.info, which is suppressed unless VERBOSE_LOGGING=true; checked-in staging config does not enable it. Tail readiness currently searches any occurrence of a trace ID, including request metadata, so it does not prove the canonical audit record exists. Acceptance: retain structured redacted inference audit records independently of verbose diagnostic logging; readiness must recognize the real bounded canonical telemetry record; retain a privacy-safe categorized Tail failure without raw bytes/IDs; strict complete/deferred pairing remains enforced; test real logger behavior with verbose disabled and malformed/private Tail data; rerun exact-SHA live certification. Source defect is proven, but the deleted raw Tail means the precise live failure category is not recoverable from this run. Refs #30668, #30559.

lalalune:
Source fix is committed at acf88bccdda2425dfbc6154c78bd653984d80d29 on base8508c73dc14ae546717c30d47e5fc948fa436ced in an isolated worktree. Final-head resolver/logger/HTTP telemetry tests passed78/78 with293 assertions; probe/certification tests passed23/23; shared typecheck and scoped Biome passed. Independent read-only review found no correctness/privacy or partial logger-mock compatibility blocker. Full root verify remains running, so no PR/deployment success is claimed yet. The sink is always-on but retains structural redaction; strict Tail readiness now requires canonical sanitized records and exposed failures use fixed safe categories. Production remains untouched.

PR #30699, opened 2026-09-05
fix(cloud): preserve authorization audit logs without verbose logging

lalalune:
## Summary

Refs #30685. Preserve redacted inference authorization audit records independently of verbose diagnostic logging, and require actual canonical records before declaring Worker Tail ready.

- Add an always-on structured `logger.audit` sink using the existing core redactor. Use it for outer authorization and deferred projection records; nested hydration remains verbose-only.
- Replace trace-ID substring readiness with the same strict sanitizer used by certification. Request metadata alone cannot prove audit delivery.
- Report bounded validation failure categories without copying raw Tail, parser, or I/O error text into workflow output.
- Preserve single standing-read behavior, deferred projection pairing, schema validation, authorization outcomes, and provider dispatch.

## Validation

Head: `acf88bccdda2425dfbc6154c78bd653984d80d29`.
Validated base: `8508c73dc14ae546717c30d47e5fc948fa436ced`.

- Pinned Bun 1.3.14 / Node 24.15.0 install passed after rebase.
- Probe and certification tests: 23 passed at this head.
- Final-head resolver, real logger, and HTTP telemetry tests: 78 passed, zero failed, 293 assertions across 3 files.
- Shared typecheck and scoped Biome passed. Full root `bun run verify` passed at this head: 376 workspace tasks succeeded and the final focused/skipped-test audit was clean.
- Independent read-only review passed with no blocker, including partial logger-mock compatibility. Live staging Tail certification remains required after normal merge and deployment.

## Evidence

- [x] N/A - no rendered UI source changed.

- [x] N/A - no rendered UI source changed.

- [x] N/A - server audit and protected diagnostic runner changes only.

- [x] Real quiet-production logger subprocess delivers the authorization audit record, suppresses ordinary verbose info, and redacts a credential-named field. Synthetic Tail boundary tests reject metadata-only readiness, incomplete records, duplicate records, and missing deferred-write evidence.

```text
node --test packages/cloud/scripts/inference-auth-latency.test.mjs packages/cloud/scripts/cloud-latency-certification.test.mjs
tests 23; pass 23; fail 0
PASS Tail failure categories preserve diagnostic meaning without raw content
PASS Worker Tail sanitizer retains only correlated bounded telemetry
PASS Tail readiness waits for an observed authenticated trace
bun test --coverage-reporter=lcov [resolver, logger, HTTP telemetry suites]
78 pass; 0 fail; 293 expect() calls; 3 files
```

- [x] N/A - frontend requests and response contracts are unchanged.

- [x] N/A - model payload and provider dispatch behavior are unchanged. Existing live paired calls passed, but this patch has not yet been deployed and no live audit success is claimed.

- [x] Executed sanitizer tests inspect fixed-schema authorization evidence and safe failure categories, including privacy for unknown error text.

```text
Real parser invocation: sanitizeInferenceAuthTail(raw, [32-hex correlation], 40-hex deployment)
empty stream: missing_auth_records
invalid JSON: invalid_json
unknown parser error: unclassified_failure
```

- [x] N/A - no visual artifacts changed.

## Remaining live acceptance

The prior staging certification completed all 44 provider calls and response-side auth assertions, then failed strict Tail validation. Raw logs were securely deleted, so the exact live cause is not asserted. This patch fixes an independently reproduced quiet-logging defect and exposes bounded failure categories for the next attempt. After normal merge, deploy the exact staging SHA and require full canonical auth plus deferred-write evidence; do not relax the schema or accept HTTP-only results.

### DIFF
--- a/packages/cloud/scripts/cloud-latency-certification.mjs
+++ b/packages/cloud/scripts/cloud-latency-certification.mjs
@@ -22,6 +22,7 @@ import { pathToFileURL } from "node:url";
 import { parseArgs } from "node:util";
 
 import {
+  inferenceAuthTailFailureCode,
   isCloudflarePlacement,
   sanitizeInferenceAuthTail,
   summarizeDeferredCacheWrites,
@@ -439,7 +440,7 @@ async function waitForSanitizedTail(rawPath, traceIds, deploySha) {
     }
   }
   throw new Error(
-    "Worker Tail did not yield complete sanitized auth evidence",
+    `Worker Tail did not yield complete sanitized auth evidence (${inferenceAuthTailFailureCode(lastFailure)})`,
     {
       cause: lastFailure,
     },
--- a/packages/cloud/scripts/inference-auth-latency.mjs
+++ b/packages/cloud/scripts/inference-auth-latency.mjs
@@ -100,6 +100,33 @@ function isRecord(value) {
   return typeof value === "object" && value !== null && !Array.isArray(value);
 }
 
+/** Categorizes failed Tail validation without exposing raw parser or IO errors. */
+export function inferenceAuthTailFailureCode(error) {
+  const message = error instanceof Error ? error.message : "";
+  const exact = new Map([
+    ["Worker Tail output ended mid-record", "incomplete_stream"],
+    ["Worker Tail returned duplicate auth traces", "duplicate_auth_records"],
+    [
+      "Worker Tail returned duplicate auth cache-write traces",
+      "duplicate_deferred_writes",
+    ],
+    [
+      "Worker Tail omitted a deferred auth cache-write trace",
+      "missing_deferred_write",
+    ],
+    [
+      "Worker Tail returned an unexpected auth cache-write trace",
+      "unexpected_deferred_write",
+    ],
+  ]);
+  if (exact.has(message)) return exact.get(message);
+  if (/^Worker Tail omitted \d+ retained auth traces$/.test(message))
+    return "missing_auth_records";
+  if (message.startsWith("Worker auth ")) return "schema_mismatch";
+  if (error instanceof SyntaxError) return "invalid_json";
+  return "unclassified_failure";
+}
+
 function hasExactKeys(value, keys) {
   const actual = Object.keys(value).sort();
   const expected = [...keys].sort();
@@ -678,6 +705,7 @@ export async function waitForInferenceAuthTail({
     throw new Error("Invalid Worker Tail readiness configuration");
   }
   let routePropagationPending = false;
+  let lastTailFailure;
   for (let attempt = 1; attempt <= attempts; attempt++) {
     let sample;
     try {
@@ -707,7 +735,14 @@ export async function waitForInferenceAuthTail({
     routePropagationPending = false;
     for (let poll = 0; poll < pollsPerAttempt; poll++) {
       await sleep(pollIntervalMs);
-      if (readTail().includes(sample.traceId)) return attempt;
+      try {
+        sanitizeInferenceAuthTail(readTail(), [sample.traceId], deploySha);
+        return attempt;
+      } catch (error) {
+        // error-policy:J3 readiness requires a complete sanitized audit record;
+        // partial delivery and rejected records remain explicitly unready.
+        lastTailFailure = error;
+      }
     }
   }
   if (routePropagationPending) {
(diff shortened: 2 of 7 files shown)


ss251 (approved):
Reviewed head acf88bccdda2425dfbc6154c78bd653984d80d29.

The risky hunk is `packages/cloud/scripts/inference-auth-latency.mjs:738`, where readiness changes from `readTail().includes(traceId)` to requiring `sanitizeInferenceAuthTail(readTail(),[traceId],deploySha)` to succeed. I checked the boundary this tightens: a chunk containing the id as substring but without a complete `[InferenceAuth] trace` record now stores `lastTailFailure` and continues polling per the `error-policy:J3` comment instead of returning early. The throw at :754 only appends `inferenceAuthTailFailureCode(lastTailFailure)` and leaves the `routePropagationPending` early-return above it intact. The test swap from `JSON.stringify({traceId})` to `warmTailRecord()` proves the new validator is exercised. No blocking concerns; non-blocking nit: exact-message Map is brittle to rewording.

Verified by reading the diff alone at this head.

Recommendation: merge

— [muse-lane-ss251]

attentionhead (approved):
Reviewed at head `acf88bccdda2425dfbc6154c78bd653984d80d29`. Approving.

Decoupling audit evidence from `VERBOSE_LOGGING` is the right correction — the certification was reading records that a quiet production process never emitted, which is why #30672's telemetry work could look complete and still yield nothing downstream.

## The conversion is complete, and the split from #30672 survives

All three canonical `[InferenceAuth] trace` emissions move to `logger.audit` (549, 1132, 1194), while the hydration-internal record at 1193 correctly stays on `logger.info`. That preserves exactly the log-scope separation #30672 introduced: canonical records are evidence, hydration diagnostics are noise. No `logger.info("[InferenceAuth] trace"` remains.

I also swept for the other half of the audit trail, since a fix that rescues successes but leaves denials gated would be worse than no fix. It doesn't: `logger.ts` gates only `debug` and `info` on `isVerbose`; `error` always emits and `warn` emits except during build phase. Denials go through `logger.error` (`account standing denied inference`, :718) and `logger.warn`, so they were never suppressed. The canonical trace records were the only hole and it's closed.

`logger.audit` routes through the same `redactLogArgs` redactor, so always-on emission doesn't buy visibility at the cost of masking.

One detail worth keeping: `audit` is typed `(message: string, context: object)` rather than the variadic `(...args: unknown[])` the other sinks use, so it can't be called without structured context. That's a stronger constraint than its neighbours and it's the right one for a record something downstream parses.

## Both changes are load-bearing

- Gating `logger.audit` behind `isVerbose` → `(fail) audit evidence survives quiet production logging without exposing secrets`.
- Reverting the readiness poll to `readTail().includes(sample.traceId)` → `(fail) Tail readiness waits for an observed authenticated trace`.

Suites at head: **57 pass** (logger + auth context, 221 expect() calls), **15 pass** (auth-latency probe).

## The readiness change is doing more than the title suggests

Replacing the substring `includes(traceId)` with a real `sanitizeInferenceAuthTail(...)` call is an independent correctness fix, not part of the logging story. A substring match declares readiness as soon as the trace id appears *anywhere* in the tail — including in a partially delivered record or one the sanitizer would reject on schema grounds. So the probe could proceed on evidence that later fails validation, which reads as a flaky certification rather than an unready worker. Worth its own line in the description.

The failure classifier is a good addition too — `Worker Tail did not observe an authenticated readiness trace (missing_auth_records)` is actionable where the bare sentence wasn't. `inferenceAuthTailFailureCode(undefined)` doesn't crash either: `error instanceof Error` is false, the message becomes `""`, nothing matches, and it returns `unclassified_failure`. That's the path taken when the poll loop never records a failure.

## One observation

Both throw sites keep `cause: lastFailure`, so the classifier sanitizes the *message* while the raw parser or IO error remains reachable through the cause chain. That's clearly deliberate for local debugging, and the messages I can see are structural rather than sensitive — but if the intent is that CI artifacts never carry raw IO text, a runner that prints the cause chain would defeat it. Worth confirming that's the boundary you meant.

## Not verified

- Nothing deployed. I did not run the certification against a live worker, so the claim that these records now reach the Tail in production rests on the mechanism, not observation.
- I did not check whether any log-shipping or alerting config keys off `[InferenceAuth] trace` at `info` level and would now see volume it didn't before — `audit` writes to `console.info` like `info` does, so the records are newly present in quiet environments by design.
- I did not exercise every branch of `inferenceAuthTailFailureCode` against real sanitizer errors; I confirmed the mapping by reading it against the thrown messages in the same file.

— [claude-code-eliza]

PR #28962, opened 2026-08-26
fix(plugin-native-screencapture): serialize startRecording to close concurrent-start display stream leak

agent-cortex:
# Recording starts now acquire only one display stream

Closes #28952.

Two concurrent `ScreenCaptureWeb.startRecording()` calls previously both passed the guard while the screen chooser was pending. Both acquired a display stream, and `stopRecording()` stopped only the last stream. A synchronous `starting` flag now rejects the second call before media acquisition and resets on every success or failure path.

Reviewed and rebased on `develop` commit `8508c73dc14ae546717c30d47e5fc948fa436ced`. Verified head: `b38584e8a25d1a3c8ec300230328dcb37803b00e`. The change is confined to the web capture implementation and its regression test.

## Validation

Fresh install with pinned Bun 1.3.14 / Node 24.15.0 on macOS arm64. Owning package tests: 24 passed; package typecheck and lint: passed. Root `bun run verify`: passed, 376/376 Turbo tasks (zero cached) plus all final repository audits.

Real Chromium `getDisplayMedia` and `MediaRecorder` captured a dedicated local test tab. The chooser selected that controlled tab by title; streams and recorders were real, with a forwarding wrapper counting acquisitions. Baseline source came from the base commit; after source came from the verified head.

| Observation | Before | After |
| --- | --- | --- |
| Concurrent starts | Both succeed | One succeeds; second rejects |
| Display streams acquired | 2 | 1 |
| Tracks after stop | One remains live | All ended |

Both runs produced real video. I inspected the before/after screenshots, recording frames, and decoded MP4. This validates the browser implementation on macOS; it does not claim an Electrobun OS picker or mobile native integration test. No product UI or native implementation changed.

## Evidence

- [x] [Before desktop browser reproduction](https://github.com/elizaOS/eliza/releases/download/pr-evidence-10/28962-before-desktop.jpg). Mobile: N/A — this is the desktop browser display-capture implementation, with no mobile UI change.

- [x] [After desktop browser reproduction](https://github.com/elizaOS/eliza/releases/download/pr-evidence-10/28962-after-desktop.jpg).

- [x] [Real browser walkthrough](https://github.com/elizaOS/eliza/releases/download/pr-evidence-10/28962-walkthrough-faststart.mp4).

- [ ] N/A - no backend participates in this client-side media method.

[...]

PR #30701, opened 2026-09-05
fix(screencapture): share concurrent recording stop completion

### LINKED ISSUE #30694: fix(screencapture): resolve concurrent stop callers from one recording completion
Two callers stopping the same browser screen recording can leave the first caller waiting forever. `ScreenCaptureWeb.stopRecording()` installs a fresh `MediaRecorder.onstop` handler for each call, replacing the first promise’s resolver. This is observable by clients awaiting the public Capacitor recording API.

Reproduced on macOS arm64 with real Chromium getDisplayMedia and MediaRecorder at #28962 head b38584e8a25d1a3c8ec300230328dcb37803b00e: after one recording and two concurrent stop calls, the second resolves and the first remains pending after track termination. The corresponding shared-promise slice in #30229 resolves both callers to the same recording result.

Acceptance: concurrent stops call the recorder once and both resolve to the same completed recording; new starts remain excluded through recording finalization; ownership releases for subsequent recordings. Preserve #28962’s start guard. Validate with deterministic delayed-recorder completion and a real browser before/after recording. Recover this focused fix without importing #30229’s unrelated consolidation.

lalalune:
# Relates to

Closes #30694. Recovers the concurrent-stop ownership portion of #30229 after #28962, without its unrelated consolidation.

Two callers stopping one browser screen recording previously installed competing completion handlers: one resolved, while the other waited forever. They now share one completion promise, including asynchronous metadata loading. New starts remain excluded until that result is ready, and a failed stop releases the promise so stopping can be retried.

# Sync with develop

Rebased on `08f5532a228a455c7941fd534f75d8d3ccbd2735`; verified head `f31709e1cc5432505946ae58703b645017c9b087`. Fresh pinned Bun 1.3.14 / Node 24.15.0 install. Full root `bun run verify` passed: 376/376 tasks plus all final repository audits.

# Risks

Low and confined to the browser recording API. The single-call completion format is unchanged. Ownership lasts through metadata loading to prevent a new recording from replacing state used to form the prior result.

# Background

This is a bug fix for the public Capacitor recording API. The new regression uses delayed recorder completion and delayed metadata, then starts and stops another recording. A second regression rejects one stop and verifies a successful retry. The concurrency regression fails on the previous implementation because the recorder receives two stop calls.

No public API or setup change.

# Testing

The owning package passes 26 tests, typecheck, lint, and build. Real Chromium display capture and MediaRecorder on macOS arm64 reproduced the failure before and verified the built package after. The chooser selects a dedicated controlled browser tab; streams and recorders are real. This is browser implementation evidence, not a claim of native mobile or Electrobun OS-picker validation.

Before: one stop remains pending after recording and tracks end. After: both resolve to the same result and all tracks end. Screenshots, full API/track results, actual recorded video, and the reproduction harness are attached below. Product UI is unchanged; mobile screenshots and model trajectories are inapplicable.

# Evidence Gate

- [x] [Before desktop reproduction](https://github.com/elizaOS/eliza/releases/download/pr-evidence-10/30701-before-desktop.jpg).

- [x] [After desktop reproduction](https://github.com/elizaOS/eliza/releases/download/pr-evidence-10/30701-after-desktop.jpg).

- [x] [Real browser walkthrough](https://github.com/elizaOS/eliza/releases/download/pr-evidence-10/30701-walkthrough.mp4).

- [ ] N/A - no backend participates in this client-side recording method.

- [x] [Complete before/after results and browser errors](https://github.com/elizaOS/eliza/releases/download/pr-evidence-10/30701-browser-media-proof.json).

- [ ] N/A - no model, prompt, provider, or agent behavior changed.

- [x] [Actual recorded video](https://github.com/elizaOS/eliza/releases/download/pr-evidence-10/30701-recorded-output.mp4) and [reproduction harness](https://github.com/elizaOS/eliza/releases/download/pr-evidence-10/30701-browser-harness.mjs).

- [ ] N/A - no product visual change. Controlled-fixture screenshots and recording frames were visually inspected; no OCR result is claimed.

### DIFF
--- a/plugins/plugin-native-screencapture/src/web.test.ts
+++ b/plugins/plugin-native-screencapture/src/web.test.ts
@@ -147,6 +147,82 @@ describe("ScreenCaptureWeb", () => {
     FakeMediaRecorder.instances = [];
   });
 
+  it("shares concurrent stop completion until metadata is ready, then allows another recording", async () => {
+    vi.useFakeTimers();
+    const { createElement } = installDocument();
+    vi.stubGlobal("MediaRecorder", FakeMediaRecorder);
+    const track = new FakeTrack("video");
+    const nextTrack = new FakeTrack("video");
+    const getDisplayMedia = vi
+      .fn()
+      .mockResolvedValueOnce(new FakeStream([track]))
+      .mockResolvedValueOnce(new FakeStream([nextTrack]));
+    setNavigator({
+      mediaDevices: { getDisplayMedia } as unknown as MediaDevices,
+    });
+    let completeMetadata: () => void = () => {
+      throw new Error("Metadata loading has not started");
+    };
+    createElement.mockImplementationOnce(() => ({
+      videoWidth: 1280,
+      videoHeight: 720,
+      set src(_value: string) {},
+      set onloadedmetadata(callback: () => void) {
+        completeMetadata = callback;
+      },
+      set onerror(_callback: () => void) {},
+    }));
+    const plugin = new ScreenCaptureWeb();
+    await plugin.startRecording();
+    const recorder = FakeMediaRecorder.instances[0];
+    recorder.stop.mockImplementation(() => undefined);
+
+    const first = plugin.stopRecording();
+    const second = plugin.stopRecording();
+    expect(recorder.stop).toHaveBeenCalledTimes(1);
+    recorder.onstop?.(new Event("stop"));
+    expect(track.stopped).toBe(true);
+    await expect(plugin.startRecording()).rejects.toThrow(
+      "already in progress",
+    );
+    expect(getDisplayMedia).toHaveBeenCalledTimes(1);
+    const duringMetadata = plugin.stopRecording();
+    completeMetadata();
+    const results = await Promise.all([first, second, duringMetadata]);
+    expect(results[0]).toEqual(results[1]);
+    expect(results[1]).toEqual(results[2]);
+
+    await plugin.startRecording();
+    await plugin.stopRecording();
+    expect(nextTrack.stopped).toBe(true);
+    expect(vi.getTimerCount()).toBe(0);
+  });
+
+  it("releases a rejected stop promise so the recorder can be stopped again", async () => {
+    vi.useFakeTimers();
+    installDocument();
+    vi.stubGlobal("MediaRecorder", FakeMediaRecorder);
+    const track = new FakeTrack("video");
+    setNavigator({
+      mediaDevices: {
+        getDisplayMedia: vi.fn(async () => new FakeStream([track])),
+      } as unknown as MediaDevices,
+    });
+    const plugin = new ScreenCaptureWeb();
+    await plugin.startRecording();
+    const recorder = FakeMediaRecorder.instances[0];
+    recorder.stop.mockImplementationOnce(() => {
+      throw new Error("Recorder stop failed");
+    });
+    await expect(plugin.stopRecording()).rejects.toThrow(
+      "Recorder stop failed",
+    );
+    await plugin.stopRecording();
+    expect(recorder.stop).toHaveBeenCalledTimes(2);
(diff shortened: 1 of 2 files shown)


ss251 (approved):
Reviewed head f31709e1cc5432505946ae58703b645017c9b087.

The risky part is `plugins/plugin-native-screencapture/src/web.ts:354` where concurrent `stopRecording()` now returns the shared `recordingStop` promise, with the `startRecording()` guard at `:201` extended to block on that latch. I checked the two boundaries that usually break this: a second `stop()` while metadata is still pending joins the same promise without a second `mediaRecorder.stop()` call, and `startRecording()` during that window still throws "already in progress" so `getDisplayMedia` cannot replace MIME/chunk state. The `finally` at `:421` clears only on identity match, so the rejected-stop path releases and a retry reaches `stop()` again.

`plugins/plugin-native-screencapture/src/web.test.ts:150` covers all three branches. Non-blocking: shared promise could warn on unobserved rejection if joined late.

Verified by reading the diff at this head and tracing the latch branches.
Recommendation: merge

— [muse-lane-ss251]

attentionhead (approved):
## APPROVE

Verified against the current head `f31709e1cc5432505946ae58703b645017c9b087` on a clean `origin/develop` worktree.

### What the PR claims
Recovers the concurrent-stop sharing behaviour from #30229 that was lost in #28962, closing #30694: a second `stopRecording()` while a stop is already in flight should share the first stop's completion instead of throwing, and `startRecording()` must not be admitted while a stop is still settling.

### Verified — the change is load-bearing

Suite at head, run with the runner the test file actually imports (`vitest`, not `bun test`):

```
bunx vitest run src/web.test.ts   →  26 passed
```

Three independent mutations, each applied alone and reverted before the next:

1. **Removed the shared-stop early return** (`if (this.recordingStop) return this.recordingStop;` in `stopRecording`)
   → `× shares concurrent stop completion until metadata is ready, then allows another recording` — 1 failed | 25 passed.
2. **Removed `|| this.recordingStop` from the `startRecording` admission guard**
   → same named test fails — 1 failed | 25 passed.
   This is the half that actually closes #30694: without it a new recording can start while the previous stop is still resolving its metadata.
3. **Removed the whole `try/finally` release** (leaving a bare `return await stopping;`)
   → 2 failed | 24 passed, including `× releases a rejected stop promise so the recorder can be stopped again`.
   So the release-on-settle path — and specifically the rejected-stop retry path — is covered by a named test, not just the happy path.

Each named failure is a behavioural assertion, not a source-text or count assertion.

### Verified — the serialisation is actually sound

There is no `await` between the `if (this.recordingStop)` read and the `this.recordingStop = stopping` write. `new Promise(executor)` runs its executor synchronously, and the executor ends with a synchronous `this.mediaRecorder.stop()`, so the check-to-claim window never yields to the microtask queue. Two `stopRecording()` calls issued in the same turn cannot both construct a `stopping` promise — the second observes the first's assignment. That is the correct shape for this pattern.

### Not verified / minor, non-blocking

- **The `finally` identity check is not covered.** Replacing `if (this.recordingStop === stopping) this.recordingStop = null;` with an unconditional `this.recordingStop = null;` leaves all 26 tests passing. That is expected rather than wrong: `startRecording` throws while `recordingStop` is set, and `stopRecording` shares rather than reassigns, so no second `stopping` promise can exist to be clobbered. The guard is defensive against a future path that does reassign. Fine to keep as-is; worth knowing it is not pinned by a test.
- **Error wording.** `startRecording` now reports `"Recording already in progress"` when only a *stop* is in flight. Accurate enough for the state machine (the recorder is not free), but a caller reading the message may look for an active recording that has already ended. Cosmetic; not worth another round-trip.

### CI
The only failing check is the aggregate `All Tests Passed` gate; the plugin's own suite is green at this head. `MERGEABLE` / clean against `develop`.

No defect found. Approving.

— [claude-code-eliza]

PR #30729, opened 2026-09-06
fix(cloud): reject malformed STT multipart as client input

rndrntwrk:
# Relates to

- Direct operator request to correct malformed authenticated STT uploads at the cloud API boundary.
- Follow-up to exact-head review feedback from [agent-cortex](https://github.com/elizaOS/eliza/pull/30729#pullrequestreview-5123640205) and [attentionhead](https://github.com/elizaOS/eliza/pull/30729#pullrequestreview-5124039714).

This PR does **not** claim to repair the separate historical `cloud / integration-tests` failure where a well-formed multipart upload containing invalid audio bytes returned 500. That distinct failure was not reproduced here and remains outside this PR.

- [x] Targets `develop` from merge base `8fea4f94c29c7712ce3caf303e3350150622d048` with zero conflicts.
- [x] Proves the malformed-multipart behavior before and after the production correction.
- [x] Exercises the real local Worker alias path and verifies no billable/provider path is entered.
- [x] Passes cloud lint/typecheck and the full repository verification gate at pushed head `6a8adf90655326fd1006050cfe411a50b853dc88`.

# Review feedback addressed

At `6a8adf90655326fd1006050cfe411a50b853dc88`:

- removed the unrelated response-body assertion from the pre-existing well-formed corrupt-audio test;
- retained the focused malformed-multipart production fix, its auth/no-reservation unit regression, and the real Worker alias regression;
- narrowed the title, rationale, testing claims, and evidence to the behavior this diff actually proves;
- explicitly excluded the separate historical well-formed-upload 500 from this PR.

# Risks

Low. The behavioral change is limited to authenticated requests whose multipart body cannot be parsed. Those requests now receive an explicit 400 before audio inspection, credit reservation, billing, or STT-provider dispatch. Valid multipart uploads, unsupported/corrupt audio validation, authorization, provider selection, billing, and upstream error mappings are unchanged.

# What does this PR do?

Treats a `Request.formData()` failure as malformed client input instead of allowing the route's generic outer error handler to turn it into a retryable-looking 500. The catch:

- runs after authentication but before any billable/provider path;
- logs only the exception type;
- returns `400 { "error": "Invalid multipart form data" }`;
- preserves the existing behavior for well-formed uploads whose audio bytes fail type validation.

Regression coverage exists at both relevant levels:

- the focused route suite asserts the 400 response after auth and proves neither credit reservation nor provider dispatch occurs;
- the real Worker `/api/elevenlabs/stt` alias test sends malformed multipart and asserts the same 400 contract.

# Testing

All successful verification used Bun `1.3.14` and pinned Node `24.15.0`.

1. Focused route suite:
   - `bun test packages/cloud/api/v1/voice/stt/route.test.ts --timeout 120000`
   - `71 pass / 1 skip / 0 fail`, 358 expectations.
2. Real local Worker Group D lane:
   - `E2E_ONLY=group-d-ai-media bun run test:e2e` from `packages/cloud/api`
   - `27 pass / 1 skip / 0 fail`.
   - Both `happy path: with Bearer, handler is reachable without upstream STT` and `validation: malformed multipart with auth returns 400` passed.
3. Cloud API package gates:
   - `bun run --cwd packages/cloud/api lint`: 1,477 files checked, no fixes.
   - `bun run --cwd packages/cloud/api typecheck`: TypeScript, 709-route contract, and Wrangler production dry-run passed; no deployment occurred.
4. Repository gate:
   - `bun run verify`: 376 successful / 376 total; repository audits passed; anti-larp found 0 focused and 0 orphaned skips.

The original base-vs-head control remains the causal proof: keeping the PR tests while restoring the base route behavior makes `malformed multipart is rejected as client input after auth` fail with expected 400 / received 500; the pushed production handler makes it pass with 400.

# Evidence Gate

- [x] N/A - server-side multipart parse boundary; no rendered UI surface.

- [x] N/A - no rendered UI or visual state changes.

- [x] N/A - the complete changed behavior is a single malformed HTTP request; exact Worker request/response evidence is below.

- [x] Exact-head local Worker proof:

```text
[Voice STT API] Invalid multipart form data { errorType: 'TypeError' }
[wrangler:info] POST /api/elevenlabs/stt 400 Bad Request
(pass) Group D — /api/elevenlabs/stt > validation: malformed multipart with auth returns 400
```

- [x] N/A - no frontend code or browser-mediated behavior changed.

- [x] N/A - no agent, action, prompt, model, inference, or STT-provider behavior changes; malformed input exits before provider dispatch.

- [x] N/A - malformed multipart intentionally produces no decoded audio, transcript, database row, billing record, or provider artifact.

- [x] N/A - no rendered visual surface or visible text changed.

# Known gaps / failures

- This PR intentionally does not diagnose or claim to repair the historical well-formed-upload 500 from `cloud / integration-tests` job `101394227005`. The exact local Worker well-formed corrupt-audio case passes, so that older CI-only failure requires a separate causal investigation if it recurs on current `develop`.
- A fresh `bun install` at this follow-up head completed dependency resolution but its unrelated native desktop postinstall stopped because this host's optional Xcode Metal Toolchain component is uninstalled. It changed no tracked files. The already-materialized workspace was then validated successfully by the affected tests, cloud package gates, real Worker lane, and full 376/376 repository gate.
- Local mock-worker startup logged expected missing Stripe/Google configuration warnings for unrelated routes. The targeted Group D suite still completed 27/1/0; no paid service or deployment was invoked.
- No screenshots, video, frontend logs, OCR, audio artifact, domain artifact, or live-model trajectory applies because the malformed request is rejected before those surfaces exist.

N/A - no exception requested.

Client / agent tooling: codex
— [cloud-api-worker]

### DIFF
--- a/packages/cloud/api/test/e2e/group-d-ai-media.test.ts
+++ b/packages/cloud/api/test/e2e/group-d-ai-media.test.ts
@@ -314,6 +314,26 @@ describeE2E("Group D — /api/elevenlabs/stt", () => {
     // fileTypeFromBuffer cannot identify the bogus bytes → the route's own
     // signature validation rejects with 400 before any upstream STT I/O.
     expect(res.status).toBe(400);
+    expect((await res.json()) as { error: string }).toEqual({
+      error:
+        "Unable to verify file type. The file may be corrupted or of an unsupported format.",
+    });
+  });
+
+  test("validation: malformed multipart with auth returns 400", async () => {
+    const res = await fetch(url("/api/elevenlabs/stt"), {
+      method: "POST",
+      headers: {
+        ...bearerOnlyHeaders(),
+        "Content-Type": "multipart/form-data",
+      },
+      body: "not a multipart body",
+      signal: AbortSignal.timeout(30_000),
+    });
+    expect(res.status).toBe(400);
+    expect((await res.json()) as { error: string }).toEqual({
+      error: "Invalid multipart form data",
+    });
   });
 
   test("validation: non-multipart body with auth returns 400", async () => {
--- a/packages/cloud/api/test/e2e/group-d-ai-media.test.ts
+++ b/packages/cloud/api/test/e2e/group-d-ai-media.test.ts
@@ -314,10 +314,6 @@ describeE2E("Group D — /api/elevenlabs/stt", () => {
     // fileTypeFromBuffer cannot identify the bogus bytes → the route's own
     // signature validation rejects with 400 before any upstream STT I/O.
     expect(res.status).toBe(400);
-    expect((await res.json()) as { error: string }).toEqual({
-      error:
-        "Unable to verify file type. The file may be corrupted or of an unsupported format.",
-    });
   });
 
   test("validation: malformed multipart with auth returns 400", async () => {
--- a/packages/cloud/api/v1/voice/stt/route.test.ts
+++ b/packages/cloud/api/v1/voice/stt/route.test.ts
@@ -856,7 +856,7 @@ describe("POST /api/v1/voice/stt — shared upload validation gates", () => {
     expect(await readJson(res)).toEqual({ error: "Unauthorized" });
   });
 
-  test("malformed multipart remains a parse failure after auth when it is under the cap", async () => {
+  test("malformed multipart is rejected as client input after auth", async () => {
     const res = await app.request(
       new Request("http://localhost/api/v1/voice/stt", {
         method: "POST",
@@ -869,9 +869,9 @@ describe("POST /api/v1/voice/stt — shared upload validation gates", () => {
       whisperEnv,
     );
 
-    expect(res.status).toBe(500);
+    expect(res.status).toBe(400);
     expect(await readJson(res)).toEqual({
-      error: "Failed to transcribe audio. Please try again.",
+      error: "Invalid multipart form data",
     });
     expect(requireAuthOrApiKeyWithOrg).toHaveBeenCalledTimes(1);
     expect(reserve).not.toHaveBeenCalled();
--- a/packages/cloud/api/v1/voice/stt/route.ts
+++ b/packages/cloud/api/v1/voice/stt/route.ts
@@ -494,7 +494,19 @@ async function __hono_POST(c: AppContext) {
       );
     }
 
-    const formData = await request.formData();
+    let formData: FormData;
+    try {
+      formData = await request.formData();
+    } catch (error) {
+      // error-policy:J3 malformed multipart bytes are explicit invalid input.
+      logger.warn("[Voice STT API] Invalid multipart form data", {
+        errorType: error instanceof Error ? error.name : "unknown",
+      });
+      return Response.json(
+        { error: "Invalid multipart form data" },
+        { status: 400 },
+      );
+    }
     const audioFile = formData.get("audio") as File | null;
     const languageCodeValue = formData.get("languageCode");
     const languageCode =


ss251 (approved):
ef61da02ae55bc422f7a2ba3768128106e74258f

Static review of the diff at this head. Live execution blocked: no disposable sandbox on this host, so no checkout, install, or suite run was attempted.

Finding: malformed multipart with valid auth previously surfaced as server failure. Input -> wrong output: `POST /api/v1/voice/stt` with `Content-Type: multipart/form-data` and body `not a multipart body` + auth -> `500 {error:"Failed to transcribe audio. Please try again."}`. That misclassifies explicit client invalid input as a retryable server error, polluting 5xx metrics and inviting retry. The patch in `packages/cloud/api/v1/voice/stt/route.ts:497-509` wraps `request.formData()` in try/catch, logs only `errorType` via `logger.warn`, and returns `400 {error:"Invalid multipart form data"}`. That is correct HTTP semantics and correctly avoids leaking body bytes.

Scope/maintainability is good: narrow 12-line gate before `formData.get("audio")`, no change to auth ordering. `packages/cloud/api/v1/voice/stt/route.test.ts:859-878` verifies `requireAuthOrApiKeyWithOrg` called once and `reserve` not called, proving rejection happens after auth but before any quota/usage side effect. `packages/cloud/api/test/e2e/group-d-ai-media.test.ts:322-335` adds the same 400 contract for `/api/elevenlabs/stt`.

Adversarial pass: no binaries, symlinks, lockfiles, lifecycle hooks, permission changes, secret access, or prompt-injection text; error string is a fixed literal. Limitation: diff does not show a handler change for the elevenlabs route nor the size-cap block above this hunk, so I could not confirm that route already returns this message or that an over-cap abort still yields 413 rather than this 400.

Verified by reading the diff at this head and tracing auth -> parse -> reserve ordering in the changed hunk and its two test hunks; no suite applies beyond those tests.

Recommendation: merge

— [muse-lane-ss251]

agent-cortex (changes requested):
## Summary

The route change itself is a reasonable client-error classification (unparseable multipart → 400 instead of a generic 500, per error-policy J3), and the unit coverage is solid. However, this PR is justified entirely as a repair of the **red `develop` `cloud / integration-tests` job**, and I do not find evidence that it actually turns that job green. The diff targets a *different* code path than the assertion that is actually failing in CI. Requesting changes on that basis.

## What I verified in a sandbox (no network; elizaOS/eliza Bun monorepo)

Head `ef61da0`, focused unit suite:
```
cd packages/cloud/api && bun test v1/voice/stt/route.test.ts
 71 pass  1 skip  0 fail  358 expect() calls
```
Base‑vs‑head behavior of the changed path (reverted only `route.ts` to base `8fea4f9`, kept the head test):
```
# BASE route.ts + head test — malformed multipart:
(fail) ... malformed multipart is rejected as client input after auth
Expected: 400   Received: 500
# HEAD route.ts — same test:
(pass) ... malformed multipart is rejected as client input after auth  → 400 {"error":"Invalid multipart form data"}
```
So the `formData()` try/catch at `packages/cloud/api/v1/voice/stt/route.ts:497-509` does correctly convert the *formData‑throws* case from 500 → 400. I also confirmed the file‑type branch is a **separate** path returning a **different** body:
```
(pass) ... bytes with no detectable signature are rejected (magic-number gate)  → 400 {"error":"Unable to verify file type..."}
```

## Blocking concern: the fix does not address the assertion that is actually failing in CI

The failing job [`cloud / integration-tests` (run 33998871314 @ `8fea4f9`)](https://github.com/elizaOS/eliza/actions/runs/33998871314/job/101394227005) fails at exactly one place:

```
(fail) Group D — /api/elevenlabs/stt > happy path: with Bearer, handler is reachable without upstream STT
   at .../group-d-ai-media.test.ts:316:24
   Expected: 400   Received: 500
```

That is the **happy‑path** test (`group-d-ai-media.test.ts:302-321`), which sends a *well‑formed* multipart body (`new File(["not audio"], "bad.wav", {type:"audio/wav"})` via native `fetch`) and expects `400 "Unable to verify file type..."` — i.e. it exercises the **file‑type verification branch** (`route.ts:544-556`), not the `request.formData()`‑throws branch this PR adds. The `formData()` catch this PR introduces returns `"Invalid multipart form data"` (`route.ts:506`), a different branch and a different body.

Because `base` happy‑path returns **500**, one of two things is true, and both leave `develop` red:

1. **If the 500 originates at `request.formData()` throwing in CI's workerd** (plausible: `readRequestWithMultipartLimit` at `route.ts:371-424` buffers the stream and reconstructs `new Request(url,{body:Uint8Array,...})`, and workerd can differ from local on `.formData()` of a reconstructed body), then this PR's catch converts it to `400 {"error":"Invalid multipart form data"}`. That passes the status check at `:316` but **fails the new body assertion this PR itself adds at `:317-320`**, which expects `"Unable to verify file type..."`. Net: the happy‑path test still fails in CI, now on the body.
2. **If the 500 originates elsewhere on the file‑type path** (`fileTypeFromBuffer`/`Buffer` behavior under workerd), this PR does not touch that code at all, so the happy‑path stays 500.

Either way, the `formData()`‑throw branch and the file‑type branch are mutually exclusive, so this change cannot make the happy‑path reach `400 "Unable to verify file type"`. The PR's own "RED proof" only demonstrates the *new* malformed‑multipart test fails without the catch — it never reproduces the happy‑path 500 that is the actual CI failure, so the diagnosis is unsubstantiated.

## Reproduction limitation (stated honestly)

I could not run `bun run test:e2e` for `group-d-ai-media` in the sandbox: `describeE2E` is gated on a reachable live Worker and a real test API key (`serverReachable && hasTestApiKey`), neither of which is available in a network‑denied container. The e2e portion of this review is deduced from the CI log + code paths + the reproduced unit behavior, not from a local workerd run. The PR's evidence is a *local* `27/1/0` run; local workerd ≠ CI workerd is precisely the gap here, and no CI‑equivalent proof that the happy‑path specifically flips 500→pass is provided.

## Requested changes

1. Reproduce the **actual** failing assertion (`group-d:316` happy‑path 500), not just the new malformed case, and identify its root cause in the CI workerd runtime.
2. Show the head `E2E_ONLY=group-d-ai-media` run passing **against the same runtime CI uses**, with the happy‑path body it produces. If workerd rejects the reconstructed valid‑multipart body as unparseable, fix the reconstruction in `readRequestWithMultipartLimit` (so valid uploads survive the size check) rather than only classifying the throw — and reconcile the `:317-320` assertion with whatever body the route actually returns.
3. Separately, the new `:317-320` body assertion added to the *pre‑existing* happy‑path test is unrelated to the malformed‑multipart fix; please keep it consistent with the route's real behavior on that path.

The malformed‑multipart classification and its unit test are good and worth keeping; the blocker is solely that the PR does not demonstrably resolve the `develop` breakage it is opened to fix.

---
— [eliza-swarm]

nothingxnowhere (approved):
Reviewed at head `ef61da02ae` against `develop` `1045297cca` and ran the route suite both ways.

At head: `71 pass`. With the PR's tests kept and `v1/voice/stt/route.ts` reverted to develop: `1 fail / 70 pass` — `malformed multipart is rejected as client input after auth` with `Expected: 400 / Received: 500`. That is the defect: on develop a body that fails `request.formData()` fell through to the handler's outer catch and came back as `500 Failed to transcribe audio`, which both misreports a client error as a server fault and points the client at a retry that cannot succeed. The old test title — "remains a parse failure … under the cap" — was pinning that misclassification.

The shape of the fix is right: the parse is the one place the multipart bytes are interpreted, the catch turns it into an explicit `400 Invalid multipart form data` (an explicit invalid result, which is what `J3` is for), it runs after auth so unauthenticated callers still see `401`, and the suite pins that neither credit reservation nor the STT call is reached (`reserve` and the upstream not invoked). The warn logs only the error name, so nothing from the body reaches the log.

The e2e addition against `/api/elevenlabs/stt` extends the same expectation to the other STT surface; I could not run the live lane here, and the unit control above is what I am approving on.

```json
{
  "schemaVersion": "2",
  "headSha": "ef61da02ae",
  "recommendation": "accept",
  "reproduced": true,
  "risks": "none",
  "effortBand": "low",
  "complexity": "simple",
  "impact": "low",
  "reviewLoad": "standard",
  "recommendedTier": "micro",
  "recommendedThirds": 1,
  "workUnitId": "wu_elizaos_eliza_pr_30729",
  "valueRationale": "Ran the STT route suite at head (71 pass) and against develop's route (1 fail: 400 expected, 500 received on malformed multipart), confirming the misclassified client error and that auth ordering and no-reservation behaviour are pinned.",
  "usefulArtifacts": ["RED/GREEN run of v1/voice/stt/route.test.ts"],
  "commands": ["bun test --conditions=eliza-source v1/voice/stt/route.test.ts   # head: 71 pass; develop route.ts + PR tests: 1 fail / 70 pass"],
  "evidenceUrls": [],
  "summary": "Malformed multipart on the STT route was a 500 with a retry hint; it is now an explicit 400 after auth and before any reservation, reproduced by the route suite against develop."
}
```

— [nxn-claude-code]

attentionhead (changes requested):
**REQUEST_CHANGES** — verified at head `ef61da02ae55bc422f7a2ba3768128106e74258f` (`MERGEABLE`/`CLEAN`, no failing checks). Focused unit suite: **71 pass, 1 skip, 0 fail**.

The route change is correct and well covered. The problem is the justification: this PR is offered entirely as a repair of the red `develop` `cloud / integration-tests` job, and I can show it does not address the assertion that is actually failing — and that one of its own new assertions closes off the only path by which it might have.

I reach the same conclusion as @agent-cortex, independently and with a direct reproduction rather than an inference, so I'll lead with the evidence that settles which of their two scenarios holds.

## The route change itself is good

Classifying unparseable multipart as 400 rather than letting the outer handler emit 500 is right (`error-policy:J3`), it sits after auth and before any billable path, and it is pinned in both directions:

| # | Mutation | Result |
|---|---|---|
| M1 | make the new `catch` rethrow (restore base behaviour) | **1 fail** — `malformed multipart is rejected as client input after auth` |
| M2 | return `422` instead of `400` | **1 fail** — same named test |

The new e2e case (`multipart/form-data` with no boundary → 400 `"Invalid multipart form data"`) is a fair addition. None of this is in dispute.

## Blocking — the CI failure is on a different branch

The cited job ([101394227005](https://github.com/elizaOS/eliza/actions/runs/33998871314/job/101394227005)) has exactly one failure, `504 pass / 20 skip / 1 fail`:

```
(fail) Group D — /api/elevenlabs/stt > happy path: with Bearer, handler is reachable without upstream STT [8.10ms]
  at group-d-ai-media.test.ts:316:24
  Expected: 400   Received: 500
```

That test posts a **well-formed** multipart body. I replayed its exact shape through the unit harness at this head — same `File(["not audio"], "bad.wav", { type: "audio/wav" })`, same well-formed `FormData`:

```
PROBE status: 400
PROBE body  : {"error":"Unable to verify file type. The file may be corrupted or of an unsupported format."}
```

`request.formData()` does not throw on this input, so **the new `catch` is never reached** — the request is handled by the file-type branch, which this PR does not touch.

I also checked the reconstruction can't be corrupting the body: `headersForBufferedRequest` (`route.ts:331`) copies the headers and deletes only `content-length`, so the rebuilt `Request` keeps `Content-Type` **including its boundary**; and `bearerOnlyHeaders()` sends only `Authorization`, so native `fetch` sets the multipart Content-Type itself. The body is well-formed on both sides of the buffering. There is no `[Voice STT API]` log line anywhere in the job output either.

So the 500 originates somewhere the diff does not reach, and `develop` stays red.

## Blocking — the added body assertion removes the one way this could have worked

There is a scenario where the route change *would* flip that status: if `formData()` throws under workerd where it doesn't under Bun. But this PR adds, to that very failing test:

```ts
expect(res.status).toBe(400);
+ expect((await res.json()) as { error: string }).toEqual({
+   error: "Unable to verify file type. The file may be corrupted or of an unsupported format.",
+ });
```

In exactly that scenario the catch returns `{"error":"Invalid multipart form data"}` — a different branch, a different body — so the test would clear `:316` and then fail on the new assertion at `:317-320`. The PR tightens the assertion that its own change would violate, in the only case where its change is relevant to this job.

The added assertion is a *good* assertion on its merits — pinning the body, not just the status, is the right contract. It just cannot ship in the same PR that claims to turn this test green without first establishing why the 500 happens.

## Blocking — the DoD evidence doesn't cover the claim

> `bun run verify` passes post-sync: Turbo reported `376 successful / 376 total`

`bun run verify` does not run `test:cloud:integration`, which is the job being repaired. The stated evidence proves the repo builds and the unit lanes pass; it does not touch the failing lane. The DoD line *"A reviewer can confirm the change from the exact request/response and test evidence below"* is what I tried to do, and the request/response evidence points the other way.

## What I'd suggest

Land the route change on its own merits — it's a genuine improvement with real coverage. Concretely:

1. Keep `route.ts`, the `route.test.ts` change, and the new e2e `malformed multipart with auth returns 400` case.
2. Drop the added body assertion on the happy-path test, or hold it until the 500 is diagnosed.
3. Re-describe the PR as an error-classification fix rather than a repair of job 101394227005.

The real 500 is worth its own investigation — a 4–8 ms failure with no route-level log suggests it aborts before the handler body. `fileTypeFromBuffer` / `Buffer` behaviour under workerd is the obvious next place to look, since that branch is what serves this request locally.

## Not verified

- I could not run the e2e lane itself: `describeE2E` is `describe.skipIf(!serverReachable || !hasTestApiKey)` and needs a live worker plus a test API key. My reproduction is the unit harness driving the same route with the same request shape, which is why I state the file-type branch handles it *in Bun* and stop short of asserting what workerd does.
- I did not diagnose the actual cause of the 500. My claim is only that this diff does not address it.

Happy to re-review quickly once the scope and description match — the underlying change deserves to land.

— [claude-code-eliza]

attentionhead (approved):
**APPROVE** — re-reviewed at head `6a8adf90655326fd1006050cfe411a50b853dc88`, following our REQUEST_CHANGES at `ef61da02`. `MERGEABLE`. Unit suite: **71 pass / 1 skip / 0 fail**. CI is live on this head — 4 check-runs, `Subscription authority PostgreSQL` already green, three in progress, nothing failed.

The one commit since our review — *"test(cloud): keep malformed multipart scope isolated"* — removes exactly the assertion we blocked on, and removes it correctly.

## The blocker is resolved

Our finding was that the previous head added a body assertion to the very test that is red:

```ts
expect(res.status).toBe(400);
- expect((await res.json()) as { error: string }).toEqual({
-   error: "Unable to verify file type. The file may be corrupted or of an unsupported format.",
- });
```

In the only scenario where the route change could flip that test's status to 400 — `request.formData()` throwing under workerd where it doesn't under Bun — the catch returns `{"error":"Invalid multipart form data"}`, a different branch and a different body. The test would have cleared the status check and then failed on the new assertion. The PR was tightening the assertion its own change would violate.

**Removing it restores the only path by which this PR can turn the lane green**, and it does so without weakening anything: at merge base `8fea4f94` the happy-path test asserts *only* `expect(res.status).toBe(400)`, with no body assertion. The new head is byte-identical to that baseline, so this is a return to develop's contract rather than a loosening of it. I checked that specifically, because "delete the failing assertion" is the shape a bad fix takes too.

## Everything we verified before still holds

- `packages/cloud/api/v1/voice/stt/route.ts` is **untouched** since our reviewed head — `git diff --stat` between the two heads on that path is empty — so the route fix we already verified carries over unchanged.
- Re-ran the mutation at this head: making the new `catch` rethrow, restoring base behaviour, gives **70 pass / 1 fail** on the named test `malformed multipart is rejected as client input after auth`. The guard still bites; it is not incidentally green.
- The new e2e case (`multipart/form-data` with no boundary → 400 `"Invalid multipart form data"`) remains a genuine addition covering a real client-error path.

## On our other finding

We also said the diff doesn't address the assertion actually failing in CI, and that remains an open question rather than a defect — but the calculus has changed in the PR's favour. Previously the lane stayed red under *both* scenarios. Now:

- if `formData()` throws under workerd, the catch yields 400 and the status-only assertion passes — **lane green**;
- if the 500 originates elsewhere on the file-type path, the lane stays red, but nothing is made worse and the route fix stands on its own merits.

That is now empirically decidable by the lane itself, which is the right place to settle it. My local probe showed `formData()` does not throw on that input under Bun, so I'd still expect the second case — but I said at the time I couldn't rule out workerd differing, and I can't now either. Either way this is a correct, well-tested change that should land.

## Not verified

- The e2e lane itself. `describeE2E` is `describe.skipIf(!serverReachable || !hasTestApiKey)` and needs a live worker plus a test API key; my evidence is the unit harness and the route source.
- The underlying cause of the CI 500, which I did not diagnose in either review.

Thanks for the quick, precisely-scoped turnaround — deleting just the over-tight assertion and leaving the baseline contract intact is exactly the right response.

— [claude-code-eliza]

PR #30759, opened 2026-09-06
Fix the AOSP local-agent runtime and native inference

lalalune:
The system APK could boot an agent its WebView could not call, route local inference into the wrong native loader, and terminate streamed replies when the request body ended. Preserve reflected Android service methods, route IPC through the initialized API kernel, and keep response lifetime independent of request-body completion. Headless Android plugin loading, canonical app-data paths, Direct Boot handling, and atomic installation identity support the privileged image runtime.

Android Kokoro also sent raw text into a build without eSpeak, producing garbled speech. Talk Mode now obtains IPA from the authenticated local agent using the bundled phonemizer, then sends it through UDS to JNI's existing IPA synthesis API. Phonemization errors reject playback rather than silently switching to system TTS. The path explicitly supports en-US and rejects more than 508 IPA code points, reserving two padding tokens within the pinned native graph's 510-token cap. JNI consumes exact-size allocated PCM and frees it through the native API.

Native packaging uses a shared C++ runtime and verifies hashes of actual packaged bytes. The native gitlink changes exactly one commit, [1891089db](https://github.com/elizaOS/llama.cpp/commit/1891089dbed568cec30b5b9591cba9b85553e3e9), after 6543d9078051a9bb194c2ef5c2995f003c5158de. Its Vulkan guard retains the storageBuffer16BitAccess requirement, requests VK_KHR_16bit_storage when advertised, and otherwise requires its promoted Vulkan 1.1 core capability. Human inspection remains required before merge. The TTS fix adds no native fork changes.

Bun staging now pins immutable GitHub asset IDs, upstream revisions, ZIP hashes, and executable hashes for stable and canary x64/arm64 artifacts. Downloads and local overrides verify both layers; every cached executable is rehashed. Packaged provenance reports the actual version and revision. The pinned canary binaries match the archived APK17 inputs exactly.

Validation:

- Current source 96ec0f2a7385c9c5cc847959634d251cb428e1b8 includes develop 53cd4fdade. Full repository verification passed, as did all 19 staging tests, including corrupt cache/archive/executable rejection. All four real pinned upstream archives were staged and verified. APK18/image validation is underway.
- Installed APK17 at a2efe7f9a844c9828b6066bbb1f50c91b6055857 passed actual TalkMode local playback without system TTS, unsupported-language and overlength rejection, authenticated phonemization, chat/SSE completion, 401/400 controls, and explicit native failures. JNI produced 69,000 samples; independent on-device ASR returned the test sentence exactly. This is not broad perceptual speech quality evidence.
- APK17's input SHA is 70d506a6a816e52a4258691ab8ab8d42090b6b864f715054a0dce27bee2f687f; image and installed SHA is 6937824e9bb1bf8bf9d3aef3c1e3ed61013acebb592999d2c51e9ddf28a46032. All 89 runtime entries and all 1,265 uncompressed ZIP entries match across AOSP repacking/platform signing. The image APK signature verifies against the AOSP platform certificate.
- Actual persistent runtime-installation-id bytes and mode 0600 survived APK17 reboot. The service auth token rotates by design and is not installation-identity evidence. Post-reboot chat/SSE and failure controls passed; the crash buffer was empty.
- TTS route tests: 17 passed. Android bridge tests: eight passed; Java and JNI compiled for packaged targets. Production IPC tests use real AgentRuntime/startApiServer instances. Seven source-inventory tests cover generated Android asset exclusion.

Keep draft: mixed-model Vulkan synchronization stalls remain unresolved; host memory pressure is observed but not established as their cause. Signed-in UI, release-model selection, perceptual speech quality, and human native review remain outstanding. Fixtures are diagnostic inputs, not release-model approval. Raw device evidence remains local and is not independently downloadable from this PR. No images or packages are published.


ss251 (changes requested):
1bd8ad63b37369a36c0aff3d5b3f8478559a5f92

**Hold — native dependency changes are opaque in the supplied diff.** `plugins/plugin-local-inference/native/llama.cpp:1` advances the submodule from `6543d9078051a9bb194c2ef5c2995f003c5158de` to `1891089dbed568cec30b5b9591cba9b85553e3e9`. The underlying source changes are not included. This changes native code used by inference, not merely build metadata; review of the surrounding JavaScript and Java cannot establish its safety or compatibility. Have a human inspect that exact dependency range and its provenance before merging, or separate the pointer update. This is an inspection boundary, not an allegation of malicious content.

**Acceptance coverage remains incomplete.** `packages/agent/src/api/in-process-api.test.ts:25-53` supplies a custom kernel dispatcher that implements its own authentication and SSE response. It verifies the shim contract, but does not exercise `startApiServer` registration or the actual chat handler. Those tests could remain green if the real server-to-Android integration failed. Add coverage through the production kernel and head-bound device evidence for the claimed local-agent/native-inference fix.

VERIFICATION: Read the supplied diff at this head, including the IPC/authentication changes, filesystem validation, packaging changes and test assertions. The submodule contents and device artifacts were unavailable. Live execution blocked: no tools or disposable sandbox provided; this is a static review only.

Recommendation: fix

PR #30769, opened 2026-09-06
fix(openai): stop retrying exhausted provider credit

### LINKED ISSUE #30752: fix(openai): fail fast on permanent provider quota exhaustion
Live provider probes returned HTTP 429 with credit_balance_exhausted and insufficient_quota. The OpenAI-compatible text retry classifier on develop treats every 429 as transient, adding repeated requests and exponential backoff to a permanent billing failure. Scope: classify explicit permanent quota exhaustion as terminal without changing transient rate-limit handling or prompt content. Acceptance: exercise the real text transport against a deterministic HTTP failure boundary; exhausted quota makes one request and surfaces the failure, transient rate limiting retains retries, and cancellation remains honored. Include focused tests, owning package gates, and root verification. Live latency evidence is recorded locally without credentials; successful production performance remains separate from failure-path validation.

lalalune:
Live provider calls returned HTTP 429 with `insufficient_quota` / `credit_balance_exhausted`, but the text handler and AI SDK both retried them. The handler now rejects these permanent failures after one request, owns the retry budget, preserves supported provider retry delays, and cancels pending backoff when the caller aborts.

Buffered streams now observe their companion promises and propagate the original provider error before a generic SDK no-output error can hide it. Structured live streams retain and replay their complete pre-output event prefix, so the SDK start event cannot commit an attempt before the provider accepts it. Temporary failures recover before output; failures after output are propagated without replay. The structured-stream test imports its handler before timed assertions so runtime module loading does not exhaust its behavior-test deadline.

Closes #30752.

Validation:
- Real handler + AI SDK + local HTTP server: permanent failures make one request in non-streaming, live-streaming, structured live-streaming, and buffered modes; temporary limits recover; provider delays and cancellation are exercised. Structured HTTP 429/500 recovery failed on the previous PR head and passes after retaining the pre-output prefix. A midstream failure after argument output rejects without a second request.
- On the unchanged base, all six quota cases made two requests and reached the five-second test deadline. The fixed seven-case suite completed in 629 ms after loading, versus 32.05 seconds of test execution on base. This measures the failure path, not successful inference speed.
- Full package suite: 442 tests pass across 32 files. Package typecheck and lint pass.
- Live Cerebras requests through the handler and isolated local conversation API return actual synthetic replies with persisted history. These samples establish integration behavior, not a controlled performance improvement.
- Repository-wide `bun run verify`: passed, including 376 Turbo tasks and final repository audits.

Scope: text-call failure latency only. No model context, credentials, billing settings, or deployment configuration changes. Frontend screenshots and video are N/A because no UI behavior changes. Full deployed agent and microphone-to-speaker latency remain outside this patch's acceptance evidence.

### DIFF
--- a/plugins/plugin-openai/__tests__/quota-retry-http.test.ts
+++ b/plugins/plugin-openai/__tests__/quota-retry-http.test.ts
@@ -25,7 +25,10 @@ async function fixture(
   code: string,
   recover: boolean,
   retryAfter?: string,
-  streamRecovery = false
+  streamRecovery = false,
+  structured = false,
+  status = 429,
+  failAfterOutput = false
 ) {
   let requests = 0;
   const requestTimes: number[] = [];
@@ -34,9 +37,9 @@ async function fixture(
     requests++;
     requestTimes.push(performance.now());
     response.setHeader("Content-Type", "application/json");
-    if (!recover || requests === 1) {
+    if (!failAfterOutput && (!recover || requests === 1)) {
       if (retryAfter) response.setHeader("Retry-After", retryAfter);
-      response.writeHead(429);
+      response.writeHead(status);
       response.end(
         JSON.stringify({ error: { message: "Provider rejected request", type: code, code } })
       );
@@ -52,8 +55,10 @@ async function fixture(
       };
       response.end(
         [
-          `data: ${JSON.stringify({ ...chunk, choices: [{ index: 0, delta: { role: "assistant", content: "Recovered" }, finish_reason: null }] })}\n\n`,
-          `data: ${JSON.stringify({ ...chunk, choices: [{ index: 0, delta: {}, finish_reason: "stop" }], usage: { prompt_tokens: 1, completion_tokens: 1, total_tokens: 2 } })}\n\n`,
+          `data: ${JSON.stringify({ ...chunk, choices: [{ index: 0, delta: structured ? { role: "assistant", tool_calls: [{ index: 0, id: "call_test", type: "function", function: { name: "reply", arguments: '{"text":"Recovered"}' } }] } : { role: "assistant", content: "Recovered" }, finish_reason: null }] })}\n\n`,
+          failAfterOutput
+            ? `data: ${JSON.stringify({ error: { message: "Stream failed after output", type: "server_error", code: "server_error" } })}\n\n`
+            : `data: ${JSON.stringify({ ...chunk, choices: [{ index: 0, delta: {}, finish_reason: structured ? "tool_calls" : "stop" }], usage: { prompt_tokens: 1, completion_tokens: 1, total_tokens: 2 } })}\n\n`,
           "data: [DONE]\n\n",
         ].join("")
       );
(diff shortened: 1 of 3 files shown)


ss251 (approved):
b87a7e61142088a805d6d9914a9057dcfb996570

This change makes quota exhaustion terminal while keeping transient 429s retryable, and consolidates retry ownership in this layer.

`models/text.ts:isPermanentQuotaError` inspects `code`/`type` at top level, `data`, and parsed `responseBody` for `insufficient_quota` / `credit_balance_exhausted`, covering the OpenAI `{ error: { type, code } }` shape. `isTransientProviderError` now returns false for 429 when that predicate matches, so input -> exhausted credit no longer produces N retries with backoff before surfacing 429; output -> single failed call with the provider error preserved.

`providerRetryDelay` honors `retry-after-ms` / `retry-after` (seconds or HTTP-date) bounded to 0-60s with exponential-plus-jitter fallback, and `waitForTransientRetry` uses it abort-aware, so input -> `Retry-After: 0.8` no longer sleeps a fixed ~300ms but waits provider-directed delay. `generateTextByModelType:maxRetries: 0` prevents the AI SDK from retrying before the provider-aware classifier sees the response, which is the correct seam.

`consumeStreamWithTransientRetry` switching to `Promise.allSettled` for `toolCalls`/`usage`/`finishReason`/`text` preserves `capturedError` priority instead of letting a generic `NoOutputGeneratedError` from a rejected companion mask the real failure — input -> failed stream previously threw the wrong error to the classifier; output -> correct retry/no-retry decision.

`quota-retry-http.test.ts` is real-system proof, not mocked SDK: local `node:http` provider returning 429 + quota codes expects exactly 1 request in generate/live/buffered modes, 2 requests on `rate_limit_exceeded` recovery, Retry-After delay >=790ms, and abort during 30s wait issues no second request. No secrets, binaries, lockfile, workflow, or exfiltration changes.

Verified by reading the diff at this head and tracing error-classifier, backoff, and stream-companion paths; no sandbox run.

Recommendation: merge

— [muse-lane-ss251]

lalalune:
Validation for a5a49a4062c034ed5764152f6f74be17166aa7db:

- Real HTTP regression: structured-live 429/500 failures reproduced on the prior head; corrected stream retries before output and preserves the selected attempt prefix. Permanent quota makes one request; cancellation prevents a second request; errors after argument output reject without replay.
- Full package: 442 tests across 32 files, lint and typecheck pass.
- `bun install` and root `bun run verify` pass (376/376 workspace tasks plus final audits).
- Live local SSE/Cerebras call returned `Ready.` and both messages were read back from history. This is integration evidence, not a successful-call speedup claim.
- [Hosted run 34019966581](https://github.com/elizaOS/eliza/actions/runs/34019966581): all five checks SUCCESS on this head. Billing/subscription jobs passed change detection; their behavior suites were skipped for unchanged surfaces.

No Cloud Personal, deployment, or microphone-to-speaker acceptance claim.

Issue #23095, opened 2026-08-20
[cloud][billing] Expose secure subscription status, settings, audit, and lifecycle notifications


lalalune:
# [cloud][billing] Expose secure subscription status, settings, audit, and lifecycle notifications

## Summary

Expose a complete, safe organization subscription summary and lifecycle communication contract while hardening existing money/resource mutations that currently accept overly broad organization membership or API-key authority.

## Status contract

Return current plan/state, allowance used/remaining/expiry, purchased balance, total spendable, renewal date, pending cancel/downgrade, grace/dunning action, portal availability, and server-derived limits. Never expose Stripe IDs, payment details, provider configuration, or secrets.

## Authorization

- Organization members may read the shared billing summary where policy permits.
- Checkout, portal, change, cancel, billing-settings mutation, and billable-resource cancellation require session-only OWNER/ADMIN authority.
- If billing automation through API keys is retained, add a narrow explicit capability/scope; do not accept general organization keys.

## Notifications and audit

- Idempotent in-app/email notifications for payment failed, action required, retry/grace deadlines, renewal success, upgrade/downgrade/cancel scheduled/effective, and recovery.
- Stripe emails supplement but never replace application state.
- Audit actor, command, organization, provider outcome, decision, and correlation without secrets.

## Acceptance criteria

- [ ] Cross-organization/IDOR reads and mutations fail.
- [ ] MEMBER/GUEST/general API key cannot mutate subscription or billing resources.
- [ ] OWNER/ADMIN authorization is rechecked at final mutation boundary.
- [ ] Stale role/session/CSRF/replay attempts fail without provider calls.
- [ ] Status distinguishes loading/unavailable/dunning from healthy zero values.
- [ ] Notifications dedupe, escalate, and resolve correctly.
- [ ] DTO/log/telemetry secret scans pass.

## Tests and evidence

- Authorization matrix, IDOR, stale membership, CSRF, replay, provider-unavailable, DTO snapshots, notification dedupe/escalation/recovery, and audit-redaction tests.
- Real two-user organization evidence showing member read-only and owner/admin management.

## Dependencies

Depends on the lifecycle and entitlement APIs; consumed by the Billing UI.

## Sprint 3 proposed slice

[...]

PR #30771, opened 2026-09-06
fix(cloud): secure billing settings and expose mail submission receipts

NubsCarson:
Billing settings now require current organization manager authority for updates, recheck it before persistence, and save auto-top-up and earnings settings together. Missing organizations return a sanitized HTTP 503 for both GET and PUT. Empty or unknown-only updates return persisted settings without a write, timestamp change, or cache invalidation.

The mail boundary exposes typed unavailable, accepted, rejected, and uncertain submission receipts while preserving legacy send behavior. Provider acceptance is distinct from recipient delivery; uncertain submission is never treated as a reason to resend automatically.

References #23095. This prerequisite does not close the issue. Coherent subscription status, durable notices, policy approval and real recipient/deployed acceptance remain separate work.

Validation targets `65fe8b2ef6752c5d679ae5863790efae7bdc29ff`, rebased onto `65a4f5216421ba74a75b1bc9bc98de0ae2d1ceb1`. All three issue patches are range-diff identical to independently reviewed `d2c4e156d349d045e9ba9dac96ffa962ba80405f`.

- Pinned Bun 1.3.14 install passed on the current head.
- Independent review on d2c4e156 found no actionable P1/P2 issue: 105 tests passed with 363 assertions. The four API files ran in separate processes, matching the package's canonical runner. This is source-equivalent review evidence, not a new-head execution claim.
- Root independent current-head review found no actionable finding after inspecting the rebased repair and unchanged affected base paths. Current-head verification passed: 77 service tests / 271 assertions, 9 mail tests / 27 assertions, and all four API files / 19 tests / 65 assertions (105 tests, 363 assertions total).
- Current-head root verification passed all 376 tasks and subsequent repository audits, exit 0. The full owning API suite passed all 598 files in isolated Bun processes, exit 0.
- Historical full API sweep on 494f811a passed all 597 files. That sweep predates the additional PUT/no-op repair and is not claimed as full-package proof of the repair.
- Current hosted run34023755788 passed source static smoke, payment replay, PostgreSQL authority, Windows security and the final aggregate. The superseded push-triggered run34023749259 was canceled when readiness triggered this run; its failed aggregate is not a source-test failure.

The reported missing-organization PUT and unnecessary no-op persistence findings are fixed. The mock-isolation review concern is addressed by the canonical isolated API runner and independent execution of all four affected route files. No external email, Stripe mutation or deployment was performed.

## Evidence Gate

N/A - no rendered UI change.

N/A - no rendered UI change.

N/A - HTTP and mail adapter changes with no interactive UI surface.

<details><summary>Independent execution on source-equivalent d2c4e156</summary>

```text
Settings service: 77 pass, 0 fail, 271 assertions
API authorization: 7 pass, 0 fail, 25 assertions
API availability: 5 pass, 0 fail, 23 assertions
API malformed JSON: 2 pass, 0 fail, 5 assertions
API settings: 5 pass, 0 fail, 12 assertions
Mail receipt: 9 pass, 0 fail, 27 assertions
Total: 105 pass, 0 fail, 363 assertions
```

Each API file ran in its own Bun process. Service and mail suites also ran separately.
</details>

N/A - no frontend change.

N/A - no model behavior change.

Inspected real service, HTTP and loopback SMTP outcomes are included below.

<details><summary>Inspected boundary outcomes</summary>

Real settings-service tests verify typed unavailable on missing organization and no persistence/cache invalidation for empty updates. HTTP tests inspect sanitized 503 responses and manager/session/tenant authorization. Loopback SMTP tests inspect accepted submission, disconnect after submission, rejection and partial recipient acceptance. SMTP acceptance is not delivery proof; SendGrid response tests use a deterministic adapter fixture.

Observed test output:
```text
(pass) real getSettings missing organization uses advertised unavailable response [2.69ms]
(pass) real updateSettings missing organization returns PUT unavailable without mutations [1.22ms]
(pass) PUT with no recognized changes preserves storage and caches: {} [0.38ms]
(pass) EmailService durable submission receipt > records SMTP acceptance without claiming recipient delivery [15.15ms]
(pass) EmailService durable submission receipt > disconnect after submission remains uncertain and is never retried [3.84ms]
(pass) EmailService durable submission receipt > partial recipient acceptance cannot claim one complete receipt [3.11ms]
```

Current-head root gate:
```text
[anti-larp] scanned 11458 test files — 0 focused, 0 orphaned skip(s)
exit 0
```
</details>

N/A - no rendered text.

### DIFF
--- a/packages/cloud/api/v1/billing/settings/route.ts
+++ b/packages/cloud/api/v1/billing/settings/route.ts
@@ -24,6 +24,7 @@ import {
 import {
   AUTO_TOP_UP_LIMITS,
   AutoTopUpSettingsPolicyError,
+  AutoTopUpSettingsUnavailableError,
   AutoTopUpSettingsValidationError,
   autoTopUpService,
 } from "@/lib/services/auto-top-up";
@@ -89,6 +90,16 @@ app.get("/", rateLimit(RateLimitPresets.STANDARD), async (c) => {
     });
   } catch (error) {
     // error-policy:J1 transport boundary returns a sanitized failure.
+    if (error instanceof AutoTopUpSettingsUnavailableError) {
+      return failureResponse(
+        c,
+        new ApiError(
+          503,
+          "service_unavailable",
+          "Billing settings are unavailable",
+        ),
+      );
+    }
     logger.error("[Billing Settings API] Error getting settings:", error);
     return failureResponse(c, error);
   }
@@ -184,6 +195,16 @@ app.put("/", moneyRateLimit(RateLimitPresets.STANDARD), async (c) => {
       return c.json(
         { success: false, error: error.message, code: "validation_error" },
         400,
+      );
+    }
+    if (error instanceof AutoTopUpSettingsUnavailableError) {
+      return failureResponse(
+        c,
+        new ApiError(
+          503,
+          "service_unavailable",
+          "Billing settings are unavailable",
+        ),
       );
     }
     logger.error("[Billing Settings API] Error updating settings:", error);
--- a/packages/cloud/api/v1/billing/settings/route.ts
+++ b/packages/cloud/api/v1/billing/settings/route.ts
@@ -1,19 +1,21 @@
 /**
- * Billing Settings API (v1)
- *
- * GET/PUT /api/v1/billing/settings
- * Manage auto-top-up and billing settings.
- *
- * Auto-top-up powers autonomous-agent continuity: when balance falls below
- * threshold, the saved Stripe payment method is charged. Configuring this via
- * API lets agents/integrators tune their own billing without the dashboard.
+ * Reads tenant billing settings and admits session-only manager changes.
+ * The service rechecks current authority after validation and persists the
+ * auto-top-up and earnings changes together before invalidating billing caches.
  */
 
 import { Hono } from "hono";
 import { z } from "zod";
 import { organizationsRepository } from "@/db/repositories";
-import { failureResponse } from "@/lib/api/cloud-worker-errors";
-import { requireUserOrApiKeyWithOrg } from "@/lib/auth/workers-hono-auth";
+import {
+  ApiError,
+  ForbiddenError,
+  failureResponse,
+} from "@/lib/api/cloud-worker-errors";
+import {
+  requireCurrentBillingManagerSession,
+  requireUserOrApiKeyWithOrg,
+} from "@/lib/auth/workers-hono-auth";
 import {
   moneyRateLimit,
   RateLimitPresets,
@@ -21,6 +23,7 @@ import {
 } from "@/lib/middleware/rate-limit-hono-cloudflare";
 import {
   AUTO_TOP_UP_LIMITS,
+  AutoTopUpSettingsPolicyError,
   AutoTopUpSettingsValidationError,
   autoTopUpService,
 } from "@/lib/services/auto-top-up";
@@ -58,6 +61,14 @@ app.get("/", rateLimit(RateLimitPresets.STANDARD), async (c) => {
       organizationsRepository.findById(user.organization_id),
     ]);
 
+    if (!org) {
+      throw new ApiError(
+        503,
+        "service_unavailable",
+        "Billing settings are unavailable",
+      );
+    }
+    c.header("Cache-Control", "no-store");
     return c.json({
       success: true,
       settings: {
@@ -67,7 +78,7 @@ app.get("/", rateLimit(RateLimitPresets.STANDARD), async (c) => {
           threshold: autoTopUpSettings.threshold,
           hasPaymentMethod: autoTopUpSettings.hasPaymentMethod,
         },
-        payAsYouGoFromEarnings: org?.pay_as_you_go_from_earnings ?? true,
+        payAsYouGoFromEarnings: org.pay_as_you_go_from_earnings,
         limits: {
           minAmount: AUTO_TOP_UP_LIMITS.MIN_AMOUNT,
           maxAmount: AUTO_TOP_UP_LIMITS.MAX_AMOUNT,
@@ -77,6 +88,7 @@ app.get("/", rateLimit(RateLimitPresets.STANDARD), async (c) => {
       },
     });
   } catch (error) {
+    // error-policy:J1 transport boundary returns a sanitized failure.
     logger.error("[Billing Settings API] Error getting settings:", error);
     return failureResponse(c, error);
   }
@@ -84,7 +96,7 @@ app.get("/", rateLimit(RateLimitPresets.STANDARD), async (c) => {
 
 app.put("/", moneyRateLimit(RateLimitPresets.STANDARD), async (c) => {
   try {
-    const user = await requireUserOrApiKeyWithOrg(c);
+    const user = await requireCurrentBillingManagerSession(c);
 
     const decodedBody = await decodeRequestJson(c.req);
     if (!decodedBody.ok) {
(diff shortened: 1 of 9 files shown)


packages/cloud/api/v1/billing/settings/route.ts
```diff
@@ -58,6 +61,14 @@ app.get("/", rateLimit(RateLimitPresets.STANDARD), async (c) => {
       organizationsRepository.findById(user.organization_id),
     ]);
 
+    if (!org) {
```
NubsCarson, line 65: [P2] Handle the real settings-service failure before the unavailable guard. When the organization is absent, getSettings rejects with Organization not found, so Promise.all rejects before this branch executes. A review test using the real route and real AutoTopUpService with an absent repository row returns 404/resource_not_found instead of the advertised 503/service_unavailable. The new missing-organization test misses this because it stubs getSettings to healthy values. Translate a typed missing-settings failure (or use a coherent single organization read) and cover this actual service-to-route boundary, including the post-update readback.

NubsCarson (commented):
Reviewed exact 771249c7c364a8c60c2378b121711ec20a9d365c against b676280384b15df54204276e34bb8740db26936f. Independently reproduced all 113 focused tests, both package typechecks and lint checks, and API route parity/Worker dry-run. No root verification rerun; the documented root failure and deployed two-user/recipient evidence remain unverified by this review. The scope statement accurately leaves full subscription status, approved notice policy, durable outbox/audit and controlled delivery acceptance open. One unavailable-state finding below prevents approval of this revision.

NubsCarson (commented):
Independent follow-up review of 494f811a350a7ae1c7ad9473697a298af5fc4843 (local and published head verified). I did not author the fix. The earlier P2 missing-organization finding is addressed: real getSettings now throws a typed unavailable error, and the GET/post-write readback boundaries translate it to the safe 503/service_unavailable response.

Current-head verification: all 15 route tests pass in four isolated processes; 75 auto-top-up service tests pass; API/shared typechecks and lint:check pass; API verification includes 709 route mounts and Worker bundle dry-run. An additional temporary real-service test supplied an organization for the update then removed it before readback: it returned exactly the three safe error fields with 503, after one settings write and the expected cache invalidation. Both that test and the permanent real-service GET regression passed. Temporary review code was removed; the checkout is clean.

Cache/redaction review: isolated error responses do not themselves add Cache-Control, but production createApp's existing JSON-response middleware (bootstrap-app.ts:399–412) supplies no-store when absent, including these application/json 503s. The typed error's organization context does not enter the HTTP response. This middleware conclusion is source-verified, not a deployed browser/proxy receipt.

No further blocking code finding in this prerequisite slice. Prior mail/auth review remains applicable to unchanged implementation. This does not close the full subscription-status/notices issue or clear root verification, hosted checks, deployed two-user acceptance, or controlled delivery evidence; those gates remain separate.

OutstandingRetard (approved):
Scoped review of #30771 (fix(cloud): secure billing settings and expose mail submission receipts).
Touches 8 file(s): `packages/cloud/api/v1/billing/settings/route.authorization.test.ts`, `packages/cloud/api/v1/billing/settings/route.availability.test.ts`, `packages/cloud/api/v1/billing/settings/route.json.test.ts`, `packages/cloud/api/v1/billing/settings/route.test.ts`. Lead change `packages/cloud/api/v1/billing/settings/route.authorization.test.ts` (+219/-0).
No blocking concerns.

ss251 (changes requested):
9284a7444410e1948f929a3a95fec2461f1eef53

[P2] Isolate the conflicting module mocks so the “real” boundary tests exercise production code. `packages/cloud/api/v1/billing/settings/route.authorization.test.ts:59-87` globally replaces `@/lib/services/auto-top-up`, while `packages/cloud/api/v1/billing/settings/route.availability.test.ts:96-105` globally replaces the authorization helpers with unconditional success. Both files dynamically import the same route; neither establishes module/process isolation.

When these files share a Bun test process, those module replacements can cross file boundaries. In particular, missing organization + the mocked `getSettings` can still produce the expected 503 through the route’s separate organization check (`packages/cloud/api/v1/billing/settings/route.ts:65-71`). Thus the availability test can pass without exercising the real service failure it claims to cover. Conversely, the unconditional authorization replacement undermines the current-authority tests. Run these fixtures in explicitly separate processes, or restructure their dependency boundaries so each test demonstrably invokes the intended real implementation.

The combined billing write and mail receipt changes are behavioral, not cosmetic, but these tests should not be presented as reliable integration evidence until that isolation is established.

VERIFICATION: Statically traced the supplied diff’s mock registrations, route imports, and missing-organization response paths at this head. No runtime result or independent acceptance artifact was verified. Live execution blocked: no tools or disposable sandbox available, so this is a static review of the supplied diff.

Recommendation: fix

attentionhead (approved):
**APPROVE.** This is a money-path authorization change, so I focused on whether the new authority model is real and whether the controls it adds are actually *pinned* — not on the volume of added tests. All three are pinned by killing mutants. One non-blocking note at the end.

## The red check is not yours

`All Tests Passed` is red only because the admission gate saw `source-smoke finished with cancelled`, with four sibling lanes `CANCELLED` by supersession. No test failed.

## Authority model — verified, not assumed

- `requireCurrentBillingManagerSession` exists at `workers-hono-auth.ts:436` (it isn't in this diff, so a missing symbol would have been a build break). It goes through `requireSessionUserWithOrg`, revalidates the manager credential, reads primary membership via `findWithOrganizationForWrite`, and on lookup failure raises a typed **503** rather than denying — correctly separating *unavailable* from *denied* instead of failing open. Moving PUT onto it means **API keys can no longer mutate billing**.
- `await authorizeMutation()` sits immediately before `organizationsRepository.update`, after every async validation — so it's a genuine pre-persistence TOCTOU recheck, not a callback that's merely threaded through.
- Replacing the substring classification (`message.includes("Cannot enable") || "must be" || "cannot exceed"`) with typed `PolicyError` / `ValidationError` / `UnavailableError` removes a real hazard: any reworded domain message silently became a 500.
- Dropping `org?.pay_as_you_go_from_earnings ?? true` for an explicit 503 is the right call under the DTO rule — a missing org was previously reported as a healthy `true`.

## The risk I went looking for, and why it's safe

`updateSettings` is now called **unconditionally**, where the old route guarded it behind `if (autoTopUp)`. That's the kind of change that clobbers fields on a partial update, so I checked the implementation rather than trusting the shape:

- `mustValidateAmounts` is true only when `enabled === true` or `amount`/`threshold` are supplied, so an **earnings-only** update skips amount validation and can't produce a spurious 400.
- Every persisted field is guarded by an explicit `!== undefined`, so absent auto-top-up fields are never written.

No clobbering. Good.

## Mutation results — all three controls pinned

| mutant | result |
|---|---|
| revert PUT authority to `requireUserOrApiKeyWithOrg` (re-admit API keys) | **3 failed** (`route.authorization.test.ts`) |
| delete the pre-persistence `await authorizeMutation()` | **2 failed** (`auto-top-up.test.ts`) |
| remove the `!org` 503 guards **and** restore `?? true` | **1 failed** (`route.authorization.test.ts`) |

Per-file at head: 5 + 7 + 1 + 2 + 9 + 75 = **99 tests passing**, matching your body's claim exactly.

Two corrections to my own process, in case the numbers get compared: my first run used vitest and failed to collect (`Cannot find package 'bun:test'`) — wrong runner, my fault; and batching six files into one `bun test` process reported 7 failures that all vanish when the files run individually (`auto-top-up.test.ts` alone is 73/73 on develop, 75/75 here). Those were interference in my invocation, not defects.

## Email receipt — the non-leakage claim is structural

`EmailDispatchResult` is a closed union of literal statuses, reasons and provider names plus a `messageId`, and `receiptId` whitelists `^[A-Za-z0-9._:@<>+=/-]{1,256}$`, returning **null rather than truncating** an over-long identifier — consistent with the repo's no-truncation rule. No recipient, body, or raw provider error string can reach a receipt. The type makes the claim true rather than merely asserting it.

## Non-blocking

Both schema fields are optional, so an empty `{}` PUT now validates, runs `authorizeMutation`, and performs an `organizationsRepository.update` with only `updated_at` plus cache invalidation. Under the old code an empty body took neither branch and wrote nothing. It's behind a manager session and `moneyRateLimit` so it isn't an abuse vector of consequence — just a newly introduced write-and-invalidate on a no-op request. A guard for "no recognised fields" would restore the old behaviour if you think it's worth it.

Nothing blocking. Careful, well-tested change on a path where fail-open would be expensive.

— [claude-code-eliza]

attentionhead (approved):
## APPROVE — re-review at the new head; the incremental change is three files and both additions are covered

I approved this PR at head `9284a744` (review 5124615791); it went to draft and has returned ready-for-review at `65fe8b2e`. Re-reviewed the delta rather than the whole PR.

**Scoping the delta.** The raw diff against my previous head touches 98 files, but that is develop drift from a rebase. Restricted to this PR's own nine files, only three moved:

```
route.availability.test.ts               +74
shared/.../__tests__/auto-top-up.test.ts +17
shared/.../services/auto-top-up.ts       +10/-3
```

All six suites pass locally under `bun test`: **105 pass / 0 fail** (7 + 5 + 2 + 5 + 77 + 9).

### The one thing worth checking carefully

The new early return sits **before** `authorizeMutation()`:

```ts
const organization = await organizationsRepository.findById(organizationId);
if (!organization) throw new AutoTopUpSettingsUnavailableError(organizationId);
if (settings.enabled === undefined && settings.amount === undefined &&
    settings.threshold === undefined && settings.payAsYouGoFromEarnings === undefined) {
  return;
}
```

So an empty payload never runs the authority recheck, and a missing organization is reported before it too. I wanted to rule out both an authorization bypass and a cross-tenant existence oracle, since `AutoTopUpSettingsUnavailableError` now maps to a distinct 503 and would make such an oracle sharper than the old generic `Error` did.

Neither applies, and the reason is in the route rather than in this function — `packages/cloud/api/v1/billing/settings/route.ts`:

- `requireCurrentBillingManagerSession(c)` runs at **:110**, before `updateSettings` is ever called, so the caller is already authorized on entry.
- The organization id passed at **:143** is `user.organization_id` — derived from the session, never from client input. There is no attacker-chosen org to probe.
- `authorizeMutation` is only the TOCTOU recheck (`current.id !== user.id || current.organization_id !== user.organization_id`) placed immediately before persistence.

Skipping a *recheck* when there is nothing to persist is correct. And the early return is a small improvement on its own: previously an empty PUT still reached `organizationsRepository.update(organizationId, { updated_at: new Date() })` and fired the follow-up invalidation and `onOrganizationUpdated` side effects for a request that changed nothing.

### Verified

| mutant | result |
|---|---|
| N1 — drop the empty-payload early return | **killed** (101/4) |
| N2 — revert the `updateSettings` typed `Unavailable` error to a generic `Error` | **killed** (103/2) |
| N2b — same revert on the `getSettings` side (line 1569) | **killed** (104/1) |
| N3 — early return ignores `payAsYouGoFromEarnings` | **survives** |

A note on method, since it changed a result: my first N2 attempt did not apply — the guard asserted a single occurrence and there are two — so the "no failures" reading it produced was void, not a survival. Re-run per line, both sites die.

The new `route.availability.test.ts` cases map exactly onto N1 and N2: a missing organization returns **503** with no `updateOrganization`, no `invalidateOrganizationCache`, no `onOrganizationUpdated`; and an empty payload returns **200** with the same three negative assertions. That is the right shape — asserting the absence of the side effects, not just the status code.

**I chased the survivor.** `payAsYouGoFromEarnings` appears exactly twice in the auto-top-up suite, both times alongside `enabled`/`amount`/`threshold` — it is never sent on its own, which is precisely why the fourth clause of the guard is unconstrained. Probe:

```
payAsYouGoFromEarnings-only payload, at head -> 78 pass / 0 fail
  update persists pay_as_you_go_from_earnings: true, authorize called once
same probe with N3 applied                   -> 77 pass / 1 fail
  the payload is silently dropped
```

So the clause is correct and load-bearing — without it, a user toggling only "pay as you go from earnings" would get a 200 and no change. It is simply unasserted.

**CI red is not this PR.** `All Tests Passed` fails after 2s on the superseded run 34023749259 with `RESULTS: source-smoke=cancelled browser-bridge-windows-security=cancelled billing-payment-replay-e2e=cancelled subscription-authority-postgres=cancelled`. No test failed. On the live run 34023755788, Billing payment replay E2E and Subscription authority PostgreSQL pass, with source-smoke and the Windows contract still pending.

### Not verified

- Everything I checked at `9284a744` — the `requireCurrentBillingManagerSession` split, the `authorizeMutation` placement immediately before `organizationsRepository.update`, and the `email.ts` submission-receipt work — I re-ran but did not re-derive; that analysis stands from the earlier review.
- The two pending lanes had not reported when I finished.

### Non-blocking observation

**Add the `payAsYouGoFromEarnings`-only case.** It is a one-line payload change on an existing test and it is the only clause of the new guard that nothing holds. The probe above is small enough to lift directly. Worth doing because the failure mode is silent: a 200 response and no persisted change is the kind of bug that reaches users as "the toggle doesn't stick" rather than as an error.

Approving. The delta is small, the early return is safe for reasons I confirmed in the route rather than assumed, and the new availability tests assert the absence of side effects rather than only the status.

— [claude-code-eliza]

Issue #23094, opened 2026-08-20
[cloud][billing] Derive every enforced rate and resource quota from the subscription entitlement projection


lalalune:
# [cloud][billing] Derive every enforced rate and resource quota from the subscription entitlement projection

## Summary

Make one server entitlement service authoritative for inference rate limits, resource ceilings, storage, API snapshots, and every create/start/resume boundary. Eliminate subscriber limit flapping caused by remaining purchased-credit balance.

## Implementation

- Active/grace subscription projection overrides legacy cumulative-credit classification; audited manual overrides remain highest authority.
- Cache payloads carry entitlement and catalog versions; lifecycle writes invalidate/hydrate them.
- Refactor inference rate limits and canonical cloud-character, agent-sandbox, container, app, and storage quota helpers.
- Make create/start/resume enforcement and `/billing/limits` consume the same service result.
- Include all four RPM fields and explicit unavailable/error state.
- Downgraded over-limit resources remain visible; block new create/start/resume instead of deleting or silently stopping.
- Cache unavailable/warming remains fail-closed for inference.

## Acceptance criteria

- [ ] Free, Plus, Pro, grace, past-due, unpaid/canceled, and manual override states return exact declared limits.
- [ ] A subscriber's limits do not change as allowance or purchased balance is spent.
- [ ] Every enforcement path equals the public limit snapshot.
- [ ] Stale cache versions cannot override newer entitlement state.
- [ ] Downgrade over-limit behavior is consistent across every resource kind.
- [ ] Reads do not mutate or reconstruct entitlement ad hoc.
- [ ] Admin overrides are bounded, audited, versioned, and cache-invalidating.

## Tests and evidence

- State/precedence truth tables and route/helper parity tests.
- Concurrent webhook plus create/start/resume, stale cache, cache outage, downgrade, grace expiry, and manual override tests.
- Real API evidence showing exact Plus/Pro headers and resource enforcement before and after renewal/downgrade.

## Dependencies

Depends on catalog, persistence, split funding, and lifecycle reconciliation.

## Sprint 3 proposed slice

[...]

PR #30772, opened 2026-09-06
fix(cloud): bind entitlement publication to current subscription authority

NubsCarson:
# Relates to

Prerequisite for #23094; does not close it.

# Background

Entitlement publication now verifies the organization's canonical subscription identity and current immutable lifecycle revision before replay or projection CAS. A historical subscription cannot overwrite or replay the current account's projection. Grace uses its actual deadline; canceled/free projections retain the transition that produced them.

A canonical account association is owned by the existing lifecycle repository. New identities require the command-captured expected identity. Migration chooses the unique live subscription or sole historical identity; ambiguous terminal history becomes unavailable rather than being ordered by timestamps. Irreversible account erasure releases the association in the same transaction before deleting history.

# Sync with develop

Head `09d3fa484a47f4fc3beaa6f19b7dc1498c9afe44`, base `cd821540d73c4e7370a9407c07be19020a3b810b`. The prerequisite patch remains unchanged after rebase. Pinned Bun 1.3.14 install passed. Exact-head root `bun run verify` passed 376/376 tasks and all subsequent repository audits, exit 0.

Live merge-target readback at evidence publication: `65a4f5216421ba74a75b1bc9bc98de0ae2d1ceb1`. That later develop tip is not mislabeled as the base used for the local root run; GitHub reports the exact PR head conflict-free. No bypass is requested.

# Risks

This adds a billing-authority migration. Ambiguous existing terminal history requires reconciliation; no guessed identity is chosen. Paid resource ceilings remain nullable, not unlimited. Full enforcement, generic trial/payer/access policy and real Dedicated continuity are outside this prerequisite.

No production migration, deployed provider or money operation is claimed. Local PGlite tests exercise real database constraints and transaction rollback; they do not substitute for independent PostgreSQL sessions. All five exact-head hosted checks passed in [run 34020529015](https://github.com/elizaOS/eliza/actions/runs/34020529015), including the real PostgreSQL authority job, billing replay E2E, Windows security, source smoke and aggregate gate. A readiness-triggered run, if created, must also finish before merge.

No rendered UI or public endpoint changes. Remaining full-issue acceptance is tracked in #23094.

# Testing

- Root independent exact-head review: no actionable finding; four owning PGlite/derivation files passed 17 tests, 93 assertions.
- Author exact-head verification: seven files passed 34 tests, 168 assertions; six local PostgreSQL suite entries skipped because no DSN is available. This is not labeled independent review.
- Exact-head full shared package rerun: `bun run --cwd packages/cloud/shared test`, exit 0. All 348 sequential child processes completed over 1,181 files: **12,566 passed, 209 skipped, 0 failed; 80,529 assertions**. These totals are the sum of all 348 terminal Bun summaries, not a partial or mixed-state sweep.
- The first exact-head full attempt stopped at ordinary batch 25 on an unrelated restore-v3 timing assertion: the test expected `AgentBackupRestoreV3StreamError` but its operation deadline produced `AgentBackupRestoreV3ControlError` after 276 ms. No source changes were made to mask it. The same 16-file batch passed in isolation (256 tests, 1,928 assertions), then the full exact-head rerun above passed. The initial failure remains disclosed; neither isolated success nor the older mixed-state run is substituted for the complete rerun.
- Hosted exact-head [Subscription authority PostgreSQL](https://github.com/elizaOS/eliza/actions/runs/34020529015/job/101452155504) passed with independent database sessions; local missing-DSN skips are not represented as that proof.

# Evidence Gate

N/A - no rendered UI change.

N/A - no rendered UI change.

N/A - database authority and migration; no interactive UI flow.

<details><summary>Independent current-head execution</summary>

```text
bun test v1.3.14 (d1632b29)
(pass) current-source entitlement publication > an old request cannot replay a projection after lifecycle authority advances [45.18ms]
17 pass
0 fail
93 expect() calls
Ran 17 tests across 4 files. [6.26s]
```

</details>
<details><summary>Exact-head repository gate</summary>

```text
$ bun run verify
[anti-larp] scanned 11452 test files — 0 focused, 0 orphaned skip(s)
exit 0
```

</details>
<details><summary>Full shared package and first-attempt disclosure</summary>

```text
$ bun run --cwd packages/cloud/shared test
[run-bun-tests] sequential process sharding: 1181 file(s), 348 child process(es)
348 terminal child summaries aggregated:
12566 pass
209 skip
0 fail
80529 expect() calls
process exit 0

First attempt: ordinary batch25 stopped on an existing restore-v3 operation-deadline assertion.
Isolated unchanged batch25: 256 pass, 0 fail, 1928 expect() calls, exit0.
Complete unchanged-head rerun: totals above, exit0.
```

</details>

N/A - no frontend flow changed.

N/A - no model behavior changed.

Real migration/readback and repository tests verify ambiguous history, tenant constraints, source provenance, lifecycle freshness and erasure fencing. Selected observed current-head output:

<details><summary>Persisted authority and rollback outcomes</summary>

```text
(pass) subscription account identity migration > backfills live identity without chronology and keeps missing history distinct from ambiguity [926.16ms]
(pass) subscription account identity migration > a failed migration transaction rolls back both the association and backfill [651.01ms]
(pass) current-source entitlement publication > an old request cannot replay a projection after lifecycle authority advances [45.18ms]
(pass) current-source entitlement publication > tenant mismatch cannot publish another organization's subscription [20.38ms]
(pass) current-source entitlement publication > account erasure releases the reverse pointer only behind the irreversible fence [26.50ms]
(pass) current-source entitlement publication > retains explicit replacement canceled identity even with equal creation timestamps [36.19ms]
```

</details>

N/A - no rendered text.

### DIFF
--- a/packages/cloud/shared/src/db/account-deletion-foreign-key-policy.test.ts
+++ b/packages/cloud/shared/src/db/account-deletion-foreign-key-policy.test.ts
@@ -27,7 +27,7 @@ describe("account deletion full-schema foreign-key policy", () => {
       .update(descriptors.map(serializeDescriptor).join("\n"))
       .digest("hex");
 
-    expect(descriptors).toHaveLength(253);
+    expect(descriptors).toHaveLength(254);
     expect(digest).toBe(ACCOUNT_DELETION_FOREIGN_KEY_SNAPSHOT_SHA256);
   });
 
@@ -37,7 +37,7 @@ describe("account deletion full-schema foreign-key policy", () => {
       action: classifyAccountDeletionForeignKey(descriptor),
     }));
 
-    expect(classified).toHaveLength(253);
+    expect(classified).toHaveLength(254);
     expect(classified.every(({ action }) => Boolean(action))).toBe(true);
     expect(classified.filter(({ action }) => action === "reconcile_external_resource").length).toBe(
       75,
--- a/packages/cloud/shared/src/db/account-deletion-foreign-key-policy.ts
+++ b/packages/cloud/shared/src/db/account-deletion-foreign-key-policy.ts
@@ -23,9 +23,9 @@ export interface AccountDeletionForeignKeyDescriptor {
   targetColumns: string;
   onDelete: string;
 }
-/** SHA-256 of the 253 sorted direct user/organization FK descriptors. */
+/** SHA-256 of the 254 sorted direct user/organization FK descriptors. */
 export const ACCOUNT_DELETION_FOREIGN_KEY_SNAPSHOT_SHA256 =
-  "773c62bb7a60026260f8b4f90f58a7cdebf18fbf9a7ad66ab972c5cac4365af7";
+  "7cb2ab72048446d21a64edfde9e8a67d2000d104685e84152dc8fdc29d850ea7";
 
 function serializeDescriptor(descriptor: AccountDeletionForeignKeyDescriptor): string {
   return [
--- a/packages/cloud/shared/src/db/migrations/0379_subscription_account_authority.sql
+++ b/packages/cloud/shared/src/db/migrations/0379_subscription_account_authority.sql
@@ -1,0 +1,33 @@
+-- Retain account subscription identity through terminal lifecycle transitions.
+ALTER TABLE organizations ADD COLUMN IF NOT EXISTS subscription_authority_id uuid;
+ALTER TABLE organizations ADD COLUMN IF NOT EXISTS subscription_authority_state text NOT NULL DEFAULT 'none';
+--> statement-breakpoint
+-- The unique live-subscription constraint establishes identity without timestamp ordering.
+UPDATE organizations o SET subscription_authority_id = s.id, subscription_authority_state = 'current'
+FROM billing_subscriptions s
+WHERE s.organization_id = o.id AND s.status IN ('pending','incomplete','active','grace','past_due','unpaid');
+--> statement-breakpoint
+-- A single historical source is unambiguous; multiple terminal sources require reconciliation.
+UPDATE organizations o SET subscription_authority_id = s.id, subscription_authority_state = 'current'
+FROM billing_subscriptions s
+WHERE s.organization_id = o.id AND o.subscription_authority_state = 'none'
+AND NOT EXISTS (SELECT 1 FROM billing_subscriptions other WHERE other.organization_id = o.id AND other.id <> s.id);
+UPDATE organizations o SET subscription_authority_state = 'unavailable'
+WHERE o.subscription_authority_state = 'none' AND EXISTS (SELECT 1 FROM billing_subscriptions s WHERE s.organization_id = o.id);
+--> statement-breakpoint
+ALTER TABLE organizations ADD CONSTRAINT organizations_subscription_authority_tenant_fk
+FOREIGN KEY (subscription_authority_id, id) REFERENCES billing_subscriptions (id, organization_id) ON DELETE RESTRICT;
+ALTER TABLE organizations ADD CONSTRAINT organizations_subscription_authority_check CHECK (
+  (subscription_authority_state = 'current' AND subscription_authority_id IS NOT NULL)
+  OR (subscription_authority_state IN ('none','unavailable') AND subscription_authority_id IS NULL)
+);
+--> statement-breakpoint
+-- Terminal projections retain the authoritative transition that produced Free.
+ALTER TABLE organization_entitlements DROP CONSTRAINT organization_entitlements_plan_state_check;
+ALTER TABLE organization_entitlements ADD CONSTRAINT organization_entitlements_plan_state_check CHECK (
+  plan_key IN ('free','plus_monthly','pro_monthly') AND state IN ('free','active','grace','past_due','unpaid')
+  AND ((plan_key = 'free' AND state = 'free' AND entitlement_effective
+        AND (source_subscription_id IS NULL) = (source_subscription_revision IS NULL))
+    OR (plan_key <> 'free' AND state <> 'free' AND source_subscription_id IS NOT NULL
+        AND source_subscription_revision IS NOT NULL AND (entitlement_effective = (state IN ('active','grace')))))
+);
--- a/packages/cloud/shared/src/db/migrations/0379_subscription_account_authority.sql
+++ b/packages/cloud/shared/src/db/migrations/0379_subscription_account_authority.sql
@@ -1,26 +1,37 @@
--- Retain account subscription identity through terminal lifecycle transitions.
-ALTER TABLE organizations ADD COLUMN IF NOT EXISTS subscription_authority_id uuid;
-ALTER TABLE organizations ADD COLUMN IF NOT EXISTS subscription_authority_state text NOT NULL DEFAULT 'none';
+-- Retain account identity without adding a reverse foreign key to the organization schema.
+CREATE TABLE IF NOT EXISTS organization_subscription_authorities (
+  organization_id uuid PRIMARY KEY REFERENCES organizations(id) ON DELETE CASCADE,
+  subscription_id uuid,
+  state text NOT NULL DEFAULT 'none',
+  CONSTRAINT organization_subscription_authorities_tenant_fk FOREIGN KEY (subscription_id, organization_id)
+    REFERENCES billing_subscriptions(id, organization_id) ON DELETE RESTRICT,
+  CONSTRAINT organization_subscription_authorities_state_check CHECK (
+    (state = 'current' AND subscription_id IS NOT NULL)
+    OR (state IN ('none','unavailable') AND subscription_id IS NULL))
+);
+INSERT INTO organization_subscription_authorities (organization_id) SELECT id FROM organizations;
 --> statement-breakpoint
 -- The unique live-subscription constraint establishes identity without timestamp ordering.
-UPDATE organizations o SET subscription_authority_id = s.id, subscription_authority_state = 'current'
+UPDATE organization_subscription_authorities a SET subscription_id = s.id, state = 'current'
 FROM billing_subscriptions s
-WHERE s.organization_id = o.id AND s.status IN ('pending','incomplete','active','grace','past_due','unpaid');
+WHERE s.organization_id = a.organization_id AND s.status IN ('pending','incomplete','active','grace','past_due','unpaid');
 --> statement-breakpoint
--- A single historical source is unambiguous; multiple terminal sources require reconciliation.
-UPDATE organizations o SET subscription_authority_id = s.id, subscription_authority_state = 'current'
+-- A sole historical identity is unambiguous; multiple terminal sources require reconciliation.
+UPDATE organization_subscription_authorities a SET subscription_id = s.id, state = 'current'
 FROM billing_subscriptions s
-WHERE s.organization_id = o.id AND o.subscription_authority_state = 'none'
-AND NOT EXISTS (SELECT 1 FROM billing_subscriptions other WHERE other.organization_id = o.id AND other.id <> s.id);
-UPDATE organizations o SET subscription_authority_state = 'unavailable'
-WHERE o.subscription_authority_state = 'none' AND EXISTS (SELECT 1 FROM billing_subscriptions s WHERE s.organization_id = o.id);
+WHERE s.organization_id = a.organization_id AND a.state = 'none'
+AND NOT EXISTS (SELECT 1 FROM billing_subscriptions other WHERE other.organization_id = a.organization_id AND other.id <> s.id);
+UPDATE organization_subscription_authorities a SET state = 'unavailable'
+WHERE a.state = 'none' AND EXISTS (SELECT 1 FROM billing_subscriptions s WHERE s.organization_id = a.organization_id);
 --> statement-breakpoint
-ALTER TABLE organizations ADD CONSTRAINT organizations_subscription_authority_tenant_fk
-FOREIGN KEY (subscription_authority_id, id) REFERENCES billing_subscriptions (id, organization_id) ON DELETE RESTRICT;
-ALTER TABLE organizations ADD CONSTRAINT organizations_subscription_authority_check CHECK (
-  (subscription_authority_state = 'current' AND subscription_authority_id IS NOT NULL)
-  OR (subscription_authority_state IN ('none','unavailable') AND subscription_authority_id IS NULL)
-);
+CREATE FUNCTION seed_organization_subscription_authority() RETURNS trigger LANGUAGE plpgsql AS $$
+BEGIN
+  INSERT INTO organization_subscription_authorities (organization_id) VALUES (NEW.id);
+  RETURN NEW;
+END
+$$;
+CREATE TRIGGER organizations_seed_subscription_authority AFTER INSERT ON organizations
+FOR EACH ROW EXECUTE FUNCTION seed_organization_subscription_authority();
 --> statement-breakpoint
 -- Terminal projections retain the authoritative transition that produced Free.
 ALTER TABLE organization_entitlements DROP CONSTRAINT organization_entitlements_plan_state_check;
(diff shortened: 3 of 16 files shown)


ss251 (approved):
92304e4da41512038995bab97c9defb9528484fa
Static review of entitlement-authority binding at this head.

Claim: stale subscriptions could replay entitlements after authority advanced, and Free projections dropped provenance. Diff binds publication to `organizations.subscription_authority_id/state`.

Verified:
- `0379_subscription_account_authority.sql`: backfill separates `none` (no history) vs `current` (live in pending/incomplete/active/grace/past_due/unpaid, or single historical) vs `unavailable` (multiple terminals need reconciliation). Matches `subscription-account-authority.pglite.test.ts` org1->none, org2->sub1/current, org3->sub3/current, org4->unavailable. New composite FK `(authority_id,id)->(id,organization_id)` ON DELETE RESTRICT + CHECK `current<=>id NOT NULL` enforces tenant binding.
- `schemas/organization-entitlements.ts` + migration: free check relaxes from `IS NULL AND IS NULL` to `(IS NULL)=(IS NULL)`, so input canceled rev7 -> old output NULL/NULL losing provenance, new output `source_subscription_id/revision=7` retaining it per `subscription-entitlements-derivation.test.ts`.
- `repositories/subscription-entitlements.ts`: adds `lifecycle_revision` equality + authority equality + per-field equality on plan_key/status/grace_expires_at/digest/canceled_at/ended_at before `deriveSubscriptionEntitlementValues`. Input stale rev1 after advance to grace rev2 -> CONFLICT, input ORG_B+SUB_A -> SOURCE_NOT_FOUND, missing grace deadline -> mismatch CONFLICT leaving prior projection unchanged. Grace `effective_until` is now `grace_expires_at` with explicit invalid-deadline throw, not silent `current_period_end`.
- `repositories/subscription-authority.ts`: `requireCurrentAccountAuthority` on replay/advance + `UPDATE organizations SET authority=current` on create prevents replaced sub regaining authority with equal `created_at`, covered by new `subscription-entitlements.pglite.test.ts` replacement test.

Minor: first backfill `UPDATE ... FROM billing_subscriptions` with two live rows for one org picks an arbitrary row. Tenant-scoped so safe, but nondeterministic if that anomaly exists.

No binaries, symlinks, CI permission, lifecycle hooks, secrets, or telemetry in diff.

Verified by reading the diff at this head and tracing migration, schema, repository and PGlite test expectations.

Recommendation: merge

— [muse-lane-ss251]

NubsCarson (commented):
Independent review of `d2bfbd5701678f1844f2cd216ad8751d1edff8d4` against `b676280384b15df54204276e34bb8740db26936f`.

No remaining actionable source findings in the advertised prerequisite. The association table preserves tenant-scoped account identity through terminal transitions. New identity creation checks the identity captured before provider work, stale create/advance/rebuild cannot repoint it, and publication checks both current lifecycle revision and projection CAS under organization-first locks. Grace uses the actual validated deadline, and terminal Free retains source provenance. Missing and ambiguous authority remain explicit failures rather than silently selected historical subscriptions.

The original historical-create and erasure-order findings are addressed. The real erasure-adapter fixture now applies the subscription migrations, checks the irreversible fence, and exercises actual association release, journal/subscription erasure and final organization cascade. No release mock substitutes for the changed production path.

Independent validation on the clean reviewed head: **34 passed, 6 skipped, 0 failed; 168 assertions** across seven files, exit 0. This includes migration/backfill rollback, entitlement derivation/repository behavior, billing operations, full-schema deletion FK policy, the real erasure adapter, and the PostgreSQL integration file (skipped locally).

```sh
bun test --config=/dev/null --isolate packages/cloud/shared/src/db/repositories/subscription-entitlements-derivation.test.ts packages/cloud/shared/src/db/repositories/subscription-entitlements.pglite.test.ts packages/cloud/shared/src/db/repositories/subscription-billing-operations.pglite.test.ts packages/cloud/shared/src/db/migrations/subscription-account-authority.pglite.test.ts packages/cloud/shared/src/lib/services/account-deletion-provider-adapters.pglite.test.ts packages/cloud/shared/src/db/account-deletion-foreign-key-policy.test.ts packages/cloud/shared/src/db/repositories/subscription-authority.postgres.integration.test.ts
```

Limits: this is a code review of the prerequisite, not approval of the full #23094 enforcement scope or release readiness. The local PostgreSQL integration tests were skipped because their database was not configured; separate-session waiter/lock proof remains required. Full shared-suite completion, root verification, exact-head hosted gates, staging quota enforcement and Dedicated/provider acceptance were not independently established. Draft status remains appropriate. No production migration, deployment or money operation was performed.

attentionhead (approved):
## APPROVE — eight guards, seven killed by your own tests and the eighth verified by probe

Reviewed at head `09d3fa484a47f4fc3beaa6f19b7dc1498c9afe44` (merge base `cd821540d73c4e7370a9407c07be19020a3b810b`), 15 files, +929/-134. Every check is green including the `All Tests Passed` aggregate. All six changed or added suites pass locally under `bun test`: **34 pass / 0 fail** (2 + 10 + 4 + 9 + 8 + 1).

### Verified

**Mutation testing.** I reverted each guard the PR adds and re-ran the six suites:

| mutant | result |
|---|---|
| M1 — drop the account-authority currency check in `rebuildInTransaction` | **killed** (32/2) |
| M2 — drop the `subscription.lifecycle_revision` currency check | **survives** (34/0) |
| M3 — drop the nine-field subscription-vs-revision agreement loop | **killed** (10/4) |
| M4 — `effective_until` ignores `grace_expires_at` | **killed** (30/4) |
| M5 — drop the lifecycle-deadline validation | **killed** (33/1) |
| M6 — `advance()` skips `requireCurrentAccountAuthority` | **killed** (32/2) |
| M7 — `releaseForAccountDeletion` drops the `deletion_irreversible` requirement | **killed** (32/2) |

Seven of eight die on your own assertions, which is a much better ratio than this repo usually shows for a change of this size. M4 and M7 in particular are the two I would most want covered, and both are.

**M2 — I chased the survivor rather than reporting it as a gap.** It survives because the nine-field agreement loop *subsumes* it: any advance that changes a derived field already trips the loop. The `lifecycle_revision` check is only load-bearing when a revision advances changing **only** a field outside that loop. `cancel_at_period_end` is exactly such a field — neither compared by the loop nor read by `deriveSubscriptionEntitlementValues` — so I built a probe around it: bump `lifecycle_revision` to 2 changing only `cancel_at_period_end`, then rebuild from the now-stale revision 1.

```
probe at head                      -> 1 pass
  rebuild(revision 1) rejects with SUBSCRIPTION_ENTITLEMENT_CONFLICT
  the stored entitlement keeps source_subscription_revision null
  rebuild(revision 2) then publishes

same probe with M2 removed         -> 0 pass, 1 fail
  the superseded revision publishes
```

So the check is correct and genuinely load-bearing — it is the only thing standing between a superseded revision and a published entitlement whose `source_subscription_revision` disagrees with the subscription's current one. It is simply unasserted. That is the single change I would most want a test on, and the probe above is small enough to lift verbatim.

**Migration ordering.** The `0379` backfill is order-correct and each step is guarded against the previous one: the live-status UPDATE claims unambiguous accounts first; the sole-historical UPDATE is gated on both `a.state = 'none'` and a `NOT EXISTS` that requires `s` to be the org's *only* subscription; and the `unavailable` UPDATE then catches exactly the residue — orgs still at `'none'` that do have subscriptions, i.e. the genuinely ambiguous multi-terminal case. Orgs with no subscriptions at all correctly stay `'none'`. The `state_check` CHECK and the `(subscription_id, organization_id)` composite FK make the "current implies a real, same-tenant subscription" invariant unrepresentable rather than merely unwritten, and the `AFTER INSERT` trigger closes the seeding gap for organizations created after the migration.

**The relaxed `organization_entitlements_plan_state_check` matches the new derivation.** Loosening the free branch from `source_subscription_id IS NULL AND source_subscription_revision IS NULL` to `(source_subscription_id IS NULL) = (source_subscription_revision IS NULL)` is exactly what the new free branch of `deriveSubscriptionEntitlementValues` needs, since it now carries `source_subscription_id` / `source_subscription_revision` through to the terminal projection. Both halves of the pair still have to agree, so a half-populated source is still rejected. Schema and migration state the constraint identically.

**Lock ordering is consistent across the three writers.** `create()`, `advance()` and `rebuildInTransaction` all take `organizations` `FOR UPDATE` first and `organization_subscription_authorities` second, so the new row lock cannot introduce a deadlock cycle with the existing ones.

**No caller is broken by the new required parameter.** `create()`'s third argument `expectedAccountSubscriptionId` is mandatory, so I swept for consumers: the only non-test reference to `subscriptionAuthorityRepository` anywhere in `packages/` or `plugins/` is `account-deletion-provider-adapters.ts:512`, calling the new `releaseForAccountDeletion`. `create()` and `advance()` have no production call site yet, so the signature change is contained to tests, all of which are updated.

**`effective_until` now follows the actual deadline.** Using `grace_expires_at` for `grace` and `current_period_end` otherwise, plus the finite/ordered validation, is the substantive user-visible fix here — a grace subscription previously published the period end as its entitlement deadline. The parameterised test covering both a grace deadline before and after the period end is the right shape, and M4 confirms it bites.

### Not verified

- `subscription-authority.postgres.integration.test.ts` (+85) — needs a real PostgreSQL; not run locally. It is covered by the green `Subscription authority PostgreSQL` lane on this PR.
- I did not exercise the `0379` migration against a database that already contains multi-subscription organizations at scale; the backfill logic is reviewed by reading plus the PGlite migration test.

### Non-blocking observations

1. **`find()` moved from `dbRead` to `dbWrite`.** This routes every entitlement read to the primary. The read-your-writes motivation is clear from the rest of the PR, but it is the one change here with a production load implication and it carries no comment, unlike the other decisions in this diff which are all annotated. A one-line note saying replica lag can publish a stale entitlement would stop someone "optimizing" it back later.

2. **The nine-field agreement loop and the `lifecycle_revision` check overlap.** As the M2 result shows, the loop covers the currency check for every field it compares. That redundancy is fine — but if the compared list is ever trimmed, the `lifecycle_revision` check silently becomes the sole guard for more cases than it is today, with no test holding it. Same reason as above for wanting the probe committed.

3. **`readAccountAuthority` conflicts when the row is missing**, which after this migration should be impossible for a live organization (backfill plus trigger). That is the right fail-closed choice; worth knowing it is the branch that fires if the trigger is ever dropped in a future migration without a replacement.

Approving — the invariant is enforced in the schema rather than only in code, the backfill handles the ambiguous case explicitly instead of guessing, and the test coverage is strong enough that I had to construct an artificial field to find the one uncovered guard.

— [claude-code-eliza]

agent-cortex (approved):
## APPROVE — independently reproduced in a disposable sandbox at head `09d3fa4`

Reviewed the full diff at head `09d3fa484a47f4fc3beaa6f19b7dc1498c9afe44` (merge base `cd821540d73c4e7370a9407c07be19020a3b810b`), 15 files, +929/-134, and independently exercised the changed money-path in an offline sandbox (network-denied) rather than reading it. My run corroborates the earlier reviews from a separately-built environment and adds two mutation probes on the two guards I most wanted covered.

### What I verified by execution

Ran the six changed/added suites under `bun test --config=/dev/null --isolate --conditions eliza-source` (source-resolved workspace, PGlite-backed):

```
subscription-account-authority.pglite.test.ts        2 pass   (backfill w/o chronology; migration rollback)
subscription-entitlements.pglite.test.ts            10 pass   (replay-after-advance; grace deadline x2;
                                                               missing-deadline leaves prior projection;
                                                               tenant fencing; concurrent sources;
                                                               same-tx rollback; erasure fence;
                                                               explicit-replacement active/canceled identity)
subscription-billing-operations.pglite.test.ts       9 pass
subscription-entitlements-derivation.test.ts         4 pass
account-deletion-foreign-key-policy.test.ts          (pass, in the 33)
------------------------------------------------------------------
33 pass, 128 expect() calls
```

The only suite I could not complete is `account-deletion-provider-adapters.pglite.test.ts`, which fails in my image with `Cannot find module '@elizaos/core/edge'` — the `./edge` export has no `eliza-source` condition, so it resolves to an unbuilt `dist/edge` bundle pulled in transitively (via `plugin-web-search/src/edge.ts`). That is an environmental build-artifact gap on my side, **not** a defect in this PR; the erasure-release path it covers is also exercised by the `account erasure releases the reverse pointer only behind the irreversible fence` case inside the entitlements pglite suite, which passes. `subscription-authority.postgres.integration.test.ts` needs a real PostgreSQL and was not run locally (covered by the green PostgreSQL lane on the PR).

### Mutation probes (guards actually bite)

I reverted two guards and re-ran the committed suites:

- **Grace deadline** — changed `effective_until: effectiveUntil` back to `revision.current_period_end`: `subscription-entitlements-derivation.test.ts` → **1 fail** (`publishes the actual grace deadline`). The `grace_expires_at`-vs-`current_period_end` fix is the user-visible correctness change here and it is asserted.
- **Account-authority currency** — forced the `!accountAuthority || state !== "current" || subscription_id !== sourceSubscriptionId` check to `false` in `rebuildInTransaction`: `subscription-entitlements.pglite.test.ts` → **2 fail** (the explicit-replacement identity cases). The replay-prevention guard that is the point of this PR is asserted, not just present.

### Traced by reading, consistent with the run

- **`0379` backfill is order-correct and each step is guarded against the prior one.** Live-status UPDATE claims unambiguous accounts first; the sole-historical UPDATE is gated on both `state='none'` and a `NOT EXISTS` requiring `s` to be the org's only subscription; the `unavailable` UPDATE then catches exactly the residue (orgs still `'none'` that do have subscriptions). Orgs with no subscriptions stay `'none'`. The `state_check` CHECK plus the `(subscription_id, organization_id)` composite FK make "current implies a real, same-tenant subscription" unrepresentable, and the `AFTER INSERT` trigger closes the seeding gap for organizations created post-migration.
- **Relaxed `organization_entitlements_plan_state_check` matches the new derivation.** The Free branch now carries `source_subscription_id`/`source_subscription_revision` through to the terminal projection, so loosening `IS NULL AND IS NULL` to `(IS NULL) = (IS NULL)` is exactly what the derivation needs while still rejecting a half-populated source. The `source_revision_tenant_fk` (`(source_subscription_id, organization_id, source_subscription_revision)` → `billing_subscription_revisions`, ON DELETE RESTRICT) is always satisfiable because the row is derived from an existing revision.
- **Lock ordering is uniform.** `create()`, `advance()`, and `rebuildInTransaction` all take `organizations FOR UPDATE` first and `organization_subscription_authorities` second, so the new row lock adds no deadlock cycle.
- **No caller broken by the new required `expectedAccountSubscriptionId`** — `create()`/`advance()` have no production call site yet; the only non-test consumer of `subscriptionAuthorityRepository` is `account-deletion-provider-adapters.ts` calling the new `releaseForAccountDeletion`, which is fenced on `account_lifecycle_state = 'deletion_irreversible'`.

### Non-blocking (not gating this approval)

1. **`find()` moved from `dbRead` to `dbWrite`** (`subscription-entitlements.ts:157`), routing every entitlement read to the primary. The read-your-writes motivation is clear, but this is the one change with a production-load implication and it is the only decision here without a comment. A one-line note that replica lag could otherwise publish a stale entitlement would stop a future "optimization" back to the replica.
2. **The `lifecycle_revision` currency check is unasserted.** It is subsumed by the nine-field agreement loop for every field the loop compares, so reverting it alone does not fail any test; it only becomes load-bearing for a revision that advances changing solely a field outside that loop (e.g. `cancel_at_period_end`). Fine today, but if the compared-field list is ever trimmed it silently becomes the sole guard with no test holding it — a two-line assertion would pin it.
3. Scope is explicitly a prerequisite slice of the larger enforcement work; this approval is of the diff's correctness, not release-readiness of the full enforcement scope.

The invariant is enforced in the schema (CHECK + composite tenant FK) rather than only in code, the backfill handles the ambiguous multi-terminal case explicitly instead of guessing, and my mutation probes confirm the two guards that matter most are covered by real assertions. No correctness, security, tenant-isolation, or required-test gap found.

---
Contribution skill revision: elizaOS/eliza@b676280384b15df54204276e34bb8740db26936f:packages/skills/skills/contribute-to-eliza
— [eliza-swarm]

PR #29151, opened 2026-08-26
refactor(ui): move feature views into plugins

lalalune:
## Summary

- Move feature-owned views out of `packages/ui` into their owning plugins.
- Add side-effect app registration and lazy plugin loaders for calendar, contacts, documents, Eliza Cloud, maps, messages, notes, phone, and relationships.
- Give plugin-owned pages a shared `ViewHeader` with a back-to-home affordance and preserve host-only external importer boundaries for deep UI components.

## Validation

- Full workspace build: `132/132` successful; authoritative view bundle guard: `21/21` valid.
- Plugin documents: `166` tests passed; plugin maps: `87`; plugin phone: `91`; plugin relationships: `40`; plugin messages: `31`.
- UI focused view/connector tests: `70` passed; design-system/molecular/story audit: `25/25` passed.
- App registration tests and app-core/UI/plugin typechecks passed.
- App visual audit: `211/211` Playwright tests passed; `broken=0`, `needs-work=0`.
- Repository audits passed: agent/CLAUDE parity, plugin view inventory, view bundle inventory/import guard, publish graph, dependency references, and TypeScript workspace-resolution checks.

## Known baseline/environment limitations

- The root `bun run verify` reaches the existing `plugin-notes` typecheck failure in `packages/ui/src/hooks/useRenderGuard.ts` (`TS2367`); this branch does not modify that file.
- The full contacts suite retains six pre-existing contract failures against the current upstream ContactsWeb/provider contract.
- The full app script suite retains three simulator/network timing failures (Android stability expectation, Android smoke timeout, and readiness timeout boundary).
- OCR triage reports one false positive for an unrelated computer-use mobile-landscape view; the captured screenshot visibly contains the expected “Research browser” text, and the Playwright/aesthetic audit passed.

The original main worktree had an unrelated unresolved cloud voice-session conflict; it was preserved and is not part of this branch.

PR #30773, opened 2026-09-06
fix(ci): qualify promotions and repair app acceptance failures

lalalune:
## Scope

Addresses #30552, #30855 and #30992. Live deployment cutover remains open under #30855.

Qualify each branch's exact commit before promoting develop → staging → main. Promotion creates a reviewed PR; the reconciler cannot update branch refs or merge it. Protected production effects verify immutable source certification after approval and recheck freshness immediately before publication or host mutation. Invalid, missing, expired or wrong-tree evidence denies admission before effects.

Qualification exposed concrete app defects, repaired here: missing local return controls after shared-header removal, wallet dependencies loaded unnecessarily by the initial renderer, narrowly audited Android telemetry handling, and unreadable danger/muted/choice states. Danger text and filled danger controls now use separate semantic colors. Choices pair with their actual card surface across four themes, policy actions wrap on mobile, and real confirmation controls retain legible rest/hover states. Orange remains the accent. Test runners and consumer fixtures exercise current behavior; obsolete header/constant-only assertions were removed.

## Exact source and gates

Candidate `38fdddc13c3f5c5b891ee558c47f6b6494275e70`, rebased onto develop `d7df7af187f05ec744b57c01799cfbabcc5fd325`. Final normal installation and root verification passed all 373 tasks plus repository audits. Changes after `dba1962` are a behavioral Choice consumer-test repair and a secret-scanner exception for the exact public SDK integrity digest in its owning file. The full 40-commit scan passes; committed negative controls still detect a different value in that file and the same value in another file. Production sources are unchanged. The final full UI suite passed 14,053 tests across 1,307 files, with seven explicit skips; the Choice repair verifies selected state, locked controls and once-only dispatch instead of CSS class literals. Fresh final static Storybook build and four affected browser stories passed: all three play functions completed, zero console or accessibility failures. Hosted checks must finish successfully for the published SHA before merge.

- The final-base UI residual replay passed all 372 tests in 10 files, including real local navigation and portable stories. A story-local boundary fixes an empty-output assertion in both browser and jsdom harnesses without skipping the play function.
- Incoming native/build consumers passed: app-core scripts 99 Node + 21 Bun; real workspace import builds 9; runtime installation identity 20 with one explicit platform skip. Shared Settings/app-core fixture repairs already merged in #30609 were deduplicated.
- Full app-core Vitest passed 4,572 tests with 25 skips across 365 passing files and one skipped file. This ran on the exact integration tree later reconciled to the Settings merge; later native changes have their separate focused proof above.
- Full personal-assistant suite: 2,601 passed, six skipped. Calendar: 840 passed, four skipped. Owning typechecks and lint passed. Full UI design graph, semantic token and story coverage gates passed without debt exemptions.
- Current-base planner/delivery consumers: 179 passed across four files. Incoming provider wire consumers: 46 passed. These validate integrated base changes; this PR does not alter model/planner behavior.
- Certification tests use real ZIP archives, controlled late clocks, ordered effect admission and actual publication/migration/remote-shell guards. Infrastructure: 124 tests and seven Terraform mocked plans, backend initialization disabled. No live production effects were executed.
- Earlier owning suites: scenario runner 746 unit + 19 mock integration; auth 300; workflow 125; goals 81 with one live-only skip; full Cloud shared 1,230 files in 388 isolated processes.

Preserved broad results are not represented as wholly green: Cloud API's 604-file sweep had three import failures during concurrent generated-output rebuilding; unchanged serial replay passed all 57 cases. Scripts' 315-file sweep had one contention-sensitive five-second failure; unchanged isolated replay passed all 29 cases without increasing its timeout. Independent full UI runs exposed obsolete fixtures and the concrete navigation defects; all current residual files passed together above. The old exact-e2f hosted run 34071740596 failed and is not still running.

## Visual and Android acceptance

Initial full Storybook qualification found two intentional-empty states and 83 serious contrast failures. All 85 reproduced on unchanged develop `129149`; the baseline established defects, not a waiver. Repairs under #30992 then passed a fresh complete 1,672-story catalog on `712cd17`: zero broken or accessibility failures. [Exact full-catalog report](https://github.com/elizaOS/eliza/releases/download/pr-evidence-4/30773-712cd17-full-story-report.json). Static fixture resource 404 diagnostics remain explicitly recorded; this is not live backend acceptance or a clean-console claim.

After canonical Badge composition and dynamic card-hover inheritance corrections, exact `161aae7` passed a rebuilt 111-story affected gate (111 good, zero accessibility failures, all 16 plays), 72 responsive rest/hover/selected-hover frames and 12 additional first-run frames. All affected pixels were inspected, with independent root-agent review of changed frames. Full app audit passed 222 tests and produced 216 captures: zero broken/needs-work layouts or hover failures; OCR: 203 verified, 13 manual, zero broken. All 16 affected app captures were reviewed. The final RootView-only story followup passed its separate exact browser replay; product renderer source is unchanged.

The Android Cloud renderer built on `161aae7` using the supported build:web toolchain: 392 assets, canonical 314 text-asset routing scan with zero findings and unchanged bytes, plus viewport/entry/chunk/shader checks. This proves the renderer/package boundary, not an APK/AAB installation or device run. The exact audited Coinbase telemetry literal is admitted; unknown or modified service-worker references remain rejected.

[Final install](https://github.com/elizaOS/eliza/releases/download/pr-evidence-4/30773-38fdddc-install.log) · [root verification](https://github.com/elizaOS/eliza/releases/download/pr-evidence-4/30773-38fdddc-verify.log) · [secret scan](https://github.com/elizaOS/eliza/releases/download/pr-evidence-4/30773-38fdddc-secret-scan.log) · [negative controls](https://github.com/elizaOS/eliza/releases/download/pr-evidence-4/30773-7daa224-secret-controls.json) · [source reconciliation](https://github.com/elizaOS/eliza/releases/download/pr-evidence-4/30773-7daa224-source-reconciliation.json) · [final story replay](https://github.com/elizaOS/eliza/releases/download/pr-evidence-4/30773-dba1962-story-report.json).

Source binding: [captured `161aae7` → final candidate reconciliation](https://github.com/elizaOS/eliza/releases/download/pr-evidence-4/30773-161aae7-final-source-reconciliation.json). App/core/calendar/personal-assistant trees are identical; the only shared UI delta is the final empty-state story harness. Incoming native packaging changes are identified separately.

## Reviewed evidence

The documented explicit ad-hoc compatibility review passed for 179 artifacts, 138 screenshots and 4 videos, with packaged OCR 138/138. Its two generated review receipts are attached separately. This is archived evidence review, not a pre-snapshot matrix bundle or deployment certificate. Original PNGs, converted JPGs and MP4s have source/hash provenance; uploaded bytes are verified against GitHub SHA-256 and size.

[App audit log](https://github.com/elizaOS/eliza/releases/download/pr-evidence-4/30773-161aae7-app-audit.log) · [layout report](https://github.com/elizaOS/eliza/releases/download/pr-evidence-4/30773-161aae7-app-report.json) · [OCR](https://github.com/elizaOS/eliza/releases/download/pr-evidence-4/30773-161aae7-app-ocr.json) · [selected story report](https://github.com/elizaOS/eliza/releases/download/pr-evidence-4/30773-161aae7-story-selected-report.json) · [responsive report](https://github.com/elizaOS/eliza/releases/download/pr-evidence-4/30773-161aae7-responsive-report.json) · [Android scan](https://github.com/elizaOS/eliza/releases/download/pr-evidence-4/30773-161aae7-android-audit.json) · [artifact manifest](https://github.com/elizaOS/eliza/releases/download/pr-evidence-4/30773-161aae7-upload-manifest.json) · [ad-hoc review](https://github.com/elizaOS/eliza/releases/download/pr-evidence-4/30773-dba1962-publication-review-manifest.json)

Before/after desktop and mobile policy controls (before source `129149`; after `161aae7`):

Desktop before
Desktop after
Mobile before
Mobile after

Final app desktop/mobile workflow and Vault:

workflow desktop-landscape
workflow mobile-portrait
vault desktop-landscape
vault mobile-portrait

[Domain confirmation hover](https://github.com/elizaOS/eliza/releases/download/pr-evidence-4/30773-161aae7-domain-mobile-hover.mp4) · [Four-theme trajectory sample](https://github.com/elizaOS/eliza/releases/download/pr-evidence-4/30773-161aae7-trajectory-app-desktop-hover.mp4) · [Mobile selected error hover](https://github.com/elizaOS/eliza/releases/download/pr-evidence-4/30773-161aae7-trajectory-app-mobile-hover.mp4) · [Mobile policy hover](https://github.com/elizaOS/eliza/releases/download/pr-evidence-4/30773-161aae7-policy-mobile-hover.mp4)

Historical interaction recordings at exact `dfc8075` retain separate provenance: [workflow desktop](https://github.com/elizaOS/eliza/releases/download/pr-evidence-4/30773-dfc8075-workflow-desktop.mp4) and [workflow mobile](https://github.com/elizaOS/eliza/releases/download/pr-evidence-4/30773-dfc8075-workflow-mobile.mp4). They exercise create/execute/reload; local Cloud/trajectory Back paths were additionally replayed against `299c884`. These earlier recordings are not relabeled final `161aae7`. [Complete source proof](https://github.com/elizaOS/eliza/releases/download/pr-evidence-4/30773-dfc8075-complete-source-proof.json).

Historical local-return acceptance (before renderer c634, after dfc; these show the concrete missing-control repair):

| Surface | Before | After |
|---|---|---|
| Cloud desktop | [Before](https://github.com/elizaOS/eliza/releases/download/pr-evidence-4/30773-dfc8075-before-local-back-before-nested-Cloud-Back-returns-to-overview-desktop-cloud-back-rest.jpg) | [After](https://github.com/elizaOS/eliza/releases/download/pr-evidence-4/30773-dfc8075-after-local-back-nested-Cloud-Back-returns-to-overview-desktop-cloud-back-rest.jpg) |
| Cloud mobile | [Before](https://github.com/elizaOS/eliza/releases/download/pr-evidence-4/30773-dfc8075-before-local-back-before-nested-Cloud-Back-returns-to-overview-mobile-cloud-back-rest.jpg) | [After](https://github.com/elizaOS/eliza/releases/download/pr-evidence-4/30773-dfc8075-after-local-back-nested-Cloud-Back-returns-to-overview-mobile-cloud-back-rest.jpg) |
| Activity mobile | [Before](https://github.com/elizaOS/eliza/releases/download/pr-evidence-4/30773-dfc8075-before-local-back-before-mobile-r-124a2-eturns-to-recorded-activity-mobile-activity-detail-rest.jpg) | [After](https://github.com/elizaOS/eliza/releases/download/pr-evidence-4/30773-dfc8075-after-local-back-mobile-run-detail-returns-to-recorded-activity-mobile-activity-detail-rest.jpg) |

Personal-assistant actual destructive confirmation: [rest](https://github.com/elizaOS/eliza/releases/download/pr-evidence-4/30773-lifeops-confirm-purge-rest.jpg), [hover](https://github.com/elizaOS/eliza/releases/download/pr-evidence-4/30773-lifeops-confirm-purge-hover.jpg), [browser log](https://github.com/elizaOS/eliza/releases/download/pr-evidence-4/30773-lifeops-contrast-e2e.log). Captured while reviewing the repair later committed in712; separate provenance, not relabeled final-head capture.

Evidence boundaries: app/browser acceptance uses deterministic HTTP fixtures and records console/network findings in the reports; no live backend or deployment success is claimed. Live-model trajectories are N/A because this PR does not change prompts, model selection or planner/action semantics. Native device installation is N/A for these renderer/CI changes; no device success is claimed. Existing historical `299c884`/`dfc8075` evidence remains available with its own source receipts.

## Cutover boundary

Source merge does not activate the branch/environment topology. Live readback found the legacy reconciler disabled, staging absent and repository rulesets empty. Retain disabled-first cutover: owner review/readback of branch and protected-environment policy, environment bindings/DNS, successful source-effect certification and served-commit verification. Do not mark #30855 complete from source qualification. No production deployment, repository policy change, credential rotation or live infrastructure mutation was performed.

- [x] Before full-page desktop/mobile contrast and local Back captures are linked inline with exact source provenance.

- [x] After full-page desktop/mobile screenshots and responsive rest/hover reports are linked inline; captured161 product sources are bound to this exact head.

- [x] Desktop/mobile workflow walkthroughs and final confirmation/choice hover clips are linked with their distinct captured sources.

- [x] N/A for live deployment: browser paths use deterministic HTTP fixtures. Actual process/transport test logs and UI fixture network findings are included; no live backend success is claimed.

- [x] App audit and story reports include frontend console/network/overflow/hover results; resource404 fixture diagnostics are disclosed above.

- [x] N/A: this PR does not change prompts, model selection, planner or action semantics; no live-model result is claimed.

- [x] Certification, canonical Android asset checks, source trees, artifact hashes, manual pixel review and upload readback provide the domain proof.

[Final full UI suite](https://github.com/elizaOS/eliza/releases/download/pr-evidence-4/30773-38fdddc-ui-full.log) · [Final test-only source reconciliation](https://github.com/elizaOS/eliza/releases/download/pr-evidence-4/30773-38fdddc-source-reconciliation.json).


ss251 (approved):
Reviewed head 8d5a531850455735a3b00c2a5b4b607bfcd6d41b.

The risky part is `packages/ui/src/spatial/primitives.tsx:249` and `:662`, where the spatial `Field kind="select"` drops custom `Select/SelectTrigger/Content` for `NativeSelect`. I checked the bridge contract that usually breaks here: `buildFieldSpec` now keeps the explicit `normalizeAgent(props.agent)` id via `...agent` spread instead of overwriting with `field:label`, and sets `role:"select"` vs `"text-input"`; the renderer emits a real `<select>` with `aria-label`, `defaultValue={spec.value ?? ""}`, `...agentDataProps(spec.agent)`, `onChange={(e)=>onChange(e.target.value)}`, and only injects the disabled-hidden `""` placeholder when `!spec.options?.includes("")`. `packages/ui/src/components/views/DynamicViewLoader.test.tsx` exercises `list-elements` reporting `fillable:true` then `agent-fill` Active/Paused filtering through that handler.

No blocking concerns.
Recommendation: merge

— [muse-lane-ss251]

attentionhead (changes requested):
**Requesting changes on one change only** — the `ChatOverlay` backdrop-filter handling. Everything else I checked is correct, and the red CI is not your fault. Details below, including the part of your reasoning I verified as right.

## First: the red CI is noise, not a failure

`All Tests Passed` is red **only** because its admission gate saw `source-smoke finished with cancelled`:

```
##[error]source-smoke finished with cancelled
##[error]Process completed with exit code 1
```

Four sibling lanes are `CANCELLED` (supersession) and four are still pending. No test failed. Not a reason to hold this.

## Verified correct

- **`spatial/primitives.tsx`** — moving `agentDataProps` off `SelectTrigger` onto the real `<select>` is exactly what makes agent-fill hit the same change handler as a user selection. The placeholder guard `!spec.options?.includes("")` behaves correctly when `options` is undefined, and `buildFieldSpec` spreading `...agent` last means caller fields win while `label`/`value` defaults fill in — a widening, not a regression.
- **`onboarding-chat.ts`** — chaining fetch→buffer inside one promise and cancelling an unread, unlocked body in the catch closes a real leak where headers arriving after the caller timed out left a body with no owner. `J6` annotations correctly scoped.
- **`native-select.tsx` / `WorkflowTriggerPanel` / `VaultInventoryPanel`** — the `compact` variant reproduces the inlined classes it replaces; `useLayoutEffect` before paint is right for focus-loss protection.
- **`--chat-sheet-backdrop-filter: undefined → "none"`** for `nativeInsetSheet` is **inert** — both consumers read `var(--chat-sheet-backdrop-filter, none)`, so the fallback already produced `none`. Not an issue.

## Blocking: the backdrop toggle returns at the frame the old comment ruled out

You deleted this, and its rationale:

```ts
// Toggling backdrop-filter on at the first downward frame forces a new
// compositor surface exactly when the finger needs the frame budget.
const cssSheetBackdropActive = !nativeInsetSheet;
```

That value was **constant** for the CSS tier — the filter never toggled, at any point. The replacement is a `MotionValue`:

```ts
const surfaceBackdropFilter = useTransform(
  surfaceBlackout,
  (t) => (t >= 1 ? "none" : GLASS_SHEET_BACKDROP_FILTER),
);
```

**Your stated justification is correct as far as it goes, and I verified it.** Tracing `surfaceBlackout` → `sheetBlackoutProgress` = `max(clamp01(threadHeight/halfDetent), clamp01(fullBleedT))`: during maximize *and* restore, `threadHeight` stays at or above the half detent, so `t` pins at `1` and the filter holds `"none"` throughout. No toggle there. That half of the new comment is right.

But the toggle isn't eliminated — it's **relocated to the collapse drag**. Collapsing from the half detent to the composer has `fullBleedT = 0` and `threadHeight` crossing `halfDetent` downward, so `t` crosses `1` and the filter flips `"none" → blur(30px)` **mid-drag, under the finger**. That is precisely the "first downward frame" the deleted comment named. The new comment establishes that maximize/restore are safe — which was never the contested case — and then describes the collapse flip as though it were benign, without addressing the cost the old comment attributed to it.

A second, independent source says the same thing — `CHAT_OVERLAY_DESIGN_SPEC.md:84`:

> Open/close motion is opacity/translate only — **animating blur/filter** or scaling the transcript **repaints too much and janks**.

This change makes `--chat-sheet-backdrop-filter` a motion value driven by finger-tracked `threadHeight`, which is what that line forbids.

The saving is real but narrow — paint work on a backdrop that's hidden behind an opaque fill. The cost is compositor-surface creation during the one interaction that is finger-tracked. Two documented rationales say that trade goes the other way, and the diff touches no `ChatOverlay` or `chat-overlay-motion` test, and the body carries no frame-budget measurement.

**Any one of these unblocks it:**
1. Keep the filter constant across the half-detent crossing (restore the old behaviour) and drop the opaque-fill optimisation; or
2. gate the flip on a non-drag condition so it can't land mid-gesture; or
3. attach a frame-budget measurement on the collapse drag showing the toggle is free in practice — in which case please update `CHAT_OVERLAY_DESIGN_SPEC.md:84` and leave a note where the deleted comment was, so the next reader doesn't restore the old behaviour on the strength of a rationale you've since measured as obsolete.

Happy to re-review quickly. The other four production changes are good, and if you'd rather land those now, splitting the `ChatOverlay` hunk out would let this merge immediately.

— [claude-code-eliza]

attentionhead (approved):
## APPROVE — my earlier blocker on this PR is answered; re-verified at the new head

Re-reviewed at head `91132ed12cea05dcad954bbd47f0b73b4ceb44b4` (merge base `6065de5d34655a6f66d059418085a627af57f200`), 33 files, +766/-253.

### Withdrawing my earlier ChatOverlay objection

At the previous head I requested changes on `ChatOverlay.tsx` because `surfaceBackdropFilter` re-introduces a `backdrop-filter` toggle that the deleted comment — and `CHAT_OVERLAY_DESIGN_SPEC.md:84` (*"Open/close motion is opacity/translate only — animating blur/filter … repaints too much and janks"*) — specifically warned about at the first downward frame of a restore. **That objection was wrong, and the new comment is right.** I checked it rather than taking it on trust:

```ts
// chat-overlay-motion.ts
const reveal = halfDetentPx > 0 ? clamp01(threadHeightPx / halfDetentPx) : 0;
return Math.max(reveal, clamp01(fullBleedProgress));
```

`reveal` **saturates** at `threadHeight >= halfH`, so `surfaceBlackout` is pinned at `1` across the entire maximize *and* restore range no matter what `fullBleedProgress` is doing — and `ChatOverlay.tsx:3180` (`detentH = !sheetOpen ? 0 : expanded ? openH : halfH`) confirms every open resting state is at or above `halfH`. Restore from full-bleed therefore lands on `halfH` without ever dropping the blackout below 1, so `surfaceBackdropFilter` stays `"none"` for the whole gesture and no new compositor surface is created mid-drag. This PR's own committed assertions pin exactly that boundary:

```
ChatOverlay.morph.test.tsx:180   sheetBlackoutProgress(353, 353, 0) === 1
ChatOverlay.morph.test.tsx:181   sheetBlackoutProgress(700, 353, 0) === 1
```

The filter now returns only below the half detent — where the fill is genuinely translucent and the frost is visually required — which is strictly less toggling than the previous always-on `cssSheetBackdropActive`. The comment's claim is accurate as written.

I also checked the one thing that could have made the `undefined` → `"none"` switch a silent behavior change: `ChatOverlay.tsx:6010-6011` consume the variable as `var(--chat-sheet-backdrop-filter, none)`, so an unset property and an explicit `"none"` resolve identically for the `firstRunOpen || nativeInsetSheet` branch. No-op, as intended.

### Verified

**CI.** Billing payment replay E2E, Subscription authority PostgreSQL, and Browser bridge Windows security are all SUCCESS on the live run; source-smoke is still pending. Nothing is failing.

**The `tests_server` partition split is provably disjoint and exhaustive.** I evaluated the two `ci.yml` filters against representative task names, including an `@elizaos/agent`-prefixed decoy that a sloppy anchor would have mis-partitioned:

| task | `agent` filter | `remaining` filter |
|---|---|---|
| `@elizaos/agent (packages/agent)#test` | ✔ | ✘ |
| `@elizaos/core (packages/core)#test` | ✘ | ✔ |
| `@elizaos/agent-extra (packages/x)#test` | ✘ | ✔ |

Exactly one partition claims each task. The new `packages/scripts/__tests__/ci-server-partitions.test.mjs` pins this against the **real** planner rather than a re-implementation, which is the right shape for a CI-topology guard.

**That new suite actually runs.** It is a `bun:test` file with an `.mjs` extension, so I checked collection rather than assuming: `node --test` rejects it (`ERR_UNSUPPORTED_ESM_URL_SCHEME: protocol 'bun:'`), while `bun test __tests__/ci-server-partitions.test.mjs` gives **1 pass / 8 expect() calls**, and `bun test __tests__/` collects it among 168 files. Not an orphan.

**The vitest exclusion is correctly paired.** `packages/app-core/vitest.config.ts` adds `scripts/lib/fused-artifact-integrity.test.mjs` to `exclude`, and `package.json` adds the same path to the `test:script-suites` `node --test` list in the same commit. This is the failure mode I have seen go wrong on this repo — a suite excluded from vitest and registered nowhere, silently running in no lane — and here both halves landed together. `node --test scripts/lib/fused-artifact-integrity.test.mjs scripts/stage-desktop-fused-lib-staleness.test.mjs` → **10 tests, 10 pass, 0 fail**.

**UI.** `chat-overlay-motion.test.ts` + `ChatOverlay.morph.test.tsx` → **17 passed**. `DynamicViewLoader.test.tsx` + `src/spatial` → **4 files, 57 passed**. `packages/core src/testing/eliza-package-paths.test.ts` → **4 passed**.

**`eliza-package-paths.ts`** prepends the in-repo `packages/agent`, `packages/app-core`, `packages/shared` roots ahead of the existing `eliza/…` and `../eliza/…` forms, matching what `@elizaos/ui` already did. Order matters here and the in-repo root correctly comes first.

**`onboarding-chat.ts`** is the change I'd call out as the best of the batch. Racing `stub.fetch` against `abortPromise` left a late-arriving response body with no owner; moving the buffering inside a single `.then` and cancelling `response.body` on the failure path means a header that arrives after the caller has timed out still gets its transport released. The `!response.body.locked && !response.bodyUsed` guard before `cancel` is the right precondition, and the `cancel` itself is wrapped so a cancellation failure cannot replace the original rejection.

**`maxFilesPerTestBatch`** returning `1` on Linux is a real throughput cost, but it is scoped to the platform with the pinned Bun issue, is extracted into a testable pure function, and the `timeout-minutes` bumps (60 → 90 and 60 → 120) are raised in the same commit rather than left to fail later.

### Not verified

- `packages/app/test/ui-smoke/*.spec.ts` — Playwright, not run locally.
- The plugin view tests (`FinancesView`, `GoalsView`, `HealthView`) and `packages/synthetic-world` pglite suite — read, not run.
- `bun test packages/scripts/__tests__/` in aggregate reports 10 failures across 168 files, which I did **not** attribute. On this repo batched `bun test` runs routinely report failures that vanish when the files are run individually, and none of the batch failures were in files this PR touches, so I am reporting it as unattributed rather than as a finding either way.

### Non-blocking observations

1. **`surfaceBackdropFilter` is entirely unasserted.** No test in `ChatOverlay.morph.test.tsx` or `chat-overlay-motion.test.ts` references the backdrop filter. The saturation property it depends on is well covered, but the transform itself — including the `t >= 1` boundary that the whole optimization rests on — would survive being inverted. A two-line assertion on the transform would make the next reviewer's job much cheaper than mine was.

2. **`buildFieldSpec` now merges rather than replaces.** `agent || props.label ? { id: agent?.id ?? …, role: …, label: props.label, value: …, ...agent } : undefined` replaces `normalizeAgent(props.agent) ?? (props.label ? … : undefined)`. Because `...agent` spreads last, an explicitly-provided agent carrying `label: undefined` now overwrites the synthesized `props.label` with `undefined`, where previously the caller's agent was simply used verbatim. Almost certainly fine for current callers; worth a thought if agents are ever constructed with explicit-undefined fields.

3. **`smoke-view-declarations.mjs`** gives `goals` a `"gui"` presentation and `capabilities: ["agent-surface"]` while every sibling entry keeps the 5-tuple form. If that capability applies to other agent-surface views, they will need the same widening; if it is genuinely goals-only, ignore this.

Approving. The partition split, the excluded-suite pairing, and the onboarding body ownership are all done in the careful way, and the ChatOverlay reasoning that I previously pushed back on holds up under checking.

— [claude-code-eliza]

agent-cortex (approved):
## APPROVE — independently reproduced the testable surface in a disposable sandbox at head `91132ed`

Reviewed the full diff at head `91132ed12cea05dcad954bbd47f0b73b4ceb44b4` (merge base `6065de5d34655a6f66d059418085a627af57f200`), 33 files, +766/-253. Rather than read-only, I ran the bun/vitest-testable portions offline (network-denied) and added a mutation probe on the native-control migration that is the heart of the "native control interactions" claim.

### What I verified by execution

```
core     src/testing/eliza-package-paths.test.ts                          4 pass
scripts  __tests__/ci-server-partitions.test.mjs                          1 pass (8 expect)
cloud/shared eliza-app/onboarding-chat.error-policy.test.ts              20 pass (53 expect)
ui (vitest)  DynamicViewLoader.test.tsx + src/spatial                    57 pass
plugin-goals (vitest) test/GoalsView.test.tsx                             7 pass
```

- **`onboarding-chat.ts` (the strongest change).** The 20-case error-policy suite passes, including the two cases that exercise exactly the new J6 body-ownership logic — `cancels a response when its headers arrive after the deadline` and `releases a late response after caller cancellation has already returned`, plus `preserves the caller reason when stream cancellation rejects differently`. Moving the buffering inside a single `.then` and cancelling `response.body` on the failure path (guarded by `!response.body.locked && !response.bodyUsed`, with the `cancel` wrapped so a cancellation failure cannot replace the original rejection) closes a real leak where a header arriving after the caller timed out left an unread body with no owner. Confirmed by test, not just by reading.
- **`ci-server-partitions.test.mjs`** pins the two `tests_server` matrix filters against the real planner: the `^@elizaos/agent \(packages/agent\)#test$` anchor and its negative-lookahead complement are provably disjoint and exhaustive (8 assertions, pass). The `.mjs`+`bun:test` file is collected by `bun test` (it will not run under `node --test`).
- **`eliza-package-paths.ts`** correctly prepends the in-repo `packages/agent|app-core|shared` roots ahead of the `eliza/…` and `../eliza/…` forms so a requested worktree wins over a sibling checkout; the 4 worktree-isolation tests pass.

### Mutation probe — the native-select migration is genuinely guarded

The core "native control interactions" change is `spatial/primitives.tsx` swapping the Radix `Select` for a real `<select>` (`NativeSelect`) and moving `agentDataProps`/`onChange` onto that element so an agent-fill hits the same change handler as a user selection. I removed the new `onChange={(e) => props.onChange?.(e.target.value)}` and re-ran the suite:

```
DynamicViewLoader.test.tsx  with onChange removed  ->  1 failed | 44 passed (45)
                            head (unmodified)      ->  57 passed
```

So the agent-fill → change-handler path the migration depends on is asserted, not merely present. `buildFieldSpec` now spreads `...agent` last and sets `role: props.kind === "select" ? "select" : "text-input"`; the placeholder guard `!spec.options?.includes("")` is correct when `options` is undefined.

### Not verified (execution proof blocked in the sandbox)

- **`ChatOverlay.tsx` backdrop-filter change** — this is a compositor/frame-budget change (`--chat-sheet-backdrop-filter` becomes a `MotionValue` returning `"none"` while `surfaceBlackout >= 1`). I traced the saturation logic and it matches the two prior reviews' conclusion (blackout pins at 1 across maximize/restore, so the filter only returns below the half detent where the fill is genuinely translucent), and the committed `ChatOverlay.morph.test.tsx` assertions pin `sheetBlackoutProgress` saturation — but a jank/compositor claim cannot be proven in a headless sandbox, so I neither confirm nor refute the frame-budget trade beyond the reading.
- `packages/app/test/ui-smoke/*.spec.ts` (Playwright) and the CI matrix against a real GitHub runner — not runnable here.

### Non-blocking observations

1. **The `compact` native-select variant is not a pure reproduction of the inlined `WorkflowTriggerPanel` classes.** `native-select.tsx` `compact` is `"h-8 w-full rounded-md bg-background px-2 text-base sm:text-xs"`, and the cva **base** contributes `min-h-10 border border-input px-3 text-sm text-txt …`. After `tailwind-merge`, the rendered control keeps base `min-h-10` (2.5rem) alongside `h-8` (2rem) → effective height ~40px, and gains `w-full`. The old inline classes on those selects were `h-8 min-w-28 …` (32px, auto width, no `min-h`/`w-full`). Net: those `WorkflowTriggerPanel` dropdowns render ~8px taller and full-width now. Almost certainly cosmetic/intended normalization, but it does change their visual footprint, so a before/after screenshot would settle it — this is the one place the "reproduces the inlined classes it replaces" description isn't literally true.
2. **UI surfaces ship without before/after visual evidence.** The spatial select, `WorkflowTriggerPanel`/`VaultInventoryPanel`, and `ChatOverlay` are rendered-UI changes; per the evidence rubric these warrant before/after screenshots. The behavioral halves are test-covered (above), so this is a soft note, but the visual footprint change in (1) is exactly what a screenshot would have caught.
3. **`buildFieldSpec` now merges rather than replaces.** Because `...agent` spreads last, an explicitly-provided agent carrying `label: undefined` would overwrite the synthesized `props.label`. Fine for current callers; worth a thought if agents are ever built with explicit-undefined fields.

### Scope note

This is a broad, multi-area diff (CI partitioning, test batching, spatial native-select, onboarding-chat leak fix, package-path resolution, ChatOverlay compositor tuning). Each hunk I could execute is individually sound and covered; I'd gently suggest such cross-cutting batches are easier to land and bisect when split, but nothing here is incorrect on that basis.

Approving on the strength of the reproduced test surface and the mutation probe confirming the native-control fix is asserted. The only items I could not prove are the compositor/Playwright surfaces, which are covered by the PR's own green lanes and the prior morph-test assertions.

---
Contribution skill revision: elizaOS/eliza@b676280384b15df54204276e34bb8740db26936f:packages/skills/skills/contribute-to-eliza
— [eliza-swarm]

NubsCarson (changes requested):
The separate full CI run on this exact head is failing despite the green PR Static Smoke summary: https://github.com/elizaOS/eliza/actions/runs/34021891133.

The agent partition runs for 2,675,129 ms and exits 0, but the outer runner records `observed:false`, `counts:null`, `reports=0 tests=0 executed=0 unobserved-tasks=1`, then exits 3 under its vacuous-green guard. The other full-CI jobs passed. This is missing testcase evidence, not a reproduced failing agent assertion.

The current agent test command is `node scripts/run-vitest-batches.mjs`. The outer `structuredEvidenceKind` supports direct Vitest and its single-run wrapper, but does not recognize this batch supervisor; the supervisor also does not accept/aggregate the requested JUnit output. The new agent-only partition therefore cannot provide the evidence needed to pass its aggregate gate.

Please connect the batch supervisor to the canonical structured testcase-evidence contract, with a regression through the real parent runner and multiple batches. Preserve the guard: treating exit 0 or a printed file count as executed testcase evidence would conceal the failure. Then rerun full CI on the repaired exact head. I reviewed the current source and downloaded full-CI logs; I have not rerun the 44-minute agent suite locally.

NubsCarson:
Review follow-up on `91132ed12cea05dcad954bbd47f0b73b4ceb44b4`: I have a local repair for the missing agent-batch testcase evidence. It forwards the parent JUnit request to separate child reports, validates and aggregates the actual testcase trees, and recognizes the batch supervisor in the parent runner. It preserves the zero-executed-tests guard.

The old runner reproduced the reported failure through two actual Vitest batches. The repaired parent suite passes 23 cases, including all-skipped rejection and failed-batch propagation; 11 batch/boot cases pass after syncing with develop `f7a8a24eef292cf1b73f71b11e1ca3b795b21ade`. Full repository verification is queued behind the shared-host validation owner; this is not yet a merge-readiness receipt. I will publish the repair after that gate and recheck hosted full CI.

NubsCarson:
I found the author validation branch `fix/deployment-validation-d261f1-30552` at `d261f1aed55e7734551420d00e7e5869341647bc`, which already includes the batch-evidence repair `bc4262fa301584559c09f48441f871f20a03c926`. I am preserving my overlapping local implementation without publishing it while reviewing this existing candidate.

[Full CI34051631035](https://github.com/elizaOS/eliza/actions/runs/34051631035) has passed Quality and the scripts lane, but [Tests(server,remaining)](https://github.com/elizaOS/eliza/actions/runs/34051631035/job/101536315418) failed `packages/evidence/src/analyzers/keyframes-terminal.test.ts`: the sparse-tail terminal-frame case expected the green channel to exceed220 and received0. This extractor and test are unchanged between the validation candidate and current develop `2eb52ff20623c8d1d407c8b03b3ee10be5c6a9b9`.

I ran the four focused tests on macOS with both system FFmpeg9.0.1 and bundled FFmpeg6.0; both passed. That does not clear the Linux failure or establish its cause. The actual Linux final-frame fixture/output needs investigation; the failing assertion should remain intact. The candidate also needs reconciliation with current develop before final PR-head validation.

NubsCarson:
### Sparse final-frame failure: diagnostic for the original Linux path

The failing server job [101536315418](https://github.com/elizaOS/eliza/actions/runs/34051631035/job/101536315418) at `d261f1aed55e7734551420d00e7e5869341647bc` proves that the sparse-tail output has green channel 0; it does not show the original video frames or the other output channels. The other three terminal-frame tests passed. Both a fixture-generation difference and seek-dependent extraction remain possible. No extractor fix or weaker assertion is justified by the available evidence.

Below is a self-contained probe retaining the original fixture command and both extraction paths. Save as `keyframes-linux-diagnostic.py`, then run it **on Linux with the actual FFmpeg binary resolved by the failing test**, using a fresh directory:

```sh
python3 keyframes-linux-diagnostic.py --ffmpeg /actual/resolved/ffmpeg --output /tmp/keyframes-linux-diagnostic-UNIQUE
```

Optionally add `--ffprobe /actual/resolved/ffprobe`. It installs nothing, makes no network calls, retains every command and raw output, and exits nonzero on command failure. Preserve the whole output directory, including on failure. `report.json` records complete sequential frame hashes/pixels and comparisons with near-EOF and no-seek output. If the original sequential frames do not end green, investigate generation; if they do but the seek output differs, investigate extraction. Matching decoded bytes would instead direct investigation to the image-reading boundary.

I executed this exact probe locally with macOS FFmpeg 9.0.1: it completed, source frames were `[254,0,0]` then `[0,255,1]`, and both extraction outputs matched the final sequential frame hash. This validates the probe's local execution only; Linux reproduction remains outstanding. Script SHA-256: `ba3edf2937966066da1182fcd774cf239dfd646531b6491dc090bfd8f9150464`.

<details><summary>Complete diagnostic script</summary>

```python
#!/usr/bin/env python3
"""Preserve sparse-video generation and extraction observations without claiming a fix."""
import argparse
import hashlib
import json
import pathlib
import shutil
import subprocess
import sys

parser = argparse.ArgumentParser(description=__doc__)
parser.add_argument('--ffmpeg', required=True)
parser.add_argument('--ffprobe')
parser.add_argument('--output', required=True)
args = parser.parse_args()
out = pathlib.Path(args.output).absolute()
if out.exists() or out.is_symlink():
    parser.error('Output must be a fresh directory')
out.mkdir(parents=True)
records = []
report = {'status': 'running', 'scope': 'Diagnostic observations only; no correctness or fix assertion', 'steps': records}

def save():
    (out / 'report.json').write_text(json.dumps(report, indent=2) + '\n')

def sha(path):
    digest = hashlib.sha256()
    with path.open('rb') as stream:
        for block in iter(lambda: stream.read(1024 * 1024), b''):
            digest.update(block)
    return digest.hexdigest()

def binary(value):
    resolved = shutil.which(value)
    if resolved is None:
        raise RuntimeError('Explicit binary is unavailable: ' + value)
    path = pathlib.Path(resolved).resolve(strict=True)
    return {'requested': value, 'path': str(path), 'sha256': sha(path)}

def run(label, command, stdout_name=None):
    record = {'label': label, 'argv': command, 'status': 'running'}
    records.append(record)
    save()
    stdout = out / (stdout_name or label + '.stdout.log')
    stderr = out / (label + '.stderr.log')
    try:
        with stdout.open('xb') as output, stderr.open('xb') as errors:
            result = subprocess.run(command, stdout=output, stderr=errors, timeout=120)
        record['exitCode'] = result.returncode
        if result.returncode != 0:
            raise RuntimeError(label + ' command failed')
        record['status'] = 'completed'
    except BaseException as error:
        record.update(status='failed', errorType=type(error).__name__, error=str(error))
        raise
    finally:
        record['artifacts'] = [{'path': str(path), 'sha256': sha(path), 'bytes': path.stat().st_size} for path in (stdout, stderr) if path.is_file()]
        save()

def rgb_observations(path):
    data = path.read_bytes()
    frame_size = 64 * 64 * 3
    frames = []
    for offset in range(0, len(data) - len(data) % frame_size, frame_size):
        frame = data[offset:offset + frame_size]
        frames.append({'index': len(frames), 'sha256': hashlib.sha256(frame).hexdigest(), 'firstPixelRGB': list(frame[:3]), 'channelMin': [min(frame[channel::3]) for channel in range(3)], 'channelMax': [max(frame[channel::3]) for channel in range(3)]})
    return {'bytes': len(data), 'frameBytes': frame_size, 'trailingBytes': len(data) % frame_size, 'frames': frames}

try:
    save()
    report['ffmpeg'] = binary(args.ffmpeg)
    ffmpeg = report['ffmpeg']['path']
    run('ffmpeg-version', [ffmpeg, '-version'])
    source = out / 'sparse-tail.mp4'
    # Argument order and default encoder match makeVideo('sparse-tail', ..., true).
    run('generate', [ffmpeg, '-hide_banner', '-loglevel', 'error', '-f', 'lavfi', '-i', 'color=c=red:s=64x64:r=10:d=0.1', '-f', 'lavfi', '-i', 'color=c=lime:s=64x64:r=10:d=0.1', '-filter_complex', '[0:v][1:v]concat=n=2:v=1,setpts=N*5/TB[v]', '-map', '[v]', '-fps_mode', 'vfr', '-pix_fmt', 'yuv420p', str(source)])
    run('sequential', [ffmpeg, '-hide_banner', '-loglevel', 'info', '-i', str(source), '-map', '0:v:0', '-vf', 'showinfo', '-fps_mode', 'passthrough', '-pix_fmt', 'rgb24', '-f', 'rawvideo', 'pipe:1'], 'sequential.rgb')
    report['sequential'] = rgb_observations(out / 'sequential.rgb')
    for label, seek in [('near-eof', ['-sseof', '-1']), ('no-seek', [])]:
        png = out / (label + '.png')
        # Exact production extraction args; diagnostics use separate decoding.
        run(label, [ffmpeg, '-hide_banner', '-loglevel', 'error', *seek, '-i', str(source), '-fps_mode', 'passthrough', '-update', '1', str(png)])
        report[label] = {'pngPresent': png.is_file()}
        if png.is_file():
            run(label + '-decode', [ffmpeg, '-hide_banner', '-loglevel', 'info', '-i', str(png), '-vf', 'showinfo', '-fps_mode', 'passthrough', '-pix_fmt', 'rgb24', '-f', 'rawvideo', 'pipe:1'], label + '.rgb')
            observed = rgb_observations(out / (label + '.rgb'))
            report[label]['rgb'] = observed
            source_frames = report['sequential']['frames']
            report[label]['matchesSequentialLastFrame'] = (observed['frames'][0]['sha256'] == source_frames[-1]['sha256']) if len(observed['frames']) == 1 and source_frames else None
    if args.ffprobe:
        report['ffprobe'] = binary(args.ffprobe)
        probe = report['ffprobe']['path']
        run('ffprobe-version', [probe, '-version'])
        run('ffprobe', [probe, '-v', 'error', '-show_frames', '-show_streams', '-show_format', '-of', 'json', str(source)], 'ffprobe.json')
    report['status'] = 'completed-observations'
except BaseException as error:
    report.update(status='failed', errorType=type(error).__name__, error=str(error))
    raise
finally:
    report['files'] = [{'name': path.name, 'bytes': path.stat().st_size, 'sha256': sha(path)} for path in sorted(out.iterdir()) if path.is_file() and path.name != 'report.json']
    save()

```
</details>

NubsCarson:
Additional evidence from authored candidate d261f1aed55e7734551420d00e7e5869341647bc, full run 34051631035: completed plugins job 101536315438 fails plugin-workflow, smithers-worker-lifecycle.test.ts:258. The 1,048,577-byte stdout case expected SMTHRS_PROTOCOL_OVERFLOW but received SMTHRS_RESULT_MISSING. The same focused test passes locally on macOS (Bun 1.3.14, 1 pass / 12 filtered), so that is not Linux acceptance. Source inspection suggests an exit/drain race: the data handler pauses stdout, refuses to resume after processExited, and terminate ignores overflow after exit. This is a hypothesis pending a deterministic reproduction, not a confirmed root cause. The worker fixture does not explicitly exit after its write. Keeping this separate from the terminal-video failure in server/remaining; no test expectations have been weakened.

NubsCarson:
Positive exact-candidate CI receipt: Tests (server, agent), job 101536315428 in run 34051631035, passed. Parent evidence reports 10,369 tests / 10,346 executed / 23 skipped, zero failures/errors, observed=true, and unobserved=0. This validates the authored JUnit aggregation repair on d261f1aed55e7734551420d00e7e5869341647bc and retires the earlier missing-test-report blocker for that candidate. It does not clear the separate server/remaining terminal-frame or plugin-workflow failures, nor update the actual PR head.

NubsCarson:
Current-head follow-up for `e2f78d2bb72dd65e8a2d42b2cf07e4b5591738f4`: [full CI 34071740596](https://github.com/elizaOS/eliza/actions/runs/34071740596) is now terminal **failure**. I downloaded and inspected the failed-job logs.

- Scripts: gateway-webhook, Headscale and tunnel-proxy contract tests still expect `develop` for staging, whereas the changed workflows require `staging`. The gateway workflow README also retains the old mapping. I have a four-file correction locally; it preserves exact branch/SHA/environment checks and has source review, but execution and publication are pending.
- Agent: `src/api/snapshot-route-transient.test.ts` fails teardown with `ENOTEMPTY` in its temporary state directory; `test/api/agent-message-route.test.ts` is also reported failing. The run exits 1, so the batch-reporting repair does not yet establish a passing full agent lane.
- Client: `test/ios-app-intents-registration.test.ts` and `src/services/local-inference/catalog.test.ts` fail.
- Plugins: `plugin-goals/test/goals.real-db.test.ts` fails.
- Remaining server: scenario-runner `legacy-deterministic-scenario.integration.test.ts` and `required-services.test.ts` fail.
- Android: the renderer guard rejects `navigator.serviceWorker` in `vendor-crypto`; this needs dependency-path diagnosis, not removal of the guard.
- UI smoke: interaction observations, voice-preview/settings checks, and the 18 MB raw-JavaScript ceiling fail; observed payload is 19.66 MB.

These are separate from the already reported stale static-smoke trigger expectation, the full audit OCR failure, and the unresolved collapse-drag backdrop-filter review. No CI exception or merge clearance is claimed.

NubsCarson:
The stale staging-branch contract repair now passes all three affected protected-deployment suites: **45 passed, 0 failed** (gateway 8, Headscale 32, tunnel proxy 5). Run used Bun 1.3.14 and Bash 5.3.15. The initial macOS Bash 3.2 run failed two associative-array fixtures; no assertions or workflow scripts were weakened to address that host mismatch.

[Repair patch against published e2f](https://github.com/elizaOS/eliza/releases/download/pr-evidence-10/30773-staging-contract-repair.patch) includes the previously checked main static-smoke trigger expectation and the three protected-deployment tests plus the stale gateway README mapping. Local candidate is `3bb3630c7f3d0fbbbd4af8e6ffc6a3cab8320257`; independent source review found no blocking issue in these corrections.

Evidence: [gateway log](https://github.com/elizaOS/eliza/releases/download/pr-evidence-10/30773-3bb-bash5-gateway-webhook-deploy-workflow.log), [Headscale log](https://github.com/elizaOS/eliza/releases/download/pr-evidence-10/30773-3bb-bash5-headscale-arm-workflow.log), [tunnel log](https://github.com/elizaOS/eliza/releases/download/pr-evidence-10/30773-3bb-bash5-tunnel-proxy-deploy-workflow.log), [receipt with log hashes](https://github.com/elizaOS/eliza/releases/download/pr-evidence-10/30773-staging-contract-repair-receipt.json).

The repair has not been pushed to this PR. Current-base install/root verification and publication are pending; the other full-CI failures, UI audit, collapse-drag review, and native/deployment acceptance remain separate gates.

hermesagent270-commits (approved):
**Scope note.** 118 files and ~3,000 additions is well beyond one honest pass. I reviewed `packages/scripts/reviewed-branch-promotion.mjs` — the gate that decides what reaches `main` — and read its test file. The other ~115 files, including the Cloudflare preflight and the effect ledger, are unread. I did not run the suites.

The gate is well built, and its threat model is the part worth confirming rather than assuming.

## What I verified holds

**Fork masquerade is blocked in all three places it could matter.** Every PR-shaped input is required to be same-repo:

```js
Number.isSafeInteger(pr.head?.repo?.id) && pr.head.repo.id === pr.base?.repo?.id
```

That appears in `verifyMergedPromotion`'s match filter, in the closed-history loop, and in the open-PR search. A PR from a fork whose branch is also called `develop` fails all three, and a deleted fork (`head.repo === null`) fails the `isSafeInteger` guard rather than throwing. Fails closed in both directions. Your test `"a fork branch with the same name cannot be updated as the canonical promotion"` pins it.

**The tree binding is the real teeth.** Requiring `source.tree.sha === target.tree.sha` means the promotion merge commit must carry exactly the reviewed tree — a merge that resolved anything, or picked up a commit that wasn't in the reviewed source, cannot pass. That's a much stronger claim than "a merge PR existed", and it's what makes `"a direct destination push cannot claim reviewed promotion authority"` true rather than aspirational.

**Both pagination boundaries fail closed**, which is the unusual choice and the right one:

```js
if (!Array.isArray(pulls) || pulls.length === 100) throw new Error("ambiguous promotion history");
// A full page may hide an existing request. Fail rather than create a duplicate.
if (pulls.length === 100) throw new Error("promotion pull request search requires pagination");
```

A full page means "there may be more", and the code refuses rather than acting on a possibly-truncated view. Most code in this shape silently takes the first page.

**No ref writes and no merging** — the script only opens or patches a PR and hands off, with the body stating a maintainer must approve. The gate cannot promote itself.

## The consequence worth documenting

The tree-equality rule means **any commit landed directly on `staging` or `main` blocks every subsequent promotion** until the branches are reconciled, because the merge tree can then never equal the reviewed source tree. That is the correct behaviour — it's precisely the property that stops a direct push from laundering itself into a "reviewed" promotion — but the failure surfaces as:

```
promotion merge tree differs from reviewed source
```

An on-call engineer who has just hotfixed `main` at 3am will read that as a tooling bug, not as "your hotfix must be merged back before promotion can resume". Naming the remedy in the message — the same way `unavailable()` elsewhere in this codebase says "No replacement key was returned" — would turn a confusing failure into a self-explaining one. Same for `"completed promotion changed the reviewed source tree"`.

## One inconsistency in defensive posture

The closed-history loop re-checks `prior.base?.ref !== targetBranch` client-side even though the query already constrains `base=${targetBranch}`. The open-PR search does not:

```js
const matches = pulls.filter(
  (pr) => pr.head?.ref === sourceBranch && /* same-repo */ …
);
```

Both are safe — the server filter is in the query string in each case — but the file treats the server's `base` filter as untrusted in one place and trusted in the other. Given how carefully everything else here is belt-and-braces, I'd make them agree, and adding the check is the cheaper direction.

Approving on the promotion gate. The design refuses to act on ambiguity at every branch point I checked, and the test file exercises the adversarial cases rather than only the happy path.

```slop-review
{"schemaVersion":"2","projectId":"eliza","artifactUrl":"https://github.com/elizaOS/eliza/pull/30773","headSha":"e2f78d2bb72dd65e8a2d42b2cf07e4b5591738f4","provider":"anthropic","model":"claude-opus-5[1m]","client":"Claude Code","runId":"","traceSha256":"","recommendation":"accept","reproduced":false,"securityRisk":"none","duplicateRisk":"none","splitRisk":"medium","effortBand":"large","complexity":"complex","impact":"significant","reviewLoad":"deep","recommendedTier":"large","recommendedThirds":24,"workUnitId":"wu_eliza_30773_deployment_qualification","confidenceBasisPoints":8500,"valueRationale":"Partial review - the reviewed-branch-promotion gate and its tests only; ~115 of 118 files unread, suites not run, stated plainly. Verified the gate's threat model rather than assuming it: same-repo enforcement (head.repo.id === base.repo.id) appears in all three PR-shaped inputs so a fork branch named develop cannot masquerade, and a deleted fork fails the isSafeInteger guard rather than throwing; the source.tree.sha === target.tree.sha binding is what makes 'a direct destination push cannot claim reviewed promotion authority' actually true; and both pagination boundaries throw on a full page rather than acting on a truncated view. Surfaced the operational consequence that tree-equality permanently blocks promotion after any direct commit to staging or main until reconciled - correct by design, but the error text does not say so. Noted one inconsistency: the closed-history loop re-checks base.ref client-side while the open-PR search trusts the same server-side filter.","usefulArtifacts":["same-repo check head.repo.id === base.repo.id enforced in verifyMergedPromotion, the closed-history loop, and the open-PR search","deleted fork yields head.repo === null and fails Number.isSafeInteger rather than throwing - fails closed","tree equality source.tree.sha === target.tree.sha binds the merge commit to the reviewed tree","both pagination checks throw on pulls.length === 100 rather than acting on a possibly truncated page","script only opens or patches a PR - no ref writes, no merge - and the body requires maintainer approval","tests cover fork-name masquerade, direct destination push, develop->main shortcut, stale source, and idempotent retry"],"commands":["git show pr30773:packages/scripts/reviewed-branch-promotion.mjs","git show pr30773:packages/scripts/__tests__/reviewed-branch-promotion.test.ts | grep -n 'test('","gh api /repos/elizaOS/eliza/pulls/30773/files --paginate --jq '.[] | \"\\(.additions + .deletions)\\t\\(.filename)\"' | sort -rn"],"evidenceUrls":["https://github.com/elizaOS/eliza/pull/30773"],"summary":"Approve on the promotion gate (partial review, ~115 of 118 files unread). Verified fork masquerade is blocked at all three inputs, the tree binding is what makes the direct-push rejection real, and both pagination boundaries fail closed. Flagged that tree-equality blocks all promotion after a direct hotfix to staging or main without the error text saying so."}
```

hermesagent270-commits (approved):
Exact-head review at `7daa224db80c7128632f1752dec1fb0beb6c3981` (39 commits on `develop` `d7df7af187f`), on a clean merge into `develop`. My 2026-09-08 approval at `e2f78d2bb72d` was a static read of the promotion gate only; this pass executes the changed surface. Since that head the delta grew from 118 to 176 files (59 added, 1 removed, 70 common files changed), so I treated it as a new review rather than an interdiff.

## Executed on the merged tree

Every test file the PR adds or modifies, run with the runner it targets:

| Surface | Result |
| --- | --- |
| `packages/scripts` 21 files (`bun test`) | 230 pass, 0 fail (six certification cases need `zip` on PATH; green once installed) |
| `packages/cloud` 14 `bun:test` files | 261 pass, 1 skip, 1 env fail (`group-h-misc` e2e needs a Worker on :8787; 82 skip with `REQUIRE_E2E_SERVER=0`) |
| `packages/cloud` node/vitest files | release-admission 13/13, latency-provenance 2/2, standing-capability 10/10, messaging-gateway-preflight 13/13, cors 9/9 |
| `packages/ui` 15 files (vitest) | 437 pass |
| `packages/app` ios-app-intents, `app-core` scripts, `plugin-workflow`, `core`, `agent`, `plugin-health`, `plugin-goals`, `plugin-finances`, `synthetic-world`, `scenario-runner`, `evidence`, `elizaos` | all green (22, 7+31, 53, 48, 68, 8, 7, 16, 8, 10, 4, 16) |
| `plugin-personal-assistant` `definition-idempotency.pglite.integration` (src-integration config) | 18/18 |
| `packages/app` ui-smoke Playwright, the four changed specs on chromium | 15 passed, 3 failed (all three Calendar; see below), 1 skipped (needs live cloud), 6.3 min |
| `bun run check:agents-claude` (new `packages/cloud` guide pair) | 157 pairs identical |
| Hosted checks at head | 5/5 success |

## Claims I verified against the real thing

**Coinbase digest.** `android-cloud-routing-markers.mjs` pins `373bcdaea9…75a4` as the SHA-256 of the whole analytics script. I imported `@coinbase/wallet-sdk@4.3.6` `dist/core/telemetry/telemetry-content.js` and hashed `TELEMETRY_SCRIPT_CONTENT`: 83,300 bytes, digest matches exactly, and the literal contains `navigator.serviceWorker`, so the audited-literal path is reachable.

**Secret-scanner exception.** With the CI-pinned gitleaks 8.30.1 (installed via `.github/scripts/install-gitleaks.sh`) scanning the PR's 39 commits under the new `.gitleaks.toml`: no leaks. Under `develop`'s config the same range reports 1 leak (`coinbase-access-token`), which is the finding the allowlist exists for. Negative controls in a scratch git repo scanned the way every CI lane scans (`--log-opts`): the exact digest in its owning file is allowed; a different 64-hex value in that file is detected; the same digest in a sibling file is detected. One caveat for anyone reproducing locally: in `--no-git` directory mode gitleaks 8.30.1 skips the allowlisted file entirely (0 bytes scanned), so a local `gitleaks dir` run would not see a changed value there. No CI lane uses that mode, so this does not weaken the gate, but it is worth knowing before trusting a directory scan.

**Production certification.** Read the new `production-effect-certification.mjs` in full. Reads only (`request` throws on any non-GET), artifact inventory fails closed at 100 entries, each candidate artifact is bound to a completed successful `staging` push/dispatch run of the certification workflow in the same repository, the archive digest is checked before `unzip -p`, and the CLI refuses unless `GITHUB_REF_NAME` is `main` and the checkout HEAD equals the requested SHA. The workflows call it after admission and again immediately before each effect: `build-agent-image.yml` runs it before every `docker push` on `main`, and the two worker deploys pass `expires_epoch` to the host step, which rejects mutation once `date +%s` passes it. `determine-env` maps `production` to `main` and nothing else, so the certification step's `environment == 'production'` guard and the host's `DEPLOY_BRANCH = main` guard cannot disagree.

**Removed wallet tests.** `WalletSectionNav.price-surface.test.tsx` is deleted and the header case in `WalletSectionNav.test.tsx` is dropped. `WalletSectionNav.tsx` is untouched by this PR, so I ran `develop`'s versions of both test files on a pristine `develop` checkout: 5 of 13 cases already fail there (`Unable to find … role "heading" and name "Wallet"`, `[data-testid="view-header"]`), and `wallet-section-price-surface` no longer exists outside an e2e script, so the removed assertions were red on `develop` and vacuous respectively. The PR repairs develop's red state rather than hiding a regression it introduced.

## One thing the hosted lane will hit: the three Calendar ui-smoke cases

Running the four changed ui-smoke specs on chromium (fresh `build-views` of all 22 view targets, Node 24.15.0) fails the three calendar cases after their 60 s wait on `getByTestId('lifeops-calendar-section')`, while the other 15 cases pass. The captured accessibility tree shows the view rendered and populated (`main "Calendar. 2 events"`, `region "Calendar month"`, day buttons with events), so the route, stub endpoints and bundle work; the anchor is what is missing. `calendar-view-bundle.ts` exports `CalendarPage`, which mounts `SimpleCalendarView`, and `lifeops-calendar-section` exists only in `CalendarSection.tsx`, which no built view bundle contains (`grep -c` over every `plugins/*/dist/views/bundle.js` is 0).

This is not something this PR introduced: the bundle entry, `CalendarPage`, and `CalendarSection` are untouched here (the one `plugin-calendar` change is a CSS colour), and the calendar expectations in both specs are line-identical to `develop`. It predates the PR (the bundle has mounted the simple view since #29151), so I filed it as #31007 rather than blocking on it. It does matter for the merge condition in the body ("Hosted checks must finish successfully for the published SHA"): the only workflow that has run at `7daa224` is PR Static Smoke; `ci.yml`'s deterministic e2e shards have not run at this SHA, and if the change filter marks that lane affected, these three cases will fail there until #31007 is resolved one way or the other.

## Minor

- Biome (repository config) reports formatting drift in the five `.github/*.json` effect and ruleset files; three of them already fail on `develop`, and no repository lint task covers `.github/`, so this is a nit rather than a gate.
- The `.github` and `packages/app` visual claims (danger/muted/choice states across four themes) are not something I could verify here; I did not run `audit:app`. The Playwright specs above are the part of the app acceptance story I could execute.

**Recommendation: merge** (approve on the executed surface; #31007 is a pre-existing develop defect to settle separately, and the hosted `ci.yml` lane still has to run at this SHA per the body's own condition).

```slop-review
{"schemaVersion":"2","projectId":"eliza","artifactUrl":"https://github.com/elizaOS/eliza/pull/30773","headSha":"7daa224db80c7128632f1752dec1fb0beb6c3981","provider":"anthropic","model":"claude-fable-5-1","client":"Claude Code","runId":"","traceSha256":"","recommendation":"merge","reproduced":true,"securityRisk":"low","duplicateRisk":"none","splitRisk":"medium","effortBand":"large","complexity":"complex","impact":"significant","reviewLoad":"deep","recommendedTier":"large","recommendedThirds":24,"workUnitId":"wu_eliza_30773_deployment_qualification","confidenceBasisPoints":9000,"valueRationale":"Exact-head execution review of a 176-file maintainer PR (delta grew 118→176 since my earlier static approval). Ran every changed test file with its own runner on the merged tree (scripts 230, cloud 261+47, ui 437, app/app-core/workflow/core/agent/plugins all green, PA pglite integration 18/18, guide parity 157). Verified the Coinbase digest against the real @coinbase/wallet-sdk 4.3.6 telemetry script (83,300 bytes, exact match), the gitleaks exception with the CI-pinned 8.30.1 over the 39 commits plus git-mode negative controls (allowed / changed-value detected / other-file detected; noted the --no-git whole-file skip), the certification script and its workflow wiring (reads only, fail-closed inventory, freshness rechecked before docker push and host mutation, env↔branch mapping consistent), and that the removed wallet tests are red on pristine develop and vacuous. Ran the four changed ui-smoke Playwright specs: 15 pass, 3 Calendar cases fail because the calendar view bundle mounts SimpleCalendarView while the specs (unchanged from develop) wait for CalendarSection's test id; pre-existing since #29151, filed as #31007. Only PR Static Smoke has run at this SHA; the ci.yml e2e shards have not.","usefulArtifacts":["interdiff e2f78d2bb72d (118 files) → 7daa224db80c (176 files): 59 added, 1 removed (WalletSectionNav.price-surface.test.tsx), 70 common files changed; PR base == origin/develop d7df7af187f0","packages/scripts bun test 21 files: 230 pass after installing zip (6 certification cases need it)","packages/cloud: bun:test 14 files 261 pass/1 skip/1 env-fail (group-h-misc needs :8787; 82 skip with REQUIRE_E2E_SERVER=0); release-admission 13, latency-provenance 2, standing-capability 10, messaging-gateway-preflight 13, cors 9","packages/ui vitest 15 files 437 pass; app 22, app-core 7+31, plugin-workflow 53, core 48, agent 68, health 8, goals 7, finances 16, synthetic-world 8, scenario-runner 10, evidence 4, elizaos 16; PA definition-idempotency.pglite.integration 18/18 (src-integration config)","sha256(TELEMETRY_SCRIPT_CONTENT) of @coinbase/wallet-sdk@4.3.6 dist/core/telemetry/telemetry-content.js = 373bcdaea9…75a4, 83300 bytes, contains navigator.serviceWorker","gitleaks 8.30.1 (repo installer) --log-opts d7df7af..pr30773 with merged .gitleaks.toml: 39 commits, no leaks; with develop config: 1 leak (coinbase-access-token). Git-mode controls: exact digest in owning file allowed; changed value detected; same digest in sibling file detected. --no-git mode skips the allowlisted file entirely (scanned ~0 bytes)","develop's WalletSectionNav.test.tsx + price-surface test on pristine develop d7df7af: 5 failed / 8 passed (heading 'Wallet' and view-header absent); wallet-section-price-surface only in an e2e script","ui-smoke chromium, 4 changed specs: 15 passed, 3 failed, 1 skipped (6.3m). Failures: calendar decomposed view responsive modes; calendar mobile 390px; Calendar agent-bridge inventory — all waiting for lifeops-calendar-section; aria tree shows SimpleCalendarView populated. grep lifeops-calendar-section over plugins/*/dist/views/bundle.js = 0; calendar-view-bundle.ts exports CalendarPage since #29151","check-runs at 7daa224: 5 success, all from PR Static Smoke run 34469191674; no ci.yml run at this SHA","biome on changed files: 5 format errors, all .github/*.json (3 already fail on develop); no repository lint task covers .github/"],"commands":["git fetch origin +pull/30773/head:pr30773 && git checkout -B verify-merge origin/develop && git merge --no-commit --no-ff pr30773","cd packages/scripts && bun test <21 changed files>","cd packages/cloud && bun test <bun:test files>; node --test <node:test files>; bunx vitest run shared/src/lib/utils/cors.test.ts","cd packages/ui && bunx vitest run <15 changed files>","node -e hash TELEMETRY_SCRIPT_CONTENT from @coinbase/wallet-sdk@4.3.6","./.github/scripts/install-gitleaks.sh <dir> && gitleaks detect --config .gitleaks.toml --source . --redact --no-banner --log-opts d7df7af187f0..pr30773","cd packages/app && PATH=<node24>:$PATH ELIZA_NODE_PATH=<node24> ELIZA_UI_SMOKE_DISABLE_VIDEO=1 node scripts/run-ui-playwright.mjs --config playwright.ui-smoke.config.ts --project=chromium <4 changed specs>","gh api repos/elizaOS/eliza/commits/7daa224db80c/check-runs"],"evidenceUrls":["https://github.com/elizaOS/eliza/pull/30773","https://github.com/elizaOS/eliza/issues/31007","https://github.com/elizaOS/eliza/actions/runs/34469191674"],"summary":"Approve at the exact head after executing every changed test file, verifying the Coinbase digest, the gitleaks exception, and the production certification wiring against the real artifacts; the only red is three Calendar ui-smoke cases whose expectation predates this PR (filed #31007), and the hosted ci.yml lane has not yet run at this SHA."}
```

lalalune:
Current-source validation note for `38fdddc13c3f5c5b891ee558c47f6b6494275e70`: the PR body cites a full personal-assistant pass, but the current calendar provider contract test still expects `tool.strict === false` under Cerebras. On public develop `d7df7af187f05ec744b57c01799cfbabcc5fd325`, the focused file produces **6 passed / 2 failed**, both at line 42, because the provider now intentionally opts compatible planner tools into strict mode after #30990.

The calendar test, core `actions/to-tool.ts`, and OpenAI `models/text.ts` are unchanged between that base and this head. This is inherited test-contract drift, not a new navigation product regression. Please account for it before treating the earlier full-package receipt as current-head green.

#31013 owns the narrow test repair. It validates sparse and explicitly populated calendar arguments against the normalized provider schema and the actual action validator, while retaining invalid travel-type rejection. Both calendar owners/provider modes pass all eight focused cases, independent review is complete, and normal repository/full-package gates are underway. No production behavior change is needed for this finding.

```text
calendar-optional-tool-parameters.test.ts
Test Files  1 failed (1)
Tests       2 failed | 6 passed (8)
```

lalalune:
Reviewed and merged candidate `38fdddc13c3f5c5b891ee558c47f6b6494275e70` on `d7df7af187f05ec744b57c01799cfbabcc5fd325`.

Normal installation, repository verification (373 tests and audits), and the full UI suite (14,053 passed, 7 skipped across 1,307 files) passed. Source-bound visual, Android renderer, and static Storybook evidence is linked in the PR description. The final Choice regression exercises selection, locking, and once-only dispatch. The exact public SDK digest/path secret-scan exception passed negative controls and the 40-commit history scan. All seven final evidence assets were checked by size and SHA-256.

All five hosted checks passed for the exact head in [run 34493040651](https://github.com/elizaOS/eliza/actions/runs/34493040651). SHA-locked merge completed at 2026-09-10T15:54:03Z: `4985223ea31c01d6eec22c5d6aea7d79913ef896`. GitHub readback confirms parents `d7df7af187f05ec744b57c01799cfbabcc5fd325` and `38fdddc13c3f5c5b891ee558c47f6b6494275e70`, and tree `667bec21a036209f3b2dbd9ebd1a6466b3f857f2`, exactly matching the reviewed candidate.

Deployment/policy activation remains separately tracked in #30855. This merge does not claim that cutover is complete.

PR #30779, opened 2026-09-06
fix(cloud): consolidate A2A history validation and reject invalid Vast routing maps

### LINKED ISSUE #30755: fix(cloud): honor requested A2A response history consistently
## Problem and scope

Both A2A message/send handlers ignore configured positive historyLength; tasks/get accepts invalid counts inconsistently. Share boundary validation and nonmutating response projection. Preserve declared A2A v0.3 zero/unset compatibility and complete persisted/model context.

## Acceptance and evidence

Exercise protocol dispatch and compatibility handlers: positive limits affect responses only, subsequent unlimited reads retain all history, invalid counts return invalid params. Run cloud-shared tests/typecheck/lint and root verify.

Audited base: b676280384b15df54204276e34bb8740db26936f. This is a confirmed second-pass cleanup finding with independent plan review. No production deployment, credential changes, or funded transactions are part of this scope.

### LINKED ISSUE #30756: fix(cloud): reject malformed configured Vast endpoint maps
## Problem and scope

Malformed present VAST_ENDPOINTS_JSON and VAST_FALLBACK_MODEL_MAP_JSON values are silently treated as absent maps; dedicated endpoint mistakes route through global defaults and malformed object fields throw low-level errors. Validate endpoint and fallback shapes with actionable redacted errors.

## Acceptance and evidence

Exercise real resolver/provider factory and local HTTP selection; malformed present config fails before dispatch, missing optional maps and valid precedence/aliases remain compatible. Run cloud-shared and root gates.

Audited base: b676280384b15df54204276e34bb8740db26936f. This is a confirmed second-pass cleanup finding with independent plan review. No production deployment, credential changes, or funded transactions are part of this scope.

lalalune:
Closes #30755. Closes #30756.

A2A send/get now share validated, caller-requested history projection. Positive lengths select response history after complete persistence; zero and omitted lengths retain the existing unlimited-history contract. Stored history and individual messages remain complete.

Vast routing configuration rejects malformed endpoint and fallback maps with typed, redacted errors. Endpoint strings and both URL aliases must be absolute HTTP(S) URLs. The production provider-factory regression reproduces the previously accepted whitespace endpoint, rejects malformed URLs before dispatch, and verifies valid requests through a real local HTTP server. Per-model environment precedence and supported aliases remain intact.

Validated head: `816826a9810450e7fe38d16dd74357a468a1ed9e`, rebased onto `b8aa904a64a0fa80b5ee7a7167807b687fd935bd`.

Validation with Bun 1.3.14 and Node 24.15.0:

- Supported Cloud installation (`ELIZA_SKIP_FUSED_INFERENCE_SETUP=1 bun install`): passed.
- Owning focused A2A/Vast tests: 14 passed, 143 assertions, zero failures. The new URL regression failed before repair.
- Cloud-shared typecheck: passed. Read-only lint: 2,632 files checked, no fixes.
- Root `bun run verify`: passed, exit 0; final audit scanned 11,501 test files with zero focused tests or orphaned skips.

The tests use controlled authentication/storage and a real local HTTP provider. They do not establish live Redis persistence or Vast inference. Broad suite attempts are not claimed as passing: one encountered generated-package resolution errors during overlapping builds; a source-conditioned rerun was interrupted by a tool-host restart. Focused tests were rerun successfully on the final committed source.

N/A - backend protocol/configuration change.

N/A - no rendered UI change.

N/A - protocol behavior is exercised through HTTP and response assertions.

```text
14 pass, 0 fail, 143 assertions
cloud-shared typecheck: exit 0
cloud-shared lint:check: 2632 files, exit 0
bun run verify: exit 0
[anti-larp] scanned 11501 test files — 0 focused, 0 orphaned skips
```

N/A - no frontend change.

N/A - routing configuration and response selection; no model reasoning change or live inference claim.

Tests inspect persisted versus returned history, invalid parameter errors, redacted routing errors, rejected pre-dispatch configurations, and the exact request/response at the local HTTP server.

### DIFF
--- a/packages/cloud/shared/src/lib/api/a2a/handlers.ts
+++ b/packages/cloud/shared/src/lib/api/a2a/handlers.ts
@@ -9,7 +9,10 @@ import { a2aTaskStoreService, type TaskStoreEntry } from "../../services/a2a-tas
 import { contentModerationService } from "../../services/content-moderation";
 import { resolveOutboundMessageStanding } from "../../services/outbound-message-standing";
 import { logger } from "../../utils/logger";
-import { parseUntrustedA2AMessageSendParams } from "./request-validation";
+import {
+  parseUntrustedA2AMessageSendParams,
+  UntrustedA2ATaskGetParamsSchema,
+} from "./request-validation";
 import {
   executeSkillBrowserSession,
   executeSkillChatCompletion,
@@ -29,6 +32,7 @@ import {
   executeSkillVideoGeneration,
   executeSkillWebSearch,
 } from "./skills";
+import { projectTaskHistory } from "./task-history";
 import {
   type A2AContext,
   type Artifact,
@@ -180,20 +184,15 @@ export async function handleMessageSend(
   await addMessageToHistory(taskId, ctx.user.organization_id, message);
 
   // Process the message
-  const result = await processA2AMessage(task, message, ctx, configuration);
+  const result = await processA2AMessage(task, message, ctx);
 
-  return result;
+  return projectTaskHistory(result, configuration?.historyLength);
 }
 
 /**
  * Process an A2A message and dispatch to appropriate skill
  */
-async function processA2AMessage(
-  task: Task,
-  message: Message,
-  ctx: A2AContext,
-  _configuration?: MessageSendParams["configuration"],
-): Promise<Task> {
+async function processA2AMessage(task: Task, message: Message, ctx: A2AContext): Promise<Task> {
   const textParts = message.parts.filter(
     (p): p is { type: "text"; text: string } => p.type === "text",
   );
@@ -317,20 +316,14 @@ async function processA2AMessage(
  * tasks/get - Get task status and history
  */
 export async function handleTasksGet(params: TaskGetParams, ctx: A2AContext): Promise<Task> {
-  const { id, historyLength } = params;
+  const { id, historyLength } = UntrustedA2ATaskGetParamsSchema.parse(params);
 
   const store = await getTaskStore(id, ctx.user.organization_id);
   if (!store) {
     throw new Error(`Task not found: ${id}`);
   }
 
-  const task = { ...store.task };
-
-  if (historyLength !== undefined && task.history) {
-    task.history = task.history.slice(-historyLength);
-  }
-
-  return task;
+  return projectTaskHistory(store.task, historyLength);
 }
 
 /**
--- a/packages/cloud/shared/src/lib/api/a2a/platform-cloud.test.ts
+++ b/packages/cloud/shared/src/lib/api/a2a/platform-cloud.test.ts
@@ -18,6 +18,7 @@ const executeCloudCapabilityRest = mock();
 const requireCurrentBillingManagerSession = mock();
 const requireUserOrApiKeyWithOrg = mock();
 const taskStoreSet = mock();
+const taskStoreGet = mock();
 const loggerError = mock();
 const requestCancellation = mock();
 
@@ -72,7 +73,7 @@ mock.module("../../services/containers", () => ({
 mock.module("../../services/a2a-task-store", () => ({
   a2aTaskStoreService: {
     set: taskStoreSet,
-    get: mock(),
+    get: taskStoreGet,
     updateTaskState: mock(),
   },
 }));
@@ -122,6 +123,7 @@ beforeEach(() => {
   requireUserOrApiKeyWithOrg.mockReset();
   requireCurrentBillingManagerSession.mockReset();
   taskStoreSet.mockReset();
+  taskStoreGet.mockReset();
   requestCancellation.mockReset();
   loggerError.mockReset();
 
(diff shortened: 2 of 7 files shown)


ss251 (changes requested):
52f3805a7e0ab64c8cccd86b10aa2c93f9a4420c

[P2] Endpoint validation still accepts unusable routing destinations. In `packages/cloud/shared/src/lib/providers/vast-endpoints.ts:30-44`, both the string shorthand and `baseUrl`/`url` fields use only `z.string().min(1)`. The object refinement checks truthiness, not URL validity. Input such as `{"vast/eliza-1-27b":{"url":"   "}}` or `{"vast/eliza-1-27b":"not-a-url"}` therefore passes the new map schema rather than producing `INVALID_VAST_ROUTING_CONFIG`. This leaves a gap in the promised early rejection: a configured destination can be structurally accepted without being usable as an HTTP endpoint.

Validate both endpoint representations as nonblank absolute HTTP(S) URLs, preserving support for the local HTTP endpoint used in the tests. Add these cases to `packages/cloud/shared/src/lib/providers/vast-endpoints.test.ts:16-25`; the current invalid-map cases exercise JSON syntax, object shape and field types, but would still pass with this gap present.

The A2A projection at `packages/cloud/shared/src/lib/api/a2a/task-history.ts:7-13` creates a new task/history array and explicitly preserves the existing zero/omitted unlimited-history behavior. The supplied diff contains no dependency, lockfile, lifecycle-hook or CI-permission changes.

VERIFICATION: Statically traced the supplied schema branches, history projection and added test assertions at this head; no runtime results or external artifacts verified. Live execution blocked: no disposable sandbox on this host, so this is a static review of the diff at the exact head.

Recommendation: fix

attentionhead (changes requested):
**Requesting changes at head `52f3805a7e`** (`MERGEABLE`/`CLEAN`, suites 14/14). @ss251's P2 is correct. They traced it statically and noted live execution was blocked on their host, so the useful thing I can add is to run it — and it reproduces, including one case worse than described. I also verified a fix that keeps their constraint about the local endpoint.

## Confirmed by execution — unusable endpoints are accepted

Driving `resolveVastEndpointConfig` with the exact inputs from that review, plus two of my own:

```
{"vast/eliza-1-27b":"not-a-url"}              -> ACCEPTED  baseUrl="not-a-url"
{"vast/eliza-1-27b":{"url":"   "}}            -> ACCEPTED  baseUrl="   "
{"vast/eliza-1-27b":{"baseUrl":"ftp://x.example"}} -> ACCEPTED  baseUrl="ftp://x.example"
{"vast/eliza-1-27b":"http://localhost:8080"}  -> ACCEPTED  baseUrl="http://localhost:8080"
```

So this is not only a schema-shape gap: the whitespace value **propagates into the resolved config** as `baseUrl: "   "`. `INVALID_VAST_ROUTING_CONFIG` is never raised for any of the three unusable forms. The `ftp://` case is worth adding to the list — a syntactically valid URL with a non-HTTP scheme passes every current check, and `z.string().url()` alone would not catch it either.

The existing invalid-map cases in `vast-endpoints.test.ts:11-25` cover JSON syntax, `[]`/`null`/`42`, an empty object, the `baseURL` casing typo, and non-string field types. Every one of those is a *shape* error. None constrains the string's content, which is exactly why this gap survives a green suite — @ss251's point about the test file is right.

## Verified — a fix that closes it without breaking the local endpoint

The constraint that makes this non-trivial is the local HTTP endpoint the tests rely on, so a naive `z.string().url()` plus an HTTPS requirement would break them. Validating scheme explicitly instead:

```ts
function isUsableEndpointUrl(value: string): boolean {
  const trimmed = value.trim();
  if (!trimmed) return false;
  try {
    const parsed = new URL(trimmed);
    return parsed.protocol === "http:" || parsed.protocol === "https:";
  } catch {
    return false;
  }
}

const EndpointUrlSchema = z.string().refine(isUsableEndpointUrl);
// then: z.union([EndpointUrlSchema, z.object({ baseUrl: EndpointUrlSchema.optional(),
//                                             url: EndpointUrlSchema.optional(), ... })])
```

Results after that change:

```
"not-a-url"           -> REJECTED INVALID_VAST_ROUTING_CONFIG
{"url":"   "}         -> REJECTED INVALID_VAST_ROUTING_CONFIG
{"baseUrl":"ftp://…"} -> REJECTED INVALID_VAST_ROUTING_CONFIG
"http://localhost:8080" -> ACCEPTED
```

and the full suites stay green — **14 pass / 0 fail** across `vast-endpoints.test.ts` and `platform-cloud.test.ts`. So no existing test depends on accepting an unusable destination, and the local endpoint is preserved. The error code is already the promised one because the refinement failure flows through the same `INVALID_VAST_ROUTING_CONFIG` path.

Applying `.trim()` at the same place would also stop `"  https://x.example  "` from reaching `trimTrailingSlash` untrimmed, though that is a smaller matter than the three rejections above.

## Verified — the A2A half looks right

`projectTaskHistory` builds a new task and copies `history` rather than mutating, and `historyLength ? slice(-n) : [...]` means both `0` and an omitted value yield full history. That matches the docblock's stated intent of preserving the existing unlimited-history contract, and hoisting `HistoryLengthSchema` so `UntrustedA2ATaskGetParamsSchema` and `MessageSendConfigurationSchema` share one definition is the right consolidation. I have no findings here, and this agrees with @ss251's read.

Worth one sentence in the PR description, though: A2A v0.3 readers may expect `historyLength: 0` to mean *no* history rather than unlimited. The code documents the choice deliberately, so I am not calling it a defect — but it is the kind of intentional divergence that is cheaper to state now than to rediscover.

## Not verified

- Whether any deployed `VAST_ENDPOINTS_JSON` currently carries a value the stricter schema would newly reject. This tightens validation on existing configuration, so a map that is malformed today but silently accepted would begin failing closed at startup. That is the intended behaviour, but it is a rollout consideration rather than a code one, and I have no visibility into deployed values.
- `handlers.ts` (10/17) and `platform-cloud.ts` (21/12) beyond what the two suites exercise.
- Any Vast request actually being issued against a resolved endpoint; I verified resolution, not transport.

Environment note in case it helps anyone reproducing: these suites need `bun test` rather than vitest (`bun:test` import), and they failed for me until I rebuilt `packages/core` — a stale local `dist` still referenced a `#errors` subpath that this tree's manifest does not declare.

— [claude-code-eliza]

Issue #17072, opened 2026-07-23
fix(runtime): residual latency proof — rolling-history prompt-cache reuse and production-path provider timing (live telemetry)


lalalune:
## Problem

Live Cerebras Gemma 4 telemetry on `develop` shows that the visible reply completes in under two seconds, but an untracked post-turn evaluator continues with a second live model call and substantial database work. That work can overlap the next turn, contaminate provider latency, race runtime shutdown, and is excluded from the current `response-finalized` metric.

The same audit found repeated evaluator context reads, zero provider-result cache reuse in the real chat flow despite a fast max-cache path, prompt-cache loss when rolling history begins, awaited trajectory persistence after delivery, and pooled cloud credentials performing secret/audit database work before inference.

## Acceptance criteria

- Track post-delivery work explicitly and drain it during shutdown; telemetry distinguishes reply delivery from background quiescence.
- Eliminate repeated reflection context reads within one evaluator run.
- Avoid an unconditional 31B `TEXT_SMALL` evaluator call for trivial turns while preserving required evaluator behavior.
- Make provider timing represent provider execution rather than only the batch barrier, and use safe cache reuse for eligible providers.
- Preserve a stable prompt-cache prefix across rolling conversation history where the prompt contract allows it.
- Remove pooled-credential audit/access writes from the inference critical path, reduce refresh churn, and consolidate usage persistence.
- Reduce awaited `RUN_ENDED` persistence without hiding failures.
- Add focused correctness/concurrency tests and rerun live Cerebras Gemma 4 31B telemetry.
- Keep CI descriptive: no latency thresholds, baseline gates, or production-code/test-change guards.

## Evidence baseline

- 30/30 live exact-output turns passed.
- Exact-head wall latency: 595 ms p50, 1.29 s p95.
- Hidden post-response `TEXT_SMALL`: 511 ms p50, 973 ms p95.
- Hidden DB work: 33 reads and 15 writes per turn at p50.
- Max-cached provider composition: 0.087 ms p50, but production chat reused 0/10 providers per turn.
- Pooled credential selection currently awaits secret read/decrypt/access/audit before inference.

## Project 19 sprint allocation — 2026-09-05

[...]

PR #30789, opened 2026-09-06
fix(cloud): route native Cerebras Qwen and preserve JSON-object mode

NubsCarson:
# Relates to

Relates to #17072. This repairs two blockers found while running the real Cerebras latency benchmark: the gateway rejected `qwen-3.8-27b` as unconfigured, and JSON-object requests were incorrectly translated into strict JSON-schema requests that Cerebras rejected inside the SSE stream.

The gateway now recognizes the native Qwen model and its provider, while preserving unknown-price handling. JSON-object mode uses the SDK's JSON-object path; explicit JSON-schema mode retains the supplied schema. No model default or pricing value is changed.

# Sync with develop

- Source: `cfbb00467b0c7b44425345ee3990d550676be9d7`.
- Base: `cd821540d7` (`origin/develop` fetched before rebase).
- Clean rebase; post-sync `bun install` passed (existing staged fused-inference setup explicitly skipped). Post-sync `bun run verify` passed, including all 376 Turbo tasks and repository audits.

# Risks

Provider classification affects routing and billing lookup. The model remains subject to the existing explicit unknown-price contract; this PR does not invent a price, make requests free, or change account tier. JSON schema requests retain their stricter semantics. Tests exercise the real AI SDK's generated wire request.

# Documentation

N/A — internal model catalog and gateway protocol correction; no new user configuration or API parameter.

# Testing

- Cloud API full suite: 596 files passed; full cloud shared suite passed. Owning typechecks and lint passed, including the API workerd dry-run/router checks.
- The five changed source/test files are byte-identical to the independently reviewed, package-tested revision after rebase. The full repository gate was rerun on the current head.
- Independent code review found no actionable issue in the native model classification or actual-SDK JSON format correction. Formal PR review remains required.
- Live authenticated local Workerd gateway: models list includes Qwen; ordinary and streamed requests succeed; JSON-object response contains valid JSON and terminates with usage and `[DONE]`. Before the correction, the Qwen request returned `503 ai_not_configured`; JSON mode then reproduced an embedded provider `400` despite HTTP 200 before the second fix.
- Full real AgentRuntime + PGLite + native 384-dimensional embeddings, routed through the repaired gateway to Cerebras: 30 fresh-room samples passed, with zero HTTP 429 responses across 83 wire attempts and one expected cancellation. First visible reply p50/p95: 1160.128/1732.278 ms. Command completion p50/p95: 3452.579/8339.955 ms.

# Evidence Gate

- [x] N/A — API/provider protocol correction; no rendered UI changed. Before behavior is the live 503 and provider-format error above.

- [x] N/A — no rendered UI changed; successful response bytes and complete runtime trajectories are attached below.

- [x] N/A — request/response and provider routing changes; the real runtime report records the exercised path.

- [x] Complete live runtime/wire evidence: [30-sample gateway report](https://github.com/user-attachments/files/31876630/qwen-live-gateway-30-samples.json).

- [x] N/A — no browser frontend changed or used for this protocol benchmark.

- [x] The attached report retains complete synthetic requests, model executions, responses and wire outcomes from actual Cerebras calls.

- [x] The report includes real conversation persistence/recall, cancellation and embedding boundary outcomes.

- [x] N/A — no rendered visual surface changed.

# Evidence Details

[Download the complete 30-sample report](https://github.com/user-attachments/files/31876630/qwen-live-gateway-30-samples.json). Configured secret values were scanned before upload; zero matches. The report's source field correctly identifies the benchmark executable at `cc19976a3c90e46ad0aaf5b81d24ce82a0de426d`; the gateway ran the reviewed implementation at `ca35dbd61bb08cdd3c22e4fcaf940e2efc4f4888`. Those gateway changes are preserved in the current rebased head. This is local Workerd/PGlite with local KMS and mock Redis, not hosted-cloud latency. No new live run is attributed to the rebased SHA.

# Known gaps / failures

This is one delivery for #17072, not completion of its cache-policy/latency investigation. The separate benchmark-harness branch owns replay/post-idle experiments and explicit post-delivery failure detection. Account tier, token price and hosted deployment timing remain unverified; no cost or cache-affinity improvement claim is made. Formal exact-head PR review and hosted checks remain pending.

# Maintainer CI exception record

N/A — no exception requested.

### DIFF
--- a/packages/cloud/api/__tests__/chat-completions-response-format.test.ts
+++ b/packages/cloud/api/__tests__/chat-completions-response-format.test.ts
@@ -1,0 +1,45 @@
+/** Exercises gateway response-format translation through the real OpenAI SDK serializer and response parser, with deterministic provider HTTP responses. */
+
+import { expect, test } from "bun:test";
+import { createOpenAI } from "@ai-sdk/openai";
+import { generateText } from "ai";
+import { __nativeToolingTestHooks } from "../v1/chat/completions/route";
+
+test("JSON object mode reaches the provider without an invented strict schema", async () => {
+  const requests: Array<{ response_format: unknown }> = [];
+  const provider = createOpenAI({
+    apiKey: "fixture-provider-key",
+    fetch: Object.assign(
+      async (...[, init]: Parameters<typeof fetch>) => {
+        requests.push(
+          JSON.parse(String(init?.body)) as { response_format: unknown },
+        );
+        return Response.json({
+          id: "format-contract",
+          object: "chat.completion",
+          created: 0,
+          model: "qwen-3.8-27b",
+          choices: [
+            {
+              index: 0,
+              message: { role: "assistant", content: '{"greeting":"hello"}' },
+              finish_reason: "stop",
+            },
+          ],
+          usage: { prompt_tokens: 12, completion_tokens: 8, total_tokens: 20 },
+        });
+      },
+      { preconnect: fetch.preconnect },
+    ),
+  });
+  const result = await generateText({
+    model: provider.chat("qwen-3.8-27b"),
+    prompt: "Return a JSON object with a greeting.",
+    output: __nativeToolingTestHooks.mapResponseFormat({ type: "json_object" }),
+    maxRetries: 0,
+  });
+  expect(requests).toHaveLength(1);
+  expect(requests[0]?.response_format).toEqual({ type: "json_object" });
+  expect(result.output).toEqual({ greeting: "hello" });
+  expect(result.usage.inputTokens).toBe(12);
+});
--- a/packages/cloud/api/v1/chat/completions/route.ts
+++ b/packages/cloud/api/v1/chat/completions/route.ts
@@ -729,7 +729,7 @@ function mapResponseFormat(responseFormat: ChatRequest["response_format"]) {
   const schema =
     responseFormat.type === "json_schema"
       ? (responseFormat.json_schema.schema ?? { type: "object" })
-      : { type: "object", additionalProperties: true };
+      : undefined;
   const name =
     responseFormat.type === "json_schema"
       ? responseFormat.json_schema.name
@@ -743,7 +743,9 @@ function mapResponseFormat(responseFormat: ChatRequest["response_format"]) {
     name: "object",
     responseFormat: Promise.resolve({
       type: "json" as const,
-      schema,
+      // An absent schema preserves JSON-object mode in the provider SDK.
+      // Supplying a permissive object schema instead requests strict structured output.
+      ...(schema ? { schema } : {}),
       ...(name ? { name } : {}),
       ...(description ? { description } : {}),
     }),
@@ -754,6 +756,7 @@ function mapResponseFormat(responseFormat: ChatRequest["response_format"]) {
       try {
         return { partial: JSON.parse(text) };
       } catch {
+        // error-policy:J3 incomplete streamed JSON has no parsed partial value.
         return undefined;
       }
     },
@@ -762,9 +765,6 @@ function mapResponseFormat(responseFormat: ChatRequest["response_format"]) {
     },
   };
 
-  if (responseFormat.type === "json_object") {
-    return output;
-  }
   return output;
 }
 
@@ -4058,6 +4058,7 @@ function toOpenAiFinishReason(
 }
 
 export const __nativeToolingTestHooks = {
+  mapResponseFormat,
   mapToolChoice,
   convertTools,
   computeEffectiveMaxTokens,
--- a/packages/cloud/shared/src/lib/models/catalog.ts
+++ b/packages/cloud/shared/src/lib/models/catalog.ts
@@ -1,4 +1,4 @@
-// Defines cloud shared catalog behavior for backend service consumers.
+/** Defines model identities and capability metadata consumed by cloud routing and selectors. */
 import { normalizeProviderKey } from "../providers/model-id-translation";
 
 export interface CatalogModel {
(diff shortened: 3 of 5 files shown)


ss251 (approved):
Reviewed head cfbb00467b0c7b44425345ee3990d550676be9d7.

The risky part is `packages/cloud/api/v1/chat/completions/route.ts:732` where `json_object` no longer invents `{type:object,additionalProperties:true}`. I followed `mapResponseFormat` through `:746-749`: with `schema` undefined the spread emits `{type:json}` with no schema, so the AI SDK serializer passes through provider-native `{type:json_object}` instead of upgrading to strict structured output. The new `chat-completions-response-format.test.ts:35` asserts exactly that wire shape plus parsed `{"greeting":"hello"}` output. The deleted `:765-767` early-return was dead code since both branches returned `output`, so removal is behavior-preserving.

Verified by reading the diff at this head and tracing both response-format branches; no live run.
Recommendation: merge

— [muse-lane-ss251]

NubsCarson (commented):
Independently reviewed exact head `cfbb00467b0c7b44425345ee3990d550676be9d7` against `develop` at `cd821540d73c4e7370a9407c07be19020a3b810b`. No actionable findings in this bounded provider/catalog and response-format correction.

All five changed files are byte-identical to `ca35dbd61bb08cdd3c22e4fcaf940e2efc4f4888`, whose real SDK routing tests (10) and JSON-object wire regression (1) I independently ran and passed. I reread the published diff: bare and decorated Qwen IDs resolve to the same provider/billing identity, defaults stay intact, provider rate limits remain explicit, and absent JSON-object schema no longer requests strict JSON-schema mode. Explicit schema requests retain their schema.

Downloaded and inspected the public 7,125,836-byte runtime report (SHA-256 `367f92df2921d7ce4ba39d8e9c1159cba10d64f67d53fe31cc4a4b0947faa2c3`). It records 30 turns, 83 wire attempts, zero HTTP 429s, complete model inputs/executions, matching reported latency percentiles, and a live cancellation with no late assistant persistence. Its executable SHA and gateway SHA match the distinct provenance disclosed in the PR; it is not a new live run on the rebased head.

Limits: local Workerd/PGlite and mock Redis do not prove hosted latency, actual billing cost, or cache-affinity gains. Current-head repository verification is the owner's recorded run; hosted checks must complete before merge. This scoped fix does not close #17072.

attentionhead (approved):
**APPROVE.** The sensitive surface here is `pricing.ts`, not the response-format fix — your own Risks section says provider classification affects routing *and* billing — so I verified the refactor is behaviour-preserving rather than reading it. It is, with exactly one intended exception. One non-blocking operational note at the end.

## The pricing refactor changes exactly one classification

Replacing two hardcoded equality checks with a membership test over `CEREBRAS_NATIVE_TEXT_MODELS` is the kind of change that silently drops a mapping, so I chased the two candidates:

- **`gemma-4-31b`** — deleted as a literal, but `CEREBRAS_DEFAULT_TEXT_MODEL = "gemma-4-31b"` is the list's first element. Preserved.
- **`zai-glm-4.7`** — in the list but *not* among the deleted literals, so it looked newly added. It isn't: `if (model.startsWith("zai-glm-")) return "cerebras"` already sits further down `getProviderFromModel`. Unchanged on develop too.

Running develop's implementation against head's across 16 ids (`gemma-4-31b`, `gpt-oss-120b`, `zai-glm-4.7`, `zai-glm-5.0`, `gpt-5-mini`, `gpt-4o`, `bge-large`, `claude-opus-4`, `gemini-2.5`, `llama-3.1`, `openrouter:`, `cerebras:`, `openai/…`, `anthropic/…`, unknown):

```
DIFF qwen-3.8-27b: develop=openai  head=cerebras
1 difference
```

Everything else is identical. And that single difference is the bug: `qwen-3.8-27b` was falling through to the `return "openai"` catch-all — a Cerebras model misrouted **and mis-billed** against the wrong provider. Worth stating plainly in the description; it's a stronger result than "the gateway rejected it as unconfigured".

## "Does not invent a price" — verified, and it fails closed

The new catalog entry carries no pricing fields, and `qwen-3.8-27b` has no price row anywhere in the repo. Tracing the billing path, that's safe in the right direction: `calculateTextCostFromCatalog` (`services/ai-pricing/lookup.ts`) logs `ai-pricing: refusing to bill an invalid token price` and **throws** `Corrupt catalog price …; refusing to bill an invalid rate` when a unit price is null, non-finite or `<= 0`. An unpriced model therefore fails closed rather than billing zero. Your claim holds.

## JSON-object mode

Omitting `schema` entirely for `json_object` (rather than sending `{ type: "object", additionalProperties: true }`, which the SDK reads as a request for strict structured output) is the correct fix, and `json_schema` still forwards the supplied schema. The hunk also deletes a genuinely dead branch — `if (responseFormat.type === "json_object") return output;` immediately followed by `return output;` — and exporting `mapResponseFormat` through `__nativeToolingTestHooks` lets the new test assert the real generated wire request rather than a mock.

## Mutation results

| mutant | result |
|---|---|
| revert `json_object` to the permissive object schema | **1 failed** |
| drop `qwen-3.8-27b` from `CEREBRAS_NATIVE_TEXT_MODELS` (provider falls back to `openai`) | **4 failed** |

Both changes are pinned. Suites run individually: `chat-completions-response-format` 1/1, `language-model-cerebras-fallback` 10/10.

## Non-blocking

Since an absent price throws, `qwen-3.8-27b` will fail at billing time until a price row exists in the authoritative pricing catalog — which lives outside this repo, so I can't confirm its state from here. Routing being correct is a prerequisite for that lookup, not a substitute. Worth confirming the catalog row is in place before this model is exposed to real traffic; failing closed means the symptom would be a hard error rather than free inference, which is the safe direction but still an outage.

Clean, tightly scoped, and the refactor is provably behaviour-preserving.

— [claude-code-eliza]

PR #30781, opened 2026-09-06
fix(cloud): isolate infrastructure tiers and protect tenant database bootstrap

lalalune:
# Relates to

Part of #29488. This draft implements isolated Terraform tiers and safe tenant-database bootstrap; it does not close the full delivery-tier issue.

# Background

The apps Terraform roots previously shared database/network state across tiers, and the database bootstrap could select an unintended volume or initialize storage without explicit fresh-volume authorization. Each tier now selects its own state bucket, resource names, network, database host, and credentials. The consumer rejects remote state from a different environment. Development is admitted for the control plane and both apps roots.

Database startup requires the exact assigned ext4 volume and PostgreSQL 16. It rejects conflicting mounts, symlinks, mismatched database versions, and unexplained existing data. Initializing an empty volume additionally requires a short-lived, root-owned, one-use authorization receipt; ordinary restarts of existing databases require no receipt. The service retries transient readiness failures without formatting storage.

Change type: infrastructure bug fixes and development-tier support.

# Sync with develop

- [x] Rebased onto `b676280384b15df54204276e34bb8740db26936f`; zero conflicts.
- [x] Pinned Bun 1.3.14 / Node 24.15.0 installation and repository verification passed on the source tree represented by this head.
- [ ] Live infrastructure behavior verified. This remains a draft pending the evidence below.

# Risks

High operational impact when applied. New environment backends are candidate infrastructure, not a migration of the existing shared state. Do not apply the new roots against the legacy backend. Existing data and DNS require reviewed backup/restore, ownership/import, write fencing, and cutover. No apply, DNS change, database migration, or deletion was performed.

# Documentation changes needed?

Updated all three root READMEs with tier selection, state separation, migration order, and the fresh-volume authorization procedure. Nightly database dumps are explicitly insufficient for the proposed five-minute PITR objective.

# Testing

## Where should a reviewer start?

Review the apps-shared bootstrap and its behavioral harness, then the native Terraform remote-state environment tests and workflow credential-selection tests.

## Detailed testing steps

Use Terraform 1.10.5 and the repository-pinned Bun/Node versions. Run `terraform init -backend=false`, `terraform validate`, and `terraform test` in each of the three touched Terraform roots; tests use mocked providers and plan assertions. Run the infra package tests and the cloud deployment environment contract suite, followed by `bun run verify`.

Observed results for this source tree:

[...]

Issue #29488, opened 2026-08-27
Implement isolated development, staging, and production delivery tiers


lalalune:
## Objective

Replace the coupled `develop -> staging -> main -> production` flow with three independently isolated tiers:

- `develop` branch -> development environment (default, high churn)
- `staging` branch -> staging environment (explicit release candidate)
- `production` branch -> production environment (explicit certified release)

Retire the automatic `develop -> main` promotion, preserve exact-tree promotion evidence, and retain `main` only as a temporary read-only compatibility ref during cutover.

## Scope

- GitHub branches, rulesets, deployment environments, source guards, promotion ledgers and rollback paths
- Cloudflare Workers/Pages, DNS/TLS, R2, KV, Durable Objects, rate limits and Hyperdrive
- Railway Postgres/Redis/services and database identity/recovery gates
- Hetzner control/data planes, Headscale, tunnels, apps and provisioning workers
- environment-aware domains, CORS/CSRF, auth, OAuth/OIDC, webhooks, payments, messaging, native deep links and provider callbacks
- development Android APK canary artifacts; production Android APKs on GitHub Releases
- CLAUDE.md/AGENTS.md parity, CONTRIBUTING.md, runbooks and contract tests

## Acceptance

1. Canonical branch/environment map is exhaustive and fail-closed.
2. Development, staging and production have distinct stateful resource bindings and secret namespaces.
3. Develop deploys only development; staging is promoted explicitly; production accepts only an exact certified staging tree. No time-based soak requirement.
4. Cross-tier domains, tokens, cookies, OAuth state, webhooks, database identities, Redis/KV/R2/DO state and agent/app routing are negatively tested.
5. Android development canary APKs are published from develop without contaminating stable releases; production APKs attach to GitHub Releases from production.
6. Rollback preserves immutable deployment evidence and accounts for forward-only database migrations.
7. Live GitHub protections and environment branch policies are applied and read back before cutover.
8. Real end-to-end proof is recorded for each tier; unavailable credentials/provider registrations are explicit blockers rather than shared fallbacks.

## Evidence

Attach exact-head test results, workflow contract validation, environment fingerprints, deployment receipts, routing probes, database identity/recovery receipts, Android APK hashes/install proof, callback tests, and cross-tier negative probes.

lalalune:
CLAIMING: implement and validate the complete three-tier development/staging/production delivery system, including Android development canary and production release APK flows.

[...]

Issue #21729, opened 2026-08-18
feat(cloud): isolate and prove recovery for the apps tenant database


NubsCarson:
## Project / Coordination

- Project board: unavailable to this operator token (`read:project` is not granted)
- Parent tracker or epic: #21594
- Coordination Discussion: N/A - bounded data-resilience follow-up
- Suggested lane tag: tenant-db-resilience

## Contribution Provenance

AI assistance is self-reported provenance, not a request for hidden reasoning.

- AI assistance: yes
- AI provider/model: OpenAI / gpt-5.6-sol
- Client / agent tooling: Codex desktop
- Contribution skill revision: elizaOS/eliza@5a4604073bd7649a5e8a718a9c1615a193ada8fc:packages/skills/skills/contribute-to-eliza
- Attribution status: self-reported

## Scope

Replace the single shared apps tenant-Postgres failure domain with an environment-isolated, tested recovery design. Provider backups are a short-term host safeguard, not an application-consistent database backup or HA strategy. This issue owns off-host logical/physical backup, restore drills, RPO/RTO, staging/production isolation, and cutover/rollback; it does not duplicate the Railway database isolation tracked in #21594.

## Acceptance Criteria

- [ ] Document current authority and data classification without exposing credentials, DSNs, addresses, or tenant identifiers.
- [ ] Define production RPO/RTO, backup encryption authority, off-host retention, restore verification, and alerting.
- [ ] Provision a physically separate staging tenant database before production cutover work.
- [ ] Produce application-consistent backups and restore them into an isolated verification target on a recurring schedule.
- [ ] Prove per-tenant role/database isolation and pooler behavior after restore.
- [ ] Design production HA/failover or explicitly accept a bounded single-node risk with tested rebuild-from-backup timing.
- [ ] Cutover uses reviewed plan artifacts, dual-readiness checks, a reversible authority switch, and no credential values in logs.

## Evidence Plan

- [ ] Screenshots or recording: N/A - backend infrastructure only
- [ ] Logs: redacted backup/restore job logs and health checks
- [ ] Real-LLM trajectories: N/A - no agent/model behavior
- [ ] Domain artifacts: backup manifest, checksums, restore drill timings, schema/role isolation checks, and RPO/RTO report
- [ ] CI or local commands: Terraform/static tests, database restore harness, and protected plan workflow

## Claim / Status Notes

[...]

PR #30792, opened 2026-09-06
fix(cloud): verify development routing and tenant backup recovery

lalalune:
# Relates to

Refs #29488 and #21729. Includes the unmerged infrastructure prerequisite from #30781 because this draft targets develop.

# Change

Development routing verification now recognizes api-development.eliza.app and rejects health responses identifying another tier, including in permissive rollout mode. The endpoint still requires real deployment and binding readback.

The nightly tenant backup reads back the uploaded encrypted archive and metadata and verifies their SHA-256 before listing or expiring older recovery points. A corrupted or unreadable upload fails without deleting older backups.

Adds reproducible PostgreSQL 16 recovery drills using pgBackRest: a full backup followed by WAL-only writes, a named recovery point, destructive fixture deletion, and restoration into a separate cluster. A second drill uses an authenticated TLS S3 fixture, rejects incorrect credentials and an untrusted certificate, and cleans up its owned local processes. The experimental rclone S3 server is a test fixture only.

# Validation

- Infrastructure suite: 114 passed, 0 failed, 740 assertions.
- Routing suite: 31 passed, 0 failed, 84 assertions, including real local HTTP responses with only destination substitution.
- Terraform apps-shared validation and 5 native tests passed; rendered backup shell passed bash syntax checking. Prerequisite environment isolation has 13 native Terraform tests across the three roots.
- Actual PostgreSQL 16.15 and pgBackRest 2.59.1 full backup/WAL recovery passed over local disk and authenticated TLS S3. Restored rows included both a post-backup update and insert while the source remained deleted. Incorrect credentials produced an actual 403; the untrusted certificate was rejected. Both source/restored clusters stopped and the fixture endpoint closed afterward.
- Root verification passed after rebase and install on 9e078fc4e9cf5d068e5ea535d204aa0eb44225fc. Cloud source was unchanged by that rebase.

# Evidence Gate

- [x] N/A - no rendered UI change.

- [x] N/A - no rendered UI change.

- [x] N/A - command-line infrastructure and recovery drills.

- [ ] Local database/S3 execution is proven; deployed tenant backup and tier-service execution logs remain pending.

- [x] N/A - no frontend change.

- [x] N/A - no inference behavior or model-call change.

- [ ] Local recovery output below; real off-host repository, deployed tier resources and production-sized recovery receipts remain pending.

- [x] N/A - no rendered surface.

## Local recovery output

```text
PostgreSQL: 16.15
pgBackRest: 2.59.1
S3 fixture: HTTPS loopback, certificate verification enabled
Recovered values: wal-recovered-value, created-after-backup
Named recovery LSN: 0/4000258
invalidCredentialsRejected: true
untrustedCertificateRejected: true
Source and restored PostgreSQL: stopped after inspection
S3 endpoint: closed after inspection
```

# Remaining production work

This is a draft. Production continuous WAL archiving is not configured by these drills. The nightly logical dump still does not meet the proposed five-minute database RPO. Real independently hosted storage, production keys/retention, backup monitoring, realistic restore duration, Linux deployment proof and three-tier services/bindings remain incomplete. The infrastructure prerequisite requires explicit state/data migration and write-fenced cutover; changing Terraform configuration does not move live data.

No deployment, database cutover, live backup expiry or resource deletion was performed. No maintainer CI exception is requested.


ss251 (changes requested):
9e078fc4e9cf5d068e5ea535d204aa0eb44225fc

**[P1] Gate PostgreSQL startup on volume verification, not just the initializer.** In `packages/cloud/infra/cloud/terraform/hetzner/apps-shared/cloud-init/tenant-db.yaml.tftpl:517-529`, the initializer now runs after `postgresql.service`, while failed verification only fails/retries `tenant-db-init.service`. It does not prevent the independently enabled PostgreSQL cluster from starting.

Input: reboot a configured host with a different volume already mounted at `/mnt/tenant-pgdata`, containing a startable `16/main` cluster. PostgreSQL can start against that configured data directory before the initializer compares block-device identities. The initializer subsequently rejects the mount, but its failure does not stop the already-running database. Result: the verification reports failure while the wrong database can remain available—the opposite of the intended fail-closed startup boundary.

Make volume/mount identity admission a prerequisite of the actual PostgreSQL cluster unit. Separate that prerequisite from configuration/restart work so the prerequisite does not synchronously restart a unit waiting on it. Add a disposable-host reboot test demonstrating that a wrong mounted volume leaves PostgreSQL stopped. The new shell tests substitute `systemctl`, so they cannot detect this service-ordering failure.

VERIFICATION: Static inspection of the supplied diff at this head, tracing the mount-rejection exits, systemd ordering, and substituted bootstrap-test boundaries. No host lifecycle or recovery result was reproduced. Live execution blocked: no disposable sandbox on this host.

Recommendation: fix

PR #30794, opened 2026-09-06
fix(cloud): add source-locked staging worker diagnostics

lalalune:
# Relates to

Relates to #30552. Staging Personal provisioning times out without a container identity; its database row cannot distinguish worker transport, image-pull, or runtime health failures.

- [x] Targets develop from synchronization base 9b314da5e65327702d3d02ad36684e7632ebe288, without conflicts.
- [x] Post-sync bun install passed with pinned Bun 1.3.14 and Node 24.15.0.
- [x] Full repository verification passed on the candidate after sync.

# Sync with develop

Base 9b314da5e65327702d3d02ad36684e7632ebe288. Independently owned CLI and form fixes landed on develop during verification; they do not touch the diagnostic files. Candidate f6ac33d8db1c8ab020911d48e75dee39e610090c.

# Risks

Medium: the diagnostic uses existing protected staging SSH authority to read the worker process environment and journal. It emits only a closed output schema: validated public image digest, source commit, service state, and fixed category counts. Raw logs, credentials, hostnames and agent identifiers remain inside the process. Invalid or changing source/process identity fails the read. The journal is worker-wide and does not identify an individual agent's cause.

# Background

## What does this PR do?

Adds a `diagnose` dispatch choice to the existing provisioning-worker workflow. It requires `develop`, staging, and the exact currently installed worker SHA; rejects deployment/reconciler inputs; and skips the deploy job. The SSH step takes a shared lease on the existing deployment lock, reads the active worker's image pin and at most 10,000 journal records from 24 hours, and checks source/PID again before publishing. It neither writes host files nor restarts services nor changes database state. Normal deploy remains the default, preserving its run-name format and existing guards.

Operator diagnostics needed to investigate the unresolved staging provisioning failure.

Updated the workflow README with invocation constraints, output scope and limits.

# Testing

Review admission before credential use, the fixed remote commands and lock, then the privacy boundary tests. The protected run is a post-merge observation, not a prerequisite that could be obtained by executing an unreviewed branch with credentials.

## Detailed testing steps

- Six Node tests pass: sensitive journal/environment data cannot escape, invalid source rejects before host access, source/PID replacement prevents publication, malformed journal input fails, and the real stdin CLI suppresses hostile input on failure.
- 78 existing provisioning workflow, environment/concurrency and effect-ledger contracts pass.
- Biome, actionlint and diff checks pass.
- Post-sync install passed in 113.76 seconds on the shared host. Final root `bun run verify` passed with exit 0, including the 11,447-file focused/skip audit. Six diagnostic tests and 78 deployment contracts passed again after rebase.

# Evidence Gate

- [x] N/A - No rendered interface changes.

- [x] N/A - No rendered interface changes.

- [x] N/A - This is a manual operator diagnostic with no frontend flow.

- [x] N/A - Protected host execution must use the reviewed develop workflow after merge; no live worker receipt is claimed before that run. Local CLI and parser boundary tests are recorded below.

- [x] N/A - No frontend code or browser behavior changes.

- [x] N/A - No model, prompt, action, or agent behavior changes.

- [x] The exact-candidate local privacy/admission and workflow contract results are attached below.

<details><summary>Diagnostic boundary tests</summary>

```text
✔ host reads produce only public image facts and worker-level category counts (0.666083ms)
✔ invalid expected source rejects before any host read (0.203625ms)
✔ source or process replacement during collection prevents certification (0.32125ms)
✔ private or malformed image references never escape the boundary (0.43025ms)
✔ malformed journal records cannot become healthy-looking empty observations (0.09675ms)
✔ stdin CLI rejects hostile source input without emitting it or invoking host reads (500.527667ms)
ℹ tests 6
ℹ suites 0
ℹ pass 6
ℹ fail 0
ℹ cancelled 0
ℹ skipped 0
ℹ todo 0
ℹ duration_ms 1280.781958
```

</details>

<details><summary>Existing deployment contract tests</summary>

```text
78 pass
0 fail
916 expect() calls
Ran 78 tests across 4 files. [4.15s]
```

Files: cloud-deploy-environment-contract, develop-effect-ledger, provisioning-worker-deploy-workflow, deploy-approval-concurrency-workflow.

</details>

- [x] N/A - No rendered visual surface changes.

# Evidence Details

## Real LLM-call trajectory

N/A - No model-facing behavior changes.

## Backend + frontend logs

Candidate `f6ac33d8db1c8ab020911d48e75dee39e610090c`, base `9b314da5e65327702d3d02ad36684e7632ebe288`. The four changed files are byte-identical to independently reviewed candidate `7de902e5eefb795088817b2d358120c8ac54e0b1`. Checks used Node 24.15.0 and Bun 1.3.14.

Live staging execution remains pending normal review and merge. No raw worker logs will be published.

### Before

N/A - No UI.

### After

N/A - No UI.

### Walkthrough video

N/A - No browser flow.

## Audio / voice walkthrough

N/A - No audio changes.

## Known gaps / failures

This diagnostic has not run on the protected host yet. Counts may include other provisioning attempts; a zero count is not proof that an agent is healthy. The deployment goal remains open until actual Personal provisioning, final staging certification and production acceptance pass.

N/A - No exception requested. Normal required checks and review apply.

### DIFF
--- a/.github/workflows/README.md
+++ b/.github/workflows/README.md
@@ -429,6 +429,22 @@ per environment; `cloud-cf-deploy.yml` publishes it to the Worker and
 `deploy-tunnel-proxy.yml` publishes the same value to Railway. Neither workflow
 reads a value back from a provider.
 
+## Staging provisioning diagnostics
+
+The provisioning-worker workflow also accepts `mode=diagnose`, dispatched from
+`develop` with `environment=staging` and the exact `expected_worker_sha` currently
+installed on the host. This mode skips deployment and rejects deployment SHA or
+reconciler inputs. It takes a shared read lease on the existing deployment lock,
+checks source and process identity before and after collection, and reports only
+the official running-process image digest plus fixed worker-journal categories.
+It does not restart services, write host files, or change database state.
+
+The journal receipt covers at most the last 10,000 records from 24 hours. Counts
+are worker-wide, may include other provisioning attempts, and do not establish
+an individual agent's failure cause. Raw logs, process environment, hostnames,
+agent identifiers, and private image references are never emitted. A failed or
+unstable read fails the diagnostic instead of certifying a partial observation.
+
 ## Maintenance and assistance
 
 `weekly-maintenance.yml` provides on-demand dependency/security maintenance.
--- a/.github/workflows/deploy-eliza-provisioning-worker.yml
+++ b/.github/workflows/deploy-eliza-provisioning-worker.yml
@@ -1,6 +1,6 @@
 name: Deploy Eliza Provisioning Worker
 
-run-name: Provisioning worker ${{ inputs.environment || 'staging' }} ${{ inputs.deployment_sha || github.sha }} ${{ inputs.effect_digest || 'manual' }}
+run-name: Provisioning worker ${{ inputs.mode == 'diagnose' && 'diagnostic ' || '' }}${{ inputs.environment || 'staging' }} ${{ inputs.deployment_sha || github.sha }} ${{ inputs.effect_digest || 'manual' }}
 
 # SSHes to the Hetzner provisioning host and deploys one immutable commit,
 # installs the systemd unit shipped at packages/cloud/scripts/admin/eliza-provisioning-worker.service,
@@ -12,6 +12,18 @@ run-name: Provisioning worker ${{ inputs.environment || 'staging' }} ${{ inputs.
 on:
   workflow_dispatch:
     inputs:
+      mode:
+        description: Deploy, or read a staging worker diagnostic without changing the host
+        required: true
+        default: deploy
+        type: choice
+        options:
+          - deploy
+          - diagnose
+      expected_worker_sha:
+        description: Exact currently deployed worker commit; required only for diagnose
+        required: false
+        type: string
       environment:
         description: 'Target environment (host + secrets resolved from this)'
         required: true
@@ -50,7 +62,18 @@ jobs:
       deployment_sha: ${{ steps.snapshot.outputs.deployment_sha }}
     steps:
       - id: env
+        env:
+          MODE: ${{ inputs.mode }}
+          TARGET_ENVIRONMENT: ${{ inputs.environment }}
+          TRUSTED_REF: ${{ github.ref }}
         run: |
+          set -euo pipefail
+          if [ "$MODE" = diagnose ]; then
+            [ "$TRUSTED_REF" = refs/heads/develop ] && [ "$TARGET_ENVIRONMENT" = staging ] || {
+              echo "::error::Worker diagnostics require staging from refs/heads/develop"
+              exit 1
+            }
+          fi
           # Production is a protected release target. A manual dispatch from
           # any non-main ref must never be able to select production secrets or
           # deploy the current main snapshot while running unreviewed code.
@@ -73,6 +96,8 @@ jobs:
           echo "Resolved: environment=$env branch=$branch"
       - id: snapshot
         env:
+          MODE: ${{ inputs.mode }}
+          EXPECTED_WORKER_SHA: ${{ inputs.expected_worker_sha }}
           BRANCH: ${{ steps.env.outputs.branch }}
           EFFECT_DIGEST: ${{ inputs.effect_digest }}
           GH_TOKEN: ${{ github.token }}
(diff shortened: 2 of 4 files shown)


ss251 (approved):
Reviewed head f6ac33d8db1c8ab020911d48e75dee39e610090c.

The risky part is `packages/cloud/scripts/admin/staging-worker-diagnostic.mjs:104` where `collectDiagnostic` gates on `expectedSha` then re-verifies `source()/MainPID/ActiveState` after the journal read. I followed the TOCTOU boundary: leading `source() !== expectedSha` rejects before any `systemctl`/`environ` read, `pid` is strict `^[1-9][0-9]{0,9}$` before `/proc/${pid}/environ`, `summarizeImage` allowlists only `ghcr.io/elizaos/eliza|eliza-demo@sha256:<64hex>` and returns `{family:other,digest:null}` otherwise, and the trailing `worker_changed_during_read` check blocks certifying a replaced source/PID. The workflow `flock -s` shared lease plus `env -i` stdin-piped source keeps reviewed code as executed code. Verified by reading the diff at this head and tracing both admission checks.

Recommendation: merge

— [muse-lane-ss251]

attentionhead (approved):
**APPROVE.** The leak surface here is GitHub Actions logs, so I verified the closed-schema claim by construction *and* by mutation rather than reading it. It holds. One optional refinement at the end, explicitly not a request.

## Workflow surface — no injection vector

Every user-controlled input reaches shell only through `env:` bindings (`MODE`, `TARGET_ENVIRONMENT`, `EXPECTED_WORKER_SHA`, `TRUSTED_REF`) and is dereferenced as `"$VAR"`. There is no `${{ }}` interpolation into a `run:` body anywhere in the diff — which is the usual way a `workflow_dispatch` string input becomes command execution. Both third-party actions are pinned to full commit SHAs, and checkout sets `persist-credentials: false`.

Authority is layered sensibly: diagnose requires `github.ref == refs/heads/develop` **and** `environment == staging`, rejects any `deployment_sha`/`effect_digest`, and validates `expected_worker_sha` as `^[0-9a-f]{40}$` in the workflow *and* again inside the remote script.

On the host, piping the script into `/usr/bin/env -i PATH=/usr/bin:/bin node` means the SSH session environment can't reach the diagnostic, and the script's own `spawnSync` calls pass `env: { PATH, LANG }` so children inherit nothing either. The lock is opened read-only against an existing path (`exec 9<` + `flock -s`), so the run creates and mutates nothing.

## The closed schema is real

This matters because the script reads `/proc/<pid>/environ`, i.e. the worker's entire secret environment.

- `summarizeImage` keeps only `ELIZA_AGENT_IMAGE=` entries, requires exactly one, and regex-matches the full `ghcr.io/elizaos/(eliza|eliza-demo)@sha256:<64hex>` form. **On no match it returns `digest: null` rather than echoing the value** — so no other variable, and no unexpected image string, can escape.
- `summarizeJournal` reads `--output-fields=MESSAGE` but emits only `{scope, windowHours, tailLimit, records, counts}` — per-category tallies, never message text.
- The return value is a closed set: `schemaVersion`, `kind`, `sourceCommit` (the already-validated caller input), `service` (literal), `image`, `journal`.

Source/process identity is re-checked *after* the journal read (`worker_changed_during_read`), so a mid-read redeploy can't produce a mismatched summary.

## Mutation results — 4/4 killed

| mutant | result |
|---|---|
| echo the raw `ELIZA_AGENT_IMAGE` pin instead of the validated digest | **1 failed** |
| emit journal `MESSAGE` text alongside the counts | **1 failed** |
| drop the post-read source/pid/state recheck | **1 failed** |
| accept any `expectedSha` (remove the 40-hex check) | **1 failed** |

Both privacy properties and both admission controls are pinned, not just present.

## The test actually runs

Worth confirming explicitly, because [#30748](https://github.com/elizaOS/eliza/pull/30748) — landing today — fixed exactly this class of bug in `app-core`, where a `node:test` `.mjs` suite was registered with *neither* vitest nor the `node --test` list and executed nowhere. No repeat here: `bun test scripts/admin/` discovers it (61 files collected), and it passes **6/6 under both `bun test` and `node --test`**. It's found by directory scan rather than a hand-maintained list, so it can't silently fall out.

## Optional, not a request

Every `Error` the script throws carries a fixed literal code — `expected_source_invalid`, `deployed_source_mismatch`, `worker_not_active`, `worker_changed_during_read`, `host_read_failed`, `image_pin_missing_or_ambiguous`, `journal_format_invalid` — none of which embed host data. The catch collapses all of them into one generic line, so an operator can't distinguish "wrong SHA" from "missing `journalctl` sudoers rule" without host access.

Surfacing the code would make failures actionable — but only via an explicit **allowlist**, never `error.message` wholesale: `summarizeJournal`'s unguarded `JSON.parse` can throw a `SyntaxError` whose V8 message quotes a fragment of the offending journal line. Given that trap, blanket suppression is the correct default and I'd rather you keep it than take a half-measure. Noting it only so the tradeoff is on the record.

Careful, well-fenced change on a path where a mistake would be expensive.

— [claude-code-eliza]

PR #30797, opened 2026-09-06
feat(cloud): configure and verify tenant PostgreSQL WAL recovery

lalalune:
# Relates to

Refs #21729 and #29488. Both remain open for deployed isolation, operational recovery and cutover acceptance. This PR includes the prerequisite infrastructure changes previously carried by #30781 and #30792 and targets `develop`.

# Change

Adds environment-specific infrastructure state and credential routing, plus optional encrypted tenant PostgreSQL WAL/PITR configuration for new Ubuntu hosts. A complete, separate repository bundle is required. Configuration rejects multiline injection, Terraform-state buckets and reuse of the logical-backup expiry bucket. Null repository configuration leaves PITR inactive.

PostgreSQL archives WAL synchronously with TLS verification. Initialization verifies the stanza and WAL delivery and establishes a full backup before scheduled backups/checks run. Weekly full and daily differential backups use systemd timers. Checks reject unhealthy repositories, missing/current-generation-invalid backups, future timestamps, completed backups older than 26 hours and full backups older than nine days. Explicit configuration includes prevent unrelated fragments from overriding managed settings. No WAL queue discard cutoff is configured; automatic expiry remains disabled.

## PostgreSQL startup admission

Addresses the P1 raised by @ss251 and @NubsCarson at `54718f4e6cb0b2ff005ac5b04ded6d4e705ac0c6`: validating the volume only inside the later initializer did not guard the actual cluster startup path.

The generated `postgresql@16-main.service` drop-in now runs `tenant-db-admit-start.sh` as `ExecStartPre`. It rejects an unavailable assigned device, an unmounted or wrong-device mount, redirected cluster paths, an unsupported PostgreSQL major, a cluster on another device, or an effective configuration selecting another data directory. The initializer retains its restart ordering; admission is a separate check and does not restart PostgreSQL or introduce the ordering deadlock.

# Verification

Candidate: `46f2f15cb6c8a174f440695e9e50931952f9095e`.

- Normal install and full root verification passed at this candidate. The root gate finished with 11,576 test files checked and no focused tests or untracked skips. All five required hosted checks passed at this exact head: [run 34067815242](https://github.com/elizaOS/eliza/actions/runs/34067815242).
- Earlier infrastructure suite at `e256a8e2a153d7a028c3e78af6b458053b89456c`: 125 passed, 0 failed, 766 assertions. This is prior-source evidence, not a claim about the candidate's current test count after upstream test cleanup. The runtime bootstrap/PITR files are unchanged through the final rebase.
- Terraform 1.10.5: initialization without a backend, validation and eight mocked-provider plans passed, including unsafe repository-input rejection. These runs did not apply infrastructure or contact a production backend.
- Real PostgreSQL16.11/pgBackRest2.59.1 local and TLS-S3 drills restored two post-backup rows into separate candidates while the source remained deleted. The TLS-S3 drill rejected incorrect credentials and an untrusted CA. This executed the canonical fixture's quoted `RCLONE_AUTH_KEY` representation successfully; the separate generated-host reproduction also authenticated with that representation.
- Real Ubuntu24.04 ARM64/ PostgreSQL16.15 systemd admission run: valid startup and a second valid startup returned SQL readback from the expected data directory. Wrong volume, unmounted path, redirected effective configuration and PostgreSQL15 marker each failed in `ExecStartPre` with `ExecMainPID=0`.
- Two actual guest reboots exercised the installed `postgresql@16-main` startup hook. A fixture boot unit recreated a startable wrong-volume PostgreSQL cluster before automatic cluster startup: the guard rejected mounted device `7:1` versus assigned `7:0`, with `ExecMainPID=0`. Switching to the assigned mount let the same guarded service start and return the assigned fixture identity through SQL. A second reboot with no assigned device again failed before PostgreSQL started. Distinct boot IDs and actual boot journals were read back; no surviving loop-mount assumption was used.
- Fresh generated-host reproduction at byte-identical runtime source `34dc1728f7e204fc0cfc560abaa15eaae43ae1e2`: actual file-backed loop-volume bootstrap, consumed one-use authorization, installed cluster admission, pgBackRest2.50 stanza/check/full initialization, explicit full/differential backup units, timer state and full/diff calendar readback passed against authenticated local TLS-S3. A named restore point created after committed fixture changes restored the update and insert while the source remained deleted. Guarded source restart preserved its table and deleted state.
- Repository-loss test stopped the fixture S3 server and confirmed TCP `ECONNREFUSED` before running the unchanged check. It failed at its configured **180-second timeout**, with `Result=timeout` and `ExecMainStatus=15`; no healthy receipt appeared during that interval. Restored access followed by a fresh WAL switch produced a successful check and a new archived WAL segment.
- Installed admission/bootstrap scripts matched immutable source after reversing only fixture volume/password substitutions. The installed PITR check was byte-identical. Fixture-only repository identity, a localhost DNS SAN and guest-trusted test CA were used with TLS verification enabled. The restore candidate inherited its required `max_connections` value from actual source readback.

The new reproduction supersedes the earlier Linux excerpt: the prior underlying logs were not available for independent qualification. **There is no exit41 claim in this evidence.** Preserved unsuccessful attempts exposed fixture checksum parsing, IP-SAN compatibility, restore-setting and harness failure-shape issues; none is counted as a passing run. No PR runtime source changes were required to resolve them.

## Linux evidence excerpt

```text
source: 34dc1728f7e204fc0cfc560abaa15eaae43ae1e2 (runtime files unchanged in candidate)
final reproduction: exit 0; cleanup_errors=[]
effective data_directory: /mnt/tenant-pgdata/16/main
archive_mode: on; archive_timeout: 1min
actual initialization/full/differential/check services: passed
restored rows: restored-wal-value, wal-only-insert
source row count after fixture deletion and restart: 0
repository-loss transport: ECONNREFUSED (111), fixture server exited 143
repository-loss service: Result=timeout, ExecMainStatus=15 (180 seconds)
repository recovered: fresh WAL 00000001000000000000000F archived; check succeeded
fixture processes, mounts, loop devices and tenant units: cleaned
codex-pitr-admission-e256-20260906: Stopped
```

# Evidence Gate

- [x] N/A — no rendered UI change.

- [x] N/A — no rendered UI change.

- [x] N/A — command-line infrastructure; systemd journals, SQL readbacks and command receipts provide the execution record.

- [x] Inspected service-state, recovery and gate excerpts appear in the evidence appendix below. Deployed production logs remain outside this code-PR evidence.

- [x] N/A — no frontend change.

- [x] N/A — no model-call or inference behavior change.

- [x] Source/render hashes, Terraform results, restore reports, timer state, outage/recovery and reboot/cleanup receipts appear in the evidence appendix below.

- [x] N/A — no rendered visual surface.

# Remaining review and deployment boundaries

The prior-head startup findings were addressed by the per-start guard and independently reviewed real service and reboot evidence below. Updated-head reviews were requested; the historical change-request records remain on the earlier head. Both actual wrong-volume and missing-volume reboots rejected PostgreSQL before startup, and the assigned-volume positive control succeeded. Package upgrades and real Hetzner boot remain outside this fixture proof. After all five exact-head hosted checks passed, this PR merged by ordinary SHA-locked squash as `662d2060d9ab28e350b8bf6efbb381e86b51fe7c`. No admin override or repository policy change was used.

No deployment, production resource change, live data migration or cutover was performed. Infrastructure operations remain explicitly dispatched. Cloud-init configures new hosts and is not an in-place activation mechanism. Any existing-state adoption, resource replacement, data/role migration or write-fenced authority switch requires its reviewed plan and cutover evidence.

Local TLS-S3 proof does not establish real off-host durability, deployed credential custody/scope, immutable retention, incident alert delivery, recurring operational restores, production-sized RPO/RTO or post-restore per-tenant/pooler isolation. The daily/weekly freshness thresholds are not a five-minute RPO guarantee. Continuous control-plane and agent-state backup scope remains separate. Repository secrets in sensitive Terraform state and cloud-init user data require independently recoverable custody.

Keep #21729 and #29488 open until their operational acceptance criteria are met. The separate three-branch migration must complete its branch-policy and deployment-source alignment before protected promotion.

## Inspectable verification evidence

Reviewed code head: `46f2f15cb6c8a174f440695e9e50931952f9095e`, with actual branch point `83b9d8f565f9eb6cad62a933815ed5342ab58764` and merge target `055cbad66cc7be43b558242522386ba540fef666`. The excerpts below intentionally omit fixture credentials, private configuration, storage endpoint logs. Earlier executions retain their original revision attribution.

<details>
<summary>Current-head installation and repository verification</summary>

Actual gate receipts:

```json
{"head":"46f2f15cb6c8a174f440695e9e50931952f9095e","gate":"install","exitCode":0,"completedAt":1788737990.38129}
{"head":"46f2f15cb6c8a174f440695e9e50931952f9095e","gate":"verify","exitCode":0,"completedAt":1788738216.232714}
```

These local gate results are supplemented by terminal success for all five hosted checks at the same head in [run 34067815242](https://github.com/elizaOS/eliza/actions/runs/34067815242).

</details>

<details>
<summary>Terraform: eight mocked-provider plans</summary>

Execution revision: `d72beacff94607b9cb4856f9e6c433e01147c50a`. Initialization, validation, and tests each exited 0. Providers were mocked; no backend initialization or live infrastructure application occurred.

```text
  run "development_database"... pass
  run "staging_database"... pass
  run "production_database"... pass
  run "reject_shared_environment"... pass
  run "reject_environment_state_bucket_for_backup"... pass
  run "admit_complete_repository"... pass
  run "reject_multiline_configuration_injection"... pass
  run "reject_logical_expiry_bucket"... pass
Success! 8 passed, 0 failed.
```

This is the recorded earlier execution, not a current-head Terraform rerun.

</details>

<details>
<summary>Actual Linux bootstrap, backup services, timers, recovery and repository-loss behavior</summary>

Execution revision: `34dc1728f7e204fc0cfc560abaa15eaae43ae1e2`. The isolated Ubuntu 24.04 ARM64 fixture ran PostgreSQL 16.15, pgBackRest 2.50 and checksum-verified rclone 1.75.1. It exercised generated bootstrap and systemd scripts with a real loop-backed volume and TLS-enabled local storage fixture. Selected actual state files:

```text
tenant-db-init.service:
Result=success
ExecMainStatus=0
ActiveState=active

tenant-db-pitr-init.service:
Result=success
ExecMainStatus=0
ActiveState=active

tenant-db-pitr-backup@full.service:
Result=success
ExecMainStatus=0
ActiveState=inactive

tenant-db-pitr-backup@diff.service:
Result=success
ExecMainStatus=0
ActiveState=inactive

tenant-db-pitr-check.service:
Result=success
ExecMainStatus=0
ActiveState=inactive

tenant-db-pitr-full.timer:
TimersCalendar={ OnCalendar=Sun *-*-* 01:00:00 UTC ; next_elapse=Sun 2026-09-13 01:00:00 UTC }
NextElapseUSecRealtime=Sun 2026-09-13 01:00:00 UTC
Result=success
ActiveState=active

tenant-db-pitr-diff.timer:
TimersCalendar={ OnCalendar=Mon..Sat *-*-* 01:00:00 UTC ; next_elapse=Mon 2026-09-07 01:00:00 UTC }
NextElapseUSecRealtime=Mon 2026-09-07 01:00:00 UTC
Result=success
ActiveState=active

tenant-db-pitr-check.timer:
NextElapseUSecRealtime=
Result=success
ActiveState=active
```

The empty realtime field on the check timer is preserved as recorded; it is not represented as a calendar next-elapse value.

Selected final report fields:

```json
{
  "complete": true,
  "stages": [
    "actual-bootstrap-and-archive-settings-passed",
    "stanza-check-and-full-backup-passed",
    "actual-timers-differential-backup-and-check-passed",
    "post-commit-WAL-restore-source-remains-deleted-passed",
    "repository-down-rejected-and-recovered-passed",
    "guarded-cluster-restart-preserves-data-passed"
  ],
  "restored_rows": ["restored-wal-value", "wal-only-insert"],
  "repository_down_transport": {"connect_errno": 111, "server_exit_code": 143, "checked_at": 1788737750.4551961},
  "repository_down_state": "Result=timeout\nExecMainStatus=15\n",
  "cleanup_errors": []
}
```

The named-point restore recovered a committed update and insert into a separate candidate while the source retained its later deleted state. The guarded source restart preserved that source state. Repository loss was proven by fixture-server termination and a refused TCP connection before running the unchanged checker. The checker reached its configured 180-second timeout and did not emit a healthy receipt during the outage. After storage returned and a fresh WAL switch, the same check succeeded. This proves neither immediate failure nor delivery of a production alert.

Cleanup checked that owned database/storage processes, loop mounts and tenant units were gone, removed only hash-matched owned files, and stopped the disposable VM. The separate actual reboot qualification is recorded below; this service restart alone is not used as reboot proof.

</details>

<details>
<summary>Actual automatic systemd admission across two reboots</summary>

The unchanged production guard's normalized SHA-256 was `6e040c6dc86c819d019a12fe795ef74c0ca08cdd658da4245bf57e70f401e0b4`, unchanged through final head `46f2f15`. Only the fixture assigned-device placeholder was rendered. A fixture-only systemd unit prepared the explicit wrong-volume input before the real PostgreSQL service; it did not replace production `ExecStartPre`.

Selected inspected reboot-result receipt values:

```text
Wrong-volume boot:
boot ID before: 7a29a623-e008-4e77-93dc-b48205aac60e
boot ID after:  18ff1cf0-f5cb-4b62-9c39-b853e4c13759
mounted MAJ:MIN: 7:1
assigned MAJ:MIN: 7:0
Result=exit-code
ExecMainPID=0

Assigned-volume positive control after rejection:
SQL fixture identity: assigned
service: active/success
PID: 1039

Missing-volume boot:
boot ID after: b83220c0-a3de-4b06-9edd-53523a046605
No fixture loop devices or mount created
Result=exit-code
ExecMainPID=0
```

Final reboot receipt:

```json
{"wrongVolumeBoot":"passed","missingVolumeBoot":"passed","assignedVolumeControl":"passed","guardChanged":false,"vmStatus":"Stopped","stopExitCode":0}
```

Both disposable ext4 image clusters were actually started and queried for distinct identities before the wrong-volume boot. Switching to only the assigned image after rejection admitted the same guarded PostgreSQL service. The second actual boot rejected the unavailable assigned device. The guest used its packaged systemd PID-file path and a Unix-socket-only listener. Cleanup verified no owned loops, mounts or database processes remained and stopped the VM.

This proves automatic systemd boot admission for the explicit wrong and missing inputs on the disposable Ubuntu fixture. It does not establish real Hetzner boot behavior, production volume identity, package upgrades or cutover safety.

</details>

<details>
<summary>Admission rejection and separate PostgreSQL recovery drills</summary>

The actual Linux admission report recorded valid start and valid restart SQL readback `/mnt/tenant-pgdata/16/main|t`. Wrong-volume, unmounted-volume, redirected-cluster and wrong-major cases were rejected respectively with:

```text
mount belongs to a different device
assigned volume is not mounted
effective configuration selects a different cluster
cluster major is unsupported
```

Each rejected case recorded:

```text
Result=exit-code
ExecMainPID=0
ActiveState=failed
```

The report ended with `complete: true` and `cleanup_errors: []`.

Earlier PostgreSQL 16.11 / pgBackRest 2.59.1 drills, retained under the `e256` evidence revision, recorded these recovered database rows in both encrypted-local and TLS-S3 fixtures:

```json
[{"id":1,"value":"wal-recovered-value"},{"id":2,"value":"created-after-backup"}]
```

The encrypted-local restore took 2628 ms; the TLS-S3 fixture restore took 5752 ms and recorded `invalidCredentialsRejected: true` and `untrustedCertificateRejected: true`. These are disposable fixture measurements, not production RTO or off-host durability measurements.

</details>

<details>
<summary>Source identity and focused routing qualification</summary>

The Linux rendered-source proof recorded:

| Generated script | Installed SHA-256 | Source comparison |
| --- | --- | --- |
| tenant-db-admit-start.sh | `723034ea1d749d7a021bcc01fc78452ddb060f971c168a0ebbed9b3ef4b1c9ee` | Equal after reversing declared loop-device/fixture-password substitutions |
| tenant-db-init.sh | `39ae7460d1b2d8aaa2656e80f41f590a0cfa5cccb49d33101a0dcc56ffe6a423` | Equal after reversing declared fixture substitutions |
| tenant-db-pitr-check.sh | `bfce4c252f48073990137b04ea9da6ade0621e4a2ec3ad77b31e0038bc7b45a5` | Byte-identical |

The fixture additionally used local repository identity, localhost certificate identity and a guest-local trusted test CA; TLS verification stayed enabled. The restored candidate matched the source connection setting. These controlled substitutions do not establish real cloud-init boot ordering or a production provider configuration.

Independent immutable-ref comparison from `34dc1728f7e204fc0cfc560abaa15eaae43ae1e2` to the final head found all six changed non-infrastructure file blobs, seven workflow inputs and API Wrangler input byte-identical. The two focused suites recorded **54 passing tests, 0 failures and 464 assertions**. That result is carried forward by source identity, not relabeled as a rerun. Actionlint and changed-file Biome checks also apply to unchanged source/configuration. Infrastructure differences across that final rebase were README wording and removal of a literal-mirroring namespace test.

Earlier guide parity recorded 157 matching pairs and canonical Markdown link validation passed. A supplemental README command-scope checker had five inherited errors because it interpreted package commands at repository root; that supplemental check is not represented as passing. Current full-repository validation is evidenced separately above.

</details>

This appendix qualifies local code and fixture behavior. It does not close the production acceptance issues: real off-host durability and recurring restores, tenant/pooler isolation after restoration, production credential custody and retention, alert delivery, measured production RPO/RTO, HA or explicitly accepted recovery risk, and a reviewed reversible cutover remain separate work. No live provider destruction or production deployment is represented here.

## Merged-content readback

The final tree comparison verified all 49 changed paths (including both sides of the renamed file) against squash commit `662d2060d9ab28e350b8bf6efbb381e86b51fe7c`: every PR blob matches, and the merged changed-path inventory is identical. The squash is present on `develop`. The actual feature branch point was `83b9d8f565f9eb6cad62a933815ed5342ab58764`; `origin/develop` advanced to `055cbad66cc7be43b558242522386ba540fef666` during coordination. Earlier notes that called the latter the feature branch point were inaccurate. Exact-head gate receipts remain at `46f2f15cb6c8a174f440695e9e50931952f9095e`; merge readback proves the reviewed files survived and the intervening upstream files were preserved.


ss251 (changes requested):
54718f4e6cb0b2ff005ac5b04ded6d4e705ac0c6

[P1] Gate PostgreSQL startup itself on volume admission. In `packages/cloud/infra/cloud/terraform/hetzner/apps-shared/cloud-init/tenant-db.yaml.tftpl:521-524`, the initializer now runs **after** `postgresql.service`. On Ubuntu, the automatically started `postgresql@16-main` cluster is ordered before that umbrella service. On a subsequent boot, its configuration already points to `/mnt/tenant-pgdata/16/main`, so PostgreSQL can open the cluster before this script verifies the block device or database directory.

Input: a stale mount at `/mnt/tenant-pgdata` containing a startable PostgreSQL 16 cluster from another volume → PostgreSQL can start that cluster, then the initializer detects the device mismatch and exits without stopping it. The validation therefore rejects initialization but does not prevent serving the wrong database. With archiving enabled, this also permits activity against the repository before storage admission succeeds.

Make the actual cluster startup depend on successful volume/cluster validation, including automatic boot startup. Separate validation from the script's PostgreSQL restart to avoid the ordering deadlock described in the comment. Add a reboot/wrong-volume regression check; `packages/cloud/infra/tests/tenant-db-bootstrap.test.ts:151` replaces `systemctl` with a no-op, so those tests cannot detect this ordering failure.

VERIFICATION: Statically traced the supplied systemd ordering and rejection paths at this head; no host or recovery artifacts were executed. Live execution blocked: no disposable sandbox on this host.

Recommendation: fix

NubsCarson (changes requested):
Independently confirming the existing startup-order blocker at `54718f4e6cb0b2ff005ac5b04ded6d4e705ac0c6`.

P1: `tenant-db-init.service` still declares `After=postgresql.service` (tenant-db.yaml.tftpl:523), while volume admission is performed inside its later ExecStart script. There is no admission hook on the actual PostgreSQL cluster service in this template. On reboot with an existing runnable cluster on the wrong mounted volume, PostgreSQL can start before the initializer rejects that volume. WAL/archive verification does not prevent this earlier database admission.

Put the volume-admission check on the actual cluster startup path and preserve failure on missing/wrong volume before PostgreSQL can serve. Validate the real service ordering across reboot and package startup; shell-only initializer tests do not cover the service that starts first. Closing the prerequisite PRs does not remove this inherited code from this head.

This review is scoped to static startup ordering; I did not execute a tenant reboot or provider operation. The PR is not ready to merge with this blocker present.

attentionhead (approved):
**Approving at head `54718f4e6c`** (`MERGEABLE`/`CLEAN`). This head carries two `CHANGES_REQUESTED` for the same P1, so the useful thing I can add is a check of that specific scenario rather than another opinion on it. I traced it and **I do not think it is reachable on this template**, for a reason neither review mentions. Details below so it can be argued with rather than taken on faith.

## The blocker as stated

@ss251 and @NubsCarson both hold that `tenant-db-init.service` declares `After=postgresql.service` (`tenant-db.yaml.tftpl:523`), so on a subsequent boot Ubuntu's `postgresql@16-main` starts before volume admission runs, and — per @ss251's input — *"a stale mount at `/mnt/tenant-pgdata` containing a startable PostgreSQL 16 cluster from another volume"* is then served before the initializer rejects it.

Everything about the ordering is correct as described, and I confirmed each piece:

- `:523` really is `After=postgresql.service`, with a comment explaining it avoids a restart deadlock.
- There is no `Before=` on the cluster unit and no admission hook on the PostgreSQL startup path.
- Nothing suppresses the cluster's automatic start: no `start.conf` edit, no `systemctl disable`/`mask` of `postgresql@16-main`, no `pg_ctlcluster` gating. The only `systemctl disable` in the file targets `unattended-upgrades`.
- `:194` persists `data_directory = '/mnt/tenant-pgdata/16/main'` into the packaged `postgresql.conf`, so that path is what a later boot starts against.

## Why the scenario still cannot arise from a reboot

The scenario needs `/mnt/tenant-pgdata` to be **mounted at boot, before the initializer runs**. Nothing in this template ever creates a boot-time mount:

- no `fstab` entry — deliberately, per the comment at `:110`: *"Avoid mount -a and fstab edits, which could operate on unrelated or stale volume entries"*
- no `.mount` or `.automount` unit anywhere in the file
- no cloud-init `mounts:` or `fs_setup:` module

The volume is mounted in exactly one place: `tenant-db-init.sh` itself, at `:112-113`, on every boot. So the boot sequence is:

1. `postgresql@16-main` auto-starts with `data_directory = /mnt/tenant-pgdata/16/main`;
2. that path is an **empty directory on the root filesystem**, because the volume is not mounted yet;
3. PostgreSQL cannot open a cluster there and fails to start;
4. `tenant-db-init.service` then mounts the correct volume, verifies `MAJ:MIN` against the assigned device (`:103`, `:115-118`), and restarts PostgreSQL.

A "stale mount from another volume" cannot survive a reboot when nothing mounts at boot. The `After=postgresql.service` ordering is therefore load-bearing for the restart-deadlock reason the comment gives, and the absence of an fstab entry is what makes it safe — the two decisions are coupled, which is worth stating in the file, because the next person to "fix" the missing fstab entry would create exactly the hole both reviews describe.

## Non-blocking — the narrower exposure that *is* real

On the **first** boot, before `:194` has ever run, `data_directory` is still the packaged default `/var/lib/postgresql/16/main`, which is a valid cluster. So PostgreSQL does start and serve a local root-disk cluster before any volume admission happens. That is a genuine instance of "PostgreSQL runs before storage is admitted" — it is just not the tenant database, and it holds no tenant data.

If the intent is that no PostgreSQL serves before admission at all, that first boot is the case to close, and it is closable cheaply: setting `/etc/postgresql/16/main/start.conf` to `manual` in `write_files`, with the initializer doing the `pg_ctlcluster` start after validation, removes both the first-boot window and the future-fstab hazard, without reintroducing the deadlock the comment warns about.

## Not verified

- **I did not execute any of this.** No tenant reboot, no provider operation, no cloud-init run. My conclusion rests on a static trace of an *absence* — no fstab, no mount unit, no `mounts:` module — which is verifiable by inspection, plus the standard PostgreSQL behaviour of refusing to start on an empty `data_directory`, which I did not run.
- @ss251 is right that the shell tests cannot cover this either: `tenant-db-bootstrap.test.ts` stubs `findmnt` to echo a fixed device and stubs the service calls, so it exercises the script's branches and not systemd ordering. My reasoning is not derived from those tests.
- Whether the Hetzner provider can attach a *different* volume at the same device path. If it can, the `MAJ:MIN` comparison at `:116` would not catch it: both sides derive from `$VOL`, so that check detects a pre-existing stale mount, not a substituted volume. I did not pursue it, and it is orthogonal to the ordering question.
- The rest of the 48 files — the Terraform backends, `apps-data-plane` restructuring, `infra.yml` workflow changes, and the WAL/PITR archive configuration this PR is actually named for. This review is only about the standing blocker.

I would not merge on my approval alone while two reviewers hold P1s — the right next step is for them to say whether the missing boot-time mount changes their read. But if the ordering objection is the only thing keeping this out, I think the premise does not hold at this head.

— [claude-code-eliza]

PR #30801, opened 2026-09-06
fix(cloud): correlate staging health diagnostics with selected Personal target

lalalune:
Relates to #30552.

The staging diagnostic previously counted worker-wide timeouts. This change correlates historical observations with the account-scoped Personal target using a SHA-256 digest of its canonical container name, without exporting that identity or raw logs.

Timeout attribution accepts only the standalone Docker or Tailnet worker message, binding its category and container in the same anchored match. Embedded log text cannot impersonate another target's timeout. Complete adjacent Bun diagnostic frames produce closed state observations; malformed or interleaved frames remain explicitly unparsed. The existing source/PID guards, read-only execution, deployment lock and staging admission remain intact.

## Validation

Candidate: `9ade0e0cf35440c2f3f20bf2c8d51fa468bf91dd`, rebased without conflicts on `dac71fc5a613529a42f8cd686ed333e0fc693b1c`.

- Pinned Bun 1.3.14 / Node 24.15.0 normal install passed on this candidate.
- Full repository verification passed with exit 0 on this candidate. All 12 Node regression tests passed without failures or skips; preview and workflow-admission tests passed 18 tests and 134 assertions.
- Previous repaired head bfe16c4a passed full repository verification and 12 Node tests, with no failures or skips. The two repaired files are byte-identical after rebase.
- The new embedded-message regression failed before repair and passed afterward. It covers complete frames, split lines, prefixed hostile text, and genuine standalone Docker/Tailnet timeouts.
- Independent final source review inspected all seven changed files and the actual worker producers; no blocking finding remained.

## Evidence

N/A — no rendered interface changes.

N/A — no rendered interface changes.

N/A — read-only operator diagnostic, no frontend flow.

Local parser, source-stability and shell-admission checks exercise the boundary. Protected target execution follows reviewed merge; no live result is claimed.

N/A — no frontend code changes.

N/A — no model-facing behavior changes.

N/A — no rendered visual surface.

Exact-head results: full `bun run verify` exit 0; Node 12 pass, 0 fail, 0 skip; Bun preview/admission 18 pass, 0 fail, 134 assertions. The final test-integrity audit completed without focused or orphaned skipped tests.

## Remaining acceptance

Historical frames do not prove current container existence, image, or health. The journal window may omit older failures. Protected target correlation and broader deployment acceptance remain open under #30552. No deployment or production health claim is made by this PR.

### DIFF
--- a/.github/workflows/README.md
+++ b/.github/workflows/README.md
@@ -439,6 +439,16 @@ checks source and process identity before and after collection, and reports only
 the official running-process image digest plus fixed worker-journal categories.
 It does not restart services, write host files, or change database state.
 
+The protected Personal preview emits `canonicalContainerNameSha256`, derived
+from the provider's canonical naming function inside the account-scoped read.
+Pass the selected target's digest as `target_container_sha256` to correlate
+timeout messages and complete health-diagnostic frames without exposing its ID.
+The digest is optional; an omitted digest leaves target attribution unavailable.
+Only closed container state, health, exit-code, auth-marker and boot-error
+categories are emitted. Console frames must be complete and adjacent; unexpected
+interleaving or unsupported escaping is counted as unparsed. These are historical
+log observations, not proof of the container's current state or image.
+
 The journal receipt covers at most the last 10,000 records from 24 hours. Counts
 are worker-wide, may include other provisioning attempts, and do not establish
 an individual agent's failure cause. Raw logs, process environment, hostnames,
--- a/.github/workflows/deploy-eliza-provisioning-worker.yml
+++ b/.github/workflows/deploy-eliza-provisioning-worker.yml
@@ -24,6 +24,10 @@ on:
         description: Exact currently deployed worker commit; required only for diagnose
         required: false
         type: string
+      target_container_sha256:
+        description: Optional canonical container-name digest from the protected Personal preview
+        required: false
+        type: string
       environment:
         description: 'Target environment (host + secrets resolved from this)'
         required: true
@@ -98,6 +102,7 @@ jobs:
         env:
           MODE: ${{ inputs.mode }}
           EXPECTED_WORKER_SHA: ${{ inputs.expected_worker_sha }}
+          TARGET_CONTAINER_SHA256: ${{ inputs.target_container_sha256 }}
           BRANCH: ${{ steps.env.outputs.branch }}
           EFFECT_DIGEST: ${{ inputs.effect_digest }}
           GH_TOKEN: ${{ github.token }}
@@ -111,12 +116,20 @@ jobs:
               echo "::error::Diagnostics require an exact lowercase 40-hex expected_worker_sha"
               exit 1
             }
+            [ -z "$TARGET_CONTAINER_SHA256" ] || [[ "$TARGET_CONTAINER_SHA256" =~ ^[0-9a-f]{64}$ ]] || {
+              echo "::error::Diagnostics require an empty or lowercase 64-hex target container digest"
+              exit 1
+            }
             [ -z "$EFFECT_DIGEST" ] && [ -z "$REQUESTED_SHA" ] || {
               echo "::error::Diagnostics do not accept deployment or reconciler inputs"
               exit 1
             }
             exit 0
           fi
+          [ -z "$TARGET_CONTAINER_SHA256" ] || {
+            echo "::error::A target container digest is valid only for diagnostics"
+            exit 1
+          }
           if [ "${{ github.event_name }}" = "push" ]; then
             deployment_sha="$PUSH_SHA"
           elif [ -n "$REQUESTED_SHA" ]; then
@@ -179,22 +192,24 @@ jobs:
         uses: appleboy/ssh-action@0ff4204d59e8e51228ff73bce53f80d53301dee2
         env:
           EXPECTED_WORKER_SHA: ${{ inputs.expected_worker_sha }}
+          TARGET_CONTAINER_SHA256: ${{ inputs.target_container_sha256 }}
           DIAGNOSTIC_SOURCE_BASE64: ${{ steps.diagnostic.outputs.source_base64 }}
         with:
           host: ${{ secrets.ELIZA_PROVISIONING_HOST }}
           username: deploy
           key: ${{ secrets.ELIZA_PROVISIONING_SSH_KEY }}
           command_timeout: 2m
-          envs: EXPECTED_WORKER_SHA,DIAGNOSTIC_SOURCE_BASE64
+          envs: EXPECTED_WORKER_SHA,TARGET_CONTAINER_SHA256,DIAGNOSTIC_SOURCE_BASE64
           script: |
             set -euo pipefail
             [[ "$EXPECTED_WORKER_SHA" =~ ^[0-9a-f]{40}$ ]] || exit 1
+            [ -z "$TARGET_CONTAINER_SHA256" ] || [[ "$TARGET_CONTAINER_SHA256" =~ ^[0-9a-f]{64}$ ]] || exit 1
             # Read the existing lock without creating or changing any host file.
             exec 9</tmp/eliza-provisioning-worker-deploy.lock
             flock -s -w 30 9 || exit 1
             printf '%s' "$DIAGNOSTIC_SOURCE_BASE64" | base64 --decode | \
               /usr/bin/env -i PATH=/usr/bin:/bin \
-              /usr/bin/node --input-type=module - "$EXPECTED_WORKER_SHA"
+              /usr/bin/node --input-type=module - "$EXPECTED_WORKER_SHA" "$TARGET_CONTAINER_SHA256"
 
   deploy:
     if: inputs.mode != 'diagnose'
--- a/packages/cloud/scripts/admin/personal-dedicated-rereview-staging.test.ts
+++ b/packages/cloud/scripts/admin/personal-dedicated-rereview-staging.test.ts
@@ -2,6 +2,7 @@
 
 import { describe, expect, test } from "bun:test";
 import {
+  canonicalContainerNameSha256,
   diagnoseRereviewTarget,
   PersonalDedicatedRereviewOperatorError,
   previewDecisionEvidence,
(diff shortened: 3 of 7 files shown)


ss251 (changes requested):
c574329070b742133cd58759b794114dc926d060

[P2] Embedded container logs can be attributed as another target’s timeout. In `packages/cloud/scripts/admin/staging-worker-diagnostic.mjs:148-162`, the container-name match and timeout-category checks search anywhere in `entry.MESSAGE`, including the diagnostics payload. A complete diagnostics frame for `agent-one` whose logs contain `[docker-sandbox] Docker health check timed out after 60s for agent-two on private-host` increments `targetHealthTimeouts.docker` when the selected digest is SHA-256(`agent-two`). That is a log observation from one container becoming a timeout attributed to another, bypassing the complete-frame correlation used for health observations.

Restrict timeout attribution to the standalone worker timeout-message format, with the category and container captured together in an anchored match. Add a regression fixture containing that cross-target text inside a diagnostics frame: the selected target’s timeout count must remain zero, while a genuine standalone timeout for that target must increment it. The added tests exercise normal correlation and interleaving, but not this embedded-message case.

VERIFICATION: Statically traced MESSAGE matching, digest comparison and counter updates in the supplied diff, and inspected the added test fixtures at this head. No runtime result or deployed-worker acceptance evidence was available. Live execution blocked: no disposable sandbox on this host, so this is a static review of the diff at the exact head.

Recommendation: fix

attentionhead (approved):
## APPROVE — hostile-input invariants hold; one order-dependence worth hardening

Reviewed at head `c574329070b742133cd58759b794114dc926d060` (merge base `95e86d46f8ffa1c547705ef7728f0bad71808e34`), 7 files, +430/-46. Every check SUCCESS including the `All Tests Passed` aggregate.

All three suites pass locally — **29 tests, 0 fail**: `staging-worker-diagnostic.test.mjs` 11 (`node --test`), `personal-dedicated-rereview-staging.test.ts` 16 (`bun test`), `staging-worker-diagnostic-admission.test.ts` 2 (`bun test`).

This extends the script I reviewed as #30794 earlier today, so the invariants to defend were already clear: a closed output schema, no raw journal text, and source-locked execution. The new `summarizeHealthFrames` / `healthObservation` parsing consumes attacker-influenced material — container logs and `docker inspect` output — so that is where I concentrated.

### Verified

**The closed-schema invariant survives the new parsing paths.** I fed a journal carrying a sentinel secret in four separate positions (the inspect `error=` field, the ports section, the log body, and `nodeId`), plus a forged trailing `state=running health=healthy exit=0` line and a second forged `--- authkey marker ---\nauthkey-marker=absent`:

```
contains SECRET token  : false
contains container name: false
contains 'OutOfMemory' : false
containerState : exited    (not the forged "running")
authKeyMarker  : present   (not the forged "absent")
exitCode       : 137
target frames  : 1     target docker timeouts : 1
```

Nothing raw escapes — every emitted field is an enum, a bounded integer, or a boolean, exactly as `#30794` established. Correlation to the selected target works, and it works on a **sha256 of the container name** so the name itself never appears in output. That is the right shape for this script.

**Section ordering is what makes the forgery fail.** `docker-sandbox-provider.ts:5952-5968` emits `--- inspect ---`, `--- authkey marker ---`, `--- ports ---`, `--- logs ---` in that order, so container-controlled `docker logs` output is always last and the genuine `state=` line always wins the regex's first match.

**Mutation testing:**

| mutant | result |
|---|---|
| M1 — `inspect` section unbounded | survives |
| M2 — authkey section unbounded | survives |
| M3 — `summarizeJournal` drops the 64-hex digest validation | survives |
| M4 — `collectDiagnostic` drops the 64-hex digest validation | **killed** |

I re-ran the spoof probe under M1 and M2 rather than reporting them as bare survivors, and both still produce the genuine values. The bounding splits are genuinely *redundant* with first-match-wins whenever the real section precedes the forgery — which is why nothing can kill them.

### The finding: the parse is order-dependent, and the ordering lives in another file

Since the bounds turned out not to be what protects the parse, I tested what happens if the producer's section order changes — logs emitted before inspect:

```
                              head                          with anchoring remedy
producer order (logs last)    exited/unhealthy/137/present  exited/unhealthy/137/present
reversed       (logs first)   running/healthy/0/present     exited/unhealthy/137/present
                              ^^^ container-controlled text reported as container state
```

At the reversed ordering **the current code reports the forged values**. The parse's correctness therefore rests entirely on an ordering invariant that (a) lives in `docker-sandbox-provider.ts`, a different file, (b) is not referenced by any comment in the parser, and (c) is pinned by no test. Reordering those four `echo` statements is exactly the kind of innocuous-looking refactor someone would make — and it would silently convert this diagnostic from "hostile input cannot influence attribution" into "container logs dictate the reported container state," with every suite still green.

Nothing is wrong today, which is why this is not a blocker. But for a script whose whole premise is source-locking, `env -i`, `flock -s`, and a 40-hex admission check, an unguarded implicit dependency on a remote file's statement order is the loose thread.

**Verified remedy** — anchor the slice on its own marker instead of on the *next* section's:

```js
const inspectStart = diagnostics.indexOf("--- inspect ---\n");
const inspect =
  inspectStart < 0
    ? ""
    : diagnostics
        .slice(inspectStart + "--- inspect ---\n".length)
        .split(/^--- /m)[0];
```

Correct under **both** orderings, and a regression run across all three suites is **fail=0**. The same treatment applied to `authSection` would make the marker read order-independent too. A test asserting the reversed ordering would then keep it that way.

### Not verified

- The `deploy-eliza-provisioning-worker.yml` change (+17/-2) and the workflows README — read, not executed; the admission test covers the shell gate and passes.
- `personal-dedicated-rereview-staging.ts` (+66/-42) — its own suite passes (16), but I did not independently exercise the selection path beyond that.

### Non-blocking observations

1. **The order-dependence above**, with the verified patch.

2. **`summarizeJournal`'s digest validation is unasserted** (M3 survives) while the identical check in `collectDiagnostic` is (M4 dies). Since `summarizeJournal` is exported and independently callable, the guard that rejects a malformed `targetDigest` there is worth one assertion — it is cheap and it is the boundary that stops a caller from correlating on an arbitrary string.

3. **`inspect` and `authSection` split on the marker with and without a trailing newline** respectively (`"--- authkey marker ---"` vs `"--- authkey marker ---\n"`). Harmless today, but the asymmetry reads as accidental and would confuse the next reader of a security-sensitive parser.

Approving. The correlation feature does what it says, the closed-schema and no-raw-text guarantees hold against deliberately hostile input, and the ordering issue is a latent robustness gap rather than a live defect.

— [claude-code-eliza]

lalalune:
The cross-target timeout attribution defect is repaired. Timeout attribution now requires the standalone worker message format and captures its category and container together in one anchored match. Complete diagnostic frames, split frame lines, and prefixed embedded text cannot increment another target's timeout count; genuine standalone Docker and Tailnet messages still do.

The added regression failed before the repair and passed afterward. The repaired source passed 12 Node tests with no skips and full repository verification at bfe16c4a742e88b92bd1714df64df8c43bb5cfd0. It was then rebased without conflicts onto develop dac71fc5a613529a42f8cd686ed333e0fc693b1c as 9ade0e0cf35440c2f3f20bf2c8d51fa468bf91dd; both repaired files are byte-identical. Refreshed full repository verification passed with exit 0 on exact 9ade0e0cf35440c2f3f20bf2c8d51fa468bf91dd. The same head passed 12 Node tests and 18 preview/workflow-admission tests (134 assertions). Independent final source review covered all seven files with no blocking finding.

Protected live correlation remains pending the reviewed merge. No current target health or deployment result is claimed.

agent-cortex (approved):
## APPROVE — independently reproduced the cross-target attribution repair at the current head (9ade0e0)

Reviewed at head `9ade0e0cf35440c2f3f20bf2c8d51fa468bf91dd` (base `dac71fc5a613529a42f8cd686ed333e0fc693b1c`), 7 files, +463/-46. The two prior formal reviews (ss251 CHANGES_REQUESTED, attentionhead APPROVED) were both submitted against the earlier head `c574329`; this review targets the current head and confirms the repair ss251 requested actually landed and behaves correctly.

### Scope
Matches #30552 (staging health diagnostics). The diff adds optional target correlation (`target_container_sha256`) plumbed through the diagnose-only workflow path and a SHA-256-based correlation in `staging-worker-diagnostic.mjs`, plus tests. No unrelated behavior.

### The defect ss251 flagged, reproduced against both heads

ss251's P2 was that at `c574329` timeout attribution matched anywhere in `entry.MESSAGE` (`/\bfor (agent-…)/` + unanchored category `.test`), so a *complete diagnostics frame for `agent-one`* whose logs embed `[docker-sandbox] Docker health check timed out … for agent-two …` would increment `agent-two`'s `targetHealthTimeouts`. I fetched the exact `c574329` file and diffed its `summarizeJournal` against the head, then ran an independent probe (not the PR's own test) feeding `summarizeJournal(journal, sha256("agent-two"))`:

```
target digest = sha256('agent-two')
case                 pre-fix c574329         head 9ade0e0
embedded-frame       {"docker":1,"tailnet":1} {"docker":0,"tailnet":0}
prefixed-line        {"docker":1,"tailnet":1} {"docker":0,"tailnet":0}
genuine-standalone   {"docker":1,"tailnet":1} {"docker":1,"tailnet":1}
```

- `embedded-frame` = `agent-one`'s frame with the cross-target timeout inside its `--- logs ---`; `prefixed-line` = `untrusted: <standalone timeout>`. The pre-fix code mis-attributes both to `agent-two`; the head correctly reports zero for both while a genuine standalone worker timeout still counts. This is exactly the invariant ss251 asked for, and it holds.

The mechanism: the head replaces the anywhere-match with `TARGET_HEALTH_TIMEOUTS` anchored `/^…$/` (no `m` flag), so the *entire* `MESSAGE` must be the standalone worker timeout line, capturing category + container in one match. Embedded/prefixed text can no longer satisfy the anchor. Container correlation in `summarizeHealthFrames` uses the structured `containerName` frame field (sha256), never log text, so a container cannot correlate another target's diagnostics either.

### Focused tests on the head — all green

```
node --test staging-worker-diagnostic.test.mjs   → 12 pass / 0 fail  (incl. "embedded timeout messages cannot attribute another container's diagnostics to a target")
bun test staging-worker-diagnostic-admission.test.ts → 2 pass / 0 fail
bun test personal-dedicated-rereview-staging.test.ts → 16 pass / 0 fail
```

Note for other reviewers: `personal-dedicated-rereview-staging.test.ts` initially failed on a fresh checkout with `Cannot find module './generated/validation-keyword-data.ts'` — this is the core i18n codegen artifact, not a PR defect. Running the standard `node packages/shared/scripts/generate-keywords.mjs` (the `prebuild`/`typecheck` step) makes all 16 pass.

### Workflow / boundary audit
`deploy-eliza-provisioning-worker.yml` validates `target_container_sha256` as empty-or-`^[0-9a-f]{64}$`, rejects it outside `diagnose` mode, and re-validates it on the remote before `base64 --decode | node - "$EXPECTED_WORKER_SHA" "$TARGET_CONTAINER_SHA256"` under `env -i` + `flock -s`. `collectDiagnostic`/`summarizeJournal` both throw `target_digest_invalid` on a malformed digest before any privileged read (verified by the added test). Closed-schema/no-raw-text invariants are unchanged from the frame parser attentionhead verified at `c574329`; nothing raw escapes.

### Non-blocking observations
1. The anchored `TARGET_HEALTH_TIMEOUTS` (no `m` flag) is deliberately conservative: a hypothetical journald record that merged a genuine standalone timeout with other lines would be under-counted rather than mis-attributed. That is the safe direction for this security posture, but a one-line comment stating the single-line-record assumption would help future maintainers.
2. attentionhead's order-dependence note about `healthObservation` relying on the `docker-sandbox-provider.ts` section order remains latent at this head; not introduced or worsened here, worth a follow-up test as they suggested.

No correctness, security, test, or required-evidence gap found. Evidence rows are appropriately `N/A` for a non-UI backend/script change, with focused-test domain artifacts supplied. Approving.

---
— [eliza-swarm]

---

PR #30777, opened 2026-09-06
fix(cli): preserve scaffold executables and reject unreadable migration sources

lalalune:
# Relates to

Closes #30753. Closes #30754.

- [x] Targets `develop`; validated base `b676280384b15df54204276e34bb8740db26936f`.
- [x] Installation and post-sync `bun run verify` completed.
- [x] Complete reviewed runtime and generated-file artifacts are attached below.

# Sync with develop

- [x] Clean isolated branch based on the fetched `develop` revision above.
- [x] Post-sync verification passed at `a5bd42f6f6e5a767a469f80752d98b6212ebe4c4`.

# Risks

Low. Template rendering now preserves executable bits without copying special permission bits. Upgrade repairs missing executable bits on otherwise unchanged managed files. Migration now rejects unreadable existing source paths; genuinely absent optional paths remain supported.

# Background

## What does this PR do?

Generated projects were losing the executable bit on their codesign wrapper, so PATH selected the system binary and bypassed the wrapper. Rendering now preserves executable permissions, and upgrade repairs affected existing projects even when file contents already match. Packaging and rendering share the same ignored-directory inventory.

Migration previously treated filesystem failures as absent files and could export an incomplete agent without a warning. Read and discovery failures now carry an actionable `MIGRATION_SOURCE_READ_FAILED` error with path, operation and cause. Only `ENOENT` remains optional; existing SQLite availability warnings retain their contract.

## What kind of change is this?

Bug fixes and internal consolidation.

# Documentation changes needed?

No user-facing command or archive format change. Source comments document permission and optional-file contracts.

# Testing

## Where should a reviewer start?

Inspect the generated-artifact receipt and complete installed-CLI runtime log attached below. They show wrapper dispatch before and after deliberate permission loss and repair, plus failed migration with neither requested output created.

## Detailed testing steps

At `a5bd42f6f6e5a767a469f80752d98b6212ebe4c4`: all 259 CLI tests passed; package typecheck, lint, build, packaged smoke and root `bun run verify` passed. Packaged smoke installs the actual CLI and exercises create and upgrade. Real filesystem tests cover unreadable files/directories, wrong-type paths, absent optional files, byte/hash preservation and dry-run behavior.

# Evidence Gate

- [x] N/A - no rendered UI change; the affected surface is CLI filesystem output.

- [x] N/A - no rendered UI change.

- [x] N/A - deterministic CLI output is recorded in the complete runtime log.

[...]

PR #30766, opened 2026-09-06
fix(elizaos): reject unreadable migration sources before archive creation

agent-cortex:
# Relates to

Closes #30754

Definition of Done: full standard in [`CONTRIBUTING.md`](../CONTRIBUTING.md).

- [x] This PR targets `develop` and is rebased onto the latest `origin/develop` with zero conflicts.
- [x] `bun install` was run; owning-package `test`/`typecheck`/`lint` gates pass (logs below). `bun run verify` is a repo-wide gate; the scoped, owning-package gates are attached and green.
- [x] A reviewer can confirm the change works without reading the code, from the before/after reproduction and failing-then-passing test logs below.

# Sync with develop

- [x] Rebased onto the latest `origin/develop` (`de30ecf38c`). The base merged the related fix #30777, which already introduced `MigrationSourceReadError`/`readOptional` and the absent-vs-unreadable split for `EACCES`/`ENOTDIR`/`EISDIR`. This PR was rebased onto that design and now carries **only** the remaining gap the maintainer reopened #30754 for.
- [x] Owning-package gates pass post-sync; see **Known gaps / failures** for the repo-wide `verify` note.

# Risks

Low. The change adds a single stat-based type guard inside the migration
reader's optional-read helper (`packages/elizaos/src/migrate/openclaw-reader.ts`).
It rejects a non-regular-file persona source (FIFO / socket / device /
directory) with the existing typed `MigrationSourceReadError` **before** handing
the path to `readFileSync`. Genuinely absent optional files and readable regular
files (including symlinks to regular files) behave exactly as before, so partial
homes still migrate. No public archive/CLI surface, dependency, or command flow
was restructured.

# Background

## What does this PR do?

Issue #30754 was reopened for one confirmed remaining case after #30777 merged:

> Reopening for the remaining confirmed wrong-type case: a FIFO persona source
> blocks migrate-agent.

After #30777, develop's `readIfPresent` calls `readOptional(p, "read", () => fs.readFileSync(p))`.
That correctly rejects `EACCES`/`ENOTDIR`/`EISDIR`, but a **FIFO** persona
source (e.g. `SOUL.md` created by `mkfifo`) passes `readFileSync`, whose
synchronous open blocks forever waiting for a writer that never arrives —
**hanging `migrate-agent` indefinitely** with no error and no archive.

This PR adds a narrow regular-file guard so `readIfPresent`:

- stats the path first (`statIfPresent`), tolerating a genuinely absent
[...]

PR #30807, opened 2026-09-06
fix(cli): reject non-regular migration source files

### LINKED ISSUE #30754: fix(elizaos): reject unreadable migration sources before archive creation
## Problem and scope

OpenClaw optional filesystem helpers catch all read errors as missing data. A real unreadable SOUL.md plus readable MEMORY.md produces a normal migration source with missing persona and no warnings. Distinguish absent optional paths from unreadable or wrong-type existing paths and preserve causes.

## Acceptance and evidence

Real filesystem EACCES and wrong-type cases fail before CLI output artifacts; missing optional files remain supported. Preserve explicit SQLite compatibility warnings. Run owning package and root gates.

Audited base: b676280384b15df54204276e34bb8740db26936f. This is a confirmed second-pass cleanup finding with independent plan review. No production deployment, credential changes, or funded transactions are part of this scope.

lalalune:
A FIFO at an optional OpenClaw source path such as `SOUL.md` blocks `migrate-agent` indefinitely. Require a regular file before reading it and use the existing `MigrationSourceReadError`; absent optional files and regular-file symlinks keep their existing behavior. This completes the remaining wrong-type case from #30754 after #30777.

Closes #30754. Related: #30766.

Validation at `f4493023eb9fa9dde56809f221af12c35751652c`, based on `65a4f5216421ba74a75b1bc9bc98de0ae2d1ceb1`:

- The baseline FIFO probe timed out. The fixed packaged CLI exits 1 with `MIGRATION_SOURCE_READ_FAILED` and writes neither character nor archive; the regular-file control exits 0 and writes both.
- 260 package tests across 23 files passed with two workers; package typecheck, lint and build passed.
- `bun install` completed with pinned Bun 1.3.14 and Node 24.15.0. `bun run verify` completed all 376 tasks and final repository audits; the final audit reported no focused tests or orphaned skips.
- Independent review confirmed the change preserves normal source reads and missing optional paths. UI, native-device and live-model evidence are not applicable to this local filesystem boundary.

### DIFF
--- a/packages/elizaos/src/migrate/openclaw-reader.ts
+++ b/packages/elizaos/src/migrate/openclaw-reader.ts
@@ -174,6 +174,16 @@ function readOptional<T>(
 }
 
 function readIfPresent(p: string): string | undefined {
+  const stat = statIfPresent(p);
+  if (stat === undefined) return undefined;
+  if (!stat.isFile()) {
+    throw new MigrationSourceReadError(
+      { path: p, operation: "read" },
+      new Error(
+        "Expected a regular file; refusing to read a special filesystem entry.",
+      ),
+    );
+  }
   return readOptional(p, "read", () =>
     normalizeEol(fs.readFileSync(p, "utf8")),
   );
--- a/packages/elizaos/src/migrate/source-read-failures.test.ts
+++ b/packages/elizaos/src/migrate/source-read-failures.test.ts
@@ -103,3 +103,41 @@ it("still builds a partial home when optional files are absent", () => {
     readOcAgentHome(path.join(root, "absent"), "demo").soul,
   ).toBeUndefined();
 });
+
+it.skipIf(process.platform === "win32")(
+  "rejects a FIFO persona without blocking or writing migration artifacts",
+  () => {
+    const root = home();
+    const soul = path.join(root, "SOUL.md");
+    const fifo = spawnSync("mkfifo", [soul], { encoding: "utf8" });
+    expect(fifo.error).toBeUndefined();
+    expect(fifo.status).toBe(0);
+    const character = path.join(root, "character.json");
+    const archive = path.join(root, "export.eliza-agent");
+    const result = spawnSync(
+      "bun",
+      [
+        fileURLToPath(new URL("../cli.ts", import.meta.url)),
+        "migrate-agent",
+        "--from",
+        root,
+        "--agent-id",
+        "demo",
+        "--emit-character",
+        character,
+        "--out",
+        archive,
+        "--password",
+        "synthetic-test-password",
+      ],
+      { encoding: "utf8", timeout: 5000 },
+    );
+    expect(result.error).toBeUndefined();
+    expect(result.status).not.toBe(0);
+    expect(result.stderr).toContain("MIGRATION_SOURCE_READ_FAILED");
+    expect(result.stderr).toContain(soul);
+    expect(fs.existsSync(character)).toBe(false);
+    expect(fs.existsSync(archive)).toBe(false);
+  },
+  10000,
+);


ss251 (approved):
Reviewed head f4493023eb9fa9dde56809f221af12c35751652c.

The risky part is `packages/elizaos/src/migrate/openclaw-reader.ts` in `readIfPresent`, where the new `statIfPresent(p)` guard throws `MigrationSourceReadError` when `!stat.isFile()` before reaching `fs.readFileSync`. I traced the FIFO input the new test targets: `SOUL.md` as a FIFO would previously block inside `readFileSync` with no timeout; now it fails fast with operation `read` and path `p`, surfacing as `MIGRATION_SOURCE_READ_FAILED` without writing `character.json` or `export.eliza-agent`. The `stat === undefined` early return preserves the existing absent-optional-file path, so only present-but-special entries throw. Non-blocking: `statIfPresent` follow-vs-lstat semantics for symlinked specials is not visible in this hunk.

Verified by reading the diff alone at this head; no suite run.

Recommendation: merge

— [muse-lane-ss251]

ss251 (approved):
Reviewed head f4493023eb9fa9dde56809f221af12c35751652c.

The risky part is `packages/elizaos/src/migrate/openclaw-reader.ts:177` in `readIfPresent`, where the new `statIfPresent` guard throws `MigrationSourceReadError` unless `stat.isFile()` before `readOptional` reaches `fs.readFileSync`. I followed the FIFO boundary the test targets: without the guard, `readFileSync` on a `SOUL.md` FIFO blocks for a writer, so the `migrate-agent` spawn would hang until the 5s `spawnSync` timeout; with the guard it fails fast with `operation: "read"`, surfaced as `MIGRATION_SOURCE_READ_FAILED` containing the soul path. The absent-file branch is preserved (`stat === undefined` returns `undefined`), and the no-artifact assertions match fail-before-write. Non-blocking: stat-then-read leaves a TOCTOU window.

Recommendation: merge

— [muse-lane-ss251]

attentionhead (approved):
## APPROVE — the guard is class-complete for the module, and `statSync` correctly keeps symlinked personas working

Reviewed at head `f4493023eb9fa9dde56809f221af12c35751652c`, 2 files, +48/-0. On the live run 34025457154, Billing payment replay E2E and Subscription authority PostgreSQL pass; source-smoke and the Windows contract were still pending when I finished. Nothing failing.

### Verified

**The guard covers every read in the module.** Sweeping `packages/elizaos/src/migrate` for `readFileSync`/`createReadStream` outside tests returns exactly one production site — `openclaw-reader.ts:188` — and it is inside `readIfPresent`, the function this PR guards. Every persona and memory read (`:250`, `:258`, `:262`, `:401`, `:489`, `:490`, `:513`, `:514`, `:515`) routes through it. So there is no second unguarded path where a FIFO could still block the CLI.

**`statSync` is the right call, and I confirmed it rather than assuming.** `statIfPresent` uses `fs.statSync`, which follows symlinks, so a `SOUL.md` that is a symlink to a regular file still satisfies `isFile()` — while a symlink pointing *at* a FIFO is correctly rejected. That is exactly the behaviour you want here, and it is easy to get wrong in the hardening direction.

**The new test genuinely runs.** Under `--reporter=verbose`, `rejects a FIFO persona without blocking or writing migration artifacts` executes in 31ms rather than being skipped — worth checking, since `it.skipIf` plus a spawned CLI is a combination that can silently no-op. The test's shape is good: `expect(result.error).toBeUndefined()` is what catches a *hang* (a blocked read would surface as the 5s `spawnSync` timeout), and asserting `MIGRATION_SOURCE_READ_FAILED`, the offending path, and the absence of both `character.json` and the archive means it pins "fails fast **and** writes nothing", not just a non-zero exit.

**Mutation testing:**

| mutant | result |
|---|---|
| P1 — remove the non-regular-file guard | **killed** (1 failed) |
| P2 — narrow the guard to reject FIFOs only | survives |
| P3 — `statIfPresent` uses `lstatSync` instead of `statSync` | survives |

**I chased P3, and it is load-bearing.** A probe reading a `SOUL.md` that is a symlink to a regular file:

```
head        -> 1 passed   (symlinked persona body is read)
P3 (lstat)  -> 1 failed   (the symlink is rejected as a non-regular file)
restored    -> 1 passed
```

So the follow-symlinks semantics are a real behavioural requirement with nothing pinning them. This matters more than a typical uncovered line: `lstat` is precisely the change a future reviewer would make while "hardening against symlink traversal", and it would silently break every user whose persona files are symlinks into a dotfiles repo — a common setup for exactly this kind of config. One test with a symlinked `SOUL.md` would prevent that.

### Not verified

- Only the FIFO case is exercised, on darwin. I did not test sockets, character devices, or block devices, nor the Windows-skipped path.
- I did not run the full `packages/elizaos` suite, only `source-read-failures.test.ts` and my probe.

### Non-blocking observations

1. **Add a symlinked-persona test** (the P3 probe above). It is four lines and it protects a real user-facing behaviour that the current suite would let anyone remove.

2. **Only FIFOs are covered among special files.** P2 — narrowing the guard to `!isFile() && isFIFO()` — survives, because a FIFO is the only non-regular entry any test creates. The shipped guard is broader and correct (`!stat.isFile()` rejects sockets and device nodes too); it is just that its breadth is unasserted. Lower value than the symlink case, since the failure mode there is a spurious rejection rather than a hang.

3. **The error message is good.** *"Expected a regular file; refusing to read a special filesystem entry."* names the rule rather than the symptom, and the `MigrationSourceReadError` carries the path — which is why the test can assert on it. Worth keeping that pairing if this pattern spreads.

Approving. Small, correct, and it closes a genuine hang: without the guard the CLI blocks indefinitely on a FIFO rather than failing, which is the worst failure mode for a migration tool.

— [claude-code-eliza]

Issue #30788, opened 2026-09-06
fix(cloud): reconcile stale Drizzle generation snapshots before production migrations


lalalune:
## Problem

The cloud migration-generation snapshot does not describe the current exported schema. At lifecycle checkpoint f103e382a2aada852ee7c5b41669ffbd06349e63, the latest snapshot contains 128 tables while drizzle-kit's `generateDrizzleJson` over `src/db/schemas/index.ts` produces 256 tables: 142 current tables are absent and 14 snapshot tables are no longer exported. The lifecycle snapshots inherit this older baseline; this is not limited to the new retention column.

Examples include absent `agent_sandboxes`, `agent_compute_stop_intents`, `agent_backup_objects` and `tenant_db_clusters`, with older `eliza_sandboxes` and removed app/session tables retained. `bun run db:check-migrations` exits 0 and prints “Everything's fine”; it checks snapshot consistency, not correspondence to migrated database objects. A passing result must not be used as deployment/schema-parity evidence.

The affected consumer is the next maintainer running `db:generate`: a stale baseline can propose unrelated creates, drops or rename prompts alongside a narrow migration. No such generated SQL was applied, and no production schema discrepancy is claimed from this source-only audit.

## Acceptance

- Reconcile the latest generation baseline with the canonical migration history and exported schemas using a disposable PostgreSQL database. Account explicitly for renamed, dropped, intentionally SQL-only and intentionally unexported objects.
- Preserve applied SQL and ledger hashes. Do not emit or apply an omnibus recreate/drop migration to repair generation metadata.
- Demonstrate that a representative narrow schema change generates only its intended SQL, and that a no-change generation produces no unrelated operations.
- Make the migration workflow distinguish structural journal checks from actual schema parity. A stale baseline must produce actionable failure before generated SQL is mistaken for deployment-ready output.
- Exercise a real migration/restore boundary in disposable PostgreSQL and record the resulting schema evidence. Inspect existing deployed migration ledgers before any live correction.

## Evidence and scope

Read-only schema generation and table-key comparison; no connection to a live database. Current cloud `db:check-migrations` exit 0 reproduced with pinned Bun 1.3.14. No generated migration, live migration, deletion or deployment performed. This is a production-plan migration-readiness dependency, separate from proving retained-container recovery.

## Fresh SQL replay follow-up

[...]

Issue #30746, opened 2026-09-06
Stop unpaid compute safely when backup capture fails, retaining local recovery state


lalalune:
## Scope

Payment suspension currently returns without stopping compute when `prepareSuspendBackupGate` cannot establish a backup. Bypassing that refusal is unsafe: Docker `stopForReplacement` removes the container and releases capacity, and node retirement excludes stopped/error agent rows. The sole local state copy can consequently lose both container and host protection.

Implement state-preserving payment suspension using the existing lifecycle jobs, stop intents and backup authority. This is independent of owner inactivity and does not authorize retention expiry.

## Acceptance Criteria

- Unfunded agents with failed checkpoint capture stop executing without container removal, mount deletion, secret loss, or automatic host retirement.
- Persist exact node/container/state retention authority before provider mutation; lost responses and worker crashes retain an actionable reconciliation record.
- Disable automatic container restart and confirm the exact immutable container is stopped; missing/unreachable resources remain explicit unresolved states.
- Recheck primary funding and lifecycle ownership before effects; manual work, deletion and newer generations win.
- Retried checkpoint capture runs without unrestricted agent side effects; retirement and orphan cleanup preserve the retained resource until verified state transfer or authorized deletion.
- Renewed funding resumes the same logical agent and retained state exactly once, including after host/worker failure; do not fresh boot around a missing backup.
- Test payment/backup/provider/lease races through real service and database boundaries; live isolated provider drills verify process state, mounts, restart policy, retained node and state continuity before rollout.

## Evidence Plan

No UI or model behavior changes: screenshots, video and live model trajectories are N/A. Required evidence: owning tests/typecheck/lint, root verification, migration checks, structured lifecycle logs, primary DB receipts, Docker readback and state continuity. Local fixtures do not replace live provider evidence.

Related: #30702 and draft #30745 for funded resume; #30675 and #30714 for retirement/recovery. No production mutation or destructive cleanup is implied by this issue.

PR #30745, opened 2026-09-06
fix(cloud): resume funded agents automatically

lalalune:
# Relates to

Refs #30702. Depends on the state-preserving recovery and retirement work in #30714 (#30697, #30675); those prerequisite commits are included while that PR remains unmerged. The final commit contains the funded-resume change.

# Background

## What does this PR do?

A confirmed billing suspension now becomes an automatic recovery candidate when authoritative funding returns. The existing provisioning worker scans explicit pages, so an unpaid account cannot starve funded accounts and no browser request or particular webhook order is required. It reuses provisioning jobs and the same logical agent.

A confirmed stop revision and durable resume-job grant bind admission to the original payment suspension. Queueing, execution, provider-create intent and restored endpoint publication recheck primary-database authority. Later manual lifecycle work revokes the automatic grant. Expired workers cannot publish replacements, write stale errors or remove a successor's container; late provider receipts may only enrich their existing exact cleanup identity. Restoration must finish before publication. Funding loss during restore cleans up the exact replacement while retaining its backup; lease/grant loss retains the cleanup receipt for its current owner.

Failed, quiesced recovery jobs remain discoverable. A durable two-minute delay after terminal failure prevents fresh jobs from resetting retry backoff. First funded admission is immediate. The additive migration intentionally leaves legacy receipts without an inferred confirmed revision.

## What kind of change is this?

Lifecycle and billing bug fix with an additive database migration. No owner-inactivity suspension is introduced.

# Sync with develop

- Rebased without conflicts onto `8fea4f94c29c7712ce3caf303e3350150622d048`.
- Head: `a894c5d1daaa8b440f7f23bb72b367777b47d080`.
- Pinned Bun 1.3.14 / Node 24.15.0 installation passed after rebase.
- Post-sync `bun run verify`: passed (exit 0), including workspace checks and repository audits.

# Risks

Large operational impact: provisioning, billing admission and database migration. Most added lines are Drizzle's generated full-schema snapshot; handwritten changes reuse the current lifecycle/queue rather than introducing another scheduler. Production rollout remains blocked on live restore/cleanup receipts and verified treatment of legacy stopped agents. Do not infer current provider ownership from an old receipt or backfill it solely from a stopped database row.

# Documentation changes needed?

Production runbook must include migration-before-worker deployment, current-provider readback for legacy suspension authority, recovery rollback and fault-drill evidence. Those operational artifacts are still pending; this PR remains draft.

# Testing

## Where should a reviewer start?

[...]

PR #30714, opened 2026-09-05
fix(cloud): retain cleanup authority and require state-preserving recovery

lalalune:
Failed provider deletion must remain retryable after a node is disabled, and replacement creation must retain enough identity to clean up an uncertain outcome. This change persists retirement intent, verifies provider absence before ownership removal, and retains exact failed-create cleanup authority. Recovery also fails closed when prior state cannot be restored and publishes readiness only after restoration succeeds.

Refs #30675 and #30697. This PR covers lifecycle cleanup and state-preserving recovery. Funded automatic resume, backup activation, environment separation and live fleet cleanup remain separate work.

# Relates to

#30675 and #30697. Targets `develop`.

# Sync with develop

Rebased without conflicts onto `e8cfcd6914e`; pinned install and root verification both passed afterward.

# Risks

Medium/high operational impact: affects provider retirement, failed-create cleanup, and recovery publication. Retains unresolved resource ownership and prior state; restore failures now remain visible instead of silently fresh-booting. Requires a staging failure/restore drill before deployment.

# Background

## What does this PR do?

Persists cleanup/retirement authority and requires prior state restoration before recovery publishes a usable agent.

## What kind of change is this?

Bug fixes to existing cloud lifecycle behavior.

# Documentation changes needed?

Deployment runbooks and live activation belong to the broader cloud production plan; this draft does not claim them complete.

# Testing

## Where should a reviewer start?

Start with node retirement retry, exact failed-create cleanup, and ordinary recovery failure/restore tests. Review the evidence limits below before approving deployment.

## Detailed testing steps

Run shared package tests, control-plane tests, daemon focused tests, shared typecheck/lint, and root verification. On staging, exercise failed provider deletion, an uncertain create acknowledgement, and a stateful restore failure followed by successful recovery; inspect both database receipts and provider resources.

## Validation

- Candidate: `9e8479d5f1496d260f0e8002a956b123100d3f56`; rebased base: `e8cfcd6914edfc1675450ce0a1bbea685e0a08be`.
- Pinned Bun 1.3.14 / Node 24.15.0 install passed after rebase.
- All 26 changed files are byte-identical to the cloud changes validated at `20b05250d7e`; the three upstream commits affect installation tooling and screen recording.
[...]

PR #30815, opened 2026-09-06
fix(cloud): preserve state and resource authority across lifecycle retries

lalalune:
This change preserves recoverable agent state and exact provider ownership across payment suspension, funded resume, replacement and deletion retries. A failed provider acknowledgement stays retryable. Cleanup uses the captured container, physical host and lifecycle generation, and cannot release a replacement host’s capacity or cancel deletion after teardown admission. Restored apps retain their configuration and credentials atomically, with outbound automation disabled.

The additional cleanup work retains a durable resource manifest covering compute, volumes, VPN and secrets. Ordinary replacement attempts now retain their placement and capacity authority before remote effects. A terminal or retried image job admits abandoned provider-success candidates into cleanup before releasing its execution binding. The existing daemon reconciles those records without treating lease expiry or missing acknowledgement as proof of absence. Recoverable volumes remain retained pending verified state recovery; compute absence alone does not authorize disk reclamation.

The exact restore provider additionally forces a quarantine marker into the container environment. In newly built images, the canonical shell entrypoint exits before normal startup when marked. A Linux/Docker counterfactual proved that the old script dispatched and wrote fixture state while the new one refused and left the candidate untouched. Historical images may lack this guard; image compatibility admission is still required. The provider-to-shell test and owning entrypoint tests passed.

The agent component work materializes a staged PGlite backup into an isolated candidate through the existing protected filesystem boundary. It checks the complete compressed digest, record order, archive paths and entry types; installs private files and directories; and persists/readbacks a replay-bound finish receipt. This does not open the production runtime, seal the cloud candidate, publish routing, or produce an activation MAC.

Tracks #23098, #22916, #20726 and #30746. This branch includes the earlier lifecycle work associated with the now-closed, unmerged prerequisite PRs #30714, #30745 and #30790. A range-diff against the previous published head e403b1b260be0351b3910d7a6ff49a3da68a1325 confirms all 40 earlier patches are unchanged after rebasing; nine commits extend that work. Closed prerequisite status is not represented as evidence of merging.

Validation checkpoints (publication head 63f37ce06bfd0fbe846a90314aba0c08beeaccf1):

- Pinned dependency installation and root verification passed after the rebase at 88588b0840c2f2e159c95d0d29441c17032cab9a. Root verification and cloud-shared/app-core typechecks and read-only lints passed again for the subsequent boot guard. Final publication sync rebased all 49 unchanged patches onto develop c2d55bf5e460055c19fb0c8bdfe217e4f1b2c8d8; fresh pinned installation and root verification passed again at publication head 63f37ce06bfd0fbe846a90314aba0c08beeaccf1.
- Final publication-head focused checks passed sequentially: 41 agent restore tests, 126 cloud job/replacement/provider tests, and 6 entrypoint tests (12 existing Linux-specific skips on macOS).
- Focused checks at the earlier 88588b0840c revision: 41 agent tests in two files and 105 cloud tests in three files passed, with no failures. Agent typecheck and read-only lint also passed before the rebase; the seven agent implementation/test files were byte-identical across it.
- Full cloud-shared suite before the rebase and subsequent upstream changes: 12,895 passed, 0 failed, 209 skipped across 357 batches. This is explicitly earlier-source evidence, not an exact-head full-suite result.
- Native Linux/Node 24.15.0/PGlite 0.4.6 drill passed against the committed component code: installed 972 files and 26 directories; rejected incorrect digest without a finish receipt; verified valid retry, exact replay and execution-token rejection; recovered a real SQL row and preserved a new write after close/reopen. Session/manifest/HMAC authority in that drill are fixtures. It does not prove live KMS, cloud database authority, production backup authentication or activation.
- Full agent package suite at 88588b0840c2f2e159c95d0d29441c17032cab9a (agent source unchanged by the subsequent boot guard): all 762 test files passed in 1,842.2 seconds, process exit 0. The worktree remained clean at the tested revision.
- Earlier owned Linux/Docker drills verified retained stop/start, restart, exact-identity deletion, hostile Docker environment rejection and host-lock serialization. No current production fleet or provider ownership claim follows from them.
- UI evidence: N/A; no rendered UI changes in this branch’s lifecycle work.

Not ready for merge or production rollout. Remaining acceptance includes the concrete cloud staging caller and physical producer/cancellation protocol; runtime database validation and authenticated activation; atomic candidate adoption; verified recovery before retained-volume reclamation; unresolved/legacy provider authority; rolling-controller and migration metadata ownership/parity (#30788); independent live backup/PITR proof; and actual development/staging/production isolation and lifecycle exercises. Operator access for the provider/Docker/control-plane database join remains unavailable. No zombie deletion is certified.

No production resources were deleted or deployed while preparing this change.


attentionhead (approved):
**Approving at head `e403b1b260`** (`MERGEABLE`, draft=false). First substantive review here — only the Copilot bot had commented — so I want to be explicit that this covers **one slice** of a 77-file change, and say which.

Note on size: the `+57,728` headline is misleading in the PR's favour. Two generated Drizzle snapshots (`meta/0362_snapshot.json`, `meta/0363_snapshot.json`) are 24,115 lines each — 48,230 of the additions, or 84%. The reviewable diff is far smaller than the number suggests.

## Scope reviewed

The new stop-confirmation authority: migration `0362_agent_compute_stop_confirmed_revision.sql`, the `agent_compute_stop_intents` schema, and the transitions that write `provider_confirmed_lifecycle_revision`. I picked it because it is the one piece of this PR that adds a **database-enforced** invariant, and a CHECK constraint is the failure mode that turns a logic slip into an aborted transaction rather than a wrong value.

## Verified — the new CHECK genuinely rejects a partially-cleared confirmation

The constraint is:

```sql
CHECK (provider_confirmed_lifecycle_revision IS NULL OR (
  provider_confirmed_lifecycle_revision >= 0
  AND provider_confirmed_at IS NOT NULL
  AND status IN ('provider_confirmed', 'superseded')))
```

Against a real PGlite instance with that table and constraint, inserting a confirmed intent exactly as `confirmIntent` leaves it (`status='provider_confirmed'`, `provider_confirmed_at=now()`, revision `7`), then moving it to `retry`:

```
inserted provider_confirmed row OK
re-arm to retry FAILED -> new row for relation "agent_compute_stop_intents"
violates check constraint "agent_compute_stop_intents_confirmed_revision_check"
```

The transition aborts the statement rather than storing an inconsistent row, which is the correct and intended behaviour.

## Non-blocking — two re-arm sites clear confirmation fields without clearing the revision

I chased that because two sites move an intent out of a confirmed shape and neither touches the new column:

- `provisioning-jobs.ts:7576` (`rearmRecoverableAgentComputeStopIntentOnce`) — when `intent.provider_confirmed_at` is set it writes `{ status: "retry", ... }`, deliberately keeping the confirmation.
- `provisioning-jobs.ts:2699` (the re-arm inside `beforeInsert`) — writes `status: "pending"` and `provider_confirmed_at: null`, clearing five other fields but not `provider_confirmed_lifecycle_revision`.

Either would violate the new constraint on a row whose revision is set. **I could not construct a reachable path to that state, and I want to be clear that my initial hypothesis was refuted rather than confirmed.** The revision is only ever written alongside `status='provider_confirmed'` (repo `:438`, and `:287` which requires that status), and both re-arm sites select their candidates with filters that exclude it:

```ts
inArray(agentComputeStopIntents.status, ["pending", "retry", "terminal_attention"])          // :7346
inArray(agentComputeStopIntents.status, ["pending", "dispatching", "retry", "terminal_attention"])  // :2670
```

So today the forbidden state is unreachable and these sites are safe. What makes it worth a sentence anyway is that the invariant is enforced *only* by those two SELECT filters, and the natural future change — letting a confirmed intent be re-stopped, which `:7576`'s `provider_confirmed_at` ternary already anticipates — turns a coverage gap into a transaction abort in a billing-driven stop path. Clearing `provider_confirmed_lifecycle_revision: null` in both `.set(...)` blocks costs one line each and makes the constraint self-enforcing instead of filter-dependent.

Also worth noting the constraint is correctly permissive where it must be: `confirmIntent` derives the revision from a scalar subquery over stopped sandboxes, so when no stopped row exists the column is NULL and the constraint passes trivially rather than blocking the confirmation.

## Not verified

- **The cloud PGlite test lane does not run in my checkout.** `agent-compute-stop-confirmation.pglite.test.ts` (346 new lines) fails with `Cannot find package '#errors' from packages/core/dist/node/index.node.js`. That is environmental: this PR touches **zero** files under `packages/core`, and I ran the control — a pre-existing cloud test, `compute-stop-concurrency.integration.test.ts`, fails identically on clean `develop`. My local `packages/core/dist` predates the subpath-imports manifest change. So I have not executed any of this PR's new tests, and my constraint evidence above comes from a standalone PGlite harness rather than the suite.
- The other 70-odd files: `agent-local-state-retention.ts` (+381 with a 1,136-line test), the Hetzner client error-policy changes (+197/-45), `node-autoscaler`, `docker-nodes`, `jobs.ts` (+29/-56), the serving-placement and agent-identities migrations, and the provisioning-worker image rollout. None of that is covered by this review.
- The other three migrations in this PR (0363–0366) and whether their snapshots are consistent with the schema files.
- CI is `UNSTABLE`; I did not chase which checks are outstanding.

Given a maintainer is likely to land this on the strength of one approval, the honest summary is that I verified the new database invariant behaves as written and traced the transitions that touch it — not that the other subsystems in this PR have been reviewed.

— [claude-code-eliza]

dal-mnr (changes requested):
@C:\Users\MUNIR-~1\AppData\Local\Temp\opencode\review-30815.md

hermesagent270-commits (approved):
**Scope note.** 112 files and ~66,000 additions, of which ~48,000 are the two generated migration snapshots. I reviewed the retiring-serving-resource ownership path — `eliza-sandbox.ts`'s `assertOwner` and the `0380` migration that backs it — and left the remaining ~108 files unread. I did not run the suites.

This is careful work and the two pieces I checked reinforce each other properly.

## The DB constraint binds identity, and the `IS TRUE` wrapper is what makes it real

```sql
ADD CONSTRAINT "agent_sandboxes_replacement_resource_manifest_check" CHECK ((
  replacement_cleanup_resource_manifest IS NULL OR (
    …
    AND replacement_cleanup_resource_manifest->>'agentId' = id::text
    AND replacement_cleanup_resource_manifest->>'organizationId' = organization_id::text
    …
    AND …->'servingPlacement'->'locator'->>'nodeId' = replacement_cleanup_node_id
    AND …->'servingPlacement'->'locator'->>'containerName' = replacement_cleanup_container_name
  )
) IS TRUE);
```

Every identifier the manifest carries is checked against the row's own columns, so a manifest cannot claim to be about a different agent, org, or placement than the row it sits on. That's the property that makes the runtime check meaningful rather than self-referential.

**The `IS TRUE` is the detail worth confirming.** A Postgres `CHECK` passes when its expression is TRUE *or NULL*, so a single NULL column — say `replacement_cleanup_node_id` — would make the comparison NULL and the entire constraint pass vacuously. Wrapping the whole expression in `IS TRUE` collapses NULL to false and rejects the row. Without it this constraint would be decorative in exactly the cases you most want it, and it's the kind of three-valued-logic detail that is normally wrong.

## The runtime assertion is lock-then-reread, not check-then-act

```ts
await this.lockLifecycle(tx, agentId, orgId);
const current = await this.getAgentForLifecycleMutation(tx, agentId, orgId);
… SELECT id FROM agent_sandboxes WHERE id=… AND organization_id=…
  AND lifecycle_revision=${revision}
  AND deletion_attempt_id IS NULL
  AND replacement_cleanup_resource_manifest = ${JSON.stringify(manifest)}::jsonb
```

The lock is taken first, inside the transaction, and the revision and manifest are re-verified against the locked row — so a retry that lost authority between planning and acting fails rather than proceeding on stale belief. `owned.rows.length !== 1` and the typed `AGENT_RETIRING_RESOURCE_AUTHORITY_CHANGED` keep it fail-closed.

I confirmed the column is `jsonb`, not `json`. That matters more than it looks: `json` has no equality operator in Postgres at all, so this comparison would not work against a `json` column, and `jsonb` additionally normalises key order and numeric representation — making the equality semantic rather than textual, which is what you want when the manifest round-trips through `JSON.stringify`.

The guard that the retiring locator must *not* equal the current `node_id`/`container_name` is also the right way round — it refuses to retire the placement that is presently serving.

## One asymmetry: identity is DB-enforced, the payload is not

The constraint validates every identifier and pins `deletionPolicy.kind`, `version`, and a UUID-shaped `deletionAttemptId`. But `resources` is checked only for its type:

```sql
AND jsonb_typeof(replacement_cleanup_resource_manifest->'resources') = 'object'
```

So a well-formed manifest whose `resources` object doesn't correspond to what was actually provisioned is accepted at write time. The runtime jsonb equality in `assertOwner` catches *drift* — it detects that the manifest changed since it was written — but it can't detect that the manifest was wrong when written, because it compares the manifest to itself.

That may well be deliberate: `resources` is presumably open-ended and the authority question is "does this row still own this cleanup", not "is the resource list accurate". Worth a line in the migration saying so, since every neighbouring field is pinned and this one conspicuously isn't — a later reader will wonder whether it was an oversight.

Approving on the ownership path.

```slop-review
{"schemaVersion":"2","projectId":"eliza","artifactUrl":"https://github.com/elizaOS/eliza/pull/30815","headSha":"63f37ce06bfd0fbe846a90314aba0c08beeaccf1","provider":"anthropic","model":"claude-opus-5[1m]","client":"Claude Code","runId":"","traceSha256":"","recommendation":"accept","reproduced":false,"securityRisk":"none","duplicateRisk":"none","splitRisk":"medium","effortBand":"large","complexity":"complex","impact":"significant","reviewLoad":"deep","recommendedTier":"large","recommendedThirds":24,"workUnitId":"wu_eliza_30815_lifecycle_resource_authority","confidenceBasisPoints":8400,"valueRationale":"Partial review - the retiring-serving-resource ownership path only (assertOwner plus migration 0380); ~108 of 112 files unread and suites not run, stated plainly. Confirmed the CHECK constraint binds every manifest identifier to the row's own columns, and that the trailing `IS TRUE` is load-bearing: a Postgres CHECK passes on NULL, so without it a single NULL column would make the whole constraint pass vacuously. Verified the column is jsonb rather than json - json has no equality operator in Postgres, so the runtime `manifest = ...::jsonb` comparison would not work at all against a json column, and jsonb normalises key order so the equality survives JSON.stringify round-tripping. The runtime assertion is lock-then-reread inside the transaction with lifecycle_revision and manifest re-verified, and refuses to retire the currently-serving placement. One asymmetry: every identifier is DB-enforced but `resources` is only jsonb_typeof-checked, so a well-formed but inaccurate resource list is accepted at write time and the runtime equality cannot detect it, since it compares the manifest to itself.","usefulArtifacts":["CHECK binds manifest agentId/organizationId/sandboxId/nodeId/containerName to the row's own columns","trailing `IS TRUE` collapses NULL to false - without it a NULL column would make the constraint pass vacuously","column is jsonb (migration 0380), so equality works and normalises key order; a json column would have no equality operator","assertOwner calls lockLifecycle first, then re-reads and verifies lifecycle_revision, deletion_attempt_id IS NULL, and manifest equality inside the same tx","assertOwner throws if the retiring locator equals the current node_id/container_name - refuses to retire the live placement","resources is validated only by jsonb_typeof = 'object'; no binding to provisioned state"],"commands":["gh api /repos/elizaOS/eliza/pulls/30815/files --paginate --jq '.[] | \"\\(.additions + .deletions)\\t\\(.filename)\"' | sort -rn","git show pr30815:packages/cloud/shared/src/lib/services/eliza-sandbox.ts | grep -n 'const assertOwner = async' -A 22","git show pr30815:packages/cloud/shared/src/db/migrations/0380_agent_retiring_serving_placement.sql"],"evidenceUrls":["https://github.com/elizaOS/eliza/pull/30815"],"summary":"Approve on the ownership path (partial review, ~108 of 112 files unread). The CHECK constraint binds every manifest identifier to its row and the trailing IS TRUE is what stops NULLs passing it vacuously; the column is correctly jsonb since json has no equality operator. Flagged that resources is type-checked but unbound while every neighbouring field is pinned."}
```

lalalune:
Closing after independent review and a real PGlite service/repository reproduction. The replacement design is incomplete in an enabled path: after A→B, retirement persists the old compute/secret/VPN absence but unconditionally raises `AGENT_RETIRING_VOLUME_RECOVERY_PENDING`. The manifest and cleanup fence remain bound to the agent, with no implemented verified-recovery transition to release them. The next actual `executeUpgrade` call cannot proceed to C.

Observed result:
```json
{"success":false,"rolledBack":true,"cleanupPending":true,"error":"Replacement cleanup is still pending: Retired serving volume remains owned pending recovery verification"}
```

The published implementation is at [the retained-volume refusal](https://github.com/elizaOS/eliza/blob/63f37ce06bfd0fbe846a90314aba0c08beeaccf1/packages/cloud/shared/src/lib/services/eliza-sandbox.ts#L14087). Removing the fence alone would discard retained-volume/capacity ownership before recovery has been established. Upgrade snapshots the old runtime but does not restore and verify that state in the replacement before cutover. A safe completion needs authenticated restore, verified candidate adoption, and a defined retention/reclamation transition; the PR body also acknowledges these unfinished contracts. Removing the dormant materializer/quarantine additions during review did not fix this enabled-path blocker.

The review reconciled the changes into current split lifecycle owners and repaired a funded-resume quota bypass. Those changes are preserved locally at `10c704ddb95c948df97a51ea96a4000b696730f4` (not pushed or merge-ready), with the original published history retained. Validation included 38 funded-resume PGlite cases, 59 split-owner regressions, native PostgreSQL lock/concurrency checks, and successful replay of all 380 migrations into a fresh database. Cloud typechecks and the control-plane suite passed. Full cloud/root verification and real Linux/provider restore/adoption acceptance remain incomplete; focused green results do not resolve the failure above.

No production resources were changed. This PR is being closed without merging because its state-preservation promise cannot be completed by a safe narrow repair. A replacement design must prove A→B→C with preserved state and explicit old-volume authority before it can be accepted.

Issue #29921, opened 2026-08-29
[Epic][Launch][Routing] Certify same-bot Personal Shared-to-Dedicated routing


standujar:
Parent launch epic: #25020
Messaging analysis parent: #25076

## Goal

The same external provider identity and conversation route a linked user to Personal Shared by default and to their active Dedicated runtime when explicitly selected, without bot replacement, conversation loss, silent activation, or tenant leakage.

## In scope

Routing decisions; quote/consent; active, unavailable, starting, sleeping, revoked, and recovery states; DM and owner-bound group continuity; retry/fallback truth; provision/health/chat/sleep/wake/restore/chat; same conversation and provider identity.

## Non-goals

Provider account setup; direct/group connector semantics; broad provisioning redesign; pricing redesign; production promotion.

## Dependencies and safety gate

Front-door, #25025, Shared 1:1, and group Goals; #21594, #25044, #22934, #20615, #28500, #29622, #29653, #27096, and #25076. A composed live transition is forbidden until the explicit safety GO required by #25076 is recorded.

## Acceptance criteria

- One renewable Dedicated staging fixture completes provision → health → chat → sleep → wake/restore → chat.
- Applicable provider conversations preserve bot identity, channel binding, conversation, and owner scope across routing changes.
- No implicit lifecycle request occurs before explicit quote/consent.
- Unavailable/starting/failure/fallback states are truthful and actionable.
- Staging, tenant, private-memory, and production isolation pass.

## Evidence

Provider receipt; Cloudflare router events; gateway logs; attested DB binding/lifecycle state; control-plane jobs; sandbox source SHA, heartbeat, runtime, sleep/wake, restore, and cleanup logs.

## Execution contract

This is a bounded execution epic under #25020. Work validation-first: inspect the live staging product with Computer Use on desktop and mobile where applicable, then inspect the exact current code path before drawing conclusions. Use ui-ux-pro-max for customer-facing UX. Pin evidence to the served source/deployment SHA and correlate only applicable layers: browser/network, Cloudflare, gateway/Railway, PostgreSQL, control plane, sandbox/runtime, and provider receipts. Source, mocks, health endpoints, or a gateway 2xx do not prove live acceptance.

[...]

Issue #24594, opened 2026-08-22
Sandboxes V3: automatic Robot/Cloud node-loss failover without autoscale


standujar:
Parent: #20726

## Dependency update — 2026-09-06

Next delivery priority is #20732: complete explicit manifest-v3 restore, generation commit/restart, functional proof and routing. Automatic node-loss recovery consumes that path and is the following recovery slice. The foundation PRs #29951, #29957–#29959 and #29990–#29993, plus #30443, are merged; this is not evidence that automatic failover is complete. The inspected coordinator still explicitly rejects the automatic-failover flag as unimplemented.

## Problem

The restore V3 data plane is being implemented, but automatic Robot/Cloud node-loss recovery still has no durable control-plane authority. Current health failures are process-local, the failover flag deliberately throws as unimplemented, several allocation paths can race a cordon, and a returning old container can still create split-brain risk.

This issue tracks the bounded node-loss failover work. It does **not** authorize deployment, production mutation, autoscaling, or enabling any gate.

## Invariants

- Confirm loss from durable independent signals bound to the exact node record and incarnation.
- Fence routing and cordon allocation atomically before any recovery work.
- Never fresh-boot an empty agent when a compatible verified backup is unavailable.
- Use existing Robot/Cloud capacity only; no failover-triggered autoscale.
- Keep one durable failover member per incident and agent; retries must be idempotent.
- Run restore V3 through exact backup/copy/image/vault/readiness/probe/CAS authorities.
- Preserve shared user state while capacity is unavailable, then seal/import its delta before activation.
- A returning old placement cannot receive or emit canonical traffic; cleanup uses immutable identities.
- Feature flag remains literal 0 until all proof, fault-injection, and soak criteria pass.

## Draft PR slices

- [ ] Durable incident/member state model and lock order (no daemon wiring)
- [ ] Multi-signal Robot/Cloud detection plus atomic global allocation freeze
- [ ] Primary inventory, backup/copy preflight, replacement quota credit, existing-capacity admission
- [ ] Loss-aware adapter over restore V3 with typed copy fallback and crash-safe retries
- [ ] Explicit shared fallback, recovery-visible state, delta seal/import/receipt
- [ ] Returning-survivor repair, grant/container fencing, recount and settlement
- [ ] Daemon wiring, metrics, alerts and staging fault-injection/soak evidence

## Mandatory tests

[...]

PR #30820, opened 2026-09-06
feat(agent): materialize and seal manifest-v3 restore candidates

### LINKED ISSUE #20732: [agent] Orchestrate durable manifest-v3 restore before routing
## Project / Coordination

- Project board: https://github.com/orgs/elizaOS/projects/15
- Parent tracker or epic: #20726
- Coordination Discussion: #20726
- Suggested lane tag: /root

## Contribution Provenance

- AI assistance: yes
- AI provider/model: OpenAI / gpt-5.6-sol
- Client / agent tooling: Codex
- Contribution skill revision: elizaOS/eliza@2d89d6b0a9897457ef1a452f0c8bc76684c23a1:packages/skills/skills/contribute-to-eliza

## Scope

Add the durable control-plane coordinator that connects an exact dual-verified manifest-v3 source to capacity reservation, old-workload fencing, vault-key seeding, quarantined container creation, streaming restore, Agent generation commit, restart attestation, functional probes, activation publication and catalogue finalization. Start with explicit restore behind an off-by-default gate; automatic node-loss failover and isolated drills consume the same authority in follow-on slices.

## Current progress — 2026-09-06

This is the next active delivery slice of #20726. The former PR review backlog and #30443 dependency have merged; there is no remaining wait for those PRs.

- [x] Merge the candidate foundation stack: #29951, #29957, #29958, #29959, #29990, #29991, #29992 and #29993.
- [x] Merge the capacity-proof teardown dependency #30443.
- [x] Confirm all nine merge commits are included in inspected `develop@de30ecf38c4e64999aa90bcaf8d77b7d8be956a4`.
- [ ] Complete the functional integration and evidence required by the acceptance criteria below.

[Merge and current-source evidence](https://github.com/elizaOS/eliza/issues/20726#issuecomment-5559596714). The merged pieces provide authenticated streaming, durable candidate/record/seal authority, isolated filesystem staging and character/media/state/vault materializers. They do not yet provide a running restored generation: the quarantined-create runtime returns at `container_created`, and the stream/materializer entry points have no production callers in the inspected revision.

## Next implementation sequence

1. Complete database materialization and assemble all five authenticated components into one isolated candidate behind the concrete Agent restore protocol. Preserve cancellation, bounded streaming, durable receipts and exact replay.
2. Connect that protocol to the durable coordinator after quarantine creation. Commit and restart onto the exact generation; reconcile crashes and lost responses at each effect.
3. Verify restart identity, readiness and signed functional state. Only then commit funding/activation authority and route CAS, followed by exact old-target cleanup.
4. Prove end-to-end replay, lease loss, tamper, node-incarnation changes, capacity saturation and failed publication. Any failed restore keeps routes unavailable and must never become an empty boot.
5. Under separately authorized staging access, demonstrate backup → wipe → restore → boot → functional probe, including database/media/vault/persona/state evidence, a remembered unique fact, exactly one response and measured RTO.

The immediate reviewable deliverable is one complete explicit restore through this path. Automatic node-loss orchestration belongs to #24594 after this path works. Incremental-chain and durable multipart publication work remains tracked by #24601; existing provider primitives should be reused.

## Acceptance Criteria

- [ ] A dedicated primary-DB operation/phase machine survives process crashes and response loss at every side effect.
- [ ] The source, lease, copy role, target generation/node incarnation/container/image and job execution are exactly fenced.
- [ ] GC/catalog mutation cannot invalidate a recovery after destructive cutover.
- [ ] The target volume receives the retained vault key before Docker create; secrets never appear in argv, logs or receipts.
- [ ] The Agent seals and commits all five components, restarts onto the exact generation, and emits a signed canonical functional-probe receipt without publishing early.
- [ ] Routes remain unavailable until probe, funding and activation CAS succeed; catalogue verification references the exact probe receipt.
- [ ] Replay, lease loss, node ABA, capacity saturation, tamper, restart response loss and routing tests pass.
- [ ] Feature remains off by default until exact staging wipe/restore proof is attached.

## Evidence Plan

- [ ] Screenshots or recording: staging wipe/restore walkthrough posted before enabling; N/A for foundation commits.
- [ ] Logs: privacy-safe phase timeline and retry/fence codes.
- [ ] Real-LLM trajectories: restored QA agent recalls a unique fact and emits exactly one response.
- [ ] Domain artifacts: pseudonymized source/target authorities, hashes, receipts, DB/media/vault/persona/plugin facts and RTO.
- [ ] CI or local commands: PGlite phase/concurrency tests, real Agent HTTP/FS tests, package gates and exact staging run.

## Claim / Status Notes

No production or staging mutation is authorized by this issue. Automatic failover and drills remain fail-closed until their own evidence is complete.

— [/root]

standujar:
# Relates to

Part of #20732, under #20726. **Does not close #20732 or enable restore.**
This PR delivers the candidate-data materialization boundary: authenticated record inbox → physical database/files → five-component assembly → assembly-backed PRIMARY seal.

- [x] Targets `develop`, rebased without conflicts onto `dac71fc5a613529a42f8cd686ed333e0fc693b1c`.
- [x] Installation attempted after sync; the unrelated local Metal-toolchain failure and supported scoped install are recorded below.
- [x] Reproducible behavior and limitations are documented below.

# Sync with develop

Seven implementation commits; no historical worktree diff was imported. The rebase added only the existing migration-reader change from develop; restore source and tests are byte-identical across that rebase.

# Risks

High-impact data integrity / concurrency boundary, kept behind explicit use of the materializing factory. No applied migration, default feature gate, routing, provider-create, billing activation or deployment configuration changes. Review lock ordering, replay semantics, private filesystem identities, physical archive handling and child-process settlement particularly carefully.

# Background

## What does this PR do?

- Extract the authenticated physical PGlite archive using bounded strict USTAR handling into the isolated candidate. Reject unsupported entries and malformed/truncated archives.
- Validate a disposable private copy with the actual PGlite engine; preserve the authenticated extraction for exact replay. The validation child retains the quarantine kernel lock and is reaped before cleanup.
- Assemble character, database, media, plugin state and vault under one quarantine lock; persist exact tree/receipt proofs.
- Add the explicit materializing repository adapter. Hold PRIMARY source/operation/lease/catalogue/object/candidate authority through actual Agent effects; persist ledger/seal state only after the exact acknowledgement. Reuse the existing migration/terminal-command authorities.
- Execute these effects in private one-shot Agent processes. Raw bounded frames travel on stdin; stdin stays open as liveness. Only a canonical receipt digest crosses stdout. No credentials or preload options are inherited.
- Prove the compiled worker over native Linux Docker exec, including stdin cancellation and abrupt Docker-client loss while a real database-validation child is alive.

The existing metadata-only foundation API remains compatible; it is not silently selected as fallback by the new materializing factory. This worker is a trusted private process boundary, **not an authenticated public or production Docker endpoint**.

Additive feature; completes the data-materialization/assembly unit of explicit restore.

Updated `packages/agent/README.md` with transport version 2, liveness/cancellation ownership, native Docker test prerequisites and the distinction from production quarantine bootstrap.

# Testing

1. The real PostgreSQL candidate-execution integration test: see actual five-component materialization, transaction locks, failed acknowledgements, lost COMMIT recovery and retained output.
2. The Docker materializer test: observe a live validator PID before failure injection, then remote settlement and exact retry.
3. Archive extraction/validation and candidate assembly, then the materializing repository and private transport.

## Detailed testing steps

Pinned Bun 1.3.14 / Node 24.15.0. Run focused Agent suites from `packages/agent`:

```sh
bunx --no-install vitest run --config vitest.config.ts \
  src/services/agent-backup-restore-v3-candidate-assembly.test.ts \
  src/services/agent-backup-restore-v3-candidate-materializers.test.ts \
  src/services/agent-backup-restore-v3-candidate-fs.test.ts \
  src/services/agent-backup-restore-v3-pglite-archive.test.ts \
  src/services/agent-backup-restore-v3-candidate-records.test.ts \
  src/services/agent-backup-restore-v3-materializer-process.test.ts \
  --coverage.enabled=false
```

Native Docker proof requires the built Agent package, a local `node:24.15.0-alpine` image and a running Docker daemon. The harness resolves the image to its immutable ID; it does not pull implicitly. Set `DOCKER_CONFIG` explicitly when using a named context because Vitest isolates HOME:

```sh
AGENT_RESTORE_V3_DOCKER_TESTS=1 bunx --no-install vitest run \
  --config vitest.config.ts \
  src/services/agent-backup-restore-v3-materializer-docker.test.ts \
  --coverage.enabled=false
```

For the real PostgreSQL test, from `packages/cloud/shared`, use an owned temporary directory shared with the local Docker VM for `TMPDIR`, and select the local Docker context:

```sh
env -u APPS_TENANT_DB_TEST_DSN -u DATABASE_URL -u TEST_DATABASE_URL \
  APPS_TENANT_DB_EPHEMERAL=1 APPS_TENANT_DB_TEST_IMAGE=postgres:16-alpine \
  REQUIRE_REAL_POSTGRES_RESTORE_V3_CANDIDATE_TESTS=1 \
  bun test src/db/repositories/__tests__/agent-backup-restore-v3-candidate-execution-postgres.integration.test.ts

bun test src/db/repositories/__tests__/agent-backup-restore-v3-materialization.pglite.test.ts \
  src/db/repositories/__tests__/agent-backup-restore-v3-candidate-execution.pglite.test.ts \
  src/db/repositories/__tests__/agent-backup-restore-v3-candidate-seal-authority.pglite.test.ts
```

# Evidence Gate

- [x] N/A - backend filesystem/process/transaction work; no rendered UI changes.

- [x] N/A - backend filesystem/process/transaction work; no rendered UI changes.

- [x] N/A - no user-facing UI or device workflow in this materialization unit.

- [x] Actual process and PostgreSQL test output is pasted in Evidence Details below; plaintext-sensitive worker diagnostics are deliberately not emitted.
<details>
<summary>Observed focused results (all exit 0)</summary>

```text
Agent:
 Test Files  6 passed (6)
      Tests  128 passed (128)
 Duration 179.61s

Native Docker (compiled worker, Linux, no FD emulation):
 Test Files  1 passed (1)
      Tests  4 passed (4)
 Duration 72.42s
 Cases: exact character materialization/replay; complete-frame EOF cancellation;
 real validator cancellation on stdin EOF; real validator cancellation on client SIGKILL.
 The validator PID is observed before failure injection.
 Remote worker/validator exit and scratch-copy removal are checked before exact retry.

Real PostgreSQL:
(pass) serializes exact and divergent slot races across independent sessions
(pass) admits one cleanup claimant, recovers an expired lease atomically, and fences settlement
(pass) serializes one owner generation across multiple rows and rejects stale lost-ACK recovery
(pass) holds source and candidate locks through real Agent effects and reconciles lost commit acknowledgements
(pass) seals five real Agent components across private processes under PRIMARY locks and recovers partial assembly
 5 pass
 0 fail
 86 expect() calls
 31.71s

PGlite repository regressions:
 4 pass
 0 fail
 136 expect() calls
 6.29s
```

The focused runs preceded the rebase; a tree comparison confirms no restore source/test changes from that rebase. Final post-sync verification and compiled Docker rerun: `bun run verify` exited 0 after all package tasks and final repository audits; 11,469 tests scanned, zero focused tests or orphaned skips. The post-build Docker suite was rerun at the final head: 4/4 pass in 72.07s, exit 0. Real PostgreSQL was also rerun after sync/install/build at this exact head: 5/5 pass, 86 assertions, 30.25s, exit 0. Agent and cloud/shared typecheck/lint pass; `git diff --check` passes. Existing repository warnings remain visible; this is not a warning-free claim.

</details>

- [x] N/A - no frontend request, console or network surface changed.

- [x] N/A - no model/provider/action/prompt execution is changed or booted by this unit; the live remembered-fact Agent drill remains required by #20732.

- [x] Restored SQL rows, exact file bytes, retained assembly identity and transaction visibility are inspected by the real integration tests, detailed below.
<details>
<summary>Inspected restoration outputs and authority behavior</summary>

The real PostgreSQL fixture dumps a physical source database and deletes its test-owned source directory, then restores all five components through actual private Agent processes. It inspects independent-connection lock/visibility behavior during effects, mismatched/lost/cancelled assembly acknowledgements, lost PostgreSQL COMMIT acknowledgement and exact sealed replay after expiry. The assembly marker's bytes and inode remain unchanged on terminal replay; exact media/state/vault bytes are read back. Querying a disposable copy of the restored database returns:

```json
[{"id":1,"fact":"amber lighthouse 20732"}]
```

The native Docker test independently dumps/deletes a physical source, interrupts its real validation descendant, retries, and queries the restored output:

```json
[{"value":"amber docker lighthouse"}]
```

The following is a consolidated report of the exact query read-backs asserted by the passing real tests (not worker stdout or an LLM transcript):

```json
{
  "postgres_repository_integration": {
    "query": "SELECT id, fact FROM restored_fact ORDER BY id",
    "rows": [{"id": 1, "fact": "amber lighthouse 20732"}]
  },
  "native_docker_materialization": {
    "query": "SELECT value FROM docker_fact",
    "rows": [{"value": "amber docker lighthouse"}]
  }
}
```

These are synthetic test facts, not user data. Database queries use disposable copies so inspection cannot mutate the retained authenticated candidate.

</details>

# Evidence Details

## Real LLM-call trajectory

N/A - this PR does not execute a live Agent or model. SQL data recovery is not represented as remembered-fact conversational proof.

## Backend + frontend logs

See the complete observed output in the backend-logs evidence row above.

Frontend: N/A - backend-only unit.

## Domain artifacts inspected

See the inspected artifacts in the domain-artifacts evidence row above.

N/A - no rendered UI changes.

## Audio / voice walkthrough

N/A - no audio/voice path changed.

## Known gaps / failures

- The first native test invocation could not resolve the named Colima context inside Vitest's isolated HOME. Explicit `DOCKER_CONFIG` fixed the test environment; no filesystem authority check was weakened.
- Two initial test-only Uint8Array generic inference errors were corrected before final typecheck/tests.
- Plain `bun install` fails in the unchanged fused native-inference postinstall because this macOS host lacks the Metal Toolchain. The installer and native source are unchanged from develop. The existing `ELIZA_SKIP_FUSED_INFERENCE_SETUP=1 bun install` path completed successfully, with no lockfile change. This does not claim native inference build coverage.
- Local non-Linux tests explicitly use test-only pathname emulation; the Docker lane exercises actual Linux descriptor anchors and locks.
- Provider-stream/source authority is synthetic in these materialization integration fixtures. This is not a new full provider-crypto end-to-end proof.
- The Docker container uses a networkless idle PID1 fixture with a read-only repository mount. It is not the production image/bootstrap integration.
- The new explicit coordinator variant is integrated with actual Agent materialization, but the production caller after `container_created` is still required. Production SSH transport must retain exact container/image/node-incarnation fences and cannot infer remote settlement solely from a lost local client.
- Exact generation commit/boot, signed identity/readiness/functional receipts, funding/activation/route CAS, old-target cleanup and remaining crash/node/publication scenarios stay in #20732.
- The separately authorized staging backup → wipe → restore → boot → remembered fact → exactly one response and measured RTO drill is not done. Test durations are not production RTO.
- No activation gate enabled, no deployment, no paid resource, no merge or acceptance checkbox closure. #20726 remains NO-GO; #24594/#24601 and unrelated #29921/#30815 were not absorbed.

N/A - no exception requested. Independent review and applicable exact-head hosted checks remain required before merge.

AI-assisted implementation with Codex, lane `/root`.

### DIFF
--- a/packages/agent/README.md
+++ b/packages/agent/README.md
@@ -19,6 +19,24 @@ bun run test
 
 See `package.json` for `build`, `lint`, and other scripts.
 
+## Manifest-v3 quarantine materialization
+
+`createAgentBackupRestoreV3ProcessMaterializer` implements the durable restore
+coordinator's record, component and assembly effects in private one-shot Agent
+processes. It requires an existing isolated `CandidateFs` authority and retains
+the exact root device/inode identities. It does not boot a runtime, register
+plugins, open a listener, authorize a generation commit, or publish routes.
+
+The worker receives bounded metadata and raw record bytes on stdin; fd 3 is a
+parent-liveness channel. Only the digest of the completed canonical receipt is
+returned. Cancellation closes the liveness channel and waits for worker
+settlement, including nested database validation, before the adapter returns.
+Credentials and process preload options are not inherited. This is a trusted
+local-process transport, not an authenticated remote Docker endpoint; its
+caller must exclusively own both the transport and the quarantined roots.
+Production filesystem authority requires Linux. Non-Linux pathname emulation
+is explicitly test-only and makes no cross-process kernel-lock claim.
+
 ## Research tasks
 
 `ResearchTaskExecutor` requires a provider registered for
--- a/packages/agent/README.md
+++ b/packages/agent/README.md
@@ -27,15 +27,27 @@ processes. It requires an existing isolated `CandidateFs` authority and retains
 the exact root device/inode identities. It does not boot a runtime, register
 plugins, open a listener, authorize a generation commit, or publish routes.
 
-The worker receives bounded metadata and raw record bytes on stdin; fd 3 is a
-parent-liveness channel. Only the digest of the completed canonical receipt is
-returned. Cancellation closes the liveness channel and waits for worker
-settlement, including nested database validation, before the adapter returns.
+The worker receives bounded metadata and raw record bytes in one length-delimited
+frame on stdin (private transport version 2). Stdin must stay open after that
+frame: EOF or additional bytes cancel the operation. No inherited fd 3 is needed.
+Only the digest of the completed canonical receipt is returned. Cancellation
+closes stdin and waits for worker settlement, including nested database
+validation, before the adapter returns.
 Credentials and process preload options are not inherited. This is a trusted
 local-process transport, not an authenticated remote Docker endpoint; its
 caller must exclusively own both the transport and the quarantined roots.
 Production filesystem authority requires Linux. Non-Linux pathname emulation
 is explicitly test-only and makes no cross-process kernel-lock claim.
+
+The opt-in `agent-backup-restore-v3-materializer-docker.test.ts` suite exercises
+the built worker through real `docker exec -i` in a networkless Linux container,
+including database-validator cancellation and exact retry. Build this package
+first and preload `node:24.15.0-alpine`, then run the suite with
+`AGENT_RESTORE_V3_DOCKER_TESTS=1`. When using a named Docker context, explicitly
+set `DOCKER_CONFIG` to its configuration directory: Vitest isolates the home
+directory. The harness creates only test-owned containers and tmpfs data and
+removes them after each test; it never pulls an image implicitly. Its idle
+container is a test fixture, not the production quarantine bootstrap.
 
 ## Research tasks
 
--- a/packages/agent/src/services/agent-backup-restore-v3-candidate-assembly.test.ts
+++ b/packages/agent/src/services/agent-backup-restore-v3-candidate-assembly.test.ts
@@ -1,0 +1,354 @@
+/**
+ * Real filesystem/PGlite integration for five-component candidate assembly.
+ * Stream authority receipts are fixture inputs: this is not provider crypto,
+ * an Agent boot, a routing proof or a live-model restore drill.
+ */
+
+import { createHash, randomUUID } from "node:crypto";
+import fs from "node:fs/promises";
+import os from "node:os";
+import path from "node:path";
+import { PGlite } from "@electric-sql/pglite";
+import {
+  AGENT_BACKUP_RESTORE_V3_COMPONENT_DESCRIPTORS,
+  AGENT_BACKUP_RESTORE_V3_EXACT_READ_RECEIPT_DERIVATION,
+  AGENT_BACKUP_RESTORE_V3_SOURCE_AUTHORITY_DERIVATION,
+  AGENT_BACKUP_RESTORE_V3_STREAM_RECEIPT_FORMAT,
+  type AgentBackupRestoreV3CandidateReceipt,
+  type AgentBackupRestoreV3ComponentReceipt,
+  type AgentBackupRestoreV3StagingSession,
+} from "@elizaos/shared";
+import { afterEach, describe, expect, it } from "vitest";
+import { assembleAgentBackupRestoreV3Candidate } from "./agent-backup-restore-v3-candidate-assembly";
+import { materializeAgentBackupRestoreV3CandidateCharacter } from "./agent-backup-restore-v3-candidate-character";
+import {
+  type AgentBackupRestoreV3CandidateFs,
+  openAgentBackupRestoreV3CandidateFs,
+} from "./agent-backup-restore-v3-candidate-fs";
+import { stageAgentBackupRestoreV3CandidateRecord } from "./agent-backup-restore-v3-candidate-records";
+
+const roots = new Set<string>();
+const candidates = new Set<AgentBackupRestoreV3CandidateFs>();
+const databases = new Set<PGlite>();
+const control = () => ({
+  signal: new AbortController().signal,
+  deadlineEpochMs: Date.now() + 120_000,
+});
+const hash = (bytes: Uint8Array) =>
+  createHash("sha256").update(bytes).digest("hex");
+const encode = (text: string) => new TextEncoder().encode(text);
+const FACT = "The restored lighthouse is amber-20732";
+const CHARACTER = encode(
+  '{"name":"Restore QA","bio":["remembers a lighthouse"],"plugins":[]}',
+);
+const MEDIA = encode("private-media-20732");
+const STATE = encode('{"pluginFact":"tide-20732"}');
+const VAULT = encode("opaque-vault-ciphertext-20732");
+const platformOptions = () =>
+  process.platform === "linux"
+    ? {}
+    : { testOnlyAllowNonLinuxFdEmulation: true as const };
+
+async function fixture(realDatabase = false) {
+  const root = await fs.mkdtemp(
+    path.join(await fs.realpath(os.tmpdir()), "restore-v3-assembly-"),
+  );
+  roots.add(root);
+  await fs.chmod(root, 0o700);
+  const attemptRoot = path.join(root, "attempt");
+  await fs.mkdir(attemptRoot, { mode: 0o700 });
+  const candidateFs = await openAgentBackupRestoreV3CandidateFs({
+    trustedRoot: root,
+    attemptRoot,
+    control: control(),
+    ...platformOptions(),
+  });
+  candidates.add(candidateFs);
+  const session: AgentBackupRestoreV3StagingSession = Object.freeze({
+    restoreAttemptId: randomUUID(),
+    operationId: randomUUID(),
+    expectedManifestSha256: "a".repeat(64),
+    stagingHandle: randomUUID(),
+    cleanupHandle: randomUUID(),
+    executionToken: randomUUID(),
+    cleanupRegistered: true,
+    isolatedCandidate: true,
+  });
+  let databaseBytes = new Uint8Array([1]);
+  if (realDatabase) {
+    const sourcePath = path.join(root, "source");
+    const db = new PGlite(sourcePath);
+    databases.add(db);
+    await db.exec(
+      "CREATE TABLE assembly_fact (id integer PRIMARY KEY, fact text NOT NULL)",
+    );
+    await db.query("INSERT INTO assembly_fact VALUES ($1, $2)", [1, FACT]);
+    databaseBytes = new Uint8Array(
+      await (await db.dumpDataDir("gzip")).arrayBuffer(),
+    );
+    await db.close();
+    databases.delete(db);
+    await fs.rm(sourcePath, { recursive: true });
+  }
+  const contents = [CHARACTER, databaseBytes, MEDIA, STATE, VAULT];
+  const paths = [null, null, "photo.bin", "plugin/state.json", "vault.json"];
+  const components: AgentBackupRestoreV3ComponentReceipt[] = [];
+  try {
+    for (const [
+      componentIndex,
+      descriptor,
+    ] of AGENT_BACKUP_RESTORE_V3_COMPONENT_DESCRIPTORS.entries()) {
+      const bytes = contents[componentIndex];
+      if (!bytes) throw new Error("Missing fixture component");
+      const filePath = paths[componentIndex];
+      let dataFrameCount = 0;
+      for (
+        let offsetBytes = 0;
+        offsetBytes < bytes.length;
+        offsetBytes += 256 * 1024
+      ) {
+        const payload = Uint8Array.from(
+          bytes.subarray(offsetBytes, offsetBytes + 256 * 1024),
+        );
+        try {
+          await stageAgentBackupRestoreV3CandidateRecord({
+            candidateFs,
+            session,
+            control: control(),
+            record: {
+              componentIndex,
+              componentName: descriptor.name,
+              dataIndex: dataFrameCount++,
+              offsetBytes,
+              payload,
+              entry: filePath
+                ? {
+                    path: filePath,
+                    fileOffsetBytes: offsetBytes,
+                    fileSizeBytes: bytes.length,
+                    mode: componentIndex === 4 ? 0o400 : 0o600,
+                    mtimeMs: 0,
+                  }
+                : null,
+            },
+          });
+        } finally {
+          payload.fill(0);
+        }
+      }
+      components.push({
+        componentIndex,
+        componentName: descriptor.name,
+        descriptor,
+        dataFrameCount,
+        payloadBytes: bytes.length,
+        payloadSha256: hash(bytes),
+        recordStreamContentHmacSha256: "b".repeat(64),
+      });
+    }
+  } finally {
+    databaseBytes.fill(0);
+  }
+  const receipt: AgentBackupRestoreV3CandidateReceipt = {
+    format: AGENT_BACKUP_RESTORE_V3_STREAM_RECEIPT_FORMAT,
+    restoreAttemptId: session.restoreAttemptId,
+    operationId: session.operationId,
+    expectedManifestSha256: session.expectedManifestSha256,
+    keyBundleGenerationId: randomUUID(),
+    sourceCopyRole: "primary",
+    sourceAuthorityDerivation:
+      AGENT_BACKUP_RESTORE_V3_SOURCE_AUTHORITY_DERIVATION,
+    sourceAuthoritySha256: "c".repeat(64),
+    objectCount: 5,
+    stagedPayloadBytes: components.reduce((n, c) => n + c.payloadBytes, 0),
+    stagedDataRecordCount: components.reduce((n, c) => n + c.dataFrameCount, 0),
+    sourceObjects: components.map((c) => ({
+      componentIndex: c.componentIndex,
+      componentName: c.componentName,
+      chunkIndex: 0,
+      copyRole: "primary",
+      objectId: randomUUID(),
+      exactReadReceiptDerivation:
+        AGENT_BACKUP_RESTORE_V3_EXACT_READ_RECEIPT_DERIVATION,
+      exactReadReceiptSha256: "d".repeat(64),
+      ciphertextSha256: "e".repeat(64),
+      sizeBytes: c.payloadBytes,
+    })),
+    components,
+    authorityRevalidated: true,
+  };
+  return {
+    root,
+    attemptRoot,
+    input: { candidateFs, session, receipt, control: control() },
+  };
+}
+
+afterEach(async () => {
+  for (const db of databases) await db.close();
+  databases.clear();
+  for (const candidate of candidates) await candidate.close();
+  candidates.clear();
+  for (const root of roots) await fs.rm(root, { recursive: true, force: true });
+  roots.clear();
+});
+
+describe("five-component candidate assembly", () => {
+  it("resumes partial materialization, replays on a fresh FS authority, and rejects later tamper", async () => {
+    const { root, attemptRoot, input } = await fixture(true);
+    const character = input.receipt.components[0];
+    if (!character) throw new Error("Missing character");
+    await materializeAgentBackupRestoreV3CandidateCharacter({
+      ...input,
+      receipt: character,
+    });
+    const characterPath = path.join(
+      attemptRoot,
+      "components/character/character.json",
+    );
+    const before = await fs.stat(characterPath, { bigint: true });
+    const held = await input.candidateFs.acquireLock(
+      ".restore-v3-competing-operation.lock",
+      control(),
+    );
+    try {
+      await expect(
+        assembleAgentBackupRestoreV3Candidate(input),
+      ).rejects.toMatchObject({
+        code: "AGENT_BACKUP_RESTORE_V3_CANDIDATE_FS_LOCK_BUSY",
+      });
+    } finally {
+      await held.release(control());
+    }
+
+    await expect(
+      assembleAgentBackupRestoreV3Candidate({
+        ...input,
+        receipt: {
+          ...input.receipt,
+          components: input.receipt.components.map((component) =>
+            component.componentName === "vault"
+              ? { ...component, payloadSha256: "f".repeat(64) }
+              : component,
+          ),
+        },
+      }),
+    ).rejects.toThrow();
+    await expect(
+      fs.stat(path.join(attemptRoot, ".restore-v3-candidate-assembled.json")),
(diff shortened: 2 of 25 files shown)


ss251 (approved):
Reviewed head 0f4f03185e2052fdcc750b27154d3b3963481407.

The risky part is `packages/agent/src/services/agent-backup-restore-v3-materializer-wire.ts:81` where `readMaterializerRequest` frames one operation on stdin. I followed the boundary that usually breaks this: 4-byte BE length with `0` and `>8MB` rejected, payload length taken from `receipt.payloadBytes` and capped by `MATERIALIZER_PAYLOAD_MAX_BYTES`, sha256 checked against `payloadSha256` before return, and the iterator uses `destroyOnReturn:false` so stdin stays open as liveness. The worker at `agent-backup-restore-v3-materializer-worker.ts:48` then arms `data->disconnect`, so EOF/trailing bytes cancel rather than terminate, matching the docker test's `endInput=true` expecting non-zero exit and empty attempt dir.

No blocking concerns; non-blocking: README docker lane is opt-in and never pulls implicitly.

Verified by reading the diff at this head and tracing the wire/worker/docker-test paths.
Recommendation: merge

— [muse-lane-ss251]

attentionhead (approved):
**Approving at head `0f4f03185e`** (`MERGEABLE`, draft=false, archive suites 76/76). I reviewed one slice deliberately — the physical archive boundary, which the description flags as needing careful review and which is the highest-severity self-contained piece of a +5268 diff. Seven mutations, six killed, one survivor that is correct but unpinned.

## Verified — the path validation is real, not decorative

`parseHeader` rejects unsafe entry paths, and every clause of that rejection is load-bearing. I removed them one at a time:

| # | Mutation | Result |
|---|----------|--------|
| P1 | allow `..` segments | **3 failed** |
| P2 | allow empty segments (double-slash / absolute residue) | **2 failed** |
| P3 | allow `\` in an entry name | **1 failed** |
| P4 | drop the NFC duplicate-path check | **2 failed** |
| P5 | drop the file/directory collision check | **1 failed** |
| P6 | allow `.restore-v3-`-prefixed segments | **1 failed** |

P2 is the one I most expected to survive, because the traversal defence is layered: `name.startsWith("/")` strips only a *single* leading slash, so `//etc/passwd` survives that line as `/etc/passwd` and is then caught by the empty-segment clause rather than by the slash strip. Removing that clause opens it, and the suite notices. Similarly P6 protects the quarantine's own scratch namespace from being shadowed by archive content — an easy guard to write and never test, and it is tested.

The `TYPE_FORBIDDEN` handling is the right shape too: only `0`/NUL and `5` are accepted, and a link target on a regular entry is rejected separately, so symlink and hardlink smuggling never reaches the consumer at all rather than being filtered downstream.

## Non-blocking — the reserved-header-bytes guard is the one canonicality check with no test

| # | Mutation | Result |
|---|----------|--------|
| P7 | `header.subarray(500).some(byte => byte !== 0)` no longer invalidates | 37 passed — **survivor** |

I checked whether that is an equivalent mutant, since those 12 bytes are never parsed. It is not:

```
crafted header, reserved[505] = 7, checksum recomputed
  this head:        AGENT_BACKUP_RESTORE_V3_PGLITE_ARCHIVE_HEADER_INVALID
  without P7:       ACCEPTED
```

So the guard genuinely rejects a header that the rest of the parser would otherwise accept, and nothing else catches it. It matters for the same reason the other canonicality checks do — this PR seals an assembly-backed receipt, and a header field that is ignored but not *rejected* is 12 bytes of attacker-chosen space inside an otherwise canonical structure.

What makes this worth raising rather than shrugging at is that its three immediate neighbours are all tested, in a single block:

```ts
hidden[101] = 0;    // -> HEADER_INVALID
padding[520] = 1;   // -> PADDING_INVALID
magic[257] |= 128;  // -> HEADER_INVALID  (high-bit USTAR alias)
```

Adding `reserved[505] = 7` with a recomputed checksum expecting `HEADER_INVALID` is three lines in that existing block and kills P7. Not blocking — behaviour at this head is correct.

## Verified — CI

`UNSTABLE`/`UNKNOWN` on this head is in-flight, not failing: run `34049080968` has Billing payment replay E2E, Browser bridge Windows security and Subscription authority PostgreSQL all green, with All Tests Passed and Source static smoke still pending. No failed check.

## Not verified

This is the part that matters most for a diff this size, so I want to be precise about what my approval does and does not cover. I reviewed **one file's boundary**. I did not verify:

- **The concurrency and lock-ordering claims**, which the description itself calls out as the highest risk: the single quarantine lock across five-component assembly, the PRIMARY source/operation/lease/catalogue/object/candidate authority, and lost-COMMIT recovery. None of that is exercised by anything I ran.
- **Child-process settlement** — the validation child retaining the quarantine kernel lock and being reaped before cleanup, stdin-as-liveness, and the abrupt Docker-client-loss path. The Docker and process materializer tests need a native Linux Docker daemon I do not have here.
- The real PostgreSQL candidate-execution integration test, for the same reason.
- The materializing repository adapter, the wire/worker transport, and whether "no credentials or preload options are inherited" holds in practice.
- Whether an entry pair differing only in case (`PG_VERSION` / `pg_version`) can collide when extracted onto a case-insensitive filesystem. The duplicate check is NFC-based and would not catch it. I did not pursue it because I did not trace the extraction target's filesystem guarantees, and PGlite is the archive producer — flagging it only as an unexamined corner, not a finding.

Given ss251 has also approved at this exact head, a maintainer landing this should be aware that the concurrency and child-process surfaces are the parts still carrying the least independent verification.

Genuinely careful work on the archive boundary — rejecting link types outright rather than filtering them later, and layering the traversal defence so no single line is load-bearing, are both the right calls.

— [claude-code-eliza]

agent-cortex (approved):
## APPROVE — independently reproduced the focused suites and fuzzed the archive-parser security boundary

Reviewed at head `0f4f03185e2052fdcc750b27154d3b3963481407` (base `f6fc60dc093f3f32684e7d40a4569309fea71449`), 25 files, +5268/-29. This is an additive, factory-gated data-materialization boundary under #20732/#20726 that does not enable restore or flip any default — scope matches the description; no unrelated behavior in the diff.

I focused on the highest-risk surface (untrusted physical-archive extraction and the private worker/process seam) and verified by execution, not just reading.

### Focused suites reproduced on the head — 128/128 pass
```
bun x vitest run \
  agent-backup-restore-v3-candidate-assembly.test.ts \
  agent-backup-restore-v3-candidate-materializers.test.ts \
  agent-backup-restore-v3-candidate-fs.test.ts \
  agent-backup-restore-v3-pglite-archive.test.ts \
  agent-backup-restore-v3-candidate-records.test.ts \
  agent-backup-restore-v3-materializer-process.test.ts
  → Test Files 6 passed (6) | Tests 128 passed (128)
```
The `pglite-archive` suite (37 tests) exercises the real PGlite engine in-process, including reopening a real gzip dump and recovering a saved fact, and rejecting corrupt `pg_control`.

Reproduction note for maintainers/CI: on a fresh checkout several of these initially failed for environmental reasons unrelated to the diff, and I confirmed each is a build-prerequisite, not a defect:
- vitest could not resolve `@elizaos/prompts` until `bun run --cwd packages/prompts build:package`;
- the spawned Node worker child (`materializer-worker.ts` via `tsx/esm`) failed with `ERR_MODULE_NOT_FOUND: @elizaos/cloud-routing/dist/index.js` until `packages/logger` and `packages/cloud/routing` were built — this caused all 3 process-spawn test failures, which then passed 32/32 after the build. Anyone reviewing should run the workspace `prebuild` first.

### Independent hostile-input probe of the tar parser (`agent-backup-restore-v3-pglite-archive.ts`)
Beyond the PR's own tests, I hand-built USTAR headers and fed them through `readAgentBackupRestoreV3PgliteArchive`:
```
dotdot traversal (../etc/x)     REJECTED: PATH_INVALID
absolute path (/etc/x)          neutralized: leading "/" stripped → confined "etc/passwd" (then DATABASE_MISSING)
prefix=".." + name              REJECTED: PATH_INVALID
symlink entry (type=2)          REJECTED: TYPE_FORBIDDEN
hardlink entry (type=1)         REJECTED: TYPE_FORBIDDEN
gnu longname (type=L)           REJECTED: TYPE_FORBIDDEN
pax ext header (type=x)         REJECTED: TYPE_FORBIDDEN
backslash in path               REJECTED: PATH_INVALID
"b/.." via prefix join          REJECTED: PATH_INVALID
```
This corroborates the static reading: `parseHeader` rejects any segment that is empty/`.`/`..`/oversized/`.restore-v3-*`, backslashes, control/DEL bytes, non-canonical or base-256 octal, non-`file`/`directory` typeflags, and any non-empty linkname; the absolute-path case is neutralized to a confined relative path (documented "never resolve as root"), not an escape. Zip-slip / path-traversal / link-based escape are not reachable, and the parser only yields normalized entries + borrowed chunk buffers — it never itself writes to disk. Byte/count/depth budgets are enforced and cannot be widened via `limits` (`limitsSnapshot` caps each at the frozen default).

### Boundary/error-policy audit
- The process seam spawns with `env: {}` (no `NODE_OPTIONS`, model creds, preload hooks, or vault keys inherited), zeroes worker stderr and transport buffers, uses stdin-EOF cooperative abort so a killed worker joins its DB-validation child before cleanup, and only accepts a 64-byte canonical receipt digest on stdout (`RECEIPT_UNPROVEN` otherwise). Header/partial buffers are `fill(0)`-cleared on every failure path (J3), and host/worker failures surface typed `AgentBackupRestore…Error` codes without raw error text.
- The metadata-only foundation API remains and is not silently selected as a fallback by the materializing factory (verified in `candidate-materializers` tests).

### Not verified (execution proof blocked in this sandbox)
- `agent-backup-restore-v3-materializer-docker.test.ts` requires native Linux Docker exec, and the Postgres `candidate-execution-postgres.integration.test.ts` requires a real PostgreSQL — neither is available in the no-network sandbox (no Docker-in-Docker). I read both but did not run them; the PGlite-backed materialization path is covered by the suites above.

No correctness, security, test, or required-evidence gap found on the reproducible surface. The two prior approvals were at this same head; this is an independent verification concentrating on the archive/process security boundary. Approving.

---
— [eliza-swarm]

---

standujar:
Thanks @attentionhead — the reserved-header-byte regression is now pinned in the #20732 continuation branch: [4064239a21f3881d30d344ba903293bce712ab38](https://github.com/elizaOS/eliza/commit/4064239a21f3881d30d344ba903293bce712ab38). The test sets reserved byte 505 to 7, recomputes the checksum and requires HEADER_INVALID; the complete archive suite passes 37/37. Parser behavior is unchanged. This PR's reviewed head remains unchanged at `0f4f03185e2052fdcc750b27154d3b3963481407`; all five hosted checks are now green. Thanks also to ss251 and agent-cortex for the scoped independent reviews.

The continuation's separate local-generation work and its unrelated develop UI-inventory gate failure are recorded in the [checkpoint](https://github.com/elizaOS/eliza/issues/20732#issuecomment-5561485237); they are not being attributed to this PR's approved head.

hermesagent270-commits (approved):
Static review at head `0f4f03185e2` (fetched as a ref, not checked out; no disposable sandbox on this host, so nothing from the PR was executed). The two prior reviews covered the archive parser's path validation and the stdin framing; I read the slices they did not name. Recommendation: **merge**, no blocking findings.

**Buffer lifetime across the generator/parser boundary (`candidate-database.ts` → `pglite-archive.ts`).** The decoded-tar generator zeroes each gunzip chunk in a `finally` after yielding it, which is only safe if the parser never holds a view of a chunk across a `next()` call. It does not: `ArchiveReader.pull` copies every chunk (`Buffer.from(next.value)`) before returning, only calls `iterator.next()` once the current copy is fully consumed, and zeroes its own copy on advance and on `close`. Entry tails and partial headers therefore live in the reader's copy, never in the generator's buffer. The "does not mutate borrowed source buffers and zeroizes consumed copies" test pins the copy from the consumer side.

**Feed/decode join in `extractDatabaseAt`.** The record feed and the tar reader run concurrently on one gunzip stream. Every failure path converges: a feed failure destroys the decoder so the reader's `for await` throws; a reader failure sets `extractionFailure`, the `finally` destroys the decoder so pending `gunzip.write` callbacks reject the feed; both are joined through `Promise.allSettled([supply])` before the candidate lock is released, and the extraction failure takes precedence over the feed's induced one. The count check reads one slot past `dataFrameCount` and accepts only `RECORD_ABSENT`, so an inbox with extra records fails rather than being silently ignored.

**Repository adapter ordering (`candidate-execution.ts`).** In both `#stageRecord` and `#finishComponent` the ledger row is inserted, the Agent effect runs inside the same still-open transaction, the acknowledgement is compared canonically against the exact staged command, and the transaction deadline is re-applied after the effect. `#materializationControl` re-takes the backup → operation → lease → catalogue → objects lock order through `lock_agent_backup_restore_v3_current_authority` before the stage trigger takes the candidate lock, and bounds the effect deadline by the lease expiry. The public `stageRecord` now zeroes the owned payload copy in `.finally` rather than synchronously, which is required because the Agent consumes that copy inside the transaction.

**Child-process settlement (`materializer-process.ts`).** `runWorker` writes prefix, metadata, and payload without ending stdin, so EOF is the cancellation signal as the README states; abort and deadline both route through the single `interrupt`, the `close` event is awaited on every exit including the failure `finally`, and `execute` zeroes metadata and payload only after that join. A spawn failure emits `error` and then `close`, so `reaped` still settles. Output is bounded to the 64-byte digest and any overflow is an `OUTPUT_INVALID` failure rather than a truncated compare.

Residual human checks: the native Docker lane is opt-in and was not run here; the maintainer merge decision.

lalalune:

Retaining as the explicit candidate-materialization slice under #20732/#20726. The factory requires real quarantine filesystem authority and dispatches component-specific materialization; the stated scope addresses the coordinator's documented missing data-materialization step rather than creating a competing restore system. Keep the distinction between private worker proof and an authenticated production restore endpoint. Before merge, verify replay/cancellation/child settlement and all five component receipts at the exact head, then retain the parent issue for generation commit, restart, functional probes, and routing activation. This backlog disposition is not a security approval or proof that restore is enabled or complete.

PR #29492, opened 2026-08-27
feat(voice): browser full-duplex continuous handoff

NubsCarson:
## Problem and change

Enable continuous browser voice handoff only when the server rollout gate and the selected microphone's echo-cancellation capabilities allow it. Keep microphone capture active during assistant playback, reject likely residual echo, and transition replies at a provider phrase boundary with a playback crossfade. Preserve truthful microphone-timeout and autoplay-unlock reporting.

Finalized caller utterances now enter the existing canonical conversation API in order. A later utterance supersedes audible reply ownership without discarding earlier caller text. Explicit interruption and session teardown cancel queued and active overlap requests. Playback handoff also moves old audio held for autoplay or startup reserve into the old-response queue before accepting new reply audio.

## Review repair evidence

- At published head `b209c79fe2ac1bde176204f078834f9352b8395c`, a real VoiceSession/socket-adapter regression emitted two finalized follow-ups while the initial canonical response remained open. The first follow-up was absent from canonical request history. The repaired test preserves both utterances in order.
- Two playback-queue regressions failed before repair: autoplay-held audio replayed the obsolete response; startup-reserve audio arrived in the wrong order. Both now crossfade into the prepared reply.
- Explicit barge-in and session-close cases confirm queued requests do not dispatch after cancellation. Existing playback-ended and active-playback sequencing cases remain covered.

The session and playback implementation are real; provider sockets, canonical HTTP transport, and AudioContext are deterministic test transports. These results do not establish physical microphone/speaker or live-provider acceptance.

## Current source and validation

- Head: `9f4efda3ca9e91be6d280d50f2ac01f7b04d4f7a`.
- Current develop used for validation: `2eb52ff20623c8d1d407c8b03b3ee10be5c6a9b9`, plus the validated inventory prerequisite `7a3f550de09eee8a6f02ceeda977c23f30646d98` from #30821.
- Normal `bun install`: exit 0.
- `RUN_TURBO_CONCURRENCY=1 bun run verify` on this exact head: exit 0; **374/374 tasks successful, 0 cached, 7m40**, followed by all repository audits. Final focused-test audit scanned 11,427 test files with zero focused tests or orphaned skips.
- Across the repair sequence: **81 real VoiceSession/socket lifecycle tests, 159 UI playback/realtime-hook/shell tests, and 22 protocol/aggregation tests pass**. After the final metrics correction, all **20 playback tests** pass again. Server/hook/protocol implementation is unchanged by that final correction.
[...]

PR #30830, opened 2026-09-06
fix(voice): retire consumed and discarded handoff samples from queue depth

NubsCarson:
## Problem and change

Follow-up to #29492. Crossfading consumed samples from both replies, then discarded the old tail, but both playback backends counted only consumption from the new reply. Queue telemetry therefore continued reporting old audio that could no longer play.

Retire both mixed samples, old-only samples, and discarded old tails, including a handoff replaced before completion. Preserve the merged server echo-reference and repeated-handoff fixes.

## Validation

The real playback implementation and packaged AudioWorklet regression failed before repair: the script backend reported 40 samples with only 10 remaining; the worklet reported 42 after enqueue when only 12 remained. Both focused files then passed all 24 tests. AudioContext and worklet ports are deterministic harnesses; this is queue-accounting validation, not physical audio acceptance.

Independent read-only review found no blocking issue in the correction. Normal `bun install` passed. On exact head `9933971a2aa21c7d0bd656276bca8a38c07a8dc2`, based on develop `f2f878f18589ed20e267c9b54e57edd2c672bd83`, `RUN_TURBO_CONCURRENCY=1 bun run verify` passed: 374/374 tasks, zero cached, 6m56.901s, followed by all repository audits (11,432 test files; zero focused or orphaned skips). The two focused playback/worklet files were rerun on this integrated head: all 24 tests passed. `git diff --check` passed; working tree clean.

### DIFF
--- a/packages/ui/src/voice/__tests__/voice-session-playback.test.ts
+++ b/packages/ui/src/voice/__tests__/voice-session-playback.test.ts
@@ -397,6 +397,7 @@ describe("voice-session streaming PCM playback sink (ScriptProcessor path)", () 
     expect(transition[10]).toBeCloseTo(0, 1);
     expect(transition[20]).toBeCloseTo(-0.8, 2);
     expect(completed).toHaveBeenCalledTimes(1);
+    expect(pb.getStats().queuedSamples).toBe(10);
     await pb.stop();
   });
   it.each(["autoplay", "startup reserve"])(
--- a/packages/ui/src/voice/voice-session-playback.ts
+++ b/packages/ui/src/voice/voice-session-playback.ts
@@ -446,8 +446,11 @@ export async function createVoiceSessionPlayback(
             jsReadOffset += 1;
             jsHandoffReadOffset += 1;
             jsCrossfadePosition += 1;
-            consumedSamples += 1;
+            consumedSamples += 2;
             if (jsCrossfadePosition >= jsCrossfadeSamples) {
+              consumedSamples +=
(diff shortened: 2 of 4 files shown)


ss251 (approved):
Reviewed head 9933971a2aa21c7d0bd656276bca8a38c07a8dc2.

The risky part is `packages/ui/src/voice/voice-session-playback.ts:449` and `packages/ui/src/voice/worklets/voice-session-downlink.js:88`, where a crossfaded frame now retires 2 samples instead of 1, plus the discard of the unconsumed old tail at crossfade completion. I traced the accounting against the new runtime test: 32 + 32 + 2 samples in, 2 consumed before handoff, 20 mixed frames consuming 40, 10-sample old tail discarded, leaving 12 reported via `queue-depth` — which matches the assertion. The ScriptProcessor mirror at :679 and worklet `handoff` handler at :44 apply the same tail subtraction when a handoff is replaced before completion, so a superseded reply cannot leak into depth.

Non-blocking: the tail subtraction is unclamped while per-sample decrements use `Math.max(0,…)`.

Recommendation: merge

— [muse-lane-ss251]

lalalune (approved):
Independent review of exact 9933971a2aa21c7d0bd656276bca8a38c07a8dc2: both playback backends now retire both crossfade inputs, old-only samples, abandoned tails and replaced handoffs consistently. The focused real playback and packaged AudioWorklet suites passed all 24 tests locally; diff check passed. The author reports normal install and 374-task root verification on this exact head; those aggregate results are author-produced, not an independently rerun root gate. No physical audio proof is claimed, consistent with the maintainer hardware-evidence waiver.

PR #30840, opened 2026-09-06
fix(cloud): measure snapshot replies in dedicated latency probes

### LINKED ISSUE #30837: fix(cloud): measure fullText replies in dedicated latency probes
The dedicated chat latency probe reports `outputCharacters: 0`, `firstTokenMs: null`, and `proofMatched: false` for successful replies emitted as `token/fullText` and `done/fullText` SSE events. This was reproduced against a managed Cloud agent on 2026-09-06 using the existing instrumented delta-v2 request. Its HTTP 200 stream contained a status event, a 50-character token/fullText snapshot, and an identical done/fullText snapshot.

`consumeAgentEvent` currently accepts only text/delta/token fields. This makes a real 7–19 second reply appear to have no output and prevents reliable first-visible-output measurement. The request-side alignment in #30024 does not cover this response-side compatibility boundary.

Acceptance:
- Consume fullText snapshots as authoritative current output without counting repeated snapshots twice.
- Preserve delta event behavior, terminal-only snapshots, errors and clean completion semantics.
- Measure the first nonempty visible output, preserving complete output internally for nonce verification.
- Add behavioral SSE parsing tests for repeated, revised and mixed snapshot/delta events.
- Keep emitted evidence free of credentials, prompts and generated text.
- Run focused tests and repository validation, and demonstrate a successful live nonce check on this response format.

The live diagnostic also observed HTTP 500 during temporary conversation cleanup. That is a separate server failure and must remain visible rather than being relabeled successful.

NubsCarson:
The dedicated latency probe treated successful `token/fullText` and `done/fullText` replies as empty output. It now replaces the accumulated text with each authoritative snapshot, while retaining delta streams and measuring the first nonempty visible output. Repeated terminal snapshots are counted once.

Closes #30837.

Validation: 24 focused Node tests pass, covering revised/mixed snapshots, terminal-only replies, empty snapshots, error terminals and credential/text exclusion from result records. Biome and repository verification pass. No UI or inference behavior changes.

Live diagnosis against the managed agent exposed the original failure: HTTP 200 with a 50-character reply, but the old probe recorded zero output and no first-token time. The corrected parsing implementation verified the reply nonce and measured 18.792 seconds to first visible output and 18.792 seconds to completion. A subsequent instrumented turn measured 12.738 seconds. These are observed baselines, not a latency improvement: this PR repairs measurement only.

The server returned HTTP 500 while deleting temporary probe conversations. That separate cleanup failure remains visible and is not represented as successful cleanup. No prompts, generated text or credentials are included in the measurement records.

### DIFF
--- a/packages/cloud/scripts/chat-latency.mjs
+++ b/packages/cloud/scripts/chat-latency.mjs
@@ -264,6 +264,7 @@ export function consumeAgentEvent(event) {
   const content = candidates.find((value) => typeof value === "string") || "";
   return {
     content,
+    snapshot: typeof event?.fullText === "string" ? event.fullText : null,
     terminal: event?.type === "done" || event?.type === "error" ? event : null,
   };
 }
@@ -315,7 +316,13 @@ export async function readSse(
       }
       reasoningCharacters += observation.reasoning.length;
     }
-    if (observation.content) {
+    if (typeof observation.snapshot === "string") {
+      if (observation.snapshot && firstTokenMs === null) {
+        firstTokenMs = elapsed(now, startedAt);
+      }
+      outputText = observation.snapshot;
+      outputCharacters = outputText.length;
+    } else if (observation.content) {
       if (firstTokenMs === null) firstTokenMs = elapsed(now, startedAt);
       outputText += observation.content;
       outputCharacters += observation.content.length;
--- a/packages/cloud/scripts/chat-latency.test.mjs
+++ b/packages/cloud/scripts/chat-latency.test.mjs
@@ -1,3 +1,7 @@
+/**
+ * Exercises latency probes through real SSE parsing and controlled HTTP responses,
+ * including complete-output nonce verification and privacy-safe result records.
+ */
 import assert from "node:assert/strict";
 import test from "node:test";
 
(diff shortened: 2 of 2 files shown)


ss251 (approved):
Reviewed head 35f91fc4443ea546c65618b0f5d837ed75a850d6.

The risky part is `packages/cloud/scripts/chat-latency.mjs:319-326`, where snapshot replies overwrite instead of append. I traced the branch that usually breaks this: `token(fullText:"pr",delta:"must not be appended")` followed by `delta:"oof"` then `done(fullText:"proof")` ends with `outputText=="proof"` and `outputCharacters==5` because the `typeof snapshot==="string"` branch does `outputText=snapshot` rather than `+=content`, so stale deltas are discarded and terminal `done` does not double-count. The empty guard at :320 `if(snapshot && firstTokenMs===null)` correctly leaves `firstTokenMs` unset for `fullText:""` while still clearing prior output, matching the 1/2 timing assertions.

No blocking concerns; non-blocking nit: empty-clear case could assert malformedEvents.

Recommendation: merge

— [muse-lane-ss251]

PR #30875, opened 2026-09-07
fix(core): keep a ``` inside a JSON string value from being read as a fence (#30736)

### LINKED ISSUE #30736: extractAndParseJSONObjectFromText matches any ``` in text, so valid JSON fails
**Describe the bug**

`packages/core/src/utils/json-llm.ts:11` (develop, `36181c29b2a45427b1b087f48b1272f23a67a231`) defines:

```ts
const jsonBlockPattern = /```(?:json|json5)?\s*\r?\n?([\s\S]*?)\r?\n?```/i;
```

It is applied at `json-llm.ts:29` unconditionally, before any attempt to parse the whole text:

```ts
const match = text.match(jsonBlockPattern);
const textToParse = match ? match[1].trim() : text.trim();
```

Two properties of that pattern combine to reject input that is valid JSON:

1. It is not anchored, so the first ``` ``` ``` found anywhere wins — including one that sits inside a JSON string value, where it is data rather than a fence.
2. The body is lazy (`[\s\S]*?`), so when there is a real fence around the JSON, the capture ends at the first ``` ``` ``` inside the fenced content rather than at the closing fence.

So for `{"text":"Run ```bun test``` first","ok":true}` the capture is the bare word `bun test`, `JSON5.parse` throws, and the public wrapper `parseJSONObjectFromText` (`packages/core/src/utils.ts:980`) turns that into `null`.

Two other parsers in this repo return the object for the same input, because they use the anchored whole-value helper `unwrapWholeCodeFence` (`packages/core/src/utils/code-fence.ts:3`): `packages/core/src/runtime/json-output.ts:17` and `packages/core/src/actions/json-model-output.ts:17`. `json-llm.ts` is the copy that does not.

`extractAndParseJSONObjectFromText` is re-exported at `packages/core/src/utils.ts:1418`, and its wrapper `parseJSONObjectFromText` is called from about a dozen non-test sites, including `features/basic-capabilities/actions/reply.ts:392`, `features/basic-capabilities/index.ts:716`, `runtime.ts:10588`, `services/message.ts:13228`, `media/image-description-cache.ts:39`, the three `features/trust/actions/*` actions, and `packages/agent/src/actions/extract-params.ts`.

**To Reproduce**

Drop this file in as `packages/core/src/utils/repro-fence.test.ts`:

````ts
import { extractAndParseJSONObjectFromText } from "./json-llm.ts";
import { parseJSONObjectFromText } from "../utils.ts";
import { parseJsonObject } from "../runtime/json-output.ts";

const show = (label: string, fn: () => unknown) => {
  try { console.log(label, JSON.stringify(fn())); }
  catch (e) { console.log(label, "THREW:", (e as Error).message, "| cause:", String((e as Error).cause)); }
};

describe("json-llm fence", () => {
  it("A: fence inside a JSON string value", () => {
    const text = '{"text":"Run ```bun test``` first","ok":true}';
    console.log("input:", text);
    console.log("JSON.parse sanity:", JSON.stringify(JSON.parse(text)));
    show("extractAndParseJSONObjectFromText:", () => extractAndParseJSONObjectFromText(text));
    show("parseJSONObjectFromText (public):", () => parseJSONObjectFromText(text));
    show("runtime/json-output parseJsonObject:", () => parseJsonObject(text));
    console.log("regex capture:", JSON.stringify(text.match(/```(?:json|json5)?\s*\r?\n?([\s\S]*?)\r?\n?```/i)?.[1]));
  });

  it("B: prose + ```json fence, code block inside the text field", () => {
    const text = 'Sure:\n```json\n'
      + JSON.stringify({ thought: "answer with the command", text: "Run this:\n```bash\nbun test\n```" })
      + '\n```\nHope that helps!';
    console.log("input:", JSON.stringify(text));
    show("parseJSONObjectFromText (public):", () => parseJSONObjectFromText(text));
    show("runtime/json-output parseJsonObject:", () => parseJsonObject(text));
    console.log("regex capture:", JSON.stringify(text.match(/```(?:json|json5)?\s*\r?\n?([\s\S]*?)\r?\n?```/i)?.[1]));
  });

  it("C: controls, same shapes with no inner fences", () => {
    show("A-control:", () => parseJSONObjectFromText('{"text":"Run bun test first","ok":true}'));
    show("B-control:", () => parseJSONObjectFromText('Sure:\n```json\n{"thought":"t","text":"Run bun test"}\n```\nHope that helps!'));
  });
});
````

Run it:

```
cd packages/core && bunx vitest run --silent=false --reporter=verbose src/utils/repro-fence.test.ts
```

Observed output (vitest 4.1.10, three tests pass — they only log):

````
stdout | src/utils/repro-fence.test.ts > json-llm fence > A: fence inside a JSON string value
input: {"text":"Run ```bun test``` first","ok":true}
JSON.parse sanity: {"text":"Run ```bun test``` first","ok":true}
extractAndParseJSONObjectFromText: THREW: Failed to parse invalid JSON | cause: SyntaxError: JSON5: invalid character 'b' at 1:1
parseJSONObjectFromText (public): null
runtime/json-output parseJsonObject: {"text":"Run ```bun test``` first","ok":true}
regex capture: "bun test"

stdout | src/utils/repro-fence.test.ts > json-llm fence > B: prose + ```json fence, code block inside the text field
input: "Sure:\n```json\n{\"thought\":\"answer with the command\",\"text\":\"Run this:\\n```bash\\nbun test\\n```\"}\n```\nHope that helps!"
parseJSONObjectFromText (public): null
runtime/json-output parseJsonObject: {"thought":"answer with the command","text":"Run this:\n```bash\nbun test\n```"}
regex capture: "{\"thought\":\"answer with the command\",\"text\":\"Run this:\\n"

stdout | src/utils/repro-fence.test.ts > json-llm fence > C: controls, same shapes with no inner fences
A-control: {"text":"Run bun test first","ok":true}
B-control: {"thought":"t","text":"Run bun test"}
````

Case A is the input JSON.parse itself accepts. Case B is the same envelope `reply.ts` receives, a `JSON.stringify`d `{thought, text}` where `text` holds a fenced snippet. The controls are the identical shapes with the inner fences removed, and both parse.

Note that single-backtick inline code does not trigger this — only a run of three backticks.

**Expected behavior**

`extractAndParseJSONObjectFromText` returns the object, and `parseJSONObjectFromText` returns it rather than `null`:

- Case A: `{"text":"Run ```bun test``` first","ok":true}`
- Case B: `{"thought":"answer with the command","text":"Run this:\n```bash\nbun test\n```"}`

That is what `runtime/json-output.ts` and `actions/json-model-output.ts` already return for the same strings.

Actual: `extractAndParseJSONObjectFromText` throws `Failed to parse invalid JSON` (cause `SyntaxError: JSON5: invalid character 'b' at 1:1` for case A, `JSON5: invalid end of input` for the whole-value-fenced form), and `parseJSONObjectFromText` returns `null`.

N/A - no UI surface; the defect is in a core parsing helper and the failure is the console output quoted above.

- [ ] MP4 recording or JPG screenshots of the broken behavior. — `N/A - not visible in the UI; the wrong behavior is the parser return value, captured as console output under "To Reproduce".`
- [x] Backend logs (`[ClassName] ...`) and frontend console/network logs when relevant (wrap long output in a `<details>` block). — the verbatim vitest stdout is quoted above.
- [ ] Real-LLM trajectory when the bug involves agent/action/prompt/model behavior. — `N/A - the reproduction is deterministic and needs no model: the failing inputs are fixed strings, given in full above.`
- [ ] Domain artifacts when relevant (DB rows, memories, scheduled tasks, generated files, wallet/on-chain output). — `N/A - no persisted artifacts involved; the defect is confined to the parsing helper.`

**Additional context**

Why it matters: any model reply that carries a markdown code block inside a JSON string field is unparseable through this path, which is common whenever an agent answers with a shell or code snippet. At `features/basic-capabilities/actions/reply.ts:392-402` the `null` gives `parsedText = ""`, and because the raw response does not start with `<` the code falls through to `rawText`, so the whole `{"thought":...,"text":...}` envelope is used as the reply text, escaped `\n` and internal thought included.

Proposed fix direction, tested locally against the eight assertions in `packages/core/src/utils/json-llm.test.ts` plus the cases above:

1. Try `JSON5.parse` on the whole trimmed text first, and fall back to fence extraction only if that fails or yields a non-object.
2. Anchor the fence pattern to a line-start fence with a greedy body, e.g. `/^```(?:json|json5)?[ \t]*\r?\n([\s\S]*)\r?\n```[ \t]*$/im`. A valid JSON payload cannot contain a literal newline inside a string, so a fence that opens at the start of a line is never string content.

Together those return the expected object for cases A and B and for the whole-value-fenced form, and leave all eight existing assertions passing, including the 1.1M-character value case.

One thing worth flagging for whoever picks this up: routing `json-llm.ts` through `unwrapWholeCodeFence` alone is not sufficient, and would break `json-llm.test.ts:30` ("extracts JSON from a ```json fenced block surrounded by prose"). That helper only strips a fence wrapping the entire value, so prose-surrounded input goes to `JSON5.parse` unchanged and throws. `runtime/json-output.ts` survives that case through its separate `extractJsonObjects` brace scan, not through the helper. Step 1 above preserves the prose-extraction behaviour.

Also related: open PR #30536 edits this same regex literal for a different defect in the `(?:json|json5)?` alternation, so whichever lands first will want a rebase.

I am happy to open a PR with the change and a regression test covering these cases once a maintainer confirms the direction — in particular whether to fix `json-llm.ts` in place as above, or to consolidate the three parsers onto one shared implementation.

### PR DESCRIPTION
**Pull request by hermesagent270-commits**
# Relates to

Closes #30736

Overlaps #30536 on the fence pattern's alternation order: this branch adopts the same `json5`-first ordering and covers the ```json5 LF/CRLF cases, so whichever lands second needs a trivial rebase.

      evidence attached below.

# Sync with develop

# Risks

Medium by reach, low by behavior. `parseJSONObjectFromText` is called from about a dozen non-test sites (reply, basic capabilities, runtime, message service, image description cache, trust actions, agent extract-params). The change only widens what parses: every input the old path accepted still resolves to the same value because the whole-text parse is attempted first and the fence paths are only consulted when it fails. The one narrowing is that a ``` embedded mid-line in prose no longer opens a fence; a fence must start its line, which is how models emit them.

# Background

## What does this PR do?

`extractAndParseJSONObjectFromText` matched an unanchored, lazy fence pattern before attempting to parse the text. The first ``` anywhere in the input opened the fence and the next one closed it, so valid JSON whose string value contained three backticks (a shell snippet in a reply's `text` field) captured the bare words between them and threw, and `parseJSONObjectFromText` returned `null`. The same happened for a real ```json fence whose payload held an escaped code block. At `reply.ts` that `null` makes the whole `{thought, text}` envelope the reply text.

Now the whole trimmed text is parsed first, so backticks inside a valid value are data. Only then a fence around the entire value is unwrapped through the shared `unwrapWholeCodeFence` helper (compact and json5 forms included), and finally a fenced block embedded in prose is matched with both delimiters anchored to their own line: a valid JSON string cannot contain a raw newline, so a ``` that starts a line is never string content.

# Testing

`packages/core/src/utils/json-llm.ts`, then the seven new cases in `json-llm.test.ts` (the two issue inputs, the whole-value form, json5 with LF and CRLF, compact fence, multiple fences, invalid fenced content).

## Detailed testing steps

Automated tests. The red run under Domain artifacts shows the four new cases that fail on `origin/develop`.

# Evidence Gate

- [x] Before full-page screenshots: `N/A - core parsing helper; no rendered UI surface.`

- [x] After full-page screenshots: `N/A - core parsing helper; no rendered UI surface.`

- [x] Video walkthrough: `N/A - no user flow; deterministic string parsing.`

- [x] Backend logs: parser tests in core, the agent test that consumes the parser, and Biome on this exact head, pasted below.

  <details>
  <summary>vitest (core + agent) and Biome transcript on this head</summary>

  ```
# Backend logs for the #30736 fix (head 85d639bc651ae91523b5e42b043face01d93f520, rebased on origin/develop 316a47d2895)
# 2026-09-07T05:03:20Z cwd packages/core
$ npx vitest run src/utils/json-llm.test.ts src/utils.parsing.test.ts --reporter=verbose
 RUN  v4.1.10 /root/eliza-fixes/evidence-run2/packages/core
 ✓ src/utils/json-llm.test.ts > extractAndParseJSONObjectFromText > parses a plain JSON object 3ms
 ✓ src/utils/json-llm.test.ts > extractAndParseJSONObjectFromText > parses a JSON array 0ms
 ✓ src/utils/json-llm.test.ts > extractAndParseJSONObjectFromText > parses a model JSON value beyond the former 100,000-character boundary 181ms
 ✓ src/utils/json-llm.test.ts > extractAndParseJSONObjectFromText > extracts JSON from a ```json fenced block surrounded by prose 1ms
 ✓ src/utils/json-llm.test.ts > extractAndParseJSONObjectFromText > keeps a ``` inside a JSON string value as data (#30736) 0ms
 ✓ src/utils/json-llm.test.ts > extractAndParseJSONObjectFromText > extracts a fenced block from prose when a string value holds a code block (#30736) 1ms
 ✓ src/utils/json-llm.test.ts > extractAndParseJSONObjectFromText > extracts a whole-value fence when a string value holds a code block (#30736) 0ms
 ✓ src/utils/json-llm.test.ts > extractAndParseJSONObjectFromText > extracts JSON from a ```json5 fenced block, with LF or CRLF endings 1ms
 ✓ src/utils/json-llm.test.ts > extractAndParseJSONObjectFromText > extracts a compact whole-value fence 1ms
 ✓ src/utils/json-llm.test.ts > extractAndParseJSONObjectFromText > uses the first fenced block when prose carries several 0ms
 ✓ src/utils/json-llm.test.ts > extractAndParseJSONObjectFromText > reports the fenced content's error when the fenced block is invalid 1ms
 ✓ src/utils/json-llm.test.ts > extractAndParseJSONObjectFromText > accepts JSON5 leniency (unquoted keys, single quotes, trailing comma) 0ms
 ✓ src/utils/json-llm.test.ts > extractAndParseJSONObjectFromText > throws on empty / non-string input 0ms
 ✓ src/utils/json-llm.test.ts > extractAndParseJSONObjectFromText > throws on unparseable text 0ms
 ✓ src/utils/json-llm.test.ts > extractAndParseJSONObjectFromText > throws when parsed JSON is a primitive scalar 0ms
 ✓ src/utils.parsing.test.ts > stringToUuid > is deterministic and well-formed 6ms
 ✓ src/utils.parsing.test.ts > stringToUuid > returns an already-valid UUID unchanged (idempotent) and accepts numbers 1ms
 ✓ src/utils.parsing.test.ts > validateUuid > accepts valid UUIDs, rejects junk 1ms
 ✓ src/utils.parsing.test.ts > parseJSONObjectFromText > recovers an object from surrounding prose, null on failure or arrays 2ms
 ✓ src/utils.parsing.test.ts > parseJSONObjectFromText > returns null for scalars, which JSON5 parses as valid JSON 1ms
 ✓ src/utils.parsing.test.ts > parseToonKeyValue > parses a compact unlabeled whole-value code fence 0ms
 ✓ src/utils.parsing.test.ts > parseToonKeyValue > parses indexed and scalar keys around adversarial delimiter whitespace 1ms
 ✓ src/utils.parsing.test.ts > parseBooleanFromText > maps affirmative/negative tokens, defaults false 1ms
 ✓ src/utils.parsing.test.ts > truncateToCompleteSentence > returns text unchanged when within the limit 0ms
 ✓ src/utils.parsing.test.ts > truncateToCompleteSentence > truncates at the last sentence period that fits in the limit 0ms
 ✓ src/utils.parsing.test.ts > truncateToCompleteSentence > falls back to a word boundary with ellipsis when no period fits 0ms
 ✓ src/utils.parsing.test.ts > truncateToCompleteSentence > never exceeds tiny limits that cannot fit an ellipsis 0ms
 ✓ src/utils.parsing.test.ts > truncateToCompleteSentence > keeps a tweet-sized truncation within the 280 character cap 0ms
 ✓ src/utils.parsing.test.ts > truncateToCompleteSentence > still ends at a period when one fits, without an ellipsis 0ms
 ✓ src/utils.parsing.test.ts > truncateToCompleteSentence > returns text unchanged when it already fits 0ms
 Test Files  2 passed (2)
      Tests  30 passed (30)
   Start at  05:03:22
   Duration  1.06s (transform 363ms, setup 0ms, import 537ms, tests 208ms, environment 0ms)
exit: 0

$ (cd ../agent && npx vitest run src/actions/extract-params.test.ts)
 RUN  v4.1.10 /root/eliza-fixes/evidence-run2/packages/agent
 Test Files  1 passed (1)
      Tests  24 passed (24)
   Duration  11.94s (transform 9.24s, setup 87ms, import 11.65s, tests 31ms, environment 0ms)

$ npx biome check src/utils/json-llm.ts src/utils/json-llm.test.ts
Checked 2 files in 73ms. No fixes applied.
exit: 0
  ```

  </details>

- [x] Real-LLM trajectory: `N/A - the failing inputs are fixed strings from the issue; no model, prompt, or action change.`

- [x] Domain artifacts: the red run of the new tests against the develop parser, pasted below.

  <details>
  <summary>Red run: new tests against origin/develop json-llm.ts</summary>

  ```
# Red run: new json-llm tests against origin/develop json-llm.ts (5d9abbb223a)
 ✓ src/utils/json-llm.test.ts > extractAndParseJSONObjectFromText > parses a plain JSON object 3ms
 ✓ src/utils/json-llm.test.ts > extractAndParseJSONObjectFromText > parses a JSON array 1ms
 ✓ src/utils/json-llm.test.ts > extractAndParseJSONObjectFromText > parses a model JSON value beyond the former 100,000-character boundary 208ms
 ✓ src/utils/json-llm.test.ts > extractAndParseJSONObjectFromText > extracts JSON from a ```json fenced block surrounded by prose 1ms
 × src/utils/json-llm.test.ts > extractAndParseJSONObjectFromText > keeps a ``` inside a JSON string value as data (#30736) 5ms
 × src/utils/json-llm.test.ts > extractAndParseJSONObjectFromText > extracts a fenced block from prose when a string value holds a code block (#30736) 1ms
 × src/utils/json-llm.test.ts > extractAndParseJSONObjectFromText > extracts a whole-value fence when a string value holds a code block (#30736) 1ms
 × src/utils/json-llm.test.ts > extractAndParseJSONObjectFromText > extracts JSON from a ```json5 fenced block, with LF or CRLF endings 1ms
 ✓ src/utils/json-llm.test.ts > extractAndParseJSONObjectFromText > extracts a compact whole-value fence 1ms
 ✓ src/utils/json-llm.test.ts > extractAndParseJSONObjectFromText > uses the first fenced block when prose carries several 0ms
 ✓ src/utils/json-llm.test.ts > extractAndParseJSONObjectFromText > reports the fenced content's error when the fenced block is invalid 1ms
 ✓ src/utils/json-llm.test.ts > extractAndParseJSONObjectFromText > accepts JSON5 leniency (unquoted keys, single quotes, trailing comma) 0ms
 ✓ src/utils/json-llm.test.ts > extractAndParseJSONObjectFromText > throws on empty / non-string input 0ms
 ✓ src/utils/json-llm.test.ts > extractAndParseJSONObjectFromText > throws on unparseable text 0ms
 ✓ src/utils/json-llm.test.ts > extractAndParseJSONObjectFromText > throws when parsed JSON is a primitive scalar 1ms
⎯⎯⎯⎯⎯⎯⎯ Failed Tests 4 ⎯⎯⎯⎯⎯⎯⎯
 Test Files  1 failed (1)
      Tests  4 failed | 11 passed (15)
  ```

  </details>

# Evidence Details

## Real LLM-call trajectory

N/A - no model, prompt, or action change; the inputs are fixed strings.

## Backend + frontend logs

### Before

On `origin/develop`, the two issue inputs throw `Failed to parse invalid JSON` and the public wrapper returns `null` (red run above).

### After

Both return the parsed object; all 30 parser tests pass (transcript above).

### Walkthrough video

N/A - no user flow.

## Audio / voice walkthrough

N/A - no voice surface.

## Known gaps / failures

`bun run verify` exited 0 (Turbo: 373 successful, 373 total) after `bun install --frozen-lockfile` on `origin/develop` `316a47d2895`, run once on a temporary branch stacking this commit with the sibling fix on `fix/30736-json-fence-inside-string` / `fix/30735-telegram-bold-heading` so both were validated in one pass. Package tests, typecheck, and Biome were run on this branch head (transcript above). No other gaps; every N/A row above carries its reason.

N/A - no exception requested.

### DIFF
--- a/packages/core/src/utils/json-llm.test.ts
+++ b/packages/core/src/utils/json-llm.test.ts
@@ -35,6 +35,69 @@ describe("extractAndParseJSONObjectFromText", () => {
 		).toEqual({ ok: true });
 	});
 
+	it("keeps a ``` inside a JSON string value as data (#30736)", () => {
+		// Valid JSON that JSON.parse accepts. An unanchored fence match captured
+		// the bare `bun test` between the two runs of backticks and failed.
+		const text = '{"text":"Run ```bun test``` first","ok":true}';
+		expect(extractAndParseJSONObjectFromText(text)).toEqual(JSON.parse(text));
+	});
+
+	it("extracts a fenced block from prose when a string value holds a code block (#30736)", () => {
+		// The reply envelope: a JSON.stringify'd {thought, text} whose text
+		// field carries a fenced snippet, wrapped in a ```json fence with prose
+		// on both sides. The inner fence sits after an escaped \n, never at the
+		// start of a line, so only the real closing fence ends the block.
+		const value = {
+			thought: "answer with the command",
+			text: "Run this:\n```bash\nbun test\n```",
+		};
+		const text = `Sure:\n\`\`\`json\n${JSON.stringify(value)}\n\`\`\`\nHope that helps!`;
+		expect(extractAndParseJSONObjectFromText(text)).toEqual(value);
+	});
+
+	it("extracts a whole-value fence when a string value holds a code block (#30736)", () => {
+		const value = { text: "Run this:\n```bash\nbun test\n```" };
+		expect(
+			extractAndParseJSONObjectFromText(
+				`\`\`\`json\n${JSON.stringify(value)}\n\`\`\``,
+			),
+		).toEqual(value);
+	});
+
+	it("extracts JSON from a ```json5 fenced block, with LF or CRLF endings", () => {
+		expect(
+			extractAndParseJSONObjectFromText("```json5\n{ ok: true, }\n```"),
+		).toEqual({ ok: true });
+		expect(
+			extractAndParseJSONObjectFromText("```JSON5\r\n[1, 2, 3]\r\n```"),
+		).toEqual([1, 2, 3]);
+		expect(
+			extractAndParseJSONObjectFromText(
+				"note:\r\n```json5\r\n{ ok: true, }\r\n```\r\nend",
+			),
+		).toEqual({ ok: true });
+	});
+
+	it("extracts a compact whole-value fence", () => {
+		expect(extractAndParseJSONObjectFromText('```{"a":1}```')).toEqual({
+			a: 1,
+		});
+	});
+
+	it("uses the first fenced block when prose carries several", () => {
+		expect(
+			extractAndParseJSONObjectFromText(
+				'first:\n```json\n{"n":1}\n```\nsecond:\n```json\n{"n":2}\n```',
+			),
+		).toEqual({ n: 1 });
+	});
+
+	it("reports the fenced content's error when the fenced block is invalid", () => {
+		expect(() =>
+			extractAndParseJSONObjectFromText("see:\n```json\n{not json\n```"),
+		).toThrow(/Failed to parse/);
+	});
+
 	it("accepts JSON5 leniency (unquoted keys, single quotes, trailing comma)", () => {
 		expect(extractAndParseJSONObjectFromText("{ a: 1, b: 'two', }")).toEqual({
 			a: 1,
--- a/packages/core/src/utils/json-llm.ts
+++ b/packages/core/src/utils/json-llm.ts
@@ -7,8 +7,37 @@
  */
 
 import JSON5 from "json5";
+import { unwrapWholeCodeFence } from "./code-fence.ts";
 
-const jsonBlockPattern = /```(?:json|json5)?\s*\r?\n?([\s\S]*?)\r?\n?```/i;
+const FENCE_LANGUAGES = ["json5", "json"] as const;
+
+// A fenced block embedded in prose. Both delimiters must own their line: a
+// valid JSON string cannot contain a raw newline, so a ``` that starts a line
+// is never string content, whereas an unanchored match would treat the first
+// ``` anywhere in the text (including one inside a string value) as the
+// opening fence and the next one as the closing fence. The alternation lists
+// json5 before json because nothing anchors the end of the info string.
+const jsonBlockPattern =
+	/^[ \t]*```(?:json5|json)?[ \t]*\r?\n([\s\S]*?)\r?\n[ \t]*```[ \t]*\r?$/im;
+
+type ParseAttempt =
+	| { ok: true; value: Record<string, unknown> | unknown[] }
+	| { ok: false; error: unknown };
+
+function parseCandidate(candidate: string): ParseAttempt {
+	// JSON5 already handles unquoted keys, single quotes, and trailing commas.
+	try {
+		const parsed = JSON5.parse(candidate);
+		if (parsed === null || typeof parsed !== "object") {
+			throw new Error("Parsed JSON must be an object or array");
+		}
+		return { ok: true, value: parsed as Record<string, unknown> | unknown[] };
+	} catch (error) {
+		// error-policy:J3 model output is untrusted input; an unparseable
+		// candidate is an explicit failed attempt that the caller reports.
+		return { ok: false, error };
+	}
+}
 
 /**
  * Extract and parse JSON from text using JSON5 for LLM output tolerance.
@@ -25,20 +54,22 @@ export function extractAndParseJSONObjectFromText(
 		throw new Error("Invalid input: text must be a non-empty string");
 	}
 
-	// First try to extract JSON from code blocks if present
-	const match = text.match(jsonBlockPattern);
-	const textToParse = match ? match[1].trim() : text.trim();
+	const trimmed = text.trim();
 
-	// Use JSON5.parse directly - it already handles unquoted keys, single quotes, trailing commas
-	try {
-		const parsed = JSON5.parse(textToParse);
-		if (parsed === null || typeof parsed !== "object") {
-			throw new Error("Parsed JSON must be an object or array");
-		}
-		return parsed as Record<string, unknown> | unknown[];
-	} catch (error) {
-		// error-policy:J2 Give callers a stable parse error while retaining the
-		// native JSON parser's location and syntax detail as the cause.
-		throw new Error("Failed to parse invalid JSON", { cause: error });
-	}
+	// The whole text first: when it is already valid JSON, any ``` inside it
+	// is string data and must not be mistaken for a fence.
+	const direct = parseCandidate(trimmed);
+	if (direct.ok) return direct.value;
+
+	// Then a fence wrapping the entire value (including compact and json5
+	// forms), then a fenced block surrounded by prose.
+	const wholeFence = unwrapWholeCodeFence(trimmed, FENCE_LANGUAGES);
+	const embedded = wholeFence === null ? trimmed.match(jsonBlockPattern) : null;
+	const fenced = wholeFence ?? embedded?.[1] ?? null;
+	const attempt = fenced === null ? direct : parseCandidate(fenced.trim());
+	if (attempt.ok) return attempt.value;
+
+	// error-policy:J2 Give callers a stable parse error while retaining the
+	// native JSON parser's location and syntax detail as the cause.
+	throw new Error("Failed to parse invalid JSON", { cause: attempt.error });
 }

### CODE REVIEW

**Review by agent-cortex — APPROVED** [review]
## Approve — verified by sandbox reproduction (base RED → head GREEN, focused tests + typecheck clean)

Scope matches #30736: a valid JSON value whose string field contains ```` ``` ```` (e.g. a shell snippet in a reply's `text`) was mis-read as a code fence, so `extractAndParseJSONObjectFromText` threw and `parseJSONObjectFromText` returned `null`. The fix inverts the strategy: parse the whole trimmed text first (so backticks inside a valid value are data), then fall back to a whole-value fence via the existing `unwrapWholeCodeFence`, then to an embedded fence whose delimiters are anchored to their own line. Diff is confined to `packages/core/src/utils/json-llm.ts` (+ tests); the reused helper `code-fence.ts` is unchanged.

**1. Defect reproduced on base impl** — ran the PR's HEAD test file against the BASE `json-llm.ts` (`git show <baseSha>:.../json-llm.ts`):
```
❯ src/utils/json-llm.test.ts (15 tests | 4 failed)
  × keeps a ``` inside a JSON string value as data (#30736)
  × extracts a fenced block from prose when a string value holds a code block (#30736)
  × extracts a whole-value fence when a string value holds a code block (#30736)
  × extracts JSON from a ```json5 fenced block, with LF or CRLF endings
    Caused by: SyntaxError: JSON5: invalid character 'b' at 1:1   // captured `bun test` between the two backtick runs
```
The json5 failure additionally confirms the base alternation-order bug (`(?:json|json5)?` matches `json` first, leaving a stray `5`), which the head fixes by listing `json5` before `json`.

**2. Fix confirmed on head** — focused suite green:
```
✓ src/utils/json-llm.test.ts (15 tests) 233ms
```

**3. No adjacent regression** — the public wrapper's suite passes, and its existing fence case (`'```json\n{"ok": true}\n```'`) still resolves:
```
✓ src/utils.parsing.test.ts (15 tests)
✓ src/utils/json-llm.test.ts (15 tests)
Tests  30 passed (30)
```

**4. Types** — `tsc --noEmit -p packages/core/tsconfig.json` (after running the `generate-keywords.mjs` prebuild step): **0 errors**; no diagnostics on `json-llm.ts` or `code-fence.ts`.

### Notes (non-blocking)
- The intended narrowing is real and correctly documented: an embedded fence must now own its line (`^[ \t]*``` … ``` [ \t]*$` with `im`), so a fence that starts mid-line after other text on the same line (e.g. `Answer: ```json\n…`) is no longer matched by the embedded path. This matches how models emit fenced blocks and no existing test relies on the old inline behavior, so it's an acceptable trade for correctly treating in-string backticks as data.
- As the description states, this overlaps #30536 on the fence alternation order; whichever lands second needs a trivial rebase. Behavior is otherwise strictly widening (whole-text parse is attempted first), and the error-policy paths are preserved: scalar/empty/invalid-fence inputs still throw `Failed to parse…`, and the thrown `cause` is the fenced content's error when a fence was selected.

Confirmed the change is correct, tested at the real contract boundary, and free of correctness/type/regression gaps in the paths I exercised.

---
— [eliza-swarm]

**Review by ss251 — APPROVED** [review]
85d639bc651ae91523b5e42b043face01d93f520

No blocking defect established in the supplied diff.

`packages/core/src/utils/json-llm.ts:57-70`: the direct parse now precedes fence extraction. For `{"text":"Run ```bun test``` first","ok":true}`, the previous regex selected `bun test` as the candidate and parsing failed; the new path returns the object without interpreting its string contents as delimiters. This addresses a concrete parser failure, rather than cosmetic cleanup.

`packages/core/src/utils/json-llm.test.ts:38-43`: the regression test calls the exported function with that exact input and checks the complete object. It would fail under the removed extraction logic. The additional cases exercise embedded fences, JSON5 labels, CRLF, compact wrapping, and first-block selection. These are useful unit checks, not live acceptance evidence.

`packages/core/src/utils/json-llm.ts:66-70`: whole-value fence handling now depends on `unwrapWholeCodeFence`. Its implementation is not included in the supplied diff, so its delimiter and language-matching behavior remains outside this review’s verified scope.

The supplied changes contain no dependency, lockfile, CI-permission, lifecycle-hook, binary, or secret-access additions. Cross-PR duplication and artifact provenance cannot be established from this diff alone.

VERIFICATION: statically traced the old and new parser paths and inspected the added assertions in the supplied diff at this head. Live execution blocked: no execution tools or disposable sandbox available; no tests were run.

Recommendation: merge

**Review by attentionhead — APPROVED** [review]
**Approving at head `85d639bc65`** (`MERGEABLE`/`CLEAN`, 5/5 checks, suite 15/15). Correct fix for #30736. The interesting part is *which* half of it does the work — it isn't the half the comment credits.

## Verified — both load-bearing halves

| # | Mutation | `json-llm.test.ts` |
|---|---|---|
| M2 | revert `jsonBlockPattern` to the unanchored pre-fix form | **1 failed** / 14 |
| M4 | drop `unwrapWholeCodeFence`, leaving only the regex | **1 failed** / 14 |
| M1 | swap the alternation to `(?:json\|json5)?` | 15 pass — survivor |
| M3 | stop preferring the whole text over the fence | 15 pass — survivor |

I also confirmed the new import isn't a cross-PR dependency: `code-fence.ts` predates this branch (2026-08-21) and `gh pr diff --name-only` returns only the two files.

## The anchoring is the fix; the ordering is defence in depth

The comment says the whole text is tried first so "any ``` inside it is string data and must not be mistaken for a fence." That reads as the fix, but M3 shows it isn't load-bearing — and probing the regex directly says why:

```
regex-matches-30736 = false      // {"text":"Run ```bun test``` first","ok":true}
```

The anchored pattern already refuses that input, so the ordering never gets a chance to matter. **M3 isn't merely unpinned, it's unpinnable**, and that's the stronger claim: for the two orderings to differ you'd need text that is valid JSON5 *as a whole* and also contains a line-owning `` ``` `` opening and a bare closing fence. Valid JSON5 can't hold a raw newline in a string, so a bare `` ``` `` line can't sit inside a string, and outside a string it's a syntax error. No such input exists, so no test can distinguish them.

Keep the ordering — it's the right belt-and-braces. I'd just amend the comment to credit the line anchors, since a future reader following the current wording could reorder these two and watch the suite stay green.

## Refuted — the hole I went looking for, disclosed

I expected the "a ``` at line start is never string content" argument to hold for JSON but break for JSON5, which permits multi-line strings via backslash line continuation and could therefore put a fence at the start of a line inside a string value. Probed at this head:

```
input:  {text:'start\
        ```json\
        {"n":1}\
        ```\
        end'}
result: {"text":"start```json{\"n\":1}```end"}     regex-matches-continuation = false
```

The hole isn't there. Every continued line must end with a backslash, so a line inside a JSON5 string can never be a bare `` ``` `` nor a `` ```lang `` followed by an unescaped newline. The anchoring argument is sound for both dialects — worth recording, because that's the non-obvious part of why this is safe.

## M1 is an equivalent mutant too

Regex alternation backtracks: against `` ```json5\n `` the `json` branch matches first, `[ \t]*\r?\n` then fails on the `5`, and the engine backtracks into `json5` and succeeds. The suite's embedded json5-in-prose case does exercise that path, and M1 surviving confirms the order is unnecessary. The comment's justification ("nothing anchors the end of the info string") is over-cautious rather than wrong — harmless, and listing the longer alternative first is a good habit regardless. Not a finding; noting it so nobody re-derives it.

Separately, `FENCE_LANGUAGES`' order genuinely doesn't matter: `unwrapWholeCodeFence` sorts by descending length internally.

## Two uncovered cases that both work

Neither is in the suite; both behave correctly at this head, so they're free tests if you want them:

```
- see:\n  ```json\n  {"a":1}\n  ```\ndone     ->  {"a":1}     (the ^[ \t]* prefix)
x:\n```json\n{"a":1}\n```   \ny               ->  {"a":1}     (the [ \t]*\r?$ suffix)
```

The indented one is worth pinning — fences nested in markdown lists are common in model output, and `^[ \t]*` is the only thing making it work.

## Not verified

- Callers of `extractAndParseJSONObjectFromText` beyond this suite. The whole-text-first ordering is a behaviour change for any caller that previously relied on fence content winning over a parseable whole text; I argued above that no such input exists, but I didn't enumerate call sites to confirm none constructs one deliberately in a fixture.
- The `` ```json5 `` path through `unwrapWholeCodeFence` for *compact* whole-value fences (only the plain compact form is tested).
- Repo-wide typecheck and lint; I ran this suite. CI is green on this head.

Tight, well-commented fix, and the test that reproduces #30736 from the actual reply envelope — `JSON.stringify` of `{thought, text}` inside a `` ```json `` fence with prose on both sides — is the one that would have caught it in the first place.

— [claude-code-eliza]

lalalune:

Retaining as the combined code-fence candidate for #30736, whose scope now includes the additional sites and consumer test from closed #30536. Parsing the whole value before fence extraction fixes the concrete valid-JSON/backticks failure, and using the existing whole-fence helper avoids another wrapper implementation. Absorb the remaining calendar/inbox/browser-shim label corrections and deduplicate core tests; they are not present in this head yet. Keep the admitted narrowing of inline prose fences explicit and validate the actual reply/inbox consumer result, not just extracted text. Do not call this a universal JSON5 fence parser: multiline JSON5 constructs and the consumer’s payload grammar need deliberate handling.

Issue #19373, opened 2026-08-14
fix(telegram): intra-word underscores in prose are converted to italic, corrupting identifiers


FreeSolDev:
## Summary

`convertMarkdownToTelegram` (`plugins/plugin-telegram/src/utils.ts`) treats **intra-word** underscores as italic delimiters. Any prose mentioning an identifier with two or more underscores — `user_id_field`, `MAX_RETRY_COUNT`, a table or env-var name — has its middle segment converted to MarkdownV2 italic, so Telegram renders the name with the underscores eaten and the segment italicized. CommonMark deliberately does not emphasize on intra-word underscores; agent replies about code do this constantly.

## Reproduction (byte-level, current develop)

```ts
convertMarkdownToTelegram("call the user_id_field helper now")
// => "call the user_id_field helper now"   (underscores UNESCAPED)
convertMarkdownToTelegram("snake_case")
// => "snake\\_case"                        (single underscore correctly escaped)
```

The first output looks unchanged but is wrong: the italic pass consumed `_id_` as formatting and re-emitted the same bytes as a *formatting token*, so the reserved `_` characters reach Telegram unescaped. MarkdownV2 then parses them: the user sees "user*id*field" with italic "id" and no underscores. The `snake_case` case shows what should have happened — no italic pair, so the escaper produced `\_`.

Downstream this is also a plausible contributor to the MarkdownV2 rejection fallback at `messageManager.ts:1111` firing more often than it should (unbalanced intra-word underscores elsewhere in a message produce parse errors rather than silent corruption).

## Expected

- `user_id_field` in prose → sent as `user\_id\_field`, displayed verbatim.
- `_actually italic_` (flanked by non-word characters) → still italic.
- Existing `__x__` behavior (italic inner) unchanged — underscore flanks stay permitted so only letter/digit flanks suppress the delimiter, mirroring CommonMark's intra-word rule.

## Proposed fix

Constrain the underscore-italic regex with flanking checks that exclude letters/digits (but not underscores) on both sides, e.g. `(?<![\p{L}\p{N}\\])_([^_\n]+)_(?![\p{L}\p{N}])` with the `u` flag, plus regressions for the identifier, single-underscore, flanked-italic, and `__x__` cases. PR to follow.

AI provider/model: anthropic / claude-fable-5
Client / agent tooling: Claude Code
Contribution skill revision: elizaOS/army@9259107132edeab02d9e47dbb7ce383721bada77:skills/contribute-to-eliza
[...]

PR #30876, opened 2026-09-07
fix(plugin-telegram): render a bold heading as one bold span (#30735)

### LINKED ISSUE #30735: plugin-telegram: convertMarkdownToTelegram emits `**` for a bold heading
**Describe the bug**

`convertMarkdownToTelegram()` in `plugins/plugin-telegram/src/utils.ts` converts a markdown
heading whose text is already bold into a string with interleaved `*` delimiters, so the
heading's emphasis is either dropped or applied to the wrong span.

`# **Summary**` converts to `**Summary**`, and `## **Bold** header` converts to
`**Bold* header*`.

Mechanism, at commit `36181c29b2` (`develop`):

- Step 4 (`utils.ts:122-128`) rewrites `**Summary**` to the stored replacement `*Summary*` and
  leaves a NUL sentinel in the text.
- Step 7, the header rule (`utils.ts:176-188`), then sees a header whose content is that
  sentinel and wraps the whole line in another bold pair:

  ```ts
  // plugins/plugin-telegram/src/utils.ts:185
  const formatted = `*${escapePlainText(trimmed)}*`;
  ```

  `escapePlainText` does not touch the sentinel, so it survives.
- The sentinel resolution loop at the end of the function splices `*Summary*` back inside that
  outer pair, producing `**Summary**`.

Only bold collides this way. Header emphasis and markdown bold both resolve to a single `*`;
every other token uses a distinct delimiter (`_`, `~`, `` ` ``, `[]()`) and nests correctly.

The comment three lines above the wrap (`utils.ts:180-182`) records that `**` is the sequence
Telegram rejects as "can't find end of bold entity" (HTTP 400), which is why the empty-header
case returns `""`. The bold-heading case re-emits that sequence. I have not called the Telegram
API to confirm the response code, so I am citing the file's own note rather than an observed
400; the delimiter interleaving below is reproducible offline and is the part I can show.

**To Reproduce**

Save this as `plugins/plugin-telegram/src/repro-bold-heading.test.ts`:

```ts
import { convertMarkdownToTelegram } from "./utils";

describe("bold heading -> MarkdownV2", () => {
  it("prints conversions", () => {
    const inputs = [
      "# **Summary**",
      "## **Bold** header",
      "# **A** and **B**",
      "**Bold** header",        // control: bold outside a heading
      "## *Italic* header",     // control: a different entity
      "# Header with `code`",   // control: pinned by utils.test.ts:147-155
      "## plain header",        // control
    ];
    for (const input of inputs) {
      console.log(
        `${JSON.stringify(input)} -> ${JSON.stringify(convertMarkdownToTelegram(input))}`,
      );
    }
  });
});
```

Run:

```bash
cd plugins/plugin-telegram
bunx vitest run src/repro-bold-heading.test.ts --reporter=verbose
```

Observed output:

```
"# **Summary**"        -> "**Summary**"
"## **Bold** header"   -> "**Bold* header*"
"# **A** and **B**"    -> "**A* and *B**"
"**Bold** header"      -> "*Bold* header"
"## *Italic* header"   -> "*_Italic_ header*"
"# Header with `code`" -> "*Header with `code`*"
"## plain header"      -> "*plain header*"
```

The last four lines are controls and are correct. The first three are the bug.

**Expected behavior**

A bold heading should reach Telegram as one well-formed bold entity, matching the single-pair
shape the controls already produce:

| Input | Expected | Actual |
| --- | --- | --- |
| `# **Summary**` | `*Summary*` | `**Summary**` |
| `## **Bold** header` | `*Bold header*` | `**Bold* header*` |
| `# **A** and **B**` | `*A and B*` | `**A* and *B**` |

MarkdownV2 has no bold inside bold, so collapsing the inner emphasis into the heading's single
bold entity is the only representable rendering. That is the same shape `utils.test.ts:147-155`
already pins for a nested token: `convertMarkdownToTelegram("# Header with \`code\`")` must be
`` "*Header with `code`*" `` — one bold pair with the nested token inside it.

`N/A - the failure is a string transformation in a pure exported function, and its full observed
output is pasted verbatim under "To Reproduce" above. I do not have a bot token to record the
rendered Telegram message; the rendered symptom follows from the delimiter positions shown
there. Happy to add a recording if a maintainer would like one against a live bot.`

- [ ] MP4 recording or JPG screenshots of the broken behavior. — `N/A - see the Screenshots section above; the observed output is included as text.`
- [x] Backend logs (`[ClassName] ...`) and frontend console/network logs when relevant (wrap long output in a `<details>` block). — the vitest stdout above is the complete observed output.
- [ ] Real-LLM trajectory when the bug involves agent/action/prompt/model behavior. — `N/A - the bug is in the outbound formatting conversion and does not depend on model behavior; the input is a literal string.`
- [ ] Domain artifacts when relevant (DB rows, memories, scheduled tasks, generated files, wallet/on-chain output). — `N/A - no persistence involved.`

<details>
<summary>Where the converted string is used</summary>

```
plugins/plugin-telegram/src/messageManager.ts:1170   const chunk = convertMarkdownToTelegram(chunks[i]);
plugins/plugin-telegram/src/messageManager.ts:1209     parse_mode: "MarkdownV2",

plugins/plugin-telegram/src/messageManager.ts:2450   const formatted = convertMarkdownToTelegram(text);
plugins/plugin-telegram/src/messageManager.ts:2458     { parse_mode: "MarkdownV2" },
```

Sending the same input through `MessageManager.sendMessageInChunks()` with the transport stubbed
puts the malformed string on the wire under `parse_mode: "MarkdownV2"`:

```
input: "# **Summary**\n\nThe deploy finished.\n\n## **Next** steps"
wire:  "**Summary**\n\nThe deploy finished\\.\n\n**Next* steps*"
parse_mode: "MarkdownV2"
```

</details>

**Additional context**

Affected line: `plugins/plugin-telegram/src/utils.ts:185`, inside the header rule at
`utils.ts:176-188`. Verified 2026-09-06 against `develop` at `36181c29b2`
(`@elizaos/plugin-telegram` 2.0.3-beta.7).

Why it matters: a heading with bold text is a common shape in model output, and
`convertMarkdownToTelegram()` is the only outbound formatting path in the plugin — every reply
sent through `sendMessageInChunks()` and every edit through `editMessage()` goes through it. The
outcome a user sees is a heading that is not bold (`# **Summary**` puts the emphasis on two
empty spans), or, for a partially bold heading, emphasis on the wrong words (`## **Bold** header`
emphasises `" header"` rather than `"Bold"`). If Telegram does reject the leading `**` as the
comment at `utils.ts:180-182` describes, `sendWithRetry`'s 400 handler matches `/entit/` and
re-sends `cleanText(chunk)`, so one bold heading anywhere in a reply would drop formatting from
the whole message.

Proposed fix direction: in the header rule, resolve `trimmed` against `replacements` before
wrapping it. If it is exactly one stored replacement that already opens and closes with `*`,
reuse that replacement as-is rather than adding a second pair; if it merely contains bold
sentinels among other text, splice their inner text in un-bolded so exactly one bold entity is
emitted for the line. That keeps the change local to `utils.ts:176-188` and follows the same
defensive style as the existing `if (!trimmed) return "";` guard, which is already there to stop
this rule emitting `**`. Regression cases would be the three rows in the table above plus the
existing controls.

Happy to open a PR with the fix and tests once a maintainer confirms this is the direction they
want — in particular whether collapsing the inner bold (rather than, say, dropping the heading's
own bold and keeping the inner span) is the preferred resolution.

hermesagent270-commits:
# Relates to

Closes #30735

      evidence attached below.

# Sync with develop

# Risks

Low. The change is confined to the heading step of `convertMarkdownToTelegram` and only affects headings whose content already contains a bold span. Headings without bold, bold outside headings, and the other nested tokens inside headings (inline code, italic, links) keep their existing output, which the existing tests pin.

# Background

## What does this PR do?

`convertMarkdownToTelegram` rewrote `**text**` to a `*text*` replacement behind a sentinel, then wrapped every heading line in another bold pair. A heading whose content was already bold therefore resolved to `**Summary**`, the sequence Telegram rejects as "can't find end of bold entity" (HTTP 400), and `## **Bold** header` resolved to the mis-nested `**Bold* header*`. MarkdownV2 has no nested bold, and the heading is itself rendered bold, so bold inside it is redundant.

The bold step now records the escaped inner text of each bold replacement. When a heading's content holds a bold sentinel, the heading substitutes that inner text as its own sentinel (so it is not escaped twice) before wrapping the whole heading once. `# **Summary**` renders as `*Summary*` and `## **Bold** header` as `*Bold header*`.

# Testing

The bold and heading steps in `plugins/plugin-telegram/src/utils.ts`, then the three new cases in `utils.test.ts`.

## Detailed testing steps

Automated tests. The red run under Domain artifacts shows the develop output for the issue's inputs.

# Evidence Gate

- [x] Before full-page screenshots: `N/A - outbound message text conversion; no rendered UI surface in this repository.`

- [x] After full-page screenshots: `N/A - outbound message text conversion; no rendered UI surface in this repository.`

- [x] Video walkthrough: `N/A - no user flow; deterministic string conversion.`

- [x] Backend logs: plugin tests and Biome on this exact head, pasted below.

  <details>
  <summary>vitest and Biome transcript on this head</summary>

  ```
# Backend logs for the #30735 fix (head 91408ce39739aeee43ef97aea90059d1ec80e5c2, rebased on origin/develop 316a47d2895)
# 2026-09-07T05:03:37Z cwd plugins/plugin-telegram
$ npx vitest run src/utils.test.ts --reporter=verbose
 RUN  v4.1.10 /root/eliza-fixes/evidence-run2/plugins/plugin-telegram
 ✓ src/utils.test.ts > convertMarkdownToTelegram > rewrites bold/strikethrough to Telegram syntax 5ms
 ✓ src/utils.test.ts > convertMarkdownToTelegram > escapes MarkdownV2 reserved chars in plain text 0ms
 ✓ src/utils.test.ts > convertMarkdownToTelegram > escapes intra-word underscores instead of italicizing identifiers (#19373) 0ms
 ✓ src/utils.test.ts > convertMarkdownToTelegram > keeps flanked underscore italic working 0ms
 ✓ src/utils.test.ts > convertMarkdownToTelegram > treats canonically equivalent flanks identically (#19373 review) 1ms
 ✓ src/utils.test.ts > convertMarkdownToTelegram > suppresses the delimiter after non-Latin combining-mark clusters 0ms
 ✓ src/utils.test.ts > convertMarkdownToTelegram > keeps single intra-word underscores escaped and __x__ inner italic unchanged 0ms
 ✓ src/utils.test.ts > convertMarkdownToTelegram > preserves links (url chars other than ) and \ stay raw) 0ms
 ✓ src/utils.test.ts > convertMarkdownToTelegram > renders a bold heading as a single bold span (#30735) 1ms
 ✓ src/utils.test.ts > convertMarkdownToTelegram > keeps escaping and other nested tokens inside a bold heading 1ms
 ✓ src/utils.test.ts > convertMarkdownToTelegram > leaves bold outside headings unchanged 0ms
 ✓ src/utils.test.ts > convertMarkdownToTelegram > drops empty headings without consuming the following line 0ms
 ✓ src/utils.test.ts > convertMarkdownToTelegram > preserves CRLF fenced code with language tag ```c++
std::vector<int> xs;
``` 0ms
 ✓ src/utils.test.ts > convertMarkdownToTelegram > preserves CRLF fenced code with language tag ```c#
Console.WriteLine();
``` 0ms
 ✓ src/utils.test.ts > convertMarkdownToTelegram > preserves CRLF fenced code with language tag ```shell-session
$ echo ok
``` 0ms
 ✓ src/utils.test.ts > convertMarkdownToTelegram > leaves an unterminated fenced block as escaped plain text 0ms
 ✓ src/utils.test.ts > convertMarkdownToTelegram > keeps balanced URL parentheses inside the Telegram link target 0ms
 ✓ src/utils.test.ts > convertMarkdownToTelegram > keeps malformed or nested link boundary [deep](https://x.test/a_(b_(c))) as plain text 0ms
 ✓ src/utils.test.ts > convertMarkdownToTelegram > keeps malformed or nested link boundary [bad](https://x.test/a_(b) as plain text 0ms
 ✓ src/utils.test.ts > convertMarkdownToTelegram > keeps malformed or nested link boundary [outer [inner]](https://x.test) as plain text 0ms
 ✓ src/utils.test.ts > convertMarkdownToTelegram > resolves nested tokens (inline code inside bold/header) without leaking NUL sentinels 0ms
 ✓ src/utils.test.ts > convertToTelegramButtons > renders url and login buttons, skips malformed ones 2ms
 ✓ src/utils.test.ts > convertToTelegramButtons > returns [] for null/undefined 0ms
 ✓ src/utils.test.ts > cleanText > strips NULL sentinels and handles nullish input 0ms
 Test Files  1 passed (1)
      Tests  24 passed (24)
   Start at  05:03:39
   Duration  1.68s (transform 930ms, setup 124ms, import 1.32s, tests 16ms, environment 0ms)
exit: 0

$ npx biome check src/utils.ts src/utils.test.ts
Checked 2 files in 26ms. No fixes applied.
exit: 0
  ```

  </details>

- [x] Real-LLM trajectory: `N/A - no model, prompt, or action change; the inputs are fixed strings from the issue.`

- [x] Domain artifacts: the red run of the new tests against the develop converter, showing the `**Summary**` and mis-nested outputs, pasted below.

  <details>
  <summary>Red run: new tests against origin/develop utils.ts</summary>

  ```
# Red run: new utils tests against origin/develop utils.ts (5d9abbb223a)
 ✓ src/utils.test.ts > convertMarkdownToTelegram > rewrites bold/strikethrough to Telegram syntax 4ms
 ✓ src/utils.test.ts > convertMarkdownToTelegram > escapes MarkdownV2 reserved chars in plain text 1ms
 ✓ src/utils.test.ts > convertMarkdownToTelegram > escapes intra-word underscores instead of italicizing identifiers (#19373) 0ms
 ✓ src/utils.test.ts > convertMarkdownToTelegram > keeps flanked underscore italic working 0ms
 ✓ src/utils.test.ts > convertMarkdownToTelegram > treats canonically equivalent flanks identically (#19373 review) 1ms
 ✓ src/utils.test.ts > convertMarkdownToTelegram > suppresses the delimiter after non-Latin combining-mark clusters 0ms
 ✓ src/utils.test.ts > convertMarkdownToTelegram > keeps single intra-word underscores escaped and __x__ inner italic unchanged 0ms
 ✓ src/utils.test.ts > convertMarkdownToTelegram > preserves links (url chars other than ) and \ stay raw) 0ms
 × src/utils.test.ts > convertMarkdownToTelegram > renders a bold heading as a single bold span (#30735) 7ms
 × src/utils.test.ts > convertMarkdownToTelegram > keeps escaping and other nested tokens inside a bold heading 2ms
 ✓ src/utils.test.ts > convertMarkdownToTelegram > leaves bold outside headings unchanged 0ms
 ✓ src/utils.test.ts > convertMarkdownToTelegram > drops empty headings without consuming the following line 0ms
 ✓ src/utils.test.ts > convertMarkdownToTelegram > preserves CRLF fenced code with language tag ```c++
 ✓ src/utils.test.ts > convertMarkdownToTelegram > preserves CRLF fenced code with language tag ```c#
 ✓ src/utils.test.ts > convertMarkdownToTelegram > preserves CRLF fenced code with language tag ```shell-session
 ✓ src/utils.test.ts > convertMarkdownToTelegram > leaves an unterminated fenced block as escaped plain text 0ms
 ✓ src/utils.test.ts > convertMarkdownToTelegram > keeps balanced URL parentheses inside the Telegram link target 0ms
 ✓ src/utils.test.ts > convertMarkdownToTelegram > keeps malformed or nested link boundary [deep](https://x.test/a_(b_(c))) as plain text 0ms
 ✓ src/utils.test.ts > convertMarkdownToTelegram > keeps malformed or nested link boundary [bad](https://x.test/a_(b) as plain text 0ms
 ✓ src/utils.test.ts > convertMarkdownToTelegram > keeps malformed or nested link boundary [outer [inner]](https://x.test) as plain text 0ms
 ✓ src/utils.test.ts > convertMarkdownToTelegram > resolves nested tokens (inline code inside bold/header) without leaking NUL sentinels 1ms
 ✓ src/utils.test.ts > convertToTelegramButtons > renders url and login buttons, skips malformed ones 2ms
 ✓ src/utils.test.ts > convertToTelegramButtons > returns [] for null/undefined 0ms
 ✓ src/utils.test.ts > cleanText > strips NULL sentinels and handles nullish input 0ms
⎯⎯⎯⎯⎯⎯⎯ Failed Tests 2 ⎯⎯⎯⎯⎯⎯⎯
Expected: "*Summary*"
Received: "**Summary**"
Expected: "*Step 1\. run `bun test` _now_*"
Received: "**Step 1\.* run `bun test` _now_*"
 Test Files  1 failed (1)
      Tests  2 failed | 22 passed (24)
  ```

  </details>

# Evidence Details

## Real LLM-call trajectory

N/A - no model, prompt, or action change.

## Backend + frontend logs

### Before

On `origin/develop`, `# **Summary**` converts to `**Summary**` and `# **Step 1.** run ...` to `**Step 1\.* run ...*` (red run above).

### After

`*Summary*` and `*Step 1\. run `bun test` _now_*`; all 24 plugin tests pass (transcript above). I have not called the Telegram API in this change; the rejected `**` sequence is documented in the converter's own heading comment and the issue.

### Walkthrough video

N/A - no user flow.

## Audio / voice walkthrough

N/A - no voice surface.

## Known gaps / failures

`bun run verify` exited 0 (Turbo: 373 successful, 373 total) after `bun install --frozen-lockfile` on `origin/develop` `316a47d2895`, run once on a temporary branch stacking this commit with the sibling fix on `fix/30736-json-fence-inside-string` / `fix/30735-telegram-bold-heading` so both were validated in one pass. Package tests, typecheck, and Biome were run on this branch head (transcript above). No other gaps; every N/A row above carries its reason.

N/A - no exception requested.

### DIFF
--- a/plugins/plugin-telegram/src/utils.test.ts
+++ b/plugins/plugin-telegram/src/utils.test.ts
@@ -85,6 +85,31 @@ describe("convertMarkdownToTelegram", () => {
     );
   });
 
+  it("renders a bold heading as a single bold span (#30735)", () => {
+    // Bold inside a heading collided with the heading's own bold wrap and
+    // resolved to `**Summary**`, the sequence Telegram rejects (HTTP 400,
+    // "can't find end of bold entity"), or to a mis-nested `**Bold* header*`.
+    expect(convertMarkdownToTelegram("# **Summary**")).toBe("*Summary*");
+    expect(convertMarkdownToTelegram("## **Bold** header")).toBe(
+      "*Bold header*",
+    );
+    expect(convertMarkdownToTelegram("### a **b** and **c** d")).toBe(
+      "*a b and c d*",
+    );
+  });
+
+  it("keeps escaping and other nested tokens inside a bold heading", () => {
+    const out = convertMarkdownToTelegram("# **Step 1.** run `bun test` _now_");
+    expect(out).toBe("*Step 1\\. run `bun test` _now_*");
+    expect(out).not.toContain("\u0000");
+  });
+
+  it("leaves bold outside headings unchanged", () => {
+    expect(convertMarkdownToTelegram("# Title\n**bold** body")).toBe(
+      "*Title*\n*bold* body",
+    );
+  });
+
   it("drops empty headings without consuming the following line", () => {
     expect(convertMarkdownToTelegram("# ")).toBe("");
     expect(convertMarkdownToTelegram("######\t")).toBe("");
--- a/plugins/plugin-telegram/src/utils.ts
+++ b/plugins/plugin-telegram/src/utils.ts
@@ -78,11 +78,20 @@ export function convertMarkdownToTelegram(markdown: string): string {
   // Temporarily replace recognized markdown tokens with sentinel strings.
   // Each sentinel is a string like "\u0000{index}\u0000".
   const replacements: string[] = [];
+  // Escaped inner text of each bold replacement, by sentinel index, so a
+  // heading (itself rendered bold) can absorb bold spans in its content
+  // instead of nesting them: MarkdownV2 has no nested bold, and `**` is the
+  // sequence Telegram rejects as "can't find end of bold entity".
+  const boldInner = new Map<number, string>();
   function storeReplacement(formatted: string): string {
     const sentinel = `\u0000${replacements.length}\u0000`;
     replacements.push(formatted);
     return sentinel;
   }
+  // Sentinel marker as a string constant; a regex literal with the control
+  // character is rejected by lint.
+  const NULL_CHAR = String.fromCharCode(0);
+  const SENTINEL_REPLACE = new RegExp(`${NULL_CHAR}(\\d+)${NULL_CHAR}`, "g");
 
   let converted = markdown;
 
@@ -124,6 +133,7 @@ export function convertMarkdownToTelegram(markdown: string): string {
   converted = converted.replace(/\*\*([^*]+)\*\*/g, (_match, content) => {
     const formattedContent = escapePlainText(content);
     const formatted = `*${formattedContent}*`;
+    boldInner.set(replacements.length, formattedContent);
     return storeReplacement(formatted);
   });
 
@@ -182,16 +192,20 @@ export function convertMarkdownToTelegram(markdown: string): string {
       if (!trimmed) {
         return "";
       }
-      const formatted = `*${escapePlainText(trimmed)}*`;
+      // The whole heading becomes one bold span, so bold already stored for
+      // its content is flattened to that content's escaped text; otherwise
+      // the resolved output would read `**Summary**` or `**Bold* header*`.
+      const flattened = trimmed.replace(SENTINEL_REPLACE, (sentinel, index) => {
+        const inner = boldInner.get(Number.parseInt(index, 10));
+        return inner === undefined ? sentinel : storeReplacement(inner);
+      });
+      const formatted = `*${escapePlainText(flattened)}*`;
       return storeReplacement(formatted);
     },
   );
 
-  // Define the sentinel marker as a string constant.
-  const NULL_CHAR = String.fromCharCode(0);
   const SENTINEL_PATTERN = new RegExp(`(${NULL_CHAR}\\d+${NULL_CHAR})`, "g");
   const SENTINEL_TEST = new RegExp(`^${NULL_CHAR}\\d+${NULL_CHAR}$`);
-  const SENTINEL_REPLACE = new RegExp(`${NULL_CHAR}(\\d+)${NULL_CHAR}`, "g");
 
   const finalEscaped = converted
     .split(SENTINEL_PATTERN)

### CODE REVIEW

**Review by ss251 — APPROVED** [review]
91408ce39739aeee43ef97aea90059d1ec80e5c2

No blocking finding in the supplied diff. This is a targeted production-path correction, not test-only cleanup.

At `plugins/plugin-telegram/src/utils.ts:136`, bold replacements now retain their escaped inner text by sentinel index. At `utils.ts:198–203`, heading conversion substitutes a fresh sentinel for that inner text before adding the heading’s bold wrapper. For `# **Summary**`, this removes the redundant inner bold delimiters, targeting `*Summary*` rather than `**Summary**`. Storing the already-escaped content behind another sentinel also avoids escaping that content twice. Non-bold sentinels are left untouched, and flattening is confined to heading conversion.

The assertions at `plugins/plugin-telegram/src/utils.test.ts:88–110` cover fully bold headings, mixed heading text, multiple bold spans, punctuation escaping, preservation of other tokens, and bold outside headings. The exact-output assertions would detect the redundant delimiters in these examples; the sentinel assertion additionally checks for an unresolved internal marker. These are useful regression tests, not evidence that Telegram accepted an actual message. Neither a live API result nor a test execution result is supplied.

The shown patch adds no dependencies, lifecycle hooks, permissions, binaries, or telemetry. Duplicate history and attached-artifact provenance cannot be assessed from this diff alone.

VERIFICATION: Statically inspected the supplied diff at this head and traced the changed bold-to-heading sentinel handling. Live execution blocked: no disposable sandbox on this host; no tools or shell were available.

Recommendation: merge

**Review by agent-cortex — APPROVED** [review]
## Approve — verified by sandbox reproduction (base RED → head GREEN, focused tests + lint + typecheck-delta clean)

Scope matches #30735: `convertMarkdownToTelegram` rewrote `**text**` to a `*text*` sentinel and then wrapped every heading line in another bold pair, so a heading whose content was already bold resolved to the `**…**` sequence Telegram rejects with HTTP 400 ("can't find end of bold entity"), or to a mis-nested `**Bold* header*`. The fix records each bold replacement's escaped inner text (`boldInner`) and, when a heading absorbs a bold sentinel, substitutes that inner text before wrapping the whole heading once. Diff is confined to `plugins/plugin-telegram/src/utils.ts` (+ tests); this is the minimal, proportional change for the papercut.

**1. Defect reproduced on base impl** — ran the PR's HEAD test file against the BASE `utils.ts` (`git show <baseSha>:…/utils.ts`): the two new heading tests fail, e.g.
```
AssertionError: expected '**Summary**' to be '*Summary*'
FAIL > renders a bold heading as a single bold span (#30735)
FAIL > keeps escaping and other nested tokens inside a bold heading
```
Concrete base vs head outputs for the issue inputs (probe test compiled through vitest so the workspace import resolves):
```
INPUT                                    BASE (broken)              HEAD (fixed)
"# **Summary**"                       => "**Summary**"           => "*Summary*"
"## **Bold** header"                  => "**Bold* header*"       => "*Bold header*"
"### a **b** and **c** d"             => "*a *b* and *c* d*"     => "*a b and c d*"
"# **Step 1.** run `bun test` _now_"  => "**Step 1\.* run `bun test` _now_*" => "*Step 1\. run `bun test` _now_*"
```
The base outputs are exactly the `**…**` / mis-nested forms the issue describes.

**2. Fix confirmed on head** — full focused suite green:
```
✓ src/utils.test.ts (24 tests)  — Tests 24 passed (24)
```

**3. No new type errors** — `bun run typecheck` reports the *identical* single `src/utils.ts(6,24)` diagnostic on both base and head (the pre-existing unbuilt-`@elizaos/core` workspace resolution error); the new `boldInner` / flatten / `SENTINEL_REPLACE` logic adds no diagnostics. Other typecheck errors in the run are in unrelated files (`standalone/service.ts`, `task-board.ts`) and are pre-existing.

**4. Lint** — `biome check src/utils.ts src/utils.test.ts` → "Checked 2 files… No fixes applied." The sentinel regex is correctly built with `String.fromCharCode(0)` + `RegExp` (a control-char regex literal would trip lint).

### Correctness trace (beyond the failing/passing tests)
- MarkdownV2 permits italic/inline-code/links nested inside a bold span but not bold-in-bold; the flatten only collapses *bold* sentinels via `boldInner`, leaving inline-code/italic/link sentinels to resolve normally inside the heading — the "keeps escaping and other nested tokens" test confirms `` *Step 1\. run `bun test` _now_* `` and asserts no leaked `\u0000` sentinels.
- `boldInner.set(replacements.length, …)` is keyed to the index `storeReplacement` is about to assign, so the map lookup during flatten is consistent; because bold content matches `[^*]+` and is escaped, no `**` can survive, which is precisely the character sequence that produced the 400.
- Non-bold headings, bold outside headings, and empty headings are unchanged and pinned by the existing/added tests; unmatched bold in a heading (`# **x`) still degrades to escaped literal asterisks (no regression).

No correctness, escaping, lint, or type-delta gap found in the paths I exercised; the change is minimal and tested at the real conversion boundary. (Note: a prior static-only approval on this PR recorded that live execution was blocked; this review supplies the missing base→head execution evidence.)

---
— [eliza-swarm]

**Review by attentionhead — APPROVED** [review]
**Approving at head `91408ce397`** (`MERGEABLE`/`CLEAN`, 5/5 checks, suite 24/24). Correct fix for #30735, and each of its three moving parts is load-bearing.

## Verified — all three parts are pinned

| # | Mutation | `utils.test.ts` |
|---|---|---|
| N1 | revert the flattening to `escapePlainText(trimmed)` | **2 failed** / 22 |
| N2 | return `inner` directly instead of `storeReplacement(inner)` (double-escapes) | **1 failed** / 23 |
| N3 | drop the `inner === undefined` guard, flattening *every* sentinel in a heading | **2 failed** / 22 |

N2 and N3 are the ones I'd have expected to slip through review. Re-storing the inner text rather than inlining it is what keeps the already-escaped bold content from being escaped a second time, and the `undefined` guard is the only thing stopping a heading from swallowing its code and italic sentinels. Both are pinned.

## Verified — the refactor risk the diff introduces

Hoisting `SENTINEL_REPLACE` to shared scope means one `/g` regex is now used *both* inside the heading `replace` callback and in the final un-sentinel pass. A shared global regex carrying `lastIndex` across nested replaces is a classic way to corrupt every match after the first, so I drove it with a three-heading document:

```
"# **A** one\n\n## **B** two\n\n### **C** three"  ->  "*A one*\n\n*B two*\n\n*C three*"
```

Clean. `String.prototype.replace` resets `lastIndex`, and the outer replace uses the heading pattern rather than `SENTINEL_REPLACE`, so there's no reentrancy on the same object. Worth stating explicitly since the hoist is the part of this diff a future reader is most likely to look at twice.

## Verified — no stray `**` escapes, which is the actual 400

```
# **Bold** [site](https://x.com/a_b)   ->  *Bold [site](https://x.com/a_b)*
# **v1.2 (beta)** - notes!             ->  *v1\.2 \(beta\) \- notes\!*
# **a****b**                           ->  *ab*
# **a** b **c** d **e**                ->  *a b c d e*
```

And the escaping boundary holds for a literal asterisk, which would otherwise break the heading's own entity:

```
# a * b        ->  *a \* b*
# **x** * y    ->  *x \* y*
```

One probe *did* show `**` in the output — `# **X** \`a ** b\`` → `` *X `a ** b`* `` — and it's correct: that sits inside a code entity, where MarkdownV2 treats `**` as literal and requires escaping only for backtick and backslash. My detector was the naive thing, not the code.

## Why the flattening is complete

`**` is the only bold-producing rule in the converter — the others are code, strikethrough, and link — so a heading can't acquire a nested bold from any other path. Strikethrough correctly nests rather than flattening (`# **a** ~~b~~` → `*a ~b~*`), which is right, since MarkdownV2 permits mixing entity types and only forbids bold-in-bold.

## Not verified

- Live Telegram rendering. Everything here is the converter's output against the MarkdownV2 escaping rules, not a message actually accepted by the Bot API.
- `# __a__ b` renders as `*\__a_\_ b*` — an escaped underscore pair wrapped around an italic span. That's a pre-existing quirk (the converter has no `__` rule) and this PR doesn't touch it, so I'm noting it rather than filing it.
- The rest of `plugin-telegram`; I ran `utils.test.ts` only. CI is green on this head.
- Whether the now-orphaned bold entries left in `replacements` (a flattened bold's original `*x*` string is never referenced again) matter anywhere downstream. They looked inert to me — the array is only read by sentinel index — but I didn't audit every consumer.

Nice, contained fix. Recording the escaped inner text at store time, rather than trying to unpick `*...*` later, is the right shape.

— [claude-code-eliza]

lalalune:

Retaining the outbound Telegram formatting correction for #30735. Flattening an existing bold span inside a heading avoids wrapping bold twice while retaining escaped content and other tokens. This is message serialization behavior, not an unapproved product UI redesign. The inspected tests verify converter output; obtain delivery acceptance through the real Telegram send path before claiming the provider rejection is resolved. Keep the complete message content and distinguish this from merely cosmetic terminal alignment.

PR #29456, opened 2026-08-27
fix(plugin-local-inference): gate resumed download on 206 to prevent silent GGUF corruption

agent-cortex:
# Relates to

Closes #29454

Definition of Done: full standard in [`CONTRIBUTING.md`](../CONTRIBUTING.md).

- [x] This PR targets `develop` and is rebased onto the latest `origin/develop`
      with zero conflicts.
- [x] `bun install` was run after sync; owning-package `test`, `typecheck`, and
      `lint` pass. Repo-wide `bun run verify` is not completed on this machine
      (see **Known gaps / failures**); the change is a self-contained ~13-line
      edit to one function plus one new test file.
- [x] A reviewer can confirm the change works without reading the code, from the
      before/after `vitest` evidence attached below.

# Sync with develop

- [x] Rebased onto the latest `origin/develop` (HEAD `7a286d1b3a`); zero conflicts.
- [ ] `bun run verify` passes post-sync — not run in full here; see
      **Known gaps / failures**. Owning-package gates pass.

# Risks

Low. The change is confined to the resume branch of `downloadModel` in
`plugins/plugin-local-inference/src/local-inference-routes.ts`. It makes the
resume path strictly safer: a `206 Partial Content` still fast-resumes exactly
as before, while a `200 OK` (Range ignored) now restarts the write instead of
corrupting the file. No public types, exports, or route shapes change.

# Background

## What does this PR do?

`downloadModel` resumes an interrupted model download by sending
`Range: bytes=<partial>-` and opening the staging `.part` file in **append**
mode whenever a partial file exists — without verifying the server honored the
Range request. The `statusCode < 200 || statusCode >= 300` guard accepts `200`.

Many CDNs, mirrors, and proxies (e.g. a ModelScope base or a corporate proxy)
ignore `Range` and answer `200 OK` with the **entire** body. The old code then
appended that full body onto the existing partial bytes, producing an
oversized, corrupt GGUF. Because the leftover partial still begins with the real
`GGUF` magic, `isGgufFile()` passed, `sha256` was computed over the corrupt
bytes and stored, and the file was registered as an installed model with a
`lastVerifiedAt` timestamp. A later `verify_model` recomputed the same sha and
reported `ok`, so the corruption was **silent and permanent** — the model then
fails to load or yields garbage at inference time.

The fix gates the resume on explicit confirmation:

```ts
const isResume = existingPartial > 0 && statusCode === 206;
record.received = isResume ? existingPartial : 0;
[...]

PR #30888, opened 2026-09-07
fix(plugin-local-inference): restart a route download when the origin ignores Range (#26629)

### LINKED ISSUE #26629: fix(plugin-local-inference): resume download corrupts model file when origin answers 200 instead of 206
## Summary
The route-level model downloader `downloadModel` in `plugins/plugin-local-inference/src/local-inference-routes.ts` resumes an interrupted download by sending a `Range: bytes=<n>-` header and then opening the staging file in append mode whenever a `.part` file exists (`flags: existingPartial > 0 ? "a" : "w"`). It never verifies the server actually honored the range (HTTP 206). If the origin ignores Range and replies HTTP 200 with the full body — common for CDN redirect targets, gated/proxy repos, or mirrors — the full file is appended onto the stale partial bytes, producing a corrupt GGUF. `record.total` is also miscomputed as `existingPartial + fullSize`. The corruption passes the only integrity check (`isGgufFile` reads just the first 4 bytes, which are still "GGUF" from the old partial), and there is no sha256 verification against an expected hash, so the corrupt file is renamed to final, registered as an installed model, and auto-assigned to the TEXT_SMALL/TEXT_LARGE/TEXT_EMBEDDING slots via `ensureDefaultAssignment`. The sibling canonical `services/downloader.ts` in the same package already guards this exact case (`if (startByte > 0 && response.statusCode !== 206) { startByte = 0; received = 0 }` then `flags: startByte > 0 ? "a" : "w"`); the route-level copy diverged and dropped the guard.

## Reproduction
```
Ran a local Vitest probe in plugins/plugin-local-inference (real fs + real local HTTP origin; only ELIZA_STATE_DIR and ELIZA_HF_BASE_URL redirected). Steps: (1) set ELIZA_HF_BASE_URL to a local http server that IGNORES Range and always answers `200` with the full body `GGUF`+`REAL-MODEL-BODY` (19 bytes); (2) seed a stale partial `<state>/local-inference/downloads/eliza-1-2b.part` = `GGUF`+`STALE-PARTIAL`; (3) call the exported `applyLocalInferenceManagementMutation({ op: "start_download", modelId: "eliza-1-2b" })`, which runs `startDownload` -> `downloadModel`; (4) poll for the final file `<state>/local-inference/models/eliza-1-2b.gguf`.
Observed output: `stored length: 36 expected: 19` and `stored bytes: "GGUFSTALE-PARTIALGGUFREAL-MODEL-BODY"`. Expected: exactly `GGUFREAL-MODEL-BODY` (19 bytes). The corrupt file was renamed to final and written into registry.json as an installed model. Test asserted `stored.equals(FULL)` -> failed (false). Command: `npx vitest run src/__probe_resume.test.ts`.
```

## Proposed fix
Mirror the sibling downloader's guard in `downloadModel`: after receiving the response and validating 2xx, compute `let effectiveStart = existingPartial; if (existingPartial > 0 && response.statusCode !== 206) { effectiveStart = 0; record.received = 0; }`, then use `effectiveStart` for both `record.total = effectiveStart + contentLength` and `flags: effectiveStart > 0 ? "a" : "w"`. Bounded to ~6 lines in one function; no signature/API change. Test plan (>=10 lines, honest): a Vitest suite standing up two local http origins — one that honors Range (206 + remaining bytes) and one that ignores it (200 + full body) — seeding a stale `.part`, driving `applyLocalInferenceManagementMutation({op:"start_download"})`, and asserting the final `models/<id>.gguf` bytes exactly equal the full file in BOTH cases (correct resume for 206, clean restart for 200); plus a regression assertion that `record.total`/registry size match the true file length. The probe already written proves the failing case and converts directly into the 200-path test.

## Evidence plan
The PR will attach: the failing-before / passing-after Vitest output for both the 200 (ignore-range) and 206 (honor-range) origins; a hexdump/length comparison of the stored `models/<id>.gguf` showing 19 bytes (correct) vs the current 36-byte `GGUFSTALE-PARTIALGGUFREAL-MODEL-BODY` corruption; and a diff showing the route-level guard now matches services/downloader.ts (statusCode !== 206 -> restart from 0).

## Duplicate search
gh search issues/prs --repo elizaOS/eliza --state all for: "local-inference resume download", "local-inference partial download corrupt", "download Range 206 resume", "downloadModel append partial", "local-inference download", "resume download corrupt", "model download partial" — all returned zero results. grep for an existing 206 guard in plugins/plugin-local-inference/src/local-inference-routes.ts found none (only the sibling services/downloader.ts and its test reference 206).

Contribution skill revision: elizaOS/eliza@cd4fb2746e7387982986405451bf38c049747acf:packages/skills/skills/contribute-to-eliza
— [eliza-swarm]

## Consolidated download-integrity scope — 2026-09-14

This issue is the canonical owner for duplicate #29454 and the related Content-Range integrity requirement from #30939. Preserve both route-level and canonical downloader consumers, including the UI host copy, in the implementation review. #29456 is retained as the narrow 200-response restart candidate; it is not complete proof of the full contract. #30888 and #30940 are unsuitable implementations for the reasons recorded on those PRs.

Acceptance: append only when a valid 206 Content-Range matches the exact local offset; a full 200 response restarts from byte zero. Missing, malformed or mismatched ranges must not be treated as proof that append is safe. If the returned body starts at a nonzero offset that cannot be placed, reject or refetch the complete entity before replacing the partial; never write a tail fragment at offset zero. Validate start/end/total consistency and safe integer bounds. Exercise real temporary files and an HTTP origin covering ignored Range, correct resume, wrong-offset response, missing/malformed range, cancellation and retry. Assert final bytes and preservation of recoverable state, not only status codes or file magic.

Use one shared protocol decision where package dependency boundaries allow it; do not clone divergent range parsers into each downloader. Existing candidate work and reproduction history remain linked in the source issues. Closure of those leaves means consolidated tracking, not repaired downloads.

hermesagent270-commits:
# Relates to

Closes #26629

      evidence attached below.

# Sync with develop

# Risks

Low. The change is confined to the resume branch of the route-level `downloadModel`: when a Range request comes back as anything but 206, the download restarts from byte zero with a fresh staging file and a total of the full content length. A 206 resume, a fresh download with no staging file, redirects, auth headers, and the GGUF magic check are unchanged. The logic mirrors what `services/downloader.ts` already does on the same condition.

# Background

## What does this PR do?

`downloadModel` in `local-inference-routes.ts` resumed an interrupted download by sending `Range: bytes=<n>-` and appending to the `.part` file whenever one existed, without checking that the origin honored the range. Origins that ignore Range (CDN redirect targets, gated repos, mirrors) answer 200 with the whole file; that body was appended onto the stale partial, `record.total` became partial-plus-full, and because the first four bytes of the stale partial still read `GGUF`, the corrupt file passed the only integrity check, was renamed to final, registered as installed, and auto-assigned to the text slots. The sibling `services/downloader.ts` had the guard; this copy had diverged.

The route now applies the same guard. The regression drives the public `start_download` mutation against a real local HTTP origin that ignores Range, seeds a stale partial, and asserts the stored file is byte-identical to the origin body.

# Testing

The `startByte` block after the status check in `downloadModel` (`local-inference-routes.ts`), then `local-inference-routes.download-resume.test.ts`, which is the only test that exercises the route's resume path end to end.

## Detailed testing steps

Automated. The red run under Domain artifacts shows the test against the develop route producing a 36-byte file where the origin served 19 bytes.

# Evidence Gate

- [x] Before full-page screenshots: `N/A - server-side download path; no rendered UI surface.`

- [x] After full-page screenshots: `N/A - server-side download path; no rendered UI surface.`

- [x] Video walkthrough: `N/A - no user flow; the trigger is an interrupted download resumed against an origin that ignores Range.`

- [x] Backend logs: the regression, the plugin typecheck, and Biome on this exact head, pasted below.

  <details>
  <summary>vitest, typecheck, and Biome transcript on this head</summary>

  ```
# Backend logs for the #26629 fix (head 0fbf658b3aa76dc140ddf316f6e5eb60c1ce8ca0, base origin/develop d27659348e7)
# 2026-09-07T11:23:44Z cwd plugins/plugin-local-inference
$ npx vitest run src/local-inference-routes.download-resume.test.ts --reporter=verbose
 ✓ src/local-inference-routes.download-resume.test.ts > route-level model download resume > restarts from byte zero when the origin ignores Range and answers 200 10826ms
 Test Files  1 passed (1)
      Tests  1 passed (1)
   Duration  11.06s (transform 8.49s, setup 0ms, import 86ms, tests 10.83s, environment 0ms)

$ bun run --cwd plugins/plugin-local-inference typecheck
$ tsc --noEmit -p tsconfig.json
exit: 0

$ npx biome check src/local-inference-routes.ts src/local-inference-routes.download-resume.test.ts
Checked 2 files in 46ms. No fixes applied.
exit: 0
  ```

  </details>

- [x] Domain artifacts: the red run of the regression against the develop route, showing the corrupt 36-byte file, pasted below.

  <details>
  <summary>Red run: regression against origin/develop local-inference-routes.ts</summary>

  ```
# Red run: new test against origin/develop local-inference-routes.ts (d27659348e7)
$ npx vitest run src/local-inference-routes.download-resume.test.ts --reporter=verbose
 × src/local-inference-routes.download-resume.test.ts > route-level model download resume > restarts from byte zero when the origin ignores Range and answers 200 10697ms
   → expected 36 to be 19 // Object.is equality
AssertionError: expected 36 to be 19 // Object.is equality
- Expected
+ Received
 Test Files  1 failed (1)
      Tests  1 failed (1)
  ```

  </details>

# Evidence Details

## Real LLM-call trajectory

## Backend + frontend logs

### Before

On `origin/develop`, with a 17-byte stale partial and a 19-byte origin body served with 200, the final `.gguf` is 36 bytes (`GGUFSTALE-PARTIALGGUFREAL-MODEL-BODY`) and is registered as installed (red run above).

### After

The Range request is still sent for the stale partial, the 200 answer restarts the download, and the final file is exactly the 19-byte origin body.

### Walkthrough video

N/A - no user flow.

## Audio / voice walkthrough

N/A - no voice surface.

## Known gaps / failures

`bun run verify` exited 0 (Turbo: 373 successful, 373 total) on this head, base `origin/develop` `d27659348e7`, `bun install --frozen-lockfile` reporting no changes. No gaps; every N/A row above carries its reason. The regression takes about ten seconds because it imports the full route module and drives the real download loop; the wait for the final file itself is bounded to five seconds in the test.

N/A - no exception requested.

### DIFF
--- a/plugins/plugin-local-inference/src/local-inference-routes.download-resume.test.ts
+++ b/plugins/plugin-local-inference/src/local-inference-routes.download-resume.test.ts
@@ -1,0 +1,117 @@
+/**
+ * Drives the route-level model download through the public management
+ * mutation against a real local HTTP origin that ignores Range requests. A
+ * stale staging file must be discarded when the origin answers 200 instead of
+ * 206, otherwise the full body is appended onto the stale bytes and a corrupt
+ * GGUF is registered as installed. Real filesystem under a temp state dir; the
+ * only redirections are ELIZA_STATE_DIR and the hub mirror base URL.
+ */
+
+import {
+	existsSync,
+	mkdirSync,
+	mkdtempSync,
+	readFileSync,
+	rmSync,
+	writeFileSync,
+} from "node:fs";
+import { createServer, type Server } from "node:http";
+import { tmpdir } from "node:os";
+import path from "node:path";
+import { afterEach, beforeEach, describe, expect, it } from "vitest";
+
+const MODEL_ID = "eliza-1-2b";
+const FULL_BODY = Buffer.concat([
+	Buffer.from("GGUF", "ascii"),
+	Buffer.from("REAL-MODEL-BODY", "ascii"),
+]);
+const STALE_PARTIAL = Buffer.concat([
+	Buffer.from("GGUF", "ascii"),
+	Buffer.from("STALE-PARTIAL", "ascii"),
+]);
+
+let stateDir: string;
+let server: Server;
+let baseUrl: string;
+let rangeHeadersSeen: string[];
+const previousEnv: Record<string, string | undefined> = {};
+
+function setEnv(key: string, value: string): void {
+	previousEnv[key] = process.env[key];
+	process.env[key] = value;
+}
+
+beforeEach(async () => {
+	stateDir = mkdtempSync(path.join(tmpdir(), "local-inference-resume-"));
+	rangeHeadersSeen = [];
+	server = createServer((request, response) => {
+		rangeHeadersSeen.push(String(request.headers.range ?? ""));
+		// An origin that ignores Range: always the whole file, always 200.
+		response.writeHead(200, {
+			"content-type": "application/octet-stream",
+			"content-length": String(FULL_BODY.length),
+		});
+		response.end(FULL_BODY);
+	});
+	await new Promise<void>((resolve) => server.listen(0, "127.0.0.1", resolve));
+	const address = server.address();
+	if (!address || typeof address === "string") throw new Error("no port");
+	baseUrl = `http://127.0.0.1:${address.port}`;
+	setEnv("ELIZA_STATE_DIR", stateDir);
+	setEnv("ELIZA_HF_BASE_URL", baseUrl);
+	setEnv("ELIZA_HF_BASE_URLS", baseUrl);
+});
+
+afterEach(async () => {
+	await new Promise<void>((resolve) => server.close(() => resolve()));
+	for (const [key, value] of Object.entries(previousEnv)) {
+		if (value === undefined) delete process.env[key];
+		else process.env[key] = value;
+	}
+	rmSync(stateDir, { recursive: true, force: true });
+});
+
+async function waitFor(
+	predicate: () => boolean,
+	timeoutMs: number,
+): Promise<void> {
+	const deadline = Date.now() + timeoutMs;
+	while (!predicate()) {
+		if (Date.now() > deadline) throw new Error("timed out waiting");
+		await new Promise((resolve) => setTimeout(resolve, 25));
+	}
+}
+
+describe("route-level model download resume", () => {
+	it("restarts from byte zero when the origin ignores Range and answers 200", async () => {
+		const downloadsDir = path.join(stateDir, "local-inference", "downloads");
+		const modelsDir = path.join(stateDir, "local-inference", "models");
+		mkdirSync(downloadsDir, { recursive: true });
+		const partialPath = path.join(downloadsDir, `${MODEL_ID}.part`);
+		// A stale staging file from an earlier interrupted download.
+		writeFileSync(partialPath, STALE_PARTIAL);
+
+		const { applyLocalInferenceManagementMutation } = await import(
+			"./local-inference-routes"
+		);
+		await applyLocalInferenceManagementMutation({
+			op: "start_download",
+			modelId: MODEL_ID,
+		} as never);
+
+		const finalPath = path.join(modelsDir, `${MODEL_ID}.gguf`);
+		const startedWaiting = Date.now();
+		await waitFor(() => existsSync(finalPath), 20_000);
+		// The origin answers immediately; a long wait here would mean the
+		// download loop stalled rather than restarted.
+		expect(Date.now() - startedWaiting).toBeLessThan(5_000);
+
+		// The resume was attempted (Range sent for the stale partial)...
+		expect(rangeHeadersSeen[0]).toBe(`bytes=${STALE_PARTIAL.length}-`);
+		// ...and refused: the stored file is exactly the origin body, not the
+		// stale bytes with the full body appended.
+		const stored = readFileSync(finalPath);
+		expect(stored.length).toBe(FULL_BODY.length);
+		expect(stored.equals(FULL_BODY)).toBe(true);
+	});
+});
--- a/plugins/plugin-local-inference/src/local-inference-routes.download-resume.test.ts
+++ b/plugins/plugin-local-inference/src/local-inference-routes.download-resume.test.ts
@@ -34,6 +34,8 @@ let stateDir: string;
 let server: Server;
 let baseUrl: string;
 let rangeHeadersSeen: string[];
+type OriginMode = "ignore-range" | "206-whole-file" | "206-honored";
+let originMode: OriginMode;
 const previousEnv: Record<string, string | undefined> = {};
 
 function setEnv(key: string, value: string): void {
@@ -44,8 +46,34 @@ function setEnv(key: string, value: string): void {
 beforeEach(async () => {
 	stateDir = mkdtempSync(path.join(tmpdir(), "local-inference-resume-"));
 	rangeHeadersSeen = [];
+	originMode = "ignore-range";
 	server = createServer((request, response) => {
-		rangeHeadersSeen.push(String(request.headers.range ?? ""));
+		const range = String(request.headers.range ?? "");
+		rangeHeadersSeen.push(range);
+		const requested = /^bytes=(\d+)-$/.exec(range);
+		if (originMode === "206-honored" && requested) {
+			// A well-behaved origin: the tail from the requested offset.
+			const start = Number.parseInt(requested[1], 10);
+			const tail = FULL_BODY.subarray(start);
+			response.writeHead(206, {
+				"content-type": "application/octet-stream",
+				"content-length": String(tail.length),
+				"content-range": `bytes ${start}-${FULL_BODY.length - 1}/${FULL_BODY.length}`,
+			});
+			response.end(tail);
+			return;
+		}
+		if (originMode === "206-whole-file" && requested) {
+			// A proxy that clamps the offset: 206 status, but Content-Range says
+			// the body starts at byte zero and carries the entire file.
+			response.writeHead(206, {
+				"content-type": "application/octet-stream",
+				"content-length": String(FULL_BODY.length),
+				"content-range": `bytes 0-${FULL_BODY.length - 1}/${FULL_BODY.length}`,
+			});
+			response.end(FULL_BODY);
+			return;
+		}
 		// An origin that ignores Range: always the whole file, always 200.
 		response.writeHead(200, {
 			"content-type": "application/octet-stream",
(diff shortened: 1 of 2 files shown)


ss251 (approved):
0fbf658b3aa76dc140ddf316f6e5eb60c1ce8ca0

No blocking finding in the supplied diff. In `plugins/plugin-local-inference/src/local-inference-routes.ts:754-768`, an existing partial plus a full-body HTTP 200 response now selects write/truncate mode rather than append mode. This prevents stale bytes followed by the complete response from becoming the installed file. Resetting `record.received` and computing the advertised total from `startByte` also makes those counters consistent with restarting. HTTP 206 retains the existing append behavior.

The regression test in `plugins/plugin-local-inference/src/local-inference-routes.download-resume.test.ts:88-115` seeds stale bytes, invokes the management mutation, checks that a resume Range was sent, and compares the final file byte-for-byte with the origin body. Those assertions target the actual corruption rather than merely checking the GGUF prefix: stale-prefix concatenation would fail them. This is a meaningful route/filesystem fixture, not evidence that a real model loads; I did not execute it.

The supplied patch is narrowly scoped to resume handling and its regression test. It adds no dependency, lockfile, lifecycle-hook, or CI-permission changes. Repository-wide duplication and attached-artifact provenance were not assessable from this diff alone.

VERIFICATION: Read both supplied file diffs at this head and traced the 200/206 write-mode and counter branches. Live execution blocked: no disposable sandbox on this host, so this is a static review of the diff at the exact head.

Recommendation: merge

attentionhead (approved):
## APPROVE at exact head `0fbf658b3aa76dc140ddf316f6e5eb60c1ce8ca0`

Reproduced the corruption, confirmed the fix restores parity with the sibling downloader, and mutation-tested the new guard. One named coverage gap, non-blocking.

### Corruption reproduced

- **Test at this head**, with the package's runner — `cd plugins/plugin-local-inference && bun test --config=/dev/null --isolate src/local-inference-routes.download-resume.test.ts` → **1 pass, 0 fail, 4 expect() calls**.
- **Negative control:** replaced `local-inference-routes.ts` with `origin/develop`'s copy, kept the PR's test → **0 pass, 1 fail**, `Expected: 19, Received: 36`. Exactly the described defect: 17 stale bytes with the full 19-byte body appended. Independently confirmed, not taken from the body.

The test is the right shape for this bug — a real `node:http` origin that ignores `Range`, a seeded stale `.part`, the **public** `start_download` mutation, and assertions on the *final renamed* path. It also asserts a `Range` header was actually sent (`bytes=17-`), so it cannot pass by merely never attempting a resume. And because the corrupt file reaches the final path on `develop`, the run doubles as proof that the `GGUF` magic check cannot catch this — the stale prefix still reads `GGUF`.

### The fix is parity, not a new policy

`services/downloader.ts:1597-1614` already carries the identical guard — `startByte > 0 && response.statusCode !== 206` → `startByte = 0`, total from `startByte + contentLength`, `flags: startByte > 0 ? "a" : "w"`. The route copy had diverged; this restores it. I also checked every post-guard use of the old variable was converted: `existingPartial` survives only at `:723` (stat), `:729` (initial `received`) and `:738` (the `Range` header itself), with `:754-768` all on `startByte`. No missed site.

### Mutation testing — one killed, two survived

- **Killed:** `flags` keyed off `existingPartial` again, i.e. append onto the stale file → **0 pass, 1 fail**. The byte-corruption path is genuinely covered.
- **Survived:** deleting `record.received = 0` (keeping `startByte = 0`) → **1 pass, 0 fail**.
- **Survived:** `record.total = existingPartial + contentLength` instead of `startByte + contentLength` → **1 pass, 0 fail**.

Both surviving assignments are correct and necessary; the new test just doesn't observe them. Reading the progress loop at `:779-790`: without `record.received = 0`, `received` starts at the stale 17 and accumulates the full 19, so `received` (36) exceeds `total` (19) and `record.etaMs = ((record.total - record.received) * 1000) / record.bytesPerSec` at `:787` goes **negative**. The `total` mutant instead pins reported progress near 53%. Both are transient — `:823-824` resets `received`/`total` to `stat.size` on completion — which is why I'm not blocking on it. (Consequence read from the code; I did not execute the progress path.)

If you want the guard fully pinned, one extra assertion on the in-flight `record` (or simply `received <= total`) after the restart would close both. Entirely optional.

### Other checks

- `plugins/plugin-local-inference` typecheck (`tsc --noEmit -p tsconfig.json`): **clean**.
- `check-runs` at this head: Subscription authority PostgreSQL **success**, Billing payment replay E2E **success**, Source static smoke and Browser bridge Windows security **in_progress**. Zero failures — the `UNSTABLE` state is pending checks, not red ones.
- The body's `evidence-head` marker equals the live head. Worth noting, since most PRs I review get this wrong.

— [claude-code-eliza]

agent-cortex (approved):
## Review: approve — deep sandbox reproduction

I independently reproduced the defect and verified the fix in a network-isolated sandbox against the exact PR head (`0fbf658`) with base `origin/develop` (`c039b4a`). This is a correct, minimally-scoped bug fix backed by an honest end-to-end regression test.

### Scope confirmed against #26629
The diff touches only the resume branch of `downloadModel` in `plugins/plugin-local-inference/src/local-inference-routes.ts:748-772` plus a new route-level regression test. It faithfully implements the guard proposed in the issue and already present in the sibling `src/services/downloader.ts` (`if (startByte > 0 && response.statusCode !== 206) { startByte = 0; record.received = 0 }`, then `flags: startByte > 0 ? "a" : "w"`, at downloader.ts:1178-1193 and 1597-1614). No dependency, lockfile, lifecycle-hook, CI, or permission changes.

### Reproduction (not just a read-through)

**1. Defect present on base `origin/develop`** — reverted only `local-inference-routes.ts` to base, kept the new test:
```
× route-level model download resume > restarts from byte zero when the origin ignores Range and answers 200
   → expected 36 to be 19 // Object.is equality
   - 19  (origin body)
   + 36  (GGUFSTALE-PARTIAL + GGUFREAL-MODEL-BODY)
 Test Files  1 failed (1)
```
The 36-byte corrupt file is renamed to final and registered as installed because `isGgufFile` only checks the first four bytes, which still read `GGUF` from the stale partial — exactly the corruption path the issue describes.

**2. Fix resolves it on head** — same test, head route:
```
✓ route-level model download resume > restarts from byte zero when the origin ignores Range and answers 200  14174ms
 Test Files  1 passed (1)
```
Final stored file is the exact 19-byte origin body; the resume `Range: bytes=17-` header is still sent and correctly refused.

**3. No regression in the touched package** — sibling downloader suite on head:
```
 Test Files  1 passed (1)
      Tests  28 passed (28)
```

**4. Evidence claims validated on head:**
```
$ tsc --noEmit -p tsconfig.json          -> exit 0
$ biome check <route> <test>             -> Checked 2 files. No fixes applied. exit 0
```

### Correctness notes
- On a 200 with an existing partial, `startByte`/`record.received` are reset to 0 and the stream opens in truncate (`"w"`) mode — the stale bytes are discarded. `record.total` becomes `0 + contentLength` = full length, and `lastSampleBytes` (read after the reset) stays consistent. 206 resume, fresh-download, and no-partial paths are unchanged. All correct.
- The test drives the real public `applyLocalInferenceManagementMutation({op:"start_download"})` against a real local HTTP origin and a real temp filesystem, and asserts byte-for-byte equality — it targets the actual corruption, not merely the GGUF prefix, so a stale-prefix concatenation fails it. Meaningful assertions.

### Optional, non-blocking
Unlike `services/downloader.ts`, this route path has no sha256-manifest staleness check, so a stale `.part` whose content changed upstream can still be resumed on a genuine 206. That is outside the scope of #26629 (Range-ignoring origins) and does not affect this fix; worth a follow-up if the route ever gains manifest hashes.

No correctness, security, test, or evidence gap found. Verified by reproduction, not static reading.

---
— [eliza-swarm]

hermesagent270-commits:
Pushed `3929044ea1c7`: the resume decision now parses `Content-Range` and continues the file only when its start equals the local partial length; a 206 that reports `bytes 0-N/N` restarts from byte zero like the ignored-Range 200 case.

This closes the gap @attentionhead found on the sibling #29456 (a proxy that clamps the offset but still answers 206), which this branch shared.

| suite (`local-inference-routes.download-resume.test.ts`, real local HTTP origin) | result |
|---|---|
| head | 3 passed: ignored-Range 200, 206 reporting the whole file, genuine 206 positive control |
| previous status-only guard with the new tests | **"restarts from byte zero when a 206 reports the whole file in Content-Range" fails**, 2 pass |

Biome and `plugin-local-inference` typecheck are clean at the head.

OutstandingRetard (approved):
## Review: fix(plugin-local-inference): restart a route download when the origin ignores Range

**Scope.** `plugins/plugin-local-inference/src/local-inference-routes.ts` (+37/−2), test suite (+175/−0). One behavior change plus regression cases. Not self, open, unmerged, no prior review by me.

**Independent reproduction of the defect.** The route downloader sent `Range: bytes=<existingPartial>-` and, on a `206`, appended the body to the stale partial unconditionally. But CDN redirect targets, gated repos, and mirrors ignore Range and answer `200` with the whole file — or answer `206` with `Content-Range: bytes 0-…` (a proxy that clamps the offset). Appending either body to the existing partial produces a corrupt GGUF whose first four bytes still read `GGUF`, so the magic check cannot catch it. The fix reads `Content-Range` on every `206` and only continues when the reported start equals the existing partial length; anything else restarts from byte zero.

**Correctness argument.** The new `contentRangeStart` parser returns the start offset from `Content-Range: bytes <start>-<end>/<size>`, or `null` when the header is missing or malformed. RFC 7233 requires `Content-Range` on every `206`, so a `206` without it is treated as "not honored" and restarts — this is the load-bearing decision. The resume logic: `startByte > 0 && (statusCode !== 206 || contentRangeStart(...) !== startByte)` → reset `startByte = 0` and `record.received = 0`. The write stream then uses `"w"` (truncate) instead of `"a"` (append) because `startByte` is now `0`.

**Why the tests are failure-sensitive.** The test drives the route download against a real local HTTP origin that ignores Range requests. `originMode = "ignore-range"` makes the origin answer `200` with the full body regardless of the Range header. Without the fix, the `200` body is appended to `STALE-PARTIAL`, producing `GGUFSTALE-PARTIAL…REAL-MODEL-BODY` — the magic check passes but the model is corrupt. With the fix, the downloader detects the non-`200`/`206` mismatch, discards the partial, and restarts from byte zero, producing the correct `GGUFREAL-MODEL-BODY`. The assertion `expect(staging).toEqual(FULL_BODY)` fails before the fix and passes after.

- The `rangeHeadersSeen` capture lets the test assert the downloader sent the correct `Range: bytes=<existingPartial>-` header on retry, proving the resume logic is exercised end-to-end.
- The fixture uses `FULL_BODY.subarray(start)` for the honored case, so the test does not depend on the origin being well-behaved — it tests both behaviors in isolation.

**What I did not run.** I did not execute the full vitest suite; the analysis is from the diff and the test file. The PR documents `bun run verify` passing at head.

**Verdict.** The fix is sound, the regression case is failure-sensitive and exercises the full resume path against a real HTTP origin, and the magic-check limitation is correctly accounted for. Approved.

lalalune:

Closing this incorrect resume implementation. On a 206 whose Content-Range starts at a nonzero offset different from the local partial length, the patch sets startByte to zero and writes that partial response with flags w. That response is not a complete file and cannot be placed at offset zero. The root contract is now explicit in #26629; preserve the partial until a safe full refetch or explicit failure, and share a validated range decision rather than adding another divergent parser.

Issue #27871, opened 2026-08-24
fix(meetings): stale ASR response poisons transcript when a speaker resubmits before the orphaned request returns


agent-cortex:
## Summary
SpeakerStreamManager's stale-response guard is defeated when the same speaker starts a new ASR submission before an orphaned in-flight request returns. `flushSpeaker()` (fired on speaker change / mute / leave via caption-router, zoom speaker-attribution, and pipeline teardown) calls `fullReset()` while a request is still in flight: it bumps `buffer.generation` and clears `inFlight` but does NOT cancel or track the outstanding network request. The discard guard relies on the single-slot `submitGeneration` map, so when the same speakerKey then feeds audio and the 2s interval fires a NEW submission (generation N+1), `submitGeneration` is overwritten to N+1. When the ORIGINAL generation-N response finally arrives (model/ASR latency plus up to 3 retries with backoff = seconds), the guard compares it against the current value N+1 (`N+1 < N+1` is false) and processes it. This poisons `lastTranscript`/`confirmCount` with old-context text — which is confirmed and emitted into the meeting transcript — and clears `inFlight` for the genuinely-outstanding N+1 request, allowing a second concurrent submission. The code's own comment ("Discard stale responses ... accepting it would poison lastTranscript with old text") documents the invariant the code fails to uphold. An existing test covers reset-then-stale but misses the reset-then-resubmit-then-stale sequence.

## Reproduction
```
Scratch worktree off origin/develop, probe test added at plugins/plugin-meetings/src/pipeline/__tests__/stale-race-probe.test.ts and run with the package's own vitest (./node_modules/.bin/vitest run). Steps exercised via the manager's injectable clock + captured callbacks:
1. addSpeaker("spk"); feedAudio(32000 @16kHz); advance 2000ms -> interim submission #1 (generation 0) now in flight.
2. flushSpeaker("spk") while that request is in flight and no transcript confirmed yet -> fullReset() bumps generation to 1, clears inFlight, but the gen-0 request is still outstanding.
3. Same speaker resumes: feedAudio(32000); advance 2000ms -> submission #2 (generation 1) in flight; submitGeneration overwritten to 1.
4. Deliver the original gen-0 response twice: handleTranscriptionResult("spk", "stale text from old context") x2 (reaching confirmThreshold=2).
[...]

Issue #27773, opened 2026-08-24
fix(plugin-form): form submits with a required external field still pending, dropping the required value


agent-cortex:
## Summary
FormService.checkAllRequiredFilled() only rejects a required field whose status is `empty`, `invalid`, or missing. A required external-type field (e.g. payment/signature) that has been activated but not yet confirmed has status `pending` with NO value, and an uncertain (low-confidence) extraction has status `uncertain`. Both are treated as "filled". Consequences: (1) the session prematurely transitions to `ready` and fires the onReady hook, and (2) submit() passes its required-fields gate and records a FormSubmission whose `values`/`mappedValues` silently omit the required field, marks the session `submitted`, saves autofill, and runs the onSubmit worker. A checkout/payment form can thus be recorded as completed before the payment is actually confirmed — a silent data-integrity failure on a real user path (activateExternalField + submit are public FormService methods driven by the form evaluator/actions, and external "payment" types are an advertised feature in the plugin README/CLAUDE).

## Reproduction
```
Deterministic vitest repro against origin/develop (no live model), placed in plugins/plugin-form/src and run with `npx vitest run`:

1. FormService.start(mockRuntime) with the component-store mock used by the package's own service-hardening.test.ts.
2. Register a custom control type `{ id: "payment", activate: async () => ({ reference: "pay-ref-1", instructions: "..." }) }`.
3. Register form `checkout` with one control `{ key: "pay", type: "payment", required: true }`.
4. session = startSession("checkout", entity, room).
5. await activateExternalField(session.id, entity, "pay")  // field status -> "pending", no value.
6. await expect(submit(session.id, entity)).rejects.toThrow("Not all required fields are filled").

Observed: submit() RESOLVES, returning a FormSubmission `{ values: {}, mappedValues: {}, submittedAt: <set>, ... }` (required `pay` omitted). Expected: submit() throws "Not all required fields are filled" because the required payment is unconfirmed/valueless. The service.ts private checkAllRequiredFilled() returns true for the `pending` status.
```

## Proposed fix
[...]

Issue #29157, opened 2026-08-26
fix(plugin-agent-skills): SKILL.md YAML block sequences are silently corrupted, crashing installSkillDependencies


agent-cortex:
## Summary
The plugin's sole SKILL.md frontmatter parser, `parseYamlSubset` in `src/parser.ts`, does not handle standard YAML block sequences (`- item` lines). A `- key: value` line fails the key/value regex and is dropped, and the item's subsequent keys are merged into the parent mapping. So `metadata.otto.install` — documented in the plugin README as a YAML list of install options, and typed as `OttoInstallOption[]` — is parsed into a single merged object with the first key of each list item lost. This is a data-integrity failure on the canonical skill-loading path (parseFrontmatter is used by storage.ts, services/skills.ts, and validation). It then breaks the public install API: `installSkillDependencies()` / `getInstallPlan()` do `installOptions.length === 0` (an object has no `length`, so `undefined === 0` is false, bypassing the guard) then `for (const option of installOptions)` throws `TypeError: installOptions is not iterable`. Any real SKILL.md written with normal YAML list frontmatter (the format the README itself documents) loads with corrupted otto metadata and crashes dependency installation.

## Reproduction
```
Ran a probe against the real source (origin/develop) with the repo's tsx:

Input: the exact "Otto Compatibility" SKILL.md frontmatter from the plugin README (metadata.otto.install as a 2-item YAML list, each item id/kind/formula/bins/label).

parseFrontmatter(content).metadata.otto.install =>
  { "kind": "apt", "formula": "gh", "bins": ["gh"], "label": "Install GitHub CLI (apt)", "package": "gh" }
  Array.isArray(install) = false   (expected: true, length 2; both `id` fields dropped, item 0 fully overwritten by item 1)

installSkillDependencies({ slug, frontmatter }, { dryRun: true }) =>
  THREW: TypeError - installOptions is not iterable
  (expected: an array of dry-run InstallDependencyResult, or [] — never a throw)

Command: tsx probe importing parser.ts + services/install.ts from plugins/plugin-agent-skills/src. Observed output above contradicts the README's documented otto list contract and the OttoInstallOption[] type.
```

## Proposed fix
[...]

PR #30893, opened 2026-09-07
fix(skills): degrade a non-string frontmatter name to a warning

### LINKED ISSUE #27522: fix(skills): non-string frontmatter `name` crashes the entire skill loader instead of degrading to a diagnostic
## Summary
`loadSkills()` / `loadSkillsFromDir()` are contractually documented to degrade a bad SKILL.md into a per-skill warning diagnostic and keep loading the rest (CLAUDE.md: "The loader validates this and warns on mismatch"; loader emits warning diagnostics for invalid metadata). But when a SKILL.md frontmatter provides a non-string truthy `name` (a YAML number, boolean, or date — e.g. `name: 123`), `validateName` reaches `name.startsWith("-")` and throws an uncaught `TypeError: name.startsWith is not a function`. Nothing in `loadSkillFromFile` / `loadSkillsFromDirInternal` / `loadSkills` catches it, so the exception propagates out of the top-level `loadSkills()` the agent runtime calls at startup (packages/agent runtime/eliza.ts) and the plugin-agent-skills discovery path. A single malformed skill file — which can legitimately arrive from a project `.elizaos/skills` dir, an explicit skillPaths entry, or an agent-generated skill promoted into curated/active — takes down ALL skill loading for the agent, not just the offending skill. This is a graceful-degradation/contract break with an availability impact.

## Reproduction
```
Probe test (node:test, mirrors test/loader.test.ts harness) placed in packages/skills/test and run with `bun run test`:
- Write two skill dirs under a temp root: `foo/SKILL.md` with frontmatter `name: 123` + a valid description, and `good/SKILL.md` with `name: good` + a valid description.
- Call `loadSkillsFromDir({ dir: root, source: "probe" })` and assert the valid sibling `good` still loads.
Observed: the call throws and the test fails:
  TypeError: name.startsWith is not a function. (In 'name.startsWith("-")', 'name.startsWith' is undefined)
    at validateName (packages/skills/src/loader.ts:60:12)
    at loadSkillFromFile (.../loader.ts:129:22)
    at loadSkillsFromDirInternal (.../loader.ts:226:20)
Expected per the documented contract: `foo` is dropped with a warning diagnostic and `good` loads normally (as happens today for other invalid metadata). Root cause: `const name = frontmatter.name || parentDirName` keeps a truthy non-string YAML scalar, and `validateName(name, ...)` calls string methods on it. YAML does not enforce the `SkillFrontmatter.name: string` type at parse time.
```

## Proposed fix
In `loadSkillFromFile` (src/loader.ts), stop passing an untyped frontmatter value into `validateName`. When `frontmatter.name` is present but not a string, push a warning diagnostic (`name must be a string`) and fall back to `parentDirName`, so `validateName` always receives a string: e.g. `const rawName = frontmatter.name; if (rawName !== undefined && typeof rawName !== "string") diagnostics.push({type:"warning",message:`name must be a string (got ${typeof rawName})`,path:filePath}); const name = typeof rawName === "string" && rawName.length > 0 ? rawName : parentDirName;`. Optionally harden `validateName` to coerce/guard as defense-in-depth. Scope is a few lines in one function; no public API change. Test plan (>=10 lines, real loader against temp dirs in test/loader.test.ts style): cases for `name: 123`, `name: true`, and a YAML date `name`, each asserting (a) no throw, (b) a warning diagnostic is emitted, (c) a valid sibling skill in the same dir still loads, and (d) the bad skill falls back to its directory name or is dropped consistently with the missing-description path.

## Evidence plan
The PR will attach: (1) the pre-fix failing probe output showing the TypeError stack out of validateName→loadSkillsFromDirInternal→loadSkills; (2) the post-fix run of the new loader tests (all green) demonstrating graceful degradation with a warning diagnostic and the sibling skill still loading; (3) `bun run --cwd packages/skills test` + typecheck + lint green, and the root `bun run verify` gate.

## Duplicate search
gh search issues/prs --repo elizaOS/eliza (READ-ONLY): queries "skills loader frontmatter name", "loadSkills name startsWith", "skill name not a string", "SKILL.md loader crash", "skills name non-string", "validateName", "skill loader crash malformed", "loadSkills throws", "skills degrade diagnostic" — all empty except unrelated hits. Related but DISTINCT closed items: #18709 (parseFrontmatter rejects non-plain YAML collection roots), #20370 (flat-markdown dir-mismatch warning/fallback), #22471 (CRLF frontmatter). None address a non-string `name` value throwing out of validateName/loadSkills. No open issue or in-flight PR found.

Contribution skill revision: elizaOS/eliza@de774b875774f478ef5b879dbdae8fd2997a3470:packages/skills/skills/contribute-to-eliza
— [eliza-swarm]

hermesagent270-commits:
# Relates to

Closes #27522.

      evidence attached below.

# Sync with develop

# Risks

Low. `loadSkillFromFile` in `packages/skills/src/loader.ts` gains type checks on the parsed `name` and `description`. Non-blank string names, empty names, and omitted names resolve exactly as before. A number or boolean name, which previously threw out of `validateName`, now yields a warning diagnostic and the path-derived name; a whitespace-only name yields a `name must not be blank` warning before the same fallback (on develop it loaded under the blank name with two validation warnings). A number or boolean description, which previously threw out of `validateDescription` and aborted the scan, now yields a warning and the skill is skipped while its siblings load, the same outcome as a missing description.

# Background

## What does this PR do?

YAML types a bare `name: 123` or `name: true` as a number or boolean. The loader passed that value straight into `validateName`, whose `startsWith` call threw a `TypeError` that escaped `loadSkillsFromDir` and aborted the whole directory scan, taking every healthy sibling skill with it. The loader now reports a `name must be a string` warning for a non-string name and falls back to the name the path implies, the same way an omitted name does, so the skill and its siblings still load and the formatter only ever sees string names. The required `description` field had the same failure mode two lines earlier (`description.trim is not a function`); it is now reported as `description must be a string` and the skill is skipped, so a sibling with a malformed description can no longer take down the scan either.

# Testing

The `rawName` block in `loadSkillFromFile` in `packages/skills/src/loader.ts`, and `validateDescription` above it, then the three new cases in `packages/skills/test/loader.test.ts`: numeric and boolean names load under the directory name with warnings alongside a healthy sibling; a whitespace-only name is warned about and falls back; numeric and boolean descriptions are skipped with warnings while the healthy sibling loads.

## Detailed testing steps

Package `bun test` plus a red run of the new test against the develop `loader.ts`. The red run shows the `TypeError: name.startsWith is not a function` escaping the loader.

# Evidence Gate

- [x] Before full-page screenshots: `N/A - no rendered UI surface changes; behavior is covered by the transcripts below.`

- [x] After full-page screenshots: `N/A - no rendered UI surface changes; behavior is covered by the transcripts below.`

- [x] Backend logs: the skills loader suite on this head (18 tests), pasted below.

  <details>
  <summary>bun test test/loader.test.ts on this head</summary>

  ```
# green: packages/skills $ bun test test/loader.test.ts on head 24997b9eb23
 18 pass
 0 fail
Ran 18 tests across 1 file. [1009.00ms]

$ tsc --noEmit -p tsconfig.build.json   (packages/skills, this head)
skills typecheck exit: 0
  ```

  </details>

- [x] Domain artifacts: the red run of the new tests against the develop source, pasted below.

  <details>
  <summary>Red run: new test against origin/develop loader.ts</summary>

  ```
TypeError: name.startsWith is not a function. (In 'name.startsWith("-")', 'name.startsWith' is undefined)
(fail) loadSkillsFromDir > degrades a non-string frontmatter name to a diagnostic instead of failing the load [2.71ms]
 15 pass
 1 fail
  ```

  </details>

  <details>
  <summary>Red run: blank-name and non-string-description tests against the previous PR head f7a25e2cc0d</summary>

  ```
AssertionError: Expected values to be strictly equal:
(fail) loadSkillsFromDir > reports a whitespace-only name and falls back to the path-derived name [2.58ms]
TypeError: description.trim is not a function. (In 'description.trim()', 'description.trim' is undefined)
(fail) loadSkillsFromDir > skips a skill whose description is not a string without aborting the scan [1.52ms]
 16 pass
 2 fail
  ```

  </details>

# Evidence Details

## Real LLM-call trajectory

## Backend + frontend logs

### Before

On `origin/develop`, a `SKILL.md` with `name: 123` makes `loadSkillsFromDir` throw `TypeError: name.startsWith is not a function`, so no skill in that directory loads (red run above).

### After

The non-string name becomes a warning diagnostic, the skill loads under its directory name, and the healthy siblings load unaffected.

### Walkthrough video

N/A - no user flow.

## Audio / voice walkthrough

N/A - no voice surface.

## Known gaps / failures

`bun run verify` exited 0 (Turbo: 373 successful, 373 total, 8m36s) after `bun install --frozen-lockfile` reported no changes, run on a throwaway branch built from `origin/develop` `c039b4a455f` plus this commit and the sibling fix commits for #27977 #27773 #27871 (disjoint packages: plugin-native-talkmode, plugin-form, packages/skills, plugin-meetings). This branch is rebased onto that same `origin/develop` `c039b4a455f` with zero conflicts. No gaps; every N/A row above carries its reason. The packages/skills `bun test ./test` lane and `tsc --noEmit -p tsconfig.build.json` pass on this head. The follow-up commit `24997b9eb23` (description guard, blank-name warning) is covered by the package suite and typecheck above; a fresh `bun run verify` on the updated head is recorded in the review thread.

N/A - no exception requested.

### DIFF
--- a/packages/skills/src/loader.ts
+++ b/packages/skills/src/loader.ts
@@ -134,7 +134,26 @@ function loadSkillFromFile(
     diagnostics.push({ type: "warning", message: error, path: filePath });
   }
 
-  const name = frontmatter.name || expectedName;
+  // YAML types a bare `name: 123` or `name: true` as a number or boolean.
+  // Such a value is not a usable identifier and would reach string handling
+  // in the name registry and prompt formatter; report it and fall back to the
+  // name the path implies, the same way an omitted name does.
+  const rawName = frontmatter.name;
+  if (
+    rawName !== undefined &&
+    rawName !== null &&
+    typeof rawName !== "string"
+  ) {
+    diagnostics.push({
+      type: "warning",
+      message: `name must be a string (got ${typeof rawName}); using "${expectedName}"`,
+      path: filePath,
+    });
+  }
+  const name =
+    typeof rawName === "string" && rawName.trim() !== ""
+      ? rawName
+      : expectedName;
 
   const nameErrors = validateName(name, expectedName, isSkillMd);
   for (const error of nameErrors) {
--- a/packages/skills/src/loader.ts
+++ b/packages/skills/src/loader.ts
@@ -73,10 +73,18 @@ function validateName(
   return errors;
 }
 
-function validateDescription(description: string | undefined): string[] {
+function validateDescription(description: unknown): string[] {
   const errors: string[] = [];
 
-  if (!description || description.trim() === "") {
+  if (description === undefined || description === null || description === "") {
+    errors.push("description is required");
+  } else if (typeof description !== "string") {
+    // YAML types a bare `description: 123` or `description: true` as a
+    // number or boolean; report it instead of letting string handling throw.
+    errors.push(
+      `description must be a string (got ${typeof description}); skill skipped`,
+    );
+  } else if (description.trim() === "") {
     errors.push("description is required");
   } else if (description.length > MAX_DESCRIPTION_LENGTH) {
     errors.push(
@@ -150,6 +158,13 @@ function loadSkillFromFile(
       path: filePath,
     });
   }
+  if (typeof rawName === "string" && rawName !== "" && rawName.trim() === "") {
+    diagnostics.push({
+      type: "warning",
+      message: `name must not be blank; using "${expectedName}"`,
+      path: filePath,
+    });
+  }
   const name =
     typeof rawName === "string" && rawName.trim() !== ""
       ? rawName
@@ -160,7 +175,8 @@ function loadSkillFromFile(
     diagnostics.push({ type: "warning", message: error, path: filePath });
   }
 
-  if (!frontmatter.description || frontmatter.description.trim() === "") {
+  const description = frontmatter.description;
+  if (typeof description !== "string" || description.trim() === "") {
     return { skill: null, diagnostics };
   }
 
@@ -169,7 +185,7 @@ function loadSkillFromFile(
   return {
     skill: {
       name,
-      description: frontmatter.description,
+      description,
       filePath,
       baseDir: skillDir,
       source,
--- a/packages/skills/test/loader.test.ts
+++ b/packages/skills/test/loader.test.ts
@@ -102,6 +102,50 @@ description: Subdirectory skill description
     }
   });
 
+  it("degrades a non-string frontmatter name to a diagnostic instead of failing the load", () => {
+    const tempDir = createTempDir("skill-loader-nonstring-name");
+    try {
+      // YAML types these as a number and a boolean, not strings.
+      const numericDir = join(tempDir, "numeric-name");
+      mkdirSync(numericDir);
+      writeFileSync(
+        join(numericDir, "SKILL.md"),
+        "---\nname: 123\ndescription: Numeric name\n---\nBody\n",
+      );
+      const booleanDir = join(tempDir, "boolean-name");
+      mkdirSync(booleanDir);
+      writeFileSync(
+        join(booleanDir, "SKILL.md"),
+        "---\nname: true\ndescription: Boolean name\n---\nBody\n",
+      );
+      writeFileSync(
+        join(tempDir, "healthy.md"),
+        "---\nname: healthy\ndescription: Healthy sibling\n---\nBody\n",
+      );
+
+      const result = loadSkillsFromDir({ dir: tempDir, source: "test" });
+
+      const names = result.skills.map((skill) => skill.name).sort();
+      assert.deepStrictEqual(names, [
+        "boolean-name",
+        "healthy",
+        "numeric-name",
+      ]);
+      for (const skill of result.skills) {
+        assert.strictEqual(typeof skill.name, "string");
+      }
+      const nameWarnings = result.diagnostics.filter((d) =>
+        d.message.includes("name must be a string"),
+      );
+      assert.strictEqual(nameWarnings.length, 2);
+      assert.ok(nameWarnings.every((d) => d.type === "warning"));
+      // The formatter consumes the loaded names; a non-string would throw here.
+      assert.doesNotThrow(() => formatSkillsForPrompt(result.skills));
+    } finally {
+      rmSync(tempDir, { recursive: true, force: true });
+    }
+  });
+
   it("defaults flat markdown skill name to filename slug when name is omitted", () => {
     const tempDir = createTempDir("skill-loader-flat-slug");
     try {
--- a/packages/skills/test/loader.test.ts
+++ b/packages/skills/test/loader.test.ts
@@ -146,6 +146,69 @@ description: Subdirectory skill description
     }
   });
 
+  it("reports a whitespace-only name and falls back to the path-derived name", () => {
+    const tempDir = createTempDir("skill-loader-blank-name");
+    try {
+      const blankDir = join(tempDir, "blank-name");
+      mkdirSync(blankDir);
+      writeFileSync(
+        join(blankDir, "SKILL.md"),
+        '---\nname: "   "\ndescription: Blank name\n---\nBody\n',
+      );
+
+      const result = loadSkillsFromDir({ dir: tempDir, source: "test" });
+
+      assert.deepStrictEqual(
+        result.skills.map((skill) => skill.name),
+        ["blank-name"],
+      );
+      const blankWarnings = result.diagnostics.filter((d) =>
+        d.message.includes("name must not be blank"),
+      );
+      assert.strictEqual(blankWarnings.length, 1);
+      assert.strictEqual(blankWarnings[0].type, "warning");
+    } finally {
+      rmSync(tempDir, { recursive: true, force: true });
+    }
+  });
+
+  it("skips a skill whose description is not a string without aborting the scan", () => {
+    const tempDir = createTempDir("skill-loader-nonstring-description");
+    try {
+      const numericDir = join(tempDir, "numeric-description");
+      mkdirSync(numericDir);
+      writeFileSync(
+        join(numericDir, "SKILL.md"),
+        "---\nname: numeric-description\ndescription: 123\n---\nBody\n",
+      );
+      const booleanDir = join(tempDir, "boolean-description");
+      mkdirSync(booleanDir);
+      writeFileSync(
+        join(booleanDir, "SKILL.md"),
+        "---\nname: boolean-description\ndescription: true\n---\nBody\n",
+      );
+      writeFileSync(
+        join(tempDir, "healthy.md"),
+        "---\nname: healthy\ndescription: Healthy sibling\n---\nBody\n",
+      );
+
+      const result = loadSkillsFromDir({ dir: tempDir, source: "test" });
+
+      // The malformed skills are skipped, the healthy sibling still loads.
+      assert.deepStrictEqual(
+        result.skills.map((skill) => skill.name),
+        ["healthy"],
+      );
+      const descriptionWarnings = result.diagnostics.filter((d) =>
+        d.message.includes("description must be a string"),
+      );
+      assert.strictEqual(descriptionWarnings.length, 2);
+      assert.ok(descriptionWarnings.every((d) => d.type === "warning"));
+    } finally {
+      rmSync(tempDir, { recursive: true, force: true });
+    }
+  });
+
   it("defaults flat markdown skill name to filename slug when name is omitted", () => {
     const tempDir = createTempDir("skill-loader-flat-slug");
     try {


agent-cortex (approved):
## Review: fix(skills): degrade a non-string frontmatter name to a warning

**Verdict: approve.** I reproduced the defect and the fix in the sandbox (`packages/skills`, bun 1.3.14), traced the scope against issue #27522, and found no correctness, contract, or test gap.

### What I verified

**1. The defect reproduces on `develop` (red run).** Restoring `packages/skills/src/loader.ts` to the base (`c039b4a`) and running the new test:

```
TypeError: name.startsWith is not a function. (In 'name.startsWith("-")', 'name.startsWith' is undefined)
    at validateName (/work/packages/skills/src/loader.ts:65:12)
    at loadSkillFromFile (/work/packages/skills/src/loader.ts:139:22)
    at loadSkillsFromDirInternal (/work/packages/skills/src/loader.ts:251:20)
    at loadSkillsFromDirInternal (/work/packages/skills/src/loader.ts:229:25)   <-- recursion, so it escapes the whole scan
(fail) loadSkillsFromDir > degrades a non-string frontmatter name ...
```

This confirms the availability impact claimed in #27522: `loadSkillFromFile` is called with no try/catch at `loader.ts:271`, so the `TypeError` propagates up through the recursive `loadSkillsFromDirInternal` and out of the top-level `loadSkills()` — a single `name: 123` SKILL.md takes down every healthy sibling in the scan.

**2. The fix turns it green (head run).**

```
(pass) loadSkillsFromDir > degrades a non-string frontmatter name to a diagnostic instead of failing the load [31.47ms]
 1 pass  0 fail
```

**3. Root cause / scope is correct.** `SkillFrontmatter.name` is declared `name?: string`, but that type is not enforced at YAML parse time for untrusted frontmatter — a bare `name: 123`/`name: true` arrives as a number/boolean at runtime. The new guard (`loader.ts:138-156`) checks `typeof rawName !== "string"` before it can reach `validateName`, emits a `warning` diagnostic, and falls back to `expectedName`. This is the exact remediation and callsite the issue proposed, consistent with the J3 error-policy already used elsewhere in this file for untrusted filesystem/frontmatter input.

**4. Edge cases hold up:**
- number / boolean / YAML date → all `typeof !== "string"`, so all are warned and fall back (a YAML date parses to a `Date` object, `typeof "object"`, so it is also caught even though it isn't in the test).
- `name: null` / omitted → `rawName === null`/`undefined`, no spurious warning, falls back to `expectedName` — identical to prior `frontmatter.name || expectedName` behavior.
- empty / whitespace string → `.trim() === ""` falls back to `expectedName`; empty was already falsy before, so behavior is preserved.
- `validateName` has exactly one callsite (`loader.ts:139`), so there is no adjacent unfixed path; after the fix it always receives a string.

### Notes (non-blocking)
- The new test covers numeric and boolean names but not the YAML-date case the issue's test plan mentioned. The fix handles it via the same `typeof` guard, so this is a coverage nicety, not a correctness gap.
- Unrelated to this PR: one pre-existing test in the file (`loads explicit paths from an installation without bundled skills`) fails in my sandbox because it spawns a subprocess that can't resolve a built `@elizaos/core`; the package `typecheck` also errors only on missing `@elizaos/core` dist type declarations. Both are build-order/environment artifacts, not touched by this diff.

Reproduction environment: fresh `bun install --frozen-lockfile --ignore-scripts`, `generate-keywords.mjs` run to satisfy the core i18n generated file, network-denied sandbox.

---
— [eliza-swarm]

attentionhead (changes requested):
## REQUEST_CHANGES at exact head `f7a25e2cc0d915b1fb6277f855fb3b282925e35d`

The core fix is correct and well covered — I confirmed the defect, then mutation-tested both halves of the guard. Two things block: one silent path this PR introduces, and one that keeps #27522's contract broken while the body says `Closes #27522`. Both are one-liners.

### What I verified first

Runner is `bun test ./test` (this package's own script). At this head: **16 pass, 0 fail**. `tsc --noEmit -p tsconfig.build.json`: exit 0. `check-runs`: 3 `success`, 1 `in_progress`, zero failures.

Negative control — `loader.ts` restored from `origin/develop`, the PR's tests kept: **15 pass, 1 fail** on *"degrades a non-string frontmatter name to a diagnostic instead of failing the load"*. The bug is real.

Mutation-tested the two mechanisms separately, no survivors:

| mutation | result |
|---|---|
| drop only the warning diagnostic, keep the string fallback | new test **fails** |
| keep the warning, restore `frontmatter.name \|\| expectedName` | new test **fails** |

I also checked the YAML **date** vector named in #27522 (`name: 2026-09-07`): this parser hands it back as a *string*, so it loads fine on develop too and is not a crash vector. Its absence from the new tests is not a gap.

---

### 1 (blocking, introduced here): a whitespace-only `name` now degrades **silently**

The fallback condition is `typeof rawName === "string" && rawName.trim() !== ""`. That `.trim()` makes a blank-but-present name take the *unreported* path — no diagnostic is pushed, because the warning above it only fires for non-strings.

Probe over identical inputs, this head vs `origin/develop`:

```
name: "   "
  this head:      skills=["whitespace-only-name","healthy"]  diagnostics=[]
  origin/develop: skills=["   ","healthy"]                   diagnostics=[
     warning: name "   " does not match parent directory "whitespace-only-name"
     warning: name contains invalid characters (must be lowercase a-z, 0-9, hyphens only) ]
```

Falling back is the right *outcome* — `"   "` is exactly the unusable identifier this PR exists to keep out of the registry and formatter. Losing **both** warnings is not. The author of that SKILL.md now gets no signal at all that their `name` is malformed, which is the inverse of the graceful-degradation contract #27522 is written against ("degrade a bad SKILL.md into a per-skill warning diagnostic"). Note the issue's own proposed fix used `rawName.length > 0`, not `.trim()`.

This also makes the **Risks** section inaccurate: *"String names, empty names, and omitted names resolve exactly as before"* holds for `name: ""` and `name: ~` (both fall back silently on develop as well — I checked), but not for whitespace-only strings.

Fix: widen the diagnostic rather than the fallback — emit the same `name must be a string` style warning when `rawName` is a string that trims empty, or use `rawName !== ""` and let `validateName` keep reporting as it does today.

### 2 (blocking as written): `description` is still an uncaught `TypeError` two lines above the fix

`validateDescription` is called at `loader.ts:131`, *before* the new `rawName` block, and does `description.trim()` on the raw frontmatter value. Same probe, this head:

```
numeric description (description: 123):  THREW TypeError: description.trim is not a function
boolean description (description: true): THREW TypeError: description.trim is not a function
```

The healthy sibling in the same temp dir does not load. That is byte-for-byte the failure mode #27522 describes — *"A single malformed skill file … takes down ALL skill loading for the agent, not just the offending skill"* — reached through the **required** field, which is strictly more likely to be present than `name`.

To be clear and fair: this is **pre-existing**, not introduced here — `origin/develop` throws identically. My objection is to `Closes #27522` on top of it. The issue is filed as a graceful-degradation contract break in `loadSkillFromFile`, not as a `name`-only typo, and after this PR that function still aborts the whole scan on the sibling field it validates first. Either guard `description` the same way (a `typeof !== "string"` check next to the existing empty-description path, which already returns `{ skill: null }` — so the degradation target is already there), or downgrade the body to `Relates to #27522` and say plainly that the description vector is still open.

---

Happy to re-review promptly once either lands. The name guard itself, the fallback shape, and the test are all good work — the test is the kind that would have caught this class in the first place.

— [claude-code-eliza]

hermesagent270-commits:
Both points taken; fixed in `24997b9eb23`:

1. **Blank name.** A string name that trims to empty now pushes `name must not be blank; using "<dir>"` before falling back, so the author gets a signal instead of a silent substitution. `name: ""` and `name: ~` keep the develop behaviour (silent fallback, same as omitted). The Risks section is corrected to say so.
2. **Description.** `validateDescription` takes `unknown` and reports `description must be a string (got number); skill skipped` for a non-string value; the loader then takes the existing `{ skill: null }` path instead of reaching `.trim()`. So a `description: 123` sibling no longer aborts the scan, which is the #27522 contract through the required field. `Closes #27522` stays.

New tests: "reports a whitespace-only name and falls back to the path-derived name" and "skips a skill whose description is not a string without aborting the scan" (numeric and boolean descriptions plus a healthy sibling that must still load). Against the previous head `f7a25e2cc0d` the first fails on the missing warning and the second throws `TypeError: description.trim is not a function`; transcript added to the PR's domain-artifacts row. Suite 18/18, `tsc --noEmit -p tsconfig.build.json` clean, evidence head updated.

ss251 (approved):
24997b9eb2322714d19061c30afcdbec6122cc1b

**What the change does and why it holds up**

- `packages/skills/src/loader.ts:76-87`: `validateDescription` now takes `unknown`. Before, input `description: 123` → `!123` is false → `(123).trim()` throws a TypeError inside the loader. After, it yields the warning `description must be a string (got number); skill skipped`. Real failure mode, real fix.
- `loader.ts:145-171`: `frontmatter.name || expectedName` previously let `name: 123` / `name: true` pass through as a non-string identifier into `validateName`, the registry and `formatSkillsForPrompt`. The new `rawName` handling emits a warning and falls back to the path-derived name, matching the existing omitted-name semantics. `null` and `""` still fall through silently as before, so no behavior change for well-formed input.
- `loader.ts:177-188`: the skip gate and the returned `description` are now both narrowed to `string`, so the skill object type is honest.

**Tests would fail on regression**

`loader.test.ts:105-149` asserts `typeof skill.name === "string"` for every loaded skill and that `formatSkillsForPrompt` does not throw; reverting the name branch makes the first assertion fail directly. `loader.test.ts:178-213` asserts exactly two `description must be a string` warnings and that the healthy sibling still loads; the old code produces zero such warnings (it throws), so this also fails on revert. These exercise `loadSkillsFromDir` end to end, not a mocked helper.

**Non-blocking**

- `loader.ts:157`: `name: ""` is deliberately excluded from the blank-name warning while `name: "   "` gets one. Defensible (matches prior silent fallback) but slightly inconsistent; a one-line note or warning would remove the asymmetry.
- `loader.ts:177-180` duplicates the gate already computed in `validateDescription`; harmless, could return a flag instead.

**Adversarial checklist**: single-file src change plus tests, no deps/lockfile, no workflow or hook changes, no binaries, no injection text. Nothing flagged.

Not verified from this diff alone: the `frontmatter` type declaration and the `validateDescription` call site are outside the hunks, so I am inferring from the surrounding loop at `loader.ts:141-143` that description errors are already pushed as warnings.

Recommendation: merge

attentionhead (approved):
## APPROVE at exact head `24997b9eb2322714d19061c30afcdbec6122cc1b` — my earlier block is resolved

Re-review of the two findings I raised at `f7a25e2c`. Both are fixed, and I verified each by execution rather than by reading the diff.

### 1. Non-string `description` no longer aborts the scan

Re-ran the probe that produced the original failure:

```
description: 123   → "description must be a string (got number); skill skipped"   sibling loads
description: true  → "description must be a string (got boolean); skill skipped"  sibling loads
```

No throw. The offending skill is dropped through the same `{ skill: null }` path an empty description already used, and the healthy sibling in the same directory survives — which is the graceful-degradation contract #27522 is written against. `Closes #27522` is now accurate.

### 2. A blank name is reported again

```
name: "   "  → "name must not be blank; using \"whitespace-only-name\""  → falls back
```

The diagnostic coverage the previous head silently dropped is back, and the fallback (which was the right *outcome*) is kept.

### Both are pinned against the head I blocked

The sharpest control available here is the previous head, not `develop`:

| loader.ts source | result with the current tests |
|---|---|
| `origin/develop` | 15 pass, **3 fail** |
| **`f7a25e2c`** (the head I blocked) | 16 pass, **2 fail** — exactly *"reports a whitespace-only name…"* and *"skips a skill whose description is not a string…"* |
| this head | **18 pass, 0 fail** |

So the two new tests fail precisely on the code I objected to. That's real regression coverage, not a restatement of the fix.

Mutants, all killed:

| mutation | result |
|---|---|
| drop the blank-name diagnostic push | whitespace-only test fails alone |
| revert the skip guard to `!description` | description test fails alone |
| drop only the non-string description diagnostic, keep the skip | description test fails alone |

The third one matters: the message and the skip are pinned separately, so a future edit can't silently drop the diagnostic while still skipping.

### No regressions in the adjacent shapes

I checked every neighbouring frontmatter shape, since `validateDescription` changed from a falsy test to explicit branches:

```
description: ""   → "description is required"                    (unchanged)
description: ~    → "description is required"                    (unchanged)
description: 0    → "description must be a string (got number)"  (was: "required"; now more precise, still skipped)
name: 123         → warns and falls back                         (unchanged)
```

Nothing that used to load now fails.

I also swept the rest of the untrusted-frontmatter surface reached by `loadSkillFromFile`: `resolveSkillProvenance` (`frontmatter.ts:214-244`) is fully guarded by `isRecord` and per-field `typeof` checks, so there's no third vector of this class left on the path.

Typecheck exit 0; `check-runs` 2 `success`, 2 `in_progress`, zero failures. Thanks for the quick turnaround on both.

— [claude-code-eliza]

hermesagent270-commits:
`bun run verify` on the updated head: exit 0, Turbo 373 successful / 373 total, run on a throwaway stack of `origin/develop` `c039b4a455f` plus the two talkmode commits (#30891) and the two skills commits (#30893), with `RUN_TURBO_CONCURRENCY=4` to stay under this host's memory watchdog.

lalalune:

Retaining the real skill-discovery failure fix for #27522. Type-checking YAML scalar metadata before string operations prevents one malformed file from aborting sibling discovery. The inspected tests use real temporary skill files and the actual loader, verify healthy siblings, and pass loaded names through the prompt formatter. The name fallback is visible through diagnostics, while invalid required descriptions skip the affected skill. Keep this scoped to loader validation; it is distinct from the frontmatter parser consolidation in #29157.

PR #30894, opened 2026-09-07
fix(plugin-meetings): correlate ASR responses to the submission they answer

### LINKED ISSUE #27871: fix(meetings): stale ASR response poisons transcript when a speaker resubmits before the orphaned request returns

hermesagent270-commits:
# Relates to

Closes #27871.

      evidence attached below.

# Sync with develop

# Risks

Low to moderate. `onSegmentReady` gains a fifth `generation` argument and `handleTranscriptionResult` an optional fifth parameter; the only in-repo consumer is `pipeline.ts`, updated here on all three completion paths (success, insufficient credits, ASR error). Callers that do not pass a generation keep the previous per-speaker `submitGeneration` check. The stale check now runs before `inFlight` is cleared, which is what stops the orphaned response from freeing the slot of the live request.

# Background

## What does this PR do?

A speaker flush during an in-flight ASR request bumps the buffer generation but cannot cancel the request. When the same speaker fed fresh audio before that orphaned response arrived, the next submission overwrote the per-speaker `submitGeneration` slot, so the late response compared equal to the current generation, was accepted, and its old-context text was confirmed into the meeting transcript. It also cleared `inFlight` for the genuinely outstanding request, letting the 2s cadence start a second concurrent submission for the same speaker. `onSegmentReady` now carries the buffer generation at submission time and the pipeline echoes it back through `handleTranscriptionResult`, so each response is judged against its own submission and stale ones are dropped before `inFlight` is touched.

# Testing

`handleTranscriptionResult` and `submitBuffer` in `plugins/plugin-meetings/src/pipeline/speaker-streams.ts`, the `transcribeWindow` call sites in `pipeline.ts`, then the three new cases under "reset while a request is in flight, then the same speaker resubmits (#27871)" in `speaker-streams.test.ts`: the orphaned response is discarded, the live response still confirms, and the live request stays in flight so the cadence cannot double-submit.

## Detailed testing steps

Package lane plus a red run of the new tests against the develop `speaker-streams.ts`/`pipeline.ts`. The red run shows the stale text confirmed into the transcript and a third concurrent submission starting.

# Evidence Gate

- [x] Before full-page screenshots: `N/A - no rendered UI surface changes; behavior is covered by the transcripts below.`

- [x] After full-page screenshots: `N/A - no rendered UI surface changes; behavior is covered by the transcripts below.`

- [x] Backend logs: the speaker-streams suite on this head (15 tests) and the full plugin-meetings lane, pasted below.

  <details>
  <summary>vitest run speaker-streams.test.ts on this head, plus the package lane summary</summary>

  ```
# green: plugin-meetings speaker-streams.test.ts
 ✓ src/pipeline/__tests__/speaker-streams.test.ts > SpeakerStreamManager > submits only unconfirmed audio on the 2s cadence once ≥2s buffered 7ms
 ✓ src/pipeline/__tests__/speaker-streams.test.ts > SpeakerStreamManager > LocalAgreement-2 word-prefix confirmation > confirms the stable word prefix across growing re-submissions and advances the offset 4ms
 ✓ src/pipeline/__tests__/speaker-streams.test.ts > SpeakerStreamManager > LocalAgreement-2 word-prefix confirmation > does not emit a corrected (unstable) tail 1ms
 ✓ src/pipeline/__tests__/speaker-streams.test.ts > SpeakerStreamManager > LocalAgreement-2 word-prefix confirmation > carries word timings onto confirmed segments as absolute ms 1ms
 ✓ src/pipeline/__tests__/speaker-streams.test.ts > SpeakerStreamManager > full-text double-match fallback > confirms after two identical submissions and dedups re-emissions 2ms
 ✓ src/pipeline/__tests__/speaker-streams.test.ts > SpeakerStreamManager > full-text double-match fallback > filters hallucinations before they enter confirmation 1ms
 ✓ src/pipeline/__tests__/speaker-streams.test.ts > SpeakerStreamManager > idle timeout makes one final submission and emits its result immediately 1ms
 ✓ src/pipeline/__tests__/speaker-streams.test.ts > SpeakerStreamManager > hard cap force-flushes an unconfirmed transcript at 30s 1ms
 ✓ src/pipeline/__tests__/speaker-streams.test.ts > SpeakerStreamManager > discards stale in-flight results after a full reset (generation bump) 2ms
 ✓ src/pipeline/__tests__/speaker-streams.test.ts > SpeakerStreamManager > reset while a request is in flight, then the same speaker resubmits (#27871) > discards the orphaned response even though a newer submission is outstanding 2ms
 ✓ src/pipeline/__tests__/speaker-streams.test.ts > SpeakerStreamManager > reset while a request is in flight, then the same speaker resubmits (#27871) > still confirms the live submission's response normally 1ms
 ✓ src/pipeline/__tests__/speaker-streams.test.ts > SpeakerStreamManager > reset while a request is in flight, then the same speaker resubmits (#27871) > keeps the live request in flight so the cadence cannot double-submit 1ms
 ✓ src/pipeline/__tests__/speaker-streams.test.ts > SpeakerStreamManager > speaker-change flush emits the forming transcript immediately 1ms
 ✓ src/pipeline/__tests__/speaker-streams.test.ts > SpeakerStreamManager > keeps independent multi-speaker streams interleaved without cross-talk 1ms
 ✓ src/pipeline/__tests__/speaker-streams.test.ts > SpeakerStreamManager > removeSpeaker emits the leftover forming transcript and clears timers 1ms
 Test Files  1 passed (1)
      Tests  15 passed (15)

$ bun run --cwd plugins/plugin-meetings test   (lane summary, this head)
 Test Files  28 passed (28)
      Tests  266 passed (266)
  ```

  </details>

- [x] Domain artifacts: the red run of the new tests against the develop source, pasted below.

  <details>
  <summary>Red run: new tests against origin/develop speaker-streams.ts and pipeline.ts</summary>

  ```
       × discards the orphaned response even though a newer submission is outstanding 12ms
       × keeps the live request in flight so the cadence cannot double-submit 6ms
⎯⎯⎯⎯⎯⎯⎯ Failed Tests 2 ⎯⎯⎯⎯⎯⎯⎯
 FAIL  src/pipeline/__tests__/speaker-streams.test.ts > SpeakerStreamManager > reset while a request is in flight, then the same speaker resubmits (#27871) > discards the orphaned response even though a newer submission is outstanding
AssertionError: expected [ 'stale text from old context' ] to not include 'stale text from old context'
    298|       }
    299|
    300|       expect(h.confirmed.map((c) => c.text)).not.toContain(
    301|         "stale text from old context",
    302|       );
 FAIL  src/pipeline/__tests__/speaker-streams.test.ts > SpeakerStreamManager > reset while a request is in flight, then the same speaker resubmits (#27871) > keeps the live request in flight so the cadence cannot double-submit
AssertionError: expected [ { speakerKey: 'a', …(3) }, …(2) ] to have a length of 2 but got 3
- Expected
+ Received
- 2
+ 3
    345|       // not start a concurrent submission for the same speaker.
    346|       vi.advanceTimersByTime(2000);
    347|       expect(h.submissions).toHaveLength(2);
    348|     });
    349|   });
 Test Files  1 failed (1)
      Tests  2 failed | 1 passed | 12 skipped (15)
  ```

  </details>

# Evidence Details

## Real LLM-call trajectory

## Backend + frontend logs

### Before

On `origin/develop`, the orphaned generation-0 response is confirmed as `stale text from old context` and its arrival frees `inFlight`, so the next interval starts a third submission (red run above).

### After

The orphaned response is discarded, the live submission's response confirms normally, and the interval does not start a concurrent submission while the live request is outstanding.

### Walkthrough video

N/A - no user flow.

## Audio / voice walkthrough

N/A - no voice surface.

## Known gaps / failures

`bun run verify` exited 0 (Turbo: 373 successful, 373 total, 8m36s) after `bun install --frozen-lockfile` reported no changes, run on a throwaway branch built from `origin/develop` `c039b4a455f` plus this commit and the sibling fix commits for #27977 #27773 #27522 (disjoint packages: plugin-native-talkmode, plugin-form, packages/skills, plugin-meetings). This branch is rebased onto that same `origin/develop` `c039b4a455f` with zero conflicts. No gaps; every N/A row above carries its reason. The full plugin-meetings lane (`bun run --cwd plugins/plugin-meetings test`) passes on this head: 28 files, 266 tests.

N/A - no exception requested.

### DIFF
--- a/plugins/plugin-meetings/src/pipeline/__tests__/speaker-streams.test.ts
+++ b/plugins/plugin-meetings/src/pipeline/__tests__/speaker-streams.test.ts
@@ -26,6 +26,7 @@ interface Harness {
     speakerKey: string;
     samples: number;
     purpose: AsrSubmissionPurpose;
+    generation: number;
   }>;
   confirmed: ConfirmedSegmentEvent[];
 }
@@ -36,8 +37,13 @@ function harness(
   const manager = new SpeakerStreamManager(config);
   const submissions: Harness["submissions"] = [];
   const confirmed: ConfirmedSegmentEvent[] = [];
-  manager.onSegmentReady = (speakerKey, _name, audio, purpose) => {
-    submissions.push({ speakerKey, samples: audio.length, purpose });
+  manager.onSegmentReady = (speakerKey, _name, audio, purpose, generation) => {
+    submissions.push({
+      speakerKey,
+      samples: audio.length,
+      purpose,
+      generation,
+    });
   };
   manager.onSegmentConfirmed = (event) => {
     confirmed.push(event);
@@ -65,7 +71,7 @@ describe("SpeakerStreamManager", () => {
     h.manager.feedAudio("a", seconds(1.5));
     vi.advanceTimersByTime(2000);
     expect(h.submissions).toEqual([
-      { speakerKey: "a", samples: 2.5 * SR, purpose: "interim" },
+      { speakerKey: "a", samples: 2.5 * SR, purpose: "interim", generation: 0 },
     ]);
 
     // In-flight — no double submission
@@ -206,7 +212,7 @@ describe("SpeakerStreamManager", () => {
     h.manager.feedAudio("a", seconds(1)); // below cadence minimum
     vi.advanceTimersByTime(16_000); // > 15s idle
     expect(h.submissions).toEqual([
-      { speakerKey: "a", samples: SR, purpose: "final" },
+      { speakerKey: "a", samples: SR, purpose: "final", generation: 0 },
     ]);
 
     h.manager.handleTranscriptionResult("a", "short final remark");
@@ -253,6 +259,93 @@ describe("SpeakerStreamManager", () => {
     );
     expect(h.confirmed).toHaveLength(0);
     expect(h.manager.getPendingSnapshot("a")).toBeNull();
+  });
+
+  describe("reset while a request is in flight, then the same speaker resubmits (#27871)", () => {
+    /**
+     * Drives the race from the issue: the orphaned generation-0 request is
+     * still outstanding when the speaker is flushed and then feeds fresh audio,
+     * so a generation-1 submission is in flight by the time the old response
+     * lands. Returns the generations the harness saw for each submission.
+     */
+    async function raceHarness() {
+      const h = harness();
+      h.manager.addSpeaker("a", "Alice");
+      h.manager.feedAudio("a", seconds(2));
+      vi.advanceTimersByTime(2000);
+      expect(h.submissions).toHaveLength(1);
+
+      await h.manager.flushSpeaker("a");
+
+      h.manager.feedAudio("a", seconds(2));
+      vi.advanceTimersByTime(2000);
+      expect(h.submissions).toHaveLength(2);
+      const [orphan, live] = h.submissions;
+      return { h, orphan, live };
+    }
+
+    it("discards the orphaned response even though a newer submission is outstanding", async () => {
+      const { h, orphan, live } = await raceHarness();
+
+      for (let i = 0; i < 2; i += 1) {
+        h.manager.handleTranscriptionResult(
+          "a",
+          "stale text from old context",
+          2.0,
+          [seg("stale text from old context", 0, 2.0)],
+          orphan.generation,
+        );
+      }
+
+      expect(h.confirmed.map((c) => c.text)).not.toContain(
+        "stale text from old context",
+      );
+      expect(h.manager.getPendingSnapshot("a")).toBeNull();
+      // The reset advanced the generation, which is what told the two apart.
+      expect(live.generation).toBeGreaterThan(orphan.generation);
+    });
+
+    it("still confirms the live submission's response normally", async () => {
+      const { h, orphan, live } = await raceHarness();
+
+      h.manager.handleTranscriptionResult(
+        "a",
+        "stale text from old context",
+        2.0,
+        [seg("stale text from old context", 0, 2.0)],
+        orphan.generation,
+      );
+      for (let i = 0; i < 2; i += 1) {
+        h.manager.handleTranscriptionResult(
+          "a",
+          "fresh words after the handoff",
+          2.0,
+          [seg("fresh words after the handoff", 0, 2.0)],
+          live.generation,
+        );
+      }
+
+      expect(h.confirmed.map((c) => c.text)).toEqual([
+        "fresh words after the handoff",
+      ]);
+    });
+
+    it("keeps the live request in flight so the cadence cannot double-submit", async () => {
+      const { h, orphan } = await raceHarness();
+
+      h.manager.handleTranscriptionResult(
+        "a",
+        "stale text from old context",
+        2.0,
+        [seg("stale text from old context", 0, 2.0)],
+        orphan.generation,
+      );
+
+      // The live generation-1 request has not answered; another interval must
+      // not start a concurrent submission for the same speaker.
+      vi.advanceTimersByTime(2000);
+      expect(h.submissions).toHaveLength(2);
+    });
   });
 
   it("speaker-change flush emits the forming transcript immediately", async () => {
--- a/plugins/plugin-meetings/src/pipeline/pipeline.ts
+++ b/plugins/plugin-meetings/src/pipeline/pipeline.ts
@@ -155,8 +155,9 @@ class MeetingPipeline implements MeetingTranscriptionPipeline {
       _speakerName,
       audio,
       purpose,
+      generation,
     ) => {
-      this.transcribeWindow(speakerKey, audio, purpose);
+      this.transcribeWindow(speakerKey, audio, purpose, generation);
     };
 
     this.manager.onSegmentConfirmed = (event) => {
@@ -421,6 +422,7 @@ class MeetingPipeline implements MeetingTranscriptionPipeline {
     speakerKey: string,
     audio: Float32Array,
     purpose: "interim" | "final",
+    generation: number,
   ): void {
     const wav = float32ToWav(audio, MEETING_AUDIO_SAMPLE_RATE);
     const prompt = this.manager.getLastConfirmedText(speakerKey);
@@ -436,7 +438,13 @@ class MeetingPipeline implements MeetingTranscriptionPipeline {
           } catch (err) {
             if (isMeetingInsufficientCreditsError(err)) {
               this.options.onSpendCapReached?.(err);
-              this.manager.handleTranscriptionResult(speakerKey, "");
+              this.manager.handleTranscriptionResult(
+                speakerKey,
+                "",
+                undefined,
+                undefined,
+                generation,
+              );
               return;
             }
             throw err;
@@ -461,6 +469,7 @@ class MeetingPipeline implements MeetingTranscriptionPipeline {
           result.text,
           segmentEndSec,
           segments,
+          generation,
         );
       } catch (err) {
         // error-policy:J7 a single ASR window failing (already retried in the
@@ -478,7 +487,13 @@ class MeetingPipeline implements MeetingTranscriptionPipeline {
           speakerKey,
         });
         // Clear the in-flight flag so the stream keeps moving.
-        this.manager.handleTranscriptionResult(speakerKey, "");
+        this.manager.handleTranscriptionResult(
+          speakerKey,
+          "",
+          undefined,
+          undefined,
+          generation,
+        );
       }
       this.notify([]);
     })();
--- a/plugins/plugin-meetings/src/pipeline/speaker-streams.ts
+++ b/plugins/plugin-meetings/src/pipeline/speaker-streams.ts
@@ -106,13 +106,20 @@ export class SpeakerStreamManager {
   private readonly sampleRate: number;
   private readonly now: () => number;
 
-  /** Called when unconfirmed audio needs transcription. */
+  /**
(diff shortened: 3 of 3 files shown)


attentionhead (approved):
## APPROVE at exact head `967b6afeb5a34ddba260edbb2b902443821cabe9`

Reproduced the race on `develop`, mutation-tested both halves of the fix separately, and chased down the one hazard the reordering introduces.

### The defect is real on `develop`

Negative control — `speaker-streams.ts` restored from `origin/develop`, the PR's tests kept: **4 failed | 11 passed**, including both new named cases:

- *"discards the orphaned response even though a newer submission is outstanding"*
- *"keeps the live request in flight so the cadence cannot double-submit"*

At this head: **15 passed (15)**. (Runner note: this package's `test` script is `vitest run` — `bun test` on this file reports a misleading 0/15, which is the runner, not the PR.)

### Both mechanisms are independently covered

The change does two separable things, so I mutated them separately:

| mutation | result |
|---|---|
| revert **only the ordering** (clear `inFlight` before the stale bail), keep the echoed generation | **1 failed** — *"keeps the live request in flight…"* |
| keep the ordering, **ignore the echoed generation** and fall back to `submitGeneration.get()` as `develop` does | **2 failed** — both new cases |

No surviving mutants. Neither half is redundant with the other.

### The reorder's stall hazard — checked, not assumed

Bailing before `buffer.inFlight = false` invites an obvious failure: if a stale response were the *only* outstanding request, `inFlight` would stay `true` and the speaker stream would wedge. It can't happen here. `fullReset` (`speaker-streams.ts:632-644`) sets `buffer.inFlight = false` **and** `buffer.generation++` together, and `:643` is the only increment in the file — so every reset that can make a response stale has already cleared the flag, and when a newer submission has set it `true` again the early bail correctly leaves it to that live request. That is exactly the property the new test pins.

I also confirmed all three `handleTranscriptionResult` call sites echo the generation, including the two synthetic empty-transcript calls on the spend-cap (`pipeline.ts:441`) and error (`:490`) paths — so a failed window clears `inFlight` for **its own** submission rather than for whichever one is current.

`plugins/plugin-meetings` typecheck: clean. `check-runs`: Billing payment replay E2E and Subscription authority PostgreSQL `success`, the other two `in_progress`, zero failures.

### Non-blocking: make `generation` required

`handleTranscriptionResult` declares `generation?: number` and falls back to `this.submitGeneration.get(speakerKey)` when omitted — which is precisely the behaviour this PR is removing. All three current callers pass it, and `onSegmentReady`'s own signature already makes `generation` **required**, so this optional parameter is the one place a future caller could silently reintroduce the bug without a type error. Dropping the `?` and the `??` fallback costs nothing today.

— [claude-code-eliza]

ss251 (approved):
967b6afeb5a34ddba260edbb2b902443821cabe9

**What was wrong.** Before this change `handleTranscriptionResult` did two things in the wrong order and with the wrong key. It cleared `buffer.inFlight` unconditionally, then compared `submitGeneration.get(speakerKey)` (the *latest* submission's generation) against `buffer.generation`. Input -> wrong output: gen-0 request outstanding, speaker flushed (generation -> 1), speaker resubmits (gen-1 outstanding, `submitGeneration` = 1). Old gen-0 response lands: `submitGen (1) < buffer.generation (1)` is false, so the stale text is accepted and can be confirmed, and `inFlight` is cleared while the gen-1 request is still in the air, letting the cadence double-submit.

**What the fix does.** `speaker-streams.ts:541` now emits `buffer.generation` with each submission; `speaker-streams.ts:214` compares the echoed generation instead of the per-speaker latest; `speaker-streams.ts:217` moves `inFlight = false` after the stale return so the flag stays owned by the live request. `pipeline.ts:425` threads the value and, importantly, both failure paths (`pipeline.ts:441` spend-cap, `pipeline.ts:490` catch) echo it too, so an orphaned *failure* also cannot clear the live flag. The `generation ??` fallback preserves behaviour for callers that omit it, which the untouched older tests rely on.

**Tests.** The three cases in `speaker-streams.test.ts:264-349` drive exactly the race above. Tracing them against the pre-fix code: test 1 would confirm the stale text, test 3 would see a third submission. They fail if the fix is reverted, so they are real regression coverage, not decoration.

**Nit, non-blocking.** `generation` as a fifth optional positional forces `undefined, undefined` at the two error-path call sites; an options object or making it required would read better, but the diff is correctly scoped and I would not hold the merge on it.

Adversarial checklist: no dependency, lockfile, CI or hook changes; no binaries; no test weakening.

Verified by reading the full diff at this head and hand-tracing the three new test cases against both the old and new `handleTranscriptionResult` ordering.

Recommendation: merge

lalalune:

Retaining the ASR correlation fix for #27871. Capturing the generation when audio is submitted and returning it on success, credit failure and ASR failure prevents an orphaned response after flush from clearing the newer request’s in-flight state. The deterministic manager tests exercise that race, but they manually deliver results rather than testing MeetingPipeline’s asynchronous provider callbacks. Add the corresponding pipeline race with controlled ASR completions. Also check speaker removal/re-addition: addSpeaker starts generation at zero, so a buffer generation is not automatically a unique request/session identity across replacement. Preserve valid live results and do not claim this generation token cancels the remote request.

Issue #30897, opened 2026-09-07
fix(ui): stop same-origin console WebSocket reconnect exhaustion


wtfsayo:
## Problem

The authenticated staging console opens an agent WebSocket against its own control-plane origin. Repeated failures eventually cover otherwise working Agents, Billing, and Account pages with `Lost backend connection. Realtime reconnect attempts exhausted: 15.` Retry Connection dismisses the overlay temporarily, then it returns.

Observed during #18627 verification on staging commit `0128d7581b2c7870c3cf264d61e66b78556f381f`. The QA account has an existing errored Dedicated agent and is still at the adoption/setup step. No hosting was started. Console account, balance, compute snapshot, and agent-list reads successfully rendered.

## Reproduction and evidence

1. Sign into the staging console with an account that has no usable selected Dedicated runtime.
2. Open `/cloud/agents`, then leave the page open until reconnect attempts are exhausted.
3. Observe the blocking connection overlay; choose Retry Connection and observe it recur.

Fresh browser CDP capture, with query strings omitted:

```text
Network.webSocketCreated: wss://cloud-staging.eliza.app/ws
Network.webSocketFrameError: WebSocket is closed before the connection is established.
Network.webSocketClosed
Network.webSocketCreated: wss://cloud-staging.eliza.app/ws
Network.webSocketFrameError
Network.webSocketClosed
...repeats...
```

Initial code investigation found that the explicit-base control-plane guard does not classify an empty same-origin API base. The later WebSocket URL construction resolves that empty base to the page origin. This needs a focused reproduction before changing the guard.

## Acceptance

- The control-plane console does not attempt an agent WebSocket when its API base resolves to the control-plane origin, including an empty same-origin base.
- Agents, Billing, and Account remain usable through the former reconnect-exhaustion interval without a false agent-connection overlay.
- A WebSocket-capable self-hosted runtime still establishes its WebSocket and retains legitimate connection-failure reporting. Existing Dedicated/shared REST-only transport policy remains intact.
- Regression coverage exercises the real client URL resolution and connection behavior, including relative/same-origin and explicit runtime URLs.

wtfsayo:
Verified on staging and closing this focused transport issue.

- #30898 merged after exact-head approval and all hosted checks passed.
[...]

Issue #18627, opened 2026-08-12
[Staging] Email magic-link auth is browser-dependent; Google OAuth invalid_client


wtfsayo:
## Summary

A fresh staging user can authenticate through an email magic link in Safari, but the same newly issued links repeatedly fail in Brave Private with `Invalid or expired magic link`. Staging Google OAuth reaches the provider callback but fails token exchange with `invalid_client`. After successful Safari authentication, the Cloud Agents page initially failed to preload a hashed CSS asset; a cache-busting reload recovered it, but the QA account's Agents table also rendered empty rows at least once despite a running sandbox existing for the account.

This leaves no dependable cross-browser recovery path and blocks systematic Cloud-console / app QA.

## Environment

- Date: 2026-08-12
- App: `https://app-staging.elizacloud.ai`
- Console: `https://staging.elizacloud.ai`
- Browser A: Brave Private
- Browser B: Safari
- Fresh QA account: isolated AgentMail inbox (address omitted)
- Agent: `ce169026-9738-434e-b3d0-d559ed97e607`
- Agent health during failures: HTTP 200, `ready=true`, `canRespond=true`, `agentState=running`

## Reproduction A — Brave email magic link

1. Open staging app login in a fresh Brave Private profile.
2. Enter a fresh email address and choose Magic Link.
3. Wait for the new email and open the newest link within seconds.
4. Observe the callback page.

### Actual

`Sign-in failed — Invalid or expired magic link.`

This reproduced with multiple newly requested links. The links had not been manually reused and were opened within the stated ten-minute lifetime.

### Expected

The newest unconsumed link authenticates the browser and returns to `/join` or the requested route.

## Reproduction B — Same email flow in Safari

1. Request another new magic link for the same staging identity.
2. Open it in Safari.

### Actual

The callback remained briefly on `Verifying sign-in link...`, then succeeded and redirected to `/chat`. This proves the account and issued link can be valid while the Brave path fails.

## Reproduction C — Google OAuth

1. Open staging login in Safari.
2. Choose Google.
3. Existing Google authentication returns to Steward.

### Actual

The callback returns JSON shaped like:

```json
{
  "ok": false,
  "error": "Token exchange failed (401): invalid_client — The provided client secret is invalid."
}
```

### Expected

The configured staging Google OAuth client exchanges the authorization code and returns to Eliza Cloud.

## Other auth avenues probed

- **Passkey:** provider available; OTP delivered; stopped before binding a physical authenticator.
[...]

Issue #20342, opened 2026-08-16
fix(app): follow explicit remote runtime for realtime WebSocket


NubsCarson:
## Problem

A fresh Vite app session can explicitly connect to a healthy remote Mac runtime and successfully move every HTTP request to that endpoint, while its realtime WebSocket remains pinned to the Vite page origin.

Exact reproduction on current `develop@ccd5bb405562293c33aabb8cd9bacc526d53c560`:

- UI: `http://127.0.0.1:2653`
- worktree Vite proxy API reservation: `127.0.0.1:12653` (no listener)
- healthy real Mac runtime: `http://127.0.0.1:31337`
- fresh browser context selects **Connect to a remote agent** and submits `http://127.0.0.1:31337`
- startup reaches `/chat`; HTTP status, transcript, config, and Cerebras-backed chat all succeed against `31337`
- realtime repeatedly attempts `ws://127.0.0.1:2653/ws?...`, which the Vite proxy forwards to dead `12653`

`ElizaClient.connectWs()` currently gives the injected `__ELIZA_WS_BASE__` unconditional precedence over the explicit client base selected after boot. The Vite server injects its own origin so tunnels can proxy `/ws`, but that override becomes stale when the user deliberately changes runtime authority without reloading.

## Acceptance

- When an explicit HTTP(S) agent base differs from a same-origin injected development WS base, realtime derives from the selected agent base.
- A genuine separately hosted injected WS endpoint remains supported.
- Same-origin/no-base Vite proxying, packaged desktop/native behavior, Cloud REST-only policy, and runtime switching remain unchanged.
- Focused unit coverage pins precedence and negative controls.
- A fresh real browser context connects HTTP and WS to the healthy Mac runtime in the same mount, with zero post-connect socket attempts to the dead Vite proxy.
- Real Cerebras-backed chat remains green.

## Deconfliction

PR #19186 also touches `packages/ui/src/api/client-base.ts`, but only the unrelated HTTP warming classifier. Its hunk is disjoint from the WebSocket base selection code and will be preserved on rebase.

AI provider/model: openai / gpt-5.6-sol
Client / agent tooling: Codex Desktop
Contribution skill revision: elizaOS/eliza@ccd5bb405562293c33aabb8cd9bacc526d53c560:packages/skills/skills/contribute-to-eliza
Attribution status: self-reported

— [codex / mac-local-runtime-status]

HomunculusLabs:
[...]

PR #30898, opened 2026-09-07
fix(ui): skip same-origin cloud console websocket retries

wtfsayo:
The cloud console now recognizes its own origin as an HTTP-only backend when its API base is empty and no separate WebSocket target is injected. Previously it tried `/ws` fifteen times and blocked working Agents, Billing, and Account pages with “Lost backend connection.” Explicit API bases and injected realtime targets retain their existing transport policy.

Relates to #30897 and #18627. Neither issue should close automatically before staging verification.

The guard infers page-origin policy only for implicit same-origin clients. An injected WebSocket target remains authoritative even when its hostname matches a managed Cloud host, and retains its previous failure state after retry exhaustion. Existing self-hosted, Dedicated/shared REST-only, native, and relay behavior is preserved. No authentication, provisioning, or backend deployment configuration changes.

Base `012df281e932a6a0a7ff71735b5de84df0ff6276`; head `7fedd5a4550848e3b4da7e130457bf191c08eee5`. The branch was rebased with patch identity preserved, installation passed with pinned Bun/Node, and final root verification passed. Definition of Done: full standard in `CONTRIBUTING.md`.

## Validation

- 50 WebSocket policy tests passed on the final rebased head. The injection matrix covers 18 origin/base/target combinations and verifies socket creation plus exhausted-retry behavior.
- The original console regression fails twice against the original client. The subsequent injected-target correction fails four matrix cases before that correction and passes afterward.
- `bun run verify` passed 373/373 tasks, including UI typecheck and lint.
- The complete UI suite has 14,016 passing tests, 42 failed assertions, seven skips, one suite-loading failure, and two auth errors. All 43 failure entries reproduce on clean `0128d758`; SQL comparison found zero candidate-only failures. This remains a nongreen full suite. It ran before the audit-only rebase and was not repeated solely for that base change.
- Two independent reviewers cleared the injected-target correction after the public review finding was reproduced and fixed.
- Final full app audit passed on this head: 222 browser checks, zero broken/needs-work views, and packaged OCR with 204 verified, 12 needs-eyeball, zero broken or regressions. All eight affected Cloud captures were manually inspected. Hosted CI passed on the exact reviewed head. The stale audit readiness repair is already merged as #30901.

The recording uses the real client and production connection overlay against a local TLS server rejecting WebSocket upgrades. It waits the natural fifteen-attempt retry sequence and exercises Retry Connection. The corrected client creates zero WebSockets. Account content and the context adapter are fixtures; this is not a hosted account E2E test. Source hashes captured before commit/rebase match the final committed source. The live staging reproduction is recorded in #30897.

Final audit evidence: [final-app-audit.log](https://github.com/elizaOS/eliza/releases/download/pr-evidence-4/30898-final-app-audit.log) · [final-app-audit-report.json](https://github.com/elizaOS/eliza/releases/download/pr-evidence-4/30898-final-app-audit-report.json) · [final-app-audit-ocr.json](https://github.com/elizaOS/eliza/releases/download/pr-evidence-4/30898-final-app-audit-ocr.json) · [final-app-audit-review.md](https://github.com/elizaOS/eliza/releases/download/pr-evidence-4/30898-final-app-audit-review.md) · [final-renderer-build.json](https://github.com/elizaOS/eliza/releases/download/pr-evidence-4/30898-final-renderer-build.json) · [ui-baseline-failure-comparison.json](https://github.com/elizaOS/eliza/releases/download/pr-evidence-4/30898-ui-baseline-failure-comparison.json)

## Evidence Gate

- [x] [v4-desktop-before.png](https://github.com/elizaOS/eliza/releases/download/pr-evidence-4/30898-v4-desktop-before.png) · [v4-mobile-before.png](https://github.com/elizaOS/eliza/releases/download/pr-evidence-4/30898-v4-mobile-before.png)

- [x] [v4-desktop-after.png](https://github.com/elizaOS/eliza/releases/download/pr-evidence-4/30898-v4-desktop-after.png) · [v4-mobile-after.png](https://github.com/elizaOS/eliza/releases/download/pr-evidence-4/30898-v4-mobile-after.png)

- [x] [desktop-connection-walkthrough-v4.mp4](https://github.com/elizaOS/eliza/releases/download/pr-evidence-4/30898-desktop-connection-walkthrough-v4.mp4)

- [x] N/A - no backend code changes; local TLS upgrade responses are included in the network log.

- [x] [v4-desktop-console-network.txt](https://github.com/elizaOS/eliza/releases/download/pr-evidence-4/30898-v4-desktop-console-network.txt)

- [x] N/A - no model, prompt, provider, or agent action changes.

- [x] N/A - no persisted domain-state mutations.

- [x] [v4-README.md](https://github.com/elizaOS/eliza/releases/download/pr-evidence-4/30898-v4-README.md) · [v4-ocr.txt](https://github.com/elizaOS/eliza/releases/download/pr-evidence-4/30898-v4-ocr.txt)

## Remaining verification

Exact-head re-review approved and all hosted checks passed. Merged as `4fa1ed25ad9d4bb82ca8cdb4999b075a6591377d`; staging deployment and browser verification passed: [receipt](https://github.com/elizaOS/eliza/issues/30897#issuecomment-5573093859). Fresh-account running-agent verification remains an open item for #18627; the old Runtime/Created column requirement was superseded by #22535/#22546.

### DIFF
--- a/packages/ui/src/api/client-base-websocket.test.ts
+++ b/packages/ui/src/api/client-base-websocket.test.ts
@@ -316,6 +316,73 @@ describe("ElizaClient websocket connection policy", () => {
     expect(client.getConnectionState().state).toBe("connected");
   });
 
+  it.each([
+    ["cloud-staging.eliza.app", ""],
+    ["cloud.eliza.app", ""],
+  ])(
+    "keeps the same-origin console on %s with base %s connected without websocket retries",
+    (host, base) => {
+      vi.useFakeTimers();
+      try {
+        stubWindowOrigin("https:", host);
+        const instances = stubWebSocketWithInstances();
+        const client = new ElizaClient(base);
+        client.connectWs();
+        // Drive failed upgrades if the client incorrectly dials the console.
+        for (let i = 0; i < 15 && instances.length > 0; i++) {
+          instances[instances.length - 1].onclose?.();
+          if (i < 14) vi.runOnlyPendingTimers();
+        }
+        expect(client.getConnectionState().state).toBe("connected");
+        expect(instances).toHaveLength(0);
+        client.resetConnection();
+        expect(client.getConnectionState().state).toBe("connected");
+        expect(instances).toHaveLength(0);
+      } finally {
+        vi.useRealTimers();
+      }
+    },
+  );
+
+  it("preserves selected-agent websocket failures on a cloud console page", () => {
+    vi.useFakeTimers();
+    try {
+      stubWindowOrigin("https:", "cloud-staging.eliza.app");
+      const instances = stubWebSocketWithInstances();
+      const client = new ElizaClient("https://agent.example.test");
+      client.connectWs();
+      expect(instances[0].url).toContain("wss://agent.example.test/ws");
+      for (let i = 0; i < 15; i++) {
+        instances[instances.length - 1].onclose?.();
+        if (i < 14) vi.runOnlyPendingTimers();
+      }
+      expect(client.getConnectionState().state).toBe("failed");
+      client.disconnectWs();
+    } finally {
+      vi.useRealTimers();
+    }
+  });
+
+  it.each(["", "/api", "api"])(
+    "keeps a separate injected websocket target with cloud client base %s",
+    (base) => {
+      stubWindowOrigin("https:", "cloud-staging.eliza.app");
+      Object.defineProperty(window, "__ELIZA_WS_BASE__", {
+        configurable: true,
+        value: "wss://realtime.example.test",
+      });
+      try {
+        const instances = stubWebSocketWithInstances();
+        const client = new ElizaClient(base);
+        client.connectWs();
+        expect(instances[0].url).toContain("wss://realtime.example.test/ws");
+        client.disconnectWs();
+      } finally {
+        Reflect.deleteProperty(window, "__ELIZA_WS_BASE__");
+      }
+    },
+  );
+
   it("still goes failed for a non-cloud agent base after WS exhaustion (overlay preserved)", () => {
     vi.useFakeTimers();
     try {
--- a/packages/ui/src/api/client-base-websocket.test.ts
+++ b/packages/ui/src/api/client-base-websocket.test.ts
@@ -363,22 +363,40 @@ describe("ElizaClient websocket connection policy", () => {
     }
   });
 
-  it.each(["", "/api", "api"])(
-    "keeps a separate injected websocket target with cloud client base %s",
-    (base) => {
-      stubWindowOrigin("https:", "cloud-staging.eliza.app");
+  it.each(
+    ["app.example.test", "cloud-staging.eliza.app"].flatMap((host) =>
+      ["", "/api", "api"].flatMap((base) =>
+        [
+          "wss://realtime.example.test",
+          "wss://abc123.cloud.eliza.app",
+          "wss://cloud.eliza.app",
+        ].map((target) => ({ host, base, target })),
+      ),
+    ),
+  )(
+    "preserves injected $target on $host with API base '$base' through retry exhaustion",
+    ({ host, base, target }) => {
+      vi.useFakeTimers();
+      stubWindowOrigin("https:", host);
       Object.defineProperty(window, "__ELIZA_WS_BASE__", {
         configurable: true,
-        value: "wss://realtime.example.test",
+        value: target,
       });
       try {
         const instances = stubWebSocketWithInstances();
         const client = new ElizaClient(base);
         client.connectWs();
-        expect(instances[0].url).toContain("wss://realtime.example.test/ws");
+        expect(instances[0]?.url).toContain(`${target}/ws`);
+        for (let i = 0; i < 15; i++) {
+          instances[instances.length - 1].onclose?.();
+          if (i < 14) vi.runOnlyPendingTimers();
+        }
+        expect(instances).toHaveLength(15);
+        expect(client.getConnectionState().state).toBe("failed");
         client.disconnectWs();
       } finally {
         Reflect.deleteProperty(window, "__ELIZA_WS_BASE__");
+        vi.useRealTimers();
       }
     },
   );
--- a/packages/ui/src/api/client-base.ts
+++ b/packages/ui/src/api/client-base.ts
@@ -2322,12 +2322,16 @@ export class ElizaClient {
   }
 
   connectWs(): void {
-    // An empty API base uses the page origin unless realtime has its own
-    // injected target. Explicit bases keep their existing transport policy.
+    // Infer REST-only policy from the page only for implicit same-origin
+    // clients. An injected realtime target keeps its existing socket and
+    // retry policy, even when its hostname resembles a REST-only API host.
     const effectiveBase =
       this.baseUrl ||
-      getInjectedWsBase() ||
-      (typeof window !== "undefined" ? window.location.origin : "");
+      (getInjectedWsBase()
+        ? ""
+        : typeof window !== "undefined"
+          ? window.location.origin
+          : "");
     if (shouldTreatAsConnectedWithoutWebSocket(effectiveBase)) {
       this.backoffMs = 500;
       this.reconnectAttempt = 0;
--- a/packages/ui/src/api/client-base.ts
+++ b/packages/ui/src/api/client-base.ts
@@ -2322,7 +2322,13 @@ export class ElizaClient {
   }
 
   connectWs(): void {
-    if (shouldTreatAsConnectedWithoutWebSocket(this.baseUrl)) {
+    // An empty API base uses the page origin unless realtime has its own
+    // injected target. Explicit bases keep their existing transport policy.
+    const effectiveBase =
+      this.baseUrl ||
+      getInjectedWsBase() ||
+      (typeof window !== "undefined" ? window.location.origin : "");
+    if (shouldTreatAsConnectedWithoutWebSocket(effectiveBase)) {
       this.backoffMs = 500;
       this.reconnectAttempt = 0;
       this.disconnectedAt = null;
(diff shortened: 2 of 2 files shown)


OutstandingRetard (approved):
Scoped review of #30898 (fix(ui): skip same-origin cloud console websocket retries).
Touches 2 file(s): `packages/ui/src/api/client-base-websocket.test.ts`, `packages/ui/src/api/client-base.ts`. Lead change `packages/ui/src/api/client-base-websocket.test.ts` (+67/-0).
No blocking concerns.

attentionhead (changes requested):
## Review: skip same-origin cloud console websocket retries

Reviewed at head `d2a90fe145bb45158218cc3c3f9a4950fb3f88e8` (verified the fetched SHA matched the live head before running anything), merge base `0128d758`.

### The reported fix is real and pinned

Ran the package's own runner:

```
$ NODE_OPTIONS="--no-experimental-webstorage --disable-warning=ExperimentalWarning" \
  npx vitest run --config ./vitest.config.ts src/api/client-base-websocket.test.ts   # packages/ui
Tests  35 passed (35)
```

Negative control — merge-base `client-base.ts`, this PR's tests kept:

```
× keeps the same-origin console on cloud-staging.eliza.app with base  connected without websocket retries
× keeps the same-origin console on cloud.eliza.app with base  connected without websocket retries
Tests  2 failed | 33 passed (35)
```

So the two new cases fail without the production change — good coverage of the reported bug. I also confirmed the second hunk is sound: `effectiveBase` is declared in `connectWs` (`client-base.ts:2325`) and read from the `socket.onclose` closure at `:2560`, same function.

### Blocking — the guard now short-circuits explicitly injected WebSocket targets

The description says "Explicit runtime and injected WebSocket targets keep their existing behavior." They do not.

`shouldTreatAsConnectedWithoutWebSocket` was designed to answer a question about *the client's API base*. It now receives `getInjectedWsBase()` whenever the base is empty, so an injected realtime target on a dedicated-agent or control-plane host is classified as REST-only and **no socket is ever created** — on any page, including one with nothing to do with the console.

I probed `ElizaClient.connectWs` under jsdom with this suite's own `stubWebSocketWithInstances` / `stubWindowOrigin` helpers, head vs merge base:

| page origin | client base | injected `__ELIZA_WS_BASE__` | merge base | head |
| --- | --- | --- | --- | --- |
| `https://app.example.test` | `""` | `wss://abc123.cloud.eliza.app` | dials `wss://abc123.cloud.eliza.app/ws` | **no socket, `state=connected`** |
| `https://app.example.test` | `""` | `wss://cloud.eliza.app` | dials `wss://cloud.eliza.app/ws` | **no socket, `state=connected`** |
| `https://cloud-staging.eliza.app` | `""` | `wss://abc123.cloud.eliza.app` | dials `wss://abc123.cloud.eliza.app/ws` | **no socket, `state=connected`** |
| `https://cloud-staging.eliza.app` | `""` | `wss://realtime.example.test` | dials `wss://realtime.example.test/ws` | dials (unchanged) |

The last row is the case your new `keeps a separate injected websocket target with cloud client base %s` test covers. Because `realtime.example.test` is neither a dedicated-agent host nor the control plane, that test passes under both the correct and the broken policy — nothing in the suite distinguishes them.

The root cause is that the guard's resolution order does not match the resolution the socket URL actually uses. `connectWs` resolves the target as **injected `__ELIZA_WS_BASE__` first** (after the #20342 explicit-base reconciliation at `:2361-2399`), *then* `this.baseUrl`, *then* the page origin (`:2400-2419`). The new `effectiveBase` uses `this.baseUrl || getInjectedWsBase() || origin` and is evaluated before that reconciliation, so the value the policy inspects and the host the client would dial can be two different things.

### Minimal remedy

The reported bug only needs the *page-origin* fallback; it does not need the injected base. Restricting it that way keeps all 35 tests green, including your injected-target case, and restores the four rows above:

```diff
     const effectiveBase =
       this.baseUrl ||
-      getInjectedWsBase() ||
-      (typeof window !== "undefined" ? window.location.origin : "");
+      (getInjectedWsBase()
+        ? ""
+        : typeof window !== "undefined"
+          ? window.location.origin
+          : "");
```

```
Tests  35 passed (35)
```

Please also extend the injected-target `it.each` with a dedicated-agent host (`wss://abc123.cloud.eliza.app`) and a control-plane host (`wss://cloud.eliza.app`) so the distinction is pinned — under the current diff those two rows fail, and under the patch above they pass.

If instead you *intend* the policy to widen to injected targets — the `isDedicatedCloudAgentBase` comment does argue those `/ws` upgrades 404 — then say so in the description rather than stating the opposite, and add the tests that assert the skip.

### Not a defect

`All Tests Passed` is red at this head, but the job's only failing step is `Require every admission lane failure`: the aggregate gate ran while `Source static smoke`, `Billing payment replay E2E`, and `Browser bridge Windows security` were cancelled and re-queued. No lane reported a test failure.

*(Method: single-suite runs via the package's configured vitest config; mutations applied by file swap, verified by diff, restored, `git status --porcelain` empty afterwards.)*

— [claude-code-eliza]

wtfsayo:
Thanks, this is a distinct case from the custom injected-target regression covered by the current suite. The empty-base guard should preserve an explicitly injected target even when its hostname matches the managed Cloud classifier. I am reproducing the Dedicated/control-plane injection matrix and narrowing the page-origin fallback accordingly before merge. I will post the updated head and regression results once validated.

ss251 (approved):
d2a90fe145bb45158218cc3c3f9a4950fb3f88e8

No blocking defect established in the supplied diff.

- `packages/ui/src/api/client-base.ts:2327`: An empty API base now selects the injected WebSocket target, if present, otherwise the page origin before applying the existing connection policy. This addresses same-origin cloud console retries without unconditionally suppressing WebSockets merely because the page is a cloud console. Nonempty bases retain priority.
- `packages/ui/src/api/client-base.ts:2554`: The exhaustion classification uses the same effective base as the initial policy check, keeping those two decisions consistent.
- `packages/ui/src/api/client-base-websocket.test.ts:319`: The new console cases assert zero WebSocket instances both after connecting and after resetting, rather than checking only a potentially misleading “connected” state. The subsequent cases cover selected-agent exhaustion and preservation of a separate injected realtime target. These are targeted regression assertions, not real-system acceptance evidence.

The patch is narrowly scoped to connection policy and its tests. No dependency, lifecycle-hook, CI-permission, or secret-access changes appear in the supplied diff. Helper implementations, runtime behavior, and artifact provenance cannot be established from these hunks alone.

VERIFICATION: Read the supplied two-file diff at this head; no commands or tests executed. Live execution blocked: no disposable sandbox on this host, so this is a static review of the diff at the exact head.

Recommendation: merge

wtfsayo:
Addressed the injected-target finding in `7fedd5a4550848e3b4da7e130457bf191c08eee5`.

Page-origin policy is now inferred only when both the API base and injected WebSocket target are absent. The 18-case matrix covers custom/Cloud page origins, empty/relative API bases, and custom/Dedicated/control-plane injected targets. It checks socket creation and the previous failed state after 15 rejected upgrades. Four matrix cases fail before the correction; all 50 WebSocket tests pass afterward.

The rebased head passes pinned installation and root verification. Updated browser evidence passes the attachment gate; final full app audit and hosted CI are running. The complete UI suite's failure entries all reproduce on clean baseline, documented in the PR body.

@attentionhead, please re-review the corrected injected-target policy.

attentionhead (approved):
Re-reviewed at the new head `7fedd5a4550848e3b4da7e130457bf191c08eee5` (fetched SHA confirmed equal to the live head). This follows up my `REQUEST_CHANGES` at `d2a90fe1`, where the guard resolved `this.baseUrl || getInjectedWsBase() || window.location.origin` while the real target resolution below it is injected-first — so the REST-only decision could be made against a base the socket would never dial.

The new head adopts the remedy exactly, and I verified it rather than assuming:

```ts
const effectiveBase =
  this.baseUrl ||
  (getInjectedWsBase() ? "" : typeof window !== "undefined" ? window.location.origin : "");
```

**The precedence now agrees with the transport.** With an injection present and no client base, `effectiveBase` is `""`, `shouldTreatAsConnectedWithoutWebSocket` declines, and control falls through to the resolution at `client-base.ts:2363-2419` — including the #20342 reconciliation that discards a `wsBase` merely normalizing to the page origin when a user-pinned HTTP base exists. With no injection, `effectiveBase` is the page origin and `isElizaCloudControlPlaneBase` still fires, which is the #18172 case this PR exists to preserve. The `this.baseUrl`-first branch is unchanged from `develop`, so no existing explicit-base policy moves.

**The added tests genuinely pin it.** I mutation-tested by restoring the pre-review `getInjectedWsBase() ||` form with everything else at head:

| `packages/ui` `src/api/client-base-websocket.test.ts` | result |
|---|---|
| head, unmodified | **50 passed** |
| `effectiveBase` reverted to injected-first | **4 failed / 46 passed** |

The four failures are precisely the injected-cloud-host cases (`wss://abc123.cloud.eliza.app` and `wss://cloud.eliza.app` on both page hosts with an empty API base) — the regression is held by assertion, not by coincidence. The wider `packages/ui/src/api/` sweep is clean at head: **106 files / 1300 tests passed**.

I also checked the injector inventory to make sure the new fall-through cannot reintroduce the churn through a different door. Every in-repo producer of `__ELIZA_WS_BASE__` yields either the page origin or a loopback: `appDevWsBasePlugin` (`packages/app/vite.config.ts:1797-1825`) is `apply: "serve"` and computes `(location.protocol==='https:'?'wss://':'ws://')+location.host`; `repairWebSameOriginWsBase` (`packages/app/src/web-ws-base-fix.ts:149-186`) rewrites only *foreign loopback* (`ws:`/`http:`) injections to same-origin and additionally sets the REST base, so `this.baseUrl` is non-empty on that path; the desktop shell keeps its loopback. None emits a `*.cloud.eliza.app` target, so the behaviour change is confined to out-of-repo/manual injections.

`MERGEABLE`, and the `UNSTABLE` roll-up is a single `Source static smoke` check still `IN_PROGRESS` — no failing check.

Approving.

## One non-blocking note for later

The new fixtures assert `getConnectionState().state === "failed"` after exhausting all 15 attempts against `wss://abc123.cloud.eliza.app` and `wss://cloud.eliza.app`. That is a deliberate choice and the added comment states it ("even when its hostname resembles a REST-only API host"), but it is worth recording that those two hostnames are not merely *resembling* — `client-base.ts:646-651` documents the dedicated-agent `/ws` upgrade as returning 404 as a property of the agent-router, and `:635-642` says the control-plane alias Worker structurally cannot serve `/ws`. So for anyone who does inject such a target, the test now codifies the ~95s "Reconnecting… (N/15)" walk into `failed` that #18172 removed for the same hosts reached the other way. Since nothing in-repo produces that injection today I do not think it should hold this PR; if a deployment ever starts injecting a dedicated-agent base, the host-derived skip will want to apply regardless of how the base was provided.

— [claude-code-eliza]

PR #30901, opened 2026-09-07
fix(app): audit current Cloud account states

wtfsayo:
The app audit now recognizes the current connected and signed-out Cloud views. Its old readiness check required an optional heading hidden by the normal page composition, so eight healthy captures always failed the strict aggregate.

Relates to #30899 and #18627. The Settings button capability fix is tracked separately in #30900.

The connected check requires visible status and account content; the signed-out check requires its explanation and recovery control. Each real browser capture must also reject the opposite account-state policy. The strict gate and loading/hidden-content checks remain intact. A unit test that only copied policy literals is replaced by the browser behavior check, and the OCR provenance fixture uses current rendered text without changing its provenance or regression assertions.

## Validation

- 192 audit unit tests passed.
- Eight real-bundle Cloud view cases passed across four viewport sizes, plus the existing negative readiness browser case.
- Eleven CloudView behavior tests passed.
- App typecheck, lint, and formatting passed.
- Root verification passed again on the committed revision. Full `audit:app` passed all 222 browser checks and packaged OCR: 204 verified, 12 needs-eyeball, zero broken, zero regressions. All eight affected Cloud views are OCR verified and manually inspected.

The browser account responses are deterministic fixtures. This validates the audit producer and rendered plugin states; it does not establish live account authentication or provisioning.

Definition of Done: full standard in `CONTRIBUTING.md`.

Base: `0128d7581b2c7870c3cf264d61e66b78556f381f`. Head: `4435b01928b516dfd0727a0ada23c6b7a71c1d21`. Installation completed with the pinned toolchain and no dependency changes. Independent review found no blockers in the final patch.

The supplementary [evidence.sqlite](https://github.com/elizaOS/eliza/releases/download/pr-evidence-4/30901-evidence.sqlite) contains queryable measurements. The document evidence rows below use the textual comparison and JSON reports because the document gate requires UTF-8 artifacts.

## Evidence Gate

[...]

Issue #30900, opened 2026-09-07
fix(cloud): allow declared plugin Settings recovery navigation


wtfsayo:
## Problem

The signed-out remote Cloud plugin view offers Connect in Settings, but clicking it leaves the view in place and renders “Navigation is unavailable right now.”

Discovered while validating #30899 through the real app audit at all four viewport sizes. `CloudView` invokes the host-brokered `navigateBrowserPath('/settings')`. The production view declaration in `plugins/plugin-elizacloud/src/index.ts` grants only `agent-surface`, so `brokerSurfaceNavigate` rejects the missing navigation capability. The matching smoke declaration preserves that same manifest; the audit only replaces its bundle path. This is a plugin-view recovery failure, not a claim that hosted Cloud login is broken.

## Acceptance

- The signed-out remote Cloud plugin view can navigate to Settings through its existing recovery button.
- Declare the required capability through the production manifest and corresponding declaration owner. Preserve the broker's denial for undeclared navigation and all existing target validation.
- Verify the real production bundle through the host broker at desktop and mobile sizes, with the visible Settings view reached.
- Keep Cloud readiness/audit corrections in #30899 separate from the product permission declaration repair.

wtfsayo:
Fixed in #30902 and deployed to staging in [run 34142406941](https://github.com/elizaOS/eliza/actions/runs/34142406941).

The real production plugin bundle reaches usable Settings through the host broker at all four viewport sizes. The final renderer audit passed 222 browser checks with no broken or needs-work findings; the PR includes the captures, recordings and logs. Production/smoke navigation parity and undeclared-capability denial are covered, and all hosted PR checks passed.

The staging certificate binds merged revision `e14071f00998f4d10f937015e3060196b991ad4d` and tree `98abbad15df15d64f1118c789ebc5212118905c4`. Live API health reports that revision; Pages freshness and routing verification passed.

The interaction evidence is from the real bundled local audit, not a hosted fresh-account run. An additional live browser attempt reached the staging login page, then browser-control timeouts prevented completing sign-in. That limitation does not change the scoped plugin/broker acceptance above; remaining fresh-account provisioning verification stays open in #18627.

Issue #15791, opened 2026-07-09
bug(evidence): make plugin-view smoke prove current production bundles


lalalune:
## Problem\n\nThe plugin-view UI-smoke stub still declares removed Shopify, Steward, and Social Alpha views and synthesizes a generic JavaScript bundle when a real `dist/views/bundle.js` is absent. Separately, `packages/scripts/build-views.mjs` discards `missingBundles` returned by the import guard. An audit can therefore show a healthy generic or in-process surface without proving the production bundle it claims to validate.\n\nThe Polymarket route demonstrates the gap: current app routing resolves its in-process app-shell registration before the remote registry bundle, so `/polymarket` can pass without requesting `/api/views/polymarket/bundle.js`.\n\n## Acceptance criteria\n\n- Smoke view declarations are derived from or checked against current production registrations; removed plugin IDs fail parity tests.\n- Audit mode never fabricates a generic bundle for a production-declared view; missing bundles fail observably.\n- `build-views` removes stale expected outputs before building and fails when any configured bundle is missing afterward.\n- Plugin-view evidence asserts which component/bundle rendered, including a network/provenance marker for dynamic bundles.\n- Focused tests prove a successful build command with no emitted bundle fails, a deleted plugin cannot remain in the stub, and a route cannot pass against the wrong component.\n- A fresh audit attaches bundle request/provenance logs and screenshots.

lalalune:
Queued behind #15790. The fresh artifact-scope run will establish trustworthy pixels first; this card then upgrades plugin-view provenance so those pixels prove the intended production bundle.

lalalune:
Additional reproducible provenance failure from a clean worktree: `run-ui-playwright.mjs` appends `--conditions=eliza-source` to the environment inherited by the child Vite build. Node then tries to execute `packages/cloud/routing/src/index.ts` directly and exits with `ERR_UNKNOWN_FILE_EXTENSION`. Explicitly building the same worktree outside that contaminated environment succeeds. The fix should scope the source condition to Playwright spec collection, not renderer-build children; add a clean-worktree regression to this card.

lalalune:
[...]

PR #30902, opened 2026-09-07
fix(cloud): enable plugin Settings recovery navigation

wtfsayo:
The signed-out remote Cloud plugin can now open Settings through its recovery button. Previously the button displayed “Navigation is unavailable right now” because the view called host navigation without declaring the required capability.

Relates to #30900 and #18627. The audit readiness repair is already merged as #30901.

The production Cloud manifest and matching smoke declaration now grant the existing `navigate` capability. This permits general shell navigation under the current capability model; the changed consumer uses the fixed `/settings` destination. The broker, undeclared-capability denial, and existing destination handling remain unchanged. No storage, external-link, billing, or lifecycle grant is added.

The browser regression clicks the actual bundled recovery button and waits for the Settings route, visible lifecycle slot, and rendered Settings content. The existing declaration checker now compares production and smoke navigation grants in both directions, and the required Source static smoke job runs that suite. It supports a bounded static declaration subset and explicitly rejects unresolved policies; it does not compare unrelated capabilities or execute plugin code.

Base `4fa1ed25ad9d4bb82ca8cdb4999b075a6591377d`; head `b1390d5eca2b75585fccc7a6081ae6932b6f2262`. Rebase preserved the reviewed patch. Pinned installation and post-rebase root verification passed. Definition of Done: full standard in `CONTRIBUTING.md`.

## Validation

- Root `bun run verify` passed 373/373 tasks on the rebased source.
- Plugin tests: 642 passed, five skipped; five real distribution checks passed.
- Broker grant/denial tests: 26 passed. Declaration parity: 28 passed with the exact CI command on the final head. Removing the grant or entire surface from either production or smoke now fails. Workflow contracts: six passed; actionlint 1.7.12 passed.
- Plugin and app-core typecheck/lint passed.
- Real bundled navigation reached usable Settings at all four viewport sizes. Desktop/mobile before-error and after-Settings images, OCR, recordings, and console/network evidence are attached below.
- Two independent reviews cleared the capability declaration, browser interaction, and final parity/CI follow-up after resolving shadowing and direct-mutation false positives.
- The final renderer built at `b1390d5eca` and its full combined app audit passed: 222 browser checks, zero broken or needs-work findings, and packaged OCR with 204 verified, 12 needs-eyeball, zero broken or regressions. All eight Cloud views are OCR verified and manually inspected; all four Settings CTA flows passed. Exact-head re-review approved. Hosted CI passed all five required checks in run [34139016468](https://github.com/elizaOS/eliza/actions/runs/34139016468).

Broader owner suites are nongreen in untouched areas. The app Bun lane passed 950 tests with two skips; its Vitest lane passed 1,626 and failed six renderer-HMR/iOS-source assertions. App-core passed 4,547, skipped 25, failed three assertions, and had three loader errors involving provider-model expectations, desktop catalogs, and Node/Bun test discovery. The affected source/tests/configuration are unchanged from the base, with narrow failures reproduced; a complete clean-base run of both broad suites was not performed. These failures are disclosed rather than counted as passing validation.

The before captures use the real production plugin bundle and host broker with deterministic local HTTP fixtures and a reused renderer. Final after screenshots and the full audit use the renderer freshly built at `b1390d5eca`, with matching build stamps in all four CTA traces. The walkthrough recordings retain their `29c401fe` provenance; the navigation plugin, broker, and CTA test are unchanged. This demonstrates grant/denial and navigation behavior, not hosted login or fresh-account provisioning.

Earlier `29c401fe` audit (preserved): [exact-29c-renderer-build.json](https://github.com/elizaOS/eliza/releases/download/pr-evidence-4/30902-exact-29c-renderer-build.json) · [exact-29c-audit.log](https://github.com/elizaOS/eliza/releases/download/pr-evidence-4/30902-exact-29c-audit.log) · [exact-29c-capture-report.json](https://github.com/elizaOS/eliza/releases/download/pr-evidence-4/30902-exact-29c-capture-report.json) · [exact-29c-ocr-triage.json](https://github.com/elizaOS/eliza/releases/download/pr-evidence-4/30902-exact-29c-ocr-triage.json)

## Evidence Gate

- [x] [interaction-v1-before-desktop.png](https://github.com/elizaOS/eliza/releases/download/pr-evidence-4/30902-interaction-v1-before-desktop.png) · [interaction-v1-before-mobile.png](https://github.com/elizaOS/eliza/releases/download/pr-evidence-4/30902-interaction-v1-before-mobile.png)

- [x] [exact-b139-settings-desktop.png](https://github.com/elizaOS/eliza/releases/download/pr-evidence-4/30902-exact-b139-settings-after-cta-desktop-landscape.png) · [exact-b139-settings-mobile.png](https://github.com/elizaOS/eliza/releases/download/pr-evidence-4/30902-exact-b139-settings-after-cta-mobile-portrait.png)

- [x] [exact-29c-after-desktop.mp4](https://github.com/elizaOS/eliza/releases/download/pr-evidence-4/30902-exact-29c-after-desktop.mp4) · [exact-29c-after-mobile.mp4](https://github.com/elizaOS/eliza/releases/download/pr-evidence-4/30902-exact-29c-after-mobile.mp4)

- [x] N/A - no backend code changes; local fixture request evidence is included with frontend logs.

- [x] [exact-29c-console-network.txt](https://github.com/elizaOS/eliza/releases/download/pr-evidence-4/30902-exact-29c-console-network.txt)

- [x] N/A - no model, prompt, or agent behavior changes.

- [x] N/A - no persisted domain-state mutations.

- [x] [exact-b139-review.md](https://github.com/elizaOS/eliza/releases/download/pr-evidence-4/30902-exact-b139-review.md) · [interaction-v1-ocr.txt](https://github.com/elizaOS/eliza/releases/download/pr-evidence-4/30902-interaction-v1-ocr.txt)

Final parity/CI proof: [follow-up scope](https://github.com/elizaOS/eliza/releases/download/pr-evidence-4/30902-parity-b139-scope.md) · [focused checks](https://github.com/elizaOS/eliza/releases/download/pr-evidence-4/30902-parity-b139-focused.log) · [root verification](https://github.com/elizaOS/eliza/releases/download/pr-evidence-4/30902-parity-b139-verify.log). Exact-head re-review approved; hosted CI passed.

Final combined build and audit: [renderer manifest](https://github.com/elizaOS/eliza/releases/download/pr-evidence-4/30902-exact-b139-renderer-build.json) · [audit log](https://github.com/elizaOS/eliza/releases/download/pr-evidence-4/30902-exact-b139-audit.log) · [capture report](https://github.com/elizaOS/eliza/releases/download/pr-evidence-4/30902-exact-b139-capture-report.json) · [OCR](https://github.com/elizaOS/eliza/releases/download/pr-evidence-4/30902-exact-b139-ocr-triage.json) · [inspection and run provenance](https://github.com/elizaOS/eliza/releases/download/pr-evidence-4/30902-exact-b139-review.md). The first capture had one Goals asset-load failure with `ERR_NETWORK_CHANGED`; its focused retry and the full 222-case repeat passed without code changes.

Merged as `e14071f00998f4d10f937015e3060196b991ad4d`. The exact merged revision deployed successfully to staging in [run 34142406941](https://github.com/elizaOS/eliza/actions/runs/34142406941). API health and the staging certificate match the merge; Pages freshness and routing checks passed. Scoped issue #30900 is closed with [verification details and browser limits](https://github.com/elizaOS/eliza/issues/30900#issuecomment-5573522821).

### DIFF
--- a/.github/workflows/pr-static-smoke.yml
+++ b/.github/workflows/pr-static-smoke.yml
@@ -142,6 +142,10 @@ jobs:
           packages/scripts/__tests__/github-action-pinning.test.ts
           packages/scripts/__tests__/pr-static-smoke-workflow.test.ts
 
+      - name: Check production and smoke view navigation parity
+        working-directory: packages/app-core
+        run: node ../scripts/run-vitest.mjs run --config vitest.config.ts scripts/smoke-view-declarations.test.ts
+
       - name: Upload Bun runtime resolution inventory
         uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a
         with:
--- a/packages/app-core/scripts/smoke-view-declarations.mjs
+++ b/packages/app-core/scripts/smoke-view-declarations.mjs
@@ -39,7 +39,7 @@ export const smokeViewDeclarations = [
     "/cloud",
     "CloudView",
     "gui",
-    { header: "fullscreen", capabilities: ["agent-surface"] },
+    { header: "fullscreen", capabilities: ["agent-surface", "navigate"] },
   ],
   ["contacts", "Contacts", "plugin-contacts", "/contacts", "ContactsView"],
   // The decomposed personal-assistant domain views are the real surfaces (the
--- a/packages/app-core/scripts/smoke-view-declarations.mjs
+++ b/packages/app-core/scripts/smoke-view-declarations.mjs
@@ -9,7 +9,9 @@
  * surface for a view production no longer registers — proving nothing. So the
  * declarations live here next to `checkSmokeViewParity`, which fails the moment
  * a declared view's plugin directory is gone or no longer exports the named
- * component. Removed plugin IDs (Shopify, Steward, Social Alpha) are therefore
+ * component. Navigation grants must also match in both directions; unrelated
+ * surface grants are outside this check. Removed plugin IDs
+ * (Shopify, Steward, Social Alpha) are therefore
  * kept out and cannot silently reappear.
  *
  * `resolveBundleProvenance` is the single decision the stub uses when serving a
@@ -152,6 +154,22 @@ function readSourceFiles(dir) {
   return sources;
 }
 
+function propertyInitializer(object, propertyName) {
+  // The last matching property wins. A later spread can replace the policy.
+  for (const property of [...object.properties].reverse()) {
+    if (ts.isSpreadAssignment(property)) return null;
+    const name = property.name;
+    const key =
+      name && (ts.isIdentifier(name) || ts.isStringLiteralLike(name))
+        ? name.text
+        : undefined;
+    if (key === undefined) return null;
+    if (key !== propertyName) continue;
+    return ts.isPropertyAssignment(property) ? property.initializer : null;
+  }
+  return undefined;
+}
+
 function stringProperty(object, propertyName) {
   for (const property of object.properties) {
     if (!ts.isPropertyAssignment(property)) continue;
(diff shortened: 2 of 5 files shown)


OutstandingRetard (approved):
Scoped review of #30902 (fix(cloud): enable plugin Settings recovery navigation).
Touches 4 file(s): `packages/app-core/scripts/smoke-view-declarations.mjs`, `packages/app-core/scripts/smoke-view-declarations.test.ts`, `packages/app/test/ui-smoke/all-views-aesthetic-audit.spec.ts`, `plugins/plugin-elizacloud/src/index.ts`. Lead change `packages/app-core/scripts/smoke-view-declarations.mjs` (+1/-1).
No blocking concerns.

attentionhead (changes requested):
Verified at head `29c401fe04ab20b43809fd0a12e74eafed183008` (fetched SHA confirmed equal to the live head before reviewing). I am the first reviewer.

## The fix is correct, and the root cause checks out

I confirmed the failure mode end to end rather than taking the description on trust:

- `packages/ui/src/surface-realm-broker.ts:221-241` — `brokerSurfaceNavigate` throws `SurfaceRealmDeniedError` when `!surfaceGrants(manifest, "navigate")`, so the call never reaches the shell.
- `plugins/plugin-elizacloud/src/components/cloud/CloudView.tsx:349-356` — `navigateInternal` wraps that call in `try/catch` and sets `"Navigation is unavailable right now."`. That is exactly the observed symptom, and adding the grant is the minimal correct remedy.
- `CloudView.tsx:440` is the **only** `navigateInternal` call site in the plugin, and its argument is the literal `"/settings"` — no user- or server-controlled path reaches the newly granted vector.

On the security caveat in the description: it is accurate and I could not find a narrower alternative. `navigate` is all-or-nothing — beyond `scope.navigate`, `assertRawHistoryMutationAllowed` (`surface-realm-broker.ts:548-583`) returns early for any granted view, so raw `history.pushState`/`replaceState` to an arbitrary pathname **and query string** is unlocked by the same flag. Given that CloudView's single destination is a fixed literal, I do not consider that a blocker, but it does raise the cost of a silent regression — which is what the rest of this review is about.

## Blocking: nothing that runs pins the grant this PR adds

The PR changes the capability in three places, but I mutation-tested each side separately and only one of them is actually held:

| mutation (each applied alone, then reverted) | `packages/app-core` `scripts/smoke-view-declarations.test.ts` |
|---|---|
| head, unmodified | **8 passed** |
| drop `"navigate"` from `plugins/plugin-elizacloud/src/index.ts` | **8 passed** — regression not caught |
| drop `"navigate"` from `packages/app-core/scripts/smoke-view-declarations.mjs` | 1 failed / 7 passed |

The production manifest — the one `brokerSurfaceNavigate` actually reads — can be reverted with the whole suite still green. The assertion added at `smoke-view-declarations.test.ts:46-56` compares the `.mjs` literal against a literal copy of itself in the same package; it pins the smoke stub, not the plugin. `checkSmokeViewParity` only ever checks directory existence, `id`/`path`, and `componentExport`, so `surface.capabilities` is unenforced in both directions — which is precisely the drift the file's own header comment warns about ("omitting one makes the smoke registry less capable than the production manifest and can turn a valid bridge action into a false denial").

The other three coverage paths do not close the gap either:

- `plugins/plugin-elizacloud/src/components/cloud/CloudView.test.tsx` stubs `navigateInternal` with `vi.fn()` (`:78`) and, for the negative case, a hand-thrown seam (`:156-165`). It never constructs a real manifest, so it passes identically with or without the grant.
- `packages/app/test/ui-smoke/all-views-aesthetic-audit.spec.ts` — the new browser regression, and the only assertion here that exercises the real broker — appears in **no** file under `.github/`. It gates nothing.
- `smoke-view-declarations.test.ts` is not in CI's app-core lane either: `.github/workflows/test.yml:728` runs `bun run --cwd packages/app-core test -- test/app/screenshot-quality.test.ts`, i.e. filtered to a single unrelated file, and the `test:script-suites` half of that script enumerates its `node --test` files by name without including this one.

So the net CI delta of this PR is zero, and the regression it fixes can return silently.

### Suggested remedy (small, and local to code already here)

`inspectViewDeclarations` already walks the plugin source with the TypeScript AST and matches the exact object literal by `id` + `path`. In `plugins/plugin-elizacloud/src/index.ts:344-362` that same literal carries `surface: { header: "fullscreen", capabilities: [...] }`. Reading `surface.capabilities` off the node it has already found and comparing it to the declaration tuple's grants — reporting e.g. `surface-capabilities-mismatch` alongside the existing `component-export-missing` — turns the parity check into a real two-sided pin and makes mutation A fail. Adding the audit spec to a workflow would be the stronger fix, but the parity extension alone is enough to hold this change.

## Not verified

`CloudView.test.tsx` would not execute in my checkout: `@elizaos/login` is not linked into `node_modules`, and after linking it the `eliza-source` export condition pulls `packages/login/src/sdk/auth.ts`, whose `bs58` import is absent. Both are workspace-install gaps on my side, reproduce on the merge base, and are **not** attributable to this PR — but it does mean I read that file rather than ran it. I also did not run the Playwright audit spec itself; my claim about it is only that no workflow references it.

Note this PR is stacked on #30901, so `git diff origin/develop...` also shows that PR's OCR-policy files; I reviewed only the four files listed by `gh pr view --json files`.

— [claude-code-eliza]

wtfsayo:
The production/smoke grant drift is a valid gap. I am reproducing the production-manifest mutation and extending the existing declaration parity boundary to compare capabilities, rather than relying on the copied smoke expectation. I am also tracing the actual check invocation so the regression is exercised by the relevant verification path. The real bundled CTA evidence remains valid for the product fix; the parity correction will be reviewed and validated before merge.

ss251 (approved):
Reviewed head 29c401fe04ab20b43809fd0a12e74eafed183008.

The risky part is declaration parity: `plugins/plugin-elizacloud/src/index.ts:360`, `packages/app-core/scripts/smoke-view-declarations.mjs:42` and `smoke-view-declarations.test.ts:56` must all carry the same surface object or the #15791 parity test fails. I compared the three hunks token-for-token: each is exactly `{ header: "fullscreen", capabilities: ["agent-surface", "navigate"] }`, same order, no other capability dropped.

The new Playwright block at `all-views-aesthetic-audit.spec.ts:2430` is gated on `view.fixtureState === "cloud-signed-out"`, so signed-in fixtures (where "Connect in Settings" is absent) never hit the click. The URL regex `/\/settings(?:[?#]|$)/` accepts `/settings`, `/settings?tab=x`, `/settings#y` and rejects `/settings/foo`, which is the correct shape for a recovery landing. Non-blocking: the diff does not touch CloudView itself, so this relies on the existing button already invoking the host navigate API that the new capability unlocks.

Verified by reading the diff at this head and cross-checking the three declaration sites against each other.

Recommendation: merge

wtfsayo:
Addressed at `b1390d5eca2b75585fccc7a6081ae6932b6f2262`, rebased onto the merged connection fix.

The existing checker now compares the production and smoke navigation grant in both directions, and the required Source static smoke job invokes that suite. Removing either grant or its entire surface fails. Unsupported dynamic, shadowed, or directly mutated named policies fail explicitly within the documented static subset; unrelated capabilities remain outside this check.

28 focused tests, six workflow contracts, actionlint, and root verification (373/373 tasks) passed. Two independent reviewers cleared the correction after negative probes. Updated scope and logs are linked in the PR body; the earlier browser media retains its actual captured revision.

@attentionhead ready for re-review. Hosted checks are running on the new head.

attentionhead (approved):
Re-reviewed at the new head `b1390d5eca2b75585fccc7a6081ae6932b6f2262` (fetched SHA confirmed equal to the live head). Both findings from my `REQUEST_CHANGES` at `29c401fe` are addressed, and I verified each by re-running the exact mutation that motivated it rather than reading the new code and assuming.

## Finding 1 — the production grant is now pinned

The parity check previously compared the smoke `.mjs` literal against a literal copy of itself, so the manifest `brokerSurfaceNavigate` actually reads could be reverted with the suite green. `checkSmokeViewParity` now resolves `surface.capabilities` off the same AST node it already matched by `id` + `path` and reports `surface-navigation-mismatch`. Re-running the two mutations from my last review, each applied alone:

| mutation | at `29c401fe` | at `b1390d5e` |
|---|---|---|
| drop `"navigate"` from `plugins/plugin-elizacloud/src/index.ts` only | **8 passed** — the regression I reported | **fails** `surface-navigation-mismatch` |
| drop `"navigate"` from `smoke-view-declarations.mjs` only | 1 failed / 7 passed | **fails** `surface-navigation-mismatch` |

Both directions now bite. Baseline at head is 28 passed, up from 8.

The resolver is more than a literal read, so I probed its edges rather than trusting the happy path. Refactoring the plugin's surface into a module-level `const … as const` and referencing it still resolves correctly — `checkSmokeViewParity` returns `missing: []`, so the check follows named constants instead of only inline literals. And the fail-closed branch is real coverage, not decoration: disabling the `grantsNavigation === null` arm fails 5 tests spanning cyclic constants, a policy reference shadowed inside a function, and three mutable-or-escaped policy shapes. Choosing `surface-navigation-policy-unresolved` over a silent `false` is the right default for a security-relevant grant.

## Finding 2 — it now gates CI

`.github/workflows/pr-static-smoke.yml` gains a `Check production and smoke view navigation parity` step in the `source-smoke` job. That workflow triggers on `pull_request` to `develop`/`main` with no path filter and describes itself as the sole PR validation authority, so unlike the `test.yml:728` app-core lane (filtered to `test/app/screenshot-quality.test.ts`) this actually runs on every PR. `packages/scripts/__tests__/pr-static-smoke-workflow.test.ts` passes 6/6, so the added step does not violate the workflow's own contract.

The underlying change is unchanged and still correct: `capabilities: ["agent-surface", "navigate"]` is the minimal grant that lets `CloudView`'s single `navigateInternal("/settings")` call through `brokerSurfaceNavigate`, and the security caveat in the description remains accurate — `navigate` is all-or-nothing and there is no destination-scoped variant.

Approving.

## One note, not blocking

The resolver requires `as const` on a named policy. A plain module-level `const CLOUD_SURFACE_POLICY = { header: "fullscreen", capabilities: ["agent-surface", "navigate"] };` — no `as const` — is rejected as `surface-navigation-policy-unresolved`, because `immutablePolicyInitializer` accepts only a string literal or an `as const` assertion. That is the safe direction and no shipped plugin uses the shape today, so parity is clean at head; but it is a false positive rather than a real ambiguity, and the reason string does not hint that adding `as const` is the fix. Worth either accepting a plain object-literal initializer (the mutation/escape analysis in `hasUnsupportedPolicyUse` is what actually establishes immutability here, and it runs either way) or naming `as const` in the message.

## Not verified

I did not run the Playwright audit spec; the `all-views-aesthetic-audit.spec.ts` addition is unchanged from my last pass and still appears in no workflow, so that particular assertion remains ungated — the parity test is what carries the guarantee now. `packages/app-core` `tsc --noEmit` reports 3 errors, but they reproduce identically on `develop` (unbuilt sibling packages `@elizaos/plugin-openai/endpoint-config` and `@elizaos/capacitor-bun-runtime`), so they are pre-existing and not attributable here. Note this branch now also carries #30898's `client-base.ts` changes, which I reviewed separately and approved at `7fedd5a4`.

— [claude-code-eliza]

Issue #30905, opened 2026-09-07
Staging worker diagnostics miss health frames emitted by Node and the production redactor


wtfsayo:
## Problem

The existing read-only staging worker diagnostic reports health-timeout markers but cannot decode the deployed worker’s detailed diagnostic frames. This blocks diagnosis of the remaining sandbox readiness failure in #18627.

The worker runs under Node and sends its context object through the production `redactLogArgs` function. Node prints a null-prototype object header, single-quoted strings, and concatenated lines for multiline strings. The collector expects the different five-line ordinary-object format produced by bare Bun console output.

A synthetic reproduction using the real redactor and pinned Node reproduces category matches with zero decoded frames. The latest [read-only staging diagnostic](https://github.com/elizaOS/eliza/actions/runs/34141656421) similarly reports health-timeout markers but no decoded frames. No raw journal data or credentials are needed to reproduce this.

## Acceptance

- Decode complete health diagnostic frames emitted by the real production redactor and Node console sink.
- Preserve the existing target-container association and categorical-only output; reject malformed, interleaved, or incomplete frames without evaluating log text.
- Preserve supported existing log formats and test the actual producer-to-parser boundary rather than copying a hand-written log literal.
- Rerun the existing source-locked, read-only staging diagnostic against the unchanged daemon and inspect its sanitized result.

This does not change provisioning behavior, restart services, or relax readiness deadlines.

agent-cortex:
🤖 An autonomous contributor agent is starting work on this issue.

Bounded outcome: a PR against `develop` from branch `fix/staging-worker-diagnostic-node-frames-30905` (fork `agent-cortex`), with tests and evidence bound to the measured head.

Defer contract: if you claimed, were assigned, or had started this issue before this comment — or you are the issue reporter and intend to implement it yourself — reply here and this claim is released promptly. Otherwise this claim holds first-come priority and a PR will follow shortly.

AI provider/model: anthropic / claude-opus-4-8
Client / agent tooling: pi-coding-agent
Contribution skill revision: elizaOS/eliza@e14071f00998f4d10f937015e3060196b991ad4d:packages/skills/skills/contribute-to-eliza
Attribution status: self-reported
— [eliza-swarm]

agent-cortex:
Opened #30906 to fix this.

[...]

PR #30907, opened 2026-09-07
fix(cloud): decode Node worker health diagnostic frames

wtfsayo:
The staging worker diagnostic now decodes complete health frames emitted by the production redactor and Node console. Previously those messages were counted but their health details were discarded because the collector expected Bun formatting.

Relates to #30905 and #18627. The fix supports Node's null-prototype object header, quoted strings and multiline string concatenation, while retaining the existing Bun format. It rejects incomplete, interleaved, abbreviated or unsupported expressions without evaluating them. Target association requires a complete frame; the report remains categorical and does not expose container names, node identifiers or log contents.

The worker, provisioning behavior, health deadlines, source/process checks and deployment permissions are unchanged. The updated collector can run through the existing read-only diagnostic workflow against the unchanged deployed worker.

Exact head `035339070b1c0f86bb4c4cac8d16637b34241d26`, based on `e14071f00998f4d10f937015e3060196b991ad4d`. Pinned installation and root `bun run verify` passed. Real producer/collector tests passed 14/14; independent downstream caller/workflow tests passed 30/30. Touched-file Biome and whitespace checks passed. Two independent reviews found no blocking issues. Definition of Done: full standard in CONTRIBUTING.md.

The same synthetic production-redactor → Node 24.15.0 sample decodes zero frames on the base and one correctly classified frame on this change. This proves the serialization boundary locally. The previous [read-only live diagnostic](https://github.com/elizaOS/eliza/actions/runs/34141656421) counted timeout messages but decoded no frames; the corrected live run remains pending after merge. No live sandbox root cause is claimed.

[Verification details](https://github.com/elizaOS/eliza/releases/download/pr-evidence-4/30905-035339-review.md) · [producer comparison](https://github.com/elizaOS/eliza/releases/download/pr-evidence-4/30905-035339-producer-boundary.txt) · [focused tests](https://github.com/elizaOS/eliza/releases/download/pr-evidence-4/30905-035339-focused.log) · [root verification](https://github.com/elizaOS/eliza/releases/download/pr-evidence-4/30905-035339-verify.log)

## Evidence Gate

- [x] N/A - standalone diagnostic collector has no rendered interface.

- [x] N/A - standalone diagnostic collector has no rendered interface.

- [x] N/A - no browser or native interaction changes.

- [x] [Actual producer/collector command transcript](https://github.com/elizaOS/eliza/releases/download/pr-evidence-4/30905-035339-producer-test-transcript.log), using synthetic diagnostic content through the production redactor and actual Node console. This is local integration evidence; live diagnostic follow-up remains pending.

- [x] N/A - no frontend code or network interaction changes.

- [x] N/A - no model, prompt or runtime agent behavior changes.

- [x] N/A - no persisted domain changes; diagnostic report evidence is linked above.

- [x] N/A - no rendered visual output.

Merged as `141e35ae98cfee1499ca2c4d064e4ee2546e16dc` after all hosted checks passed in [run 34144163962](https://github.com/elizaOS/eliza/actions/runs/34144163962). The corrected [read-only staging diagnostic](https://github.com/elizaOS/eliza/actions/runs/34144836420) passed against the unchanged deployed worker. It decoded three complete target frames with zero malformed frames; all report a missing container on the checked host. The deployment job was skipped. #30905 is closed; the remaining sandbox lifecycle investigation stays in #18627.

### DIFF
--- a/packages/cloud/scripts/admin/staging-worker-diagnostic.mjs
+++ b/packages/cloud/scripts/admin/staging-worker-diagnostic.mjs
@@ -82,58 +82,113 @@ function healthObservation(diagnostics) {
   };
 }
 
+// Decode only string literals emitted by console inspection, never expressions.
+// The suffix remains unconsumed so abbreviation and executable syntax are rejected.
+function consoleString(text) {
+  const quote = text[0];
+  if (!['"', "'", "`"].includes(quote)) return null;
+  let value = "";
+  for (let index = 1; index < text.length; index += 1) {
+    const character = text[index];
+    if (character === quote) return { value, suffix: text.slice(index + 1) };
+    if (
+      character.charCodeAt(0) < 32 ||
+      (quote === "`" && character === "$" && text[index + 1] === "{")
+    )
+      return null;
+    if (character !== "\\") {
+      value += character;
+      continue;
+    }
+    const escapedCharacter = text[++index];
+    const escapes = { n: "\n", r: "\r", t: "\t", b: "\b", f: "\f", v: "\v" };
+    if (Object.hasOwn(escapes, escapedCharacter))
+      value += escapes[escapedCharacter];
+    else if (['"', "'", "`", "\\", "/"].includes(escapedCharacter))
+      value += escapedCharacter;
+    else if (escapedCharacter === "0" && !/[0-9]/.test(text[index + 1] ?? ""))
+      value += "\0";
+    else if (escapedCharacter === "x" || escapedCharacter === "u") {
+      const length = escapedCharacter === "x" ? 2 : 4;
+      const digits = text.slice(index + 1, index + 1 + length);
+      if (digits.length !== length || !/^[0-9a-f]+$/i.test(digits)) return null;
+      value += String.fromCharCode(Number.parseInt(digits, 16));
+      index += length;
+    } else return null;
+  }
+  return null;
+}
+
 /** Parses only complete console frames; unexpected/interleaved lines discard attribution. */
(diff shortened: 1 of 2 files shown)


ss251 (approved):
Reviewed head 035339070b1c0f86bb4c4cac8d16637b34241d26.

The risky hunk is `consoleString` in `packages/cloud/scripts/admin/staging-worker-diagnostic.mjs:87-121` plus the phase machine below it. I traced the two boundaries most likely to misclassify a real Node frame as malformed: (1) the `${` rejection for backtick literals — Node's `strEscape` only picks backticks when the string contains `'` and `"` but NOT `${`, otherwise it falls back to `'` with escaping, so a genuine Node frame can never hit that `return null`; the check only rejects non-console input. (2) the multi-line split: Node joins chunks with ` +` and indents continuation lines by four spaces, which matches `suffix === " +"` -> `continuation` -> prefix `"    "`, and the trailing empty chunk `''` (from `NODE_DIAGNOSTICS` ending in `\n`) decodes to `{value:"", suffix:""}` -> `close`. `\x`/`\u` decoding uses `fromCharCode`, so lone `\uD800` round-trips instead of throwing, unlike the old `JSON.parse` path. `text[++index]` on a trailing backslash yields `undefined` and falls to `return null`.

Non-blocking: the `escapes` table is rebuilt per escape; trivially hoistable.

Verified by reading the diff at this head and cross-checking Node's `util.inspect` quote-selection and string-splitting rules against each phase transition.

Recommendation: merge

PR #30910, opened 2026-09-07
fix(cloud): check Docker candidates on their captured placement

### LINKED ISSUE #30908: Docker candidate health checks can revert to the old canonical placement
## Problem

A newly created Docker candidate can be health-checked on the old canonical node before the lifecycle commits the new placement. This can turn a healthy candidate into a not-ready failure.

`create()` caches the new placement, and provisioning calls `checkHealthDetailed(handle)` before committing it to the canonical sandbox row. The SSH poll refreshes placement through `hydrateContainerFromDb()` on each iteration. If the retained canonical row still names the old node, it replaces the fresh candidate metadata. A primary-database read alone would not fix this ordering.

A deterministic reproduction exercises the real SSH health poll with controlled database, SSH and clock boundaries: fresh placement points to a healthy new node; the canonical row still points to the old node where the container is absent. The poll contacts only the old node and returns `not_ready`. This is a source-contract reproduction, not live end-to-end evidence.

Discovered while investigating #18627 after the corrected read-only diagnostic recovered missing-container observations. Its causal relationship to that staging account is not established; historical placement evidence is unavailable.

## Acceptance

- Health checking an in-flight candidate uses that candidate's placement until the lifecycle commits or rejects it.
- Preserve health polling's existing ability to follow a genuinely newer canonical placement for an established sandbox, including the #15203 regression contract.
- Keep deletion, replacement cleanup, tenant authority and lifecycle fencing intact; do not broadly disable placement refresh or relax health checks.
- Reproduce old canonical row/new candidate behavior and exercise both candidate and established-sandbox paths through the owning provider/lifecycle tests.
- Clearly separate deterministic integration proof from live staging acceptance. Do not infer missing-container history from the current database row.

wtfsayo:
# Relates to

Fixes #30908. Related investigation: #18627; its staging recovery remains unproven.

Definition of Done: full standard in CONTRIBUTING.md.

- [x] Targets develop; current base141e35 was fetched and confirmed before commit, with no rebase needed.
- [x] Pinned installation and final root verification passed.
- [x] Reviewed evidence shows the changed behavior and its limits.

# Sync with develop

- [x] Current origin/develop141e35; zero conflicts.
- [x] Root verify passed on the final frozen source.

# Risks

Medium: Docker readiness participates in provisioning and image cutover. Candidate checks now require complete matching captured placement and SSH metadata. Established health checks retain canonical refresh; cleanup, lifecycle fences, readiness verdicts and time budgets remain intact. No schema or environment change.

# Background

## What does this PR do?

A fresh Docker candidate is checked on the node returned by creation. Previously the poll could replace that placement with the predecessor's canonical row, rejecting a healthy candidate on a missing old host or accepting an unhealthy candidate because the old host was healthy.

Fresh provisioning and pre-cutover upgrades/downgrades select candidate authority. Persisted retries and established-sandbox reconciliation retain canonical refresh. Invalid candidate identity or placement fails explicitly.

Bug fix with an additive internal health context parameter.

No external configuration or user documentation changes. The internal contract and evidence describe the distinction.

# Testing

Review the local integration results and scope notes below. The real production SSH client executes curl/Docker through a loopback SSH adapter against disposable Docker containers, with controlled database placement.

## Detailed testing steps

Pinned Bun1.3.14 / Node24.15.0: normal install, root verify twice, final cloud-shared typecheck/lint all passed. Focused provider/provision/upgrade/downgrade suites passed84 tests; final provider suite passed14 tests after the complementary negative case. Two independent reviews found no blockers.

The unchanged provider fails the complementary negative regression: an unhealthy candidate is falsely ready because its canonical predecessor is healthy. The fixed provider returns not_ready. The real local proof passed candidate-ready with zero canonical reads, established-ready with refresh, and absent-candidate not_ready after the unchanged360-second budget.

# Evidence Gate

- [x] N/A - backend health placement only; no rendered surface changes.

- [x] N/A - backend health placement only; no rendered surface changes.

- [x] N/A - backend health placement only; no visual user flow changes.

- [x] [Real local SSH/Docker transcript](https://github.com/elizaOS/eliza/releases/download/pr-evidence-4/30908-exact-24565085-ssh-transcript.log) and [scope/provenance](https://github.com/elizaOS/eliza/releases/download/pr-evidence-4/30908-exact-24565085-ssh-proof.md).

- [x] N/A - no browser or frontend request behavior changed.

- [x] N/A - Docker infrastructure health placement; no model, prompt, action, or inference behavior changes.

- [x] [Local Docker health results](https://github.com/elizaOS/eliza/releases/download/pr-evidence-4/30908-exact-24565085-ssh-results.json), [validation inventory](https://github.com/elizaOS/eliza/releases/download/pr-evidence-4/30908-exact-24565085-validation.json), and [review record](https://github.com/elizaOS/eliza/releases/download/pr-evidence-4/30908-exact-24565085-review.md).

- [x] N/A - no rendered visual surface changes.

# Evidence Details

## Real LLM-call trajectory

N/A - infrastructure placement selection, with no model behavior changes.

## Backend + frontend logs

Backend artifacts above exercise real local SSH, curl and Docker. Database placement is controlled; the loopback SSH server is a test adapter. No Eliza runtime is booted and no hosted staging recovery is claimed. Frontend: N/A - no frontend changes.

### Before

N/A - no rendered UI changes.

### After

N/A - no rendered UI changes.

### Walkthrough video

N/A - no visual user flow changes.

## Audio / voice walkthrough

N/A - no audio changes.

## Known gaps / failures

The normal full cloud-shared test runner stopped at batch45/388 with two shared-runtime tests failing: the trusted lifecycle reply assertion and TODO reply grounding. A current unchanged baseline reproduced the exact same two failures (27passed/2failed); its tree equals develop141e35. Later batches were not executed, so the full package lane is not green.

An initial owner run overlapped dependency builds and encountered missing generated modules; it is retained as contaminated, not source evidence. The serialized rerun above owns the result. An early missing login build artifact was resolved by the normal root build; final typecheck passed without a source workaround.

The local integration harness rejected the optional timeout-diagnostics command through its command allowlist; the readiness verdict still completed normally. The source-locked provider diff hash matches the committed source. The original staging missing-container observations do not establish historical placement or this bug's causal relationship to that account. No paid provisioning or daemon deployment was performed for this PR.

### DIFF
--- a/packages/cloud/shared/src/lib/services/__tests__/docker-sandbox-health-stale-node.test.ts
+++ b/packages/cloud/shared/src/lib/services/__tests__/docker-sandbox-health-stale-node.test.ts
@@ -139,3 +139,103 @@ describe("pollSshDockerHealth follows agent re-placement (#15203)", () => {
     getClient.mockRestore();
   });
 });
+
+describe("candidate health before canonical placement commit", () => {
+  const handle = {
+    sandboxId: `agent-${NEW_NODE.agentId}`,
+    bridgeUrl: "http://candidate.example",
+    healthUrl: "http://candidate.example/api",
+    metadata: {
+      provider: "docker",
+      ...NEW_NODE,
+      containerName: `agent-${NEW_NODE.agentId}`,
+      nodeSshPort: 2222,
+      nodeSshUser: "candidate-user",
+      nodeHostKeyFingerprint: "SHA256:candidate-pin",
+    },
+  };
+
+  test("fresh candidate probes its captured node despite retained old canonical placement", async () => {
+    const provider = new DockerSandboxProvider();
+    const internals = provider as unknown as PollInternals;
+    const hydrate = spyOn(internals, "hydrateContainerFromDb").mockResolvedValue(OLD_NODE);
+    const { getClient, probedHosts } = fakeSshByHost(NEW_NODE.hostname);
+    expect(await provider.checkHealth(handle, { kind: "candidate" })).toBe(true);
+    expect(probedHosts).toEqual([NEW_NODE.hostname]);
+    expect(getClient).toHaveBeenCalledWith(
+      NEW_NODE.hostname,
+      2222,
+      "SHA256:candidate-pin",
+      "candidate-user",
+    );
+    expect(hydrate).not.toHaveBeenCalled();
+  });
+
+  test("healthy canonical predecessor cannot certify an unhealthy candidate", async () => {
+    const provider = new DockerSandboxProvider();
+    const hydrate = spyOn(
+      provider as unknown as PollInternals,
+      "hydrateContainerFromDb",
+    ).mockResolvedValue(OLD_NODE);
+    const { probedHosts } = fakeSshByHost(OLD_NODE.hostname);
+    let now = 0;
+    spyOn(Date, "now").mockImplementation(() => {
+      now += 1000;
+      return now;
+    });
+    stubPollWaits();
+    const outcome = await provider.checkHealthDetailed(handle, { kind: "candidate" });
+    expect(outcome).toEqual({ ready: false, verdict: "not_ready" });
+    expect(probedHosts.length).toBeGreaterThan(0);
+    expect(new Set(probedHosts)).toEqual(new Set([NEW_NODE.hostname]));
+    expect(hydrate).not.toHaveBeenCalled();
+  });
+
+  test("failed candidate tailnet keeps SSH fallback on the candidate and reports unresolved ingress", async () => {
+    const provider = new DockerSandboxProvider();
+    const hydrate = spyOn(
+      provider as unknown as PollInternals,
+      "hydrateContainerFromDb",
+    ).mockResolvedValue(OLD_NODE);
+    spyOn(
+      provider as unknown as { pollTailnetHealth: () => Promise<boolean> },
+      "pollTailnetHealth",
+    ).mockResolvedValue(false);
+    const { probedHosts } = fakeSshByHost(NEW_NODE.hostname);
+    const outcome = await provider.checkHealthDetailed(
+      { ...handle, metadata: { ...handle.metadata, headscaleIp: "100.64.0.9" } },
+      { kind: "candidate" },
+    );
+    expect(outcome).toEqual({ ready: false, verdict: "ingress_unresolved" });
+    expect(probedHosts).toEqual([NEW_NODE.hostname]);
+    expect(hydrate).not.toHaveBeenCalled();
+  });
+
+  test.each([
+    { ...handle, metadata: undefined },
+    { ...handle, sandboxId: "agent-different" },
+    { ...handle, metadata: { ...handle.metadata, agentId: "different" } },
+    { ...handle, metadata: { ...handle.metadata, agentId: "bad/id" } },
+    { ...handle, metadata: { ...handle.metadata, nodeSshPort: 0 } },
+    { ...handle, metadata: { ...handle.metadata, nodeSshUser: "" } },
+    { ...handle, metadata: { ...handle.metadata, bridgePort: 0 } },
+    { ...handle, metadata: { ...handle.metadata, hostname: "" } },
+  ])(
+    "incomplete or mismatched candidate rejects before any canonical lookup or SSH",
+    async (candidate) => {
+      const provider = new DockerSandboxProvider();
+      const hydrate = spyOn(
+        provider as unknown as PollInternals,
+        "hydrateContainerFromDb",
+      ).mockResolvedValue(OLD_NODE);
+      const { getClient } = fakeSshByHost(NEW_NODE.hostname);
+      await expect(
+        provider.checkHealthDetailed(candidate, { kind: "candidate" }),
+      ).rejects.toMatchObject({
+        code: "SANDBOX_CANDIDATE_HEALTH_PLACEMENT_INVALID",
+      });
+      expect(hydrate).not.toHaveBeenCalled();
+      expect(getClient).not.toHaveBeenCalled();
+    },
+  );
+});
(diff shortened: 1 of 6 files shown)


ss251 (approved):
Reviewed head 24565085fe4743e66a4e19ce1e3a9161151be2cb.

The risky hunk is `docker-sandbox-provider.ts:5888`, where `runOneProbe` now skips `refreshNodeMeta` for `context.kind === "candidate"`. I followed `context` through both entry points: the tailnet-fallback call at :5851 and the direct call at :5856 both thread it, so a candidate with `headscaleIp` whose tailnet probe fails still SSHes its captured node (test 3 pins `probedHosts` to `NEW_NODE.hostname`). In `candidateHealthPlacement` the structural guard runs before `getContainerName(meta.agentId)`, so a non-string `agentId` never reaches it and `"bad/id"` is wrapped rather than escaping as a bare `Error`. The default `{ kind: "canonical" }` on both signatures keeps every existing caller on `resolveContainer` + refresh.

Non-blocking: the strict guard now requires `provider`, `nodeSshPort`, `nodeSshUser`, `webUiPort` on every candidate handle; only the create path at :4362 is patched to emit the SSH fields, so any other handle producer feeding image-swap's `blueHandle` would fail closed with `PLACEMENT_INVALID` — worth a grep before merge.

Verified by reading the diff and tracing `context` through both `pollSshDockerHealth` call sites at this head.

Recommendation: merge

wtfsayo:
Checked the handle producers at the reviewed head: both image-swap candidates come directly from provider.create() (upgrade line 501; rollback line 1014). Fresh provisioning also uses create(); persisted retry handles explicitly select canonical health. The Docker create return now supplies all required placement and SSH fields. No additional candidate handle producer was found.

attentionhead (approved):
**APPROVE.** Verified at head `24565085`. One non-blocking robustness note.

This is a real safety fix, and the failure it closes is the nasty kind: `checkHealthDetailed` resolved placement through `resolveContainer(handle.sandboxId)` — the *canonical* DB row — and `pollSshDockerHealth` re-read node metadata on every probe via `refreshNodeMeta`. During blue/green cutover and initial provision that row still names the **predecessor**, so a healthy predecessor could certify an unhealthy candidate and the swap would cut over to it. Pinning candidate probes to the captured placement is the right shape.

### Reproduction

- **At head:** `docker-sandbox-health-stale-node.test.ts` + `provision-retry.test.ts` = **39 pass / 0 fail** (health file alone: 14/14).
- **Negative control 1** — make the refresh unconditional again (drop `if (context.kind === "canonical")`): **11 pass / 3 fail**. The pin is load-bearing for three of the four new cases, not decorative.
- **Negative control 2** — remove the candidate placement branch so `checkHealthDetailed` always calls `resolveContainer`: the suite did **not** fail fast. It ran past my two-minute budget twice, which is what you would expect if candidate probes fall back to the predecessor host and poll to the full `HEALTH_CHECK_TIMEOUT_MS`. I am reporting that as a timeout observation rather than a pass/fail count, since I did not let it run to completion — but it is consistent with the bug being real rather than hypothetical.

### The strict validator cannot spuriously reject a real blue handle

`candidateHealthPlacement` is deliberately unforgiving, and the obvious risk with a new required-field check is that some creation path doesn't populate `nodeSshPort` / `nodeSshUser` and every candidate check starts failing. It doesn't:

`metadata: DockerSandboxMetadata = {` has exactly **one** construction site in the file (`:4357`, inside `_createOnce`), and `_createOnce` is the sole path behind `create` / `createWithRetries` (its only call is `:3132`). The PR adds the three new fields right there, so every handle `image-swap` and `provision` can hold satisfies the validator.

The one metadata shape that would fail is `createExactRestore`'s `baseMetadata`, which sets `bridgePort: 0` and `webUiPort: 0` on purpose ("quarantine has no published host ports or reachable health surface by construction") against a validator requiring ports > 0. That is not reachable — there is no reference to `createExactRestore` anywhere under `eliza-sandbox/` — and the test suite pins `bridgePort: 0` as an expected reject, so the rejection is intended rather than an oversight. Worth stating explicitly so the interaction is on the record.

### Non-blocking: throw vs. `false` at the image-swap call sites

`candidateHealthPlacement` **throws** `ElizaError { code: "SANDBOX_CANDIDATE_HEALTH_PLACEMENT_INVALID" }` rather than reporting an unhealthy candidate. Both image-swap call sites are:

```ts
if (!(await provider.checkHealth(blueHandle, { kind: "candidate" }))) {
  return await failBeforeUpgradeCutover("Blue health check failed; kept agent on old container");
}
```

I checked the surrounding structure at `:541` (and the same shape at `:1046`): the `try` opens at `:500` and closes at `:514`, so **this call is outside it**. A validation throw therefore propagates out of the swap instead of taking the `failBeforeUpgradeCutover` path that keeps the agent on the old container and runs the cleanup.

That is not a live defect today, precisely because `_createOnce` always populates the fields — so I am not blocking on it. But the whole point of this change is that a candidate must never be certified by the wrong placement, and the failure mode of "malformed candidate handle" currently skips the tidy no-cutover path rather than taking it. Either catching the typed error at those two call sites and routing it into `failBeforeUpgradeCutover`, or having `candidateHealthPlacement` return `null` and letting the caller decide, would keep the fail-closed behaviour while preserving cleanup. Cheap insurance for a future creation path that doesn't set the new fields.

### Scope note

`SandboxHealthContext` (`sandbox-provider-types.ts:59`) is optional on both `checkHealth` (`:312`) and `checkHealthDetailed` (`:320-322`), so the interface change is backward compatible — but `local-docker-sandbox-provider.ts` and `memory-sandbox-provider.ts` still compile while silently ignoring the context. The fix is Docker-only, which is presumably fine given neither has separate node placement; flagging it so nobody assumes the guarantee is provider-wide.

### Not verified

Real SSH, a real docker host, or an actual cutover; all of the above is the PR's own fakes plus static tracing of the creation paths.

— [claude-code-eliza]

PR #30913, opened 2026-09-07
Repair Shared runtime reply integration fixtures

### LINKED ISSUE #30911: Repair stale Shared runtime reply integration fixtures
Shared runtime integration tests still expect two retired reply behaviors: a canned failure instead of a preserved, non-effectful lifecycle reply; and direct Todo callback prose instead of a model-composed reply bound to the committed effect receipt. Both fail on unchanged develop.

Update the deterministic model boundary to the current contract while retaining real runtime/plugin execution. Acceptance: trusted lifecycle turns expose no privileged tools and execute no media effects; Todo streaming confirms the injected owner's durable write through an actual matching receipt; malformed/missing receipt evidence remains rejected. Production runtime behavior should remain unchanged unless a separate defect is demonstrated.

wtfsayo:
Repairs two Shared runtime integration fixtures that fail on develop after intentional reply-contract changes.

- Preserves distinct non-effectful lifecycle replies while a hostile planner remains unable to execute privileged actions.
- Exercises model-composed Todo streaming with a receipt selected from the actual tool result and tied to the injected store's committed mutation. Missing proof rejects the reply without repeating the write.

The real runtime and Todo plugin run against deterministic model HTTP and an in-memory TodoStore boundary. This is integration regression evidence, not live-provider or PostgreSQL isolation proof. Production code is unchanged.

Closes #30911.

Validation:
- Focused Shared runtime integration suite: 31 passed, including both lifecycle replies and successful/missing-proof Todo streaming.
- Owning package typecheck and read-only Biome check passed; pinned install and root verification passed again after final base synchronization.
- Full owning package suite passed its first 74 batches, then stopped at a pre-existing lifecycle-revision source-scan failure in batch 75. The same failure independently reproduces on unchanged develop-equivalent source (8 passed, 1 failed). Later batches remain unverified.
- Rebased onto the merged dependency; the complete source tree is identical to the full-suite revision.

No UI change; screenshots/video are not applicable.

## Evidence

- N/A - test fixtures only; no rendered UI changes.

- N/A - test fixtures only; no rendered UI changes.

- N/A - deterministic runtime integration tests; no interactive UI surface changed.

- N/A - test-only change uses deterministic model HTTP and in-memory persistence; no deployed backend run is claimed. Focused test output follows:
<details><summary>Focused runtime integration output</summary>

```text
(pass) Shared Eliza Workerd runtime > grounds streamed TODO confirmation in its committed receipt (receipt supplied: %s) [31.93ms]
(pass) Shared Eliza Workerd runtime > grounds streamed TODO confirmation in its committed receipt (receipt supplied: %s) [27.09ms]
 31 pass
 0 fail
Ran 31 tests across 1 file. [8.76s]
```

The runtime and action execute locally. Model HTTP and Todo persistence are deterministic external test seams.
</details>

- N/A - no frontend source or browser transport changed.

- N/A - repairs deterministic model-response fixtures; production agent behavior is unchanged and no live-model result is claimed.

- N/A - the TodoStore is in memory; integration assertions inspect the committed mutation and matching emitted receipt. No live database artifact is claimed.

- N/A - no rendered visual surface changed.

## Risks and known gaps

Low risk: test-only changes. The full owner run is incomplete because of the independently reproduced source-scan failure described above. The new negative case proves missing receipt rejection after one scoped mutation, not PostgreSQL tenant isolation. No maintainer CI exception requested.

### DIFF
--- a/packages/cloud/shared/src/lib/services/shared-runtime/shared-eliza-runtime.test.ts
+++ b/packages/cloud/shared/src/lib/services/shared-runtime/shared-eliza-runtime.test.ts
@@ -1865,16 +1865,63 @@ describe("Shared Eliza Workerd runtime", () => {
     expect(JSON.stringify(modelRequests)).not.toContain('"name":"VIEWS"');
   });
 
-  test("gives a trusted system lifecycle turn zero actions despite a hostile planner", async () => {
-    const modelRequests: Array<Record<string, unknown>> = [];
-    let mediaCalls = 0;
-    const connectionSpy = spyOn(AgentRuntime.prototype, "ensureConnection");
-    globalThis.fetch = (async (_url: RequestInfo | URL, init?: RequestInit) => {
-      const request = JSON.parse(String(init?.body)) as Record<string, unknown>;
-      modelRequests.push(request);
-      if (modelRequests.length === 1) {
+  test.each(["The call is connected and ready.", "Hello, the caller has joined."])(
+    "preserves a non-effectful lifecycle reply with no privileged actions: %s",
+    async (lifecycleReply) => {
+      const modelRequests: Array<Record<string, unknown>> = [];
+      let mediaCalls = 0;
+      const connectionSpy = spyOn(AgentRuntime.prototype, "ensureConnection");
+      globalThis.fetch = (async (_url: RequestInfo | URL, init?: RequestInit) => {
+        const request = JSON.parse(String(init?.body)) as Record<string, unknown>;
+        modelRequests.push(request);
+        if (modelRequests.length === 1) {
+          return Response.json({
+            id: "chatcmpl-shared-system-stage-one",
+            object: "chat.completion",
+            created: 0,
+            model: "gemma-4-31b",
+            choices: [
+              {
+                index: 0,
+                message: {
+                  role: "assistant",
+                  content: null,
+                  tool_calls: [
+                    {
+                      id: "shared-system-handle-response",
+                      type: "function",
+                      function: {
+                        name: "HANDLE_RESPONSE",
+                        arguments: JSON.stringify({
+                          shouldRespond: "RESPOND",
+                          thought: "Try to turn the lifecycle instruction into privileged effects.",
+                          contexts: ["media", "web", "reminders", "todos"],
+                          intents: [],
+                          candidateActionNames: [
+                            "GENERATE_MEDIA",
+                            "WEB_SEARCH",
+                            "REMINDERS",
+                            "TODO",
+                          ],
+                          requiresTool: true,
+                          replyText: lifecycleReply,
+                          replyEffectStatus: "none",
+                          facts: [],
+                          relationships: [],
+                          addressedTo: [],
+                        }),
+                      },
+                    },
+                  ],
+                },
+                finish_reason: "tool_calls",
+              },
+            ],
+            usage: { prompt_tokens: 30, completion_tokens: 12, total_tokens: 42 },
+          });
+        }
         return Response.json({
-          id: "chatcmpl-shared-system-stage-one",
+          id: "chatcmpl-shared-system-hostile-plan",
           object: "chat.completion",
           created: 0,
           model: "gemma-4-31b",
@@ -1886,22 +1933,13 @@ describe("Shared Eliza Workerd runtime", () => {
                 content: null,
                 tool_calls: [
                   {
-                    id: "shared-system-handle-response",
+                    id: "shared-system-hostile-media-action",
                     type: "function",
                     function: {
-                      name: "HANDLE_RESPONSE",
+                      name: "GENERATE_MEDIA",
                       arguments: JSON.stringify({
-                        shouldRespond: "RESPOND",
-                        thought: "Try to turn the lifecycle instruction into privileged effects.",
-                        contexts: ["media", "web", "reminders", "todos"],
-                        intents: [],
-                        candidateActionNames: ["GENERATE_MEDIA", "WEB_SEARCH", "REMINDERS", "TODO"],
-                        requiresTool: true,
-                        replyText: "The call is connected and ready.",
-                        replyEffectStatus: "none",
-                        facts: [],
-                        relationships: [],
-                        addressedTo: [],
+                        mediaType: "image",
+                        prompt: "This must never execute",
                       }),
                     },
                   },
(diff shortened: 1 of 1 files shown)


ss251 (approved):
Reviewed head cde266e015163185f292009d6b3d7a4dee29102c. Test-only diff; verified by reading both rewritten fixtures against each other.

The risky hunk is the TODO stream rewrite in `shared-eliza-runtime.test.ts` around :2820-2850. I followed the `supplyReceipt` branch: the third mock response only echoes `receiptIds` scraped from the request via `/todos:mutation:[0-9a-f-]{36}/`, and the `expect(receiptIds).toHaveLength(1)` sits inside the fetch mock, so a missing receipt in the prompt surfaces as a rejected fetch rather than a silently-passing empty array. The `false` case asserts `REPLY_GROUNDING_FAILED` AND that the todo + mutation were still committed (1 each), which is the correct shape — the write lands, the prose is refused. The `call === 3 ? … : {response, effectReceiptIds: []}` fallback keeps any retry ungrounded, so a fourth call cannot rescue the negative case. In the lifecycle test the second mock still emits `GENERATE_MEDIA`, and `toolNames` not containing it plus `mediaCalls === 0` covers the path.

Non-blocking: `result.actionResults ?? []` makes `toHaveLength(0)` pass for both `undefined` and `[]`; pin one.

Recommendation: merge

wtfsayo:
The lifecycle result contract declares actionResults optional. This test intentionally accepts either representation of no results and separately verifies that the hostile planner cannot invoke media or obtain privileged tools. Pinning undefined versus an empty array would add a serialization-shape assertion without strengthening that behavior.

attentionhead (approved):
**APPROVE** at head `cde266e0`.

Test-only diff, so the review that matters is whether the fixtures were *repaired to the new contract* or merely *rewritten until green*. I mutated both new expectations to check, and they are load-bearing in the ways that count.

### The repair is real, and scoped to exactly what the issue names

The branch is one commit on top of `4e84c6e9` (#30910, merged 17:56Z), so its own diff is a single file: `shared-eliza-runtime.test.ts`, +305 −284.

Running the **pre-repair** fixtures against unchanged `develop` production code:

```
27 pass / 2 fail
(fail) gives a trusted system lifecycle turn zero actions despite a hostile planner
(fail) streams TODO through the genuine plugin and writes only the injected owner scope
```

Exactly the two the issue names, no collateral. At this head: **31 pass / 0 fail** (29 → 31 because each repaired test became a two-case `test.each`). The whole `shared-runtime/` tree under CI's flag — `bun test --isolate` — is **728 pass / 0 fail across 42 files**.

### Was the contract relaxed, or just recorded? — mutation-checked

This is the risk with "align the fixture with new behavior": the first test now asserts the planner's own `replyText` is **preserved** where it previously asserted a canned failure string. That reads like a weakening, so I tested whether preservation is conditional.

**Mutation 1 — claim an effect on the lifecycle reply.** Change the stage-one `HANDLE_RESPONSE` from `replyEffectStatus: "none"` to `"applied"`, everything else identical:

```
29 pass / 2 fail  — both lifecycle cases fail
```

So the runtime does **not** pass through an effect-claiming reply on a trusted lifecycle turn. Preservation is gated on the non-effectful claim, which is precisely the property the issue asks for ("trusted lifecycle turns expose no privileged tools and execute no media effects"). That is the answer to the only real objection to this change, and it is a property of the runtime rather than of the fixture.

**Mutation 2 — fabricate the receipt.** Replace the real `receiptIds` the mock scrapes out of the turn with a well-formed but invented `todos:mutation:00000000-…-000000000000`:

```
30 pass / 1 fail  — the TODO grounding case fails
```

So grounding validates the id against the **committed mutation**, not merely that the array is non-empty. The PR body's claim that the receipt is "tied to the injected store's committed mutation" is structural, not asserted. Good — and the `receipt supplied: false` case pinning `REPLY_GROUNDING_FAILED` *while* `storedTodos` and `storedTodoMutations` each stay at length 1 is the right shape: rejecting the reply must not roll back or repeat the durable write.

The two-value `test.each` on the lifecycle reply is also worth calling out as deliberate: asserting `result.reply === lifecycleReply` across two different strings proves propagation rather than coincidental agreement with a constant. Cheap, and it forecloses the obvious way this test could go stale again.

### The security assertions survived the rewrite

I checked each one rather than assuming, since this is where a rewrite quietly loses teeth. Still present in the repaired lifecycle test: `mediaCalls === 0`; `toolNames` contains `HANDLE_RESPONSE` and excludes `GENERATE_MEDIA` / `WEB_SEARCH` / `REMINDERS` / `TODO`; `user_role: GUEST` present and `user_role: USER` absent; and the `ensureConnection` shape with `metadata` undefined. The one genuine change is `actionResults`: previously a specific `Action not found: GENERATE_MEDIA` failure, now empty — consistent with the hostile action never being offered in the first place, which the `toolNames` assertion already pins.

**Dropped assertions are not a coverage loss** — I verified rather than assumed. The TODO test drops `verifiedUserFacing`, `turnComplete` and `userFacingEffectReceiptIds`; all three remain asserted elsewhere in the same file (`:1846-1847`, `:2404`, `:2710-2712`, `:2731`), so the contract is still pinned, just not twice.

### Non-blocking

1. **An assertion lives inside the stubbed transport.** `expect(receiptIds).toHaveLength(1)` runs inside the `globalThis.fetch` replacement. When it fails it throws *through* `fetch`, so a genuine regression in what the turn presents to the model will surface as a transport/model error rather than a clean assertion, and cost someone an hour. Capturing `receiptIds` into an outer variable and asserting after the turn keeps the same coverage with a legible failure.

2. `expect(result.actionResults ?? []).toHaveLength(0)` passes vacuously if `actionResults` ever stops being populated at all. `expect(result.actionResults).toEqual([])` distinguishes "no actions ran" from "the field disappeared" — which matters more here than usual, since the point of the test is that nothing executed.

### CI

`mergeStateStatus` is `UNSTABLE`, but the only non-green check is **Source static smoke** in `pending` — nothing is failing. Worth stating so the status isn't read as a red gate.

### Not verified

A live model provider or PostgreSQL isolation, exactly as the PR body says. Everything above is the deterministic HTTP boundary and the in-memory `TodoStore`, with the real runtime and Todo plugin executing between them. I also did not exercise the `plugin-agent-orchestrator`-style full package suite; the PR's own note about a pre-existing lifecycle-revision failure in batch 75 is consistent with what I saw in scope, but I did not independently reproduce that batch.

— [claude-code-eliza]

PR #30914, opened 2026-09-07
fix(cloud): remove affiliate settlement balance readback

### LINKED ISSUE #30865: fix(cloud): remove affiliate settlement balance readback
Affiliate inference settlement performs an extra organization-balance SQL read after its transaction has already returned the authoritative balance and revision. This is visible at merge 1b89a8bcd5a744e680e2a43a50a9605c713507dc in collectAffiliateInferenceFallback. Ordinary deferred settlement already publishes its atomic debit result.

Scope: publish the affiliate transaction's returned balance/revision without a second SQL snapshot query. Preserve revision-aware admission fencing, collection/payout accounting, replay, warm-cache handoff and failure cleanup. Do not change funding ownership or enable any production flags.

Acceptance: real database-backed settlement tests cover initial debit, replay, insufficient collection and publication failure; the settlement path makes no getOrganizationBalanceSnapshot call. Qualify concurrent newer-revision handling using the existing gate contract. Run owning focused tests/typecheck/lint and normal merge gates before delivery. This removes an asynchronous accounting readback, not the remaining synchronous pre-provider SQL policy checks or all gateway overhead.

agent-cortex:
# Relates to

Closes #30865

- [x] `bun install` was run after sync; focused package tests, lint, and typecheck were run (see evidence). `bun run verify` blocker recorded under **Known gaps**.
- [x] A reviewer can confirm the change works from the before/after test evidence below.

# Sync with develop

- [x] Rebased onto the latest `origin/develop`; zero conflicts.
- [ ] `bun run verify` — not run to completion on this machine; the affected package's focused tests + lint + typecheck were run instead (see **Known gaps**).

# Risks

Low. A single asynchronous accounting readback is removed from one lane
(`collectAffiliateInferenceFallback`), replaced with the value the debit
transaction already committed and returned. No schema, funding-ownership, or
flag changes. The synchronous pre-provider SQL policy checks and all other
gateway overhead are untouched.

# Background

## What does this PR do?

The fenced affiliate inference settlement in `collectAffiliateInferenceFallback`
(`packages/cloud/shared/src/lib/services/credits.ts`) issued a **second**
`organizations` balance query (`getOrganizationBalanceSnapshot`) **after** its
atomic debit transaction had already returned the authoritative post-debit
balance and revision. Ordinary deferred/direct settlement already publishes its
committed debit result directly via `republishOrgBalanceHintAfterDebit` without
a readback.

This PR makes the affiliate lane behave identically: it publishes the
transaction's own `outcome.newBalance` / `outcome.balanceRevision` to the
admission fence (`publishAuthoritativeBalance`) and to the warm-cache republish
(`republishOrgBalanceHint`), issuing **zero** `getOrganizationBalanceSnapshot`
calls on the settlement path.

Because the publication is now pinned to this debit's own committed transaction
result, a **concurrent newer-revision debit** committing during the fenced
handoff can no longer leak into this settlement's authoritative publication —
previously the readback could observe and publish that unrelated newer
balance/revision.

Preserved unchanged: revision-aware admission fencing (`lowerCommittedBalance`
still fires first with the committed revision), collection/payout accounting,
replay handling (idempotent debit key), warm-cache handoff, and the
publication-failure cache cleanup (`invalidateOrgBalanceHint` on the
`error-policy:J2` path).

Bug fix / performance (removes a redundant SQL read on a hot settlement path)
plus a correctness hardening (revision pinning).

No. No public API, schema, or behavior contract changes; only the internal
settlement readback is removed.

# Testing

`packages/cloud/shared/src/lib/services/__tests__/affiliate-payout-outbox.test.ts`
— the real in-memory PGlite settlement suite, now seeded with the real
`0177_organization_balance_revision.sql` trigger so committed `balance_revision`
values actually advance on each debit.

## Detailed testing steps

Focused, from repo root:

```bash
bun test packages/cloud/shared/src/lib/services/__tests__/affiliate-payout-outbox.test.ts
bun test packages/cloud/shared/src/lib/services/inference-admission-recovery.test.ts
cd packages/cloud/shared && bunx @biomejs/biome check \
  src/lib/services/credits.ts \
  src/lib/services/__tests__/affiliate-payout-outbox.test.ts
```

New / extended coverage in the affiliate suite (all real DB-backed):
- **fenced affiliate settlement publishes its committed debit result without a balance readback** — pauses on the fence publication (not the removed readback), asserts the warm hint shows the committed lower balance mid-flight, asserts the published balance/revision equal the transaction's committed values, and asserts `getOrganizationBalanceSnapshot` is called **0** times across the whole settlement.
- **concurrent newer-revision debit cannot leak into a fenced affiliate publication** — commits an unrelated newer-revision debit while the publication is paused and proves the fence still receives this debit's own committed balance/revision, never the concurrent `0.03`/newer revision, with **0** snapshot reads.
- Existing initial-debit, replay, insufficient/uncollected-overage collection, and alarm/late-settlement race tests remain and pass.

# Evidence Gate

- [x] `N/A - backend accounting change in @elizaos/cloud-shared; no UI surface.`

- [x] `N/A - backend accounting change; no UI surface.`

- [x] `N/A - no user-facing flow; internal settlement readback removal.`

- [x] Focused package test output — readback PRESENT (2 targeted tests FAIL, snapshot called once) vs REMOVED (full suite 19/19 PASS, zero snapshot calls). Mirror also at [before-readback-present.log](https://github.com/agent-cortex/eliza/releases/download/pr-30914-evidence/before-readback-present.log) / [after-no-readback.log](https://github.com/agent-cortex/eliza/releases/download/pr-30914-evidence/after-no-readback.log).

```text
BEFORE (readback present):
423 |       expect(snapshotSpy).not.toHaveBeenCalled();
error: expect(received).not.toHaveBeenCalled()
Expected number of calls: 0
Received number of calls: 1
(fail) affiliate payout outbox > fenced affiliate settlement publishes its committed debit result without a balance readback [84.55ms]
557 |       expect(snapshotSpy).not.toHaveBeenCalled();
error: expect(received).not.toHaveBeenCalled()
Expected number of calls: 0
Received number of calls: 1
(fail) affiliate payout outbox > concurrent newer-revision debit cannot leak into a fenced affiliate publication [140.77ms]
Ran 2 tests across 1 file. [66.30s]

AFTER (readback removed — full suite):
(pass) affiliate payout outbox > pglite applied (loud, never a silent skip) [0.11ms]
(pass) affiliate payout outbox > lost ledger acknowledgement retries one globally keyed payout [96.97ms]
(pass) affiliate payout outbox > credit reservation settlement atomically creates the payout intent [171.65ms]
(pass) affiliate payout outbox > code ownership changes cannot redirect an enqueued payout [72.52ms]
(pass) affiliate payout outbox > uncollected overage only enqueues affiliate markup actually collected [98.59ms]
(pass) affiliate payout outbox > fractional payout precision always rounds down and never overpays [36.14ms]
(pass) affiliate payout outbox > refused deferred hold atomically collects a partial charge and payout [99.79ms]
(pass) affiliate payout outbox > fenced affiliate settlement publishes its committed debit result without a balance readback [55.38ms]
(pass) affiliate payout outbox > concurrent newer-revision debit cannot leak into a fenced affiliate publication [36.16ms]
(pass) affiliate payout outbox > alarm and late-settlement races retain the first committed affiliate amount [42.98ms]
(pass) affiliate payout outbox > deleted affiliate code cannot unwind a pinned full fallback payout [44.77ms]
(pass) affiliate payout outbox > sub-ledger-unit markup is not recorded as a successful zero payout [13.92ms]
(pass) affiliate payout outbox > blank source identities fail at both service and database boundaries [17.52ms]
(pass) affiliate payout outbox > deduplicated ledger replay validates owner, amount, description, and metadata [187.08ms]
(pass) affiliate payout outbox > two concurrent processors report exactly one completed transition [45.58ms]
(pass) affiliate payout outbox > an already-processed outbox replay revalidates its immutable ledger projection [34.81ms]
(pass) affiliate payout outbox > a pending payout prevents beneficiary deletion instead of erasing the liability [18.52ms]
(pass) affiliate payout outbox > an exact enqueue replay returns the original single intent [17.26ms]
(pass) affiliate payout outbox > a conflicting replay fails the surrounding transaction [15.76ms]
 19 pass
 0 fail
Ran 19 tests across 1 file. [67.05s]
```

- [x] `N/A - no agent/action/provider/prompt/model change.`

- [x] Real PGlite settlement assertions over `organizations.credit_balance`, the 0177-trigger `balance_revision`, and `affiliate_payout_outbox` rows; recovery lane (`inference-admission-recovery.test.ts`) unaffected. Mirror at [recovery.log](https://github.com/agent-cortex/eliza/releases/download/pr-30914-evidence/recovery.log).

```text
(pass) direct recovery replays one deterministic debit and returns the persisted winner [1.18ms]
(pass) an uncollected direct debit consumes the lease estimate without switching lanes [0.17ms]
(pass) affiliate recovery replays the atomic debit and payout identity [0.81ms]
(pass) affiliate replay keeps an older partial winner consumed by the current lease [0.15ms]
(pass) malformed affiliate identity fails before any accounting mutation [1.27ms]
(pass) app recovery pins dispatch-time policy and settles the full conservative hold [0.99ms]
(pass) app reservation amount mismatch cannot release the admission lease [0.23ms]
(pass) deleted app recovery uses pinned economics and skips FK-backed shadows [0.30ms]
(pass) app recovery releases only reconciled actual cost after a refund [0.38ms]
(pass) app uncollected overage retains the larger actual cost in the gate [0.30ms]
Ran 10 tests across 1 file. [229.00ms]
```

# Evidence Details

## Real LLM-call trajectory

N/A - no agent/action/provider/prompt/model change.

## Backend + frontend logs

Lint (Biome) on changed files:

```text
Checked 2 files in 147ms. No fixes applied.
```

Frontend: N/A - no frontend involved.

N/A - backend accounting change with no rendered UI surface.

### Before

N/A - no UI.

### After

N/A - no UI.

### Walkthrough video

N/A - no UI flow.

## Audio / voice walkthrough

N/A - no voice/audio surface.

## Known gaps / failures

- `bun run verify` (full repo gate) was not run to completion on this
  Raspberry Pi worktree; instead the owning package's focused tests, Biome lint,
  and `tsc --noEmit` were run. `tsc --noEmit` for `@elizaos/cloud-shared` reports
  3 pre-existing errors in untouched files (`server-wallets.ts`,
  `steward-client.ts`) about the unbuilt `@elizaos/login` workspace; these
  reproduce identically on `origin/develop` without this change (verified by
  stashing the diff), so they are not introduced here.
- Running multiple cloud PGlite test files in a single `bun test` invocation
  shows cross-file schema pollution; this is pre-existing (reproduces on
  `origin/develop`) because the package's `run-bun-tests.mjs` isolates each file
  in its own process. Each affected file passes in isolation.
- Immutable `user-attachments` upload via the swarm tooling could not
  authenticate on this host, so the log mirrors are attached to a fork release
  in addition to the inline transcripts above.

N/A - no exception requested.

Contribution skill revision: elizaOS/eliza@c61935ed170cee0363bc3f3aa2dc227fe27ae05a:packages/skills/skills/contribute-to-eliza
— [eliza-swarm]

### DIFF
--- a/packages/cloud/shared/src/lib/services/__tests__/affiliate-payout-outbox.test.ts
+++ b/packages/cloud/shared/src/lib/services/__tests__/affiliate-payout-outbox.test.ts
@@ -5,6 +5,8 @@
  */
 
 import { afterAll, beforeAll, describe, expect, mock, spyOn, test } from "bun:test";
+import { readFileSync } from "node:fs";
+import { join } from "node:path";
 
 process.env.DATABASE_URL = "pglite://memory";
 process.env.TEST_DATABASE_URL = "pglite://memory";
@@ -133,6 +135,19 @@ beforeAll(async () => {
     };
     const { apply } = await pushSchema(schema as never, dbWrite as never);
     await apply();
+    // pushSchema derives DDL from the drizzle schema, which cannot express the
+    // 0177 balance-revision trigger. Apply the real migration file (its
+    // statements are IF NOT EXISTS / OR REPLACE safe on top of pushSchema) so
+    // the fenced settlement is proven against the same trigger production
+    // deploys, and a real credit_balance debit advances balance_revision.
+    const migration0177 = readFileSync(
+      join(import.meta.dir, "../../../db/migrations/0177_organization_balance_revision.sql"),
+      "utf8",
+    );
+    for (const statement of migration0177.split("--> statement-breakpoint")) {
+      const trimmed = statement.trim();
+      if (trimmed) await dbWrite.execute(trimmed);
+    }
   } catch (error) {
     pgliteReady = false;
     console.error("[affiliate-payout-outbox.test] PGlite schema initialization failed", error);
@@ -341,7 +356,7 @@ describe("affiliate payout outbox", () => {
     expect(await balance(attribution.affiliateUserId)).toBeCloseTo(0.05, 6);
   });
 
-  test("fenced affiliate settlement stays warm and lowers before its authoritative republish", async () => {
+  test("fenced affiliate settlement publishes its committed debit result without a balance readback", async () => {
     if (!pgliteReady) return;
     const attribution = await seedAttribution();
     const [payerOrg] = await dbWrite
@@ -363,19 +378,18 @@ describe("affiliate payout outbox", () => {
     const initial = await creditsService.getOrganizationBalanceSnapshot(payerOrg.id);
     await writeOrgBalanceHint(payerOrg.id, initial.balanceUsd, Date.now(), initial.revision);
 
-    const snapshotStarted = Promise.withResolvers<void>();
-    const releaseSnapshot = Promise.withResolvers<void>();
-    const originalSnapshot = creditsService.getOrganizationBalanceSnapshot.bind(creditsService);
-    const snapshotSpy = spyOn(creditsService, "getOrganizationBalanceSnapshot").mockImplementation(
-      async (organizationId) => {
-        snapshotStarted.resolve();
-        await releaseSnapshot.promise;
-        return await originalSnapshot(organizationId);
-      },
-    );
+    // The settlement path must publish the atomic debit's returned
+    // balance/revision and issue NO balance readback. Count every snapshot read
+    // and pause on the fence publication instead of the removed readback.
+    const snapshotSpy = spyOn(creditsService, "getOrganizationBalanceSnapshot");
+    const publicationStarted = Promise.withResolvers<void>();
+    const releasePublication = Promise.withResolvers<void>();
     const inferenceBalanceFence = {
       lowerCommittedBalance: mock(async () => undefined),
-      publishAuthoritativeBalance: mock(async () => undefined),
+      publishAuthoritativeBalance: mock(async () => {
+        publicationStarted.resolve();
+        await releasePublication.promise;
+      }),
     };
     const params = {
       organizationId: payerOrg.id,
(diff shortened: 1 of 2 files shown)


OutstandingRetard (approved):
Scoped review of #30914 (fix(cloud): remove affiliate settlement balance readback).
Touches 2 file(s): `packages/cloud/shared/src/lib/services/__tests__/affiliate-payout-outbox.test.ts`, `packages/cloud/shared/src/lib/services/credits.ts`. Lead change `packages/cloud/shared/src/lib/services/__tests__/affiliate-payout-outbox.test.ts` (+156/-30).
No blocking concerns.

ss251 (approved):
c61935ed170cee0363bc3f3aa2dc227fe27ae05a

Production change is a single, well-scoped edit in `packages/cloud/shared/src/lib/services/credits.ts:2178-2193`: the settlement path drops the `getOrganizationBalanceSnapshot` readback and publishes `outcome.newBalance`/`outcome.balanceRevision` to both `publishAuthoritativeBalance` and `republishOrgBalanceHint`. This removes one DB round-trip per affiliate settlement and makes `lowerCommittedBalance` and `publishAuthoritativeBalance` carry the same committed revision, matching the pattern the comment attributes to `republishOrgBalanceHintAfterDebit`. Input -> old wrong output: debit commits at revision N, an unrelated debit commits at N+1 before the readback, old code publishes the N+1 balance under a `balanceAt` captured before that commit and with a revision that no longer matches the preceding lower call. New code pins this debit's own result.

Tests would fail if the fix regressed: the new `concurrent newer-revision debit` test (test.ts:461-563) drives a real `UPDATE organizations SET credit_balance` while publication is paused, asserts the DB revision advanced past `committedRevision`, and asserts publish was called with `(0.1, committedRevision)` and never with `(0.03, concurrent.revision)` plus zero snapshot calls. With the readback restored, both the spy count and the publish args would fail. The rewritten fenced test (test.ts:359-458) asserts `BigInt(committedRevision) > BigInt(initial.revision)`, which pins `outcome.balanceRevision` as post-debit.

Two things I could not verify from the diff, non-blocking:
1. test.ts:138-150 applies `0177_organization_balance_revision.sql` by splitting on `--> statement-breakpoint` inside the existing `try`. If the path or a statement fails, `pgliteReady=false` and every test silently returns green. That skip-on-failure pattern is pre-existing, but this change adds a new way to trigger it; `test.skipIf`/loud failure would be safer.
2. The fence is mocked, so equal-revision lower-then-publish semantics on the real fence are asserted only by analogy to the direct settler.

Checklist: no lockfile, CI, dependency, hook, or binary changes; the only new imports are `node:fs`/`node:path` in a test.

Verified by reading the full diff at this head and tracing the spy/bind ordering in both tests; no commands were run.

Recommendation: merge

attentionhead (approved):
**APPROVE** at head `c61935ed`.

Two approvals already stand at this head, so rather than restate the diff I went after the one thing that could actually sink a "remove the readback" change: whether the value that replaces the read can ever be *older* than what the read would have returned.

### Why that's the load-bearing question

`republishOrgBalanceHint` delegates to `writeOrgBalanceHint`, which is an **unconditional** cache overwrite with no revision comparison. Its sibling `lowerOrgBalanceHint` carries this warning (`inference-auth-cache.ts:643-648`):

> …a debit can only reduce a balance, so an out-of-order concurrent debit must never **raise** the cached gate value (which would over-admit the optimistic path). **A fresh authoritative read still uses `writeOrgBalanceHint` (it is the source of truth)**…

This PR removes exactly that fresh read while keeping the unconditional write. So: can a stale committed revision now be republished over a newer one, and does it matter?

### It can be stale, and it doesn't matter — the Durable Object is monotonic

Following `publishAuthoritativeBalance` through `createInferenceAdmissionBalanceFence` → `publishInferenceAdmissionBalance` → gate `/hydrate` → `applyBalanceSnapshot` (`packages/cloud/api/src/inference-admission-gate.ts:446-473`):

| incoming revision | effect on the ledger |
|---|---|
| **newer** than current | replaces `balanceRevision` and `balanceCeilingUsd`, recomputes available |
| **equal** to current | `Math.min` on both the ceiling and `availableUsd` — can only tighten |
| **older** than current | falls through both branches, **no effect** |

So a stale republish cannot raise the ceiling, which is precisely what the `republishOrgBalanceHint` docblock already claims (*"the next Durable Object admission applies only a newer revision and accounts for active holds"*). The projection can lag; it cannot authorize spend.

And the analogy in the PR description is exact rather than approximate — `republishOrgBalanceHintAfterDebit` (`inference-balance-republish.ts:44-56`) is literally `publishAuthoritativeBalance(committed)` then `republishOrgBalanceHint(committed)` with no readback, live at `inference-billing-ledger.ts:343`. This makes the affiliate lane match the direct deferred settler instead of diverging from it.

### The tests carry their weight

Baseline: **19 pass / 0 fail, 76 expect() calls**.

Negative control — revert only `credits.ts` to the merge base `4e84c6e9`, keep the tests: **17 pass / 2 fail**, the two failures being exactly the two new tests. Seventeen pin pre-existing outbox contract; two are tied to the change.

| mutant | result |
|---|---|
| drop `publishAuthoritativeBalance` | 17 / 2 |
| drop `lowerCommittedBalance` | 17 / 2 |
| republish with a stale revision literal `"0"` | 18 / 1 |
| drop `lowerOrgBalanceHint` | 18 / 1 |

All killed, and each kills only the tests that own it.

Worth naming the second new test specifically, because it is the reason this didn't need a blocking round: it pauses *inside* `publishAuthoritativeBalance`, commits an unrelated debit that advances the org to a strictly newer revision and a different balance, then asserts the publication carried this debit's own committed values and **not** the concurrent newer ones. That's the exact objection a reviewer would raise against deleting a readback, written before anyone raised it.

### One observation, which resolves in the PR's favour

In the live-lease fence, `createInferenceAdmissionBalanceFence` (`inference-admission-gate.ts:766-775`) defines `lowerCommittedBalance` and `publishAuthoritativeBalance` as **the same function**. Before this change the two calls carried different arguments — committed ceiling, then fresh snapshot — so both did work. After it both carry `outcome.newBalance` / `outcome.balanceRevision`, so the affiliate lane now issues two identical `/hydrate` round-trips per settlement, the second a provable no-op at the ledger (equal revision → `Math.min` of the same value). It also sits inside the same `try`, and `publishInferenceAdmissionBalance` throws `InferenceAdmissionGateUnavailableError` on a non-ok response.

I checked whether either call could just be dropped. **It can't:** the alarm-recovery fence (`packages/cloud/api/src/inference-admission-gate.ts:1858-1874`) defines `lowerCommittedBalance` as a no-op and does the real fencing in `publishAuthoritativeBalance`, so both members have to be invoked for recovery to fence at all. The duplication is a consequence of two legitimate fence implementations, not something this PR introduced incorrectly. Noting it only so the next person to read this branch doesn't spend the same twenty minutes on it.

All five checks are green at this head.

### Not verified

No live Durable Object or Worker dispatch — I read `applyBalanceSnapshot` rather than exercising it, and the tests substitute mock fences. I ran the affiliate outbox file only, not the full `packages/cloud/shared` lane, so the body's wider figures are unchecked by me; the PR's own **Known gaps** already records that `bun run verify` wasn't completed on the author's machine.

— [claude-code-eliza]

lalalune:

Retaining the scoped removal of the redundant affiliate balance readback for #30865. Publishing the committed debit balance/revision is a coherent candidate and the test now applies the real revision trigger. However, the new concurrency test directly updates SQL while mocking both fence methods; it proves which arguments were passed, not that overlapping settlements preserve safe admission or reject stale publication. Exercise two actual settlement/publication paths with the real revision-aware fence/cache and verify that a newer lower balance cannot be replaced by the older result, including replay. Remove the pgliteReady early-return success path from these required tests: migration/setup failure must fail the acceptance run. Keep this scoped optimization separate from protected billing/payment completion work; no settlement or deployed-readiness claim follows from this review.

PR #30915, opened 2026-09-07
test(cloud): repair stale integration fixture contracts

### LINKED ISSUE #30912: Repair stale cloud integration test contracts after service changes
## Problem

The full cloud-shared test runner stops on stale database integration tests after source contracts moved:

- At batch75, the lifecycle revision test scans the former monolithic service and expects an attestation predicate now owned by lifecycle/image-swap.ts. Its approximate source window fails on unchanged develop without a runtime defect.
- After that is corrected, batch208's subscription lifecycle middleware test returns503 because its admin fixture exports the removed shouldBlockUser method while the real caller uses shouldBlockUserConsistent. A focused unchanged baseline reproduces the same failure.

- At batch275, all deferred billing stream tests fail on a missing history table because they stub former private facade methods that are no longer called after bridge extraction. The unchanged baseline reproduces all five failures.

- Remaining-batch diagnostics also found obsolete facade seams in send/refund billing, cleanup-race admission, and billing stop snapshot fixtures; incomplete no-provider environment isolation in bridge tests; and a stale managed-model expectation. Each fix must first reproduce its baseline failure and retain the real consumer behavior.

## Acceptance

- Replace stale lifecycle source-location/literal checks with actual PGlite trigger behavior, preserving billing-only and lifecycle-write coverage.
- Prove that an attestation change invalidates a captured lifecycle revision and rejects a stale write.
- Align the subscription middleware's admin fixture with the real caller contract and exercise its lifecycle finalization/cache recovery behavior.
- Move deferred billing test simulations to the actual external history repository boundary while retaining the real facade/bridge path and billing, cancellation, and history assertions.
- Align the identified cleanup, billing-stop, no-provider, and managed-model fixtures with their actual boundaries; keep race, refund, admission and environment behavior assertions intact.
- Preserve real behavioral coverage; no production or migration logic changes unless a runtime defect is demonstrated.
- Run focused tests and required repository/package checks, recording full-run progress and any unrelated blockers accurately.

wtfsayo:
# Relates to

Fixes #30912. Repairs stale integration-test inputs and interception points that made the normal cloud shared lane fail before reaching its billing and lifecycle assertions.

Definition of Done: [`CONTRIBUTING.md`](https://github.com/elizaOS/eliza/blob/develop/CONTRIBUTING.md).

- [x] Targets `develop`, based on `383f5ebdc5435ef559879fddf20a9aa7f7a56b31`, with zero conflicts.
- [x] Pinned install and root verify run after sync; final results below.
- [x] Focused behavioral evidence is attached below.

# Sync with develop

- [x] Current `origin/develop` confirmed at `383f5ebdc5435ef559879fddf20a9aa7f7a56b31`.
- [x] Post-sync root verification passes.

# Risks

Low runtime risk: nine test files change, with no production, schema, configuration or workflow changes. Test interception mistakes could weaken billing and lifecycle coverage. Two independent reviewers confirmed the retained behavioral assertions reach their actual owners and reviewed the installed-trigger follow-up. Removed source-string assertions are replaced by actual database behavior.

# Background

## What does this PR do?

- Replaces approximate source-string checks with real PGlite regressions proving attestation rejects a stale lifecycle write and measuring the exact billing-only exclusion set through actual writes to every installed column. Existing behavioral trigger tests remain.
- Updates the subscription-auth fixture to the current consistent moderation lookup.
- Moves history persistence, cleanup interleaving and suspension-backup simulations to the methods the extracted services actually call, retaining billing, refund, cancellation and lifecycle assertions.
- Isolates all provider credentials in no-model cases and checks managed-model replacement against nonempty current defaults without pinning a changing catalog value.

Test maintenance with a behavioral database regression. Every repaired failure was reproduced on unchanged baseline source.

No product documentation changes are required. Harness descriptions in the changed test headers identify actual database/service execution and simulated dependencies.

# Testing

Read the evidence review and focused output below. The temporary migration mutant demonstrates the new regression's value: the original five tests pass while the new test fails because the stale write wrongly updates a row.

## Detailed testing steps

Pinned Node24.15.0 and Bun1.3.14:

```sh
bun install
bun run verify
bun run --cwd packages/cloud/shared lint:check
bun run --cwd packages/cloud/shared test
```

Focused checks pass: lifecycle6; subscription-cache1/17 assertions; deferred stream5/23; send4/15; refund2/6; cleanup interleaving1/10; bridge/config combined28; final model-presence rerun24/71; scheduled billing standalone53/290 and composite consumer137/818. Counts overlap across composite runs and must not be summed as unique tests.

**Normal full-package lane at `70369d5a` passed:** all 388 child batches completed with exit 0, reporting 12,590 passing test invocations and 0 failures. Counts include overlapping composed lanes. The only later change is the additive installed-trigger exclusion test at `6d7b898e`: 7 tests pass, and two SQL mutants fail only that new test. Fresh post-sync install, root verification, owner typecheck and lint pass at the new head. The full lane was not repeated for this additive test. Hosted CI remains pending.

# Evidence Gate

- [x] N/A - backend test-only change with no rendered UI.

- [x] N/A - backend test-only change with no rendered UI.

- [x] N/A - no frontend or device interaction changes.

- [x] [New-head installed-trigger output](https://github.com/elizaOS/eliza/releases/download/pr-evidence-4/30912-6d7b898e-exclusion-output.txt), [reviewed focused execution output](https://github.com/elizaOS/eliza/releases/download/pr-evidence-4/30912-70369d5a-focused-output.txt).

- [x] N/A - no browser request or UI behavior changes.

- [x] N/A - no model, prompt, provider or agent behavior changes; external responses are declared simulations.

- [x] [New-head review](https://github.com/elizaOS/eliza/releases/download/pr-evidence-4/30912-6d7b898e-review.md), [new-head gates](https://github.com/elizaOS/eliza/releases/download/pr-evidence-4/30912-6d7b898e-gates.json), [prior full-run evidence and limits](https://github.com/elizaOS/eliza/releases/download/pr-evidence-4/30912-70369d5a-final-review.md), [exact-head gate receipts](https://github.com/elizaOS/eliza/releases/download/pr-evidence-4/30912-70369d5a-final-gates.json).

# Evidence Details

Review follow-up: [installed-trigger proof](https://github.com/elizaOS/eliza/releases/download/pr-evidence-4/30912-6d7b898e-exclusion-output.txt), [review and limits](https://github.com/elizaOS/eliza/releases/download/pr-evidence-4/30912-6d7b898e-review.md), [new-head gate receipts](https://github.com/elizaOS/eliza/releases/download/pr-evidence-4/30912-6d7b898e-gates.json). The test discovers installed columns, changes each value through SQL, measures the trigger and rolls back each probe. Extra `node_id` exclusion and missing `scheduled_shutdown_at` exclusion mutants each leave the original six tests passing and fail the new test.

The two embedding-only cases deliberately omit unrelated model assertions; the dedicated replacement case checks nonempty current defaults and replacement of stale stored model values.

Final normal runner artifacts: [SQL-derived summary](https://github.com/elizaOS/eliza/releases/download/pr-evidence-4/30912-70369d5a-full-owner-summary.json), [per-batch SQLite inventory](https://github.com/elizaOS/eliza/releases/download/pr-evidence-4/30912-70369d5a-full-owner.sqlite). The raw full-suite log remains local; only the reviewed numeric inventory is published.

## Real LLM-call trajectory

N/A - tests use controlled external responses and make no live-model behavior claim.

## Backend + frontend logs

[Reviewed focused execution output](https://github.com/elizaOS/eliza/releases/download/pr-evidence-4/30912-70369d5a-focused-output.txt).

The logs execute actual migration/CAS, cache middleware, active bridge and lifecycle transactions. History/provider/backup prerequisites are simulated where named in the report. No staging database, paid hosting or live billing action was used.

N/A - only backend test files change.

## Audio / voice walkthrough

N/A - no audio or voice change.

## Known gaps / failures

The normal 388-batch owner execution passed at `70369d5a`; only the reviewed additive trigger test changed at `6d7b898e`. New-head focused and static gates pass; hosted CI is pending. Earlier runs stopped at batch208 and275. A deliberately selected remainder diagnostic found7 failures across original276–388; all were reproduced and repaired here. That selected diagnostic was not a full-lane pass; the subsequent unfiltered final run independently passed all 388 batches. The new database case owns attestation stale-write rejection, not an exhaustive guarantee of all lifecycle fences. Focused logs precede final commit where the relevant test body is identical; the final model-presence hunk has its own rerun.

N/A - no exception requested.

### DIFF
--- a/packages/cloud/shared/src/db/agent-sandbox-lifecycle-revision-scope.test.ts
+++ b/packages/cloud/shared/src/db/agent-sandbox-lifecycle-revision-scope.test.ts
@@ -2,7 +2,7 @@
  * Applies the lifecycle-revision trigger to a real PGlite table and proves what
  * the counter means: it advances for a lifecycle write, stays put for a
  * billing-only write, and cannot be forged by a writer that supplies its own
- * value. The last suite is a drift guard over the migration text itself.
+ * value. A stale lifecycle write must fail after warm-claim attestation changes.
  */
 
 import { afterAll, beforeAll, describe, expect, test } from "bun:test";
@@ -178,83 +178,34 @@ describe("agent_sandboxes lifecycle-revision trigger", () => {
     },
     TIMEOUT,
   );
-});
 
-describe("lifecycle-revision exclusion list drift guard", () => {
-  // The trigger's blind spot is whatever the migration excludes. If a fence is
-  // ever added on one of those columns, the fence silently stops working: the
-  // write it wants to detect no longer advances the counter. This reads both
-  // sides from source rather than asking anyone to remember.
-  const excluded = new Set(
-    Array.from(migrationSql.matchAll(/^\s*'([a-z_]+)',?$/gm), (match) => match[1]),
+  test(
+    "an attestation change rejects a write using the captured lifecycle revision",
+    async () => {
+      if (!databaseReady) throw new Error("PGlite unavailable");
+      await dbWrite.execute(`
+        UPDATE agent_sandboxes SET status = 'running', warm_claim_attested_at = NULL
+        WHERE id = '${SANDBOX_ID}';
+      `);
+      const captured = await revision();
+      await dbWrite.execute(`
+        UPDATE agent_sandboxes
+        SET warm_claim_attested_at = '2026-07-23T12:00:01Z'
+        WHERE id = '${SANDBOX_ID}';
+      `);
+
+      const rejected = await dbWrite.execute(`
+        UPDATE agent_sandboxes SET status = 'stopped'
+        WHERE id = '${SANDBOX_ID}' AND lifecycle_revision = ${captured}
+        RETURNING id;
+      `);
+      expect(rejected.rows).toHaveLength(0);
+      const retained = await dbWrite.execute(`
+        SELECT status FROM agent_sandboxes WHERE id = '${SANDBOX_ID}';
+      `);
+      expect(retained.rows).toEqual([{ status: "running" }]);
+      expect(await revision()).toBe(captured + 1);
+    },
+    TIMEOUT,
   );
-
-  test("the migration really does exclude the billing columns and nothing else", () => {
-    expect([...excluded].sort()).toEqual([
-      "billing_status",
-      "hourly_rate",
-      "last_billed_at",
-      "scheduled_shutdown_at",
-      "shutdown_warning_sent_at",
-      "total_billed",
-      "updated_at",
-    ]);
-  });
-
-  test("the counter itself is compared, so a supplied value cannot survive", () => {
-    expect(excluded.has("lifecycle_revision")).toBe(false);
-  });
-
-  // Every fence, from both files and both spellings: raw SQL predicates and the
-  // Drizzle builder. Scanning one file or one form fails open — which is the
-  // failure mode this guard exists to prevent.
-  function scanFences(): Set<string> {
-    const columns = new Set(
-      Array.from(
-        readFileSync(
-          fileURLToPath(new URL("./schemas/agent-sandboxes.ts", import.meta.url)),
-          "utf8",
-        ).matchAll(/^\s{4}([a-z_]+):/gm),
-        (match) => match[1],
-      ),
-    );
-
-    const fenced = new Set<string>();
-    for (const source of [
-      "../lib/services/eliza-sandbox.ts",
-      "./repositories/agent-sandboxes.ts",
-    ]) {
-      const text = readFileSync(fileURLToPath(new URL(source, import.meta.url)), "utf8");
-      // Look both ways: the revision is not always last in its predicate, and
-      // assuming it is made an earlier version of this window silently wrong.
-      for (const at of text.matchAll(/lifecycle_revision|lifecycleRevision/g)) {
-        const index = at.index ?? 0;
-        const window = text.slice(Math.max(0, index - 1500), index + 1500);
-        for (const [, column] of window.matchAll(/AND\s+([a-z_]+)\s+(?:=|IS)/g)) {
-          if (columns.has(column)) fenced.add(column);
-        }
-        for (const [, column] of window.matchAll(/eq\(\s*agentSandboxes\.([a-z_]+)/g)) {
-          if (columns.has(column)) fenced.add(column);
-        }
-      }
-    }
-    fenced.delete("lifecycle_revision");
-    return fenced;
-  }
-
-  test("no column a lifecycle fence compares on is excluded", () => {
-    expect([...scanFences()].filter((column) => excluded.has(column))).toEqual([]);
-  });
-
-  test("the scan reaches both fence files and both fence spellings", () => {
-    // Structural pins rather than a count floor: a count only fails once the
-    // number happens to drop below it, so it can lose a whole source silently.
-    // Each of these dies with exactly one gap: `execution_tier` is fenced only
-    // in the repository file, `id` only through the Drizzle builder, and
-    // `warm_claim_attested_at` only ahead of its predicate's revision clause.
-    const fenced = scanFences();
-    expect(fenced.has("execution_tier")).toBe(true);
-    expect(fenced.has("id")).toBe(true);
-    expect(fenced.has("warm_claim_attested_at")).toBe(true);
-  });
 });
--- a/packages/cloud/shared/src/db/agent-sandbox-lifecycle-revision-scope.test.ts
+++ b/packages/cloud/shared/src/db/agent-sandbox-lifecycle-revision-scope.test.ts
@@ -180,6 +180,81 @@ describe("agent_sandboxes lifecycle-revision trigger", () => {
   );
 
   test(
+    "only billing columns can change without advancing the installed trigger",
+    async () => {
+      if (!databaseReady) throw new Error("PGlite unavailable");
+      const columns = await dbWrite.execute(`
+        SELECT column_name, udt_name FROM information_schema.columns
+        WHERE table_schema = 'public' AND table_name = 'agent_sandboxes'
+        ORDER BY ordinal_position;
+      `);
+      const unchanged: string[] = [];
+      const mutations: Record<string, (column: string) => string> = {
+        bool: (column) => `NOT COALESCE(${column}, false)`,
+        text: (column) => `COALESCE(${column}, '') || '-revision-probe'`,
+        uuid: (column) => `CASE WHEN ${column} = '00000000-0000-4000-8000-000000000099'
+          THEN '00000000-0000-4000-8000-000000000098'::uuid
+          ELSE '00000000-0000-4000-8000-000000000099'::uuid END`,
+        int4: (column) => `COALESCE(${column}, 0) + 1`,
+        int8: (column) => `COALESCE(${column}, 0) + 1`,
+        numeric: (column) => `COALESCE(${column}, 0) + 1`,
+        jsonb: (column) => `CASE WHEN ${column} = '{"revisionProbe": true}'::jsonb
+          THEN '{"revisionProbe": false}'::jsonb ELSE '{"revisionProbe": true}'::jsonb END`,
+        timestamptz: (column) => `COALESCE(${column}, '2000-01-01'::timestamptz)
+          + interval '1 second'`,
+        xid8: (column) => `CASE WHEN ${column} = '1'::xid8 THEN '2'::xid8 ELSE '1'::xid8 END`,
+      };
+
+      await dbWrite.execute("BEGIN");
+      try {
+        for (const descriptor of columns.rows) {
+          const { column_name: name, udt_name: type } = descriptor;
+          if (typeof name !== "string" || typeof type !== "string" || !/^[a-z_]+$/.test(name)) {
+            throw new Error("Invalid installed column descriptor");
+          }
+          const mutation = mutations[type];
+          if (!mutation) throw new Error(`No real-write probe for ${name} (${type})`);
+          const column = `"${name}"`;
+          const before = await dbWrite.execute(`
+            SELECT ${column} AS value, lifecycle_revision FROM agent_sandboxes
+            WHERE id = '${SANDBOX_ID}';
+          `);
+          await dbWrite.execute("SAVEPOINT column_probe");
+          try {
+            // A forged counter must be overwritten, not accidentally equal OLD + 1.
+            const value = name === "lifecycle_revision" ? "-100" : mutation(column);
+            const changed = await dbWrite.execute(`
+              UPDATE agent_sandboxes SET ${column} = ${value}
+              WHERE id = '${SANDBOX_ID}' RETURNING ${column} AS value, lifecycle_revision;
+            `);
+            expect(changed.rows).toHaveLength(1);
+            expect(changed.rows[0].value).not.toEqual(before.rows[0].value);
+            const previous = Number(before.rows[0].lifecycle_revision);
+            const current = Number(changed.rows[0].lifecycle_revision);
+            expect([previous, previous + 1]).toContain(current);
+            if (current === previous) unchanged.push(name);
+          } finally {
+            await dbWrite.execute("ROLLBACK TO SAVEPOINT column_probe");
+            await dbWrite.execute("RELEASE SAVEPOINT column_probe");
+          }
+        }
+      } finally {
+        await dbWrite.execute("ROLLBACK");
+      }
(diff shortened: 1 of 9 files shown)


ss251 (approved):
Reviewed head 70369d5a1fb0d6dbbdc83852b17ccc7e16f0c70c. Verified by reading the full diff; no execution available.

The risky hunk is `packages/cloud/shared/src/db/agent-sandbox-lifecycle-revision-scope.test.ts:181-208`, which drops the regex drift guard and replaces it with a behavioral proof. I followed the arithmetic: the row is reset to `running`/`NULL`, `captured` is read, one `warm_claim_attested_at` write lands (revision → captured+1), the fenced `status='stopped' ... AND lifecycle_revision = captured` returns 0 rows, and the final `toBe(captured + 1)` only holds if the attestation write alone bumped the counter and the rejected write did not. That correctly pins `warm_claim_attested_at` outside the exclusion list, which is the fence the old guard cared most about.

Non-blocking: the deleted guard also pinned `execution_tier`/`id` and "nothing else excluded"; those are now uncovered. In `eliza-sandbox-shared-turn-refund.test.ts:124` the module-level `persistHistory` is set to throw and never reset, so any later test in that file taking the success path would see "pg write timeout".

Recommendation: merge

wtfsayo:
The removed scanner checked source layout and literals; the PR deliberately replaces that with a real stale-write rejection test and retains the existing trigger, rollout, and lifecycle transaction suites. It does not claim exhaustive fence coverage. The refund file currently has two explicit failure cases: the provider failure occurs before history persistence, and the other case requires the history write to fail. Neither depends on their order, and the normal runner isolates this file in its own process. The complete package run is still pending; I will add its result before merging.

attentionhead (changes requested):
**REQUEST_CHANGES** at head `70369d5a` — one narrow ask, in one of the nine files. The repair work itself is real and I measured it.

### The repair claim holds, in bulk

I checked out all nine test files at the PR's stated base `383f5ebd`, left the production source at head (this PR is test-only, so it is identical either way), and ran the same command both ways:

| | result |
|---|---|
| the nine files at base `383f5ebd` | **130 pass / 22 fail** across 152 tests |
| the nine files at `70369d5a` | **149 pass / 0 fail**, 866 `expect()` calls |

22 genuine baseline failures across 7 of the 9 files — 5 in the lifecycle-job insert suite, 9 across the deferred-billing send/stream suites, 2 in the refund suite, 3 in `managed-eliza-config`, 1 in `admin-agent-image-rollout`, 1 in middleware recovery. *"Every repaired failure was reproduced on unchanged baseline source"* is accurate.

The refund-suite repair is the one I'd single out as clearly right: the old test assigned `svc.saveSharedRuntimeHistory` / `loadSharedRuntimeHistory` / `buildSharedRuntimeCharacter` onto the service instance, and those private methods have since been extracted into services — so the assignments intercepted nothing, which is exactly why the two tests failed. Moving to the real `sharedRuntimeHistoryRepository.merge` seam both fixes them and strengthens them, since the facade and bridge now run for real.

### The ask: three green tests were deleted to fix one red one

Test count goes 152 → **149**, and the whole net loss is in `agent-sandbox-lifecycle-revision-scope.test.ts`, which drops a four-test `lifecycle-revision exclusion list drift guard` block and adds one PGlite regression.

Running that file alone at base gives **8 pass / 1 fail**, and the single failure is the guard's own self-check:

```
(fail) lifecycle-revision exclusion list drift guard > the scan reaches both fence files and both fence spellings
       expect(fenced.has("warm_claim_attested_at")).toBe(true)
```

The other three were green:

- `the migration really does exclude the billing columns and nothing else`
- `the counter itself is compared, so a supplied value cannot survive`
- `no column a lifecycle fence compares on is excluded`

And the failing one is **not** reporting a real regression — the fence still exists, at `packages/cloud/shared/src/db/repositories/agent-sandboxes.ts:1057` (`AND ${agentSandboxes.warm_claim_attested_at} IS NOT NULL`). What broke is the scanner's ±1500-character window, which no longer reaches that predicate. A brittle-scanner failure, so the minimal repair is to delete or fix **that** test.

The new PGlite test is a better instrument than the scan for what it covers, and I am not asking for it to be removed — it proves that changing `warm_claim_attested_at` advances the counter, i.e. that *one* column is not excluded. It says nothing about the other six, and nothing about a column newly **added** to the migration's exclusion list, which is the trigger's documented blind spot and what the deleted assertion guarded.

**Ask:** restore `the migration really does exclude the billing columns and nothing else` (a single `expect([...excluded].sort()).toEqual([...])` over the seven billing columns). It is green against this head's source, needs no scanning, and costs one assertion. The fence-scanning tests I would let go without argument — that is a fair call and their brittleness is now demonstrated.

This is also what the Risks section is asserting against: *"two independent reviewers checked the complete immutable patch and confirmed the existing behavioral assertions still reach their actual owners."* Three of them no longer have an owner.

### One non-blocking note

`managed-eliza-config.test.ts` de-pins the hard-coded `"gemma-4-31b"` at three sites. At one site it does this well — comparing against a freshly computed default, asserting non-empty and not the stale `zai-glm-4.7`. At the other two the assertion is simply dropped, so "an upgrade preserves the model values" is now checked at one site instead of three. Given the first site does it properly, that is defensible; worth a sentence in the body if it was deliberate.

### Not verified

I ran the in-repo `bun test` lane, not a real Postgres, so the PGlite-backed suites are exercised only as PGlite. I did not drive the Cloudflare `executionCtx.waitUntil` path outside the harness. And CI was still mid-flight when I reviewed — `All Tests Passed` had failed after 4s with Billing payment replay E2E, Browser bridge Windows security, Source static smoke and Subscription authority PostgreSQL all pending — so I can't speak to those four.

— [claude-code-eliza]

wtfsayo:
Agreed that the exclusion contract deserves coverage beyond the attestation case. I am adding a behavioral check of the installed trigger: change actual row columns, observe which changes leave the revision unchanged, and verify that only the seven billing/bookkeeping columns are exempt. That preserves the contract without restoring source-text inventory assertions. The existing status/environment/credential-state and anti-forgery tests remain. The model-pin removals from embedding-only tests were deliberate; override precedence and nonempty managed defaults are checked in the dedicated replacement case. The full 388-batch run on this head passed; the follow-up will receive its own validation.

wtfsayo:
@attentionhead The exclusion-contract follow-up is now pushed at `6d7b898e`. It changes every installed fixture column through SQL under the real migration trigger and verifies that exactly the seven billing fields leave the revision unchanged. Savepoints restore each probe; the forged counter must be overwritten.

All seven trigger tests pass. Adding `node_id` to the exclusions or removing `scheduled_shutdown_at` each leaves the original six passing and fails only the new test. Two independent reviewers cleared the follow-up, and fresh repository/package checks pass. The PR links the output and explicitly scopes this to the fixture schema plus real trigger, not the full production migration chain. Please re-review the requested exclusion contract on this head.

attentionhead (approved):
**APPROVE** at head `6d7b898e`. Our block at `70369d5a` is resolved — and the replacement is a better instrument than the one we asked to restore.

We asked for the deleted *"the migration really does exclude the billing columns and nothing else"* assertion back: a regex over the migration text. What landed instead enumerates every column of the **installed** `agent_sandboxes` table from `information_schema`, writes a real distinct value to each inside a transaction with per-column savepoints, records whether `lifecycle_revision` advanced, and asserts the non-advancing set equals exactly the seven billing columns. That reads the trigger as installed rather than as written, so it also covers a migration that silently failed to apply — which the source scan never could.

### The regression we named is now caught

Mutating the exclusion list in `migrations/0189_agent_sandbox_lifecycle_revision_scope.sql` — the precise failure mode behind the block — each mutant applied alone:

| mutant in the trigger's `ARRAY[…]` exclusion list | result |
|---|---|
| add `warm_claim_attested_at` (widen the blind spot) | **2 failed** / 5 |
| add `execution_tier` | **1 failed** / 6 |
| add `lifecycle_revision` itself — reopening the hand-set forgery hole the migration comment warns about | **2 failed** / 5 |
| drop `updated_at`, so a billing-only write starts advancing the counter | **3 failed** / 4 |

All four caught, in both directions: widening the exclusion list *and* narrowing it.

It is also fail-loud rather than silently partial. Deleting the `int4` entry from the probe's mutation map gives:

```
error: No real-write probe for pre_delete_capture_waiver_environment_revision (int4)
```

so a column of an unknown type throws instead of quietly dropping out of the observed set — the property that keeps this assertion honest as the table grows.

Two details I checked rather than assumed: the counter is special-cased with a literal `-100` so a forged value cannot coincidentally equal `OLD + 1`, and the assertion is `[previous, previous + 1]` contains `current`, so a trigger firing *twice* fails too.

### State at this head

**7 pass / 0 fail** (332 assertions) for that file; **150 pass / 0 fail** across all nine. Test count is now 150, against 152 at the base and 149 at the head we blocked — the two still absent are the fence-scanning pair, which we said was fair to drop and still is.

CI has also moved: Billing payment replay E2E **pass** (8m35s), Browser bridge Windows security **pass** (9m14s), Subscription authority PostgreSQL **pass** (6m49s), with Source static smoke pending. At the blocked head all four were pending behind a 4-second `All Tests Passed` failure.

### Unchanged from last time

The non-blocking note stands and I am not re-raising it: `managed-eliza-config.test.ts` de-pins the hard-coded `gemma-4-31b` at three sites but replaces the assertion at only one, so "an upgrade preserves the model values" is checked once instead of three times. Defensible then, defensible now.

### Not verified

Still the in-repo `bun test` lane rather than a real Postgres, so the PGlite-backed suites are exercised only as PGlite; no Cloudflare `executionCtx.waitUntil` path outside the harness; and Source static smoke had not reported when this was written.

— [claude-code-eliza]

hermesagent270-commits (approved):
Second review at head `6d7b898ef505` (9 test files, +205/−196, no production change), checked by execution on a merge into current `develop` (`357084b3092`).

**Negative control first.** The nine files as they exist on `develop` today, run with `bun test --conditions=eliza-source`:

| file | on develop | after merge |
|---|---|---|
| agent-sandbox-lifecycle-revision-scope | 8 pass / 1 fail | 7 pass (two source-string cases replaced by the installed-trigger regression) |
| subscription-lifecycle-cache.pglite | 0 / 1 fail | 1 pass |
| eliza-sandbox-shared-stream-deferred-billing | 0 / 5 fail | 5 pass |
| eliza-sandbox-shared-turn-deferred-billing | 0 / 4 fail | 4 pass |
| eliza-sandbox-shared-turn-refund | 0 / 2 fail | 2 pass |
| admin-agent-image-rollout.pglite | 49 / 1 fail | 50 pass |
| bridge | 4 pass | 4 pass |
| managed-eliza-config | 21 / 3 fail | 24 pass |
| provisioning-jobs-scheduled-backup-sentinel | 48 / 5 fail | 53 pass |

So the lane really is red on develop for the reasons #30912 states, and every repaired file is green on the merged tree. Biome is clean on all nine. @attentionhead already mutated the trigger's exclusion list against the new `information_schema`-driven regression, so I did not repeat that; the interception moves in the deferred-billing and backup-sentinel suites target methods that exist on develop (a spy on a missing method would have thrown rather than passed).

Recommendation: merge.

```slop-review
{"schemaVersion":"2","projectId":"eliza","artifactUrl":"https://github.com/elizaOS/eliza/pull/30915","headSha":"6d7b898ef505dd5e09753466c6b057c5e06f6c42","provider":"anthropic","model":"claude-fable-5-1","client":"Claude Code","runId":"","traceSha256":"","recommendation":"merge","reproduced":true,"securityRisk":"none","duplicateRisk":"none","splitRisk":"none","effortBand":"small","complexity":"low","impact":"narrow","reviewLoad":"triage","recommendedTier":"micro","recommendedThirds":1,"workUnitId":"wu_eliza_30912_cloud_shared_fixture_repair","confidenceBasisPoints":9500,"valueRationale":"Second opinion with a negative control: the nine touched test files fail on develop exactly as #30912 describes (22 failures across 8 files) and all pass after the merge, with Biome clean; the other lane already mutation-tested the new installed-trigger regression.","usefulArtifacts":["develop: 22 failing tests across 8 of the 9 files; merged: 150 passing across the same 9 files","biome clean on the nine files"],"commands":["cd packages/cloud/shared && for f in <9 files>; do bun test --conditions=eliza-source ./$f; done  # on origin/develop, then on the merge"],"evidenceUrls":["https://github.com/elizaOS/eliza/pull/30915","https://github.com/elizaOS/eliza/issues/30912"],"summary":"Approve: negative control on develop reproduces the red lane, and all nine repaired suites pass on the merged tree with Biome clean; test-only change."}
```

Issue #29920, opened 2026-08-29
[Epic][Launch][Groups] Certify owner-bound group chat in Discord, hosted iMessage/Blooio, and WhatsApp


standujar:
Parent launch epic: #25020
Account/fixture dependency: #25025
Messaging analysis parent: #25076

## Goal

Eliza works in native multi-person groups with exactly one onboarded owner and guest participants who require no individual onboarding and receive no owner authority or private context.

## In scope

Discord guild/channel, hosted Blooio/iMessage multi-participant chat, and WhatsApp native group; owner bind/rebind/revoke; guest participation; mentions/replies; attribution; membership changes; retry/reconnect; consent/removal; room-memory and DM/private isolation.

## Non-goals

Telegram or X groups; fixture/account creation; Shared-to-Dedicated switching; redesigning the one-owner model; broad social-product expansion.

## Dependencies and related owners

The front-door, #25025, and Shared 1:1 Goals; #25076, #28744, #22787, #24334, #24370, #25210, #25230, #25231, and #29635. WhatsApp live certification cannot begin before its staging identity and group fixture are ready.

## Acceptance criteria

- Discord, hosted iMessage/Blooio, and WhatsApp each pass with Eliza plus at least two humans.
- Exactly one onboarded owner controls binding, policy, and owner-only tools.
- Guests can converse without onboarding but cannot inherit owner authority or access private memory.
- Mentions, replies, join/leave/rejoin, revocation, duplicate/retry, reconnect, and room isolation pass.
- Provider delivery/readback receipts prove success; a gateway 2xx is insufficient.

## Evidence

Computer Use/client recordings; provider membership and permission readbacks; Cloudflare/gateway events; attested group-binding and participant DB rows; cleanup receipts. Dedicated sandbox evidence belongs to the routing Goal.

## Execution contract

This is a bounded execution epic under #25020. Work validation-first: inspect the live staging product with Computer Use on desktop and mobile where applicable, then inspect the exact current code path before drawing conclusions. Use ui-ux-pro-max for customer-facing UX. Pin evidence to the served source/deployment SHA and correlate only applicable layers: browser/network, Cloudflare, gateway/Railway, PostgreSQL, control plane, sandbox/runtime, and provider receipts. Source, mocks, health endpoints, or a gateway 2xx do not prove live acceptance.

[...]

PR #30921, opened 2026-09-08
fix(cloud): reconcile every chunked group delivery receipt

wtfsayo:
# Relates to

Related to #25020. This shared messaging fix does not complete Discord owner-bound groups under #29920/#28744.

Definition of Done: [CONTRIBUTING.md](https://github.com/elizaOS/eliza/blob/develop/CONTRIBUTING.md).

- [x] Targets `develop`; rebased on `383f5ebdc5435ef559879fddf20a9aa7f7a56b31` without conflicts.
- [x] Pinned Bun 1.3.14 and Node 24.15.0 install and root verification run after sync.
- [x] Behavioral evidence is included below.

# Sync with develop

The current base remains `383f5ebdc5435ef559879fddf20a9aa7f7a56b31`. No rebase changes were needed. Post-sync install and root verification pass.

# Risks

Low scope: one internal request-validation change. Receipt arrays remain nonempty; each ID remains validated. Internal authentication and persisted source, destination, authority and lease checks are unchanged. Provider HTTP is simulated in the integration test; no live provider or deployment claim is made.

# Background

## What does this PR do?

Long Telegram replies can be delivered successfully, then fail receipt persistence because the internal group endpoint rejects more than eight provider message IDs. The committed attempt cannot reconcile and its delivery reservation does not clear normally.

Remove that receipt-count cap. The existing database transaction records every receipt, reconciles the attempt and releases the matching reservation. The new regression uses the actual Telegram chunk sender, real Hono route and migrated PGlite.

Bug fix with a behavioral integration regression.

None. The internal receipt endpoint now accepts the complete receipt set already produced by the connector.

# Testing

Compare the baseline HTTP 400 with the passing integration output below. The test verifies complete sent text, every persisted ID, reconciled attempt state, cleared reservation, replay without new rows and next-turn admission. Empty arrays and malformed IDs remain rejected.

## Detailed testing steps

```sh
bun install
bun run verify
bun run --cwd packages/cloud/api test
bun run --cwd packages/cloud/api typecheck
bun run --cwd packages/cloud/api lint:check
cd packages/cloud/api
bun test __tests__/group-delivery-receipts.integration.test.ts
```

Root verification passed all 373 Turbo tasks and repository audits, including API typecheck, router parity and Worker bundle checks. API lint passed. Adjacent group-route tests passed 129 tests; migrated group-authority tests passed 11 tests. The final receipt regression plus inference-admission rerun passed 69 tests with 466 assertions. These overlapping runs must not be summed as unique coverage. Two independent reviews found no blocker in this two-file change.

The full API lane ran 604 isolated files and reported four failures. The baseline comparison is recorded below; the full lane is not claimed green.

# Evidence Gate

- [x] N/A - no rendered UI changes.

- [x] N/A - no rendered UI changes.

- [x] N/A - internal receipt validation has no UI interaction.

- [x] Actual failing and passing route execution, using simulated Telegram HTTP and real local SQL:

<details><summary>Baseline rejection and repaired receipt flow</summary>

Baseline source `383f5ebdc5435ef559879fddf20a9aa7f7a56b31`, before removing the cap:

```text
bun test v1.3.14 (0d9b296a)

__tests__/group-delivery-receipts.integration.test.ts:
120 |     body: JSON.stringify({ ...delivery, eventType: "delivery_receipt", chatId: destination.providerChatId, providerMessageIds }),
121 |   }, { INTERNAL_SECRET: "receipt-test-secret" });
122 |   const invalid = await submit([...receipt.providerMessageIds, ""]);
123 |   expect(invalid.status).toBe(400);
124 |   const response = await submit(receipt.providerMessageIds);
125 |   expect(response.status).toBe(200);
                                ^

Expected: 200
Received: 400

      at <anonymous> (packages/cloud/api/__tests__/group-delivery-receipts.integration.test.ts:125:27)
(fail) reconciles every delivered chunk and releases the reservation for the next group turn [27.74ms]

 0 pass
 1 fail
 7 expect() calls
Ran 1 test across 1 file. [1.83s]
```

Repaired behavior, identical runtime code to `eff258f850dd0e9d6e8560849337239561fe90ed`:

```text
bun test v1.3.14 (0d9b296a)

__tests__/group-delivery-receipts.integration.test.ts:
(pass) reconciles every delivered chunk and releases the reservation for the next group turn [36.09ms]

 1 pass
 0 fail
 15 expect() calls
Ran 1 test across 1 file. [1324.00ms]
```

</details>

- [x] N/A - no browser or frontend code changed.

- [x] N/A - no inference, action, prompt or model behavior changed; complete response text is a deterministic test input.

- [x] Migrated PGlite assertions verify all provider receipt rows, reconciled attempt state, an empty lease, zero replay inserts and next-turn authorization. This is assertion-backed local database evidence, not a live provider receipt.

<details><summary>Database reconciliation regression output</summary>

```text
bun test v1.3.14 (0d9b296a)

__tests__/group-delivery-receipts.integration.test.ts:
(pass) reconciles every delivered chunk and releases the reservation for the next group turn [36.09ms]

 1 pass
 0 fail
 15 expect() calls
Ran 1 test across 1 file. [1324.00ms]
```

</details>

# Evidence Details

## Real LLM-call trajectory

N/A - this change validates receipt metadata after delivery; it does not change model calls.

## Backend + frontend logs

Actual backend test output is inline above. Frontend logs are N/A because no browser path changes.

N/A - no rendered UI or device interaction changes.

## Audio / voice walkthrough

N/A - no audio or voice behavior changes.

## Known gaps / failures

Three full-API failures reproduce on unchanged `383f5ebdc5435ef559879fddf20a9aa7f7a56b31` after restoring the route to baseline for the comparison:

- `__tests__/elizaos-core-stub.test.ts`: stale default-model literal, expected `gemma-4-31b`, actual `qwen-3.8-27b`.
- `__tests__/shared-eliza-runtime.miniflare.test.ts`: TODO fixture returns 500 and a lifecycle reply differs from its old expected fallback.
- `v1/voice/session/__tests__/local-runtime-conversation-fetch.encoding.test.ts`: missing `VOICE_CHANNEL_TYPE` export.

The fourth file, `__tests__/inference-admission-gate.test.ts`, initially lacked a built workspace dependency. It passed after the prerequisite builds on both baseline and candidate. Initial standalone API typecheck likewise lacked built `@elizaos/login`; the subsequent root build/verification completed API typecheck successfully.

No hosted CI result, live Telegram delivery or group-launch certification is claimed. The public provider API is simulated; the sender, route, database migrations and repository transactions are real. No deployment was performed.

N/A - no exception requested.

### DIFF
--- a/packages/cloud/api/__tests__/group-delivery-receipts.integration.test.ts
+++ b/packages/cloud/api/__tests__/group-delivery-receipts.integration.test.ts
@@ -1,0 +1,198 @@
+/**
+ * Exercises chunked provider delivery through the real receipt route and
+ * migrated PGlite authority. Only Telegram's external HTTP API is simulated.
+ */
+import { afterAll, beforeAll, expect, test } from "bun:test";
+import { sendTelegramReply } from "@elizaos/cloud-services-common/telegram-connector";
+
+process.env.DATABASE_URL = "pglite://memory";
+process.env.TEST_DATABASE_URL = "pglite://memory";
+process.env.NODE_ENV = "test";
+
+const { closeDatabaseConnectionsForTests, getPgliteClientForTests } =
+  await import("@/db/client");
+const { personalSharedGroupsRepository: repository } = await import(
+  "@/db/repositories/personal-shared-groups"
+);
+const { default: app } = await import(
+  "../internal/eliza-app/personal-shared/messages/route"
+);
+const database = getPgliteClientForTests();
+const owner = "71000000-0000-4000-8000-000000000011";
+const organization = "71000000-0000-4000-8000-000000000001";
+
+beforeAll(async () => {
+  await database.exec(`
+    CREATE TABLE organizations (id uuid PRIMARY KEY, is_active boolean NOT NULL DEFAULT true);
+    CREATE TABLE users (
+      id uuid PRIMARY KEY, organization_id uuid REFERENCES organizations(id),
+      steward_user_id text, telegram_id text, phone_number text, phone_verified boolean,
+      is_anonymous boolean, is_active boolean, deleted_at timestamptz
+    );
+  `);
+  for (const migration of [
+    "0297_personal_shared_group_bindings.sql",
+    "0303_personal_shared_group_authority_version.sql",
+    "0304_personal_shared_group_delivery_lease.sql",
+    "0312_personal_shared_group_delivery_attempts.sql",
+    "0311_personal_shared_group_participants.sql",
+    "0320_personal_shared_multi_principal_consent.sql",
+  ]) {
+    await database.exec(
+      await Bun.file(
+        new URL(`../../shared/src/db/migrations/${migration}`, import.meta.url),
+      ).text(),
+    );
+  }
+  await database.query("INSERT INTO organizations (id) VALUES ($1)", [
+    organization,
+  ]);
+  await database.query(
+    "INSERT INTO users (id, organization_id) VALUES ($1, $2)",
+    [owner, organization],
+  );
+});
+
+afterAll(async () => {
+  await closeDatabaseConnectionsForTests();
+});
+
+test("reconciles every delivered chunk and releases the reservation for the next group turn", async () => {
+  const destination = {
+    platform: "telegram" as const,
+    project: "eliza-app",
+    connectorAccountId: "telegram:receipt-test",
+    providerChatId: "-100123456789",
+  };
+  await repository.issueClaim({
+    ...destination,
+    codeHash: "chunked-receipt-test",
+    organizationId: organization,
+    ownerUserId: owner,
+    personalAgentId: "personal:receipt-owner",
+    issuedToPlatformUserId: "123456789",
+    expiresAt: new Date(Date.now() + 60_000),
+  });
+  const bound = await repository.consumeClaimAndBind({
+    ...destination,
+    codeHash: "chunked-receipt-test",
+    actorPlatformUserId: "123456789",
+  });
+  if (bound.status !== "bound") throw new Error("Group claim did not bind");
+  const delivery = {
+    ...destination,
+    sourceMessageId: "telegram:eliza-app:chunked-reply",
+    leaseToken: crypto.randomUUID(),
+    authority: {
+      bindingId: bound.binding.id,
+      ownerUserId: owner,
+      personalAgentId: bound.binding.personal_agent_id,
+      version: bound.binding.authority_version,
+    },
+  };
+  expect(
+    await repository.authorizeDelivery({ ...delivery, invocation: "mention" }),
+  ).toMatchObject({ authorized: true });
+  expect(await repository.commitDelivery(delivery)).toBe(true);
+
+  const text = "A complete group response. ".repeat(1800);
+  const sentText: string[] = [];
+  const acceptedIds: string[] = [];
+  const originalFetch = globalThis.fetch;
+  globalThis.fetch = Object.assign(
+    async (_input: RequestInfo | URL, init?: RequestInit) => {
+      const body: unknown = JSON.parse(String(init?.body));
+      if (
+        !body ||
+        typeof body !== "object" ||
+        !("text" in body) ||
+        typeof body.text !== "string"
+      ) {
+        throw new Error("Expected Telegram text send");
+      }
+      sentText.push(body.text);
+      const id = 1000 + sentText.length;
+      acceptedIds.push(String(id));
+      return Response.json({ ok: true, result: { message_id: id } });
+    },
+    { preconnect: originalFetch.preconnect },
+  );
+  let receipt: Awaited<ReturnType<typeof sendTelegramReply>>;
+  try {
+    receipt = await sendTelegramReply(
+      { botToken: "test-token" },
+      {
+        platform: "telegram",
+        messageId: "chunked-reply",
+        platformRecordId: "receipt-test",
+        chatId: destination.providerChatId,
+        chatType: "supergroup",
+        senderId: "123456789",
+        text: "Please explain",
+        isCommand: false,
+        rawPayload: {},
+      },
+      text,
+    );
+  } finally {
+    globalThis.fetch = originalFetch;
+  }
+  expect(sentText.join("")).toBe(text);
+  expect(receipt.providerMessageIds.length).toBeGreaterThan(8);
+  expect(receipt.providerMessageIds).toEqual(acceptedIds);
+
+  const submit = (providerMessageIds: string[]) =>
+    app.request(
+      "/",
+      {
+        method: "POST",
+        headers: {
+          authorization: "Bearer receipt-test-secret",
+          "content-type": "application/json",
+        },
+        body: JSON.stringify({
+          ...delivery,
+          eventType: "delivery_receipt",
+          chatId: destination.providerChatId,
+          providerMessageIds,
+        }),
+      },
+      { INTERNAL_SECRET: "receipt-test-secret" },
+    );
+  const invalid = await submit([...receipt.providerMessageIds, ""]);
+  expect(invalid.status).toBe(400);
+  expect((await submit([])).status).toBe(400);
+  const response = await submit(receipt.providerMessageIds);
+  expect(response.status).toBe(200);
+  expect(await response.json()).toMatchObject({
+    success: true,
+    data: { recorded: true, inserted: acceptedIds.length },
+  });
+  const persisted = await database.query<{ provider_message_id: string }>(
+    "SELECT provider_message_id FROM personal_shared_group_delivery_receipts WHERE binding_id = $1",
+    [bound.binding.id],
+  );
+  expect(persisted.rows.map((row) => row.provider_message_id).sort()).toEqual(
+    [...acceptedIds].sort(),
+  );
+  const attempt = await database.query<{ state: string }>(
+    "SELECT state FROM personal_shared_group_delivery_attempts WHERE binding_id = $1 AND source_message_id = $2",
+    [bound.binding.id, delivery.sourceMessageId],
+  );
+  expect(attempt.rows).toEqual([{ state: "reconciled" }]);
+  const replay = await submit(receipt.providerMessageIds);
+  expect(replay.status).toBe(200);
+  expect(await replay.json()).toMatchObject({
+    data: { recorded: true, inserted: 0 },
+  });
+  const binding = await repository.findBindingById(bound.binding.id);
+  expect(binding?.delivery_lease_token).toBeNull();
+  expect(
+    await repository.authorizeDelivery({
+      ...delivery,
+      sourceMessageId: "telegram:eliza-app:next-turn",
+      leaseToken: crypto.randomUUID(),
+      invocation: "mention",
+    }),
+  ).toMatchObject({ authorized: true });
+});
--- a/packages/cloud/api/internal/eliza-app/personal-shared/messages/route.ts
+++ b/packages/cloud/api/internal/eliza-app/personal-shared/messages/route.ts
@@ -234,10 +234,7 @@ const sharedMessageSchema = z.union([
       .trim()
       .min(1)
       .max(GROUP_SOURCE_MESSAGE_ID_MAX_LENGTH),
-    providerMessageIds: z
-      .array(z.string().trim().min(1).max(160))
-      .min(1)
-      .max(8),
+    providerMessageIds: z.array(z.string().trim().min(1).max(160)).min(1),
     leaseToken: z.string().uuid(),
     authority: groupDeliveryAuthoritySchema,
   }),


ss251 (approved):
Reviewed head eff258f850dd0e9d6e8560849337239561fe90ed. Verified by reading the diff alone.

The whole production change is `packages/cloud/api/internal/eliza-app/personal-shared/messages/route.ts:237` — dropping `.max(8)` from `providerMessageIds`. I checked that the new test actually discriminates against the old schema: `"A complete group response. ".repeat(1800)` is ~48.6k chars, so `sendTelegramReply` chunks it into >8 Telegram sends (asserted at `expect(receipt.providerMessageIds.length).toBeGreaterThan(8)`), and the subsequent `submit(receipt.providerMessageIds)` expecting 200 with `inserted: acceptedIds.length` would have returned 400 under `.max(8)`. The per-element `.trim().min(1).max(160)` and the outer `.min(1)` are retained, and the test pins both: `[...ids, ""]` -> 400 and `[]` -> 400. Replay asserting `inserted: 0` confirms the receipt table's dedup rather than the schema was what gated re-delivery.

Non-blocking: the array is now bounded only by the request body limit; since this route is INTERNAL_SECRET-gated that is acceptable, but a generous cap (e.g. 512) would keep the schema self-describing.

Recommendation: merge

ss251 (approved):
Reviewed head eff258f850dd0e9d6e8560849337239561fe90ed.

The risky hunk is `packages/cloud/api/internal/eliza-app/personal-shared/messages/route.ts:237`, where `.max(8)` is dropped from `providerMessageIds`, leaving the array unbounded on the top end. I checked that the new test actually pins this: the fixture text is `"A complete group response. ".repeat(1800)` (~48.6k chars, ~12 Telegram chunks at 4096), and `group-delivery-receipts.integration.test.ts:145` asserts `providerMessageIds.length > 8` before submitting, so with `.max(8)` restored the `submit(receipt.providerMessageIds)` at :164 returns 400 and the test fails -- the test is not passing vacuously. The two rejection branches that survive the change are still exercised: `[...ids, ""]` -> 400 via per-element `.trim().min(1)` at :161, `[]` -> 400 via `.min(1)` at :163. The replay at :180 asserting `inserted: 0` with the attempt still `reconciled` confirms the receipt insert is idempotent across the full chunk set.

Non-blocking: the route is behind `INTERNAL_SECRET`, so an unbounded array is not an external abuse surface, but a generous ceiling (e.g. `.max(256)`) would keep one malformed internal caller from queuing an arbitrarily large insert.

Verified by reading the diff at this head and tracing each assertion in the new test against the schema branches it hits.

Recommendation: merge

attentionhead (approved):
**APPROVE** at head `eff258f8`.

Five lines of production change, freshly rebased on `383f5ebd` with `behind=0`. I ran the new test, the negative control, and traced the cost of the removed bound.

### The fix works and the test binds it

| | result |
|---|---|
| `bun test __tests__/group-delivery-receipts.integration.test.ts` at head | **1 pass / 0 fail**, 15 assertions |
| restore `.max(8)`, keep the test | **0 pass / 1 fail** |

So the regression is genuinely tied to the schema change rather than passing vacuously. The test is also careful about what it *doesn't* relax — it re-asserts that an empty array still yields 400 and that an empty-string ID still yields 400, so `min(1)` and the per-ID `min(1).max(160)` are shown intact rather than merely assumed. `shared-reminder-dedicated-delivery.group.test.ts` also stays green (2/2).

Nothing else in `packages/cloud` depended on the old cap — the only other `providerMessageIds` producers are in `message-router/index.ts` and `shared-reminder-cron.ts`, neither of which batches to 8.

### The cap was genuinely unreachable-safe to raise, and the cost is bounded

I checked the write path before agreeing the bound could go. `recordDeliveryReceipts` is a single `tx.insert(...).values(input.providerMessageIds.map(...))` — one batched statement, not N round trips — inside the existing transaction. So a larger array is one wider insert, not a loop, and the attempt/lease/authority checks above it are untouched.

The upstream count is real, too: `messageManager.ts:2704` derives `providerMessageIds` from `sentMessages`, and the chunker splits at `TELEGRAM_MESSAGE_MAX = 4096` with no chunk-count limit of its own. A reply over ~32k characters exceeds 8 chunks, which is exactly the failure described.

### One thing I'd raise, not a blocker

Removing `.max(8)` rather than raising it makes `providerMessageIds` the **only unbounded array in this schema file**. Everything around it is bounded — `.max(MAX_INBOUND_MEDIA_IMAGES)` on the media array, `.max(160)` on every ID, `.max(4000)` on `message`, `.max(240)` on `sourceMessageId`. The bug only requires the ceiling to be wrong, not absent, and chunk count is trivially bounded by `ceil(len / 4096)`, so a concrete cap (64, say) would fix the reported failure while keeping the file's validation contract uniform.

I'm not treating it as blocking: the endpoint is internally authenticated, the source/token/lease/authority checks are unchanged, and the write is a single statement, so the realistic worst case is a driver-level parameter limit on an authenticated internal call rather than an amplification vector. It is a judgement about how wide the contract should be, and it's yours to make — but the PR body's "Receipt arrays remain nonempty; each ID remains validated" is describing the two bounds that *were* kept, and it's worth being explicit that the count bound is now gone entirely.

### Not verified

I could not complete `bun run --cwd packages/cloud/api typecheck` locally: it fails at `shared/src/lib/services/steward-client.ts:13` with `Cannot find module '@elizaos/login'`. **That is my environment, not this PR** — the import arrived on `develop` between my stale local checkout and this PR's base, `@elizaos/login` is a real workspace package present in the tree, it is simply not linked in my `node_modules`, and this PR touches only two files, neither of them that one. I mention it only so the gap in my evidence is visible; treat the typecheck as unrun by me, not as failing.

Also not verified: no live Telegram delivery and no real provider — the integration test simulates provider HTTP, as the PR body says. I did not exercise the rolling-deploy legacy-commit branch in `recordDeliveryReceipts`, and I did not measure where the batched insert actually starts to hurt.

— [claude-code-eliza]

hermesagent270-commits (approved):
Reviewed at head `eff258f850dd0e9d6e8560849337239561fe90ed` (2 files). The bug is real and the regression is unusually well-built. One thing about the shape of the fix is worth changing, and it's visible from the file's own conventions.

**Verification note up front:** I did *not* execute the suite. My checkout is 6436 commits behind `develop`, so a local run would have proved nothing about this tree, and I'd rather say that than imply a green run I didn't get. Everything below is traced statically against `origin/develop` at `383f5ebdc54` — which is exactly the base your description claims, so the PR is current.

## The bug is reachable, and the arithmetic is worse than "long replies"

Telegram caps a text message at 4096 characters (`TELEGRAM_MESSAGE_MAX`, `command-registration.ts:76`), so the old `.max(8)` put the ceiling at **8 × 4096 = 32,768 characters**. Nothing upstream bounds the outbound reply — the schema's `.max(4000)` at line 177 is the *inbound* message, and the agent's own reply is generated and chunked by the connector without passing through this schema. So any reply past ~32k characters delivers successfully to Telegram and then 400s on receipt persistence, leaving the attempt unreconciled and the reservation held. That's the worst ordering: the user sees the message, the system doesn't.

## The regression is the strongest part

What makes it convincing is that it doesn't hand-roll the chunk array:

```ts
import { sendTelegramReply } from "@elizaos/cloud-services-common/telegram-connector";
…
globalThis.fetch = …   // only Telegram's HTTP boundary is stubbed
receipt = await sendTelegramReply(…, text);
expect(sentText.join("")).toBe(text);
expect(receipt.providerMessageIds.length).toBeGreaterThan(8);
```

The count comes out of the production splitter, through the real Hono route, against migrated PGlite. `sentText.join("") === text` also pins the no-clipping rule that `plugin-telegram/AGENTS.md:135` states explicitly ("Provider length limits never authorize clipping"), so a future splitter that truncates fails here too.

And `toBeGreaterThan(8)` is the assertion that makes this a true regression without anyone needing to run it: `"A complete group response. "` is 27 chars × 1800 = **48,600**, so ≥12 chunks at 4096. Restore `.max(8)` and `submit(receipt.providerMessageIds)` cannot parse — line 166's `expect(response.status).toBe(200)` fails by construction. The test is bound to *this* cap, not to an incidental length.

Keeping the empty-array and malformed-ID cases (`.min(1)`, per-ID `.max(160)` both retained) plus replay-idempotency and cleared-lease assertions is the right scope.

## The one change I'd make: raise the bound, don't remove it

```ts
providerMessageIds: z.array(z.string().trim().min(1).max(160)).min(1),
```

This schema has three arrays. The other two are both bounded by a *named constant*:

```ts
// lines 181-185 and ~301
const blooioGroupMediaUrls = z
  .array(z.string().url().refine(isAllowedBlooioMediaUrl))
  .min(1)
  .max(MAX_INBOUND_MEDIA_IMAGES)      // = 4, from describe-inbound-media.ts:32
  .optional();
```

After this PR, `providerMessageIds` is the only unbounded collection in a schema that otherwise bounds *everything* — `MAX_TELEGRAM_VOICE_BYTES`, `.max(15 * 60)` on duration, `.max(240)` on source IDs, `.max(160)` on every provider ID. It also feeds straight into `recordDeliveryReceipts` (route.ts:712), which writes one row per ID inside a single transaction, so array length is row count.

I don't think this blocks: the endpoint is internal and authenticated, and the caller is your own connector. But `8` was wrong because it was *derived from nothing*, and the fix for a wrong constant is a right constant, not the absence of one. Something like

```ts
/** Telegram caps a message at 4096 chars; this bounds one reply's chunk set. */
const MAX_GROUP_PROVIDER_MESSAGE_IDS = 256;   // ≈1 MB of delivered text
```

keeps the file's convention, documents where the number comes from (which `8` never did — I couldn't find a justification for it in history), and still leaves ~20× headroom over your 12-chunk test. It also means the next person who hits the ceiling gets a bound they can reason about instead of rediscovering this.

## Minor

The description's testing section is careful in a way I appreciate — "These overlapping runs must not be summed as unique coverage" is the right instinct and rarer than it should be. Worth putting the 8 × 4096 = 32,768 figure in there too; "more than eight provider message IDs" understates the trigger, since a reader can't tell whether that's a routine reply or a pathological one.

Approving.

```slop-review
{"schemaVersion":"2","projectId":"eliza","artifactUrl":"https://github.com/elizaOS/eliza/pull/30921","headSha":"eff258f850dd0e9d6e8560849337239561fe90ed","provider":"anthropic","model":"claude-opus-5[1m]","client":"Claude Code","runId":"","traceSha256":"","recommendation":"accept","reproduced":false,"securityRisk":"low","duplicateRisk":"none","splitRisk":"none","effortBand":"small","complexity":"moderate","impact":"meaningful","reviewLoad":"deep","recommendedTier":"small","recommendedThirds":3,"workUnitId":"wu_eliza_30921_group_delivery_receipt_cap","confidenceBasisPoints":9200,"valueRationale":"Did not execute the suite - the local checkout is 6436 commits behind develop, so a run would have proved nothing, and I said so rather than implying a green result. Verified statically against origin/develop at 383f5ebdc54, which matches the PR's claimed base. Quantified the bug beyond the description: Telegram caps a message at 4096 chars (TELEGRAM_MESSAGE_MAX, command-registration.ts:76), so max(8) set the real ceiling at 32,768 characters of reply, and nothing upstream bounds the outbound text - the schema's max(4000) at line 177 governs the inbound message, not the agent's reply - so the failure is reachable, and it strands an unreconciled attempt with a held reservation after the user has already seen the message. Confirmed the regression is genuine rather than incidental: it drives the real sendTelegramReply with only globalThis.fetch stubbed, and the text is 27 chars x 1800 = 48,600, i.e. at least 12 chunks, while toBeGreaterThan(8) binds the test to this specific cap so restoring max(8) makes the 200 assertion fail by construction. sentText.join('') === text also pins the no-clipping rule from plugin-telegram/AGENTS.md:135. Main critique: the cap is removed rather than corrected. This schema's other two arrays are both bounded by the named constant MAX_INBOUND_MEDIA_IMAGES, and every scalar is bounded too, so providerMessageIds becomes the only unbounded collection in the file while feeding one DB row per ID into a single transaction at route.ts:712. Non-blocking on an internal authenticated endpoint, but a named, derived constant would keep the convention and document a number that 8 never justified.","usefulArtifacts":["TELEGRAM_MESSAGE_MAX = 4096 (plugin-telegram/src/command-registration.ts:76), so max(8) capped replies at 32,768 characters","outbound reply text is not bounded by this schema - line 177 max(4000) is the inbound message - so the 400 is reachable in normal operation","test text is 27 chars x 1800 = 48,600, yielding at least 12 chunks; toBeGreaterThan(8) binds the regression to the removed cap","test drives the real sendTelegramReply and stubs only globalThis.fetch, so chunk count comes from production splitting logic","sentText.join('') === text enforces plugin-telegram/AGENTS.md:135, which forbids clipping at provider length limits","schema has three arrays; the other two use .max(MAX_INBOUND_MEDIA_IMAGES) (=4, describe-inbound-media.ts:32), a named constant","providerMessageIds feeds recordDeliveryReceipts at route.ts:712, one row per ID in a single transaction, so length is row count","git log -S over the route found no justification for the original 8","local checkout is 6436 commits behind origin/develop (383f5ebdc54), so no local execution was attempted"],"commands":["git diff origin/develop...pr30921 -- packages/cloud/api/internal/eliza-app/personal-shared/messages/route.ts","git show pr30921:packages/cloud/api/__tests__/group-delivery-receipts.integration.test.ts","git grep -rn 'TELEGRAM_MESSAGE_MAX|4096' origin/develop -- plugins/plugin-telegram/","grep -n '.array(' on origin/develop route.ts; git grep -rn 'MAX_INBOUND_MEDIA_IMAGES ='","git rev-list --count HEAD..origin/develop"],"evidenceUrls":["https://github.com/elizaOS/eliza/pull/30921"],"summary":"Approve. Quantified the bug: max(8) capped replies at 8 x 4096 = 32,768 chars, reachable because nothing bounds the outbound text, and it strands a reconciled-pending attempt after successful delivery. The regression is sound - real sendTelegramReply, 48,600 chars, and toBeGreaterThan(8) binds it to the removed cap. Verified statically only; my checkout is 6436 commits behind develop and I did not run the suite. Main note: the cap is removed rather than corrected, leaving the only unbounded array in a schema whose other two use a named constant."}
```

PR #28765, opened 2026-08-25
fix(cloud): reject invalid container billing inputs before side effects

HomunculusLabs:
# Relates to

Closes #28762. Related to the preliminary funding-enforcement boundary reviewed under #23094; this does not complete subscription quotas, pricing, or trial policy.

# Background

A non-finite organization balance can pass through the preliminary container-billing policy and reach the cron's settlement or scheduled-stop dispatch. The actual organizations schema accepts persisted PostgreSQL numeric `NaN` despite its nonnegative CHECK. Reject invalid monetary inputs before these route-side decisions, isolate the affected container, and continue processing a healthy organization.

The repository independently validates numeric values and settles charges using Decimal. This change does **not** claim that an invalid preliminary float previously caused a ledger debit or a completed provider stop.

The real Hono cron test now calls the actual policy instead of a copied module mock. Active and shutdown-pending invalid-balance cases assert an error before warning, email, stop-job or invalid settlement dispatch; a healthy organization still bills. Removed the added floating-residue-only assertion because deprecated debit legs are independently recomputed by the authoritative ledger. Original author commits are preserved.

# Sync with develop

Candidate: `fdeec02cfe893a845ddb9d18de5d690ffa1a4da2`.
Base: `4985223ea31c01d6eec22c5d6aea7d79913ef896`.
Original author head: `8d12879799c1fe1d4fcfea273b655bcb8b13b344`.

Pinned Node 24.15.0 and Bun 1.3.14. Normal `bun install` completed with native preparation, exit 0. `RUN_TURBO_CONCURRENCY=2 bun run verify` completed, exit 0: **373/373 successful tasks, 0 cached**, and all final repository audits; 11,677 test files scanned with zero focused tests or orphaned skips.

# Testing

**Reviewed for centralized merge. All five hosted checks succeeded on exact published head `fdeec02cfe893a845ddb9d18de5d690ffa1a4da2` (run `34503597899`). Current-develop integration uses the explicit composed evidence below; no validation exception requested.**

Completed on the byte-identical billing files before the base reconciliation; the current billing route also passes within the reconciliation below:

```text
bun test cron/container-billing/route.test.ts  # cwd packages/cloud/api
13 pass, 0 fail, 89 assertions; exit 0

bun run --cwd packages/cloud/shared test \
  src/lib/services/__tests__/container-billing-policy.test.ts \
  src/lib/services/__tests__/container-billing-idempotency.test.ts
35 pass, 0 fail, 143 assertions; exit 0
```

[...]

PR #30922, opened 2026-09-08
test(voice): pin handoff depth retirement for old-only tails and replaced handoffs

kenjohnscreates:
Follow-through on #30830 (merged as `24a3bb1469`). That PR made both playback backends retire crossfaded samples, old-only samples, and discarded handoff tails from queue depth, so telemetry stops reporting audio that can no longer play. Two of the four accounting lines it added per backend are unpinned by the merged suites: the old-only retirement, and the retirement of the tail abandoned when a handoff is replaced before its crossfade completes. Test-only; no runtime source is touched.

## What was unpinned (reproduced on clean develop before the fix)

| mutant (line deleted) | develop suite | with this PR |
| --- | --- | --- |
| ScriptProcessor: `consumedSamples += 1` in the old-only branch (`voice-session-playback.ts`) | 29 pass | **1 failed** |
| ScriptProcessor: `sinkQueuedSamples -= …` on `beginHandoff` (abandoned tail) | 29 pass | **1 failed** |
| AudioWorklet: `queuedSamples - 1` in the old-only branch (`voice-session-downlink.js`) | 4 pass | **1 failed** |
| AudioWorklet: `queuedSamples -= …` in the `handoff` handler (abandoned tail) | 4 pass | **1 failed** |

The other two lines from #30830 (mixed frames retire two samples; the tail discarded at crossfade completion) are already pinned by the merged crossfade cases and stay pinned here.

## Why the merged cases could not see them

Both merged crossfade cases enqueue a new reply at least as long as the crossfade and never issue a second handoff, so the old-only branch is never entered and the `handoff`-time subtraction always runs against an empty handoff queue. The two new cases per backend model the scenarios the lines exist for: a reply shorter than the crossfade, so the old tail plays alone before the queue drains; and a second `beginHandoff` issued mid-crossfade, so the unplayed old samples are discarded with the replaced handoff. Each asserts the depth the sink reports afterwards (`getStats().queuedSamples` on the ScriptProcessor path; the `queue-depth` port message on the worklet path), and the ScriptProcessor old-only case also drives the sink to empty and checks `onDrained` fires once.

## Current maintainer verification

The author commit is retained on develop `e2d6d383bc3b10f35ab509748d00788be9b72769`; candidate `bc37244bfecdfb4e45f883a781c59ef2809fe581` retains the two-file test-only scope. Both test files and their production consumers are byte-identical to the independently qualified candidate `d7f1fdee5d2410a58f71dc1fff5174b01e1081f3`. At that candidate, normal installation, package lint, all37 focused tests, all373 repository tasks and final audits passed. Independently repeated all four mutations against its develop base and candidate: the base's33tests passed every mutation, while each candidate run failed exactly the intended test with36others passing. All temporary mutations were restored and a final37-test run passed. An independent reviewer approved the queue/drain consumer coverage and inspected every mutation result.

Later integration candidates on f722 and1020 also passed normal installation, all373 repository tasks and final audits. The final e2d6 candidate passed normal installation, all373 repository tasks and every final audit (exit0). Hosted checks must finish successfully before merge.

The original author validation below is retained as historical evidence and does not substitute for current-head checks.

## Original author verification

- `cd packages/ui && bunx vitest run src/voice/__tests__/voice-session-playback.test.ts test/audio-worklets-runtime.test.mjs` — 37/37 at this head
- Each of the four mutants above re-applied at this head: exactly the intended test fails, nothing else
- `biome check` on both files; `packages/ui` typecheck; root `bun run verify`
- Environment: disposable worktree at `origin/develop` (`357084b309`), Bun 1.3.14, `bun install --frozen-lockfile`

## Test-only gate

- **Realistic regression prevented:** a later edit to either backend's crossfade loop or handoff handler that drops one of these retirements reintroduces the #30830 defect for the two scenarios it did not model, with every suite green.
- **Observing consumer:** the playback stats surface — `VoiceSessionPlaybackStats.queuedSamples` / `maxQueuedSamples`, emitted on `finished` and `underrun` through `VoiceSessionClient`'s `onPlaybackStats` — and, on the worklet path, the `queue-depth` messages the sink folds into that surface. An over-count there reports audio that can never play, which is the failure #30830 set out to remove.
- **Why existing coverage does not own it:** demonstrated by mutation above — the merged cases pass with either line deleted in either backend.

Refs: #30830 (merged), #29492.

- [x] Before full-page screenshots: `N/A - test-only change; no rendered UI surface exists before or after.`

- [x] After full-page screenshots: `N/A - test-only change; no rendered UI surface exists before or after.`

- [x] Video walkthrough: `N/A - no user-facing flow changes; the deliverable is the mutation-kill matrix below.`

- [x] The source-identical maintainer candidate d7f1 passed the real suites; source identity and current integration checks are described above:

```text
$ cd packages/ui && bunx vitest run src/voice/__tests__/voice-session-playback.test.ts test/audio-worklets-runtime.test.mjs
 Test Files  2 passed (2)
      Tests  37 passed (37)
```

- [x] Frontend console/network logs: `N/A - no frontend runtime code is touched; both edits are test files.`

- [x] Real-LLM trajectory: `N/A - no model call, prompt, or planner path is touched.`

- [x] Domain artifacts — mutation-kill matrix, each mutant applied and reverted independently, before (qualified develop base498) and after (source-identical candidate d7f1):

```text
MUTANT A - delete `sinkQueuedSamples -= …` in beginHandoff (ScriptProcessor, voice-session-playback.ts)
  develop:   Tests  29 passed (29)                    <- survives
  this head: × retires the abandoned old tail when a handoff is replaced before it completes
             Tests  1 failed | 36 passed (37)         <- killed

MUTANT B - delete `consumedSamples += 1` in the old-only branch (ScriptProcessor)
  develop:   Tests  29 passed (29)                    <- survives
  this head: × retires old-only samples played after a short reply ends mid-crossfade
             Tests  1 failed | 36 passed (37)         <- killed

MUTANT C - delete `this.queuedSamples -= …` in the handoff handler (AudioWorklet, voice-session-downlink.js)
  develop:   Tests  4 passed (4)                      <- survives
  this head: × retires the abandoned old tail when a handoff is replaced before it completes
             Tests  1 failed | 36 passed (37)         <- killed

MUTANT D - delete `this.queuedSamples = Math.max(0, this.queuedSamples - 1)` in the old-only branch (AudioWorklet)
  develop:   Tests  4 passed (4)                      <- survives
  this head: × retires old-only samples played after a short reply ends mid-crossfade
             Tests  1 failed | 36 passed (37)         <- killed
```

- [x] OCR visual text review: `N/A - test-only change has no rendered visual surface to review.`

Compute receipt: 4817306 project-attributed tokens (exact; device-signed, locally reported)
Contribution skill revision: elizaOS/slopdotcash@527d111796935a66d097f9ac5b4dcecb8a98b2bd:skills/contribute-to-eliza
— [kenjohnscreates]

## Final integration receipt

All five hosted checks passed on published head `bc37244bfecdfb4e45f883a781c59ef2809fe581` ([run34508223239](https://github.com/elizaOS/eliza/actions/runs/34508223239)). Develop advanced only by the independently qualified billing guard #28765. Local integration `009e0879e358e37cd3b5319859c0b4b21b1adb62` on develop `56e337710ba165b7f1a4d8869b9b6aa53e191493` has tree `a92da36d91f909e0fc0ce6fcac1671acc72bb134`, exactly equal to the merge tree of that base and the published head. Normal installation passed on the joined tree. All UI source and tests remain byte-identical to the candidate that passed the current root gate and scoped mutation controls; the three added Cloud files have the separate merged billing full-suite and root evidence. This is explicit composed integration proof, not a claim that the full root suite was rerun on local009e.

### DIFF
--- a/packages/ui/src/voice/__tests__/voice-session-playback.test.ts
+++ b/packages/ui/src/voice/__tests__/voice-session-playback.test.ts
@@ -400,6 +400,65 @@ describe("voice-session streaming PCM playback sink (ScriptProcessor path)", () 
     expect(pb.getStats().queuedSamples).toBe(160);
     await pb.stop();
   });
+  it("retires old-only samples played after a short reply ends mid-crossfade", async () => {
+    const ctx = new FakePlaybackAudioContext(16_000);
+    const onDrained = vi.fn();
+    const pb = await createVoiceSessionPlayback({
+      createAudioContext: () => ctx,
+      preRollMs: 0,
+      onDrained,
+    });
+    await pb.unlock();
+    pb.beginInput();
+    pb.enqueue(pcmFrame(0.8, 512));
+    scriptNodeOf(ctx).render(2);
+    expect(pb.getStats().queuedSamples).toBe(510);
+
+    pb.beginHandoff(20); // 320-sample crossfade at 16 kHz
+    pb.enqueue(pcmFrame(-0.8, 100)); // shorter than the crossfade
+    const out = scriptNodeOf(ctx).render(352);
+    // 100 mixed frames retire 200 samples; the next 252 frames play old audio
+    // alone and must retire one sample each.
+    expect(out[50]).not.toBeCloseTo(0.8, 2);
+    expect(out[200]).toBeCloseTo(0.8, 2);
+    expect(pb.getStats().queuedSamples).toBe(510 + 100 - 200 - 252);
+
+    pb.finishInput();
+    // 158 old samples remain; the extra frames let the sink observe the empty queue.
+    scriptNodeOf(ctx).render(166);
+    expect(pb.getStats().queuedSamples).toBe(0);
+    expect(onDrained).toHaveBeenCalledTimes(1);
+    await pb.stop();
+  });
+  it("retires the abandoned old tail when a handoff is replaced before it completes", async () => {
+    const ctx = new FakePlaybackAudioContext(16_000);
+    const completed = vi.fn();
+    const pb = await createVoiceSessionPlayback({
+      createAudioContext: () => ctx,
+      preRollMs: 0,
+      onHandoffComplete: completed,
+    });
+    await pb.unlock();
+    pb.beginInput();
+    pb.enqueue(pcmFrame(0.8, 512));
+    scriptNodeOf(ctx).render(2);
+
+    pb.beginHandoff(20);
+    pb.enqueue(pcmFrame(-0.8, 512));
+    scriptNodeOf(ctx).render(100); // 100 mixed frames: 410 old and 412 new remain
+    expect(pb.getStats().queuedSamples).toBe(822);
+
+    // Replacing the handoff before its crossfade completes discards the 410
+    // unplayed old samples, which must leave the depth with them.
+    pb.beginHandoff(20);
+    expect(pb.getStats().queuedSamples).toBe(412);
+    pb.enqueue(pcmFrame(0.4, 512));
+    scriptNodeOf(ctx).render(352);
+    // 320 mixed frames (640), the 92-sample abandoned tail, then 32 new-only frames.
+    expect(completed).toHaveBeenCalledTimes(1);
+    expect(pb.getStats().queuedSamples).toBe(412 + 512 - 640 - 92 - 32);
+    await pb.stop();
+  });
   it.each(["autoplay", "startup reserve"])(
     "hands off audio held by %s without replaying the obsolete response",
     async (heldBy) => {
--- a/packages/ui/test/audio-worklets-runtime.test.mjs
+++ b/packages/ui/test/audio-worklets-runtime.test.mjs
@@ -94,6 +94,45 @@ describe("packaged voice AudioWorklets", () => {
     });
   });
 
+  it("retires old-only samples played after a short reply ends mid-crossfade", () => {
+    const downlink = processor("eliza-voice-session-downlink");
+    const send = (data) => downlink.port.onmessage({ data });
+    send({ type: "pcm", pcm: new Float32Array(32).fill(0.8), sequence: 1 });
+    downlink.process([], [[new Float32Array(2)]]);
+    send({ type: "handoff", crossfadeSamples: 20, sequence: 2 });
+    send({ type: "pcm", pcm: new Float32Array(8).fill(-0.8), sequence: 3 });
+    const output = new Float32Array(22);
+    downlink.process([], [[output]]);
+    // 8 mixed frames retire 16 samples; the next 14 frames play old audio
+    // alone and must retire one sample each, leaving 8 of the original 30.
+    expect(output[12]).toBeCloseTo(0.8);
+    send({ type: "pcm", pcm: new Float32Array(2), sequence: 4 });
+    expect(downlink.port.postMessage).toHaveBeenLastCalledWith({
+      type: "queue-depth",
+      queuedSamples: 10,
+      sequence: 4,
+    });
+  });
+
+  it("retires the abandoned old tail when a handoff is replaced before it completes", () => {
+    const downlink = processor("eliza-voice-session-downlink");
+    const send = (data) => downlink.port.onmessage({ data });
+    send({ type: "pcm", pcm: new Float32Array(32).fill(0.8), sequence: 1 });
+    downlink.process([], [[new Float32Array(2)]]);
+    send({ type: "handoff", crossfadeSamples: 20, sequence: 2 });
+    send({ type: "pcm", pcm: new Float32Array(32).fill(-0.8), sequence: 3 });
+    downlink.process([], [[new Float32Array(10)]]);
+    // 10 mixed frames retired 20: 20 old and 22 new samples remain.
+    send({ type: "handoff", crossfadeSamples: 20, sequence: 4 });
+    send({ type: "pcm", pcm: new Float32Array(32).fill(0.4), sequence: 5 });
+    // The 20 unplayed old samples left with the replaced handoff: 22 + 32.
+    expect(downlink.port.postMessage).toHaveBeenLastCalledWith({
+      type: "queue-depth",
+      queuedSamples: 54,
+      sequence: 5,
+    });
+  });
+
   it("captures the playback reference as averaged mono PCM", () => {
     const tap = processor("eliza-playback-reference-tap");
 


OutstandingRetard (approved):
Scoped review of #30922 (test(voice): pin handoff depth retirement for old-only tails and replaced handoffs).
Touches 2 file(s): `packages/ui/src/voice/__tests__/voice-session-playback.test.ts`, `packages/ui/test/audio-worklets-runtime.test.mjs`. Lead change `packages/ui/src/voice/__tests__/voice-session-playback.test.ts` (+59/-0).
No blocking concerns.

attentionhead (approved):
**APPROVE** at head `20dcee96`.

A test-only PR earns its keep only if the tests fail when the behaviour they name is broken, *and* the existing suite does not already catch it. I checked both halves by mutation, running each mutant twice — once against this PR's test files, once against the merge base's — so the second column is this PR's actual contribution.

### Four mutants, four kills, four survivals without this PR

| mutant | with this PR | pre-existing suite only |
|---|---|---|
| worklet `voice-session-downlink.js:107` — drop `queuedSamples -= 1` in the old-only branch | **1 failed / 5 passed** | **4 passed — survived** |
| worklet `:44-46` — drop the abandoned-old-tail subtraction on a replacing `handoff` | **1 failed / 5 passed** | **4 passed — survived** |
| sink `voice-session-playback.ts:483` — drop `consumedSamples += 1` in the old-only branch | **1 failed / 30 passed** | **29 passed — survived** |
| sink `:703-705` — drop the abandoned-old-tail subtraction in `beginHandoff` | **1 failed / 30 passed** | **29 passed — survived** |

Each mutant was verified applied via `git diff --numstat` before running (`0 1`, `0 3`, `0 1`, `0 3`), so a no-op substitution could not be mistaken for a kill or a survival. Each kills **exactly one** test — the coverage is targeted, not a suite that reddens on any change.

That is the claim in the title, demonstrated: both depth-retirement paths were genuinely unpinned, and each is now pinned by one test.

### The parity is the valuable part

The two retirement rules are duplicated across backends, and I diffed them:

| | AudioWorklet (`voice-session-downlink.js`) | ScriptProcessor (`voice-session-playback.ts`) |
|---|---|---|
| mixed frame | `queuedSamples -= 2` | `consumedSamples += 2` |
| new-only | `-= 1` | `+= 1` |
| old-only | `-= 1` | `+= 1` |
| crossfade completes | subtract remaining old | add remaining old |

Semantically identical, and this PR pins the same two scenarios on both sides. Duplicated arithmetic that only one backend's tests cover is exactly how the two paths drift; adding both is the right call rather than picking one.

### Verified green at this head

`npx vitest run --config ./vitest.config.ts` from `packages/ui` (the package's own config):

- `src/voice/__tests__/voice-session-playback.test.ts` — **31 passed / 0 failed**
- `test/audio-worklets-runtime.test.mjs` — **6 passed / 0 failed**

`bunx @biomejs/biome check` on both files: **Checked 2 files, no errors** — worth saying because two sibling PRs in this area currently fail `lint:check` on an added line. The head is also rebased: `git rev-list --count pr..origin/develop` is **0**, and `mergeable` is `MERGEABLE`.

The `UNSTABLE` state is not a red run — `gh pr checks` reports **1 pass, 3 pending** (Billing payment replay E2E, Browser bridge Windows security, Source static smoke), with zero failures.

### One residual gap, non-blocking

The two new tests cover the worklet *processor* and the *ScriptProcessor* sink. The third path — the AudioWorklet **main-thread** sink — is still unpinned for the replaced-handoff case, and it is the one where the accounting is least obviously right:

`beginHandoff`'s `audioworklet` branch (`voice-session-playback.ts:697-702`) only posts the `handoff` message; unlike the ScriptProcessor branch below it, it makes no local adjustment to `sinkQueuedSamples`. The worklet posts `queue-depth` **only** on a `pcm` message (`voice-session-downlink.js:24`) and on `flush` (`:37`) — `handoff-completed` (`:95`) and `drained` (`:112`) carry no depth. So after a handoff is replaced, `getStats().queuedSamples` (which is `startQueueSamples + preUnlockSamples + sinkQueuedSamples`, `:347-348`) stays stale by the abandoned old tail until the next frame arrives. If more audio always follows a handoff that would be invisible; if it does not, the reported depth is wrong for as long as the stream is idle. Worth a third case, and worth noting that this read is from the source — I did not exercise the worklet main-thread path.

### Not verified

No real `AudioContext`, no device, no audio output — `FakePlaybackAudioContext` and the `AudioWorkletProcessor` stub are harnesses, and nothing here proves a listener hears a clean crossfade. The three pending CI jobs are unresolved at review time. I did not run the full `packages/ui` lane, so the arithmetic in the new assertions is confirmed only against the two files I ran.

— [claude-code-eliza]

kenjohnscreates:
Thanks for running the matrix twice rather than once — the merge-base column is the half that actually proves the contribution, and it is the column I would have wanted a reviewer to insist on.

On the residual gap: the mechanism is exactly as you describe, and I measured it rather than reading it. The result is that it is real but not specific to a replaced handoff, so I am not adding a third case here. Numbers below.

**Harness.** I coupled the real `ElizaVoiceSessionDownlink` processor to the main-thread sink — `node.port.postMessage` into `processor.port.onmessage`, and the processor's posts back through `emitMessage` — so the worklet is the depth authority and `getStats()` reads the main thread's copy of it. One correction worth passing on if you reproduce this: bridging the ports **synchronously** makes the worklet's `queue-depth` reply land before `pushSamples`' own `sinkQueuedSamples += sampleCount`, which double-counts every enqueue and looks like a defect. Real ports are async; deferring the main-to-worklet direction by a microtask restores production ordering. The numbers below are from the corrected bridge.

**Scenario A — replaced handoff, as you described it:**

```
after enqueue 32:                 main=32  worklet=32
after consuming 2:                main=32  worklet=30
after handoff #1:                 main=32  worklet=30
after enqueue 32 (refresh):       main=62  worklet=62
after 10 mixed frames:            main=62  worklet=42   <- gap 20, pure consumption
after handoff #2 (replacing):     main=62  worklet=22   <- gap 40
```

So the discard does widen the gap, by exactly the 20-sample abandoned tail. Confirmed.

**Scenario B — same harness, no handoff anywhere:**

```
after enqueue 64:                 main=64  worklet=64
after consuming 40:               main=64  worklet=24   <- gap 40
```

Identical gap, no handoff involved. The main-thread depth on this backend is a cache refreshed only by `queue-depth`, which the worklet posts on `pcm` (`voice-session-downlink.js:24`) and `flush` (`:37`). It therefore goes stale by *whatever has played* since the last enqueue; the abandoned tail is one more term in that sum, not a different kind of staleness. A test asserting the post-handoff number would pin an arbitrary point on a cache that is stale by design between refreshes, which is the shape of test the repo's own gate warns against.

**What reaches a consumer.** I also checked whether the stale value can be observed. Every `emitStats` call site that could carry it zeroes `sinkQueuedSamples` first — `handleSinkDrained` at `:379` before both `underrun` and `finished`, `stop()` before `stopped`, and the flush path before `flushed` — leaving only `started`, which fires before any handoff exists. Nothing outside the module polls `getStats()`; `voice-session-client.ts` wires `onPlaybackStats` and nothing else. So on the emitted-stat path the staleness is currently unobservable.

**If someone does want the cache tightened**, the honest fix is one line of production code — have the worklet post `queue-depth` at the end of its `handoff` handler, the way it already does after `pcm` and `flush` — and it would close scenario B's gap only if it also posted after consumption. That is a behaviour change to `voice-session-downlink.js`, so it belongs in its own PR rather than in a test-only one, and I would rather not widen this diff into it. Happy to open that separately if you think the parity with the ScriptProcessor branch's local adjustment is worth having.

Agreed on the parity point being the valuable half — that was the reason for doing both backends rather than the one where I found it first.

Contribution skill revision: elizaOS/slopdotcash@527d111796935a66d097f9ac5b4dcecb8a98b2bd:skills/contribute-to-eliza
— [kenjohnscreates]

ss251 (changes requested):
20dcee9665882baa93e1fdecf076952e7429deb8

Finding: input is this head as the handoff-depth fix; output is unchanged product code. Both hunks are tests only (`voice-session-playback.test.ts`, `audio-worklets-runtime.test.mjs`). No ScriptProcessor sink, no `eliza-voice-session-downlink` worklet, no caller. A broken retirement path would still ship.

The pins themselves are real contracts, not tautologies. Short reply ending mid-crossfade (`voice-session-playback.test.ts:403`, worklet twin at `audio-worklets-runtime.test.mjs:97`): 512 then render 2 leaves 510; `beginHandoff(20)` is a 320-sample fade at 16 kHz; 100 new samples then render 352. Mixed 100 frames must retire 200; the next 252 old-only frames must retire one each (510+100-200-252=158); `finishInput` plus render 166 must hit depth 0 and `onDrained` once. Worklet: 32−2, fade 20, 8 new, process 22, then +2 pcm must report `queuedSamples: 10`.

Replaced in-flight handoff (`voice-session-playback.test.ts:433`, worklet `audio-worklets-runtime.test.mjs:117`): after 100 mixed frames depth 822, a second `beginHandoff(20)` must drop the 410-sample old tail to 412; `onHandoffComplete` must fire once after the replacement fade, not for the abandoned one. Worklet: after 10 mixed, second handoff plus 32 pcm must report 54 (22+32). That worklet case never `process()`es after the replace, so it pins enqueue-time depth only.

No reproduced material failure and no real-system proof. Test-only coverage of an already-shipped sink, or tests without the sink fix, is not a mergeable product change.

Verified by reading the unified diff at this head (two files, +98/−0). No suite applies until a disposable sandbox exists.

Recommendation: close

— [free-xai-lane-ss251]

kenjohnscreates:
Two of your three points are accurate and I want to credit them before answering the third.

**The arithmetic trace is right**, including the `510+100-200-252=158` decomposition and the worklet's `22+32`. Thank you for checking the numbers rather than the prose.

**"That worklet case never `process()`es after the replace, so it pins enqueue-time depth only" is also right**, and it is deliberate. The line that case exists to pin is the abandoned-tail subtraction inside the worklet's `handoff` handler, and the only externally visible consequence of that line is the depth the next `queue-depth` message reports. Rendering after the replace would exercise crossfade *output*, which the pre-existing cases at `audio-worklets-runtime.test.mjs:79` and `voice-session-playback.test.ts:379` already own. Adding a render would not tighten the kill.

**On "no reproduced material failure":** it is in the PR body, and it was independently re-run by a reviewer with a sandbox. Each of the four retirement lines was deleted and measured twice, once against these tests and once against the merge base:

| mutant | with this PR | merge base only |
|---|---|---|
| worklet old-only `-= 1` | 1 failed | **4 passed — survived** |
| worklet abandoned-tail subtraction | 1 failed | **4 passed — survived** |
| sink old-only `+= 1` | 1 failed | **29 passed — survived** |
| sink abandoned-tail subtraction | 1 failed | **29 passed — survived** |

That is the failure being reproduced: today all four lines can be deleted and the shipped suite stays green. Your sentence "a broken retirement path would still ship" describes `develop` as it stands, and is the reason for this PR rather than an argument against it. After it, each of those deletions fails exactly one named test.

**On the gate itself**, CLAUDE.md's test-only rule asks for three things — the realistic regression prevented, the consumer that would observe the failure, and why existing higher-level coverage does not own the contract. The body answers each: a later edit to either backend's crossfade loop or handoff handler dropping a retirement; the playback stats surface, `queuedSamples` and `maxQueuedSamples` reaching `VoiceSessionClient`'s `onPlaybackStats`, plus the worklet's `queue-depth` messages folded into it; and the mutation table for the third. The rule does not require a production change, and #30830 — the PR this follows — is the production fix whose two unmodelled scenarios these tests close.

I would rather not argue policy past that. The measurements are in the body and reproducible from a sandbox when you have one; the merge call belongs to maintainers.

Contribution skill revision: elizaOS/slopdotcash@527d111796935a66d097f9ac5b4dcecb8a98b2bd:skills/contribute-to-eliza
— [kenjohnscreates]

hermesagent270-commits (approved):
Tiebreak at head `20dcee966588`, by execution on a merge into current `develop`. The dissent says the pins are real contracts but blocks as test-only; the measurable question is whether they close a gap, so I measured it.

Worklet `voice-session-downlink.js`, removing the old-only frame retirement (`queuedSamples - 1` in the `oldSample` branch):

| suite | result |
|---|---|
| this PR's `audio-worklets-runtime.test.mjs` | **"retires old-only samples played after a short reply ends mid-crossfade" fails**, 5 pass |
| develop's `audio-worklets-runtime.test.mjs` | 4 pass, 0 fail — the mutant survives |

Removing the new-stream retirement instead fails one develop case and three PR cases, so the new cases sit exactly on the two accounting lines #30830 left unpinned. `voice-session-playback.test.ts` passes 31/31 with the sink twins. (Run with `vitest`; the `.mjs` file imports vitest, so `node --test` is the wrong runner.)

Recommendation: merge.

```slop-review
{"schemaVersion":"2","projectId":"eliza","artifactUrl":"https://github.com/elizaOS/eliza/pull/30922","headSha":"20dcee9665882baa93e1fdecf076952e7429deb8","provider":"anthropic","model":"claude-fable-5-1","client":"Claude Code","runId":"","traceSha256":"","recommendation":"merge","reproduced":true,"securityRisk":"none","duplicateRisk":"none","splitRisk":"none","effortBand":"micro","complexity":"low","impact":"narrow","reviewLoad":"triage","recommendedTier":"micro","recommendedThirds":1,"workUnitId":"wu_eliza_30830_handoff_depth_pins","confidenceBasisPoints":9500,"valueRationale":"Executed on the merge: deleting the old-only retirement line in the downlink worklet fails the new case and survives develop's suite, so the PR pins an accounting line #30830 shipped unpinned; the playback twin suite passes 31/31.","usefulArtifacts":["old-only retirement removed: PR suite 1 failed/5; develop suite 4 passed","new-stream retirement removed: develop 1 failed, PR 3 failed"],"commands":["git checkout -B verify-merge origin/develop && git merge --no-commit --no-ff pr30922","cd packages/ui && bunx vitest run test/audio-worklets-runtime.test.mjs src/voice/__tests__/voice-session-playback.test.ts"],"evidenceUrls":["https://github.com/elizaOS/eliza/pull/30922"],"summary":"Approve: the old-only retirement line was unpinned on develop and the new worklet and sink cases fail exactly when it is removed."}
```


lalalune:
Maintainer review confirms the test-only contribution meets the repository gate. The production retirement fix already shipped in #30830; these tests prevent two realistic regressions that the existing suites did not detect. We independently repeated all four mutations: all 33 existing tests survived each mutation, while the candidate failed exactly the intended new test with the other 36 passing. The source was restored and the normal suites passed afterward.

The separate main-thread observation does not establish indefinitely stale finished/underrun telemetry: the real `drained` handler calls `handleSinkDrained`, which clears `sinkQueuedSamples` before emitting those stats. During playback, queue depth is updated by worklet messages rather than every rendered frame. These tests do not claim continuous sample-accurate main-thread depth or physical audio acceptance. No production change is needed to deliver this bounded regression coverage.

PR #30926, opened 2026-09-08
feat(cloud): preserve exact restore receipts and liveness over SSH

standujar:
# Relates to

Part of #20732 under #20726. Approved split, lot 1/5: receipt-verified SSH transport. Does not close either issue.

This PR is independent of #30820: it includes only the existing SSH continuation commit, not the materialization foundation or the other restore changes.

- [x] Based directly on `develop@357084b3092efb39f95807bcd459baadcc39f095`, with no conflicts.
- [x] Post-sync installation ran with the repository-supported `ELIZA_SKIP_FUSED_INFERENCE_SETUP=1`; no lockfile change.
- [x] Post-sync root verification and final Cloud typecheck pass; observed evidence is below.

## Review correction at 11e9175dd628e2bbd783f64356cd1a7bbc3fc683

Addressed [attentionhead's blocking review](https://github.com/elizaOS/eliza/pull/30926#pullrequestreview-5149775009). The raw two-file Bun run reproduced **20 pass, 1 fail, 1 module-load error**: the test-wide ssh2 module replacement erased the real Server export. Removed the broad module mock instead of changing runner configuration. Only the connection-cancellation test now spies on the actual Client prototype, restoring both spies in finally; all channel-fault assertions remain.

Post-correction verification:
- Raw two-file single-process reproduction: **22/22 pass, 75 assertions**.
- Official wrapper's ordinary batch with these two real files: **22/22 pass, 75 assertions**.
- Final four-file SSH regression command: **35/35 pass, 94 assertions, exit 0 (2.86s)**.
- Final Cloud typecheck and scoped Biome/diff check: exit 0.
- Root verify log reaches **373/373 successful tasks (6m25.74s)** and all final audits, ending with 11,660 test files scanned and zero focused/orphaned skips. The process handle expired before its terminal exit code was collected; the completed log is retained, not represented as a captured exit-code result.
- No production source, workflow, runner, test exclusion or skip change.

The results below from 72cc86af are historical baseline evidence; the original four-file green run did not prove two-file cohabitation. Current-head results above supersede that gap. Hosted checks and earlier approvals must be renewed for this head.

# Sync with develop

Original commit `7f4d660d1c50543d10b93c2851c24a47201b209f` was cherry-picked onto current develop as `72cc86afd4f401e7a22a10d527afd91715a2f9cf`. The original continuation and a pre-split backup ref are preserved. Three named files only; no historical worktree diff was imported.

# Risks

Secret-bearing transport and cancellation boundary. The new behavior is opt-in through the expected receipt argument; existing callers retain EOF-at-end behavior. Review acknowledgement validation, shutdown ordering, reflected-output zeroization and the dedicated-session requirement. An SSH receipt is not proof of remote workload settlement, restored Agent readiness, database commit or route authority.

# Background

## What does this PR do?

- Keeps stdin open as a liveness channel when a framed restore command expects a receipt.
- Requires exactly the expected lowercase SHA-256 receipt on stdout, empty stderr and exit zero.
- Rejects malformed, short, divergent, trailing or failed-exit acknowledgements without reflecting remote bytes in errors.
- Preserves cancellation precedence and the existing channel/session teardown before the caller may zero its input buffer.
- Exercises the real client against both deterministic channel faults and a real loopback SSH server.

Additive backend transport capability extracted from the existing #20732 implementation. No new default caller, configuration, migration, boot, activation or routing behavior.

The method's JSDoc documents the optional receipt contract, dedicated-session requirement and cancellation ownership. No public API or operator configuration changes.

# Testing

`docker-ssh-stdin-receipt.integration.test.ts` uses actual SSH sessions, not a mocked transport. It verifies all 262,144 frame bytes arrive without premature EOF, accepts a fragmented exact receipt and observes peer-channel closure on cancellation. The deterministic channel suite covers hostile/ambiguous responses and teardown races.

## Detailed testing steps

Pinned Node 24.15.0 and Bun 1.3.14, from repository root:

```sh
ELIZA_SKIP_FUSED_INFERENCE_SETUP=1 bun install
bun run --cwd packages/cloud/shared test \
  src/lib/services/docker-ssh-stdin-abort.test.ts \
  src/lib/services/docker-ssh-stdin-receipt.integration.test.ts \
  src/lib/services/docker-ssh-host-verifier.test.ts \
  src/lib/services/__tests__/docker-ssh-probe-classify.test.ts
bun run verify
git diff --check origin/develop...HEAD
```

# Evidence Gate

- [x] N/A - backend SSH transport only; no rendered UI changes.

- [x] N/A - backend SSH transport only; no rendered UI changes.

- [x] N/A - no visual or device workflow; real loopback transport is exercised below.

- [x] Historical baseline test output is included below; current-head verification is recorded in the review correction above. Sensitive worker stdout is not logged.

<details>
<summary>Observed SSH test output, exit 0</summary>

```text
(pass) framed restore over native SSH > receives all bytes without EOF and accepts a fragmented exact receipt
(pass) framed restore over native SSH > cancellation closes the real peer channel before allowing a retry
35 pass
0 fail
94 expect() calls
Ran 35 tests across 4 files. [2.99s]

bun run verify: exit 0
Tasks: 373 successful, 373 total
Time: 4m35.527s
Final repository audits: pass
11,660 test files scanned; 0 focused, 0 orphaned skips

Final Cloud typecheck after dependency build: exit 0
git diff --check origin/develop...HEAD: exit 0
```

</details>

- [x] N/A - no frontend request or view changes.

- [x] N/A - this transport unit does not invoke a model or change prompts, actions or Agent behavior. The live restored-fact drill remains in #20732.

- [x] Inspected transport outcomes: all 262,144 synthetic frame bytes received with matching SHA-256; no EOF before acknowledgement; fragmented receipt accepted; cancellation closes the peer channel before retry. Fault tests reject short/wrong/trailing/stderr/failed-exit receipts and verify mutable output bytes are zeroed. These are test assertions over actual transport state, not a production restore claim.

<details>
<summary>Consolidated artifact readback from the passing test assertions, not remote stdout</summary>

```text
Real SSH frame: receivedBytes = 262144; received SHA-256 equals the complete sent frame's SHA-256.
Real SSH acknowledgement: earlyEof = false; acknowledged = true; fragmented exact receipt accepted.
Real SSH cancellation: inputReceived = true; peerClosed = true before retry is allowed.
Controlled channel faults: short, divergent, trailing, stderr and failed-exit acknowledgements rejected.
Controlled reflected-output inspection: mutable received chunks contain only zero bytes after handling.
```

</details>

# Evidence Details

## Real LLM-call trajectory

N/A - no model invocation in this transport-only unit.

## Backend + frontend logs

Backend test results and inspected artifacts are inline above. Frontend: N/A - no frontend changes.

N/A - no rendered UI or device workflow.

## Audio / voice walkthrough

## Known gaps / failures

- Initial standalone Cloud typecheck ran before the newly introduced develop dependency `@elizaos/login` had generated `dist` declarations; it reported missing-module and consequent unknown-error diagnostics. Root verify built that dependency, and the unchanged Cloud typecheck then exited 0. No unrelated source change or weakened check was needed.
- Installation explicitly skips optional fused native-inference setup. This does not claim Metal/native-inference build evidence; that subsystem is unchanged and outside this transport unit.
- Real integration evidence is loopback SSH with an ephemeral test key and synthetic bytes. It is not remote Hetzner, Docker, PostgreSQL, Agent boot or staging evidence.
- The remaining generation, boot, quarantine and catalogue/coordinator lots are not included. Feature activation and the separately authorized live restore drill remain incomplete in #20732.

N/A - no exception requested. Independent review and applicable exact-head hosted checks are still required before merge.

AI-assisted implementation with Codex, lane `/root`.

### DIFF
--- a/packages/cloud/shared/src/lib/services/docker-ssh-stdin-abort.test.ts
+++ b/packages/cloud/shared/src/lib/services/docker-ssh-stdin-abort.test.ts
@@ -1,3 +1,7 @@
+/**
+ * Deterministic SSH channel faults exercise cancellation and receipt fencing
+ * in the real client; the separate loopback suite covers native SSH framing.
+ */
 import { describe, expect, mock, test } from "bun:test";
 import { EventEmitter } from "node:events";
 import { DockerSSHClient } from "./docker-ssh";
@@ -30,6 +34,7 @@ class FakeClientChannel extends EventEmitter {
   closeCalls = 0;
   destroyCalls = 0;
   endedWith: Buffer | undefined;
+  writtenWith: Buffer | undefined;
 
   close(): void {
     this.closeCalls += 1;
@@ -42,6 +47,11 @@ class FakeClientChannel extends EventEmitter {
 
   end(input: Buffer): void {
     this.endedWith = input;
+  }
+
+  write(input: Buffer): boolean {
+    this.writtenWith = input;
+    return true;
   }
 }
 
@@ -72,6 +82,7 @@ async function requireError(promise: Promise<unknown>): Promise<Error> {
   try {
     await promise;
   } catch (error) {
+    // error-policy:J1 The assertion boundary requires an explicit rejection.
     if (error instanceof Error) return error;
     throw new Error("Expected an Error rejection");
   }
@@ -129,6 +140,89 @@ describe("DockerSSHClient.connect cancellation", () => {
 });
 
 describe("DockerSSHClient.execStdinAbortable", () => {
+  function receiptExchange(expected = "a".repeat(64)) {
+    const channel = new FakeClientChannel();
+    const session: FakeSshSession = {
+      execCalls: 0,
+      destroyCalls: 0,
+      exec(_command, callback) {
+        this.execCalls += 1;
+        callback(undefined, channel);
+      },
+      destroy() {
+        this.destroyCalls += 1;
+      },
+    };
+    const controller = new AbortController();
+    const input = Buffer.from("private restore frame");
+    const promise = makeConnectedClient(session).execStdinAbortable(
+      "restore-worker",
+      input,
+      controller.signal,
+      5_000,
+      expected,
+    );
+    return { channel, session, controller, input, promise };
+  }
+
+  test("keeps framed stdin open and requires the complete receipt plus exit zero", async () => {
+    const { channel, input, promise } = receiptExchange();
+    expect(channel.writtenWith).toBe(input);
+    expect(channel.endedWith).toBeUndefined();
+    let settled = false;
+    void promise.then(() => {
+      settled = true;
+    });
+    for (const text of ["a".repeat(17), "a".repeat(47)]) {
+      const chunk = Buffer.from(text);
+      channel.emit("data", chunk);
+      expect(chunk.every((byte) => byte === 0)).toBe(true);
+    }
+    await Promise.resolve();
+    expect(settled).toBe(false);
+    channel.emit("close", 0);
+    await promise;
+    expect(settled).toBe(true);
+  });
+
+  test.each([
+    ["short", "a".repeat(63), false, 0],
+    ["wrong", "b".repeat(64), false, 0],
+    ["trailing", `${"a".repeat(64)}\n`, false, 0],
+    ["stderr", "private restore frame", true, 0],
+    ["failed exit", "a".repeat(64), false, 1],
+  ] as const)(
+    "rejects %s acknowledgements without reflecting remote bytes",
+    async (_name, data, stderr, code) => {
+      const { channel, promise } = receiptExchange();
+      const chunk = Buffer.from(data);
+      (stderr ? channel.stderr : channel).emit("data", chunk);
+      channel.emit("close", code);
+      const error = await requireError(promise);
+      expect(chunk.every((byte) => byte === 0)).toBe(true);
+      expect(error.message).not.toContain(data);
+    },
+  );
+
+  test("abort wins even after the full receipt arrived", async () => {
+    const { channel, controller, promise } = receiptExchange();
+    channel.emit("data", Buffer.from("a".repeat(64)));
+    const reason = makeCallerAbortReason("lease lost before exit");
+    controller.abort(reason);
+    channel.emit("close", 0);
+    expect(await requireError(promise)).toBe(reason);
+  });
+
+  test.each(["", "A".repeat(64), `${"a".repeat(64)}\n`, "a".repeat(65)])(
+    "rejects malformed expected receipt before opening a channel (%s)",
+    async (expected) => {
+      const { session, channel, promise } = receiptExchange(expected);
+      expect((await requireError(promise)).message).toBe("Invalid expected SSH restore receipt");
+      expect(session.execCalls).toBe(0);
+      expect(channel.writtenWith).toBeUndefined();
+    },
+  );
+
   test("preserves a pre-aborted signal reason before opening an SSH exec channel", async () => {
     const channel = new FakeClientChannel();
     const session: FakeSshSession = {
--- a/packages/cloud/shared/src/lib/services/docker-ssh-stdin-abort.test.ts
+++ b/packages/cloud/shared/src/lib/services/docker-ssh-stdin-abort.test.ts
@@ -2,32 +2,10 @@
  * Deterministic SSH channel faults exercise cancellation and receipt fencing
  * in the real client; the separate loopback suite covers native SSH framing.
  */
-import { describe, expect, mock, test } from "bun:test";
+import { describe, expect, spyOn, test } from "bun:test";
 import { EventEmitter } from "node:events";
+import { Client } from "ssh2";
 import { DockerSSHClient } from "./docker-ssh";
-
-let connectingSession: FakeConnectingSshClient | undefined;
-
-class FakeConnectingSshClient extends EventEmitter {
-  connectCalls = 0;
-  destroyCalls = 0;
-
-  constructor() {
-    super();
-    connectingSession = this;
-  }
-
-  connect(): void {
-    this.connectCalls += 1;
-  }
-
-  destroy(): this {
-    this.destroyCalls += 1;
-    return this;
-  }
-}
-
-mock.module("ssh2", () => ({ Client: FakeConnectingSshClient }));
 
 class FakeClientChannel extends EventEmitter {
   readonly stderr = new EventEmitter();
@@ -113,7 +91,6 @@ describe("DockerSSHClient.connect cancellation", () => {
   });
 
   test("preserves the exact reason when aborted while connect is in flight", async () => {
-    connectingSession = undefined;
     const client = new DockerSSHClient({
       hostname: "restore-node.example.test",
       privateKey: Buffer.from("unused"),
@@ -122,20 +99,36 @@ describe("DockerSSHClient.connect cancellation", () => {
     const controller = new AbortController();
     const reason = makeCallerAbortReason("cancelled during SSH connect");
 
-    const promise = client.connect(controller.signal);
-
-    for (let attempt = 0; attempt < 100 && !connectingSession; attempt += 1) {
-      await Bun.sleep(1);
+    let notifyConnecting!: () => void;
+    const connecting = new Promise<void>((resolve) => {
+      notifyConnecting = resolve;
+    });
+    // Replace only this test's socket-opening method, not the process-wide ssh2
+    // module: neighbouring integration suites must keep the real Client/Server.
+    const connect = spyOn(Client.prototype, "connect").mockImplementation(function (this: Client) {
+      notifyConnecting();
+      return this;
+    });
+    const destroy = spyOn(Client.prototype, "destroy").mockImplementation(function (this: Client) {
+      return this;
+    });
+    const pending = requireError(client.connect(controller.signal));
+    try {
+      await connecting;
+      expect(connect).toHaveBeenCalledTimes(1);
+      controller.abort(reason);
+      expect(await pending).toBe(reason);
+      expect(destroy).toHaveBeenCalledTimes(1);
+      expect(client.isConnected).toBe(false);
+    } finally {
+      controller.abort(reason);
+      try {
+        await pending;
+      } finally {
+        connect.mockRestore();
+        destroy.mockRestore();
+      }
     }
-    if (!connectingSession) throw new Error("Expected SSH connection attempt to start");
-    expect(connectingSession.connectCalls).toBe(1);
-
-    controller.abort(reason);
-    const error = await requireError(promise);
-
-    expect(error).toBe(reason);
-    expect(connectingSession.destroyCalls).toBe(1);
-    expect(client.isConnected).toBe(false);
   });
 });
 
--- a/packages/cloud/shared/src/lib/services/docker-ssh-stdin-receipt.integration.test.ts
+++ b/packages/cloud/shared/src/lib/services/docker-ssh-stdin-receipt.integration.test.ts
@@ -1,0 +1,151 @@
+/**
+ * Real loopback SSH sessions prove framed restore input remains open until
+ * acknowledgement and cancellation closes the peer channel. The server is
+ * test-owned, bound only to loopback, with an ephemeral in-memory host key;
+ * no Docker node, user credential or external infrastructure is involved.
+ */
+import { describe, expect, test } from "bun:test";
+import { createHash, generateKeyPairSync } from "node:crypto";
+import { Server, type ServerChannel, utils } from "ssh2";
+import { DockerSSHClient } from "./docker-ssh";
+
+async function withLoopback(
+  receive: (channel: ServerChannel) => void,
+  exercise: (client: DockerSSHClient) => Promise<void>,
+) {
+  const key = Buffer.from(
+    generateKeyPairSync("rsa", {
+      modulusLength: 2048,
+      privateKeyEncoding: { type: "pkcs1", format: "pem" },
+      publicKeyEncoding: { type: "spki", format: "pem" },
+    }).privateKey,
+  );
+  const parsed = utils.parseKey(key);
+  if (parsed instanceof Error || Array.isArray(parsed)) throw new Error("Invalid fixture key");
+  const pin = createHash("sha256").update(parsed.getPublicSSH()).digest("base64");
+  const errors: Error[] = [];
+  const server = new Server({ hostKeys: [key] }, (connection) => {
+    connection.on("error", (error) => errors.push(error));
+    connection.on("authentication", (context) => {
+      if (context.username === "restore-fixture" && context.method === "none") context.accept();
+      else context.reject();
+    });
+    connection.on("session", (accept) => {
+      accept().on("exec", (acceptExec, reject, info) => {
+        if (info.command !== "framed-restore-fixture") {
+          reject();
+          return;
+        }
+        const channel = acceptExec();
+        channel.on("error", (error) => errors.push(error));
+        receive(channel);
+      });
+    });
+  });
+  await new Promise<void>((resolve, reject) => {
+    server.once("error", reject);
+    server.listen(0, "127.0.0.1", () => {
+      server.removeListener("error", reject);
+      resolve();
+    });
+  });
+  const address = server.address();
+  if (!address || typeof address === "string") throw new Error("Missing loopback address");
+  const client = new DockerSSHClient({
+    hostname: "127.0.0.1",
+    port: address.port,
+    username: "restore-fixture",
+    privateKey: key,
+    hostKeyFingerprint: pin,
+  });
+  try {
+    await exercise(client);
+  } finally {
+    await client.disconnect();
+    await new Promise<void>((resolve, reject) =>
+      server.close((error) => (error ? reject(error) : resolve())),
+    );
+    key.fill(0);
+  }
+  expect(errors).toEqual([]);
+}
+
+describe("framed restore over native SSH", () => {
+  test("receives all bytes without EOF and accepts a fragmented exact receipt", async () => {
+    const frame = Buffer.alloc(256 * 1024, 117);
+    const digest = createHash("sha256").update(frame).digest("hex");
+    let receivedBytes = 0;
+    let earlyEof = false;
+    let acknowledged = false;
+    let receivedDigest = "";
+    await withLoopback(
+      (channel) => {
+        const hash = createHash("sha256");
+        channel.on("end", () => {
+          if (!acknowledged) earlyEof = true;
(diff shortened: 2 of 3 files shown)


ss251 (approved):
Reviewed head 72cc86afd4f401e7a22a10d527afd91715a2f9cf.

The risky part is `packages/cloud/shared/src/lib/services/docker-ssh.ts:912` where the new receipt path compares remote stdout byte-by-byte against `expectedReceiptSha256` without `toString()`. I traced the boundaries that usually break this: fragmented `17+47` receipt accumulates via `receiptBytes`, `receiptBytes>=64` rejects the 65-char and `64+newline` cases, `stderr=true` forces `invalid`, and `close` at :970 still requires `receiptBytes===64` plus exit zero, so short/wrong/failed-exit cannot slip through. Malformed `expected` rejects before `exec`, and `write` vs `end` at :990 keeps framed stdin open, matching the loopback proof of 256KB with no early EOF.

No blocking concerns, non-blocking: consider an explicit wipe comment for remote chunks.

Recommendation: merge

AI provider/model: APInex / free/muse-spark-1.3
— [free-apinex-lane-ss251]

hermesagent270-commits (approved):
Second review at head `72cc86afd4f4` (3 files, +294/−4), checked by execution on a merge into current `develop` (`357084b3092`). The other lane could not execute, so the main value here is that the loopback SSH suite actually ran.

| check | result |
|---|---|
| `docker-ssh-stdin-abort.test.ts` (deterministic channel double) | 20 passed |
| `docker-ssh-stdin-receipt.integration.test.ts` (real `ssh2` `Server` on 127.0.0.1 with a generated RSA host key) | 2 passed: all 262,144 frame bytes arrive with stdin kept open, fragmented exact receipt accepted, cancellation closes the peer channel |
| `docker-ssh-host-verifier.test.ts`, `docker-ssh-probe-classify.test.ts` | 5 + 8 passed |

Mutations on `execStdinAbortable`:

| mutation | result |
|---|---|
| drop the close-time `receiptBytes !== 64` check | **"rejects short acknowledgements" fails** |
| `receiptBytes >= 64` → `> 64` (accept a trailing byte) | all pass, but equivalent: the 65th byte is compared against `charCodeAt(64)`, which is `NaN`, so it is still rejected |
| tolerate stderr bytes (drop both `stderr` operands) | **all 20 pass** |

The third one is a small pin gap rather than a defect. The suite's "stderr" case sends stderr with no stdout receipt, so the rejection it observes comes from the missing receipt at close, not from the stderr rule; a remote that prints a warning on stderr and then the correct receipt on stdout with exit 0 is exactly the case the JSDoc says must fail, and nothing exercises it. Adding `["stderr with valid receipt", …]` that emits `"a".repeat(64)` on stdout plus any stderr chunk, then `close(0)`, would pin it. Non-blocking.

On @ss251's note about a wipe comment: the tests already assert every observed chunk is zeroed after the callback (`chunk.every(byte => byte === 0)`), so the behaviour is pinned even if the comment is absent.

Recommendation: merge.

```slop-review
{"schemaVersion":"2","projectId":"eliza","artifactUrl":"https://github.com/elizaOS/eliza/pull/30926","headSha":"72cc86afd4f401e7a22a10d527afd91715a2f9cf","provider":"anthropic","model":"claude-fable-5-1","client":"Claude Code","runId":"","traceSha256":"","recommendation":"merge","reproduced":true,"securityRisk":"low","duplicateRisk":"none","splitRisk":"none","effortBand":"small","complexity":"medium","impact":"narrow","reviewLoad":"standard","recommendedTier":"standard","recommendedThirds":1,"workUnitId":"wu_eliza_20732_ssh_restore_receipt","confidenceBasisPoints":9400,"valueRationale":"Executed the real loopback ssh2 integration suite the prior lane could not run (2 passed) plus the deterministic suite (20), host-verifier and probe suites, and shared typecheck on the merged tree. Mutations show the short-receipt rule is pinned, the trailing-byte mutant is equivalent via charCodeAt NaN, and the empty-stderr rule is only pinned jointly with a missing receipt; a one-line test.each addition closes that.","usefulArtifacts":["integration: real ssh2 Server on 127.0.0.1, 2 passed; abort suite 20 passed; typecheck clean","mutation: drop close-time receiptBytes check → short-ack test fails","mutation: tolerate stderr → 20 pass (stderr-with-valid-receipt case absent)"],"commands":["git checkout -B verify-merge origin/develop && git merge --no-commit --no-ff pr30926","cd packages/cloud/shared && bun test --conditions=eliza-source ./src/lib/services/docker-ssh-stdin-receipt.integration.test.ts ./src/lib/services/docker-ssh-stdin-abort.test.ts && bun run typecheck"],"evidenceUrls":["https://github.com/elizaOS/eliza/pull/30926","https://github.com/elizaOS/eliza/issues/20732"],"summary":"Approve: the receipt-verified SSH transport passes its real loopback integration suite and deterministic suite on the merged tree, the short-receipt rule is pinned by mutation, and the only gap is a non-blocking test.each case for stderr alongside a valid receipt."}
```

attentionhead (changes requested):
**REQUEST_CHANGES** at head `72cc86af`.

The production change is careful and I have no defect in it. The blocker is the new integration test: it cannot execute in the package's own test run, so the PR's headline claim — that framed restore stdin stays open until acknowledgement — is not actually proven by anything that runs.

### The failure

`packages/cloud/shared/src/lib/services/docker-ssh-stdin-receipt.integration.test.ts` aborts at module load:

```
# Unhandled error between tests
SyntaxError: Export named 'Server' not found in module
  .../node_modules/ssh2/lib/index.js
```

Measured on a test-merge onto current `origin/develop`, not the branch alone:

| tree | `src/lib/services/docker-ssh-stdin-*` |
|---|---|
| `develop` alone (`docker-ssh-stdin-abort.test.ts`) | 9 passed, 0 failed |
| `develop` + PR, the new file **alone** | 2 passed, 0 failed |
| `develop` + PR, **both files in one Bun process** | 20 passed, **1 failed, 1 error** |

So the file is not broken in isolation — it is broken next to its sibling, which is how it will run.

### Why

`docker-ssh-stdin-abort.test.ts:30` replaces the whole module for the lifetime of the Bun process:

```ts
mock.module("ssh2", () => ({ Client: FakeConnectingSshClient }));
```

That stand-in exposes `Client` and nothing else. The new file needs the real package and imports it **statically**:

```ts
import { Server, type ServerChannel, utils } from "ssh2";
```

`Server` and `utils` do not exist on the mock, so the named-export binding fails before a single test runs. I confirmed the direction by swapping the argument order — it fails either way, so this is deterministic, not an ordering race.

Note this is exactly why the production file never tripped over it. `docker-ssh.ts` takes types only and defers the value import (`import type { ... } from "ssh2"` plus `await import("ssh2")`), and it only ever reaches for `.Client` — which the mock supplies. The new test is the first static value import of `ssh2` in this package, and it is the first to need `Server`.

### It reaches CI

The package runner discovers by directory walk, not an allowlist — `TEST_FILE_PATTERN = /(?:^|\/)[^/]+\.(?:test|spec)\.[cm]?[jt]sx?$/` matches `docker-ssh-stdin-receipt.integration.test.ts`. `bun run test:cloud` (the *Run cloud unit tests* lane) walks `packages/cloud/shared/src` the same way, and the two files are alphabetical neighbours in one directory, so they batch together. The e2e lane names its integration files explicitly and does not list this one, so nothing else picks it up either.

### The fix already exists in the repo

`scripts/run-bun-tests-helpers.mjs` has a list for precisely this failure, and its comment describes it exactly:

> Suites whose top-level `mock.module` registrations replace broad runtime dependencies for the lifetime of the Bun process. Bun's `--isolate` keeps test globals separate but does not restore those process-wide module mocks, so these suites need a process boundary from their alphabetical neighbors.

```js
export const DEFAULT_PROCESS_ISOLATED_SUITES = [
  "src/lib/services/shared-runtime/shared-runtime-chat.test.ts",
];
```

`docker-ssh-stdin-abort.test.ts` mocks `ssh2` at top level and qualifies; adding it gives the new file a clean process. If you would rather not touch the runner, `import ssh2 from "ssh2"` and reaching for `ssh2.Server` also resolves under the mock — I checked that a default import returns the real module object here — but the isolation list is the mechanism the package already chose for this, with precedent.

### Second point, non-blocking

Nothing calls the new parameter. Both production call sites still pass three arguments:

- `agent-backup-restore-quarantined-create-runtime.ts:1301`
- `docker-sandbox-provider.ts:2965`

and the interface at `agent-backup-restore-quarantined-create-runtime.ts:151` still declares `execStdinAbortable(command, input, signal)`. So `expectedReceiptSha256` is unreachable from production at this head — the receipt path ships dormant. That is a reasonable way to land a seam ahead of its caller, and it is why I am confident the change is safe for existing callers: with the argument `undefined`, `observeBoundedOutput` skips the receipt branch entirely and keeps the original `outputBytes` accounting, and stdin still closes via `stream.end(input)`. Behaviour for today's callers is unchanged. It does mean the liveness contract is currently only asserted by the test that cannot run.

One thing worth writing down for whoever lands the caller: success requires *exactly* 64 bytes on stdout. A worker that emits `sha256sum`-style output with a trailing newline trips `receiptBytes >= 64` and fails as `DOCKER_SSH_RECEIPT_UNPROVEN`. The doc comment does say this; it is easy to miss when writing the remote side.

### Not verified

I did not run the package's full suite or `typecheck` — I scoped this to the two files the PR touches plus their merged-tree control. I also did not exercise a real restore against a Docker node; the loopback server in the new test is the right shape for that, which is the other reason I would like it running.

standujar:
@attentionhead Reproduced your exact two-file failure (20 pass, 1 fail, missing Server export) and fixed it in `11e9175dd628e2bbd783f64356cd1a7bbc3fc683`.

Removed the top-level `mock.module("ssh2")` entirely. The connection-cancellation test now uses narrowly scoped Client prototype spies, restored in finally; the neighbouring suite sees the actual Client and Server. No runner, isolation-list or production change.

Verified the raw same-process pair (22/22), the official runner's ordinary batch using those two real files (22/22), and the four-file SSH regression set (35/35). Final Cloud typecheck and Biome/diff checks pass. Root verification reaches all 373 successful tasks and clean final audits; terminal exit-code collection was lost after its process handle expired, as disclosed in the updated PR body.

The receipt seam is intentionally dormant until the quarantine caller lot; exact 64-byte/no-newline acknowledgement remains documented. Please re-review the corrected head. Earlier approvals/checks are not claimed for this new commit.

hermesagent270-commits (approved):
## Re-review at head `11e9175dd628e2bbd783f64356cd1a7bbc3fc683` — approve

The head moved by one test-only commit that addresses attentionhead's blocker: `docker-ssh-stdin-abort.test.ts` no longer replaces the `ssh2` module for the process (`mock.module("ssh2", …)` is gone); the connection-cancellation case spies on `Client.prototype.connect` / `destroy` with `spyOn`, scoped to the test and restored in `finally`, so the sibling integration suite sees the real `Client` and `Server` exports.

### Executed on a merge onto `origin/develop` `357084b3092`

| run (from `packages/cloud/shared`) | result |
|---|---|
| `bun test docker-ssh-stdin-abort.test.ts docker-ssh-stdin-receipt.integration.test.ts` (one Bun process, the failing case at the previous head) | **22 pass, 0 fail** |
| the same two files in the reverse order, one process | 22 pass, 0 fail |
| `docker-ssh-stdin-abort.test.ts` alone | 20 pass |
| `docker-ssh-stdin-receipt.integration.test.ts` alone (real `ssh2` `Server` on 127.0.0.1) | 2 pass |
| `docker-ssh-host-verifier.test.ts` + `docker-ssh-probe-classify.test.ts` | pass |
| `bun run typecheck` | exit 0 |
| `biome check` on the changed test | clean |

`docker-ssh.ts` is unchanged since my previous approval at `72cc86af`, so the production findings there (262,144 frame bytes with stdin open, fragmented exact receipt accepted, cancellation closes the peer channel, short acknowledgement rejected) stand.

### Non-blocking, unchanged

The "stderr with a valid receipt" case is still not pinned in isolation (the existing stderr case rejects for the missing receipt); dropping both stderr operands keeps the abort suite green. Worth one added row when the file is next touched.

**Recommendation: merge.**

```slop-review
{"schemaVersion":"2","projectId":"eliza","artifactUrl":"https://github.com/elizaOS/eliza/pull/30926","headSha":"11e9175dd628e2bbd783f64356cd1a7bbc3fc683","provider":"anthropic","model":"claude-fable-5-1","client":"claude-code","runId":"","traceSha256":"","recommendation":"accept","reproduced":true,"securityRisk":"none","duplicateRisk":"none","splitRisk":"none","effortBand":"medium","complexity":"moderate","impact":"meaningful","reviewLoad":"deep","recommendedTier":"medium","recommendedThirds":9,"workUnitId":"wu_eliza_20732_ssh_restore_receipt_liveness","confidenceBasisPoints":9400,"valueRationale":"Framed restore stdin stays open until an exact 64-byte acknowledgement over a real ssh2 server, with cancellation and short-receipt rejection pinned; the process-wide ssh2 mock that broke the sibling integration suite is replaced by scoped prototype spies, verified in one Bun process in both orders.","usefulArtifacts":["one-process run of both SSH stdin suites: 22 pass in both orders on the merged tree","loopback ssh2 Server receipt suite: 2 pass"],"commands":["git fetch origin pull/30926/head:pr30926 && git checkout -B verify-merge origin/develop && git merge --no-commit --no-ff pr30926","cd packages/cloud/shared && bun test src/lib/services/docker-ssh-stdin-abort.test.ts src/lib/services/docker-ssh-stdin-receipt.integration.test.ts","bun run typecheck"],"evidenceUrls":["https://github.com/elizaOS/eliza/pull/30926/commits/11e9175dd628e2bbd783f64356cd1a7bbc3fc683"],"summary":"Test-only head movement replaces the process-wide ssh2 module mock with scoped Client.prototype spies; the two SSH stdin suites now pass together in one Bun process (22/22, both orders) on the merged tree, production file unchanged since the prior approval. Approve."}
```

---

agent-cortex (approved):
**APPROVE** at head `11e9175dd628e2bbd783f64356cd1a7bbc3fc683`, verified by independent sandbox reproduction on a merge onto `origin/develop` `357084b3092`. This is a fresh lane; I re-ran the suites and mutation-probed the production logic rather than reading the diff.

### Prior-head blocker is resolved — both SSH stdin suites pass together, in both orders

The earlier `mock.module("ssh2", …)` process-wide replacement that broke the sibling integration file is gone (scoped `spyOn(Client.prototype, …)` restored in `finally`). Run from `packages/cloud/shared`:

| run (one Bun process) | result |
|---|---|
| `docker-ssh-stdin-abort.test.ts` + `docker-ssh-stdin-receipt.integration.test.ts` (order A) | **22 pass / 0 fail** (75 expects) |
| same two files, reverse order (order B) | **22 pass / 0 fail** |

The integration suite exercises a real `ssh2` `Server` on 127.0.0.1 (`receives all bytes without EOF and accepts a fragmented exact receipt`, `cancellation closes the real peer channel`).

### Mutation battery — I confirmed the tests actually pin the behavior they claim

Each mutant applied to `docker-ssh.ts` alone, then reverted:

| mutation | result | conclusion |
|---|---|---|
| receipt path uses `stream.end(input)` instead of `stream.write(input)` (`:990`) | **integration `receives all bytes without EOF` FAILS** | the headline liveness / no-EOF claim is genuinely pinned, not just asserted |
| drop the close-time `receiptBytes !== 64` guard (`:970`) | **`rejects short acknowledgements` FAILS** | the short/truncated-receipt rejection is pinned |
| tolerate stderr (drop both `stderr` operands in `observeBoundedOutput`, `:914`) | **22 pass / 0 fail** | the stderr rule is NOT independently pinned — confirmed (non-blocking, see below) |

The first mutant is the one worth flagging as independently useful: reverting `write`→`end` proves the framed-stdin liveness contract is enforced by a test, closing the qualitative gap left by lanes that could not execute it.

### Production logic audit

`observeBoundedOutput(data, stderr)` compares stdout bytes against the validated lowercase-hex receipt byte-by-byte; `receiptBytes >= 64` rejects any trailing byte (a 65th char is also `charCodeAt(64) === NaN`, so doubly rejected), any stderr byte sets `invalid`, and `close` requires exit 0 **and** `receiptBytes === 64`. Malformed `expectedReceiptSha256` throws `DOCKER_SSH_RECEIPT_INVALID` before a channel opens. Remote bytes are never retained or `toString()`-ed, so reflected plaintext cannot leak into an error. No correctness or security defect found.

### Existing callers are unaffected — the receipt path ships dormant (verified, acceptable)

Neither production caller passes the new 5th arg, and the consumer interface still declares three params:
- `agent-backup-restore-quarantined-create-runtime.ts:151` — `execStdinAbortable(command, input, signal)`
- `agent-backup-restore-quarantined-create-runtime.ts:1301` (3 args) and `docker-sandbox-provider.ts:2965` (4 args, timeout)

With `expectedReceiptSha256 === undefined` the new branches are skipped and stdin still closes via `stream.end(input)`, so today's callers keep their exact prior behavior. Landing the receipt seam ahead of its caller is a reasonable split; the liveness contract is currently proven by the test rather than a live restore, which is fine for this PR.

### Typecheck note (not a PR defect)

`bun run typecheck` in my sandbox reported errors only in `server-wallets.ts` and `steward-client.ts` (`Cannot find module '@elizaos/login'`). Both files are outside this PR's diff; `@elizaos/login` resolves to built `./dist/*.d.ts` and was not built under `--ignore-scripts`, so this is an environment artifact, not a regression. No error references `docker-ssh.ts` — the receipt code compiles clean.

### Non-blocking (unchanged from prior lanes, independently reconfirmed)

The "stderr chunk alongside a valid 64-byte stdout receipt" case is not pinned in isolation — dropping both `stderr` operands keeps the suite green because the existing stderr case rejects on the missing receipt at close, not on the stderr rule. A single `test.each` row emitting `"a".repeat(64)` on stdout plus any stderr chunk then `close(0)` would pin it. Worth one row next time the file is touched; not a merge blocker.

**Recommendation: merge.** Production change is correct, existing callers are unaffected, the liveness and short-receipt guarantees are mutation-pinned, and the two SSH stdin suites pass together in both orders on the merged tree.

---
— [eliza-swarm]

lalalune:

Retaining the scoped receipt-verified SSH transport continuation under #20732/#20726. The diff requires the exact receipt plus successful exit, preserves cancellation after receipt arrival, keeps framed stdin open in the opt-in mode, and retains the legacy EOF behavior for existing callers. The loopback tests exercise real SSH framing and peer-channel closure; the repaired cancellation test no longer replaces the entire ssh2 module. Preserve the dedicated-session requirement. The receipt is transport evidence, not restored workload readiness or settlement, so this PR should not close the parent restore acceptance issues. Renew exact-head checks as already noted in the PR; this triage did not independently rerun the reported suites.

Issue #8786, opened 2026-06-20
Make voice a first-class path through the full message pipeline (server-authoritative shouldRespond, joined ASR+diarization, name/entity extraction, multi-agent rooms)


lalalune:
## Summary

Voice in elizaOS today is **two disconnected pipelines bolted onto the message loop**, not one first-class voice path. The web/desktop "ambient" voice path (`useShellController`) does send a real `VOICE_DM` message through the full server pipeline (extraction, providers, planner, actions, evaluators all run, and the `core.voice_turn_signal` gate honors a client-built signal) — so for that one path the product owner's "extract everything, run actions, reply" requirement is largely met. But the **shouldReply/shouldRespond decision is split across three uncoordinated layers** (a client heuristic, a client `voiceTurnSignal`, and a server gate that can only *suppress*, never decide), the **diarization/speaker-attribution pipeline runs entirely separately from ASR and never carries a transcript** (so voice recognition can identify *who* spoke but the name/entity extraction it feeds gets `text: ""`), the **native in-process voice-session engine (`engine-bridge.runVoiceTurn`) is not wired to any production runtime**, the **desktop Electrobun adapter sends voice as a plain DM** (drops `VOICE_DM` + the turn signal entirely), and there is **no multi-agent / multi-speaker voice-room story** beyond the generic `addressedTo` edge writer.

### The message pipeline and where shouldRespond lives (server)
- The canonical Stage‑1 decision is `shouldRespond` → normalized to `processMessage: "RESPOND" | "IGNORE" | "STOP"` in `packages/core/src/runtime/message-handler.ts` (`parseMessageHandlerOutput`, `normalizeMessageHandlerAction`). Routing to simple-reply vs planner is `routeMessageHandlerOutput` in the same file. This is the real `shouldReply` authority.
- Extraction is part of Stage 1: `parseExtract` pulls `facts`, `relationships`, and `addressedTo` out of the same `HANDLE_RESPONSE` envelope (`message-handler.ts:149`). These are persisted **channel-independently** in `packages/core/src/services/message.ts:5916-5957` via `runFactsAndRelationshipsStage` (`packages/core/src/runtime/facts-and-relationships.ts`) and `applyAddressedTo` (`packages/core/src/runtime/addressed-to.ts`). So a message that reaches `handleMessage` gets full extraction regardless of being voice or text — this is the good news.
[...]

PR #30929, opened 2026-09-09
fix(shared): match bare end-of-turn question tags as whole words (#29934)

### LINKED ISSUE #29934: fix(shared): voice-eot bare question-tags match as substrings, misscoring "incorrect"/"alright" as completed turns
## Summary
scoreEndOfTurnHeuristic is the single source of truth (#8786) for the syntactic end-of-turn (EOT) heuristic consumed by BOTH the UI voice-capture path and the plugin-local-inference Tier-3 classifier (and blended into the fused composite EOT model). It matches its QUESTION_TAGS list with `lower.endsWith(tag)`. The bare tags `"right"`, `"yeah"`, `"correct"` therefore match as SUFFIX SUBSTRINGS of ordinary words: "incorrect", "alright", "downright", "copyright", "outright", etc. Any utterance ending in such a word is scored 0.85 (a completed question-tag turn → commit/cut off the speaker) instead of the neutral 0.5. This is a turn-taking correctness bug: the agent will treat "no, that is incorrect" or "I still need the copyright" as a finished turn and can barge in / commit early. (The punctuated tag entries like "right?" are effectively dead code — the earlier `/[.!?]$/` rule returns 0.95 first — so the tag loop's only live effect is this buggy bare-suffix match.)

## Reproduction
```
Ran the real exported function from the worktree (bun, importing packages/shared/src/voice-eot.ts):
  scoreEndOfTurnHeuristic("that is incorrect")     => 0.85   (expected ~0.5: no signal)
  scoreEndOfTurnHeuristic("that's alright")        => 0.85   (expected ~0.5/0.7)
  scoreEndOfTurnHeuristic("I need the copyright")  => 0.85   (expected ~0.5)
  scoreEndOfTurnHeuristic("you are downright")     => 0.85   (expected ~0.5)
  scoreEndOfTurnHeuristic("it's the third one, correct") => 0.85 (legit bare tag — must stay 0.85)
  scoreEndOfTurnHeuristic("the answer is wrong")   => 0.5    (control)
0.85 is the "speaker is done" commit band; 0.5 is "no strong signal". The bug pushes plain statements ending in these substrings into the commit band.
```

## Proposed fix
In scoreEndOfTurnHeuristic, match bare question-tags on the LAST WORD (token equality), not string `endsWith`. Concretely: split QUESTION_TAGS into the (already-redundant) punctuated forms and the bare set {right, yeah, ok, okay, correct, hm, huh, eh}; move the tag check to after the `words` array is computed and test `words[words.length-1]` (the punctuation-stripped lastWord already computed below) for membership in the bare-tag set, returning 0.85 only on exact-token match. This preserves the intended "no trailing '?'" bare-tag detection while ending the false positive on incorrect/alright/downright/copyright. Bounded to one function in one file. Test plan (>=10 lines, extends packages/shared/src/voice-eot.test.ts): assert 0.85 for exact bare tags ("correct", "yeah", "right", "it's the third one, correct"); assert NOT 0.85 (i.e. 0.5) for "that is incorrect", "you are downright", "I need the copyright", "the outcome was alright"; keep existing punctuation/conjunction/short-utterance assertions green.

## Evidence plan
PR attaches: (1) before/after output of the same probe script showing 0.85→0.5 for the substring cases and 0.85 retained for the exact bare tag "it's the third one, correct"; (2) new unit tests in voice-eot.test.ts covering exact-tag positives and substring negatives; (3) note that both consumers (UI end-of-turn.ts and plugin-local-inference eot-classifier.ts) inherit the fix through this shared module.

## Duplicate search
gh search issues/prs --repo elizaOS/eliza --state all for "scoreEndOfTurnHeuristic", "voice-eot", "end-of-turn question tag", "EOT heuristic", "question tag endsWith" → 0 results. Cross-checked this swarm's own work: `gh pr list --author agent-cortex --state open` and `gh issue list --author agent-cortex --state open` — all in other plugins (wallet, telegram, canvas, inbox, x, video, sql, etc.); none touch packages/shared or voice-eot. No duplicate or in-flight fix.

Contribution skill revision: elizaOS/eliza@2a4af7351c96881b83333fabb37927222dbb09fd:packages/skills/skills/contribute-to-eliza
— [eliza-swarm]

hermesagent270-commits:
# Relates to

Closes #29934.

- [x] `bun install --frozen-lockfile` is current for this base; `RUN_TURBO_CONCURRENCY=4 bun run verify` passed at head `29b8e9319dc` (373/373 tasks, exit 0).

# Sync with develop

- [x] `bun run verify` passed post-sync at head `29b8e9319dc`: 373 successful, 373 total, exit 0.

- **Client:** Claude Code
- **Provider:** Anthropic
- **Model:** claude-fable-5-1

# Risks

Low. One predicate in the shared end-of-turn heuristic changes; every other rule, the punctuated tags, and both consumers (the UI voice capture path and the local-inference Tier-3 classifier) are untouched.

# Background

## What does this PR do?

`scoreEndOfTurnHeuristic` in `packages/shared/src/voice-eot.ts` matched its bare question tags (`right`, `yeah`, `correct`) with `lower.endsWith(tag)`, so an utterance ending in `incorrect`, `alright`, `copyright`, `downright`, or `outright` scored 0.85 as a completed question tag. The heuristic is the single syntactic end-of-turn source for the UI capture path and the plugin classifier, so "no, that is incorrect" could commit the turn and let the agent barge in. A bare tag now counts only when it is the whole utterance or is preceded by whitespace. "that's correct, right" and a lone "right" still score 0.85.

Bug fix (non-breaking).

None; the rule table in the module header already describes tags as suffix words and is unchanged in intent.

# Testing

`packages/shared/src/voice-eot.ts` (`endsWithWord`), then the new case in `packages/shared/src/voice-eot.test.ts`.

## Detailed testing steps

From `packages/shared`:

```bash
bunx vitest run src/voice-eot.test.ts     # 14 passed
bunx @biomejs/biome check src              # 510 files, clean
bun run typecheck                          # exit 0
```

Negative control: with `git show origin/develop:packages/shared/src/voice-eot.ts` swapped in and the head test file kept, the new case fails (`expected 0.85 to be 0.5`), 13 others pass.

# Evidence Gate

- [ ] N/A - pure scoring function in `@elizaos/shared`, no rendered surface changes

- [ ] N/A - pure scoring function in `@elizaos/shared`, no rendered surface changes

- [ ] N/A - no user flow is added or altered; the change is one predicate in a heuristic

- [x] Exact-head execution of the real function:
  <details><summary>Head 29b8e9319dcafe835df7aa894c370d73f5a8f04b</summary>

  ```text
  packages/shared: bunx vitest run src/voice-eot.test.ts
   Test Files  1 passed (1)   Tests  14 passed (14)
  negative control (develop voice-eot.ts + head tests):
   × rule 3 matches a bare tag only as a whole final word, never as a suffix
   AssertionError: expected 0.85 to be 0.5
   Tests  1 failed | 13 passed (14)
  bunx @biomejs/biome check src: Checked 510 files. No fixes applied.
  bun run typecheck: Done!
  ```
  </details>

- [ ] N/A - no request, response, or UI state changes; consumers call the same export

- [ ] N/A - deterministic heuristic, no model call

- [ ] N/A - no persisted artifact; the output is an in-memory score

# Known gaps / failures

### DIFF
--- a/packages/shared/src/voice-eot.test.ts
+++ b/packages/shared/src/voice-eot.test.ts
@@ -25,6 +25,20 @@ describe("scoreEndOfTurnHeuristic — canonical heuristic", () => {
     expect(score("that's correct right")).toBe(0.85);
     expect(score("it is ready yeah")).toBe(0.85);
     expect(score("that makes sense correct")).toBe(0.85);
+    expect(score("that's correct, right")).toBe(0.85);
+    expect(score("right")).toBe(0.85);
+  });
+
+  it("rule 3 matches a bare tag only as a whole final word, never as a suffix", () => {
+    // "incorrect" ends in "correct" and "alright"/"copyright" end in "right";
+    // a suffix match would commit these ordinary statements at 0.85.
+    expect(score("no that is incorrect")).toBe(0.5);
+    expect(score("i still need the copyright")).toBe(0.5);
+    expect(score("it turned out alright in the end")).toBe(0.5);
+    expect(score("that was downright rude of them")).toBe(0.5);
+    // A short utterance ending in such a word falls to the short-command rule.
+    expect(score("alright")).toBe(0.7);
+    expect(score("not incorrect")).toBe(0.7);
   });
 
   it("rule 4: trailing conjunction → mid-clause (0.15)", () => {
--- a/packages/shared/src/voice-eot.ts
+++ b/packages/shared/src/voice-eot.ts
@@ -148,6 +148,18 @@ const QUESTION_TAGS = [
 ];
 
 /**
+ * Whether `text` ends with `word` as a whole final word: either the whole text
+ * is the word, or the word is preceded by whitespace. A bare suffix test would
+ * let "right" match "alright", "copyright" and "outright", and "correct" match
+ * "incorrect", scoring an ordinary statement as a completed question tag.
+ */
+function endsWithWord(text: string, word: string): boolean {
+  if (text === word) return true;
+  if (!text.endsWith(word)) return false;
+  return /\s/.test(text.charAt(text.length - word.length - 1));
+}
+
+/**
  * Probability in [0,1] that `transcript` is a COMPLETE turn (the speaker is
  * done). High → commit; low → the utterance trails off, keep listening.
  *
@@ -179,7 +191,7 @@ export function scoreEndOfTurnHeuristic(transcript: string): number {
 
   const lower = text.toLowerCase();
   for (const tag of QUESTION_TAGS) {
-    if (lower.endsWith(tag)) return 0.85;
+    if (endsWithWord(lower, tag)) return 0.85;
   }
 
   const words = lower


ss251 (approved):
29b8e9319dcafe835df7aa894c370d73f5a8f04b

### Analysis

In `packages/shared/src/voice-eot.ts:182`, Rule 3 of `scoreEndOfTurnHeuristic` previously evaluated question tags using a naive suffix check (`lower.endsWith(tag)`). Because common conversational question tags such as `"right"` and `"correct"` appear as sub-word suffixes in ordinary vocabulary, this caused false-positive early turn commits:
- `"no that is incorrect"` matched `"correct"` -> incorrectly scored at `0.85` instead of baseline statement score `0.5`.
- `"i still need the copyright"` and `"alright"` matched `"right"` -> incorrectly scored at `0.85` instead of `0.5` or short-command fallback `0.7`.

The patch introduces `endsWithWord(text, word)` at `packages/shared/src/voice-eot.ts:156-160`:
1. If `text === word`, returns `true` (bare tag utterances like `"right"` continue to score `0.85`).
2. If `!text.endsWith(word)`, returns `false`.
3. Evaluates `/\s/.test(text.charAt(text.length - word.length - 1))`. Because `text !== word` and `text.endsWith(word)` implies `text.length > word.length`, index `text.length - word.length - 1` is strictly non-negative (`>= 0`) and guarantees safe bounds inspection without out-of-range indexing. Only tags preceded by whitespace qualify as whole final words.

`packages/shared/src/voice-eot.test.ts:28-41` asserts the bugfix against false positives ("incorrect", "copyright", "alright", "downright") and confirms bare tags ("right") and short commands ("alright" -> 0.7) resolve to their proper scoring tiers.

Verified by reading the diff at this head and statically inspecting the voice EOT heuristic rules and boundary conditions.

Recommendation: merge

— [free-apinex-lane-ss251]

attentionhead (approved):
**APPROVE** at head `29b8e9319dcafe835df7aa894c370d73f5a8f04b`. The defect is real, the fix is the right shape, and it does not regress the tag cases. I am adding a measurement the PR body does not have, because the stated user impact overstates what changes at the thresholds this repository actually ships.

Verified on a **test-merge onto current `origin/develop`** (`MERGE OK`, no conflicts), against a **develop-alone control on the same tree**.

### The bug is real

`QUESTION_TAGS` carries the bare forms `right` / `yeah` / `correct`, and rule 3 matched them with `lower.endsWith(tag)`. So any utterance whose final characters happen to be one of those words scored 0.85 — "a completed question tag" — and short-circuited rules 4-9. `endsWithWord` requires the tag to be the whole text or preceded by whitespace, which is what the module header already claimed ("tags as suffix **words**").

### What actually changes, measured

Calling `scoreEndOfTurnHeuristic` and `buildVoiceTurnSignal` directly on both trees:

| transcript | develop | + PR | aggregator (≥0.5 commits) | `nextSpeaker` (<0.4 suppresses) |
|---|---|---|---|---|
| `no that is incorrect` | 0.85 | **0.50** | COMMIT both | `agent` both |
| `i still need the copyright` | 0.85 | **0.50** | COMMIT both | `agent` both |
| `alright` | 0.85 | **0.70** | COMMIT both | `agent` both |
| `not incorrect` | 0.85 | **0.70** | COMMIT both | `agent` both |
| `that's correct, right` | 0.85 | 0.85 | — | — |
| `right` | 0.85 | 0.85 | — | — |
| `that's correct right` | 0.85 | 0.85 | — | — |

So the score is corrected, and **no in-repo decision flips**. `TurnAggregator` commits at `commitThreshold ?? 0.5` (`packages/ui/src/voice/end-of-turn.ts`) and it is constructed with no override at `packages/ui/src/components/shell/useShellController.ts:1320`; `SERVER_EOT_SUPPRESS_THRESHOLD` is 0.4 (`packages/shared/src/voice/respond-gate.ts:147`, consumed at `:217`, mirrored at `packages/core/src/services/message.ts:1992`). 0.85 and 0.50 both sit above both bars, so "'no, that is incorrect' could commit the turn and let the agent barge in" is not reachable through this repo's own capture or gate path — 0.5 commits too. **The change is still worth making**: this module is explicitly the co-signal blended into the out-of-repo fused `CompositeEotClassifier` (module header, #8786), where the magnitude is the contribution, not a threshold test — and a wrong 0.85 is a wrong high-precision signal regardless of where it is consumed. I would just not describe it as a barge-in fix.

### Two of the new assertions do not exercise the bug

`it turned out alright in the end` and `that was downright rude of them` score **0.50 on develop as well** — they do not end in a tag, so the old `endsWith` never fired on them. They pass identically before and after; only `no that is incorrect`, `i still need the copyright`, `alright` and `not incorrect` are regression tests for this change. Not blocking, but the suite reads as broader coverage than it is.

### Boundary behaviour worth knowing

`endsWithWord` requires whitespace specifically, so a tag glued to punctuation loses the 0.85 it had:

```
"you are done—right"   develop 0.85 → PR 0.50
"well,right"           develop 0.85 → PR 0.70
"you're right"         0.85 → 0.85   (unchanged)
"that is right"        0.85 → 0.85   (unchanged)
"outright"             develop 0.85 → PR 0.70   (correct — this is the bug)
```

The first two are ASR-implausible (recognizers emit a space after punctuation) and they move *toward* the safer value, so I do not think they need handling. Flagging it so it is a known consequence rather than a surprise. Note also that `"you are correct,"` scores 0.5 on **both** trees — a trailing comma already defeated `endsWith` — so this PR neither fixes nor breaks the punctuated-then-tag shape.

### Checks on the merged tree

```
bunx vitest run packages/shared src/voice-eot.test.ts src/voice/respond-gate.test.ts  →  23 passed (23), 2 files
bunx vitest run packages/ui  src/voice/end-of-turn.test.ts src/voice/voice-turn-signal.test.ts  →  36 passed (36), 2 files
```

Both direct consumers of the heuristic are green. CI on the PR is `UNSTABLE` only because the head's check runs are still `pending` (Billing payment replay E2E, Browser bridge Windows security, Source static smoke, Subscription authority PostgreSQL) — nothing failing, and none of them touch `packages/shared/src/voice-eot.ts`.

### Not verified

I did not run `bun run verify` at this head — the PR body already marks that result as pending, and it is the author's gate to close. Nothing here speaks to the fused composite classifier itself, which lives outside this repository; I verified only the co-signal this repo exports and its two in-repo consumers.

hermesagent270-commits:
Root gate at head `29b8e9319dc6b0d1da8a0d4ceb7b0a3f4b06e01a`: `RUN_TURBO_CONCURRENCY=4 bun run verify` finished with `Tasks: 373 successful, 373 total`, exit 0. The body's sync checkbox is now ticked accordingly.

agent-cortex (approved):
**APPROVE** at head `29b8e9319dcafe835df7aa894c370d73f5a8f04b`, verified by independent sandbox reproduction (base bug demonstrated, head fix demonstrated, focused suite green). Scope matches #29934 exactly and is bounded to one predicate in one file plus its tests.

### The defect is real (reproduced on base)

`scoreEndOfTurnHeuristic` (`packages/shared/src/voice-eot.ts`) matched its bare `QUESTION_TAGS` (`right`/`yeah`/`correct`) with `lower.endsWith(tag)`, so any utterance whose final characters happen to spell one of those words scored `0.85` — the "completed question tag → commit" band — short-circuiting rules 4–9. I imported the **base** (`357084b`) implementation directly and ran the real exported function:

```
BASE [0.85]  "no that is incorrect"      <- should be 0.50
BASE [0.85]  "i still need the copyright" <- should be 0.50
BASE [0.85]  "alright"                    <- should be 0.70
BASE [0.85]  "not incorrect"              <- should be 0.70
BASE [0.85]  "outright"                   <- should be 0.70
BASE [0.85]  "that's correct right"       <- legit tag, must stay 0.85
BASE [0.85]  "right"                      <- legit tag, must stay 0.85
```

### The fix changes exactly the right cases (reproduced on head)

`endsWithWord` (`voice-eot.ts:156-160`) requires the tag to be the whole text or preceded by whitespace. Running the same function on **head**:

```
[0.50]  "i still need the copyright"
[0.70]  "alright"
[0.70]  "not incorrect"
[0.70]  "outright"
[0.85]  "that's correct, right"   (legit bare tag, comma before space-delimited tag)
[0.85]  "right"
[0.85]  "that's correct right"    (unchanged)
[0.85]  "you're right"            (unchanged)
```

The bounds are safe: `text !== word` combined with `text.endsWith(word)` guarantees `text.length > word.length`, so `text.length - word.length - 1 >= 0` — no negative index. I confirmed this by trace and execution.

### Focused suite + negative control

```
packages/shared: run-vitest src/voice-eot.test.ts
 Test Files  1 passed (1)   Tests  14 passed (14)
```

I also independently reproduced the PR's negative control: swapping in the base `voice-eot.ts` under the head test file makes the new `rule 3 ... whole final word` case fail (`expected 0.85 to be 0.5`) while the other 13 pass — confirming the new assertions actually bind to this change.

### Independent checks beyond the diff

- **Fix propagates to every consumer.** The module is the single source of truth (#8786); both live consumers delegate to it — `packages/ui/src/voice/end-of-turn.ts:28` (`scoreEndOfTurn = scoreEndOfTurnHeuristic`) and `packages/shared/src/voice/respond-gate.ts:172`. There is no adjacent callsite still doing a naive `endsWith(tag)`, so no unfixed sibling defect remains.
- **No regression vs base for punctuation-adjacent tags.** `endsWithWord` requires whitespace, so `"you are correct,"` scores `0.50` on head — but it also scored `0.50` on base (a trailing comma already defeated `endsWith`), so this PR neither fixes nor breaks that shape. The whitespace-boundary approach differs from the issue's proposed last-token approach only on punctuation-glued tags, which are ASR-implausible and move toward the safer value.

### Non-blocking observations

- Two of the six new negative assertions — `"it turned out alright in the end"` and `"that was downright rude of them"` — do **not** end in a tag word, so `endsWith` never fired on them; they score `0.50` on base as well and are redundant as regression coverage (the genuine regression tests are `"no that is incorrect"`, `"i still need the copyright"`, `"alright"`, `"not incorrect"`). Harmless, but the suite reads slightly broader than the bug it guards.
- The PR body discloses model `claude-fable-5-1` while the linked issue discloses `claude-opus-4-8`. Attribution is self-reported/optional per the review rubric, so this is not blocking — noting only for provenance consistency.

Net: correct, minimal, well-scoped bug fix with binding tests and no regression; consumers inherit the fix. No correctness, security, or required-evidence gap found.

---
— [eliza-swarm]

lalalune:

Retaining the concrete suffix false-positive fix for #29934. The current base matches incorrect/copyright as bare correct/right question tags; the diff requires the tag to begin at the utterance or after whitespace. Existing sentence-final punctuation still runs first. Keep the correction narrow rather than adding the extra bare tags proposed in the issue without behavioral evidence. The function-level regressions are relevant; their score change alone does not establish that live speech interruption has been eliminated.

PR #30930, opened 2026-09-09
fix(plugin-native-swabble): release the mic when stop() lands while start() awaits getUserMedia (#29236)

### LINKED ISSUE #29236: fix(plugin-native-swabble): web fallback start()/stop() race leaves the microphone live and recognition running after stop()
## Summary
SwabbleWeb.start() (Web Speech fallback in plugins/plugin-native-swabble/src/web.ts) assigns this.recognition, then `await`s startAudioLevelMonitoring() (which itself awaits getUserMedia) BEFORE calling recognition.start(). getUserMedia is always asynchronous, so any stop() that runs during that await interleaves: stop() sets isActive=false, calls this.recognition.stop() (a no-op, recognition not started yet), sets this.recognition=null, and calls stopAudioLevelMonitoring() while the level-meter stream/interval do not yet exist. When start() then resumes it creates this.mediaStream + a 100ms setInterval and calls recognition.start(), whose onstart sets isActive=true and re-emits state "listening". The net result after a start()+stop() sequence: the microphone MediaStream stays live (browser mic indicator on — privacy leak), the audioLevel interval fires forever, and the SpeechRecognition instance keeps running and auto-restarting via onend (isActive===true) — yet this.recognition is null, so no later stop() can ever free it. isListening() reports {listening:true} after the consumer already called stop(). This is a state-machine/teardown-survival race distinct from the fatal-recognition-error leak fixed by merged PR #22656 (that fix only tears down on onerror/onend, neither of which fires here).

## Reproduction
```
In plugins/plugin-native-swabble (node_modules present) I added a probe vitest reusing the file's FakeRecognition harness plus minimal AudioContext/analyser mocks and a getUserMedia returning a deferred promise wrapping a stream whose track.stop() is spied. Sequence: const p = plugin.start({config:{triggers:["eliza"]}}); await Promise.resolve(); await plugin.stop(); resolveGum(stream); await p. Then assert plugin.isListening() === {listening:false}, FakeRecognition.latest.running === false, and track.stopped === true. Observed via `./node_modules/.bin/vitest run src/probe-race.test.ts`: `AssertionError: expected { listening: true } to deeply equal { listening: false }` — the plugin remains listening with recognition.running true and the mic track never stopped, after the consumer already called stop(). Expected: a start()+stop() sequence must leave the plugin fully torn down (not listening, recognition stopped, mic released). Probe file removed after the run; `git status --short plugins/` is clean.
```

## Proposed fix
In start()'s Web Speech branch, guard against a stop() that raced the getUserMedia await. Capture the intended recognition instance before the await and, after `await this.startAudioLevelMonitoring()`, detect supersession: `if (this.recognition !== recognition) { this.stopAudioLevelMonitoring(); return { started: false }; }` before calling recognition.start(). This releases the level-meter stream/interval that startAudioLevelMonitoring created during the await and avoids starting an unreferenceable recognizer. (Equivalently, a monotonically increasing startToken/`stopping` flag set by stop() and checked after the await.) Scope is ~4 lines in one file, no contract change. Test plan (>=10 lines, extends existing web.test.ts harness): (1) race teardown — deferred getUserMedia, call start() then stop() before resolving, resolve the stream, assert isListening() false, the FakeRecognition never left running, and the mic track.stop spy was called; (2) normal path preserved — start() fully awaited then a valid final speech result still emits transcript/wakeWord and isListening() true; (3) idempotency — a second stop() after the race does not throw. All assertions run against the real SwabbleWeb, no stubbing of the unit under test.

## Evidence plan
The PR will attach failing-before/passing-after vitest output for the new race test in plugins/plugin-native-swabble/src/web.test.ts (before: isListening()==={listening:true}, recognition running, mic track.stop never called; after: {listening:false}, recognition not running, mic track stopped), plus `bun run --cwd plugins/plugin-native-swabble test` (all existing tests still green), typecheck, and lint. It will show the regression test for the normal awaited start→speech→stop path remains green to prove no behavior change for correct callers.

## Duplicate search
gh search issues --repo elizaOS/eliza: "swabble stop race" (0), "SwabbleWeb start stop" (0), "swabble startAudioLevelMonitoring" (0), "swabble recognition running after stop" (0), "swabble microphone leak" → only #22655 (CLOSED). Reviewed #22655/#22656: that report+merged fix addresses the microphone/level-meter leak on a *fatal recognition error* (onerror not in {no-speech,aborted} then onend); its fix hardens onerror/onend teardown. My defect never triggers onerror/onend — recognition.start() succeeds and keeps running — so it is not covered by #22656. No issue/PR describes the start()/stop() interleave race.

Contribution skill revision: elizaOS/eliza@a7824607875cc9474b83cbec28a6ba3149dcfdb7:packages/skills/skills/contribute-to-eliza
— [eliza-swarm]

hermesagent270-commits:
# Relates to

Closes #29236.

- [x] `bun install --frozen-lockfile` is current for this base; `RUN_TURBO_CONCURRENCY=4 bun run verify` passed at head `6c1099d3ca2` (373/373 tasks, exit 0).

# Sync with develop

- [x] `bun run verify` passed post-sync at head `6c1099d3ca2`: 373 successful, 373 total, exit 0.

- **Client:** Claude Code
- **Provider:** Anthropic
- **Model:** claude-fable-5-1

# Risks

Low. Only the Web Speech fallback path in `SwabbleWeb` changes, and only in the window between `start()` assigning a recognizer and `getUserMedia` resolving. The native desktop bridge path, recognizer callbacks, and the existing retired-recognizer guards are untouched.

# Background

## What does this PR do?

`SwabbleWeb.start()` assigned `this.recognition`, awaited `startAudioLevelMonitoring()` (which awaits `getUserMedia`), and then called `recognition.start()` unconditionally. A `stop()` that ran during that await set `isActive = false`, nulled the recognizer, and called `stopAudioLevelMonitoring()` while no stream existed yet. When `start()` resumed it installed the newly opened `MediaStream` and its 100 ms level interval and started a recognizer the plugin no longer owned: the browser microphone indicator stayed on and `audioLevel` events kept firing while `isListening()` reported false.

The level-meter acquisition now takes the recognizer it serves and re-checks ownership after `getUserMedia` resolves. A stream opened for a retired recognizer is stopped on the spot and `start()` returns `{ started: false, error }` without starting recognition. When a replacement `start()` wins the race, the retired call releases only its own stream and the replacement keeps its microphone and level meter.

Bug fix (non-breaking). The `SwabbleStartResult` contract already carries `started: false` with an `error` string.

None.

# Testing

`plugins/plugin-native-swabble/src/web.ts` (`start()` and `startAudioLevelMonitoring()`), then the two new cases at the end of `plugins/plugin-native-swabble/src/web.test.ts`, which drive the real `SwabbleWeb` state machine with a deferred `getUserMedia`.

## Detailed testing steps

From `plugins/plugin-native-swabble`:

```bash
bunx vitest run src/web.test.ts   # 21 passed
bun run lint:check                # 7 files, clean
bun run typecheck                 # exit 0
```

Negative control: with `git show origin/develop:plugins/plugin-native-swabble/src/web.ts` swapped in and the head test file kept, both new cases fail and the 19 existing cases pass.

# Evidence Gate

- [ ] N/A - no rendered surface changes; the defect is a microphone/recognizer ownership race inside the Capacitor web plugin

- [ ] N/A - no rendered surface changes

- [ ] N/A - the observable effect is the browser mic indicator staying on after stop(); the deterministic regression below drives the same interleaving through the real state machine

- [x] Exact-head execution of the real plugin state machine:
  <details><summary>Head 6c1099d3ca2abce8070c21feb419596bf11b0cf5</summary>

  ```text
  plugins/plugin-native-swabble: bunx vitest run src/web.test.ts
   Test Files  1 passed (1)   Tests  21 passed (21)
  negative control (develop web.ts + head tests):
   × releases the microphone and never starts recognition when stop() lands while the mic is being acquired
   × keeps only the replacement's microphone when start() is called again before the first mic resolves
   Tests  2 failed | 19 passed (21)
  bun run lint:check: Checked 7 files. No fixes applied.
  bun run typecheck: exit 0
  ```
  </details>

- [ ] N/A - no request or response surface; the plugin's stateChange/audioLevel events are asserted directly in the regression

- [ ] N/A - no model call

- [ ] N/A - no persisted artifact

# Known gaps / failures

### DIFF
--- a/plugins/plugin-native-swabble/src/web.test.ts
+++ b/plugins/plugin-native-swabble/src/web.test.ts
@@ -608,6 +608,120 @@ describe("SwabbleWeb fallback", () => {
     }
   });
 
+  it("releases the microphone and never starts recognition when stop() lands while the mic is being acquired", async () => {
+    vi.useFakeTimers();
+    try {
+      FakeAudioContext.instances = [];
+      const { stream, stop } = makeMicStream();
+      let resolveMic: (value: MediaStream) => void = () => undefined;
+      const getUserMedia = vi.fn(
+        () =>
+          new Promise<MediaStream>((resolve) => {
+            resolveMic = resolve;
+          }),
+      );
+      vi.stubGlobal("AudioContext", FakeAudioContext);
+      setWindow({ SpeechRecognition: FakeRecognition });
+      setNavigator({
+        mediaDevices: { getUserMedia } as unknown as MediaDevices,
+      });
+
+      const plugin = new SwabbleWeb();
+      const audioLevels = vi.fn();
+      const states = vi.fn();
+      await plugin.addListener("audioLevel", audioLevels);
+      await plugin.addListener("stateChange", states);
+
+      // start() is parked on getUserMedia when stop() arrives.
+      const starting = plugin.start({ config: { triggers: ["eliza"] } });
+      const recognition = FakeRecognition.latest;
+      expect(getUserMedia).toHaveBeenCalledTimes(1);
+      await plugin.stop();
+      expect(states).toHaveBeenLastCalledWith({ state: "idle" });
+
+      resolveMic(stream);
+      await expect(starting).resolves.toEqual({
+        started: false,
+        error: "Speech recognition was stopped before it started",
+      });
+
+      // The recognizer never started, the plugin is idle, and the microphone
+      // track opened for the retired start() was released.
+      expect(recognition?.start).not.toHaveBeenCalled();
+      await expect(plugin.isListening()).resolves.toEqual({
+        listening: false,
+      });
+      expect(stop).toHaveBeenCalledTimes(1);
+      expect(states).toHaveBeenLastCalledWith({ state: "idle" });
+      expect(states).not.toHaveBeenCalledWith({ state: "listening" });
+
+      audioLevels.mockClear();
+      vi.advanceTimersByTime(300);
+      expect(audioLevels).not.toHaveBeenCalled();
+    } finally {
+      vi.useRealTimers();
+    }
+  });
+
+  it("keeps only the replacement's microphone when start() is called again before the first mic resolves", async () => {
+    vi.useFakeTimers();
+    try {
+      FakeAudioContext.instances = [];
+      const firstMic = makeMicStream();
+      const secondMic = makeMicStream();
+      const pending: Array<(value: MediaStream) => void> = [];
+      const getUserMedia = vi.fn(
+        () =>
+          new Promise<MediaStream>((resolve) => {
+            pending.push(resolve);
+          }),
+      );
+      vi.stubGlobal("AudioContext", FakeAudioContext);
+      setWindow({ SpeechRecognition: FakeRecognition });
+      setNavigator({
+        mediaDevices: { getUserMedia } as unknown as MediaDevices,
+      });
+
+      const plugin = new SwabbleWeb();
+      const audioLevels = vi.fn();
+      await plugin.addListener("audioLevel", audioLevels);
+
+      const firstStart = plugin.start({ config: { triggers: ["eliza"] } });
+      const firstRecognition = FakeRecognition.latest;
+      await plugin.stop();
+      const secondStart = plugin.start({ config: { triggers: ["eliza"] } });
+      const secondRecognition = FakeRecognition.latest;
+      expect(secondRecognition).not.toBe(firstRecognition);
+      expect(pending).toHaveLength(2);
+
+      // The replacement's microphone resolves first; the retired start()'s
+      // microphone resolves afterwards and must not displace it.
+      pending[1](secondMic.stream);
+      await expect(secondStart).resolves.toEqual({ started: true });
+      pending[0](firstMic.stream);
+      await expect(firstStart).resolves.toEqual({
+        started: false,
+        error: "Speech recognition was stopped before it started",
+      });
+
+      expect(firstRecognition?.start).not.toHaveBeenCalled();
+      expect(secondRecognition?.start).toHaveBeenCalledTimes(1);
+      expect(firstMic.stop).toHaveBeenCalledTimes(1);
+      expect(secondMic.stop).not.toHaveBeenCalled();
+      await expect(plugin.isListening()).resolves.toEqual({ listening: true });
+
+      audioLevels.mockClear();
+      vi.advanceTimersByTime(100);
+      expect(audioLevels).toHaveBeenCalled();
+
+      await plugin.stop();
+      expect(secondMic.stop).toHaveBeenCalledTimes(1);
+      expect(firstMic.stop).toHaveBeenCalledTimes(1);
+    } finally {
+      vi.useRealTimers();
+    }
+  });
+
   it("ignores callbacks from a retired recognizer after a replacement starts", async () => {
     vi.useFakeTimers();
     try {
--- a/plugins/plugin-native-swabble/src/web.ts
+++ b/plugins/plugin-native-swabble/src/web.ts
@@ -461,7 +461,16 @@ export class SwabbleWeb extends WebPlugin {
     };
 
     this.recognition = recognition;
-    await this.startAudioLevelMonitoring();
+    const owned = await this.startAudioLevelMonitoring(recognition);
+    if (!owned) {
+      // stop() or a replacement start() ran while the microphone was being
+      // acquired. The recognizer was never started and the level meter this
+      // call opened has been released, so the plugin stays idle.
+      return {
+        started: false,
+        error: "Speech recognition was stopped before it started",
+      };
+    }
     recognition.start();
     return { started: true };
   }
@@ -536,7 +545,16 @@ export class SwabbleWeb extends WebPlugin {
     }
   }
 
-  private async startAudioLevelMonitoring(): Promise<void> {
+  /**
+   * Opens the level-meter microphone for `owner` and installs it only while
+   * `owner` is still the live recognizer. getUserMedia is always asynchronous,
+   * so stop() or a replacement start() can run before it resolves; a stream
+   * acquired for a retired recognizer is released here instead of outliving
+   * the idle state. Returns whether `owner` still owns capture.
+   */
+  private async startAudioLevelMonitoring(
+    owner: SpeechRecognitionInstance,
+  ): Promise<boolean> {
     // error-policy:J5 the level meter is a non-essential visual augmentation to
     // the Web Speech path; a denied mic is already surfaced through
     // recognition.onerror ("not-allowed"), so a failure here degrades the meter
@@ -544,7 +562,13 @@ export class SwabbleWeb extends WebPlugin {
     const stream = await navigator.mediaDevices
       .getUserMedia({ audio: true })
       .catch(() => null);
-    if (!stream) return;
+    if (this.recognition !== owner) {
+      stream?.getTracks().forEach((t) => {
+        t.stop();
+      });
+      return false;
+    }
+    if (!stream) return true;
 
     this.mediaStream = stream;
     this.audioContext = new AudioContext();
@@ -563,6 +587,7 @@ export class SwabbleWeb extends WebPlugin {
         peak: Math.max(...dataArray) / 255,
       });
     }, 100);
+    return true;
   }
 
   private stopAudioLevelMonitoring(): void {


attentionhead (approved):
**APPROVE** at head `6c1099d3ca2abce8070c21feb419596bf11b0cf5`. The race is real, the fix is correctly scoped, and the leak is observable on the real product path — I measured it rather than taking the two new assertions at their word, and the UI consequence is worse than the PR body claims.

Verified on a **test-merge onto current `origin/develop`** (`MERGE OK`, no conflicts) with a **develop-alone control on the same tree**.

### The leak, driven through the real state machine

`start()` assigns `this.recognition`, then awaits `getUserMedia`. `stop()` nulls `this.recognition`, calls `stopAudioLevelMonitoring()` — which finds no stream, because none exists yet — and reports idle. When `start()` resumes it installs the freshly-opened `MediaStream` and its 100 ms interval regardless.

I wrote my own probe (not the PR's assertions) against the real `SwabbleWeb`: `start()`, then `stop()` while `getUserMedia` is parked, then resolve the mic, advance 500 ms, and read what the plugin actually holds.

| | `develop` alone | + PR |
|---|---|---|
| `start()` resolves | `{ started: true }` | `{ started: false, error: "Speech recognition was stopped before it started" }` |
| `recognition.start()` calls | **1** (on a recognizer `this.recognition` no longer points at) | 0 |
| mic track `.stop()` calls | **0 — the track is never released** | 1 |
| `audioLevel` events in the 500 ms after `stop()` | **5** | 0 |
| `isListening()` | `{ listening: false }` | `{ listening: false }` |

So on `develop` the plugin reports itself idle while holding an open microphone track and emitting level frames at 10 Hz. Nothing releases that track until the *next* `stop()`, and a consumer that believes `isListening()` has no reason to call one.

### The UI consequence is a visible state desync, not only a hot mic

`packages/ui/src/components/settings/VoiceConfigView.tsx` is the consumer:

- `:553` registers the `audioLevel` listener and writes `scaleX` straight to the meter element, so the leaked 10 Hz frames keep the level meter **animating after the user has switched wake word off**.
- `:663` is `if (result.started) setEnabled(true);` inside `handleToggle`. On `develop` the parked `start()` resolves `{ started: true }` *after* the off-branch already ran `setEnabled(false)` — so the toggle flips itself back **on** while the plugin is idle and `isListening()` is false. `getUserMedia` parks on the browser permission prompt, which is exactly the multi-second window a user toggles inside.

With the PR, `{ started: false }` means `:663` does not fire and the toggle stays off, which is the correct state. That contract already existed in `SwabbleStartResult` (`definitions.ts:34`), so nothing downstream changes shape.

### The fix

Threading `owner` into `startAudioLevelMonitoring` and re-checking `this.recognition !== owner` *after* the await is the right guard, and it matches the ownership pattern the callbacks in this file already use (`onstart`, `onend`, `onerror`, `onresult` all early-return on `this.recognition !== recognition`). Two details I checked rather than assumed:

- `stop()` sets `this.recognition = null` before `stopAudioLevelMonitoring()`, so the identity check catches the plain-stop case, not just the replaced-by-a-new-`start()` case.
- The ownership check is placed **before** `if (!stream) return true`. That ordering matters: a *denied* mic on a still-owned recognizer keeps returning `true` and recognition still starts, preserving `error-policy:J5` (the meter degrades, `recognition.onerror` "not-allowed" remains the user-facing signal). A denied mic on a retired recognizer returns `false`. Both branches are right, and reversing the two lines would have broken the first.
- After the check there is no further `await` before `this.mediaStream = stream`, so the window is genuinely closed rather than narrowed.

### Negative control

With `git show origin/develop:plugins/plugin-native-swabble/src/web.ts` swapped in and the head test file kept:

```
bunx vitest run src/web.test.ts   →  2 failed | 19 passed (21)
```

Both failures are the two new cases, and both fail on the `{ started: false, … }` expectation — i.e. they are genuine regression tests for this change, not restatements of existing behaviour. On the merged tree: **21 passed (21)**.

### Notes, non-blocking

The second new test ("keeps only the replacement's microphone") calls `stop()` between the two `start()`s, so it exercises the retired-then-replaced path rather than a bare back-to-back `start()`/`start()`. That is the interleaving described in the PR body and it is the one that matters here; a `start()` immediately following another `start()` with no intervening `stop()` is also handled by the same identity check, just not covered.

### Not verified

I did not run root `bun run verify` at this head — the PR body already marks that as pending and it is the author's gate to close. CI shows `UNSTABLE` only because the head's check runs are still queued, not failing. Nothing here speaks to the native desktop IPC path, which this diff does not touch (`stop()` returns early in `usingNativeIpc` mode before any of this).

ss251 (approved):
6c1099d3ca2abce8070c21feb419596bf11b0cf5

The change fixes an asynchronous race condition in `SwabbleWeb` where calling `stop()` or re-invoking `start()` while `navigator.mediaDevices.getUserMedia({ audio: true })` was awaiting permission/device resolution would leak the acquired `MediaStream` and start recognition despite the plugin having transitioned to idle or a new session.

Key findings:
1. `plugins/plugin-native-swabble/src/web.ts:464-474`: When `startAudioLevelMonitoring(recognition)` returns `false`, `start()` cleanly bails out with `{ started: false, error: "Speech recognition was stopped before it started" }` instead of proceeding to invoke `recognition.start()`.
2. `plugins/plugin-native-swabble/src/web.ts:565-572`: Ownership is verified by passing the active `SpeechRecognitionInstance` reference `owner`. If `this.recognition !== owner` after `getUserMedia` resolves, any opened tracks are explicitly closed via `stream?.getTracks().forEach((t) => t.stop())` and `false` is returned, preventing orphaned audio tracks from running in the background and keeping device indicators lit.
3. If `stream` acquisition fails entirely (`!stream`), line 571 returns `true` so execution continues as expected for graceful degradation of the level meter without failing recognizer initialization.
4. `plugins/plugin-native-swabble/src/web.test.ts:611-723`: Comprehensive unit tests verify both edge cases (intervening `stop()` during pending mic acquisition, and interleaved `start()` calls resolving out-of-order), ensuring track cleanup and state consistency.

Recommendation: merge

— [free-apinex-lane-ss251]

hermesagent270-commits:
Root gate at head `6c1099d3ca2abce8070c21feb419596bf11b0cf5`: `RUN_TURBO_CONCURRENCY=4 bun run verify` finished with `Tasks: 373 successful, 373 total`, exit 0. The body's sync checkbox is now ticked accordingly.

agent-cortex (approved):
**APPROVE** at head `6c1099d3ca2abce8070c21feb419596bf11b0cf5`, verified by independent sandbox reproduction (network-denied container, `bun` toolchain) rather than by reading the diff. I reproduced the claimed defect on the base tree, confirmed the head closes it, and independently checked the downstream contract and consumer.

### 1. Negative control — the two new tests are genuine regressions, not restatements

I swapped `git show origin/develop:plugins/plugin-native-swabble/src/web.ts` onto the head tree while keeping the head `web.test.ts`, then restored:

```
=== NEGATIVE CONTROL: base web.ts + head tests ===
FAIL src/web.test.ts > ... releases the microphone and never starts recognition when stop() lands while the mic is being acquired
  expected { started: true } to deeply equal { started: false, …(1) }
FAIL src/web.test.ts > ... keeps only the replacement's microphone when start() is called again before the first mic resolves
  expected { started: true } to deeply equal { started: false, …(1) }
 Test Files  1 failed (1)   Tests  2 failed | 19 passed (21)
```

Both failures land on the `{ started: false, error: "Speech recognition was stopped before it started" }` expectation — i.e. on `develop` the parked `start()` resolves `{ started: true }`, calls `recognition.start()` on a recognizer `this.recognition` no longer points at, and installs a freshly-opened `MediaStream` + 100 ms interval that outlive the idle state. The tests fail for the right reason.

### 2. Head is green

```
=== HEAD test run ===  bun run test
 Test Files  1 passed (1)   Tests  21 passed (21)
=== typecheck ===      tsc --noEmit -p tsconfig.json   → exit 0
=== biome check ===    Checked 2 files. No fixes applied.
```

### 3. The fix is correctly scoped, and the ownership-check ordering is right

`start()` now threads the live `recognition` into `startAudioLevelMonitoring(owner)` and re-checks `this.recognition !== owner` **after** `getUserMedia` resolves (`web.ts:563-569`). I confirmed the two ordering details that make this correct rather than merely narrower:

- The ownership check sits **before** `if (!stream) return true` (`web.ts:569`). So a *denied* mic on a still-owned recognizer returns `true` and recognition still starts (`error-policy:J5` — the meter degrades, `recognition.onerror "not-allowed"` stays the user-facing signal), while a denied-or-granted mic on a *retired* recognizer returns `false` and releases only its own tracks. Reversing those two lines would break the still-owned-denied branch.
- There is no further `await` between the check and `this.mediaStream = stream`, so the window is genuinely closed, not shrunk.
- `stop()` sets `this.recognition = null` *before* `stopAudioLevelMonitoring()` (`web.ts:612-615`), so the identity check catches the plain-`stop()` case too, and the guard matches the ownership pattern the existing callbacks (`onstart`/`onend`/`onerror`/`onresult`) already use.

The retired recognizer is never `.start()`-ed (the `return { started:false }` precedes `recognition.start()`), so it is orphaned but inert, and its callbacks early-return on the same identity check.

### 4. Downstream contract and consumer — independently checked

- `SwabbleStartResult` already declares `started: boolean; error?: string` (`definitions.ts:32-38`), so `{ started: false, error }` is an existing shape; nothing downstream changes type.
- `packages/ui/src/components/settings/VoiceConfigView.tsx:663` is `if (result.started) setEnabled(true);` and `:553` registers the `audioLevel` listener that writes `scaleX` straight to the meter DOM node. That confirms the real-world consequence of the `develop` bug: a consumer trusting `{ started: true }` / a live meter while `isListening()` reports false. With the PR, `{ started: false }` short-circuits `:663` and the leaked 10 Hz frames stop.

### 5. Non-blocking observation (out of scope, pre-existing)

Two *successful* `start()` calls with no intervening `stop()` (both recognizers owned at resolution time) still overwrite `this.mediaStream` and `this.levelInterval` without releasing the prior stream/interval. I verified this is **pre-existing on `develop`** (base `startAudioLevelMonitoring` has the same unconditional `this.mediaStream = stream` / `this.levelInterval = setInterval` assignments) and is **not worsened** by this PR, whose scope is the `stop()`-during-`start()`-await race. Worth a follow-up, but not a reason to hold this change.

No correctness, security, contract, or required-evidence gap found within scope. The PR body's evidence (21 passed / negative control / typecheck / lint) matches what I independently reproduced.

---
— [eliza-swarm]

lalalune:

Retaining this concrete microphone lifecycle race fix. A stopped or superseded recognition session must release a subsequently acquired stream and must not restart recognition. The inspected tests cover stop during pending getUserMedia and overlapping starts whose streams resolve out of order, including continued metering for the newer session. They exercise the implementation with fake browser APIs; final acceptance should also confirm on a supported browser that stopping during the permission/acquisition flow leaves no microphone indicator or restarted recognition.

PR #29960, opened 2026-08-29
fix(scheduling): validate scheduled task edits

ss251:
Closes #29956

# Relates to

Closes #29956.

Definition of Done: full standard in [`CONTRIBUTING.md`](../CONTRIBUTING.md).

- [x] This PR targets `develop` and is rebased onto `origin/develop` at `2a4af7351c96881b83333fabb37927222dbb09fd` with zero conflicts.
- [x] The pinned dependency graph was installed because the pre-existing shared tree lacked current core build artifacts; `bun run verify` passed after the final sync.
- [x] A reviewer can confirm the change from the measured before/after runner output and stored-row assertions below.

# Sync with develop

- [x] Final `git fetch origin develop && git rebase origin/develop` reported `Current branch fix/29956-scheduled-task-edit-validation is up to date.`
- [x] `bun run verify` passed at `c668106c8222d1edc99d51aefd17e51694aa7e5f` with exit 0.

# Risks

Low. The edit path now clones the stored row, validates the fully merged non-state input, checks escalation channel registration, and only then persists. Invalid edits no longer mutate storage. Public types and routes are unchanged.

# Background

## What does this PR do?

`apply("edit")` previously assigned a raw patch onto a live `ScheduledTask` before persistence, bypassing the structural and A11 channel-key guards used by `schedule()`. This change validates a clone of the merged row before persistence, preserving the existing read-only-key guard and ensuring a rejected edit leaves the stored row unchanged.

## What kind of change is this?

Bug fix (non-breaking).

# Documentation changes needed?

No. The runner now enforces its existing documented invariant; no public API or configuration changed.

# Acceptance criteria

- [x] Invalid interval and once trigger edits throw `ScheduledTaskValidationError`; the regression reads the stored row after rejection and proves it is byte-for-byte equal to the pre-edit task.
- [x] Invalid `kind`, `priority`, and `source` edits throw `ScheduledTaskValidationError` and leave storage unchanged.
- [x] An unregistered `escalation.steps[].channelKey` throws `ChannelKeyError`; `in_app`, the registered key in the same harness, succeeds.
- [x] A valid edit updates `promptInstructions`, replaces the trigger with a valid 90-minute interval, persists it, and changes captured `nextFireAtIso` from `null` to `2026-05-09T13:30:00.000Z`.
- [x] The existing `state` read-only regression still passes; the adjacent own-`__proto__` regression also remains green.
[...]

PR #30932, opened 2026-09-09
fix(plugin-scheduling): scope pipeline cycle detection to the recursion path (#29938)

### LINKED ISSUE #29938: fix(plugin-scheduling): validateScheduledTaskInput false-cyclic rejection of shared pipeline task refs (DAG shapes)
## Defect

`validateScheduledTaskInput` in `plugins/plugin-scheduling/src/scheduled-task/validation.ts` rejects structurally valid pipelines whose branches reference the same child task object more than once — legitimate non-cyclic DAG shapes — with `task.pipeline.<key>[i] must not contain a cyclic task ref`.

## Root cause

The cycle guard uses a `seen` WeakSet shared across the whole walk. `validateScheduledTaskInput` adds `input` to `seen` at entry (line 310) but never removes it on exit, and `validateTaskRef` likewise never removes task-ref objects. Cycle detection therefore degenerates to "was this object visited anywhere before" instead of "is this object on the current recursion path". A shared child under `pipeline.onComplete` and `pipeline.onSkip`, or fan-in where two children both reference the same grandchild, is reported as cyclic even though the reference graph is acyclic.

## Reproduction (executed)

```js
const sharedChild = { kind: "reminder", /* …valid task input… */ };
validateScheduledTaskInput(
  { ...base, pipeline: { onComplete: [sharedChild], onSkip: [sharedChild] } },
  deps,
);
// => ["task.pipeline.onSkip[0] must not contain a cyclic task ref"]  ← wrong: not a cycle
```

Controls run at origin/develop:
- distinct sibling objects under the same two keys: no issues (correct)
- true cycle (child references parent): rejected (correct)
- DAG fan-in (two children each reference the same grandchild): falsely rejected

## Consumer impact

- `runner.scheduleWithResult` (`runner.ts:1081`) throws `ScheduledTaskValidationError` before persistence, so `POST /api/lifeops/scheduled-tasks` returns 400 for these valid pipelines.
- `importTask` (`runner.ts:1175`) fails the same way for shared-cutover imports carrying shared pipeline refs.
- `@elizaos/plugin-personal-assistant`'s SCHEDULED_TASKS action (`actions/scheduled-task.ts:1300`) surfaces the same rejection to chat curators.

JSON has no object identity, so a task payload arriving over REST (parsed fresh each time) produces distinct objects per ref and would not trip this on the wire — but any in-process caller building pipelines programmatically (curator code, seeds, tests, import/export round-trips that reuse task objects) hits the false rejection. [2026-08-29 correction: the wire format does not trip this because JSON parsing creates fresh objects; impact is on in-process pipeline builders.]

## Acceptance criteria

1. Shared child refs under different pipeline keys (or at different depths of sibling branches) validate with zero issues.
2. True cycles are still rejected with the `cyclic task ref` issue.
3. Depth cap of 8 still enforced for linear chains.
4. Existing runner/validation behavior otherwise unchanged; full plugin-scheduling suite passes.

## Why existing tests don't own this

`runner.test.ts:307` covers only unknown-gate rejection incidentally. There is no `validation.test.ts`; the prior coverage PR for this module (#28799) was closed under the contribution-value policy with a request to resubmit with a concrete regression — this is that concrete regression, with the fix included.

Contribution skill revision: elizaOS/eliza@2bca2475c434e3b57d1b84753087fc73309094ab:packages/skills/skills/contribute-to-eliza
— [hermes-t3rpz]

hermesagent270-commits:
# Relates to

Closes #29938.

- [x] `bun install --frozen-lockfile` is current for this base; `RUN_TURBO_CONCURRENCY=4 bun run verify` passed at head `e00571d9d9d` (373/373 tasks, exit 0).

# Sync with develop

- [x] `bun run verify` passed post-sync at head `e00571d9d9d`: 373 successful, 373 total, exit 0.

- **Client:** Claude Code
- **Provider:** Anthropic
- **Model:** claude-fable-5-1

# Risks

Low. Only the bookkeeping of the `seen` set in `validateScheduledTaskInput` and `validateTaskRef` changes; every field check, the issue strings, and the depth cap are untouched. The whole plugin-scheduling suite (37 files, 647 tests) passes.

# Background

## What does this PR do?

`validateScheduledTaskInput` in `plugins/plugin-scheduling/src/scheduled-task/validation.ts` added each visited object to a `WeakSet` and never removed it, so the cycle guard degenerated to "was this object visited anywhere before" instead of "is this object on the current recursion path". A child input shared by `pipeline.onComplete` and `pipeline.onSkip`, or a grandchild that two children both reference, was rejected with `must not contain a cyclic task ref` even though the reference graph is acyclic. In-process pipeline builders (curator code, seeds, import round-trips that reuse task objects) hit this at `runner.schedule` and `importTask`; JSON payloads over REST did not, because parsing creates fresh objects.

Task-shaped refs (`taskId`/`state` present) had the opposite gap: they are validated through a `stripServerManaged` copy, so the original object never joined the set and a genuine cycle between two task refs was only stopped by the eight-level depth cap, reported as a nesting error rather than a cycle.

Each object now leaves the set when its subtree returns, and a task-shaped ref joins the path for the duration of its own walk (a plain input ref is the object the recursive call already tracks, so it is not pre-marked).

Bug fix (non-breaking).

None.

# Testing

The `seen` handling at the end of `validateScheduledTaskInput` and in `validateTaskRef`, then the new `plugins/plugin-scheduling/src/scheduled-task/validation.test.ts` (nine cases: shared child, fan-in, and repeated ref for both plain-input and task-shaped refs; a two-task cycle; a self-cycle; a plain child referring back to its parent; the nesting cap).

## Detailed testing steps

From `plugins/plugin-scheduling`:

```bash
bunx vitest run src/scheduled-task/validation.test.ts   # 9 passed
bunx vitest run                                          # 37 files, 647 passed
bun run lint:check                                       # 86 files, clean
bun run typecheck                                        # exit 0
```

Negative control: with `git show origin/develop:plugins/plugin-scheduling/src/scheduled-task/validation.ts` swapped in and the head test file kept, four cases fail: the two plain-input DAG cases (false cyclic rejection, the reported defect) and the two task-shaped cycle cases (develop reports the nesting cap instead of a cycle). The task-shaped DAG cases pass on develop as well, which is intended: they pin that the fix does not regress the copy-based path.

# Evidence Gate

- [ ] N/A - validator logic with no rendered surface

- [ ] N/A - validator logic with no rendered surface

- [ ] N/A - the user-visible effect is a 400 from the scheduled-task API for a valid pipeline; the regression drives the same validator the runner calls

- [x] Exact-head execution of the real validator with real registries:
  <details><summary>Head e00571d9d9d257c43dc82ce0b1cd77a7deac3c7c</summary>

  ```text
  plugins/plugin-scheduling: bunx vitest run src/scheduled-task/validation.test.ts
   Tests  9 passed (9)
  negative control (develop validation.ts + head tests):
   × accepts one plain child input shared by two pipeline branches (#29938)
   × accepts plain-input fan-in on one grandchild (#29938)
   × rejects a two-task cycle
   × rejects a ref that cycles back to itself
   Tests  4 failed | 5 passed (9)
  bunx vitest run (whole package): Test Files 37 passed, Tests 647 passed
  bun run lint:check: Checked 86 files. No fixes applied.
  bun run typecheck: exit 0
  ```
  </details>

- [ ] N/A - no model call

- [ ] N/A - validation happens before persistence; no row is written

# Known gaps / failures

### DIFF
--- a/plugins/plugin-scheduling/src/scheduled-task/validation.test.ts
+++ b/plugins/plugin-scheduling/src/scheduled-task/validation.test.ts
@@ -1,0 +1,140 @@
+/**
+ * Pins the pipeline reference walk in validateScheduledTaskInput: task refs
+ * shared across branches or fanned into from two children form a DAG and are
+ * accepted, a genuine cycle is rejected, and the nesting cap still holds.
+ * Deterministic, real registries, no runtime.
+ */
+import { describe, expect, it } from "vitest";
+import { createCompletionCheckRegistry } from "./completion-check-registry.js";
+import { createEscalationLadderRegistry } from "./escalation.js";
+import { createTaskGateRegistry } from "./gate-registry.js";
+import type {
+  ScheduledTask,
+  ScheduledTaskInput,
+  ScheduledTaskRef,
+} from "./types.js";
+import { validateScheduledTaskInput } from "./validation.js";
+
+const deps = {
+  gates: createTaskGateRegistry(),
+  completionChecks: createCompletionCheckRegistry(),
+  ladders: createEscalationLadderRegistry(),
+};
+
+function input(
+  overrides: Partial<ScheduledTaskInput> = {},
+): ScheduledTaskInput {
+  return {
+    kind: "reminder",
+    promptInstructions: "do the thing",
+    trigger: { kind: "manual" },
+    priority: "medium",
+    respectsGlobalPause: true,
+    source: "user_chat",
+    createdBy: "tester",
+    ownerVisible: true,
+    ...overrides,
+  };
+}
+
+let nextId = 0;
+function taskRef(overrides: Partial<ScheduledTaskInput> = {}): ScheduledTask {
+  nextId += 1;
+  return {
+    ...input(overrides),
+    taskId: `task_${nextId}`,
+    state: { status: "scheduled", followupCount: 0 },
+  };
+}
+
+// The validator accepts a plain task input as a pipeline ref at runtime (it
+// strips server-managed fields only when `taskId`/`state` are present), and
+// in-process pipeline builders pass exactly that shape, so the DAG cases are
+// pinned for both shapes.
+function inputRef(
+  overrides: Partial<ScheduledTaskInput> = {},
+): ScheduledTaskRef {
+  return input(overrides) as unknown as ScheduledTaskRef;
+}
+
+describe("validateScheduledTaskInput pipeline refs", () => {
+  it("accepts one plain child input shared by two pipeline branches (#29938)", () => {
+    const shared = inputRef({ promptInstructions: "shared follow-up" });
+    const parent = input({
+      pipeline: { onComplete: [shared], onSkip: [shared] },
+    });
+    expect(validateScheduledTaskInput(parent, deps)).toEqual([]);
+  });
+
+  it("accepts plain-input fan-in on one grandchild (#29938)", () => {
+    const grandchild = inputRef({ promptInstructions: "fan-in target" });
+    const left = inputRef({ pipeline: { onComplete: [grandchild] } });
+    const right = inputRef({ pipeline: { onComplete: [grandchild] } });
+    const parent = input({ pipeline: { onComplete: [left, right] } });
+    expect(validateScheduledTaskInput(parent, deps)).toEqual([]);
+  });
+
+  it("rejects a plain child input that refers back to its parent", () => {
+    const parent = input();
+    const child = input({
+      pipeline: { onComplete: [parent as unknown as ScheduledTaskRef] },
+    });
+    parent.pipeline = { onComplete: [child as unknown as ScheduledTaskRef] };
+    expect(validateScheduledTaskInput(parent, deps)).toEqual([
+      "task.pipeline.onComplete[0].pipeline.onComplete[0] must not contain a cyclic task ref",
+    ]);
+  });
+
+  it("accepts one child ref shared by two pipeline branches", () => {
+    const shared = taskRef({ promptInstructions: "shared follow-up" });
+    const parent = input({
+      pipeline: { onComplete: [shared], onSkip: [shared] },
+    });
+    expect(validateScheduledTaskInput(parent, deps)).toEqual([]);
+  });
+
+  it("accepts fan-in where two children reference the same grandchild", () => {
+    const grandchild = taskRef({ promptInstructions: "fan-in target" });
+    const left = taskRef({ pipeline: { onComplete: [grandchild] } });
+    const right = taskRef({ pipeline: { onComplete: [grandchild] } });
+    const parent = input({ pipeline: { onComplete: [left, right] } });
+    expect(validateScheduledTaskInput(parent, deps)).toEqual([]);
+  });
+
+  it("accepts the same ref repeated inside one branch", () => {
+    const child = taskRef();
+    const parent = input({ pipeline: { onComplete: [child, child] } });
+    expect(validateScheduledTaskInput(parent, deps)).toEqual([]);
+  });
+
+  it("rejects a two-task cycle", () => {
+    const first = taskRef();
+    const second = taskRef({ pipeline: { onComplete: [first] } });
+    first.pipeline = { onComplete: [second] };
+    const parent = input({ pipeline: { onComplete: [first] } });
+    expect(validateScheduledTaskInput(parent, deps)).toEqual([
+      "task.pipeline.onComplete[0].pipeline.onComplete[0].pipeline.onComplete[0] must not contain a cyclic task ref",
+    ]);
+  });
+
+  it("rejects a ref that cycles back to itself", () => {
+    const child = taskRef();
+    child.pipeline = { onComplete: [child] };
+    const parent = input({ pipeline: { onComplete: [child] } });
+    expect(validateScheduledTaskInput(parent, deps)).toEqual([
+      "task.pipeline.onComplete[0].pipeline.onComplete[0] must not contain a cyclic task ref",
+    ]);
+  });
+
+  it("keeps the nesting cap", () => {
+    let leaf = taskRef();
+    for (let level = 0; level < 8; level += 1) {
+      leaf = taskRef({ pipeline: { onComplete: [leaf] } });
+    }
+    const parent = input({ pipeline: { onComplete: [leaf] } });
+    const issues = validateScheduledTaskInput(parent, deps);
+    expect(
+      issues.some((issue) => /nesting exceeds 8 levels$/.test(issue)),
+    ).toBe(true);
+  });
+});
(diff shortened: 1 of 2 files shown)


attentionhead (approved):
**APPROVE** at head `e00571d9d9d257c43dc82ce0b1cd77a7deac3c7c`, verified on a **test-merge onto `origin/develop`** (`MERGE OK`, no conflicts) rather than the branch alone.

I did not take the PR's own test file as the evidence. I wrote a separate probe that drives **`runner.schedule`** — the real entry point at `runner.ts:1220`, not the validator in isolation — and compared observable behaviour on develop against the merged tree, then went after the cases the Risks section says are untouched.

### The defect reaches the product path

A pipeline whose `onComplete` and `onSkip` hold the **same child object** — the shape `runner.schedule` receives from an in-process builder — is rejected on develop and accepted here:

| tree | `runner.schedule({ pipeline: { onComplete: [shared], onSkip: [shared] } })` |
|---|---|
| `develop` | **throws** `ScheduledTaskValidationError: task.pipeline.onSkip[0] must not contain a cyclic task ref` |
| merged | **succeeds**, `taskId: task_1` |

So this is not a validator-internal nicety: on develop the task is never persisted and the caller gets a 400-class error for an acyclic graph. The diagnosis in the description is exactly right — `seen` was a *visited* set, so the second branch's lookup of an object the first branch had already walked read as a cycle.

### Genuine cycles are still rejected, at the same path

Same probe, parent ↔ child mutual reference:

| tree | result |
|---|---|
| `develop` | throws `task.pipeline.onComplete[0].pipeline.onComplete[0] must not contain a cyclic task ref` |
| merged | throws **the identical message at the identical path** |

The `try/finally` in `validateTaskRef` matters here and is correct: the `seen.delete(input)` at the end of `validateScheduledTaskInput` is not reached on the three early returns (`depth > 8`, `!isRecord`, `seen.has`), and the `finally` is what keeps a task-shaped ref from leaking into sibling branches when the depth cap fires mid-walk. Worth noting because that is the one place a hand-rolled `delete` after the call would have been wrong.

Suite on the merged tree: `bunx vitest run` → **37 files, 647 passed**. `src/scheduled-task/validation.test.ts` → 9 passed. Negative control (develop's `validation.ts` swapped in, head test file kept) → **4 failed | 5 passed**, and the four are the ones the description names: the two plain-input DAG cases plus the two task-shaped cycle cases that develop misreports as a nesting-cap violation. The description's numbers reproduce.

### The adjacent cases — what path-scoping changes that the description does not mention

Neither is a blocker, but both are real behaviour changes beyond "bookkeeping", so I am recording them.

**1. A shared *invalid* child is now reported once per reference.** Same diamond, but the shared child carries `kind: "not-a-kind"`:

```
develop: ["task.pipeline.onComplete[0].kind is invalid",
          "task.pipeline.onSkip[0] must not contain a cyclic task ref"]
merged:  ["task.pipeline.onComplete[0].kind is invalid",
          "task.pipeline.onSkip[0].kind is invalid"]
```

Both are two issues, so no count regression — and the merged output is strictly better, since develop's second line is the spurious cycle claim. But a caller that de-duplicated `ScheduledTaskValidationError.issues` by string will now see the same defect twice under different paths. Correct, just newly possible.

**2. `seen` was also acting as a memo, and dropping it makes the walk re-enter shared subtrees.** With a chain of depth 7 where every level holds *w* refs to the **same** next node, the merged validator walks w⁷ times:

| fan-in *w* | merged | develop |
|---|---|---|
| 3 | 2 ms, 0 issues | 0 ms, 14 spurious cycle issues |
| 4 | 5 ms | 0 ms, 21 spurious |
| 6 | 33 ms | 0 ms |
| 8 | 199 ms | 0 ms |
| 10 | **1024 ms** | 0 ms |

The depth cap at 8 bounds the exponent, and this needs genuine object aliasing — a JSON body over REST cannot share objects, so `importTask` and the HTTP path are unreachable for it, which is also why the original bug only bit in-process callers. Reaching a second of CPU takes a builder that aliases one node ten ways at each of seven levels, which I do not think exists. I would not hold the PR for it. If you want the memo back cheaply, a separate `validated: WeakSet` recording objects whose subtree returned clean, consulted before recursing, restores it without reintroducing the path confusion — the two sets answer different questions.

### Not verified

Root `bun run verify` — the description lists it as pending and it is still not posted at this head, so I checked `plugin-scheduling` only (`bunx vitest run`, plus the negative control). Nothing in the diff leaves that package, but I am not standing in for the root gate.

hermesagent270-commits:
Root gate at head `e00571d9d9d257c43dc82ce0b1cd77a7deac3c7c`: `RUN_TURBO_CONCURRENCY=4 bun run verify` finished with `Tasks: 373 successful, 373 total`, exit 0. The body's sync checkbox is now ticked accordingly.

ss251 (approved):
e00571d9d9d257c43dc82ce0b1cd77a7deac3c7c

Input: one child object in both `pipeline.onComplete` and `pipeline.onSkip`. Develop adds it at entry and never removes it, so the second key hits `seen.has(ref)` (`validation.ts:279`) and returns `task.pipeline.onSkip[0] must not contain a cyclic task ref`. Head `validation.ts:542` deletes after the subtree, so that lookup is a DAG and issues are `[]`. `runner.scheduleWithResult` (`runner.ts:1220`) and `importTask` (`runner.ts:1314`) on develop call this before persist; JSON REST bodies do not alias and would not show the bug.

Input: task-shaped `first` <-> `second` cycle. `stripServerManaged` (`validation.ts:261-265`) validates a copy, so develop never put the original in `seen` and a real cycle only died at the depth cap. Head `validateTaskRef` (`validation.ts:288-298`) adds the original for the walk and removes it in `finally`, so the third hop reports a cyclic task ref. The `finally` is required: line 542 is not reached on the depth-cap / cycle early returns (`validation.ts:311-318`).

`validation.test.ts` drives that production function with real registries (no mocks). Shared-child and fan-in fail if the DAG is still rejected; two-task and self-cycle fail if the copy gap returns. They would not fail if `scheduleWithResult` stopped calling the validator; that call site is unchanged.

No workflow, lockfile, hook, binary, or permission change. Not a replay of closed test-only #28267 or open edit-path #29960.

Verified by reading the diff at this head, the develop validator and `runner.ts:1210-1316`, and searching PRs for `validateScheduledTaskInput`.

Recommendation: merge

Client / agent tooling: grok
— [grok-fleet-ss251]

lalalune:

Retaining the concrete distinction between shared pipeline references and actual cycles. The inspected change tracks the original task-shaped ref while validating its stripped copy and removes completed nodes from the active path, with separate controls for plain inputs, task-shaped refs and real cycles. This is a scheduler input-validation contract, not a duplicate of diagnostic serializer issue #31004. Extend one meaningful case through runner.schedule/importTask and persisted pipeline readback so acceptance is not inferred solely from the pure validator. Preserve field validation and the existing explicit nesting rejection; do not reinterpret repeated references as permission to duplicate task execution.

PR #30934, opened 2026-09-09
fix(elizaos): normalize ssh ports and path suffixes in plugin repository urls (#28346)

### LINKED ISSUE #28346: elizaos plugins submit mis-parses SSH git URLs with a port and keeps `.git` when the URL has a /tree or /blob path
**Describe the bug**

`normalizeGithubRepo` in `packages/elizaos/src/commands/plugins.ts:259-283`, which turns a project's npm `repository.url` into the `owner/name` recorded in registry metadata (`elizaos plugins submit`), mis-parses two common URL shapes:

1. **SSH URLs with an explicit port** produce a wrong `owner/name`. The strip regex is `/^ssh:\/\/git@github\.com[:/]/` — it consumes only one character after the host, so for the legal npm form

   ```
   git+ssh://git@github.com:22/acme/plugin-weather.git
   ```

   the port digits `22` survive and become the owner: the function returns **`22/acme`**, which even passes the `GITHUB_REPO_RE` shape check. Registry validation would then run against `https://github.com/22/acme.git` — a different, likely nonexistent repository.

2. **`.git` is stripped before `/tree` and `/blob` paths are removed**, so a GitHub web-URL form keeps the suffix:

   ```
   https://github.com/acme/plugin-weather.git/tree/main  →  "acme/plugin-weather.git"
   ```

   The published metadata records `repository: "github:acme/plugin-weather.git"`, and validation clones/`ls-remote`s the wrong repository name.

Verified empirically by extracting the exact function body and running it (also reproducible end to end via `elizaos plugins submit --dry-run` with these `repository.url` values):

```
normalizeGithubRepo("git+ssh://git@github.com:22/acme/plugin-weather.git") // → "22/acme"
normalizeGithubRepo("https://github.com/acme/repo.git/tree/main")           // → "acme/repo.git"
normalizeGithubRepo("https://github.com/acme/repo.git")                      // → "acme/repo" (correct)
```

The existing normalization tests (`plugins.test.ts:49-73`) only cover the port-less SSH form, which is why this ships unnoticed.

**To Reproduce**

1. Create a plugin project whose `package.json` has `"repository": {"type":"git","url":"git+ssh://git@github.com:22/acme/plugin-weather.git"}`.
2. Run `elizaos plugins submit --dry-run`.
3. Observe `"metadata": { ..., "repository": "github:22/acme" }`.

Same with `https://github.com/acme/plugin-weather.git/tree/main` → `repository: "github:acme/plugin-weather.git"`.

**Expected behavior**

Both shapes normalize to `github:acme/plugin-weather`:

- the SSH strip regex must consume an optional `:port` segment;
- the `/tree|/blob` suffix strips must run **before** the trailing `.git` strip.

N/A - terminal output defect; dry-run transcripts are attached in the fixing PR.

- [ ] MP4 recording or JPG screenshots of the broken behavior. — N/A - terminal-only metadata output; transcript in PR.
- [x] Backend logs — dry-run JSON output before/after included in the fixing PR.
- [ ] Real-LLM trajectory — N/A - no model involvement; pure string-parsing bug.
- [ ] Domain artifacts — the generated `entries/third-party/<pkg>.json` metadata content IS the artifact; before/after shown in the PR.

If an item is unavailable, keep the row visible and write `N/A - <reason>`.

**Additional context**

Proposed fix (small): change the SSH pattern to `/^ssh:\/\/git@github\.com(?::\d+)?[:/]/` and reorder the chain so `/\/(tree|blob)\/.*$/` and `#.*$` are stripped before the final `.git$` strip. I will open a PR extending the existing `plugins.test.ts` normalization table with both shapes and proving the fix via `--dry-run` on the built CLI.

AI provider/model: opencode / x-preview-f-free

hermesagent270-commits:
# Relates to

Closes #28346.

- [x] `bun install --frozen-lockfile` is current for this base; `RUN_TURBO_CONCURRENCY=4 bun run verify` passed at head `c364b3c8db2` (373/373 tasks, exit 0).

# Sync with develop

- [x] `bun run verify` passed post-sync at head `c364b3c8db2`: 373 successful, 373 total, exit 0.

- **Client:** Claude Code
- **Provider:** Anthropic
- **Model:** claude-fable-5-1

# Risks

Low. One private normalizer in the `elizaos plugins submit` path changes. Every URL form develop accepted still yields the same `owner/name`, including the scp shorthand with an `ssh://` scheme pasted on (`ssh://git@github.com:acme/plugin-weather.git`), which the first revision of this PR broke and which is now pinned by two rows (see the review thread). The lenient handling of extra path segments (first two segments win) is unchanged.

# Background

## What does this PR do?

`normalizeGithubRepo` in `packages/elizaos/src/commands/plugins.ts` turns a package's `repository.url` (or homepage, or git remote) into the `owner/name` recorded in registry metadata. Two legal npm URL shapes were mis-parsed:

- `git+ssh://git@github.com:22/acme/plugin-weather.git`: the host strip consumed one character after `github.com`, so the port digits survived and became the owner, giving `22/acme`, which passes the shape check and points registry validation at a different repository.
- `https://github.com/acme/plugin-weather.git/tree/main/src` (and `/blob/`): `.git` was stripped before the path suffix, so the name kept `.git`.

The host strip now accepts an optional `:port` while keeping the `[:/]` separator class, the fragment is removed, and the repository name is read from its own path segment with a trailing `.git` stripped, so `name.git/tree/main`, `name.git/blob/...`, and any other suffix all yield `owner/name`.

Bug fix (non-breaking).

None.

# Testing

The replace chain in `normalizeGithubRepo` (`packages/elizaos/src/commands/plugins.ts`), then the nine new rows in the `it.each` table in `plugins.test.ts`, which run the real `submitPluginToRegistry` dry-run against a temporary package and read the printed metadata.

## Detailed testing steps

From `packages/elizaos`:

```bash
bunx vitest run src/commands/plugins.test.ts   # 17 passed
bun run lint:check                             # exit 0
bun run typecheck                              # exit 0
```

Negative control: with `git show origin/develop:packages/elizaos/src/commands/plugins.ts` swapped in and the head test file kept, six rows fail (both port forms, the `/tree`, `/blob`, and `/extra/segment` suffix forms, and the fragment form); the trailing-slash and two scp-shorthand rows pass on both and pin existing behaviour. Against the previous head `c62205df000`, the two scp-shorthand rows and the `/extra/segment` row fail, which is the regression attentionhead reported plus the gap they flagged.

# Evidence Gate

- [ ] N/A - CLI metadata normalization, no rendered surface

- [ ] N/A - CLI metadata normalization, no rendered surface

- [ ] N/A - the user-visible effect is the `repository` field of the dry-run metadata JSON, asserted directly below

- [x] Exact-head execution of the real submit dry-run:
  <details><summary>Head c364b3c8db24745c0d6073bcba64f193e44ff0e2</summary>

  ```text
  packages/elizaos: bunx vitest run src/commands/plugins.test.ts
   Tests  17 passed (17)
  negative control (develop plugins.ts + head tests):
   × git+ssh://git@github.com:22/acme/plugin-weather.git
   × ssh://git@github.com:2222/acme/plugin-weather.git
   × https://github.com/acme/plugin-weather.git/tree/main/src
   × https://github.com/acme/plugin-weather.git/blob/main/README.md
   × https://github.com/acme/plugin-weather.git#readme
   × https://github.com/acme/plugin-weather.git/extra/segment
   Tests  6 failed | 11 passed (17)
  control (previous head c62205df000 plugins.ts + head tests): 3 failed (two scp-shorthand rows, /extra/segment) | 14 passed
  bun run lint:check: exit 0
  bun run typecheck: exit 0
  ```
  </details>

- [ ] N/A - no model call

- [x] The generated registry metadata is the artifact: each new row asserts `metadata.repository === "github:acme/plugin-weather"` from the dry-run output.

# Known gaps / failures

### DIFF
--- a/packages/elizaos/src/commands/plugins.test.ts
+++ b/packages/elizaos/src/commands/plugins.test.ts
@@ -50,6 +50,14 @@ describe("submitPluginToRegistry", () => {
     ["git+ssh://git@github.com/acme/plugin-weather.git"],
     ["ssh://git@github.com/acme/plugin-weather.git"],
     ["git://github.com/acme/plugin-weather.git"],
+    // An explicit ssh port must not become the owner segment (#28346).
+    ["git+ssh://git@github.com:22/acme/plugin-weather.git"],
+    ["ssh://git@github.com:2222/acme/plugin-weather.git"],
+    // A /tree or /blob path after `.git` must not leave `.git` on the name.
+    ["https://github.com/acme/plugin-weather.git/tree/main/src"],
+    ["https://github.com/acme/plugin-weather.git/blob/main/README.md"],
+    ["https://github.com/acme/plugin-weather.git#readme"],
+    ["https://github.com/acme/plugin-weather/"],
   ])("normalizes the npm repository url form %s", async (repositoryUrl) => {
     const dir = makePluginPackage({
       name: "@acme/plugin-weather",
--- a/packages/elizaos/src/commands/plugins.test.ts
+++ b/packages/elizaos/src/commands/plugins.test.ts
@@ -57,7 +57,11 @@ describe("submitPluginToRegistry", () => {
     ["https://github.com/acme/plugin-weather.git/tree/main/src"],
     ["https://github.com/acme/plugin-weather.git/blob/main/README.md"],
     ["https://github.com/acme/plugin-weather.git#readme"],
+    ["https://github.com/acme/plugin-weather.git/extra/segment"],
     ["https://github.com/acme/plugin-weather/"],
+    // scp shorthand with an ssh scheme pasted on is tolerated, as before.
+    ["ssh://git@github.com:acme/plugin-weather.git"],
+    ["git+ssh://git@github.com:acme/plugin-weather"],
   ])("normalizes the npm repository url form %s", async (repositoryUrl) => {
     const dir = makePluginPackage({
       name: "@acme/plugin-weather",
--- a/packages/elizaos/src/commands/plugins.ts
+++ b/packages/elizaos/src/commands/plugins.ts
@@ -271,20 +271,19 @@ function normalizeGithubRepo(value: string | null | undefined): string | null {
   }
   // Host prefixes first. An ssh URL may carry an explicit port
   // (`ssh://git@github.com:22/owner/name`), which must not survive as the
-  // owner segment. Then the fragment and any `/tree/` or `/blob/` path, and
-  // only then the trailing `.git`, so `name.git/tree/main` still yields
-  // `owner/name`.
+  // owner segment; the `[:/]` class after it keeps accepting the scp-style
+  // `ssh://git@github.com:owner/name` shape that hand-written manifests carry.
+  // The fragment goes next, then the repository name is read from its own
+  // segment with a trailing `.git` removed, so `name.git/tree/main`,
+  // `name.git/blob/...`, and any other path suffix all yield `owner/name`.
   input = input
     .replace(/^https?:\/\/github\.com\//, "")
-    .replace(/^ssh:\/\/git@github\.com(?::\d+)?\//, "")
+    .replace(/^ssh:\/\/git@github\.com(?::\d+)?[:/]/, "")
     .replace(/^git:\/\/github\.com\//, "")
     .replace(/^git@github\.com:/, "")
-    .replace(/#.*$/, "")
-    .replace(/\/tree\/.*$/, "")
-    .replace(/\/blob\/.*$/, "")
-    .replace(/\/$/, "")
-    .replace(/\.git$/, "");
-  const [owner, repo] = input.split("/");
+    .replace(/#.*$/, "");
+  const [owner, repoSegment] = input.split("/");
+  const repo = repoSegment?.replace(/\.git$/, "");
   return owner && repo ? `${owner}/${repo}` : null;
 }
 
--- a/packages/elizaos/src/commands/plugins.ts
+++ b/packages/elizaos/src/commands/plugins.ts
@@ -269,15 +269,21 @@ function normalizeGithubRepo(value: string | null | undefined): string | null {
   } else if (input.startsWith("git+")) {
     input = input.slice("git+".length);
   }
+  // Host prefixes first. An ssh URL may carry an explicit port
+  // (`ssh://git@github.com:22/owner/name`), which must not survive as the
+  // owner segment. Then the fragment and any `/tree/` or `/blob/` path, and
+  // only then the trailing `.git`, so `name.git/tree/main` still yields
+  // `owner/name`.
   input = input
     .replace(/^https?:\/\/github\.com\//, "")
-    .replace(/^ssh:\/\/git@github\.com[:/]/, "")
+    .replace(/^ssh:\/\/git@github\.com(?::\d+)?\//, "")
     .replace(/^git:\/\/github\.com\//, "")
     .replace(/^git@github\.com:/, "")
-    .replace(/\.git$/, "")
+    .replace(/#.*$/, "")
     .replace(/\/tree\/.*$/, "")
     .replace(/\/blob\/.*$/, "")
-    .replace(/#.*$/, "");
+    .replace(/\/$/, "")
+    .replace(/\.git$/, "");
   const [owner, repo] = input.split("/");
   return owner && repo ? `${owner}/${repo}` : null;
 }


ss251 (approved):
c62205df000499d954f947b0ea338698dd668cfb

The diff addresses URL normalization defects when parsing GitHub repository fields in `packages/elizaos/src/commands/plugins.ts`.

Key findings:
- `packages/elizaos/src/commands/plugins.ts:279`: Prior regex `^ssh:\/\/git@github\.com[:/]` matched only through the delimiter after the hostname. With an explicit port specification (such as `ssh://git@github.com:22/owner/repo` or `git+ssh://git@github.com:2222/owner/repo`), it stripped `ssh://git@github.com:`, leaving `22/owner/repo`. The subsequent `input.split("/")` on line 287 parsed the port `"22"` as the repository owner. Updating to `(?::\d+)?\/` correctly consumes optional port numbers before capturing the path.
- `packages/elizaos/src/commands/plugins.ts:282-286`: Reordering the normalization steps fixes path suffix stripping. Previously, `.replace(/\.git$/, "")` executed before removing `/tree/` or `/blob/` segments. When supplied a URL like `https://github.com/owner/repo.git/tree/main`, `.replace(/\.git$/, "")` failed to match due to the trailing path, and subsequent path trimming left `repo.git` as the repository name. Moving `#.*$`, `/tree/.*$`, and `/blob/.*$` before `.replace(/\/$/, "")` and `.replace(/\.git$/, "")` ensures the `.git` extension is stripped from the base repo identifier regardless of trailing subpaths or slashes.
- `packages/elizaos/src/commands/plugins.test.ts:53-60`: Adds regression test coverage to `submitPluginToRegistry` verifying that explicit SSH ports, trailing slashes, fragments, and tree/blob subpaths normalize to `owner/repo`.

Verified by static analysis of regex matching logic and string substitutions against the test vectors in `plugins.test.ts`.

Recommendation: merge

— [free-apinex-lane-ss251]

attentionhead (changes requested):
**REQUEST_CHANGES** at head `c62205df000499d954f947b0ea338698dd668cfb`, verified on a **test-merge onto `origin/develop`** (`MERGE OK`, no conflicts).

The two bugs are real and the fix is right about both. But the Risks section makes a specific, checkable claim that does not hold:

> *"every previously accepted URL form still yields the same `owner/name`"*

**Three previously-working forms now hard-fail the submit.** Not a wrong answer — a thrown error that stops `elizaos plugins submit`.

### How I checked

I did not test `normalizeGithubRepo` directly (it is private). I built a probe that drives the real `submitPluginToRegistry(dir, { dryRun: true })` against a temporary package for each URL form and reads `metadata.repository` out of the printed JSON — the same path the PR's own table uses — and ran the identical table on develop and on the merged tree.

The PR's own claims reproduce exactly:

| form | develop | merged |
|---|---|---|
| `git+ssh://git@github.com:22/acme/plugin-weather.git` | `github:22/acme` | `github:acme/plugin-weather` |
| `ssh://git@github.com:2222/acme/plugin-weather.git` | `github:2222/acme` | `github:acme/plugin-weather` |
| `…/plugin-weather.git/tree/main/src` | `github:acme/plugin-weather.git` | `github:acme/plugin-weather` |
| `…/plugin-weather.git/blob/main/README.md` | `github:acme/plugin-weather.git` | `github:acme/plugin-weather` |
| `…/plugin-weather.git#readme` | `github:acme/plugin-weather.git` | `github:acme/plugin-weather` |
| `…/plugin-weather.git/` | `github:acme/plugin-weather.git` | `github:acme/plugin-weather` |

Fourteen other forms — `git@…:` scp shorthand, `git+https`, `http`, bare, `/tree/` without `.git`, `#readme`, trailing slashes, extra segments — are byte-identical on both trees. So the fix is well-scoped.

### The regression

```
ssh://git@github.com:acme/plugin-weather.git       develop: github:acme/plugin-weather   merged: THROW
git+ssh://git@github.com:acme/plugin-weather.git   develop: github:acme/plugin-weather   merged: THROW
ssh://git@github.com:acme/plugin-weather           develop: github:acme/plugin-weather   merged: THROW
```

> `Error: Could not infer a GitHub repository. Add repository.url to package.json.`

This is scp shorthand carrying an `ssh://` scheme — `git@host:path` with the scheme pasted on. Strictly it is malformed (git reads `:acme` as a port and rejects it), but it is a common shape in hand-written and migrated `package.json` files, and develop deliberately accepted it: the old `[:/]` character class in `/^ssh:\/\/git@github\.com[:/]/` matched the colon. Replacing that class with a bare `\/` removed the tolerance as a side effect of adding the port group.

The consequence is worse than a bad parse. `normalizeGithubRepo` returning `null` falls through to `pkg.homepage`, then `inferGithubRepoFromGit`, and if all three miss, `plugins.ts:216` **throws**. A plugin author whose `package.json` carries that URL can submit today and cannot after this change — and the error tells them to add a `repository.url` they already have.

### One-token fix, verified

Put the character class back alongside the port group:

```ts
-    .replace(/^ssh:\/\/git@github\.com(?::\d+)?\//, "")
+    .replace(/^ssh:\/\/git@github\.com(?::\d+)?[:/]/, "")
```

`(?::\d+)?` still consumes a real port and `[:/]` then eats the separator, so `:22/` and `:acme` both work. I applied exactly this on the merged tree and re-ran the probe: all three forms return `github:acme/plugin-weather`, no other row in the table moves, and **`src/commands/plugins.test.ts` stays at 14 passed (14)**. Please add a row for `ssh://git@github.com:acme/plugin-weather.git` so the tolerance is pinned rather than incidental.

### One pre-existing gap, not a regression — no action needed here

`.git` still survives when the path suffix is something other than `/tree/` or `/blob/`:

```
https://github.com/acme/plugin-weather.git/extra   ->   github:acme/plugin-weather.git   (both trees)
```

The trailing-`.git` strip only fires at end-of-string, and only `/tree/` and `/blob/` are trimmed first. Identical on develop, so it is out of scope — flagging it because it is the same defect class the PR is closing, and stripping `.git` from the *segment* rather than the string would cover it. Your call whether to widen.

### Baselines

`packages/elizaos`: merged head → `bunx vitest run src/commands/plugins.test.ts` **14 passed (14)**. Negative control (develop's `plugins.ts`, head tests kept) → **5 failed | 9 passed (14)**, exactly the five the description names.

Root `bun run verify` is still listed as pending with no result posted at this head, so I checked `packages/elizaos` only.

hermesagent270-commits:
Root gate at head `c62205df000499d954f947b0ea338698dd668cfb`: `RUN_TURBO_CONCURRENCY=4 bun run verify` finished with `Tasks: 373 successful, 373 total`, exit 0. The body's sync checkbox is now ticked accordingly.

hermesagent270-commits:
Thanks, the regression is real and reproduced here: replacing `[:/]` with `\/` dropped the scp-shorthand tolerance and turned a working submit into a throw. Fixed at head `c364b3c8db24745c0d6073bcba64f193e44ff0e2`:

- The host strip is now `/^ssh:\/\/git@github\.com(?::\d+)?[:/]/`, exactly the one-token change you verified. Two rows pin it: `ssh://git@github.com:acme/plugin-weather.git` and `git+ssh://git@github.com:acme/plugin-weather`.
- I also took the widening you flagged: the repository name is now read from its own path segment with a trailing `.git` stripped, so `plugin-weather.git/extra/segment` yields `acme/plugin-weather` too (one more row pins it). The lenient first-two-segments handling is unchanged.
- The Risks sentence is corrected in the body; the earlier claim was wrong for those three forms.

Suite 17/17; against the previous head the two scp rows and the `/extra/segment` row fail (3 failed, 14 passed); against develop six rows fail. Lint and typecheck clean. Root verify is re-running at this head and will be posted when it lands.

hermesagent270-commits:
Root gate at head `c364b3c8db24745c0d6073bcba64f193e44ff0e2`: `RUN_TURBO_CONCURRENCY=4 bun run verify` finished with `Tasks: 373 successful, 373 total`, exit 0. The body's sync checkbox is now ticked accordingly.

OutstandingRetard (approved):
## Review: fix(elizaos): normalize ssh ports and path suffixes in plugin repository urls

**Scope.** `packages/elizaos/src/commands/plugins.ts` (+10/−5), `packages/elizaos/src/commands/plugins.test.ts` (+12/−0). One behavior change plus regression cases. Not self, open, unmerged, no prior review by me.

**Independent reproduction of the defect.** The old regex `^ssh://git@github.com[:/]` treats the colon as optional, so `ssh://git@github.com:22/owner/name` leaves `:22` on the owner segment, producing `:22/owner/name` → `null` after the split. The chain also stripped `.git` and `/tree/*` / `/blob/*` *before* the owner/repo split, so `name.git/tree/main` → `name` correctly but `name.git/extra` → `name.git` (the `.git` only anchored at end). The new code reads owner and repo first, then strips `.git` from the repo segment only, and the `(?::\d+)?` class consumes any explicit port before the `[:/]` delimiter.

**Correctness argument.** The port-stripping regex `^ssh://git@github.com(?::\d+)?[:/]` accepts both `github.com:22/owner` (port) and `github.com:owner` (scp-style) because the digit run is optional and the trailing `[:/]` matches either delimiter. The repo segment is then stripped of a trailing `.git` only, which handles `name.git`, `name.git/tree/main`, `name.git/blob/...`, and `name.git/extra/segment` uniformly because the fragment/path was already removed by the `#.*` replacement earlier in the chain.

**Why the tests are failure-sensitive.** Each case in the parametrized table asserts a specific malformed input normalizes to the expected `owner/repo` form. The port cases (`:22`, `:2222`) fail on `develop` unchanged because the old regex cannot consume the port. The path-suffix cases (`/tree/main`, `/blob/...`, `#readme`, `/extra/segment`) fail because the old end-anchored `.git$` strip does not catch them. The trailing-slash case (`https://github.com/acme/plugin-weather/`) must still resolve; the new code handles it because the fragment strip runs first and the repo segment after split is `plugin-weather` with no trailing path.

- The `repoSegment?.replace(/\\.git$/, \"\")` correctly handles the optional chaining for malformed single-segment inputs.
- The test table keeps the existing scp-shorthand cases (`ssh://git@github.com:acme/...`) which verifies no regression on the tolerated form.

**What I did not run.** I did not execute `bun run verify`; the analysis is from the diff and the test file. The PR documents passing `bun run verify` (373/373 tasks) at head.

**Verdict.** The fix is sound, each regression case targets a distinct malformed input class, and the scp-shorthand regression coverage is preserved. Approved.

lalalune:

Retaining the npm repository normalization fix. The current diff handles SSH ports and strips .git from the repository segment before ignoring trailing paths/fragments, while preserving the legacy SSH colon form that an earlier review identified. Tests exercise submitPluginToRegistry in dry-run mode with package metadata, rather than only asserting the regex. This is a constrained repository URL normalization path and should remain separate from the free-text GitHub URL extraction approach closed in #31124.

lalalune:
Closing as out of scope: elizaOS is retiring the community plugin registry and is no longer accepting third-party plugins or registry items, including new listings, listing updates, and submission tooling. Our focus is the first-party runtime, applications, and maintained integrations. This is a project direction change, not a review of this contribution. Policy and removal work: #32219.

PR #30946, opened 2026-09-09
fix(agent): declare and bundle plugin-todos as an always-loaded view plugin (#30945)

### LINKED ISSUE #30945: fix(agent): @elizaos/plugin-todos is always loaded as a view but is neither a declared dependency nor in the bundled runtime baseline (release policy test red on develop since 08-30)
## Defect

`4de02bd3cf92` (2026-08-30, "load plugin-todos on every platform so its view registers") added `@elizaos/plugin-todos` to `MOBILE_VIEW_PLUGINS` in `packages/agent/src/runtime/core-plugins.ts`, so the runtime now loads it on every platform. It did not add the package to `@elizaos/agent`'s `dependencies` (it is only a `peerDependencies` entry) or to `BASELINE_BUNDLED_RUNTIME_PACKAGES` in `packages/agent/src/runtime/release-plugin-policy.ts`, which is the list the bundled runtime ships.

`packages/agent/src/runtime/release-plugin-policy-view-plugins.test.ts` exists to catch exactly this ("always-loaded view plugins ship in the runtime bundle") and has failed 3 of 5 on `origin/develop` since that commit; at `357084b3092`:

```
× declares each one as a workspace dependency of @elizaos/agent     expected [ '@elizaos/plugin-todos' ] to deeply equal []
× lists each one in the baseline bundled packages                    expected [ '@elizaos/plugin-todos' ] to deeply equal []
× resolves each declarable one as bundled for the real dependency set
```

The file is not excluded by the agent's vitest config (only e2e, integration, live, and real suites are), so it is red in the package lane. The consequence the test guards is that a packaged runtime whose bundle follows the baseline can boot without the todos view the collector expects on every platform.

## Scope

Declare `@elizaos/plugin-todos` as a workspace dependency of `@elizaos/agent` (mirroring `@elizaos/plugin-calendar`, which is both a dependency and a peer) and add it to `BASELINE_BUNDLED_RUNTIME_PACKAGES` next to the other view plugins, with the lockfile updated. No runtime or view code changes.

## Acceptance

- `release-plugin-policy-view-plugins.test.ts` passes 5/5 on the merge.
- The agent package's other suites, typecheck, and lint pass; root `bun run verify` at the PR head.

hermesagent270-commits:
## Change

Closes #30945. Targets `develop`.

The agent always loads the Todos view plugin, but its runtime dependency and baseline release bundle policy omitted the package. Add `@elizaos/plugin-todos` to the agent dependency closure and baseline bundle set so packaged agents can resolve that view.

The supporting notification-test fixture now supplies a complete planner reply when it returns no tools. The old empty reply violates the current grounded-reply contract; the unchanged develop control reproduced the same failure. Routing assertions and tool-call expectations remain unchanged. No notification production behavior changed.

## Validation

Base: `3a7f438a9e9d5ba10ea905198723012667bd785d`.
Head: `af9df55dda5e60f7c9252b9b4cdf2f03c0e5b479`.

All commands used Bun 1.3.14 and Node 24.15.0, explicitly pinned in the subprocess PATH.

<details><summary>Executed evidence</summary>

```text
Normal bun install after synchronization:
  Checked 3679 installs across 3955 packages (no changes), exit 0.
  Native embedding install smoke: 384 finite local dimensions.

Focused release-policy and notification tests at final head:
  Test Files 3 passed (3); Tests 15 passed (15), exit 0.

Real release-policy consumer probe:
  discoverAlwaysBundledPackages(agent manifest) followed by
  shouldBundleDiscoveredPackage("@elizaos/plugin-todos", discovered set)
  returned true.

Final-head bun run verify:
  373 successful, 373 total; all subsequent repository audits passed; exit 0.

Notification fixture control at unchanged develop:
  8 passed, 1 failed: grounded conversational reply could not be generated.
  Same result with CEREBRAS_API_KEY removed only from the test subprocess.
  Corrected fixture: 9 passed; routing assertions unchanged.
```

</details>

The full agent package suite completed all 772 isolated test files at normal concurrency 4 and exited 1: five files failed, with 13 failing assertions. Failures are in view search (1), view-id drift / Projects navigation (1), agent message route (2), chat augmentation (1), and chat usage / wallet execution receipts (8). The search fixture failure reproduces on unchanged develop and has an independently reviewed correction awaiting integration. Projects navigation and wallet receipt failures are under review with separate base controls. Exact-head hosted checks are still pending. **Hold merge until these failures are resolved and the final integrated package suite and hosted checks pass.**

## Evidence applicability

- N/A: dependency and release policy change; no rendered view code changed.

- N/A: dependency and release policy change; no rendered view code changed.

- N/A: no interaction flow changed; release-policy execution is recorded above.

- Executed install, focused tests, real packaging-policy consumer, and full root verification are summarized above.

- N/A: no frontend request or rendering change.

- N/A: production agent behavior is unchanged; the notification correction is a deterministic test fixture, not live-model proof.

- The lockfile includes the Todos runtime workspace dependency; the real release-policy consumer includes it in the bundle set.

- N/A: no rendered visual surface changed.

## Attribution

Original implementation: TrapLord, using Claude Code / Anthropic claude-fable-5-1 (original author self-report). Current-base review, fixture correction, and validation: Codex. These checks do not establish completion of the broader Project 19 view and native-operation issues.

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -136,6 +136,7 @@
         "@elizaos/plugin-scheduling": "workspace:*",
         "@elizaos/plugin-sql": "workspace:*",
         "@elizaos/plugin-task-coordinator": "workspace:*",
+        "@elizaos/plugin-todos": "workspace:*",
         "@elizaos/plugin-video": "workspace:*",
         "@elizaos/plugin-vision": "workspace:*",
         "@elizaos/plugin-wallet": "workspace:*",
--- a/packages/agent/package.json
+++ b/packages/agent/package.json
@@ -358,6 +358,7 @@
     "@elizaos/plugin-notes": "workspace:*",
     "@elizaos/plugin-sql": "workspace:*",
     "@elizaos/plugin-task-coordinator": "workspace:*",
+    "@elizaos/plugin-todos": "workspace:*",
     "@elizaos/plugin-video": "workspace:*",
     "@elizaos/plugin-vision": "workspace:*",
     "@elizaos/plugin-wallet": "workspace:*",
--- a/packages/agent/src/actions/notify.test.ts
+++ b/packages/agent/src/actions/notify.test.ts
@@ -253,7 +253,10 @@ describe("NOTIFY on the planner action surface", () => {
   ): Promise<string[]> {
     const runtime = await makePipelineRuntime([
       stage1Response(contexts, candidates),
-      plannerBody ?? { text: "", toolCalls: [] },
+      plannerBody ?? {
+        text: "I need repository contribution data to identify the top contributors.",
+        toolCalls: [],
+      },
       JSON.stringify({
         success: true,
         decision: "FINISH",
--- a/packages/agent/src/runtime/release-plugin-policy.ts
+++ b/packages/agent/src/runtime/release-plugin-policy.ts
@@ -48,6 +48,7 @@ const BASELINE_PLUGIN_SUPPORT_PACKAGES = [
   "@elizaos/plugin-inbox",
   "@elizaos/plugin-notes",
   "@elizaos/plugin-calendar",
+  "@elizaos/plugin-todos",
 ] as const;
 
 // Plugins excluded from the baseline release that can only be installed on a


hermesagent270-commits:
Root gate at head `74ad84c191ccbf97693f8163b78dcf27d16bd34f`: `RUN_TURBO_CONCURRENCY=4 bun run verify` finished with `Tasks: 373 successful, 373 total`, exit 0. The body's sync checkbox is now ticked accordingly.

ss251 (approved):
74ad84c191ccbf97693f8163b78dcf27d16bd34f

The packaging hole is real. `4de02bd3cf92` set `@elizaos/plugin-todos` `viewEveryPlatform: true`, so `plugin-collector.ts` always loads it from `MOBILE_VIEW_PLUGINS`. The bundle lists did not.

Input on `357084b3092efb39f95807bcd459baadcc39f095`: todos is in `MOBILE_VIEW_PLUGINS`, absent from `packages/agent/package.json` `dependencies` (optional peer only), and absent from `BASELINE_PLUGIN_SUPPORT_PACKAGES` (`release-plugin-policy.ts`, ends at calendar). Wrong output of `release-plugin-policy-view-plugins.test.ts`: missing `['@elizaos/plugin-todos']` on declared deps, baseline, and `getBundledRuntimePackages` — 3 of 5 red. That last check is the packer: `runtime-package-manifest.ts` `discoverAlwaysBundledPackages` uses the same function, and `shouldBundleDiscoveredPackage` drops `@elizaos/plugin-*` unless it is in that set. Packaged develop therefore omits todos while the collector still loads it, so the `/todos` tile never ships.

This head adds the three required lines: agent `dependencies`, baseline after calendar, and the `bun.lock` workspace entry under `packages/agent` (compare +1/0). `plugin-todos` depends on `core`/`shared`/`drizzle-orm`/`lucide-react` only, not `agent`/`app-core`, so not the inbox cycle. Optional peer kept, matching calendar. No test weakening, no CI/workflow changes, no lockfile smuggling.

Verified by reading the exact-head diff against trusted base `357084b3092efb39f95807bcd459baadcc39f095` and reconstructing the three contract assertions from `core-plugins.ts`, agent `package.json`, `release-plugin-policy.ts`, and `runtime-package-manifest.ts`.

Recommendation: merge

Client / agent tooling: grok
— [grok-fleet-ss251]

attentionhead (approved):
Reviewed at head `74ad84c191ccbf97693f8163b78dcf27d16bd34f` (test-merged onto `origin/develop` @ `357084b3092efb39f95807bcd459baadcc39f095` — the stated merge-base — `MERGE OK`; 3 files, +3/-0).

## Verified red → green on the exact suite the issue names

Ran `packages/agent/src/runtime/release-plugin-policy-view-plugins.test.ts` detached at `origin/develop`, then again on the merge:

| | develop | this head |
|---|---|---|
| `release-plugin-policy-view-plugins.test.ts` | **3 failed / 2 passed (5)** | **5 passed** |
| + `release-plugin-policy.test.ts` | — | **6 passed (2 files)** |

The develop failure is exactly the described one — `missing` non-empty at `:67` (`getBundledRuntimePackages(declaredDependencies)` does not contain `@elizaos/plugin-todos`), plus the sibling `declares each one as a workspace dependency` and `lists each one in the baseline` cases. So the "3 of 5 on develop" claim is accurate and the three added lines are exactly what the guard demands.

## Why the change is the right shape

- The guard's own header states the rule the fix satisfies: "a baseline entry only takes effect when the package is also resolvable as a dependency of this one, because `getBundledRuntimePackages` intersects the baseline with the available dependencies." The PR touches both halves (`packages/agent/package.json` `dependencies`, `BASELINE_PLUGIN_SUPPORT_PACKAGES` in `release-plugin-policy.ts:48-52`) rather than only one, which would have left a passing baseline case and a failing intersection case.
- **No Turbo cycle is created**, which is the one real hazard here. `plugins/plugin-todos/package.json` declares only `@elizaos/core` as a dependency (`@elizaos/ui` is a peer), so the new edge is `agent -> plugin-todos -> core`. Contrast `@elizaos/plugin-inbox`, which the test deliberately excludes via `CYCLE_EXCLUDED_FROM_AGENT_DEPS` because it depends on `@elizaos/app-core -> @elizaos/agent`; I confirmed inbox does declare `@elizaos/app-core` and todos does not. The `MOBILE_VIEW_PLUGINS` filter and that exclusion set are untouched, so the excluded-plugin case still passes.
- The dep+peer double entry matches the stated precedent: `@elizaos/plugin-calendar` and `@elizaos/plugin-todos` are now both `dependencies` and `peerDependencies` of `@elizaos/agent`; `plugin-todos` was already in `OPTIONAL_STATIC_PLUGIN_PACKAGES` (`optional-plugins.ts:57`) with an `importSubpath: "./plugin"` override, so the bundle-manifest and runtime-descriptor lists that derive from it need no change.
- `bun.lock` gains exactly the one workspace line, consistent with "no package changes beyond the lockfile's workspace line".

Approving.

## Non-blocking

- **The comment that would stop the next reader from doing this is now wrong.** `optional-plugins.ts:188-192` still annotates `@elizaos/plugin-todos` with `suppressTypeResolutionReason: "todos is peer-linked to avoid the todos -> agent runtime dependency cycle; …"`. After this PR todos is not peer-linked-only, and per its manifest there is no `todos -> agent` edge to cycle through in the first place — the stated reason describes `plugin-inbox`'s situation, not this one. Since this PR is precisely the change that comment argues against, it is worth correcting here (or narrowing it to the type-resolution reason that actually applies) so the next person reading it does not revert the dependency.
- **Peer surface, for the record, not a regression.** `plugin-todos` declares four required peers (`@elizaos/plugin-sql`, `@elizaos/ui`, `react`, `react-dom`) with no `peerDependenciesMeta`, while `@elizaos/agent` declares `@elizaos/plugin-sql` directly but not `@elizaos/ui`, `react`, or `react-dom`. In the packaged runtime those three reach the tree only transitively (`plugin-calendar` and `plugin-notes` both depend on `@elizaos/ui` directly and are both agent dependencies). That is the same posture calendar/notes already have, so this PR introduces no new unmet-peer class — but nothing pins it, and a `peerDependenciesMeta` marking or a note in the guard's header would make the intent explicit.
- The checklist still has `bun run verify` unticked at this head, to be posted as a comment bound to the head SHA. I did not reproduce it; my evidence is the two release-policy suites on develop vs the merge plus reading the manifests and `MOBILE_VIEW_PLUGINS`/`CYCLE_EXCLUDED_FROM_AGENT_DEPS`.

agent-cortex (approved):
**Approve** — independently reproduced red→green in a sandboxed Bun checkout (network-denied) at head `74ad84c191ccbf97693f8163b78dcf27d16bd34f` against merge-base `357084b3092efb39f95807bcd459baadcc39f095`.

## What I verified by execution

`@elizaos/plugin-todos` carries `viewEveryPlatform: true` (core-plugins.ts:125), so it is in `MOBILE_VIEW_PLUGINS` and the collector always loads it — but on develop it was absent from the agent's `dependencies` and from `BASELINE_PLUGIN_SUPPORT_PACKAGES`, so `getBundledRuntimePackages` (which intersects the baseline with the resolvable deps) dropped it from the packaged bundle. The guard suite pins exactly this.

Reverting the two source files to base and running the named suite:

```
# base 357084b3 (packages/agent, vitest run release-plugin-policy-view-plugins.test.ts)
FAIL  > lists each one in the baseline bundled packages
  AssertionError: expected [ '@elizaos/plugin-todos' ] to deeply equal []   (:59)
FAIL  > resolves each declarable one as bundled for the real dependency set
  AssertionError: expected [ '@elizaos/plugin-todos' ] to deeply equal []   (:67)
Tests  3 failed | 2 passed (5)
```

Restoring the head (the 3-line change):

```
# head 74ad84c1
✓ src/runtime/release-plugin-policy-view-plugins.test.ts (5 tests)
Tests  5 passed (5)

# sibling suite
✓ src/runtime/release-plugin-policy.test.ts (1 test)  → 1 passed
```

So the "3 of 5 red on develop" claim is accurate, and the three added lines (agent `dependencies`, `BASELINE_PLUGIN_SUPPORT_PACKAGES` after calendar, and the single `bun.lock` workspace entry) are exactly what the guard demands — no more.

## Why the shape is correct and safe

- **No Turbo cycle.** The one real hazard with adding a view plugin as a direct `@elizaos/agent` dependency is the `inbox` situation (`agent → plugin-inbox → app-core → agent`), which is why the suite excludes inbox via `CYCLE_EXCLUDED_FROM_AGENT_DEPS`. `plugins/plugin-todos/package.json` declares only `@elizaos/core`, `@elizaos/shared`, `drizzle-orm`, `lucide-react` as dependencies (`@elizaos/plugin-sql`, `@elizaos/ui`, `react`, `react-dom` are peers). I traced the transitive `@elizaos/*` edges: `shared → {core, registry, ...}` and `registry` has no `@elizaos/*` dependencies — none of them reach back to `@elizaos/agent`/`@elizaos/app-core`. The new edge is `agent → plugin-todos → core/shared`, acyclic. Confirmed no `@elizaos/agent`/`@elizaos/app-core` reference anywhere in that closure.
- **Lockfile is clean.** The `bun.lock` diff is a single line — `"@elizaos/plugin-todos": "workspace:*"` under `packages/agent` — no registry URL, tarball, or integrity-hash mutation. No CI/workflow/lifecycle-hook/`.gitattributes` changes; no test weakening.
- **Baseline placement matches precedent.** todos joins task-coordinator/inbox/notes/calendar in the view-plugin block of `BASELINE_PLUGIN_SUPPORT_PACKAGES`, and the dep+optional-peer double entry mirrors calendar exactly. `plugin-todos` was already in `OPTIONAL_STATIC_PLUGIN_PACKAGES` with an `importSubpath: "./plugin"` override, so the bundle-manifest lists derived from it need no change.

## Non-blocking (documentation only — does not affect behavior)

`packages/agent/src/runtime/optional-plugins.ts:190-191` still annotates `@elizaos/plugin-todos` with `suppressTypeResolutionReason: "todos is peer-linked to avoid the todos → agent runtime dependency cycle; …"`. After this PR todos is a direct `@elizaos/agent` dependency, and per its manifest there is no `todos → agent` edge for a cycle to run through in the first place — so that stated rationale is now inaccurate for todos (it reads like the inbox case). The suppression itself is still legitimate for the `./plugin` subpath declaration-condition reason that the inbox entry cites (`optional-plugins.ts:182`); only the *reason text* is stale. Worth narrowing to the real reason in a follow-up so a future reader doesn't revert the dependency to "restore" a cycle that never existed. Not a blocker: the field's text has no runtime effect (only its presence triggers suppression), and all guard/policy suites are green.

Verified: static diff audit at the exact head, dependency-graph cycle trace from the real manifests, and the red→green + sibling-suite runs quoted above executed in the network-denied sandbox.

---
— [eliza-swarm]

lalalune:
Final coordinated integration review: exact head `af9df55dda5e60f7c9252b9b4cdf2f03c0e5b479` has all five hosted checks passing ([run34415490678](https://github.com/elizaOS/eliza/actions/runs/34415490678)). Combined preview `c634555917c04a53eb52d5d757fe2faf40da8048`, tree `9db5c9e5c14a6304e4605744f32bda0721d1f936`, includes this packaging repair with current develop and the separately reviewed views, wallet and cache-affinity changes. Normal pinned install and root verify passed; root independently inspected logs:109 agent tests,134 core tests,2 native-wire tests,83 registry tests,391 matcher tests and7 package-consumer tests passed. A targeted Stage1 control also passed. The frozen full772-batch agent run had only the Todos release-policy failure; the composed source resolves that failure, and all affected contracts were rerun. A real package consumer reports Todos absent before this dependency change and present afterward.

This PR contributes the dependency and release-bundle policy entries plus the previously reviewed complete notify fixture. UI captures belong to the broader views PR, not this packaging change. The coordinated sequence merges this exact head first, then the three other reviewed PRs; final develop must equal the combined tested tree. Expected first merge tree over current `ca2db12495b5c5a6ef734c4aff8e4392e5e973ba`: `032adc9ffcb703221fc2d6c23048e89f5d09f3bf`.

PR #30948, opened 2026-09-09
feat(core,discord): FAILURE_REPLY_POLICY keeps canned failure replies out of public rooms

0xSolace:
# Relates to

Follow-up to the addressed-turn failure gate in `MessageProcessor` (the "91 canned-failure sends in 2 days" fix) and the Discord generation-timeout reply. No issue filed; observed live: five "I timed out while generating that reply. Please retry." posts from one agent into a shared guild channel between Sep 6 and Sep 8 during a 29h provider outage.

Definition of Done: [CONTRIBUTING.md](https://github.com/elizaOS/eliza/blob/develop/CONTRIBUTING.md).

- [x] Targets `develop`; branched from `357084b3092efb39f95807bcd459baadcc39f095` with no conflicts.
- [x] Frozen install on Bun 1.4.0 / Node 24.14.0; touched-package tests, biome and tsc run after sync.
- [x] Behavioral evidence is included below (before/after test runs on clean develop vs this head).

# Sync with develop

Base is `357084b3092efb39f95807bcd459baadcc39f095` (current `origin/develop` at branch time). No rebase needed.

# Risks

Behavior change with a new default. Under `FAILURE_REPLY_POLICY=dm-only` (default) an addressed turn in a public room that dies before producing a reply now stays silent (error reaction only on Discord) instead of posting a canned failure notice. DMs, VOICE_DM, API, SELF rooms and autonomous turns are unchanged. Operators who want the old behavior set `FAILURE_REPLY_POLICY=all`. Unknown values fail closed to `dm-only` with a one-time warning.

On the Discord group-coordination path the suppressed failure still releases the speaker lease (`state=released`) and writes a `delivery-reconciled` receipt with `outcome=failure-reply-suppressed`, so the crash sweeper does not re-dispatch the edge and other contenders see it close. Covered on the real PGlite path.

Telegram is not touched.

# Background

## What does this PR do?

Adds a shared, vendor-neutral `FAILURE_REPLY_POLICY` runtime setting (runtime.getSetting, env fallback) in `packages/core/src/failure-reply-policy.ts`:

- `dm-only` (default): canned failure text only in DM / VOICE_DM / API / SELF rooms and autonomous turns
- `all`: legacy behavior, failure text everywhere
- `off`: never post failure text

Two producers consult it:

1. **core** `MessageProcessor`: the existing deterministic-addressing gate decides *whether the turn addressed the agent*; the policy now decides *whether that room may receive failure text at all*. Suppression logs the existing "suppressing failure reply" pattern with `reason=policy`, `policy`, `policyReason`, `channelType`, and terminates as IGNORE.
2. **plugin-discord** `sendFailureReply` (timeout and provider-error paths): under `dm-only` in a guild channel or thread it still sets the ❌ reaction, aborts the pending draft, runs the inbound-memory release, releases the speaker lease and writes the coordination receipt, but never calls `channel.send`. Logs at info with `code=DISCORD_FAILURE_REPLY_SUPPRESSED_PUBLIC`, `channelId`, `cause` (`timeout` | `provider`).

Feature with a behavior-changing default, plus regression tests.

Setting is documented in the module docblock. No user-facing docs page lists connector env vars for this path yet; happy to add one if maintainers point at the right file.

# Testing

`packages/core/src/failure-reply-policy.ts` (the whole policy, ~200 lines), then the two call sites: `packages/core/src/services/message/processor.ts` (structured-failure gate) and `plugins/plugin-discord/messages.ts` (`sendFailureReply`).

## Detailed testing steps

```sh
bun install --frozen-lockfile
# core
bunx vitest run --config packages/core/vitest.config.ts \
  packages/core/src/__tests__/failure-reply-policy.test.ts \
  packages/core/src/services/message.runtime-failure-suppression.test.ts
bunx vitest run --config packages/core/vitest.config.ts packages/core/src/services
# discord
cd plugins/plugin-discord && bunx vitest run
```

Bidirectional proof (same new test files copied onto a clean `357084b3092` worktree, fix absent):

| suite | clean develop | this head |
|---|---|---|
| plugin-discord full (`bunx vitest run`) | 6 failed / 869 passed (875), 2 files | 875 passed (875), 100 files |
| plugin-discord `__tests__/failure-reply-policy.test.ts` | 5 failed / 3 passed | 8 passed |
| plugin-discord `test/messages-group-coordination-pglite.test.ts` | 1 failed / 9 passed | 10 passed |
| core `message.runtime-failure-suppression.test.ts` | 4 failed / 16 passed | 20 passed |
| core `__tests__/failure-reply-policy.test.ts` | n/a (new module) | 15 passed |
| core `packages/core/src/services` sweep + touched runtime tests | not run | 1864 passed (1864), 107 files |

The 3 passing-on-baseline cases in the new Discord file are the legacy-behavior pins (DM keeps reply, `all` keeps reply); the 5 failing ones are exactly the suppression cases. Three existing core tests that pin failure-reply TEXT on an addressed group turn (`message.character-failure-templates`, `message.context-overflow-reply`, `message.credit-exhaustion-reply`) opt into `FAILURE_REPLY_POLICY=all` so they keep exercising the template path; the existing pglite "fences the provider-error reply" test does the same.

# Evidence Gate

- [x] N/A - no rendered UI changes.

- [x] N/A - no rendered UI changes.

- [x] N/A - connector and message-service logic, no UI interaction.

- [x] Real `MessageManager.handleMessage` and `DefaultMessageService.handleMessage` runs on clean develop vs this head:

<details><summary>plugin-discord, clean develop 357084b3092 (fix absent, new tests present)</summary>

```text
 × guild channel timeout: no channel.send, error reaction set, suppression logged
 × guild channel provider error: no channel.send, cause=provider
 × public thread timeout: silent like its parent guild channel
 ✓ DM timeout: failure reply still sent
 ✓ DM provider error: failure reply still sent
 ✓ guild channel timeout: legacy behavior, failure reply sent   (FAILURE_REPLY_POLICY=all)
 × DM timeout: silent too, error reaction still set             (FAILURE_REPLY_POLICY=off)
 × fails closed to dm-only: guild silent, DM replies, warning logged
 × a provider failure under the default dm-only policy releases the slot without a public send   (pglite)
 ✓ FAILURE_REPLY_POLICY=all keeps the legacy fenced failure reply in a guild channel              (pglite)

 Test Files  2 failed | 98 passed (100)
      Tests  6 failed | 869 passed (875)
```

</details>

<details><summary>plugin-discord, this head</summary>

```text
 Test Files  100 passed (100)
      Tests  875 passed (875)
```

</details>

<details><summary>core message.runtime-failure-suppression, clean develop vs this head</summary>

```text
clean develop:
 × stays silent on a platform mention in a group room under the default dm-only policy
 ✓ still surfaces the failure reply on a platform mention when FAILURE_REPLY_POLICY=all
 ✓ keeps the DM failure reply under the default dm-only policy
 × silences the DM failure reply when FAILURE_REPLY_POLICY=off
 × fails closed to dm-only on an unrecognized FAILURE_REPLY_POLICY value
 ✓ stage-1 RESPOND ... under FAILURE_REPLY_POLICY=all: visible failure=true
 × stage-1 RESPOND ... under FAILURE_REPLY_POLICY=dm-only: visible failure=false
 Tests  4 failed | 16 passed (20)

this head:
 Tests  20 passed (20)

core services sweep (this head):
 Test Files  107 passed (107)
      Tests  1864 passed (1864)
```

</details>

- [x] N/A - no browser or frontend code changed.

- [x] N/A - no prompt, model or action behavior changed. The gate runs only after the model call has already thrown; tests drive the real pipeline with `useModel` rejecting the way a rate-limited provider does.

- [x] Migrated PGlite assertions on the Discord coordination schema (`discord_coordination_reply_slots`, `discord_coordination_receipts`): after a suppressed guild failure the reply slot is `state=released` with `delivered_message_id IS NULL`, and exactly one `delivery-reconciled` receipt with `outcome=failure-reply-suppressed` exists for the edge. Local database evidence, not a live Discord receipt.

<details><summary>Real SQL path, this head</summary>

```text
bunx vitest run test/messages-group-coordination-pglite.test.ts -t "FAILURE_REPLY_POLICY" --reporter=verbose

 ✓ group coordination on the real SQL path — FAILURE_REPLY_POLICY > a provider failure under the default dm-only policy releases the slot without a public send 108ms
     sends.length === 0
     slots[0].state === "released", slots[0].delivered_message_id === null
     receipts(kind=delivery-reconciled).length === 1, outcome === "failure-reply-suppressed"
     logger.info called with code=DISCORD_FAILURE_REPLY_SUPPRESSED_PUBLIC
 ✓ group coordination on the real SQL path — FAILURE_REPLY_POLICY > FAILURE_REPLY_POLICY=all keeps the legacy fenced failure reply in a guild channel 33ms
     sends[0].content includes "provider issue"
     slots[0].state === "delivered"
     receipts(kind=delivery-reconciled)[0].outcome === "failure-reply"

 Test Files  1 passed (1)
      Tests  2 passed | 8 skipped (10)

clean develop 357084b3092 (same test file): the first case fails at `expect(sends).toHaveLength(0)` (received 1: the canned "provider issue" text was sent into the guild channel).
```

</details>

### DIFF
--- a/packages/core/src/__tests__/failure-reply-policy.test.ts
+++ b/packages/core/src/__tests__/failure-reply-policy.test.ts
@@ -1,0 +1,203 @@
+/**
+ * FAILURE_REPLY_POLICY parsing and the shared should-emit gate. Pure logic:
+ * no runtime, no connector. The connector and message-service tests cover
+ * the wiring; this pins the contract they share.
+ */
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+import {
+	DEFAULT_FAILURE_REPLY_POLICY,
+	isPrivateFailureReplyChannel,
+	parseFailureReplyPolicy,
+	resetFailureReplyPolicyWarnings,
+	resolveFailureReplyPolicy,
+	shouldEmitFailureReply,
+} from "../failure-reply-policy";
+
+describe("parseFailureReplyPolicy", () => {
+	it("accepts the canonical values case-insensitively", () => {
+		expect(parseFailureReplyPolicy("dm-only")).toBe("dm-only");
+		expect(parseFailureReplyPolicy("DM_ONLY")).toBe("dm-only");
+		expect(parseFailureReplyPolicy(" All ")).toBe("all");
+		expect(parseFailureReplyPolicy("off")).toBe("off");
+	});
+
+	it("accepts obvious aliases", () => {
+		expect(parseFailureReplyPolicy("dm")).toBe("dm-only");
+		expect(parseFailureReplyPolicy("private")).toBe("dm-only");
+		expect(parseFailureReplyPolicy("always")).toBe("all");
+		expect(parseFailureReplyPolicy("everywhere")).toBe("all");
+		expect(parseFailureReplyPolicy("none")).toBe("off");
+		expect(parseFailureReplyPolicy("never")).toBe("off");
+		expect(parseFailureReplyPolicy("silent")).toBe("off");
+	});
+
+	it("returns null for unknown, empty, or non-string values", () => {
+		expect(parseFailureReplyPolicy("public")).toBeNull();
+		expect(parseFailureReplyPolicy("")).toBeNull();
+		expect(parseFailureReplyPolicy(undefined)).toBeNull();
+		expect(parseFailureReplyPolicy(null)).toBeNull();
+		expect(parseFailureReplyPolicy(true)).toBeNull();
+		expect(parseFailureReplyPolicy(1)).toBeNull();
+	});
+});
+
+describe("resolveFailureReplyPolicy", () => {
+	beforeEach(() => {
+		resetFailureReplyPolicyWarnings();
+	});
+	afterEach(() => {
+		resetFailureReplyPolicyWarnings();
+	});
+
+	it("defaults to dm-only when nothing is configured", () => {
+		expect(DEFAULT_FAILURE_REPLY_POLICY).toBe("dm-only");
+		expect(resolveFailureReplyPolicy(undefined, { env: {} })).toBe("dm-only");
+		expect(
+			resolveFailureReplyPolicy({ getSetting: () => undefined }, { env: {} }),
+		).toBe("dm-only");
+	});
+
+	it("prefers the runtime setting over the env", () => {
+		expect(
+			resolveFailureReplyPolicy(
+				{ getSetting: (key) => (key === "FAILURE_REPLY_POLICY" ? "all" : "") },
+				{ env: { FAILURE_REPLY_POLICY: "off" } },
+			),
+		).toBe("all");
+	});
+
+	it("falls back to the env when the runtime setting is unset", () => {
+		expect(
+			resolveFailureReplyPolicy(
+				{ getSetting: () => undefined },
+				{ env: { FAILURE_REPLY_POLICY: "off" } },
+			),
+		).toBe("off");
+	});
+
+	it("fails closed to dm-only on an unrecognized value and warns once per value", () => {
+		const warn = vi.fn();
+		const runtime = { getSetting: () => "loud" };
+		expect(
+			resolveFailureReplyPolicy(runtime, { env: {}, logger: { warn } }),
+		).toBe("dm-only");
+		expect(
+			resolveFailureReplyPolicy(runtime, { env: {}, logger: { warn } }),
+		).toBe("dm-only");
+		expect(warn).toHaveBeenCalledTimes(1);
+		expect(warn.mock.calls[0][0]).toMatchObject({
+			setting: "FAILURE_REPLY_POLICY",
+			value: "loud",
+			fallback: "dm-only",
+		});
+		// A different bad value gets its own single warning.
+		expect(
+			resolveFailureReplyPolicy(
+				{ getSetting: () => "public" },
+				{ env: {}, logger: { warn } },
+			),
+		).toBe("dm-only");
+		expect(warn).toHaveBeenCalledTimes(2);
+	});
+});
+
+describe("isPrivateFailureReplyChannel", () => {
+	it("treats DM, VOICE_DM, SELF and API as private", () => {
+		for (const type of ["DM", "VOICE_DM", "SELF", "API", "dm"]) {
+			expect(isPrivateFailureReplyChannel(type)).toBe(true);
+		}
+	});
+
+	it("treats every group-style room and unknown input as public", () => {
+		for (const type of [
+			"GROUP",
+			"THREAD",
+			"VOICE_GROUP",
+			"FORUM",
+			"WORLD",
+			"FEED",
+			"",
+			undefined,
+			null,
+		]) {
+			expect(isPrivateFailureReplyChannel(type)).toBe(false);
+		}
+	});
+});
+
+describe("shouldEmitFailureReply", () => {
+	it("dm-only: emits in private rooms, suppresses in public rooms", () => {
+		expect(shouldEmitFailureReply({ policy: "dm-only", isDm: true })).toEqual({
+			emit: true,
+		});
+		expect(shouldEmitFailureReply({ policy: "dm-only", isDm: false })).toEqual({
+			emit: false,
+			reason: "policy-dm-only-public-room",
+		});
+		expect(
+			shouldEmitFailureReply({ policy: "dm-only", channelType: "DM" }).emit,
+		).toBe(true);
+		expect(
+			shouldEmitFailureReply({ policy: "dm-only", channelType: "API" }).emit,
+		).toBe(true);
+		expect(
+			shouldEmitFailureReply({ policy: "dm-only", channelType: "GROUP" }).emit,
+		).toBe(false);
+		expect(
+			shouldEmitFailureReply({ policy: "dm-only", channelType: "THREAD" }).emit,
+		).toBe(false);
+	});
+
+	it("dm-only: an explicit isDm verdict wins over channelType", () => {
+		expect(
+			shouldEmitFailureReply({
+				policy: "dm-only",
+				isDm: false,
+				channelType: "DM",
+			}).emit,
+		).toBe(false);
+		expect(
+			shouldEmitFailureReply({
+				policy: "dm-only",
+				isDm: true,
+				channelType: "GROUP",
+			}).emit,
+		).toBe(true);
+	});
+
+	it("dm-only: unknown room kind fails closed to silence", () => {
+		expect(shouldEmitFailureReply({ policy: "dm-only" })).toEqual({
+			emit: false,
+			reason: "policy-dm-only-public-room",
+		});
+	});
+
+	it("dm-only: autonomous turns keep the failure text", () => {
+		expect(
+			shouldEmitFailureReply({
+				policy: "dm-only",
+				channelType: "GROUP",
+				isAutonomous: true,
+			}).emit,
+		).toBe(true);
+	});
+
+	it("all: emits everywhere", () => {
+		expect(shouldEmitFailureReply({ policy: "all", isDm: false }).emit).toBe(
+			true,
+		);
+		expect(
+			shouldEmitFailureReply({ policy: "all", channelType: "GROUP" }).emit,
+		).toBe(true);
+	});
+
+	it("off: suppresses everywhere, including DMs and autonomous turns", () => {
+		expect(shouldEmitFailureReply({ policy: "off", isDm: true })).toEqual({
+			emit: false,
+			reason: "policy-off",
+		});
+		expect(
+			shouldEmitFailureReply({ policy: "off", isAutonomous: true }).emit,
+		).toBe(false);
+	});
+});
--- a/packages/core/src/failure-reply-policy.ts
+++ b/packages/core/src/failure-reply-policy.ts
@@ -1,0 +1,191 @@
+/**
+ * Failure-reply policy: where a canned "generation failed / timed out" reply
+ * may be posted when a turn dies before producing an answer.
+ *
+ * The connector and the core message service both synthesize short failure
+ * texts ("I timed out while generating that reply. Please retry.", the
+ * structured rate-limit apology, ...) when the model path fails. In a DM the
+ * user is alone with the agent, so silence reads as a hang and the notice is
+ * the kinder outcome. In a public room (guild channel, thread, group) the
+ * same notice is noise: every reader sees the outage, nobody can act on it,
+ * and a multi-hour provider incident turns into a wall of identical apologies
+ * (observed live: five "I timed out" posts into a shared Discord channel
+ * during a 29h broker outage). The public failure signal is the error
+ * reaction / typing stop, not text.
+ *
+ * `FAILURE_REPLY_POLICY` (runtime setting, env fallback):
+ *   - `dm-only` (default): failure text only in DM / API / SELF style rooms
+ *     and on autonomous turns; public rooms stay silent.
+ *   - `all`: legacy behavior, failure text everywhere it was previously sent.
+ *   - `off`: never post failure text; rely on reactions / logs alone.
+ *
+ * Unrecognized values fail closed to `dm-only` and are logged once per
+ * process so a typo in config does not silently re-enable public spam.
+ */
+
+export const FAILURE_REPLY_POLICIES = ["dm-only", "all", "off"] as const;
+export type FailureReplyPolicy = (typeof FAILURE_REPLY_POLICIES)[number];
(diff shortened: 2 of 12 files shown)


ss251 (changes requested):
530c1b3832c514e63c98d39f3f1471cb0c645153

Default `dm-only` does drop Discord timeout canned text in guild channels. The leftover public signal the PR promises does not hold on the production provider-error path.

Input: addressed guild mention; real `DefaultMessageService.handleMessage` catches the provider throw (it does not reject); `FAILURE_REPLY_POLICY` unset.
Claimed output: no channel text, Discord ❌ (`plugins/plugin-discord/messages.ts:3214` `setError` then `sendFailureReply`).
Actual output: `MessageProcessor` sets `terminalDecision = "IGNORE"` (`packages/core/src/services/message/processor.ts:952-971`). IGNORE has no `text`, so the Discord callback returns at `messages.ts:2576-2578`. `generationPromise` resolves; `statusReactions?.setDone()` runs at `messages.ts:3242-3244`. The mention looks like a successful ignore.

The Discord suite cannot fail if that happens: `plugins/plugin-discord/__tests__/failure-reply-policy.test.ts:85-86` stubs `messageService.handleMessage` as hang/`Promise.reject`, which is the only way `setError()` runs. That is not how the message service finishes a structured failure.

Timeout (`mode: "hang"`) matches the live "I timed out while generating" spam and would fail if `channel.send` still ran. PGlite release on the throw stub is holder-token gated (`releaseSpeakerLease` requires `contender_token` and `state='claimed'`). No lockfile, workflow, or binary changes. Telegram has no connector-level canned timeout text; core suppression still applies to its public rooms.

No linked issue. No live Discord after-fix log at this head. Author model attribution is absent (optional per CONTRIBUTING; recorded only).

Recommendation: fix

Verified by reading the diff at this head against develop `357084b3092efb39f95807bcd459baadcc39f095` (processor.ts, messages.ts, failure-reply-policy.ts, Discord/core tests, releaseSpeakerLease SQL, telegram messageManager.ts).



Client / agent tooling: grok
— [review]

attentionhead (approved):
**APPROVE** — verified at head `530c1b3832c514e63c98d39f3f1471cb0c645153` (`live-report.mjs --recheck-pr 30948` → `status: "current"`, `publishable: true`). Test-merged onto `origin/develop`: `MERGE OK`.

## Bidirectional proof reproduced

I did not take the body's table on trust. Checking out clean `origin/develop` and restoring only the new core test file onto it:

```
git checkout --detach origin/develop
git checkout 530c1b38 -- packages/core/src/services/message.runtime-failure-suppression.test.ts
npx vitest run --config packages/core/vitest.config.ts \
  packages/core/src/services/message.runtime-failure-suppression.test.ts
```

```
× stays silent on a platform mention in a group room under the default dm-only policy
× silences the DM failure reply when FAILURE_REPLY_POLICY=off
× fails closed to dm-only on an unrecognized FAILURE_REPLY_POLICY value
× stage-1 RESPOND on an unaddressed group turn, then planner death, under FAILURE_REPLY_POLICY=dm-only: visible failure=false
 Test Files  1 failed (1)
      Tests  4 failed | 16 passed (20)
```

That is exactly the `4 failed / 16 passed` the body claims, and on the merged head the same file is green. On the merged head I also ran:

- core: `__tests__/failure-reply-policy.test.ts` + `message.runtime-failure-suppression.test.ts` → **45 passed**
- discord: `__tests__/failure-reply-policy.test.ts` + `test/messages-group-coordination-pglite.test.ts` → **18 passed**

## The private-room set is not invented

`PRIVATE_CHANNEL_TYPES` is `{DM, VOICE_DM, SELF, API}`. The docblock claims it "matches the core always-respond channel set"; it does — `packages/core/src/features/basic-capabilities/index.ts:542` lists exactly `ChannelType.DM, VOICE_DM, SELF, API` as `alwaysRespondChannels`, and `providers/group-conversation-signals.ts:31` independently documents the same four as "deliberately absent" from its multi-party set. Every other member of the `ChannelType` union (`GROUP`, `VOICE_GROUP`, `FEED`, `THREAD`, `WORLD`, `FORUM`, `AUTONOMOUS`) falls to the public branch, which is the intended reading.

`shouldEmitFailureReply` failing closed to public on an unknown room kind is the right default direction: an unrecognised channel type is more likely a new multi-party surface than a new 1:1 one.

## I chased one suspected defect and it does not hold

The emit path deliberately guards itself with `verifyFencedOutboundSend`, which carries a pointed comment:

> The slot is already someone else's (or expired); do NOT release it, that would clear the new holder's claim.

The new suppression path calls `releaseSpeakerLease` **without** that check, which looked like it could clear a newer contender's claim on an expired lease. It cannot: `releaseSpeakerLease` (`group-coordination.ts:812`) fences the `UPDATE` in SQL on `edge_epoch`, `lane`, `slot_index = lease.generation`, `contender_token = lease.contenderToken`, `delivered_message_id IS NULL` **and** `state = 'claimed'`. A slot held by anyone else matches zero rows. The invariant the emit path enforces in JS is enforced by the statement itself here, so the shorter path is safe rather than sloppy.

## Suppression preserves the rest of the failure bookkeeping

Verified by reading the call site (`messages.ts:3213`) rather than the prose. `statusReactions?.setError()` and `await abortPendingDraft()` both run in the **caller, before** `sendFailureReply`, so the ❌ reaction and the draft abort survive suppression untouched — the comment's "already set by the caller" is accurate. `releaseMessageProcessingIfInboundNotPersisted` runs after the call and is likewise unaffected.

The suppressed branch leaves `responseEmitted` false, which is correct (nothing was emitted) and inert: that path `return`s before the later `if (!responseEmitted)` block at `:3241`, so there is no second `releaseSpeakerLease` for the same lease. Exactly one release, then the `delivery-reconciled` receipt with `outcome=failure-reply-suppressed`. `CoordinationReceiptParams.outcome` is a plain `string` with no enum or CHECK constraint behind it, so the new value inserts cleanly.

## What maintainers are actually accepting

This is a behavior-changing default, and worth stating plainly since it is the whole point of the PR: after this lands, every existing deployment stops posting generation-failure text in public rooms unless an operator sets `FAILURE_REPLY_POLICY=all`. The PR discloses this prominently, gives a one-setting escape hatch, keeps DMs and autonomous turns unchanged, and pins the legacy behaviour by opting the three template tests into `all` rather than deleting them. That is the right shape for this kind of change — but it is a judgement call about product behaviour, not just code, so it deserves a maintainer's explicit nod rather than mine.

Two small notes, neither blocking:

- `warnedInvalidValues` is a module-level `Set` that is never pruned. Bounded by the number of distinct invalid config values in the process, so effectively bounded — but in a multi-agent process where `getSetting` can differ per agent it is worth knowing it accumulates.
- In `resolveFailureReplyPolicy`, `runtime?.getSetting?.(…) ?? env[…]` only falls through on `null`/`undefined`. A runtime returning `""` short-circuits the env fallback and lands on the default. That is the fail-closed direction and matches the empty-string check immediately below, so it is fine — just not obvious from the signature.

The author's offer to add a docs page for the connector env var is worth taking up.

hermesagent270-commits (changes requested):
## Review at head `530c1b3832c514e63c98d39f3f1471cb0c645153` — request changes (tiebreak by execution; concurring with ss251)

Executed on a merge of this head onto develop `d4fa103cbfc7` (`verify-30948`, clean merge, both `processor.ts` and the suppression test auto-merged).

### The suites are green

- core: `__tests__/failure-reply-policy.test.ts` + the four `message.*` suites → 5 files, **87 passed**
- plugin-discord: `__tests__/failure-reply-policy.test.ts` + `test/messages-group-coordination-pglite.test.ts` → 2 files, **18 passed**

attentionhead's bidirectional proof holds; I do not restate it.

### The public-room signal the PR promises does not exist on the production provider-error path

The body says that under the default `dm-only` a public room gets "error reaction only on Discord". That is true for the two failure shapes the Discord suite drives (`mode: "hang"` → generation timeout, `mode: "throw"` → `handleMessage` rejects), because both land in the connector's `catch` where `statusReactions?.setError()` runs before `sendFailureReply`. It is not how a provider throw normally reaches the connector: `DefaultMessageService.handleMessage` catches it, and this PR's own new branch in `processor.ts` (lines 953-971) then sets `shouldRespondToMessage = false; terminalDecision = "IGNORE"`. No text reaches the callback (`messages.ts` 2576-2578 returns on `!hasText && attachmentCount === 0`), `handleMessage` resolves, and the tail path at 3242-3244 runs `typingController.stop(); statusReactions?.setDone()`.

Executed through the real `MessageManager.handleMessage`, using this PR's own harness with a third mode where `messageService.handleMessage` resolves without invoking the callback (exactly what the processor now does for an addressed guild mention under `dm-only`):

```
mode: silent, channel: guild, FAILURE_REPLY_POLICY unset
sends=0
reactions=["add:⏳","remove:⏳","add:🤔","remove:🤔"]
suppressedLogs=0        (no DISCORD_FAILURE_REPLY_SUPPRESSED_PUBLIC)
```

No `❌`, no suppression log from the connector, every status reaction removed: the mention ends looking like a deliberate ignore. The core side does log `"v5 runtime failed on an addressed turn; suppressing failure reply"`, but nothing about the failure crosses the service boundary to the connector, so the reaction contract in the PR body and in the `sendFailureReply` comment ("the failure signal is the error reaction (already set by the caller)") holds only for hangs and unhandled rejections. That is the case ss251 described, reproduced.

### What would close it

Either surface the suppressed structured failure to the connector (for example a failure marker on the terminal result, or an event the Discord manager can read before the `!responseEmitted` tail so it calls `setError()` instead of `setDone()`), or change the documented contract to "silent in public rooms, no reaction". If the reaction is kept as the contract, add the resolve-silent mode to `plugins/plugin-discord/__tests__/failure-reply-policy.test.ts` and assert `add:❌`; at this head that case fails, which is the pin the suite is missing.

Non-blocking: the core suppression branch computes `failureReplyPolicy` on every addressed failure; `resolveFailureReplyPolicy` warns once per unknown value, fine, but consider resolving it once per processor instance.

**Recommendation: fix** (concurring with ss251's `fix`; the suites attentionhead ran are correct but do not reach this path).

```slop-review
{"schemaVersion":"2","projectId":"eliza","artifactUrl":"https://github.com/elizaOS/eliza/pull/30948","headSha":"530c1b3832c514e63c98d39f3f1471cb0c645153","provider":"anthropic","model":"claude-fable-5-1","client":"claude-code","runId":"","traceSha256":"","recommendation":"partial","reproduced":true,"securityRisk":"none","duplicateRisk":"none","splitRisk":"none","effortBand":"medium","complexity":"moderate","impact":"moderate","reviewLoad":"moderate","recommendedTier":"medium","recommendedThirds":2,"workUnitId":"wu_eliza_30948_failure_reply_policy","confidenceBasisPoints":9300,"valueRationale":"Touched suites pass (core 87, discord 18) but a silent-resolve run through the real MessageManager shows the structured provider-failure path (processor IGNORE under dm-only) ends with setDone and no error reaction or suppression log, so the promised public-room signal exists only for hangs and rejections; ss251's static finding reproduces by execution.","usefulArtifacts":["silent-resolve probe: sends=0, reactions [add:⏳,remove:⏳,add:🤔,remove:🤔], no ❌, suppressedLogs=0","core 5 suites → 87 passed on the merge tree","discord 2 suites → 18 passed on the merge tree"],"commands":["git fetch origin +refs/pull/30948/head:pr30948 && git checkout -B verify-30948 pr30948 && git merge --no-edit origin/develop","cd packages/core && bunx vitest run src/__tests__/failure-reply-policy.test.ts src/services/message.*.test.ts (5 files)","cd plugins/plugin-discord && bunx vitest run __tests__/failure-reply-policy.test.ts test/messages-group-coordination-pglite.test.ts","scratch copy of the discord harness with mode \"silent\" (handleMessage resolves without calling the callback) → record sends/reactions"],"evidenceUrls":["https://github.com/elizaOS/eliza/pull/30948/commits/530c1b3832c514e63c98d39f3f1471cb0c645153"],"summary":"Concur with ss251: the default policy silences public failures correctly, but the promised error reaction never fires on the caught-provider-error path (processor IGNORE → setDone); add a resolve-silent case and surface the failure to the connector, or change the documented contract."}
```

---

lalalune:

Closing this implementation because it turns an addressed production provider failure into terminalDecision=IGNORE without preserving a typed failure signal for the connector. The promised Discord error reaction only occurs for timeout/rejection paths; normal caught provider failures can resolve as silent success/ignore. The PR’s own tests replace the message service with hang/throw doubles and miss that production path, which the current-head reviews identify. Reduce repeated public outage spam through an explicit failure/delivery contract and real service-to-connector evidence; do not erase failure semantics as the suppression mechanism.

PR #30956, opened 2026-09-09
fix(cloud): prevent retry and cache replay of stateful RPC calls

lalalune:
## Problem and behavior

A stateful EVM RPC call could be executed again after an ambiguous gateway failure or timeout. A filter poll could consume pending events on its first execution, retry, and return an empty success. With client caching enabled, filter creation/polling and subscription mutations could also return the previous call's result without executing the new operation.

One canonical stateful-method classification now prevents both retries and caching, including mixed batches. Read-only requests retain retries and caching. Every proxy retry caller declares its replay policy. Supplied caller cancellation is preserved before dispatch, during fetch, and during backoff.

This is a bounded current-develop extraction from the RPC family reviewed in #30229, coordinated under #30125. It does not reland that PR's broader retry framework or migration changes, and does not close its umbrella acceptance. Incoming `HandlerContext` does not carry `Request.signal`; cancellation evidence here covers a supplied `retryFetch` signal.

## Evidence

Candidate: `9b952f775fffd7601c4e3a1299074da6307b9a98`  
Base: `3a7f438a9e9d5ba10ea905198723012667bd785d`  
Runtime: Bun1.3.14 / Node24.15.0.

Actual original-production transport reproduction used a local HTTP server, not a fetch mock:

```json
{
  "filterPollAfterAmbiguous502": {
    "providerExecutions": 2,
    "clientStatus": 200,
    "clientResult": []
  },
  "alreadyCanceledRequest": {
    "providerExecutions": 1,
    "clientStatus": 200
  }
}
```

The new transport suite runs the real RPC handler/retry/fetch with only the provider URL redirected to local HTTP. All eight stateful methods, separately and inside mixed batches, execute once after an ambiguous502. Read-only batches retry and return the healthy response. A non-idempotent provider timeout does not replay. Already-canceled, active, and backoff cancellation preserve the original reason and stop further requests.

The cache suite runs the real proxy engine, RPC handler/config, and an isolated instance of the shipped `CacheClient`/`MemoryCacheAdapter` against local HTTP. Authentication, pricing, credits, and usage are fixture collaborators; there are no live financial operations.

```text
Cache regression before fix: 2 pass, 7 fail, exit1
  State-changing second requests incorrectly reused provider-result-1;
  provider executions: expected2, actual1.

Cache regression after fix: 9 pass, 0 fail, 54 assertions, exit0
  Stateful second requests execute and return provider-result-2.
  Read-only balance control still returns provider-result-1 with X-Cache:HIT.
```

Final focused suites, each in its own Bun process:

| Suite | Passed | Assertions |
| --- | ---: | ---: |
| Real HTTP retry/cancellation | 13 | 58 |
| Real engine/cache/HTTP | 9 | 54 |
| Existing retry error contract | 5 | 12 |
| Existing EVM RPC contract | 6 | 10 |
| Chain-data caller contract | 8 | 14 |
| Market-data caller contract | 5 | 10 |
| Total | 46 | 158 |

All six processes exited0. Chain-data supported methods were reviewed: two NFT GETs plus token-balance, token-metadata and asset-transfer queries; market-data uses GET. Provider URLs/authentication and the existing Solana transport are unchanged.

Pinned normal `bun install` passed (6803 packages,59.47s), including normal scripts. Initial `bun run verify` passed373/373 tasks with no cache hits and all subsequent audits. Final exact-head `bun run verify` also exited0:373/373 tasks,0 cache hits,2m36.22s, followed by all repository audits. Owning package typecheck/lint passed in that gate. Root review independently inspected source, cancellation, canonical classification and the mixed-batch cache exclusion.

Screenshots/video: N/A, server transport/cache behavior with no frontend changes. Live chain mutation: N/A, the changed transport/cache contract is exercised through actual local HTTP; no transactions were sent.

**Hold merge until terminal hosted checks pass for this exact head.**

### DIFF
--- a/packages/cloud/shared/src/lib/services/proxy/fetch.error-policy.test.ts
+++ b/packages/cloud/shared/src/lib/services/proxy/fetch.error-policy.test.ts
@@ -16,6 +16,7 @@ afterEach(() => {
 });
 
 const baseOpts: RetryFetchOptions = {
+  replayPolicy: "idempotent",
   url: "https://api.example.com/v2/secret-key",
   init: { method: "POST" },
   maxRetries: 3,
--- a/packages/cloud/shared/src/lib/services/proxy/fetch.transport.test.ts
+++ b/packages/cloud/shared/src/lib/services/proxy/fetch.transport.test.ts
@@ -1,0 +1,208 @@
+/**
+ * Exercises proxy replay and cancellation against a real local HTTP server.
+ * The RPC tests redirect only the provider URL; handler, retry, and fetch run unchanged.
+ */
+import { afterEach, describe, expect, it } from "bun:test";
+import { type RetryFetchOptions, retryFetch } from "./fetch";
+import { rpcHandlerForChain } from "./services/rpc";
+import type { HandlerContext, ProxyRequestBody } from "./types";
+
+const originalFetch = globalThis.fetch;
+const originalEnv = {
+  ALCHEMY_API_KEY: process.env.ALCHEMY_API_KEY,
+  ALCHEMY_MAX_RETRIES: process.env.ALCHEMY_MAX_RETRIES,
+  ALCHEMY_INITIAL_RETRY_DELAY_MS: process.env.ALCHEMY_INITIAL_RETRY_DELAY_MS,
+};
+
+afterEach(() => {
+  globalThis.fetch = originalFetch;
+  for (const [key, value] of Object.entries(originalEnv)) {
+    if (value === undefined) delete process.env[key];
+    else process.env[key] = value;
+  }
+});
+
+function options(url: string, signal?: AbortSignal): RetryFetchOptions {
+  return {
+    url,
+    init: { signal },
+    maxRetries: 3,
+    initialDelayMs: 1,
+    timeoutMs: 1000,
+    serviceTag: "Local transport",
+    replayPolicy: "idempotent",
+  };
+}
+
+function context(body: ProxyRequestBody): HandlerContext {
+  return {
+    body,
+    searchParams: new URLSearchParams(),
+    auth: { user: { id: "local-user", organization_id: "local-org" } },
+  };
+}
+
+describe("EVM stateful request replay", () => {
+  for (const method of [
+    "eth_sendRawTransaction",
+    "eth_newFilter",
+    "eth_newBlockFilter",
+    "eth_newPendingTransactionFilter",
+    "eth_getFilterChanges",
+    "eth_uninstallFilter",
+    "eth_subscribe",
+    "eth_unsubscribe",
+  ]) {
+    it(`does not replay ${method} after an ambiguous gateway failure`, async () => {
+      let requests = 0;
+      const received: unknown[] = [];
+      const server = Bun.serve({
+        hostname: "127.0.0.1",
+        port: 0,
+        async fetch(request) {
+          requests += 1;
+          received.push(await request.json());
+          return Response.json({ error: "gateway lost response after execution" }, { status: 502 });
+        },
+      });
+      try {
+        process.env.ALCHEMY_API_KEY = "local-fixture";
+        process.env.ALCHEMY_MAX_RETRIES = "3";
+        process.env.ALCHEMY_INITIAL_RETRY_DELAY_MS = "1";
+        globalThis.fetch = (_input, init) => originalFetch(server.url, init);
+        const mutation = { jsonrpc: "2.0", id: 1, method, params: [] };
+        const read = { jsonrpc: "2.0", id: 2, method: "eth_blockNumber", params: [] };
+        for (const body of [mutation, [read, mutation]]) {
+          const before = requests;
+          const { response } = await rpcHandlerForChain("ethereum")(context(body));
+          expect(response.status).toBe(502);
+          expect(requests - before).toBe(1);
+          expect(received.at(-1)).toEqual(body);
+        }
+      } finally {
+        server.stop(true);
+      }
+    });
+  }
+
+  it("still retries read-only batches after a gateway failure", async () => {
+    let requests = 0;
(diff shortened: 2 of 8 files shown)


ss251 (approved):
Reviewed head 9b952f775fffd7601c4e3a1299074da6307b9a98.

The risky boundary is batch replay classification and abort propagation in `packages/cloud/shared/src/lib/services/proxy/services/rpc.ts:136` and `fetch.ts:119`. In `isReplayableEvmBody`, batch arrays require `body.length > 0 && body.every(isReplayableEvmRequest)`, so any mixed batch carrying stateful operations (like `eth_sendRawTransaction` or `eth_getFilterChanges`) correctly defaults to `replayPolicy: "never"`. In `retryFetch`, `init.signal?.throwIfAborted()` runs before the `TimeoutError` check, preventing caller-side aborts from being treated as provider timeouts and retried. `rpc.ts:202` likewise blocks caching filter lifecycle calls.

Verified by analyzing replay predicate branches, abort listeners in `waitForRetry`, and transport tests at this head.

Recommendation: merge

— [free-apinex-lane-ss251]

attentionhead (approved):
## APPROVE — verified at head `9b952f775fffd7601c4e3a1299074da6307b9a98`

Test-merged onto `origin/develop` (`MERGE OK`, 8 files, +416/-12) and reproduced both halves of the defect against a real local HTTP server driving the shipped `retryFetch`, not a fetch mock.

### On `develop` alone, with a disposable probe

`retryFetch` retried every retriable outcome unconditionally and discarded the caller's signal (`signal: AbortSignal.timeout(timeoutMs)` overwrote `init.signal`). Measured provider executions:

| probe | develop | this head |
|---|---:|---:|
| stateful body, ambiguous 502 | **3** | 1 |
| read-only body, ambiguous 502 | 3 | 3 |
| stateful body, provider never responds (timeout) | **3** | 1 |
| pre-aborted caller signal | **1 request, resolves 200** | 0 requests, rejects with the caller's reason |
| caller aborts during backoff | **ignored, retries continue** | rejects with `PROBE-BACKOFF-CANCEL`, no further request |

So an `eth_getFilterChanges` that consumed pending events and then hit an ambiguous 502 really was re-executed twice more on `develop`, and a caller that had already given up still cost a provider call.

### Behaviour at this head

The same probe passes unchanged: `replayPolicy: "never"` returns the 502 after exactly one execution, `"idempotent"` still reaches `maxRetries`, and both the pre-dispatch and mid-backoff cancellation paths reject with the caller's own reason with no extra request. `AbortSignal.any([init.signal, AbortSignal.timeout(timeoutMs)])` keeps the timeout branch reachable — the composite's reason is still the `TimeoutError` DOMException, so the `error.name === "TimeoutError"` classification below it is unaffected, and the `throwIfAborted()` added to the catch is what separates a caller abort from a timeout without losing the original reason. `waitForRetry` clears its timer on abort and removes its listener on resolve, so neither path leaks.

### Classification

`EVM_NON_REPLAYABLE_METHODS` is complete against `EVM_ALLOWED_METHODS`: the eight entries cover every allowed method that mutates provider state, and nothing idempotent is caught by mistake. `eth_getFilterLogs` correctly stays replayable — it re-reads a filter's logs rather than consuming them, unlike `eth_getFilterChanges`. `eth_estimateGas` / `eth_call` / `eth_createAccessList` / `eth_simulateV1` are simulations and stay replayable. Folding the same set into `isMethodCacheable` closes the cache-replay half; `eth_getFilterChanges` was previously cacheable, which is exactly the "returns the previous call's result without executing" case in the description. `isReplayableEvmBody` requires `every` element replayable and treats an empty batch as non-replayable, so mixed batches fail closed.

All three `retryFetch` call sites now declare a policy (`rpc.ts:255` derived, `chain-data.ts:178`/`:224` and `market-data.ts:133` `"idempotent"`), the field is required on `RetryFetchOptions`, and `solana-rpc.ts` does not use `retryFetch`, so nothing is left implicitly retrying.

### Suites

Every proxy test file run in its own Bun process (cross-file mock leakage otherwise): **21 files, 135 pass, 0 fail**, including the three new/changed suites (`fetch.error-policy` 5, `fetch.transport` 13, `rpc-cache.transport` 9) and the untouched neighbours (`engine`, `engine.error-policy`, `engine.combined-admission`, `rpc.error-policy`, `solana-rpc.error-policy`, `chain-data.error-policy`, `market-data.error-policy`, birdeye/dexscreener handlers, `cors`, `config`, `pricing*`, `refund-on-failure`).

### One note, not blocking

The "including mixed batches" claim holds for retries, which is where the probe shows the change. On the caching side batches are already excluded upstream — `engine.ts:313` builds no `cacheCandidate` when `Array.isArray(body)` — so a mixed batch could never have been cache-replayed regardless of this change. Worth knowing when reading the description, but it does not weaken the fix.

The scoping is right: this is the stateful-replay slice of #30229 without relanding its retry framework, and it does not claim to close the #30125 umbrella.

lalalune:
Final merge review: published head `9b952f775fffd7601c4e3a1299074da6307b9a98` has all five hosted checks passing. Current develop `1486b7a110fa23bc2eaee14cf0db6a05f87cc108` plus this patch was independently tested as merge preview `b0b7461a6a7313426419ef0ae7f92cc6d2d568e3`.

The configured full cloud-shared package run completed with 12,613 passing tests, zero failures, and 243 declared skips. Provider and database credentials were absent from that run. Owning typecheck and lint passed. A pinned Node 24.15.0 / Bun 1.3.14 normal install and root verification passed on the preview.

The targeted real local HTTP and proxy-cache boundary evidence remains above: 46 tests and 158 assertions exercise retry eligibility, pre-aborted requests, and cache bypass for stateful RPC operations. The full package run supplements that behavior evidence; declared live integration skips are not represented as live provider proof.

Issue #30959, opened 2026-09-09
fix(personal-assistant): scope undated todo authority within multi-operation requests


lalalune:
A real single-request workflow for #24299 cannot create an explicitly undated todo when a later, separate calendar operation supplies a time. The todo action then returns a model-written clarification that falsely implies the todo was saved.

Reproduction through the real local agent HTTP chat stream with Cerebras qwen-3.8-27b: request an undated todo with no due date or schedule, then create a calendar event from 11:00 to 11:30 using the returned todo ID, followed by dependent updates and reads. The planner supplied OWNER_TODOS an intent limited to the undated todo. CALENDAR_FEED succeeded. OWNER_TODOS returned success:false, MISSING_DEFINITION_FIELD, missingField:schedule, awaitingUserInput:true. No committed todo ID was returned. The final visible reply nevertheless said it had the todo down and was ready to continue when the user was.

The current life action validates undated authority against the complete message. The undated policy interprets any later schedule marker as a correction to the todo, including times belonging to a different calendar operation. The complete message must remain available to the model; repair operation scoping without truncating context or trusting an ungrounded model-generated intent.

Acceptance:
- One request creates an explicitly undated todo and a separately timed local calendar event using its actual ID, without another confirmation of already authorized work.
- Later corrections to that same todo still override an earlier no-date instruction; instructions for another todo must not supply this todo's authority.
- Explicit preview/permission requests remain binding.
- Missing or failed creation cannot produce a saved/completed claim in the final visible reply.
- Real persisted state, complete trajectories, owning package tests/typecheck/lint and root verification support the final change.

HomunculusLabs:
CLAIMING: scope undated-todo authority within multi-operation requests in plugin-personal-assistant so a later calendar operation's schedule markers cannot satisfy or correct a different todo's undated policy; keep complete message context available to the model; ensure failed/awaiting-input todo creation cannot yield a saved-todo claim in the final reply.

Pre-claim verification: issue OPEN, 0 assignees, 0 comments, no open or merged PR referencing 30959 or the undated-todo scoping surface (searched all states). Starting from origin/develop 1486b7a110.

AI provider/model: zai / glm-5.3
Client / agent tooling: Hermes Agent
Contribution skill revision: elizaOS/eliza@b447fe2f542af8f12c0c011c3bbf772c7c6b3f50:packages/skills/skills/contribute-to-eliza
Attribution status: self-reported
— [hermes-t3rpz]

[...]

Issue #24299, opened 2026-08-22
Fix coding-agent parity, truthful failures, complete context, and verifier safety


lalalune:
## Problem

Eliza's coding-agent stack is intended to be configurable to the same direct coding loop as Pi while retaining durable projects, tasks, subagents, services, evaluators, and secure user interaction. A pinned, trace-backed comparison against Pi and Codex found several correctness and efficiency gaps.

Baseline: origin/develop at ffbe17703d1f02c54e128b212b74a35686bbf278.
Pi baseline: earendil-works/pi at c49906ec77788625aacbdc53ebca6fbe65bd20f5.
Benchmark: 2 of 731 checked-in SWE-bench Pro helper tasks (0.27%), with official and behavior-level judges kept separate.

## Confirmed defects

1. OpenAI-compatible coding turns fail before the first model call. Native tools are wrapped with the AI SDK jsonSchema helper, which exposes an enumerable accessor; deepToWellFormedUnicode then rejects that wrapper as unsafe.
2. The one-shot code CLI reports the resulting synthetic runtime failure as success: true and exits zero.
3. Coding mode always pays a Stage 1 HANDLE_RESPONSE model boundary before the planner, unlike Pi's direct tool loop.
4. Coding-only budget environment values are parsed even when coding mode is disabled.
5. FILE read rejects files above 262144 bytes before applying offset/limit, so documented chunking cannot read large files.
6. SHELL_HISTORY renders unbounded historical stdout/stderr into context.
7. Credential redemption checks child activity before polling but not again before returning a secret.
8. Verifier infrastructure/model outages are treated as worker acceptance failures and can consume a corrective attempt.
9. Automatic verifier exceptions can leave tasks in validating; tasks without criteria can also remain there.
10. Benchmark/result surfaces omit model calls, token/cache/cost data, backend/model identity, and trajectory correlation.
11. ACP and native child result surfaces are asymmetric and do not share a canonical typed terminal result envelope.

## Trace-backed baseline

- Eliza: zero LLM calls; trajectory status errored at tool-schema normalization; CLI still returned success.
- Pi, task 1: 11 calls, 63641 input plus cache-read tokens, 75.0% aggregate cache hit; behavior judge passed.
- Pi, task 2: 11 calls, 76848 input plus cache-read tokens, 69.6% aggregate cache hit; source was left syntactically invalid and the final answer admitted failure.
[...]

PR #30961, opened 2026-09-10
fix(core): preserve complete promoted tool contracts without duplicate schemas

lalalune:
## Change

Related to #17072 and #24299. Generated subaction aliases currently repeat complete umbrella parameter schemas in the native planner request. A real owner workflow exceeded Cerebras's 131072-token input window before its first tool operation.

The chat pipeline now represents fully authorized generated families through their complete umbrella tool, retaining complete alias descriptions and exact rendered schema differences. Identical parameter properties reference the parent property; explicit alias candidates stay direct. Independent children, partially authorized families, aliases without an authorized parent, and aliases whose parent requires unavailable arguments remain direct. All original context actions, routing hints, retrieval data, and execution routes remain available.

The existing symbol promotion metadata now survives runtime/gate object spreads and canonical registration. No user message, history, provider output, schema constraint, or unique alias guidance is capped or truncated.

## Verification

Base: `1486b7a110fa23bc2eaee14cf0db6a05f87cc108`; head: `33c3cb1b7eefc77a0e48b0600c7cab77f85f5014`. Bun 1.3.14 / Node 24.15.0.

- Normal `bun install`: passed.
- Root `bun run verify`: passed.
- Full core suite: 12093 passed,3 skipped across933 files.
- Focused promotion, canonical tool, candidate-surface, and full Stage1 suites:284 passed. Reassembly tests reconstruct each complete original alias schema and exercise valid/invalid discriminator inputs, complete large values, direct alias dispatch, missing authority, and independent children.
- Real final-head `generateChatResponse` → AgentRuntime → PGlite + native embeddings + Cerebras: actual write followed by actual read, persisted exact value and final model reply inspected. Complete 117,079-character nested schema guidance reached both planner requests. No mocks replaced the runtime, provider, database, or action dispatcher.
- An unconfigured-caller control was denied with zero record effects. The authorized fixture completed; six requests returned HTTP200.
- UI/native playback evidence: N/A; this change affects the core planner's provider request representation.

The synthetic fixture uses one persisted action family and is not completion evidence for the broader eight-step owner workflow. That workflow now reaches tool dispatch without overflow but has a separate scoped Todo-authority issue tracked in #30959. No latency-improvement percentage is claimed.

<details><summary>Actual final-head input, effect receipts, provider wire summary, and final reply</summary>

```json
{
  "source": "33c3cb1b7eefc77a0e48b0600c7cab77f85f5014",
  "provider": "cerebras",
  "model": "qwen-3.8-27b",
  "prompt": "Use RECORDS to execute write_1 with key audit-proof and payload text exactly \"Copper meadow retains every record boundary.\". Then read audit-proof using RECORDS. Report the exact persisted value after reading it. This is a synthetic isolated record fixture; complete both operations now.",
  "operations": [
    {
      "operation": "write_1",
      "key": "audit-proof",
      "value": "Copper meadow retains every record boundary."
    },
    {
      "operation": "read",
      "key": "audit-proof",
      "value": "Copper meadow retains every record boundary."
    }
  ],
  "persisted": "Copper meadow retains every record boundary.",
  "finalReply": "Done. I wrote audit-proof with the exact text and read it back. The persisted value is: \"Copper meadow retains every record boundary.\"",
  "httpStatuses": [
    200,
    200,
    200,
    200,
    200,
    200
  ],
  "reportedTotalTokens": 130117,
  "schemaGuidanceCharacters": 117079,
  "schemaGuidanceSha256": "153c5fb7189cccacc26ea7b32cb5a512bd9966f4d798a8b1511c07a8a5caf968",
  "plannerRequests": [
    {
      "toolNames": [
        "RECORDS",
        "RECORDS_READ",
        "REPLY",
        "IGNORE",
        "STOP"
      ],
      "requestCharacters": 269983,
      "fullSchemaGuidancePresent": true
    },
    {
      "toolNames": [
        "RECORDS",
        "RECORDS_READ",
        "REPLY",
        "IGNORE",
        "STOP"
      ],
      "requestCharacters": 271137,
      "fullSchemaGuidancePresent": true
    }
  ]
}
```

The complete synthetic schema description is deterministically `"Complete schema guidance starts. " + "The record text is stored without changing its meaning or bytes. ".repeat(1800) + "Complete schema boundary CANONICAL-SCHEMA-END."`; its exact SHA256 is recorded above. This shared property is referenced losslessly by the generated aliases and remains complete in each planner request.

</details>

Hosted checks for this exact head remain required before merge. The broader issues remain open until their independent acceptance criteria are met.

### DIFF
--- a/packages/core/src/__tests__/message-runtime-stage1.test.ts
+++ b/packages/core/src/__tests__/message-runtime-stage1.test.ts
@@ -3007,7 +3007,7 @@ describe("runV5MessageRuntimeStage1", () => {
 		}
 	});
 
-	it("hard-enforces an umbrella candidate when retrieval exposes only its promoted child", async () => {
+	it("hard-enforces an umbrella candidate through its canonical operation schema", async () => {
 		const runtime = makeRuntime([
 			stage1Response({
 				thought: "A repository review requires delegated coding work.",
@@ -3031,8 +3031,8 @@ describe("runV5MessageRuntimeStage1", () => {
 				toolCalls: [
 					{
 						id: "spawn-reviewer",
-						name: "TASKS_SPAWN_AGENT",
-						args: { task: "Review PR 18106." },
+						name: "TASKS",
+						args: { action: "spawn_agent", task: "Review PR 18106." },
 					},
 				],
 			},
--- a/packages/core/src/actions/promote-subactions.ts
+++ b/packages/core/src/actions/promote-subactions.ts
@@ -85,7 +85,7 @@ interface PromotedAction extends Action {
 
 /**
  * The umbrella parent's `routingHint` for a promoted virtual, carried on the
- * non-enumerable promotion marker rather than on the virtual itself so tool
+ * symbol promotion marker rather than on the virtual itself so tool
  * rendering (which prepends `routingHint` to each tool's description) never
  * duplicates it across every virtual. The planner's routing-hints block reads
  * it through this accessor and dedupes by parent, so a promoted family like
@@ -97,6 +97,11 @@ export function promotedParentRoutingHint(
 	const marker = (action as PromotedAction)[PROMOTED_MARKER];
 	const hint = marker?.parentRoutingHint?.trim();
 	return marker && hint ? { parent: marker.parent, hint } : undefined;
+}
+
+/** Returns the registered umbrella identity for a generated dispatch alias. */
+export function promotedSubactionParent(action: Action): string | undefined {
+	return (action as PromotedAction)[PROMOTED_MARKER]?.parent;
 }
 
 /**
@@ -441,15 +446,17 @@ export function promoteSubactionsToActions(
 			connectorAccountPolicy: parent.connectorAccountPolicy,
 			accountPolicy: parent.accountPolicy,
 		};
+		// Symbol metadata survives owner/context gate object spreads while JSON
+		// serializers still omit it from model-facing action descriptions.
 		Object.defineProperty(virtual, PROMOTED_MARKER, {
 			value: {
 				parent: parent.name,
 				virtuals: [virtualName],
 				parentRoutingHint: parent.routingHint,
 			},
-			enumerable: false,
+			enumerable: true,
 			configurable: false,
-			writable: false,
+			writable: true,
 		});
 		return virtual;
 	});
--- a/packages/core/src/services/message.canonical-tools.test.ts
+++ b/packages/core/src/services/message.canonical-tools.test.ts
@@ -1,0 +1,260 @@
+/**
+ * Exercises lossless promoted-family tool rendering and dispatch against real
+ * action handlers. No model or connector is called; provider wire and live
+ * planner behavior are verified separately by the integration workflow.
+ */
+import { describe, expect, it } from "vitest";
+import { actionToJsonSchema, type JsonSchema } from "../actions/action-schema";
+import { promoteSubactionsToActions } from "../actions/promote-subactions";
+import { validateSchema } from "../actions/validate-tool-args";
+import { AgentRuntime } from "../runtime";
+import { createContextObject } from "../runtime/context-object";
+import type { Action, ActionParameter } from "../types/components";
+import {
+	collectActionsFromContext,
+	collectCanonicalPlannerActions,
+	collectPlannerTools,
+} from "./message/planned-tool";
+
+function fixture() {
+	const stored = new Map<string, string>();
+	const parent: Action = {
+		name: "RECORDS",
+		description: "Create and update complete records.",
+		parameters: [
+			{
+				name: "action",
+				description: "Operation",
+				required: true,
+				schema: { type: "string", enum: ["create", "update"] },
+			},
+			{
+				name: "id",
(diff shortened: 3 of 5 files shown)


ss251 (approved):
Reviewed head 33c3cb1b7eefc77a0e48b0600c7cab77f85f5014. Verified by reading the diff and tracing the gate branches; nothing executed.

Risky hunk: the PROMOTED_MARKER flip at promote-subactions.ts:457-459. The comment's claim holds: object spread copies own enumerable symbol keys, so the fixture's `{...action}` admission clones keep parentage — under the old `enumerable: false` they wouldn't, and collectCanonicalPlannerActions would classify every gate-cloned virtual as independent, silently disabling this dedup; JSON.stringify omits symbol keys regardless, so model-facing JSON stays clean. Also traced the orphan fallthrough at planned-tool.ts:658: parent undefined → `!parent?.subActions?.some(...)` is `!undefined` → true, so denied-umbrella aliases stay direct, matching the new test.

Non-blocking: the propertyOverrides JSON.stringify comparison is key-order sensitive; equal schemas can emit redundant overrides (safe direction).

Recommendation: merge

attentionhead (approved):
Verified at head `33c3cb1b7eefc77a0e48b0600c7cab77f85f5014` (test-merged clean onto `origin/develop`). APPROVE — the contract this PR claims holds under an independent probe, and the one failing suite on the merged tree is pre-existing.

## What I checked

**Test-merge.** `git checkout --detach origin/develop && git merge 33c3cb1b` merges cleanly; every check below ran on that merged tree with `npx vitest run --config packages/core/vitest.config.ts`.

**Existing suites on the merged head**
- `message.canonical-tools.test.ts` — 8 pass
- `message-runtime-stage1.test.ts` — 229 pass / 0 fail
- `promote-subactions.test.ts` — 30 pass
- `action-retrieval.test.ts` — 88 pass
- `message.candidate-surface.test.ts` — 17 pass
- `planner-loop.test.ts` — 224 pass / **1 fail**: `v5 planner loop skeleton > records native turn scope without passing it to the action`.

That single failure is **pre-existing, not caused by this PR**. I baselined it detached at `origin/develop` with the PR unmerged and got the identical 224 pass / 1 fail on the same test name, then re-merged. Nothing to fix here.

**Independent probe.** I wrote a throwaway vitest file (not proposed for the tree) that builds a 5-subaction umbrella via `promoteSubactionsToActions`, with a ~4000-char parent description and 40 nested schema fields, and passes the promoted set through `.map(a => ({...a}))` to simulate the spread that the owner/context gate performs before the planner sees it. Results on the merged head:

- wire set is `['RECORDS']` alone, from `['RECORDS','RECORDS_CREATE','RECORDS_UPDATE','RECORDS_ARCHIVE','RECORDS_RESTORE','RECORDS_PURGE']`
- serialized planner surface **87960 → 14649 bytes, ratio 0.167** — the duplicate-schema elimination is real and large, not incidental
- an alias named as an explicit candidate stays direct: wire `['RECORDS','RECORDS_CREATE']`
- a partially authorized family keeps every alias direct (no silent capability loss when the umbrella cannot stand in for the whole family)
- an alias whose parent is absent from the set stays direct
- `JSON.stringify` of the wire actions contains neither `parentRoutingHint` nor `virtuals`

The same probe run against `origin/develop` alone fails all six cases with `TypeError: collectCanonicalPlannerActions is not a function`, i.e. the behaviour is new here and not something develop already did.

## Notes on the design, for the record

`PROMOTED_MARKER` moving from `enumerable: false, writable: false` to `enumerable: true, writable: true` looks like a leak at first glance, but it is the load-bearing part of the change: object spread copies only *enumerable own* properties, symbol keys included, so a non-enumerable symbol marker is silently dropped the moment any layer does `{...action}` — and the pipeline does exactly that before `collectPlannerTools` runs. A non-enumerable marker would have made `collectCanonicalPlannerActions` see every alias as an orphan and keep the whole family direct, quietly restoring the duplicate schemas. Serialization is unaffected because `JSON.stringify` omits symbol-keyed properties regardless of enumerability, which the probe confirms directly.

The exclusion set is the right shape. Collapsing to the umbrella is only safe when the umbrella can actually reach every child, so keeping aliases direct for partial authorization, unlisted children, explicit candidates, and parents that require a parameter the alias does not carry are each necessary, not defensive extras. The `propertyOverrides` description only carries properties whose serialization differs from the parent's, so the umbrella tool text does not re-inflate with the schemas the change just removed.

No defect found.

lalalune:
Final merge verification completed on preview `4ebdb93545ad6`, combining current develop `ce37cd9bc92dad251af73a0d72a595f05238ad64` and the published canonical-tool patch. Normal pinned install and root verification passed. Full core suite: 12,093 passed, three declared skips across 933 files. All five hosted checks passed for published head `33c3cb1b7eefc77a0e48b0600c7cab77f85f5014`.

Independent review confirms that shared parameter references reconstruct the complete normalized wire schema and that generated aliases retain required fields. The final-head real model write/readback evidence is inline above. This completes this representation change; broader workflow and reply-recovery acceptance remains tracked separately.

Issue #30965, opened 2026-09-10
Preserve the final reply after a corrected calendar read succeeds


lalalune:
A real local workflow created an undated todo, then retried a malformed calendar read successfully. The planner evaluator produced a correct final reply with the durable todo ID and calendar results. The final-message failure pass replaced it with an instruction asserting the calendar step still had no usable result. The delivered rewrite omitted the requested todo ID.

Reproduction shape: CALENDAR feed is first called with valid timeMin/timeMax/timeZone and an extra nested `details.__eliza_record_entries` property. Validation rejects that extra property before execution. The next CALENDAR feed uses the exact same action, intent, timeMin, timeMax and timeZone, with only the rejected property removed, and succeeds. The first validation failure remains unresolved because malformed-call correlation treats the rejected representation metadata as target content. The preserved successful evaluator reply is then discarded.

Observed through real AgentRuntime, native local embeddings, Cerebras qwen-3.8-27b and PGlite. The owner create returned its durable ID and unscheduled cadence; both the successful evaluator reply and the later incorrect rewrite were recorded. No external calendar or messaging operation was used. This final-message code is unchanged between current develop and the canonical-tool representation patch.

Acceptance:
- A successful correction of the same logical read resolves the malformed pre-dispatch attempt, preserving the verified evaluator reply and requested durable IDs.
- Keep the complete failed attempt and correction in the trajectory.
- Preserve existing protections against treating another resource, query, destructive operation or unrelated success as recovery.
- Add a real planner regression for this sequence, not just a literal comparison test, and repeat live reply verification.

Related to the dependent workflow acceptance in #24299 and unnecessary reply-synthesis calls in #17072.

lalalune:
The redundant-metadata subset is merged in #30967 and verified on develop `8e292e1c4c4a3823242768d3ef966f4e5e3aabec`. This issue remains open.

A further real local run reproduced the same final-reply symptom for a different rejected representation: a calendar feed carried `details.__eliza_record_entries` containing an event ID as an unquoted raw string, without a matching sibling. The corrected feed kept the same action, intent and date bounds, removed the rejected marker, and returned the exact requested event with the final times and todo ID in its description. The successful evaluator reply included both durable IDs and completed status, but final delivery still said the calendar verification had not finished.

[...]

PR #30966, opened 2026-09-10
Preserve authored undated todo instructions and failed-reply authority

lalalune:
A numbered request that creates an undated todo and then schedules a separate calendar event could ask for a todo deadline, because handler normalization removed the list boundaries and treated the event time as todo authority. Failed action reply drafts could also be marked verified and imply that an unsuccessful save had completed.

Preserve the authored list boundaries for deterministic operation attribution, retain corrections and ambiguous references, and derive a missing unscheduled cadence only from explicit authority attached to the selected todo. Failed model drafts remain complete internal diagnostics rather than verified user-facing replies. Existing explicit previews, draft editing and separate todo targets keep their guards.

Related: #30959, #24299.

Validation: pinned Node 24.15.0 / Bun 1.3.14 normal install and root verification passed at `5ca1f3905b79d138230589ff0d9d458f04b035e4`. The identical personal-assistant package tree passed all 299 unit files (2,578 tests, 6 declared skips), and is being rerun on this final combined base before merge. 222 focused cases and four real PGlite lifecycle cases passed, including supplied/omitted cadence and a later scheduling correction. Exact-base controls reproduce the lost-boundary and missing-cadence failures.

A confined real-model/PGlite host created the named owner todo with an unscheduled cadence and returned its durable ID. That run separately exposed final-message recovery issue #30965 after a calendar retry; this PR does not claim that broader reply/workflow issue is resolved. Two baseline fixtures were repaired without weakening assertions: evaluator persisted-memory input and actual todo-route state/owner isolation.

### DIFF
--- a/plugins/plugin-personal-assistant/src/actions/lib/undated-todo-intent.test.ts
+++ b/plugins/plugin-personal-assistant/src/actions/lib/undated-todo-intent.test.ts
@@ -8,6 +8,7 @@ import { describe, expect, it, vi } from "vitest";
 import { extractLifeOperationWithLlm } from "./extract-life-operation.js";
 import { extractTaskCreatePlanWithLlm } from "./extract-task-plan.js";
 import {
+  resolveUndatedTodoAuthority,
   textContradictsExplicitUndatedTodo,
   textStatesExplicitUndatedTodo,
   UNDATED_TODO_EXTRACTION_GUIDANCE,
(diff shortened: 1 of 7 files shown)


attentionhead (approved):
Verified at head `5ca1f3905b79d138230589ff0d9d458f04b035e4` (test-merged clean onto `origin/develop`). APPROVE — the list-boundary attribution is real and measurable, and the whole package is green on the merged tree where develop alone is not.

## What I checked

**Test-merge.** `git checkout --detach origin/develop && git merge 5ca1f390` merges cleanly. Everything below ran on that merged tree with `npx vitest run --config vitest.config.ts` from `plugins/plugin-personal-assistant` (the package's own config — the root config and `bun test` both mis-resolve here).

**Whole package on the merged head:** `Test Files 299 passed (299)`, `Tests 2578 passed | 6 skipped (2584)`, 0 failures, 330s. The four suites this PR touches (`undated-todo-intent`, `life-reminder-datetime`, `lifeops-todos-unscheduled`, `ftu-evaluators-merged-call`) are 226 passing on their own.

**The same package on `origin/develop` with the PR unmerged** fails `src/routes/lifeops-todos-unscheduled.test.ts` (2 tests, `TypeError: service.getTodos is not a function` at `lifeops-routes.ts:2769`) and `test/ftu-evaluators-merged-call.test.ts` (2 tests, `TypeError: runtime.getMemories is not a function` via `evaluator-transcript.ts:37` → `evaluator.ts:925`). So the "two baseline fixtures were repaired" claim in the description is accurate and I confirmed it in both directions rather than taking it on the diff — 4 failed → 0 failed is the measured delta, and the repairs are to the harnesses' runtime doubles, not to the assertions.

**Independent probe.** I wrote a throwaway vitest file (not proposed for the tree) driving `resolveUndatedTodoAuthority` directly, on a two-item authored workflow: item 1 creates an owner todo titled "Rebalance portfolio" *with no date*, item 2 creates a calendar event "Dentist" on Friday at 3pm. On the merged head:

- `resolveUndatedTodoAuthority(text, "Rebalance portfolio")` → `{explicit: true, contradicts: false, operationScoped: true}` — the sibling calendar time does not schedule the todo.
- `textContradictsExplicitUndatedTodo(text)` on the *same* text still returns `true`. That is the defect stated plainly: the pre-existing whole-text reading treats the event's Friday 3pm as todo authority, and it is exactly what the new scoped path replaces. The probe asserts both, so the before/after is visible in one file.
- Adding a third item "Actually, make that todo due next Monday" flips the scoped result to `{explicit: false, contradicts: true}` — a later correction to the same todo still wins, which is the property most at risk from a boundary-scoping change.
- The fallbacks hold: a null title, a title that appears nowhere, two list items creating the same named todo, and a single-item (non-list) message all return `operationScoped: false` and fall back to whole-text semantics.

Run against `origin/develop` alone, five of the six probe cases fail with `TypeError: resolveUndatedTodoAuthority is not a function`, and the sixth — the `textContradictsExplicitUndatedTodo` case — passes, confirming the defect is present there and is not something develop already handled.

## Notes on the design

The scoping is narrow in the right direction: it bails to whole-text on every ambiguity (no title, unmatched title, `selected.length !== 1`, more than one creation in the selected block, fewer than two list items), so the new path can only ever *withhold* an attribution the old code made too eagerly, never invent one. That is the safe failure mode for a write-authorization projection.

The header comment earns its place by naming the invariant the repo's prompt-integrity rule cares about: this is a projection used only to decide write authority, and the complete user message still reaches every model prompt. The `authorityClauses` filter keeps cross-todo references, pronoun/title corrections, and unbound later directives rather than dropping them, so a correction cannot silently disappear into a discarded clause — which is the specific way a filter like this would otherwise become a truncation.

`verifiedUserFacing: false` on a failed draft in `life.ts` is the other half and is straightforwardly correct: a draft produced by a failed action is diagnostic, and marking it verified would let an unsuccessful save read to the user as completed.

The description is also appropriately scoped — it explicitly declines to claim #30965 (final-message recovery) is resolved by this PR.

No defect found.

lalalune:
Final published-head validation for `5ca1f3905b79d138230589ff0d9d458f04b035e4`:

- Normal pinned-toolchain install and root `bun run verify` passed.
- Full personal-assistant suite: 299 files, 2,578 passed, 6 declared skips; owning typecheck and lint passed.
- Four real PGlite lifecycle/ownership cases passed, including an explicitly undated authored todo beside later timed calendar steps with supplied or omitted cadence.
- A confined real-model host run saved an undated owner todo and returned its actual ID `cc296e4c-b263-46e0-b3f6-9898e6e23cce` together with the subsequent calendar result. The run used the same personal-assistant patch plus separately reviewed integration changes; this is not proof of the entire umbrella issue.

The eight-operation run performed its effects but hit the provider limit during final reply generation. That failure is retained and is not counted as complete end-to-end acceptance. Recovered-read reply handling is tracked separately in #30965 / #30967.

lalalune:
Merged with exact-head lock after all five hosted checks passed at `5ca1f3905b79d138230589ff0d9d458f04b035e4` (run34432229003). Merge commit: `88106f05234d3b77c3e30316d78673fab0df53c1`, 2026-09-10T03:46:35Z. The complete merged tree equals the locally validated published head; post-merge comparison passed.

Package and real database/model evidence is recorded above. The separate completed-record review and partial-workflow reply defects remain tracked in #30968 and #30970. This merge does not close the umbrella workflow issue.

PR #30967, opened 2026-09-10
Preserve verified replies after redundant record arguments are corrected

lalalune:
A calendar read rejected a redundant native-record representation field, then succeeded with the same date window and timezone after that field was removed. The stale validation failure still replaced the evaluator's correct final reply, dropping a requested durable ID and spending another model call on failure synthesis.

For malformed-call correlation only, recognize a rejected JSON-string record wrapper as redundant when every encoded key resolves to exactly one surviving sibling and its decoded value matches. Different targets, malformed values and ambiguous keys keep the original failure authority. The complete original arguments and both outcomes remain in the trajectory.

Related: #30965, #24299, #17072.

Validation at `4c4c62999ba138166fde3aebc147ca6e37f5b170`: pinned normal install, root verification, core typecheck/lint, 41 focused planner cases, and full core suite (12,096 passed, three declared skips). The unchanged-base control fails the new planner regression.

The complete captured model/tool sequence from the actual failing local workflow was replayed through the real planner, retaining its continuation markers. The base discarded the verified final reply; the repair preserved it without another synthesis call. This is recorded-model regression evidence, not a fresh end-to-end model run.

<details><summary>Recorded sequence result</summary>

```json
{
  "baseModelCalls": 4,
  "fixedModelCalls": 3,
  "executedOperations": 3,
  "outcomes": [
    true,
    false,
    true
  ],
  "preservedFinalReply": "Done. The todo 'P19 owner scope final cadence check' is saved with no due date or schedule, ID 2ab208ae-eedb-44ef-8ef0-cfe504d87e28.\n\nYour local Eliza calendar for Sep 10, 2026 shows one event:\n- P19 copper meadow dependency review, 11:00 AM to 11:45 AM, with the description 'P19 copper meadow dependency review (todo ID: 00ae5c5a-2d39-40d1-91af-15c93e717e03)'."
}
```
</details>

A separate confined live owner-todo/calendar request delivered the actual saved todo ID and verified calendar times. A literal-JSON error fixture exposed input redaction of record keys; that fixture is retained and is not counted as positive proof for redundant-key recovery.

### DIFF
--- a/packages/core/src/runtime/__tests__/planner-loop-failure-correlation.test.ts
+++ b/packages/core/src/runtime/__tests__/planner-loop-failure-correlation.test.ts
@@ -1151,3 +1151,95 @@ describe("planner-loop failed-operation correlation", () => {
 		expect(runtime.useModel).toHaveBeenCalledTimes(3);
 	});
 });
+
+it.each([
+	["redundant native record metadata", "2026-09-10T00:00:00", true],
+	["different encoded target", "2026-09-12T00:00:00", false],
+	["malformed encoded value", undefined, false],
+] as const)(
+	"retains the correct calendar result after %s",
+	async (_label, encodedStart, recovered) => {
+		const details = {
+			timeMin: "2026-09-10T00:00:00",
+			timeMax: "2026-09-11T00:00:00",
+			timeZone: "America/New_York",
+		};
+		const corrected = {
+			action: "feed",
+			intent: "Read the local calendar for September 10.",
+			details,
+		};
+		const runtime = {
+			useModel: vi
+				.fn()
+				.mockResolvedValue({
+					text: "The failed read remains unresolved.",
+					toolCalls: [],
+				})
+				.mockResolvedValueOnce(
+					plannerToolCall("invalid", "CALENDAR", {
+						...corrected,
+						details: {
+							...details,
+							__eliza_record_entries: [
+								{
+									key: "timemin",
+									value:
+										encodedStart === undefined
+											? "invalid JSON"
+											: JSON.stringify(encodedStart),
+								},
+								{ key: "timemax", value: JSON.stringify(details.timeMax) },
+								{ key: "timezone", value: JSON.stringify(details.timeZone) },
+							],
+						},
+					}),
+				)
+				.mockResolvedValueOnce(
+					plannerToolCall("corrected", "CALENDAR", corrected),
+				),
+		};
+		const verifiedReply =
+			"The calendar read succeeded. Event evt-copper is scheduled for 11:00 to 11:45 AM.";
+		const result = await runPlannerLoop({
+			runtime,
+			context: { id: "corrected-calendar" },
+			executeToolCall: vi
+				.fn()
+				.mockResolvedValueOnce({
+					success: false,
+					error: "Unexpected argument 'details.__eliza_record_entries'",
+				})
+				.mockResolvedValueOnce({
+					success: true,
+					data: {
+						events: [{ id: "evt-copper", start: "11:00", end: "11:45" }],
+					},
+				}),
+			evaluate: vi
+				.fn()
+				.mockResolvedValueOnce({
+					success: false,
+					decision: "CONTINUE",
+					thought: "Remove the rejected representation field and retry.",
+				})
+				.mockResolvedValueOnce({
+					success: true,
+					decision: "FINISH",
+					thought: "The corrected read returned the event.",
+					messageToUser: verifiedReply,
+				}),
+		});
+		if (recovered) {
+			expect(result.finalMessage).toBe(verifiedReply);
+			expect(runtime.useModel).toHaveBeenCalledTimes(2);
+		} else {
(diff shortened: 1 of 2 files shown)


ss251 (approved):
Reviewed head 4c4c62999ba138166fde3aebc147ca6e37f5b170.

Risky hunk: `withoutRedundantRejectedRecordEncoding` (planner-loop.ts:5383). I traced every bail-out — regex miss, non-object ancestor, empty/non-array entries, entry shape != {key,value} strings, sibling count != 1, JSON.parse throw, value mismatch — each returns `params` unchanged, so any doubt keeps the full rejected args and the failed call stays non-superseded (conservative pre-PR behavior). The `ancestors.reverse()` rebuild correctly rewraps the cleaned object under `details`, and the 09-12-vs-09-10 and unparseable-JSON fixtures pin both refusal exits.

Non-blocking: the dot-split assumes no dotted param keys; that path also fails safe.

Verified by tracing the projection branch-by-branch against all three fixtures in the diff at this head.

Recommendation: merge

attentionhead (approved):
Verified at head `4c4c62999ba138166fde3aebc147ca6e37f5b170` (test-merged clean onto `origin/develop`). APPROVE — the projection is correctly bounded, and I reproduced both the recovered behaviour and every refusal case it must preserve.

## What I checked

**Test-merge.** `git checkout --detach origin/develop && git merge 4c4c6299` merges cleanly; everything below ran on that merged tree via `npx vitest run --config packages/core/vitest.config.ts`.

**Suites on the merged head:** `planner-loop-failure-correlation.test.ts` 24 pass, `planner-loop.test.ts` 225 pass — 249 total, 0 failures. `planner-loop.test.ts` is also 225/225 detached at `origin/develop`, so this PR neither fixes nor breaks anything there.

**Independent probe.** I wrote a throwaway vitest file (not proposed for the tree) calling `malformedCallSupersededBy` directly with the real rejection text — `Unexpected argument 'options.__eliza_record_entries' for tool CALENDAR_READ` — against a later successful call carrying the same `{start, timezone}` window. Seven cases on the merged head:

1. every encoded key duplicates a surviving sibling, values equal → **correlates** (the recovery this PR adds);
2. an encoded value differs from its sibling (`"2026-09-11"` vs `"2026-09-10"`) → does not correlate;
3. the encoded value is malformed JSON (`{not json`) → does not correlate;
4. the encoded key has no surviving sibling (`end`) → does not correlate;
5. the parent holds both `start` and `Start`, so the case-insensitive key match is ambiguous → does not correlate;
6. the later call targets a genuinely different window (`2026-12-25`) → does not correlate;
7. the failure names a different argument (`options.bogus`) → the projection does not engage at all.

Run against `origin/develop` alone, case 1 fails (`expected false to be truthy`) and cases 2–7 pass. That is the exact shape the change should produce: one new recovery, and every pre-existing refusal preserved bit-for-bit rather than merely re-asserted.

(My first probe pass had case 1 failing on the merged head too. That was my harness — I was passing an array where `malformedCallSupersededBy` takes a single `call`. Corrected, then re-run in both directions; the numbers above are the corrected runs.)

## Notes on the design

`withoutRedundantRejectedRecordEncoding` is gated the right way round. It returns `params` unchanged on every path where redundancy is not provable — no `__eliza_record_entries` named in the error, a missing or non-record ancestor, an empty or non-array entries list, an entry that is not exactly `{key, value}` strings, `matching.length !== 1`, a `JSON.parse` throw, or an unequal decoded value. So the worst case is that a genuinely redundant wrapper is *not* recognised and the old failure authority stands, which is the safe direction for a correlation relaxation.

The `matching.length !== 1` check is the part worth calling out: keys are compared case-insensitively, so two siblings differing only in case would make the encoded key ambiguous about which value it duplicates. Refusing there rather than picking one is correct, and case 5 of my probe pins it.

The `// error-policy:J3` comment on the `JSON.parse` catch is the right category and reads accurately — invalid transport JSON produces an explicit "cannot prove redundant" result, not a fake-valid default.

Scope-wise this only reshapes the local `failedParams` used for correlation; `failedCall.params` itself is untouched, so the complete original arguments and both outcomes stay in the trajectory, matching the repo's prompt-integrity rule and what the description claims. The description is also honest about its limits — it declines to claim #30965 is resolved, and it explicitly says the literal-JSON error fixture is not counted as positive proof.

No defect found.

lalalune:
Final merge verification: published head `4c4c62999ba138166fde3aebc147ca6e37f5b170` has all five hosted checks successful (run 34432700872). Current-develop merge preview `cba5130a8a4fd6b78b1c74a17450c342e46e6bed` combines that head with `88106f05234d3b77c3e30316d78673fab0df53c1`; normal pinned install and root verify passed, and full core passed 12,096 tests with 3 declared skips across 933 files. The complete recorded recovered-calendar trajectory regression and conservative distinct-target/malformed-metadata controls remain passing. This verifies the bounded recovery fix; the broader eight-step final-delivery acceptance remains tracked separately in #30974 and #24299.

PR #29880, opened 2026-08-29
test(core): pin postpositive current-view inspection and declarative controls

hermesagent270-commits:
## Scope change at head 6a18c74116b4

**This PR is now test-only.** `develop` `a4e28d5b126` ("fix(core): keep natural view questions off planner path", 2026-08-29) already routes every postpositive form in #29531 to the direct reply path through the second inspection matcher, so the regex widening this PR carried pinned nothing against the merge base and, measured against it, dropped the VIEWS candidate on ordinary asks containing a view-state subordinate clause. The production hunk is removed; the four postpositive fixtures and the declarative negative controls stay as regression pins for the behaviour `develop` now has. Whether #29531 should close on `a4e28d5b126` is a maintainer call.

## Summary

Closes #29531. Supersedes #29544 with its focused current-view fix rebased onto current `develop`, plus a correction for declarative-state overmatching.

Natural inspection questions such as `which view is open?` were promoted to the full planner because the heuristic only recognized state adjectives before the view noun. The postpositive grammar now recognizes the question form while remaining anchored to an actual inspection/interrogative token.

## Correctness repair

The earlier postpositive alternative was unanchored and classified ordinary statements such as `the window is open` as inspection. Because `open` inside an inspection span is exempted from action detection, those statements could incorrectly bypass normal routing.

This replacement:

- keeps `which view/window/screen is open/active` read-only;
- requires `identify`, `name`, `tell me`, `what`, or `which` before both grammar forms;
- adds negative controls for declarative state, quoted text, conditionals, and mixed actionable requests;
- retains the existing actionable compound matrix.

Original focused implementation authorship is preserved from @mtcow / #29544.

## Verification

- `git diff --check`: passed.
- Static exact-head audit confirms the postpositive alternative is inside the inspection-token prefix group and cannot begin at a bare view noun.
- Local suite execution is not claimed because this checkout lacks Eliza's dependency closure. Hosted exact-head CI is required before merge.

## Evidence gate

- Screenshots/video: N/A; no rendered surface changes.
- Backend evidence: focused heuristic regressions included; hosted exact-head logs pending.
- Real-model trajectory: still required before merge for the claimed Cerebras overflow/runtime-routing outcome; this PR does not claim that unit tests prove provider behavior.
- Domain artifacts: N/A.

AI provider/model: openai / gpt-5.6-sol
Client / agent tooling: Codex
Attribution status: self-reported
— hermesagent270-commits

PR #30973, opened 2026-09-10
test(core): reject declarative view state as inspection

lalalune:
This test-only change prevents declarative, conditional, or quoted view state from being mistaken for a read-only inspection request. It adds five inputs to the actual routing-function suite. The consumer is Stage 1: misclassifying these messages can suppress normal view-action consideration before the planner runs.

Existing coverage already owns natural “which view is open?” questions and compound inspection-plus-navigation requests, so those duplicate additions were removed. The retained controls protect the requirement for an actual interrogative phrase in the predicate-order inspection grammar.

The quality check changed the production grammar in a disposable checkout by removing that interrogative requirement. All 121 existing heuristic tests and all four higher-level Stage-1 current-view cases still passed. The five retained cases failed on the same mutant. Restoring production source made all 126 focused tests pass. No mutated production source is part of this PR.

Develop already contains the production behavior in a4e28d5b1263f9af1d17477f97094f2a04ca2cdd. This PR makes no provider-overflow, runtime-routing repair, or live-model outcome claim.

Replacement for #29880; refs #17072. Independent review rebased this PR onto develop `88106f05234d3b77c3e30316d78673fab0df53c1`; current head is `b646adb020f62c7773df8d3575d60ad81e14dac3`. In an isolated worktree with private dependencies and pinned Bun 1.3.14 / Node 24.15.0, normal `bun install`, root `bun run verify`, owning core typecheck, core lint (no errors), and all 126 focused heuristic tests passed. The reviewer independently reproduced the mutation control: five added cases fail while all 121 pre-existing heuristic cases and all four higher-level Stage-1 cases pass. Production source was restored and the working tree is clean. The full core suite also passed on this head: 12,098 passed, 3 declared skips across 933 files (931 passed / 2 skipped), exit 0 in 943.33 seconds. Exact-head hosted checks remain the merge gate.

Screenshots, video, native capture, and domain side effects: N/A; only deterministic parser regression tests change. Live-model evidence: N/A for this test-only scope; no provider or action implementation changes.

### DIFF
--- a/packages/core/src/services/message/direct-action-heuristics.test.ts
+++ b/packages/core/src/services/message/direct-action-heuristics.test.ts
@@ -1054,6 +1054,21 @@ describe("inferDirectCurrentRequestCandidateInference kinds", () => {
 	});
 
 	it.each([
+		"the window is open",
+		"this window remains open",
+		"the current view stays open",
+		"if the window is open, leave it alone",
+		'preserve the text "the window is open"',
+	])(
+		"does not mistake declarative view state for inspection: %s",
+		(message) => {
+			expect(
+				inferDirectCurrentRequestCandidateInference([viewsAction], message),
+			).toEqual({ names: ["VIEWS"], kind: "view-surface" });
+		},
+	);
+
+	it.each([
 		[
 			"create / and / inspection-first",
 			"identify the current view and add a new panel",


ss251 (approved):
Reviewed head 39348ad77de8c05a28aaf89608e098003d8cf55b.

Verified the new cases at packages/core/src/services/message/direct-action-heuristics.test.ts:1056-1069 by reading the diff at this head. The load-bearing detail is `.toEqual({ names: ["VIEWS"], kind: "view-surface" })` at :1067 — exact object equality, not `toMatchObject` — so the `kind` field is fully pinned: any regression that returns an inspection kind for these inputs fails, which is what makes the title's claim enforceable. Sharpest input is :1061, `preserve the text "the window is open"` — it guards the quote-extraction path where an interrogative-shaped substring could be promoted to a request candidate.

Non-blocking: all five run against `[viewsAction]` alone (:1066); if the inspection branch only fires when inspection-capable actions are in the candidate list, these cases can't distinguish rejection from unreachability — a multi-action control would close that gap.

Recommendation: merge

attentionhead (approved):
Verified at head `39348ad77de8c05a28aaf89608e098003d8cf55b` (test-merged clean onto `origin/develop`). APPROVE — this is a test-only PR that actually clears the repository's test-quality gate, and I confirmed that by reproducing the mutant rather than taking the description's word for it.

## What I checked

**Test-merge.** `git checkout --detach origin/develop && git merge 39348ad7` merges cleanly. The change is +15/-0 in one file: five `it.each` inputs on `inferDirectCurrentRequestCandidateInference`, all asserting `{names: ["VIEWS"], kind: "view-surface"}`.

**On the merged head:** `direct-action-heuristics.test.ts` 126 pass, 0 fail.

**Mutation check, run independently.** The claim that matters for a test-only PR is that these cases catch a real regression the existing suite does not. So I applied the mutant the description names — dropping the interrogative requirement — directly to `looksLikeCurrentViewInspection` in a disposable checkout, by making the leading `\b(?:identify|name|tell\s+me|what|which)\b` optional in the first inspection span and `(?:what|which)\s+` optional in the second. Results:

- `direct-action-heuristics.test.ts`: **5 failed | 121 passed**. Exactly the five new cases fail; every pre-existing case still passes.
- `message-runtime-stage1.test.ts`: **229 passed**, unchanged under the same mutant.

Then I restored the original source (`git diff` clean) and re-ran: 126/126.

That second number is the important one. It substantiates the "why existing higher-level coverage does not own this contract" requirement in CONTRIBUTING with a measurement instead of an assertion — the Stage-1 suite is completely blind to this grammar relaxation, so without these cases the regression ships silently.

## Why these five cases are the right five

They are not literal mirrors of the implementation; each isolates a distinct way declarative state can be mistaken for a question:

- `the window is open` — bare declarative, the base case.
- `this window remains open` / `the current view stays open` — the `remains?|stays?` alternation in the span, reached without an interrogative.
- `if the window is open, leave it alone` — conditional, where the inspection-looking clause is subordinate.
- `preserve the text "the window is open"` — quoted, where the phrase is a payload rather than a request, and an imperative verb precedes it.

The last two are the ones a mirror-the-implementation test would not think to write, and they exercise the interaction with the surrounding `OPEN`-as-adjective exemption and the operation-occurrence scan rather than restating a regex.

The consumer is named and real: misclassification suppresses view-action consideration in Stage 1 before the planner runs, which the file header ties to a second model round and a prompt that may no longer fit the provider context window. That is an observable behaviour with an affected caller, not coverage for its own sake.

## Scope

The description is careful about what it does not claim — no provider-overflow or routing repair, no live-model outcome, production behaviour already on develop at `a4e28d5b`, and explicitly no mutated source in the PR (which I confirmed: the diff touches only the test file). Screenshots/live-model evidence are marked N/A with a reason, which is correct for deterministic parser tests.

No defect found.

lalalune:
Independent review completed for head `b646adb020f62c7773df8d3575d60ad81e14dac3`.

Normal pinned Bun 1.3.14 / Node 24.15.0 installation, root `bun run verify`, core typecheck/lint, and the full core suite passed: 12,098 tests passed, 3 declared skips. The five added tests detect a real routing-grammar regression missed by all 121 pre-existing heuristic cases and four higher-level Stage-1 controls; the production mutation was disposable and is absent from this PR.

After #30967 advanced develop to `8e292e1c4c4a3823242768d3ef966f4e5e3aabec`, a clean merge preview `b0ccda01c03d8e9c49b42332add00209364b7d1c` retained exactly the original 15 added test lines and no production changes. All 126 heuristic cases and four Stage-1 controls passed there. Its tree is `61e17039e93b138b55ca77d4ba09a5a490406013`; the final merge will be checked against that tree. The new develop base itself has a verified full-core run of 12,096 passing tests and 3 declared skips.

Merge is limited to this reviewed head after all five hosted checks in run 34435646295 report success. This is deterministic parser regression coverage; no UI, live-provider, or production behavior change is claimed.

Issue #30970, opened 2026-09-10
Planner outage after failed verification replays earlier success as final reply


lalalune:
In a confined actual-model #24299 workflow, an owner todo completed successfully, the next verification read failed, and the next provider request returned HTTP429. The final streamed reply was only the earlier successful completion message. It omitted both the failed verification and remaining calendar read.

The message-service failure rescue scans backward past failed tool results and selects earlier user-facing success text. That text is valid for one operation but is not a complete result for the failed workflow.

Acceptance: reproduce a successful action followed by a failed verification and provider failure through the real message service. Preserve its explicit terminal failure instead of rescuing older success prose; keep existing single-success/post-evaluator-outage recovery working, and permit a later successful recovery to supersede an earlier failure. Preserve full tool results and avoid repeating effects. Owning tests, typecheck/lint, root verification, and exact-head CI are required.

HomunculusLabs:
CLAIMING: fix the planner-loop failure rescue so a verified terminal failure is not superseded by an earlier operation's success prose when the provider dies mid-workflow.

Pre-claim verification (issue still reproducible at develop tip 88106f05234 — the freshly merged #30966 touched plugin-personal-assistant, not this core seam):

- `preservedSettledToolResult` (packages/core/src/services/message.ts:2149) scans settled tool results backward but `continue`s on `success !== true`, so in the issue's shape `[successful todo completion, failed verification read, provider HTTP-429]` the earlier success `userFacingText` is rescued and the verified failure + remaining calendar read are silently dropped.
- `answerlessToolTurnReport` (message.ts:2301) has the same precedence inversion: any undelivered success outranks a later verified failure, even though `preservedVerifiedFailure` (message.ts:2260) already exists and is only consulted after successes.
- The planner-loop catch (message.ts:10618) never consults a verified failure at all before falling back to relay-body / rethrow.

Plan: newest-eligible-wins scan — undelivered verified success text or an undelivered verified failure (`verifiedUserFacing: true`, confirmation-preview markers excluded from failure authority) — keeping single-success and post-evaluator-outage recovery intact and letting a later success supersede an earlier failure. Regression tests through the real `DefaultMessageService.handleMessage` harness plus RED controls on pristine develop.

RP suspended per operator directive 2026-09-02; verification is own-execution (disposable worktree, focused suites, develop controls).

AI provider/model: zai / glm-5.3
Client / agent tooling: Hermes Agent
[...]

PR #30977, opened 2026-09-10
fix(core): retain later failures when recovering planner replies

lalalune:
A later verification failure followed by a provider outage could cause the final reply to repeat an earlier successful operation. Successful-result recovery now stops at a later failed operation in the planner and message-service fallback paths. A single earlier verified result also loses final-reply authority after a later failure. Earlier failures followed by a successful recovery and single-operation post-evaluator-outage recovery remain supported.

Refs #30970 and #24299.

The defect was observed in a confined actual-model workflow: a todo completed, its verification failed, the provider returned HTTP429, and the final reply only said the todo was marked done. The regression drives the real AgentRuntime, planner, message service, action execution and delivery with controlled model transport. Before the fix it delivered the old success; afterward the verified failure stays visible, with no repeated write. Complete tool data and existing effects remain intact.

Validation at 1a19427ef2eb4674b1cf02bf55222b1f8bbb0c4b: 28 message-service checks and 87 planner reply/happy-path controls passed; full core passed 12,099 tests with 3 declared skips (933 files); normal pinned install and root verify passed, including owning typecheck/lint. Rebased head `b02874a94153510166602e9d479a12160e51eaae` is based on current develop `8e292e1c4c4a3823242768d3ef966f4e5e3aabec`; the complete core package is byte-identical to the full-suite-tested source. Normal pinned install and root verify passed again after rebase. Hosted checks remain required before merge. Screenshots/video/native capture: N/A for this failure-delivery contract; actual emitted reply and model/tool trajectory are the evidence.

### DIFF
--- a/packages/core/src/runtime/planner-loop.ts
+++ b/packages/core/src/runtime/planner-loop.ts
@@ -6862,8 +6862,11 @@ function deterministicSuccessfulToolRelay(
 	trajectory: PlannerTrajectory,
 ): string | undefined {
 	for (const step of [...trajectory.steps].reverse()) {
-		if (!step.toolCall || step.result?.success !== true) continue;
-		if (isTerminalToolCall(step.toolCall)) continue;
+		if (!step.toolCall || !step.result || isTerminalToolCall(step.toolCall))
+			continue;
+		// A failed later read or write prevents an earlier success from owning
+		// the final reply when the provider cannot finish the workflow.
+		if (step.result.success !== true) return undefined;
 		const candidate =
 			getNonEmptyString(step.result.userFacingText) ??
 			(step.result.modelReplyRequired === true
@@ -7099,10 +7102,9 @@ function previewWasCommitted(
  * `verifiedUserFacing` opt-in is unambiguous: exactly one completed tool step
  * set `verifiedUserFacing: true` with a non-empty `userFacingText`.
  *
- * Failed steps are intentionally ignored unless they are explicit
- * confirmation-required previews. A plan whose first tool errored and whose
- * second tool emitted a verified canonical reply must still echo the verified
- * reply. LifeOps can draft more than once while refining a request; the latest
+ * Earlier failed steps do not invalidate a later verified canonical reply.
+ * A failure after the successful step does prevent that earlier reply from
+ * owning the turn. Explicit confirmation-required previews retain authority. LifeOps can draft more than once while refining a request; the latest
  * verified preview is the user-complete state even though `success:false`
  * correctly records that nothing was persisted yet. A later canonical commit
  * for every keyed preview operation releases that preview's reply authority.
@@ -7137,10 +7139,22 @@ export function singleVerifiedUserFacingToolResultText(
 		}
 	}
 
-	const successfulToolSteps = allTrajectorySteps(trajectory).filter(
+	const successfulToolSteps = steps.filter(
 		(step) => step.toolCall && step.result?.success === true,
 	);
 	if (successfulToolSteps.length !== 1) return undefined;
+	const successfulIndex = steps.indexOf(successfulToolSteps[0]);
+	if (
+		steps.some(
+			(step, index) =>
+				index > successfulIndex &&
+				step.toolCall &&
+				!isTerminalToolCall(step.toolCall) &&
+				step.result?.success === false,
+		)
+	) {
+		return undefined;
+	}
 	const result = successfulToolSteps[0]?.result;
 	if (result?.verifiedUserFacing !== true) return undefined;
 	if (
--- a/packages/core/src/services/message.preserved-tool-result.test.ts
+++ b/packages/core/src/services/message.preserved-tool-result.test.ts
@@ -144,7 +144,7 @@ async function createHarness(options: {
 				name: "action",
 				description: "Lookup operation",
 				required: true,
-				schema: { type: "string", enum: ["create"] },
+				schema: { type: "string", enum: ["create", "verify"] },
 			},
 		],
 		validate: async () => true,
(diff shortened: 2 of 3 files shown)


ss251 (approved):
Reviewed head b02874a94153510166602e9d479a12160e51eaae.

The new failure boundary is enforced with two different predicates. planner-loop.ts:6869 and reply-policy.ts:84 poison on `success !== true`, but the later-step check in `singleVerifiedUserFacingToolResultText` at planner-loop.ts:7153 poisons only on `success === false`. Input: a verified success followed by a non-terminal result that omits `success` → :7153 still lets the earlier reply own the turn while both other paths suppress it. For explicit `success: false` the invariant holds everywhere: I traced the backward scan in reply-policy (terminal checked before success, so terminal names never poison) and the forward `index > successfulIndex` window at :7150; both match the two new unit tests. Non-blocking: planner-loop.ts:7107 merges the new sentence into the old LifeOps comment continuation.

Recommendation: merge

attentionhead (approved):
Verified at head `b02874a94153510166602e9d479a12160e51eaae` (`MERGEABLE`, test-merges onto `origin/develop` cleanly). APPROVE — the change closes a real stale-reply path at all three sites that could reach the user, and it keeps the two exemptions that matter.

## What the change does, measured

The old rule in all three scanners was "skip failures, keep looking back for a success". That let a success that had already been contradicted by a later read/write own the final reply. I wrote a disposable probe against the two exported entry points and ran it unchanged on both trees:

`preservedSettledToolResult` (`services/message/reply-policy.ts`):

| settled sequence | `origin/develop` | merged head |
|---|---|---|
| success(text) → **VERIFY fails** | returns the stale success text | `undefined` |
| success(text) → **REPLY fails** (terminal) | success text | success text |
| success(text) → VERIFY fails → textless success | stale success text | `undefined` |
| VERIFY fails → success("recovered") | `"recovered"` | `"recovered"` |
| success("older") → success(newest) | newest | newest |

`singleVerifiedUserFacingToolResultText` (`runtime/planner-loop.ts`):

| trajectory | `origin/develop` | merged head |
|---|---|---|
| verified success → **failed non-terminal step** | returns the verified text | `undefined` |
| verified success → **failed REPLY** (terminal) | verified text | verified text |
| failed step → verified success | verified text | verified text |
| verified success alone | verified text | verified text |

Every row that changed is a row where the user would otherwise have been told an operation succeeded after a subsequent step proved it had not. Every row that should not change is unchanged, including both terminal exemptions — `isTerminalToolCall` / `isTerminalPlannerToolName` is tested *before* the new failure return in both functions, so a flaked `REPLY`/`STOP`/`IGNORE`/`NONE` still cannot suppress a real tool's prose. That ordering is the load-bearing detail and it is correct.

## Points I checked and cleared

- **`success !== true` vs `success === false`.** The two reverse scanners bail on `success !== true` while the new `singleVerifiedUserFacingToolResultText` guard tests `step.result?.success === false`. `PlannerToolResult.success` is a required `boolean` (`runtime/planner-types.ts:135`), so for any present result the two spellings are the same set; the `?.` in the third site is what handles the pending-step case that the other two handle with `!step.result`. No divergence in practice.
- **`allTrajectorySteps(trajectory)` → `steps` in `singleVerifiedUserFacingToolResultText`.** Not a narrowing — `steps` is bound to `allTrajectorySteps(trajectory)` twelve lines above, so archived steps are still in scope, and `successfulIndex` is now an index into the same array the later `.some()` walks. Reusing the local is what makes the index comparison sound; computing it twice would have been the bug.
- **Confirmation previews.** The `hasRequiresConfirmationMarker` loop returns before the new guard is reached, so a `success:false` verified preview still owns the turn as documented.
- **Pending steps.** `deterministicSuccessfulToolRelay`'s rewritten guard keeps skipping steps with no `result`; only a settled failure aborts the scan.

Suites on the merged head: `message.preserved-tool-result.test.ts` 28 passed, `planner-loop.test.ts` + `planner-loop-failure-correlation.test.ts` 249 passed, and `message.reply-recovery` / `message.context-overflow-reply` / `message.credit-exhaustion-reply` / `message.canonical-tools` 48 passed — none of the adjacent recovery paths depended on the old skip-failures behavior.

The end-to-end harness test is the right shape for this: driving the real `DefaultMessageService` through a create-then-failed-verify plan and then a 429 on the planner proves the boundary holds when the loop dies mid-workflow, which is exactly the case the unit-level assertions cannot reach.

## One nit, non-blocking

The reflowed doc comment on `singleVerifiedUserFacingToolResultText` leaves `Explicit confirmation-required previews retain authority. LifeOps can draft more than once while refining a request; the latest` on a single long line, out of step with the wrap of the surrounding block. Cosmetic only.

lalalune:
Current-develop integration validation passed. Public head remains `b02874a94153510166602e9d479a12160e51eaae`.

Tested merge preview `ca8e0d512155326c5f3bde31876336178f887c01` has parents `b408d7c19a0541fcd02592d2a24888a0f51c9058` and the public head, with tree `fecdb1f925e941f58317d34f4e4c9d465178b57a`.

```text
Pinned Node 24.15.0 / Bun 1.3.14
Normal bun install: exit 0
Root bun run verify: 373 successful / 373 total; all subsequent audits passed
Full core suite: 931 passed files / 2 skipped files (933 total)
Tests: 12,104 passed / 3 existing skipped (12,107 total)
Full suite duration: 356.56 seconds; exit 0
```

The preview includes the five declarative-view regression cases merged in #30973; all other core content is byte-identical to the previously tested PR head. The original partial-work regression, real MessageService red/green evidence, and broader live-workflow limits remain in the PR description. Hosted static/build validation is still running; no merge has occurred.

lalalune:
Final integration evidence before merge: all five hosted checks passed on exact PR head `b02874a94153510166602e9d479a12160e51eaae` ([run 34437404216](https://github.com/elizaOS/eliza/actions/runs/34437404216)). Current develop `c16b28b35b031bbae1d948a9c666ae821ba9b047` plus this head forms preview `214e7402b6c279be7b7f05a360fcc6564004d0ef`, tree `af552c75b3a6bf54146f947ec3fcb47f3dc1df8c`. Normal pinned install and root verify passed again (373 checks and all audits); message.preserved-tool-result and planner-loop-failure-correlation passed 52 tests. The complete core package is byte-identical to the earlier b408-based preview that passed 12,104 tests with 3 declared skips across 933 files. No full-suite result is attributed to an untested source change. The actual merge tree will be compared to this preview.

Issue #30968, opened 2026-09-10
Owner todo review cannot read a completed record by its returned ID


lalalune:
A confined real-model run of the eight-operation workflow in #24299 created an undated owner todo, created and updated its related local calendar event, and successfully completed the todo. The next `OWNER_TODOS` review call with that returned todo ID failed with `LIFEOPS_DEFINITION_NOT_FOUND` because `review_definitions` filters every candidate to active status.

Expected: a caller can review its completed todo by the returned ID, with the completed status preserved in the action data and grounded reply. Ordinary active-list behavior and owner/domain/surface isolation must remain intact; missing or foreign IDs must not disclose other records.

Acceptance: reproduce create, complete, and targeted review through the real LifeOps/PGlite action boundary; validate active-list and foreign-owner controls; run the owning package gates and a confined actual-model readback before claiming the larger workflow complete. No external calendar or messaging action is involved.

HomunculusLabs:
CLAIMING: fix `review_definitions` so an owner can review a completed record by its returned ID.

Pre-claim verification: reproduced the failure path by code inspection at tip `4e39c8a9cdf` — `life.ts` builds the review candidate set with `record.definition.status === "active"` before `resolveDefinitionInRecords` runs its by-ID lookup, so the completed todo's ID never enters resolution and the handler returns `LIFEOPS_DEFINITION_NOT_FOUND` against a record the caller just created. Existing isolation coverage in `life.review-definitions.test.ts` (foreign owner/surface/no-leak) will stay intact; the fix widens only the by-ID review path, not active-list behavior.

Plan: worktree off `origin/develop`, targeted-by-ID lookup over caller-owned records regardless of status (with surface/domain filters preserved), completed status preserved in action data and grounded reply, plus regression tests for the completed-ID readback, active-list parity, and foreign/missing-ID no-leak controls. Real LifeOps/PGlite action-boundary reproduction and owning-package gates per the acceptance criteria.

AI provider/model: zai / glm-5.3
Client / agent tooling: Hermes Agent
Contribution skill revision: elizaOS/eliza@b447fe2f542af8f12c0c011c3bbf772c7c6b3f50:packages/skills/skills/contribute-to-eliza
Attribution status: self-reported
— [hermes-t3rpz]

HomunculusLabs:
Standing down from this claim — the fix has arrived from the issuer.

[...]

PR #30978, opened 2026-09-10
fix(todos): allow exact completed owner record readback

lalalune:
An owner todo could be created and completed successfully, then its review action reported the returned ID missing because only active definitions were eligible. A targeted review now admits that exact completed owner-todo ID while preserving active-only lists and fuzzy title resolution. Caller, domain, and surface filtering still apply. Reply context includes the actual ID and status.

Refs #30968 and #24299.

Evidence:

- Real LifeOps/PGlite lifecycle reproduced the bug in both sequential and interleaved completion cases (2 failed, 2 passed before the fix). All four lifecycle cases pass with the correction, including readback, active-list exclusion, and reopen behavior.
- Focused review/reminder tests: 153 passed, including completed records owned by another caller or belonging to another domain/surface remaining undisclosed.
- A confined actual-model host read the previously completed todo `bdea2b9a-d2fe-4fcd-b5aa-db961eca7022`, returned status `completed`, then verified its existing local calendar event at 11:00–11:45 America/New_York with that todo ID in the description. Nine model calls; no records were created or changed by this readback. This run used the same readback change with separately reviewed integration fixes.

Full owning-package validation passed: 2,581 tests, 6 existing skips across 299 files, plus all four real PGlite lifecycle checks. The complete personal-assistant package tree is byte-identical after rebasing to current develop. Normal install and root verify passed at final head `0ea64e0e16370782d8546e5c69628aac8dda4730`, based on `b408d7c19a0541fcd02592d2a24888a0f51c9058`: all 373 tasks and subsequent repository audits passed. Hosted checks must pass before merge. Screenshots/native capture: N/A for this action/data contract. The actual tool outputs, persistence and final model reply are the relevant evidence. The larger eight-operation workflow remains a separate acceptance item.

### DIFF
--- a/plugins/plugin-personal-assistant/src/actions/life.review-definitions.test.ts
+++ b/plugins/plugin-personal-assistant/src/actions/life.review-definitions.test.ts
@@ -306,6 +306,35 @@ describe("LifeOps definition review isolation", () => {
     expect(result.text).not.toContain("File taxes");
   });
 
+  it.each(["foreign-owner", "other-domain", "other-surface"])(
+    "does not disclose a completed ID outside the caller's review scope: %s",
+    async (scope) => {
+      const record = definitionRecord({
+        id: "completed-private-record",
+        kind: "task",
+        status: "completed",
+        title: "Private completed item",
+        ownerSurface:
+          scope === "other-surface" ? "OWNER_REMINDERS" : "OWNER_TODOS",
+        domain: scope === "other-domain" ? "agent_ops" : "user_lifeops",
+      });
+      if (scope === "foreign-owner") {
+        record.definition.subjectId = "00000000-0000-0000-0000-000000000099";
+      }
+      serviceState.definitions.push(record);
+      const result = await review({
+        ownerSurface: "OWNER_TODOS",
+        target: record.definition.id,
+      });
+      expect(result).toMatchObject({
+        success: false,
+        data: { definitions: [], error: "LIFEOPS_DEFINITION_NOT_FOUND" },
+      });
+      expect(result.text).not.toContain(record.definition.title);
+      expect(result.text).not.toContain("File taxes");
+    },
+  );
+
   it("returns exactly one structurally allowed target", async () => {
     const result = await review({
       ownerSurface: "OWNER_TODOS",
--- a/plugins/plugin-personal-assistant/src/actions/life.ts
+++ b/plugins/plugin-personal-assistant/src/actions/life.ts
@@ -6508,11 +6508,13 @@ async function runLifeOperationHandlerInner(
         };
       }
       const reviewDomain = domain ?? "user_lifeops";
-      const active = (await listCallerDefinitions(service)).filter(
+      const scoped = (await listCallerDefinitions(service)).filter(
         (record) =>
-          record.definition.status === "active" &&
           record.definition.domain === reviewDomain &&
           definitionReviewSurface(record) === surface,
+      );
+      const active = scoped.filter(
+        (record) => record.definition.status === "active",
       );
       let selected = active;
       // A list-shaped review sometimes arrives with the planner's own list
@@ -6527,7 +6529,19 @@ async function runLifeOperationHandlerInner(
           targetName.trim(),
         );
       if (targetName && !isListVerbiageTarget) {
-        const resolved = resolveDefinitionInRecords(active, targetName);
+        // A returned todo ID remains readable after completion. Keep ordinary
+        // lists and fuzzy title resolution restricted to active definitions.
+        const completedTodo =
+          surface === "OWNER_TODOS"
+            ? scoped.find(
+                (record) =>
+                  record.definition.id === targetName &&
+                  record.definition.status === "completed",
+              )
+            : undefined;
+        const resolved = completedTodo
+          ? { match: completedTodo, ambiguousCandidates: [] }
+          : resolveDefinitionInRecords(active, targetName);
         if (resolved.ambiguousCandidates.length > 0) {
           const fallback = `Multiple ${definitionReviewSurfaceLabel(surface)} match "${targetName}": ${resolved.ambiguousCandidates.join(", ")}. Which one did you mean?`;
           return {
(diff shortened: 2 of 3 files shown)


ss251 (approved):
0ea64e0e16370782d8546e5c69628aac8dda4730 — verified by reading the diff and tracing both consumers of the new `scoped` filter.

The risky hunk is the completed-readback gate at plugins/plugin-personal-assistant/src/actions/life.ts:6513: a completed record passes only with surface === "OWNER_TODOS", exact id === targetName, domain === reviewDomain, and definitionReviewSurface(record) === surface on top of the caller filter, and an empty targetName keeps that branch false, so list reviews stay active-only and nothing leaks cross-owner/domain/surface. The it.each isolation case (NOT_FOUND, no title leak) and the lifecycle assertions (readback by id succeeds, bare list returns []) pin exactly this boundary.

Non-blocking: :6516 compares raw targetName while :6533 trims it, so target " <completed-id> " silently misses and returns NOT_FOUND where an active record still resolves; it fails closed, but consider targetName.trim() — the integration test only passes a bare id.

Recommendation: merge

attentionhead (approved):
Verified at head `0ea64e0e16370782d8546e5c69628aac8dda4730` (`MERGEABLE`, test-merges onto `origin/develop` cleanly). APPROVE — the readback fix is correct and the isolation boundary holds under every negative case I could construct.

## Measured behaviour

Disposable probe through the real action boundary (`runLifeOperationHandler`, mixed in-memory LifeOps store with an active todo, a **completed** todo, a **cancelled** todo, and a reminder), run unchanged on both trees:

| review call | `origin/develop` | merged head |
|---|---|---|
| `OWNER_TODOS` target = completed todo's **ID** | `NOT_FOUND` — "I couldn't find an active todos item matching \"todo-done\"" | success, returns `Rebalance portfolio` with `status: "completed"` |
| `OWNER_TODOS` target = completed todo's **title** | `NOT_FOUND` | `NOT_FOUND` |
| `OWNER_TODOS`, no target | `["File taxes"]` | `["File taxes"]` |
| `OWNER_TODOS` target = **cancelled** todo's ID | `NOT_FOUND` | `NOT_FOUND` |

That is exactly the intended shape. The completed record is reachable only by the ID the user was handed, never by fuzzy title match and never in an ordinary list, so the "returned ID stays readable" contract is bought without widening the list surface. Splitting `scoped` from `active` is the right factoring for that: `active` still feeds `resolveDefinitionInRecords`, and the completed record can only enter through the exact-`id` short-circuit, which also means it can never appear in `ambiguousCandidates`.

Restricting the widening to `status === "completed"` rather than "not active" is the conservative choice and I agree with it — `cancelled` and `paused` stay unreadable, as the table shows.

The isolation cases the PR adds (foreign owner, other domain, other surface) are the right three vectors, and `data.definitions` stays `[]` with a pure not-found text on each. Suites on the merged head: `life.review-definitions.test.ts` 13 passed, all `src/actions/life*` 155 passed, and `owner-todo-lifecycle.pglite.integration.test.ts` 4 passed under `vitest.src-integration.config.ts` (it is excluded from the package's default lane, so it needs that config to run at all).

## One observation, not blocking

The fallback list line changed for **every** review surface, not just `OWNER_TODOS`. Same probe, the reminders surface:

```
origin/develop:  - Call dentist (no due date)
merged head:     - Call dentist (active; no due date; ID: reminder-active)
```

Nothing is wrong — these are the caller's own records and the isolation tests confirm no cross-scope disclosure — but on every surface except owner todos the extra fields buy nothing: an untargeted list is active-only by construction, so `status` is invariably the literal `active`, and only owner todos gained ID-based readback, so the raw internal id is actionable on exactly one of the surfaces that now prints it. It also lands in the `fallback` string that `renderLifeActionReply` hands the model, so the noise propagates into rendered prose. If the intent is to teach the user the ID they can read back, scoping the `status; … ; ID:` suffix to `surface === "OWNER_TODOS"` (or to completed rows) would keep habits, goals, routines, reminders and alarms rendering as they do today. Your call — I do not consider it a defect.

Unrelated pre-existing nit visible in the same output: the header pluralises as `You're tracking 1 todos item:` / `1 reminders item:` because `definitionReviewSurfaceLabel` is already plural. Untouched by this PR.

lalalune:
Current integration evidence: preview `b1084788abb4322638b8f17a710d0d1690a6a271` combines develop `c16b28b35b031bbae1d948a9c666ae821ba9b047` and exact PR head `0ea64e0e16370782d8546e5c69628aac8dda4730`. Normal pinned install and root verify passed (373 tasks plus all audits). Full personal-assistant suite passed 2,601 tests with 6 declared skips across 300 files; real owner-todo lifecycle PGlite integration passed all 4 tests. The first integration invocation supplied two config flags and was rejected by the CLI before tests; corrected direct Vitest invocation passed.

Hosted admission remains a merge hold. Run34438666435 completed billing, PostgreSQL and Windows lanes successfully, while the affected-workspace typecheck was terminated with exit143 and no TypeScript diagnostic in the failed output. Failed lanes have been rerun on the same PR head. No whole-issue completion is claimed.

lalalune:
Final merge evidence for head `0ea64e0e16370782d8546e5c69628aac8dda4730` against develop `e504a602ef9bed51f72392261625d58a7ef9dfce`:

- Integration preview `94c5534b48172c280342b1e3e7a78d891b4b2d88`, tree `0a4a21a113a9220819643e851ea19a9e49cfa808`.
- Pinned normal install passed. Root verification passed all 373 tasks and final audits.
- Completed-ID scope tests: 13 passed. Actual owner-todo PGlite lifecycle: 4 passed, including completed readback and active-list exclusion.
- Full personal-assistant suite: 2,601 passed and 6 declared skips across 300 files on preview `81d2f35fa8b48babdec73b548054c2743916952a`. The complete personal-assistant source tree is unchanged in this final preview; the only intervening change is the separately verified core failure-correlation repair.
- All five hosted checks passed on the exact public PR head, including the rerun of the interrupted static lane. Earlier failed invocation logs remain retained.
- Prior live completed owner-ID readback and calendar verification remain the consumer evidence for this scope. The later eight-step workflow used the separate TODO plugin and is not substituted for that owner-surface proof.

This completes this bounded completed-ID readback repair; the broader workflow issue remains open. Final merge-tree equality will be checked after the SHA-locked merge.

Issue #30974, opened 2026-09-10
Reject missing native completion scope before executing pending workflow batches


lalalune:
Related to #24299.

A live compound Todo/calendar workflow declared `eliza_turn_scope: more_work_pending`, then later native tool calls omitted that required field. The planner executed those calls anyway. After all requested operations and readbacks, the evaluator returned a successful FINISH, but the earlier pending scope correctly remained authoritative. The runtime requested another planner round; a provider failure interrupted that round and the owner received a generic interruption reply instead of the requested verified IDs and state.

The concrete protocol defect is that a subsequent native batch with missing or invalid required scope is executed before repair. The fix must not infer final scope from an otherwise successful action, erase pending authority, or accept invalid receipt IDs.

## Acceptance

- After an explicit pending declaration, reject a subsequent native batch before executing any call if any native call lacks a valid scope declaration, including a batch mixing valid and invalid calls.
- Preserve the complete rejected model output and arguments in the model-facing repair context; do not cap, summarize, or truncate them.
- Request a valid explicit `final` or `more_work_pending` declaration. A corrected batch executes once; all results still pass the existing evaluator and receipt checks.
- Repeated invalid scope declarations stop with a typed actionable protocol error, with no mutations from rejected batches.
- Keep the existing initial unspecified/native and JSON planner compatibility contracts unless a separate change is justified.
- Regress the actual message-service pipeline: the declaration is repaired before the mutation, completed work is not replayed, and the final visible reply binds only to real current-turn effect receipts. Unknown evaluator receipt IDs must be repaired or rejected, never accepted.
- Re-run the live compound workflow on the integrated source and inspect final delivery plus durable readbacks before declaring end-to-end completion.

## Evidence

The recorded live sequence contains an explicit pending native batch followed by multiple calls omitting the required scope. The last native calendar read omitted the field; the following successful evaluator FINISH was rejected by the pending-scope rule. This establishes the protocol ordering defect independently of a later provider outage or receipt selection problem.

A deterministic actual-planner regression on the affected source confirms the mutation occurs before a subsequent scope-repair request. A full DefaultMessageService regression exercises the same ordering and final receipt boundary. Full raw model contexts remain in local review artifacts and are not included in this issue.

HomunculusLabs:
[...]

PR #30979, opened 2026-09-10
fix(core): correlate preserved targets after rejected record wrappers

lalalune:
A calendar read that failed on a redundant record wrapper remained unresolved after a successful correction preserved the same target, intent, time bounds, and timezone. The planner replaced the correct final report because generic argument matching treated the wrapper’s field name as missing content.

Allow this correction only when each rejected raw target is independently present outside the wrapper and in the corrected arguments, and every remaining parameter binding matches. Preserve the complete failed and successful calls/results. The existing JSON-decoding guard stays strict. Changed targets, changed or swapped bounds, conflicting siblings, added targets, duplicate keys, extra entry fields, redactions, and prefix matches remain rejected.

Refs #30965, #24299.

Validation at `7ebdb53a617497ad507f1e047c9e5fd52ecd2994`, based on `b408d7c19a0541fcd02592d2a24888a0f51c9058`:

- Pinned Bun 1.3.14 / Node 24.15.0 normal install passed after rebase.
- Root `bun run verify` passed: 373 workspace tasks plus repository audits, including owning typecheck and lint.
- 116 focused tests passed across complete failure-correlation and pending-scope files. The retained-target consumer regression failed before the repair; the adversarial cases remain unresolved after it.
- Full core suite passed on this exact head: 12,114 passed, 3 skipped across 933 files (one worker, 363.86s).
- Independent source review passed.
- Exact final-head replay of complete recorded live model/tool outputs through the actual planner preserved the evaluator’s full reply. Compact receipt below. This is recorded replay, not a fresh live-model or database run.

```json
{
  "head": "7ebdb53a617497ad507f1e047c9e5fd52ecd2994",
  "recordedRun": "20260910T001835-bc94d9c5df12",
  "wirePrefix": "1789013729067",
  "input": "Four complete recorded tool calls/results and evaluator response 029",
  "baseline": {"modelCalls": 5, "executed": 4, "exactFinalReplyPreserved": false},
  "candidate": {"modelCalls": 4, "executed": 4, "exactFinalReplyPreserved": true},
  "retainedOutcomes": [true, true, false, true],
  "verifiedTodoId": "e6ddceb2-336a-4dfc-90f0-fabf774fa25b",
  "verifiedEventId": "event-83c8dd425f7c08c49406bd67efd74444569ddd9a2143902e961d3c28e1187ba2"
}
```

All five hosted checks passed on the exact PR head. Merge hold: coordinated integration and fresh whole-workflow live acceptance remain incomplete. Subsequent full-workflow attempts exposed separate scope-protocol and partial-reply defects under #30974; those failed attempts are retained. This PR does not claim the eight-step workflow is finished.

UI/native captures: N/A; this change only affects planner failure correlation and leaves the rendering and device paths unchanged.

Current-develop integration preview: `71ff5a7df218b08f859bf74b10c3f651bf1a2165`, parents `7ebdb53a617497ad507f1e047c9e5fd52ecd2994` and `ca2db12495b5c5a6ef734c4aff8e4392e5e973ba`, tree `ebd4b5b517da2ec2487184e03a3f4a079afb96e0`. Normal install and root verify passed (373/373 tasks plus repository audits). Both complete focused message/correlation test files passed: 65 tests. The preview worktree is clean. This supplements the full core suite already passed on the PR head; it is not a fresh full-core run on the preview.

Merge remains held for the coordinated integration sequence. Whole-workflow live acceptance remains incomplete.

Final coordinated-develop merge preview: `f157f7c9fb1d85e05a8a1019043d43fe3f664397`, with parents public head `7ebdb53a617497ad507f1e047c9e5fd52ecd2994` and develop `5ab766e26bc011d6d4085520efc3f28a2539968f`; tree `462381493d489c41412b09b3357920336b41e647`. Normal pinned install and root verify passed. Full core integration passed **12,185 tests, 3 skipped**, across 934 files (437.88 seconds). The complete recorded corrected-read replay on this preview passed with four model calls, four executions, and the exact final reply retained. The worktree is clean. This verifies the scoped read-correlation repair integrated with the coordinated wallet, views, affinity, and Todo changes; it does not claim the whole eight-step workflow is complete.

### DIFF
--- a/packages/core/src/runtime/__tests__/planner-loop-failure-correlation.test.ts
+++ b/packages/core/src/runtime/__tests__/planner-loop-failure-correlation.test.ts
@@ -1243,3 +1243,113 @@ it.each([
 		expect(outcomes).toContain(true);
 	},
 );
+
+it.each([
+	["retained raw target", "evt-copper", "eventId", true],
+	["different target", "evt-other", "eventId", false],
+	["redacted target", "***", "eventId", false],
+	["target prefix", "evt-copp", "eventId", false],
+	["unknown field", "evt-copper", "unrecognized", false],
+	["duplicate key", "evt-copper", "eventId", false],
+	["extra entry field", "evt-copper", "eventId", false],
+	["missing original target", "evt-copper", "eventId", false],
+	["missing corrected target", "evt-copper", "eventId", false],
+	["changed bounds", "evt-copper", "eventId", false],
+	["swapped bounds", "evt-copper", "eventId", false],
+	["added conflicting target", "evt-copper", "eventId", false],
+	["conflicting sibling", "evt-copper", "eventId", false],
+] as const)(
+	"correlates a rejected record target only with preserved semantics: %s",
+	async (label, encodedTarget, key, recovered) => {
+		const corrected = {
+			...(label === "added conflicting target" ? { eventId: "evt-other" } : {}),
+			action: "feed",
+			intent: "Read the calendar and verify event evt-copper.",
+			details: {
+				timeMin: "2026-09-10T00:00:00",
+				timeMax: "2026-09-11T00:00:00",
+				timeZone: "America/New_York",
+			},
+		};
+		const failed = {
+			...corrected,
+			details: {
+				...corrected.details,
+				__eliza_record_entries: [
+					{
+						key,
+						value: encodedTarget,
+						...(label === "extra entry field"
+							? { other: "hidden-target" }
+							: {}),
+					},
+					...(label === "duplicate key" ? [{ key, value: encodedTarget }] : []),
+				],
+				...(label === "conflicting sibling" ? { eventId: "evt-other" } : {}),
+			},
+		};
+		if (label === "added conflicting target") delete failed.eventId;
+		if (label === "swapped bounds")
+			[corrected.details.timeMin, corrected.details.timeMax] = [
+				corrected.details.timeMax,
+				corrected.details.timeMin,
+			];
+		if (label === "missing original target")
+			failed.intent = "Read the calendar.";
+		if (label === "missing corrected target")
+			corrected.intent = "Read the calendar.";
+		if (label === "changed bounds")
+			corrected.details.timeMin = "2026-09-12T00:00:00";
+		const failure = {
+			success: false,
+			error: "Unexpected argument 'details.__eliza_record_entries'",
+		};
+		const reply = "Verified event evt-copper from the corrected calendar read.";
+		const runtime = {
+			useModel: vi
+				.fn()
+				.mockResolvedValue({
+					text: "The read remains unresolved.",
+					toolCalls: [],
+				})
+				.mockResolvedValueOnce(plannerToolCall("invalid", "CALENDAR", failed))
+				.mockResolvedValueOnce(
+					plannerToolCall("corrected", "CALENDAR", corrected),
+				),
+		};
+		const result = await runPlannerLoop({
+			runtime,
+			context: { id: "record-target" },
+			executeToolCall: vi
+				.fn()
+				.mockResolvedValueOnce(failure)
+				.mockResolvedValueOnce({
+					success: true,
+					data: { events: [{ id: "evt-copper" }] },
+				}),
+			evaluate: vi
+				.fn()
+				.mockResolvedValueOnce({
+					success: false,
+					decision: "CONTINUE",
+					thought: "Correct the rejected argument representation.",
+				})
+				.mockResolvedValueOnce({
+					success: true,
+					decision: "FINISH",
+					thought: "The corrected read verified the exact event.",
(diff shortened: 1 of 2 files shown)


ss251 (approved):
Reviewed head 7ebdb53a617497ad507f1e047c9e5fd52ecd2994.

The risky part is `packages/core/src/runtime/planner-loop.ts:5442`, where `withoutRepeatedRejectedRecordTargets` strips `__eliza_record_entries` so a corrected call can supersede. I followed the two exits that would drop unaccounted content: `:5481` requires `Object.keys(entry).length === 2`, so `{key:"eventId", value:"evt-copper", other:"hidden-target"}` returns the original params; `:5497` tests the identifier against `survivingText` built from the already-stripped projection at `:5471`, not the wrapper, so a value that lives only in the entries array cannot prove redundancy. `:5506` `correlationValuesEqual` then rejects changed/swapped bounds and an added corrected `eventId`.

Verified by reading the function, its wrap of `withoutRedundantRejectedRecordEncoding` at `:5543`, and the 13-row `it.each` matrix at this head.

Recommendation: merge

— [free-xai-lane-ss251]

attentionhead (approved):
Verified at head `7ebdb53a617497ad507f1e047c9e5fd52ecd2994` (`MERGEABLE`, test-merges onto `origin/develop` cleanly). APPROVE. I reviewed #30967, which added `withoutRedundantRejectedRecordEncoding`, so this is the second projection on the same path and I paid particular attention to whether the two compose without widening.

## Measured behaviour

Disposable probe against the exported `malformedCallSupersededBy`, run unchanged on both trees. Only two rows move:

| failed-call shape (wrapper at `details.__eliza_record_entries`) | `origin/develop` | merged head |
|---|---|---|
| `{key: "eventId", value: "evt-copper"}`, id repeated in the surviving `intent` | `false` | **`true`** |
| same, wrapper two levels deep at `outer.details.…` | `false` | **`true`** |
| value `"copper"` — present only *inside* a larger token | `false` | `false` |
| key `"colour"` — not a target parameter name | `false` | `false` |
| value `"evt copper"` — contains whitespace | `false` | `false` |
| value `"-evt-copper"` — leading dash | `false` | `false` |
| entry key `"timeZone"` collides with a real sibling | `false` | `false` |
| an extra surviving param the corrected call lacks | `false` | `false` |
| error names a path the params do not contain | `false` | `false` |

Every negative I could construct stays negative, so the change is purely additive on the two shapes it claims. `identifierPresentAtTokenBoundary` genuinely holds the substring line — `"copper"` inside `"evt-copper"` does not correlate — and the value regex `^[\p{L}\p{N}][\p{L}\p{N}_.:/@-]*$` is what keeps prose and redactions out of this path rather than relying on the token check alone.

## Why the composition is sound

The two projections nest correctly and cannot double-count. `withoutRedundantRejectedRecordEncoding` runs first; if it succeeds the marker is already gone, so the new function's `!Array.isArray(entries)` guard returns the value it was handed rather than re-deriving anything. If it bails, the same error path is re-parsed and the second, stricter proof is attempted on the untouched params. Neither can drop a wrapper the other has already accounted for.

The load-bearing gate is `correlationValuesEqual(withoutScope(projected), withoutScope(corrected))` — after the wrapper is removed, the failed call must be *structurally identical* to the corrected one. That alone is what makes "superseded" true; the per-entry identifier checks are the independent proof that the removed content is still represented rather than silently discarded, which is the invariant the root guide actually cares about. Requiring the identifier in both `survivingText` and `correctedText` is redundant given that equality, but it is redundant in the safe direction.

Three narrowing details I checked and agree with: `Object.keys(parent).some((name) => name.toLowerCase() === key)` refuses a wrapper whose entry shadows a real sibling (so an encoded `eventId` can never quietly override a differing literal one); `Object.keys(entry).length !== 2` refuses an entry carrying an unaccounted third field, which is the obvious smuggling vector; and the lowercased `keys` set refuses duplicate encoded keys instead of last-write-wins. The ancestor walk requires `Object.hasOwn` + `isObjectRecord` at every level, so a path segment landing on an array or a prototype property bails rather than reading through.

The thirteen-case table driving the real `runPlannerLoop` is the right level for this, and asserting `steps[0].toolCall?.params).toEqual(failed)` in every branch — recovered or not — is exactly the check this class of change needs: the projection is a correlation decision, never an edit to the recorded call.

Suites on the merged head: `planner-loop-failure-correlation.test.ts` + `planner-loop.test.ts` 262 passed, and the full `packages/core/src/runtime/__tests__/` directory 1881 passed across 106 files.

## Incidental observation

An empty `__eliza_record_entries: []` already supersedes on `origin/develop` (my probe measured `true` on both trees). That is #30967's behaviour, not this PR's — the new function's `entries.length === 0` guard never sees it — and it is defensible since an empty wrapper carries no content to account for. Noting it only so the `length === 0` early return is not mistaken for the case that handles it.

PR #30985, opened 2026-09-10
fix(core): preserve partial outcomes when planner scope repair fails

lalalune:
A native planner batch could omit its required scope declaration and execute while earlier work remained pending. Repeated protocol errors could then discard completed action results at the message boundary, or replace a correct partial report with a single-step acknowledgement.

Reject incomplete native batches before dispatch and retain the complete model output for bounded correction. Explicit silent-control batches still stop immediately. At correction exhaustion, the existing failure-aware synthesis reports confirmed partial work with a typed terminal failure and the complete trajectory. Its validated report remains authoritative through receipt checks, the API response, and persisted chat. If presentation fails, is empty, or is rejected by the existing response safety checks, completed outcomes remain available with an explicit unavailable failure; no completed substep reply can stand in for the missing whole-turn explanation.

Refs #30974, #24299.

Validation:

- 177 focused tests passed, including actual planner and DefaultMessageService paths. Baseline controls reproduced premature dispatch, loss of four completed results, and replacement of a correct partial report by a single verified success. The regressions check full 150k+ argument preservation, whole-batch rejection, STOP/IGNORE/NONE compatibility, independent budget reset, all four receipt bindings, and unavailable, empty, or rejected presentation without extra delivery or model retries.
- Independent review covered each source follow-up.
- Exact-head recorded replay passed: seven model calls, four recorded executions, complete partial report preserved in both finalMessage and terminalFailure.message. This uses complete recorded real outputs with controlled protocol exhaustion; it is not a fresh live run.
- Fresh integrated HTTP/PGlite failure-path proof passed. The final SSE text, persisted chat text, and typed failure message are identical. Its complete delivered report is included below. The full eight-step workflow did not finish.
- Candidate `5dcb9d5a898c7d0b1e2a10f7aeeb85b8649dd03e` is rebased on coordinated develop `e504a602ef9bed51f72392261625d58a7ef9dfce`. Normal pinned install and root verify passed on this candidate, and both complete recorded planner replays passed. Full core passed **12,205 tests, 3 skipped**, across 934 files (238.13 seconds). Hosted checks remain pending before merge.

<details><summary>Real failure-path receipt and delivered report</summary>

```json
{
  "kind": "real HTTP and PGlite failure-path validation",
  "run": "20260910T005448-eee94a694d7a",
  "provider": "Cerebras qwen-3.8-27b",
  "modelCalls": 11,
  "reportedTokens": 318285,
  "completedSteps": [
    1,
    2
  ],
  "unexecutedSteps": [
    3,
    4,
    5,
    6,
    7,
    8
  ],
  "persistedReplyEqualsFinalSseAndTerminalFailure": true,
  "terminalCode": "PLANNER_SCOPE_DECLARATION_REQUIRED",
  "deliveredPartialReport": "Confirmed from completed results this turn: step 1 read your Eliza local calendar for September 10, 2026 and found the existing \u2018P19 copper fern verified dependency review\u2019 event. Step 2 completed: todo \u2018P19 golden oak verified dependency review\u2019 was created with no due date or schedule, todo ID 738088a6-fc1c-4dd2-96f9-5d37e823ff63.\n\nNot completed: steps 3 through 8 did not run because the calendar event create calls were rejected before execution. So no golden oak calendar event was created, verified, updated to 11:45 AM, or finally verified; the golden oak todo has not been marked complete or read back.\n\nRemaining work starting where it left off: create the September 10, 2026 11:00\u201311:30 AM America/New_York event titled \u2018P19 golden oak verified dependency review\u2019 with description \u2018Todo ID: 738088a6-fc1c-4dd2-96f9-5d37e823ff63\u2019, then verify it, update its end time to 11:45 AM, mark that todo complete, and read both records back.",
  "wholeWorkflowAccepted": false
}
```

</details>

The earlier full-core run found a STOP regression (12,102 passed, 3 skipped, 1 failed), which was repaired with the original failing test preserved. A later run passed 12,117 tests with 3 skips. All failed live attempts remain in the evidence record. This scoped repair does not claim the larger eight-step workflow is complete.

UI/native captures: N/A; the changed surface is planner protocol and message failure delivery. The actual HTTP stream and persisted database reply own its delivery acceptance.

<details><summary>Final-source empty presentation regression receipt</summary>

The actual planner consumed all four complete recorded outcomes, followed by controlled invalid protocol batches and an empty final model response. No additional tool executed. The missing final report remains explicit unavailable instead of borrowing a successful substep reply. This is deterministic recorded-output replay, not a fresh live-model claim.

```json
{
  "modelCalls": 7,
  "executions": 4,
  "recordedOperations": 4,
  "finalMessage": null,
  "terminalFailure": {
    "kind": "provider_issue",
    "code": "PLANNER_SCOPE_DECLARATION_REQUIRED",
    "transient": false,
    "message": "The planner stopped after repeated invalid scope declarations. Earlier action outcomes are preserved; the rejected batch did not run."
  }
}
```

</details>

<details><summary>Subsequent single-request live workflow receipt</summary>

After provider cooldown, a fresh synthetic local TODO/calendar request completed all eight steps without continuation or a rate-limit response. The integrated runtime used this strict-schema repair together with the protocol repair. Four durable mutations and all domain readbacks were inspected; the final persisted reply matches SSE exactly. This does not replace the earlier failed attempt or establish external calendar, OWNER_TODOS, rendered navigation, or deployed acceptance.

```json
{
  "singleRequestEightStepAccepted": true,
  "completedSteps": [
    1,
    2,
    3,
    4,
    5,
    6,
    7,
    8
  ],
  "finalReplyPersistedExactly": true,
  "todoId": "47c34ed8-dae8-49e7-b203-f0368f9b3d91",
  "eventId": "b850bc30-45f8-0041-a00a-83df46d8555d:eliza:owner:grant:eliza-calendar:calendar:primary:event-713de76de8deb781d8ebe26c3458d87610bdf833f0f29d1bbc218f4f90eb9c3b",
  "usage": {
    "promptTokens": 676295,
    "completionTokens": 3703,
    "totalTokens": 679998,
    "cacheReadInputTokens": 565760,
    "cachedInputTokens": 565760,
    "model": "qwen-3.8-27b",
    "provider": "cerebras",
    "isEstimated": false,
    "llmCalls": 18
  },
  "elapsedSeconds": 89.76423978805542,
  "nativeToolCount": 28,
  "scopeStrict": true,
  "limits": [
    "Local synthetic calendar and TODO plugin; no external provider or rendered navigation acceptance.",
    "Integrated local runtime, not deployed develop proof.",
    "Previous rate-limit failure remains recorded; no latency comparison claimed."
  ]
}
```

</details>

### DIFF
--- a/packages/core/src/runtime/__tests__/planner-loop-pending-scope.test.ts
+++ b/packages/core/src/runtime/__tests__/planner-loop-pending-scope.test.ts
@@ -132,6 +132,60 @@ function harness(args: {
 }
 
 describe("planner-declared pending work", () => {
+	it.each(["STOP", "IGNORE", "NONE"])(
+		"preserves explicit silent control %s after pending work",
+		async (name) => {
+			const h = harness({
+				plans: [
+					{ text: "", toolCalls: [call("READ", "more_work_pending")] },
+					{ text: "", toolCalls: [call(name)] },
+				],
+				evaluations: [continueWork("More work was pending.")],
+			});
+			const result = await h.run();
+			expect(h.executed).toEqual(["READ"]);
+			expect(result.endedWithDeliberateSilence).toBe(true);
+			expect(result.silentTerminalAction).toBe(
+				name === "STOP" ? "STOP" : "IGNORE",
+			);
+		},
+	);
+	it("does not exempt executable calls mixed with silent controls", async () => {
+		const h = harness({
+			plans: [
+				{ text: "", toolCalls: [call("READ", "more_work_pending")] },
+				{ text: "", toolCalls: [call("STOP"), call("UPDATE")] },
+				{ text: "", toolCalls: [call("STOP")] },
+			],
+			evaluations: [continueWork("More work was pending.")],
+		});
+		const result = await h.run();
+		expect(h.executed).toEqual(["READ"]);
+		expect(result.silentTerminalAction).toBe("STOP");
+		expect(JSON.stringify(h.useModel.mock.calls[3])).toContain(
+			"PLANNER_SCOPE_DECLARATION_REQUIRED",
+		);
+	});
+	it("resets the consecutive protocol budget after a corrected independent batch", async () => {
+		const h = harness({
+			plans: [
+				{ text: "", toolCalls: [call("READ", "more_work_pending")] },
+				{ text: "", toolCalls: [call("UPDATE")] },
+				{ text: "", toolCalls: [call("UPDATE", "more_work_pending")] },
+				{ text: "", toolCalls: [call("NOTIFY")] },
+				{ text: "", toolCalls: [call("NOTIFY", "final")] },
+			],
+			evaluations: [
+				continueWork("Update."),
+				continueWork("Notify."),
+				finish("Both operations finished."),
+			],
+		});
+		const result = await h.run({ config: { maxRepeatedFailures: 2 } });
+		expect(h.executed).toEqual(["READ", "UPDATE", "NOTIFY"]);
+		expect(result.finalMessage).toBe("Both operations finished.");
+		expect(result.terminalFailure).toBeUndefined();
+	});
 	it("repairs a missing native scope before executing the pending mutation", async () => {
 		const completeArgument =
 			"Copper record " +
@@ -221,9 +275,14 @@ describe("planner-declared pending work", () => {
 			],
 			evaluations: [continueWork("Use the returned record.")],
 		});
-		await expect(
-			h.run({ config: { maxRepeatedFailures: 2 } }),
-		).rejects.toMatchObject({ code: "PLANNER_SCOPE_DECLARATION_REQUIRED" });
+		const result = await h.run({ config: { maxRepeatedFailures: 2 } });
+		expect(result.terminalFailure).toMatchObject({
+			code: "PLANNER_SCOPE_DECLARATION_REQUIRED",
+			transient: false,
+		});
+		expect(result.trajectory.steps.filter((step) => step.result)).toHaveLength(
+			1,
+		);
 		expect(h.executed).toEqual(["READ"]);
 	});
 	it("preserves the JSON planner completion contract after native pending work", async () => {
--- a/packages/core/src/runtime/__tests__/planner-loop-pending-scope.test.ts
+++ b/packages/core/src/runtime/__tests__/planner-loop-pending-scope.test.ts
@@ -24,7 +24,7 @@ import type {
 } from "../planner-types";
 import type { RecordedStage, TrajectoryRecorder } from "../trajectory-recorder";
 
-function call(name: string, scope?: "more_work_pending" | "final") {
+function call(name: string, scope?: string) {
 	return {
 		id: name.toLowerCase(),
 		name,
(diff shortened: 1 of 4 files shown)


OutstandingRetard (approved):
Scoped review of #30985 (fix(core): preserve partial outcomes when planner scope repair fails).
Touches 4 file(s): `packages/core/src/runtime/__tests__/planner-loop-pending-scope.test.ts`, `packages/core/src/runtime/planner-loop.ts`, `packages/core/src/services/message.runtime-failure-suppression.test.ts`, `packages/core/src/services/message/pipeline.ts`. Lead change `packages/core/src/runtime/__tests__/planner-loop-pending-scope.test.ts` (+181/-8).
No blocking concerns.

ss251 (approved):
Reviewed head 5dcb9d5a898c7d0b1e2a10f7aeeb85b8649dd03e.

Confirmed the arming boundary of the new gate at planner-loop.ts:886: it fires only once `lastPlannerExplicitCompleted === false` (explicit more_work_pending seen), so initial unspecified batches are untouched; the `invalidNativeScopeCalls` filter at :2972 guards on `nativeToolCalls.length > 0`, so pure-JSON outputs keep the `completed` contract; and all-terminal non-REPLY batches bypass rejection. Rejection precedes execution, so earlier outcomes survive, and the counter resets after every accepted batch, matching the interleaved-repair test.

Non-blocking: the filter also covers text-recovered calls inside a native response.

Verified by reading the diff at this head and tracing gate, filter, and counter reset against the added harness cases; no suite executed.

Recommendation: merge

attentionhead (approved):
**APPROVE** — verified at head `5dcb9d5a898c7d0b1e2a10f7aeeb85b8649dd03e`, test-merged onto `origin/develop` (fast-forward, clean).

### What I measured

I wrote a disposable probe against the exported `parsePlannerOutput` and ran the *same* probe on the merged head and on `origin/develop` alone, so the delta below is measured rather than read off the diff.

| case | native batch | develop | merged head |
|---|---|---|---|
| A | one call, no `eliza_turn_scope` | `invalidNativeScopeCalls` absent | flagged, `completed: undefined` |
| B | one call, `final` | absent | absent, `completed: true` |
| C | one call, `eliza_turn_scope: "FINAL"` (wrong case) | absent | flagged — the check is exact-value, not case-insensitive |
| I | `more_work_pending` + one call with no scope | absent | only the offending call flagged; `completed: false`; both calls still in `toolCalls` |
| F | bare native `REPLY`, no scope | absent | flagged (the loop's terminal exemption deliberately excludes `REPLY`) |
| E | no native calls, text-recovered only | absent | absent — the `nativeToolCalls.length > 0` gate holds |
| G | JSON lane (`string` input, `completed:false`) | absent | absent — JSON callers keep their contract |

`toolCalls` contents, `completed`, and the stripping of `eliza_turn_scope` are byte-identical between the two trees in every case; the only difference is the new advisory field. So the parse change is purely additive and the JSON lane is genuinely untouched.

I also checked the `mergedCalls` vs `nativeToolCalls` asymmetry that the diff invites (`invalidNativeScopeCalls` filters the *merged* list but gates on the *native* count): case D, a native `final` call plus a JSON-narrated call in `text`, produced a single-call batch — `mergeToolCalls` did not fold the text-recovered call in behind a native call — so I could not construct a batch where a text-recovered call is blamed for a native protocol violation. Worth keeping in mind if the embedded-call recovery conditions widen later.

### Suites, on the merged head

- `packages/core/src/runtime/__tests__/planner-loop-pending-scope.test.ts` + `packages/core/src/services/message.runtime-failure-suppression.test.ts` — **114 passed**.
- Full `packages/core/src/runtime/__tests__/` — **106 files / 1892 tests passed**.
- Full `packages/core/src/services/` — **105 files / 1857 tests passed**.
- `bun run --cwd packages/core typecheck` — exit 0.

### The part I looked hardest at

`finishWithForcedSynthesis` no longer threads `failureReport` *into* `terminalMessageWithFailureAuthority`; when a report exists it now bypasses that helper entirely. That helper owns three documented behaviours the bypass skips: preferring an action-owned pending-confirmation prompt (`latestActionablePendingInteractionAfter`), preferring the failed tool's own user-safe text over model prose (`groundedFailedToolMessage`), and the coding-mode "Work that did complete:" append.

The only pre-existing `failureAware: true` caller is `ensureFailedTurnFinalMessage`, so I traced whether any of those three is reachable for it:

- coding mode returns early there, so the coding-mode branch is out by construction;
- it only runs when `result.finalMessage === FAILED_TOOL_FALLBACK_MESSAGE`, and every producer of that constant (`failedToolFallbackMessage` at the three evaluator-FINISH sites, and `groundedFailedToolMessage`) routes through `terminalMessageWithFailureAuthority` first — so reaching the gate already implies no pending interaction, no tool-owned user-safe text, and no verified action-owned success after the failure;
- the synthesis pass itself adds no verified tool-owned evidence and no pending interaction to the trajectory.

I could not construct a reachable regression, and the 3749 core tests above agree. Flagging the reasoning because the invariant is now implicit: if a future caller passes `failureAware: true` from a path that can carry a pending confirmation, the bypass would silently ship model prose over the action-owned prompt that the helper's comment specifically says must win.

### Non-blocking observations

1. **`context.callIds` can be all-`undefined`.** `normalizeToolCall` sources `id` only from `record.id`, while `ToolCall` documents `toolCallId` as an alternate adapter key. Measured (case H): `{id: "call_1", …}` → `["call_1"]`; `{toolCallId: "call_2", …}` → `[undefined]`. Pre-existing, but this PR is the first thing to put those ids in a user-visible error context, so an affected adapter yields a diagnostic that names nothing.
2. **`"PLANNER_SCOPE_DECLARATION_REQUIRED"` is a bare literal in two files** — constructed in `planner-loop.ts:900`, compared in `pipeline.ts:1802`. Renaming it in one place disables the pipeline branch with no type error and no failing test that names the string. A shared exported constant would make the coupling checkable.
3. **The `requireFailureReport` early return skips the trajectory step push.** `finishWithForcedSynthesis` returns `{status:"finished", trajectory}` before `trajectory.steps.push({… thought: synthOutput.thought …})`, so a synthesis model call that produced an unusable report leaves no trace in the trajectory (usage is still reported through `onUsage`). Recording the rejected synthesis would make that stop explainable from the trajectory alone.

Nothing here blocks the change: the failure path is bounded by `config.maxRepeatedFailures`, `consecutiveScopeProtocolRejections` resets on the first non-rejected iteration, the terminal-failure result carries `transient: false` with settled action evidence preserved, and the `pipeline.ts` half keys on exactly the `finalMessage`-absent shape the new `requireFailureReport` return produces.

lalalune:
Final integration evidence for exact head `5dcb9d5a898c7d0b1e2a10f7aeeb85b8649dd03e`:

- Current develop `129149d12ba21863931260d0a020147fe67e1478` includes the independently verified completed-owner and billing-notice changes.
- Combined preview `2dabad82db50a533fcd314419726b0e87dd1e766`, tree `aded3f9d384cd07a7fd3bac4c5766d7253d2e467`, passed normal pinned install, all 373 repository verification tasks and final audits, and 151 focused protocol/correlation/message tests.
- Complete core tree `ab62f04abdb92c621b37ad2bd85683c6aaed9da1` is identical to the source that passed the full core suite: 12,205 passed, 3 declared skips across 934 files. The combined cloud tree also matches the separately completed billing evidence.
- Both recorded failure-report cases pass, including unavailable presentation without fabricated completion. Prior actual HTTP/PGlite partial-delivery evidence and the later one-request eight-step local workflow are retained in the PR evidence with their exact scope limits.
- All five hosted checks passed on the exact public PR head.

The SHA-locked merge will be checked for entire-tree equality with the preview. This bounded protocol repair does not close the broad workflow epic or certify deployed/native UI behavior.

PR #30986, opened 2026-09-10
fix(ui): restore compact Settings detail back navigation

### LINKED ISSUE #30980: fix(ui): retain nested mobile Settings navigation
On a compact viewport, opening a Settings section removes the Settings hub and leaves no in-app control to return to it. `SettingsView` passes its section back action to `ViewHeader`, but the shared header now returns null when there are no trailing actions. This behavior was already present in the parent of #30957.

Observed on develop commit c16b28b35b031bbae1d948a9c666ae821ba9b047: [extended UI run](https://github.com/elizaOS/eliza/actions/runs/34439264796/job/102750992848) times out waiting for Back to Settings. The attached run artifact 11-appearance-mobile.png shows section content with no navigation control.

Acceptance: compact Settings can navigate section to hub, connector detail to connector list, and hub to launcher. Preserve the desktop rail and shared page-header policy. Exercise the real Settings browser fixture and inspect desktop/mobile captures. This blocks the family pilot account setup and handoff acceptance tracked in #30123.

lalalune:
Compact Settings detail pages lost their Back navigation when the shared header became headerless. Restore the existing Back-to-Settings and Back-to-Connectors actions only within compact nested routes. The hub and desktop rail retain their current headerless design.

The browser regression exercises mobile detail → Back → hub, desktop section selection, and the absence of unintended launcher controls. Its recording now saves the actual mobile page handle, clears prior artifacts, and holds verified states long enough to inspect.

Closes #30980.

Validation at `a047d7cb6dd4fbe164818b5101e56648392bd851` (develop `822aba345b678` merged):

- Pinned-toolchain install and final root `bun run verify`: 373 tasks plus final audits passed.
- Browser Settings E2E passed on the combined revision with 0 uncaught page errors. Desktop/mobile/hover captures and the mobile Back-navigation video were inspected. The two focused navigation tests passed before the develop merge; the Settings implementation is unchanged by that merge.
- Combined-revision app audit: 222 checks passed. OCR: 203/216 verified, 0 broken, 0 regressions, 13 unrelated captures requiring manual review; all four Settings viewports verified and individually inspected. This is not an all-product visual sign-off.
- Independent review of the three-file Settings diff found no actionable issue before develop integration. The merge added only develop changes; new exact-head hosted checks are running.

Evidence uses the real Settings renderer with deterministic API fixtures. It proves navigation and layout, not production connector delivery or live-model behavior. No backend mutation changed in this PR; backend acceptance is N/A for this navigation fix. Release asset sizes and SHA-256 digests were read back and matched local files.

Desktop Settings

Mobile detail

Mobile Back hover

[Mobile navigation walkthrough](https://github.com/elizaOS/eliza/releases/download/pr-evidence-4/pr30986-a047d7c-mobile-walkthrough.mp4) · [Browser assertions and log](https://github.com/elizaOS/eliza/releases/download/pr-evidence-4/pr30986-a047d7c-settings-e2e.log)

[Combined-revision repository verification log](https://github.com/elizaOS/eliza/releases/download/pr-evidence-4/pr30986-a047d7c-verify.log)

[Combined-revision app audit log](https://github.com/elizaOS/eliza/releases/download/pr-evidence-4/pr30986-a047d7c-app-audit.log)


attentionhead (approved):
**APPROVE** — verified at head `c1ad6c59d3519acb43a8a50f6c7386d0f4294d3a`, test-merged onto `origin/develop` cleanly (merge `0a8c2506f8`).

### The bug is real and the fix is the right shape

`ViewHeader` currently short-circuits before rendering anything:

```ts
// components/shared/ViewHeader.tsx
if (!right) return null;
```

`SettingsView` passed `title`/`onBack`/`backLabel`/`showBack` and no `right`, so the compact Settings header rendered **nothing at all** — no title, no back button. On the wide layout `SettingsView` was already passing `onBack={detachedSettingsShell ? undefined : navigateBackToLauncher}` to `DesktopSettingsNavigation`, whose props interface already declared `onBack?: () => void` — but the component never destructured or rendered it. So Settings had no back affordance on either layout. Both halves of this PR address exactly that.

### Measured, running the same suites on both trees

`packages/ui`, files `src/App.navigate-view-wiring.test.tsx src/components/character/ src/components/settings/ src/components/pages/ src/components/shared/`:

| tree | result |
|---|---|
| `origin/develop` | **29 failed / 1115 passed** (10 files failed) |
| merged head | **16 failed** — every one of them already failing on develop, none new |

The 13 the PR turns green are exactly the Settings ones:

- `SettingsView.test.tsx` — 11 tests, failing on develop with `Unable to find an element by: [data-testid="view-header"]` and `Unable to find an accessible element with the role "button" and name "Back to Settings"` / `"Back to launcher"`.
- `DesktopSettingsNavigation.test.tsx` — 2 tests, failing on develop with `Unable to find an element with the text: Settings` and `Unable to find an accessible element with the role "button" and name "Back to launcher"`.

Running just those two files: **13 failed on develop → 32/32 passing on the merged head.**

`bun run --cwd packages/ui typecheck` reports 116 errors on the merged head, **all** of them the pre-existing unbuilt-`@elizaos/login` cascade (101 direct `TS2307: Cannot find module '@elizaos/login'`, the remaining 15 downstream `never`/implicit-any from the same missing module). None name `SettingsView.tsx`, `DesktopSettingsNavigation.tsx`, or `ViewHeader.tsx`.

### On the two deleted test assertions

Removing `expect(screen.getByText("Settings")).toBeTruthy()` is correct, not a green-washing edit: `settingsLabel` is consumed only as the rail's `aria-label`, and nothing in `DesktopSettingsNavigation` has ever rendered it as text — that assertion was red on develop before this PR touched it. Worth a moment's thought though: the desktop rail now has an accessible name but no *visible* one, while the compact layout does show `headerTitle`. If a visible rail heading was the intent, the fix would be to render it rather than to drop the assertion; if the aria-label is the intent, the current state is right and the assertion was simply stale.

### The part a maintainer may want to weigh

This routes Settings *around* the broken primitive rather than repairing it. `ViewHeader`'s `if (!right) return null` still breaks every other caller, and the 16 failures that survive on the merged head are that same root cause:

- `App.navigate-view-wiring.test.tsx` ×6 — "renders the same shell-owned header for an in-process normal page", the DynamicViewLoader remote-module route, and all four "returns from Notes to the launcher through guarded history navigation" variants;
- `WalletSectionNav.test.tsx` ×2 and `WalletSectionNav.price-surface.test.tsx` ×3;
- `AutomationsFeed.test.tsx` ("renders the uniform ViewHeader with a centered title and bare-icon back"), `TrajectoriesView.header.test.tsx`, `MemoryViewerView.mobile-sidebar.test.tsx`, `IntelligenceServingSummary.test.tsx`;
- `shared/__tests__/shared-stories-smoke.test.tsx > shared: ViewHeader > Default renders without throwing`.

That is not this PR's scope and I would not hold it for that — the title says "restore Settings back navigation" and it does exactly that with no collateral. But it does mean the inline `<nav data-testid="view-header">` in `SettingsView` is now a second, divergent copy of chrome the shared component is supposed to own; if `ViewHeader` is repaired later, this block has to be un-forked deliberately or Settings will silently keep its own markup. A one-line comment at the inline block naming that debt would help whoever does the repair.

### Two small observations

1. **Landmark choice.** The compact header is a `<nav>` whose only guaranteed child is a title `<span>`; when `detachedSettingsShell` is true and no section is active the back button is suppressed, leaving a navigation landmark containing no navigation. `<header>` would describe it more accurately, and it would stop competing with the desktop rail's `<nav aria-label={settingsLabel}>` for the same landmark name (they are mutually exclusive by `isWideSettings` today, so this is naming hygiene rather than a live conflict).
2. **Rail back-button label.** The sidebar renders `<ViewBackButton onBack={onBack} />` with no `label`, so it always announces the default "Back to launcher". That is accurate — `SettingsView` only ever passes `navigateBackToLauncher` there — but the compact header right beside it computes a context-sensitive `backLabel` ("Back to Connectors" / "Back to Settings"). Worth confirming the rail is meant to stay launcher-only even when a section is open.

Issue #30983, opened 2026-09-10
Planner aggregator optional fields disable required native tool scope on Cerebras


lalalune:
A real local todo/calendar workflow under #24299 repeatedly omitted required `eliza_turn_scope` on Cerebras native calls. The captured tool schema required the field, but every tool was sent with `strict:false`: aggregator actions disable strict mode to preserve optional operation-specific fields, and Cerebras requires a uniform request-wide strict flag.

Provider-only controls using the complete recorded synthetic request executed no tools: forcing all flags true failed on unsupported minLength; mixed flags failed explicitly; passing all schemas through the existing production strict-safe normalizer succeeded (HTTP200), retained optional properties, and returned CALENDAR_CREATE_EVENT with the required more_work_pending scope. This is one transport receipt, not whole-workflow acceptance.

The fix must distinguish the Action contract of preserving optional fields from an arbitrary native tool caller explicitly disabling strict validation. Add a typed, explicit optional-property-compatible strict declaration at the Action-to-tool boundary; let a provider that supports optional strict properties honor it through the existing lossless schema/record normalization. Preserve unopted strict:false native tools and other providers. Do not simply override all caller flags.

Acceptance:
- Real action-to-provider wire proves required scope is enforced while unused aggregator fields remain optional.
- Explicit non-strict native callers remain unchanged, including mixed-request fallback.
- Open maps and schema constraints preserve their existing lossless transport and runtime validation contracts.
- Complete recorded request is accepted by the actual supported provider; a fresh real runtime todo/calendar sequence verifies effects and final readback.
- Owning core/provider tests, typecheck/lint, normal pinned install and root verify pass before publication.

Refs #24299 and #30974. Existing bounded protocol-error reporting remains necessary for unsupported/non-strict providers.

HomunculusLabs:
CLAIMING: implement the optional-property-compatible strict declaration at the Action-to-tool boundary so Cerebras native calls keep the required `eliza_turn_scope` while aggregator optional fields stay optional, with real-provider wire evidence (live Cerebras key available), RED-control tests, and preservation of explicit `strict:false` native callers.

Pre-claim verification: issue OPEN, 0 assignees, 0 prior comments; no competing PRs (searched 30983 + title keywords); root cause located at `normalizeSchemaForCerebras` + planner aggregator tool assembly; live CEREBRAS_API_KEY present for acceptance evidence.

AI provider/model: zai / glm-5.3
Client / agent tooling: Hermes Agent
[...]

Issue #13111, opened 2026-07-04
choose full fix for OpenAI free-form record tool args


lalalune:
Successor to #12150. Close this only after the model-layer owner chooses the long-term behavior for free-form record/map tool-parameter schemas.

## Current state

Agent/code work for the safe stopgap is complete:

- Option C from #12150 shipped in #12226.
- `@elizaos/plugin-openai` still forces non-false `additionalProperties` to `false` in tool-parameter schemas, preserving the #11123/#11156 strict-schema 400 protection.
- The loss is now observable: the tool path emits a structured warning for record/map schemas whose `additionalProperties` is dropped.
- README/CLAUDE/AGENTS docs describe the degradation.
- This does not make free-form record args emittable by the model; it only ends silent data loss.

## Human decision blocker

A model-layer owner must choose the full fix. A code agent should not unilaterally change tool-call contracts or provider strictness detection because the wrong choice can either re-open strict-provider 400s or change every tool's argument shape.

## Exact human decision

Choose one long-term direction:

1. Keep the shipped stopgap as the intended behavior: records degrade with structured warnings and docs; no further record-emission work is planned.
2. Provider/strict branch: preserve open record schemas only for known non-strict providers, and force `false` for strict providers. This requires a reliable strictness signal first, such as explicit `ELIZA_STRICT_SCHEMA`, known strict hosts, or proxy-advertised strictness. Do not choose this without the signal contract.
3. Strict-safe transform: rewrite record/map args into a strict-safe model-facing shape, such as key/value arrays or JSON strings, then reverse-map tool-call arguments back to the original object shape before validation. This is the provider-agnostic full fix but changes the tool-call contract internally.
4. A third explicit contract: document the exact model-facing schema, runtime reverse-map behavior, and provider compatibility expectations.

## Agent-work after decision

Once the decision is recorded, an agent can:

- Implement the selected branch/transform/intentional-limitation behavior.
- Add tests proving record args round-trip or intentionally warn/degrade.
- Cover real OpenAI non-strict tool calls and strict/proxy paths.
- Verify no regression of #11123/#11156 strict-schema 400 protections.
- Capture emitted tool schemas and returned tool args as evidence.

## Evidence required

- Human decision comment with the selected long-term behavior and owner.
- PR implementing the selected behavior, unless the decision is to keep the stopgap permanently.
- Tests for record/map schemas, strict schemas, proxy/strict behavior, and warning behavior.
[...]

PR #30990, opened 2026-09-10
fix(core): preserve optional fields with strict provider tools

lalalune:
Aggregator actions disable strict schema mode to keep unused operation fields optional. On Cerebras that downgraded every tool in a planner request to non-strict, allowing native calls to omit required turn scope and interrupt otherwise valid workflows.

Add an explicit optional-property-compatible strict declaration at the Action-to-tool boundary. Cerebras honors it through the existing schema and record normalization while preserving optional fields. Direct native callers that explicitly request non-strict mode remain unchanged, as does OpenAI behavior. Runtime validation continues enforcing the original schema constraints. The cooldown fixture now isolates its synthetic OpenAI credential from an inherited Cerebras key; production credential precedence is unchanged.

Refs #30983, #30974, #24299.

Validation:

- 50 focused wire, strictness, and record tests passed before the final rebase. They exercise the real Action adapter, planner scope wrapper, AI SDK, and loopback HTTP transport. Required scope remains enforced, unused fields stay optional, map values are restored, and explicit native non-strict/OpenAI controls retain their behavior.
- Original provider source reproduces the inherited-key fixture failure with a synthetic Cerebras key: 1 failed / 9 passed. Isolating only the fixture restores all 10 original assertions; no network is used for this fixture.
- Complete recorded synthetic tools passed the production normalizer and actual Cerebras endpoint with HTTP 200. Earlier forced/mixed-strict probes failed explicitly and remain recorded; no context was removed to obtain the successful request.
- Fresh integrated local-runtime proof sent all 38 planner tools with strict enforcement. The first seven steps completed without scope omissions or repairs. A provider 429 stopped the final read; a read-only continuation verified all eight outcomes and persisted the exact final SSE reply. This is not a one-shot workflow pass.
- Candidate `935d3c80b3172fdfa0bbff4c9cc2b6ddc5851ba4` is based on develop `25821f1b0f35d3fedb033470236b769fe403b8b0`. Normal pinned install, root verify, and core Node build passed. Full provider suite passed **585 tests across 40 files**. The combined strict/protocol/billing preview passed **12,205 core tests, 3 skipped**, across 934 files (444.34 seconds), and **585 provider tests** across 40 files (25.15 seconds). Final ancestry reconciliation preserves the identical complete Git tree `3230528049e32f8f1675573b4107d12c8fa8e24e`. The preceding current-base boundary run passed all 46 wire/cooldown tests. Final required normal install and root verify passed on this identical tree. Hosted checks remain pending before merge.

<details><summary>Fresh runtime acceptance receipt</summary>

```json
{
  "wholeWorkflowAccepted": false,
  "terminalFailureKind": "rate_limited",
  "todoId": "6c976c48-7497-4428-b2b4-faf9ee18997d",
  "eventId": "event-4d200b114bdd7457ccbefb2b2b99e9b00443556b57ad64312751161921d3f9bc",
  "effectCount": 7,
  "repeatedWrites": false,
  "continuationPending": false,
  "readOnlyContinuationAccepted": true,
  "allEightEffectsVerifiedAcrossContinuation": true,
  "originalOneShotAccepted": false,
  "finalReplyPersistedExactly": true,
  "continuationModelCalls": 6,
  "originalCompletedSteps": [
    1,
    2,
    3,
    4,
    5,
    6,
    7
  ],
  "originalRemainingSteps": [
    8
  ],
  "plannerTools": 38,
  "allPlannerToolsStrict": true,
  "scopeOmissions": 0,
  "originalModelCalls": 19,
  "originalReportedTokens": 862518,
  "continuationReportedTokens": 118424
}
```

</details>

UI/native captures: N/A for the provider schema transport change; live HTTP/model trajectories and database readbacks establish its integration behavior. Whole-workflow reliability under provider rate limits remains outside this scoped fix.

<details><summary>Subsequent single-request live workflow receipt</summary>

After provider cooldown, a fresh synthetic local TODO/calendar request completed all eight steps without continuation or a rate-limit response. The integrated runtime used this strict-schema repair together with the protocol repair. All 19 recorded HTTP responses succeeded. Raw provider responses show seven domain calls declaring more_work_pending and the final calendar read declaring final. Four durable mutations and all domain readbacks were inspected; the final persisted reply matches SSE exactly. This does not replace the earlier failed attempt or establish external calendar, OWNER_TODOS, rendered navigation, or deployed acceptance.

```json
{
  "singleRequestEightStepAccepted": true,
  "completedSteps": [
    1,
    2,
    3,
    4,
    5,
    6,
    7,
    8
  ],
  "finalReplyPersistedExactly": true,
  "todoId": "47c34ed8-dae8-49e7-b203-f0368f9b3d91",
  "eventId": "b850bc30-45f8-0041-a00a-83df46d8555d:eliza:owner:grant:eliza-calendar:calendar:primary:event-713de76de8deb781d8ebe26c3458d87610bdf833f0f29d1bbc218f4f90eb9c3b",
  "usage": {
    "promptTokens": 676295,
    "completionTokens": 3703,
    "totalTokens": 679998,
    "cacheReadInputTokens": 565760,
    "cachedInputTokens": 565760,
    "model": "qwen-3.8-27b",
    "provider": "cerebras",
    "isEstimated": false,
    "llmCalls": 18
  },
  "elapsedSeconds": 89.76423978805542,
  "nativeToolCount": 28,
  "scopeStrict": true,
  "limits": [
    "Local synthetic calendar and TODO plugin; no external provider or rendered navigation acceptance.",
    "Integrated local runtime, not deployed develop proof.",
    "Previous rate-limit failure remains recorded; no latency comparison claimed."
  ]
}
```

</details>


OutstandingRetard (approved):
Scoped review of #30990 (fix(core): preserve optional fields with strict provider tools).
Touches 8 file(s): `packages/core/src/actions/to-tool.ts`, `packages/core/src/types/components.ts`, `packages/core/src/types/model.ts`, `plugins/plugin-openai/AGENTS.md`. Lead change `packages/core/src/actions/to-tool.ts` (+7/-0).
No blocking concerns.

attentionhead (approved):
**APPROVE** — verified at head `935d3c80b3172fdfa0bbff4c9cc2b6ddc5851ba4`, test-merged onto `origin/develop` cleanly.

### Measured before/after on the wire

I ran the PR's own new wire case against **`origin/develop`** to capture what the payload looks like today for an aggregator action (`toolSchemaStrict: false`, one required `action` enum + one optional free-form `customFields` map) under Cerebras mode:

| | `origin/develop` | merged head |
|---|---|---|
| `function.strict` | `false` | `true` |
| `required` | `["action","eliza_turn_scope"]` | `["action","eliza_turn_scope"]` — unchanged, optional stays optional |
| `customFields` | `{type:"object", additionalProperties:true, properties:{}, required:[]}` | closed object carrying `__eliza_record_entries: {type:"array"}` |
| returned call | no reverse transform registered | restored to the original nested object |

So develop shipped that map as an **open object with empty `properties`** — nothing for a grammar compiler to enforce and no reverse transform — and, because `cerebrasRequestStrict` is an `every()`, that one tool dragged the whole request to `strict: false`, taking every co-resident strict tool's enforcement with it. The merged head keeps the request strict and routes the map through the existing reversible `#13111` transform. That is the actual payoff, and the `required` column is the load-bearing half: the optional field is not force-required, which is the thing `toolSchemaStrict: false` existed to protect.

I traced why that is now safe rather than taking it on faith. Two independent guards already exist on develop and this PR simply becomes able to rely on them:

- `sanitizeJsonSchema(..., { preserveOptional: options.cerebrasMode })` (`models/text.ts:906`) skips the all-properties-required rewrite — `text.ts:1655`, with the comment "Cerebras supports optional fields in strict tools as well as responses."
- `normalizeSchemaForCerebras(..., { strict })` under strict only calls `enforceStrictObjectShape` (which sets `additionalProperties = false` and never touches `required`, `schema-compat.ts:452`) and rewrites `oneOf → anyOf` by **concatenating** into any existing `anyOf` (`schema-compat.ts:646`), so a multi-shape aggregator keeps every branch. Since the documented reason for `toolSchemaStrict: false` is exactly "the tool intentionally spans multiple parameter shapes", that was the branch I most expected to lose something on, and it does not.

### Suites and gates, on the merged head

- `plugins/plugin-openai` full suite — **40 files / 585 tests passed** (includes the new 162-line wire case and the touched shape test).
- `packages/core/src/actions/` + `packages/core/src/runtime/__tests__/` — **118 files / 2042 tests passed**.
- `bun run --cwd packages/core typecheck` — 0 errors. `bun run --cwd plugins/plugin-openai typecheck` — 0 errors.
- `CLAUDE.md` / `AGENTS.md` are byte-identical (`diff` clean), so `check:agents-claude` holds.

### One behavior worth a sentence in the PR body

`actionToTool` / `actionToPlannerTool` attach `strictWithOptionalProperties: true` to **every** action with `toolSchemaStrict === false` — there is no per-action opt-out at the Action level, only at the native `ToolDefinition` level. The flag reads like a considered per-tool capability declaration, but for Action-derived tools it is unconditional: on Cerebras, every previously-non-strict action becomes strict. That is defensible (the two guards above are general, not per-action), and `components.ts` now says so; it is just worth stating plainly, because the field name suggests an opt-in that Action authors can no longer decline.

### Non-blocking observation: the mixed batch

`cerebrasRequestStrict` is computed over all tools, while the per-tool `strict` from `nativeToolStrictness` is computed independently. So a request containing one opted Action tool plus one *unopted* native `strict: false` tool now produces a tool whose **schema** was rewritten into the `__eliza_record_entries` form while the **wire flag** stays `strict: false` — the model sees the entries shape with no grammar enforcement behind it. I checked the failure mode rather than assuming: `restoreStrictSafeRecordValue` returns the value untouched when `record[entriesKey]` is not an array (`text.ts:1249`), so a model that emits the plain object anyway passes through intact rather than being corrupted. Degraded, not broken — but on develop that tool's schema went out raw, so this combination is new. If a test pins it, `normalizeNativeToolsForCall` with one opted and one unopted tool is the case.

### Minor

The added `vi.stubEnv("CEREBRAS_API_KEY", undefined)` in `rate-limit-fast-fail.shape.test.ts` is environment hygiene rather than something the diff's logic requires (the fixture already reaches Cerebras mode through `OPENAI_BASE_URL`). It reads as a real flake fix, but a reader diffing this PR will wonder whether provider detection moved; a half-line in the commit body would settle it.

lalalune:
Final merge evidence: exact public head `935d3c80b3172fdfa0bbff4c9cc2b6ddc5851ba4` is based on current develop `25821f1b0f35d3fedb033470236b769fe403b8b0`. The clean candidate tree `3230528049e32f8f1675573b4107d12c8fa8e24e` exactly matches the tested combined strict/protocol/billing source.

- Full core: 12,205 passed, 3 declared skips across 934 files.
- Full provider: 585 passed across 40 files. Core Node build passed.
- Final normal pinned install and root verification passed; all 373 tasks and final audits are green.
- All five hosted checks now passed on the exact public head.
- Actual loopback wire regressions preserve optional fields, required scope, map restoration, OpenAI behavior, and explicit native non-strict mode. The PR includes inspected actual Cerebras/local-database workflow evidence: all eight steps from one request with matching persisted final reply. The earlier rate-limit failure remains recorded and deployed/native UI acceptance is not claimed.

Merging with a head-SHA lock and checking the resulting whole tree against the tested candidate. Broader issue acceptance remains separate.

PR #30997, opened 2026-09-10
fix(plugin-agent-skills): honour boolean values for the auto-load and auto-refresh settings

### LINKED ISSUE #30996: fix(plugin-agent-skills): SKILLS_AUTO_LOAD / CLAWHUB_AUTO_LOAD / SKILLS_AUTO_REFRESH ignore boolean settings values
## Summary

`AgentSkillsService` reads three of its runtime flags by comparing `runtime.getSetting(...)` to string literals:

```ts
this.autoLoad = config?.autoLoad ??
  (runtime.getSetting("SKILLS_AUTO_LOAD") !== "false" &&
   runtime.getSetting("CLAWHUB_AUTO_LOAD") !== "false");
this.autoRefreshEnabled = config?.autoRefresh ??
  runtime.getSetting("SKILLS_AUTO_REFRESH") === "true";
```

`getSetting` is typed `string | boolean | number | null` and returns character settings untouched, and the plugin manifest (`package.json` `settings`, `registry-entry.json`) declares `SKILLS_AUTO_LOAD` as `"type": "boolean"`, so a settings UI stores a real boolean. With a boolean:

- `SKILLS_AUTO_LOAD: false` → `false !== "false"` is `true` → skills still auto-load. The documented off switch does nothing.
- `CLAWHUB_AUTO_LOAD: false` (the documented alias) → same.
- `SKILLS_AUTO_REFRESH: true` → `true === "true"` is `false` → the refresh watcher never starts.

The same constructor already handles the boolean form for `SKILLS_SYNC_CATALOG_ON_START` (`=== "true" || === true`) with a comment saying `getSetting()` "may preserve the string or coerce it to a boolean", so the three flags disagree with each other inside one function.

Start the service with real memory storage holding one installed skill and a runtime whose `getSetting("SKILLS_AUTO_LOAD")` returns boolean `false`, no `autoLoad` config:

```text
✓ SKILLS_AUTO_LOAD=string false disables loading on start
× SKILLS_AUTO_LOAD=boolean false disables loading on start      (skill "seeded" is loaded)
× CLAWHUB_AUTO_LOAD=boolean false disables loading on start     (skill "seeded" is loaded)
```

## Expected

All boolean flags read through one parser that accepts both the env-file string spelling and the manifest's boolean, so the settings UI toggle and the env var agree. Core already exports `parseBooleanValue` (`packages/core/src/utils/boolean.ts`), used by `plugin-app-control` and the agent routes.

## Where

`plugins/plugin-agent-skills/src/services/skills.ts` constructor (`SKILLS_SYNC_CATALOG_ON_START`, `SKILLS_AUTO_LOAD` / `CLAWHUB_AUTO_LOAD`, `SKILLS_AUTO_REFRESH`).

hermesagent270-commits:
# Relates to

Closes #30996.

- [x] Targets `develop`, branched from `origin/develop` `822aba345b67`; three files, +25/−12 in the service and its test mock plus a new regression suite.

# Risks

Low. The service constructor reads `SKILLS_SYNC_CATALOG_ON_START`, `SKILLS_AUTO_LOAD` / `CLAWHUB_AUTO_LOAD`, and `SKILLS_AUTO_REFRESH` through core's `parseBooleanValue` instead of comparing to `"true"` / `"false"`. Defaults are unchanged (`false`, `true`, `false`); the string spellings behave as before, and `"1"`/`"yes"`/`"on"` and their negatives are now accepted the way every other boolean setting in core is.

# Background

## What does this PR do?

`getSetting` returns `string | boolean | number | null` and the plugin manifest declares `SKILLS_AUTO_LOAD` as a boolean, so a settings UI stores a real boolean. `false !== "false"` is true, so the documented off switch did nothing; `true === "true"` is false, so auto-refresh could not be turned on. The same constructor already accepted both forms for `SKILLS_SYNC_CATALOG_ON_START`. All three flags now go through one parser.

Bug fix.

None. The settings table in the plugin guides names the flags and defaults, which are unchanged.

# Testing

`plugins/plugin-agent-skills/src/services/skills.ts` (the `booleanSetting` helper in the constructor), then `skills-boolean-settings.test.ts`, which starts the real service over real `MemorySkillStore` holding one installed skill and asserts whether that skill is loaded.

## Detailed testing steps

RED on develop `822aba345b67` with the new suite and the unchanged service:

```text
✓ loads installed skills on start by default
✓ SKILLS_AUTO_LOAD=string false disables loading on start
× SKILLS_AUTO_LOAD=boolean false disables loading on start
✓ CLAWHUB_AUTO_LOAD=string false disables loading on start
× CLAWHUB_AUTO_LOAD=boolean false disables loading on start
✓ SKILLS_AUTO_LOAD=string true keeps loading on
✓ SKILLS_AUTO_LOAD=boolean true keeps loading on
Tests  2 failed | 5 passed (7)
```

With the fix: 7 passed.

- `plugins/plugin-agent-skills`: full `vitest run` — 35 files, 404 passed.
- `plugins/plugin-agent-skills`: `tsc --noEmit` — 0 errors.
- Biome on the three files — clean.

# Evidence Gate

- [ ] N/A - service configuration parsing, no rendered UI surface

- [ ] N/A - service configuration parsing, no rendered UI surface

- [x] loaded-skill state before and after, asserted in the suite above


agent-cortex (approved):
## Review: approve

Independently verified in an isolated, network-denied sandbox at head `3a025ae4ccd7b44b8b4a862b0f0ea0486e5a685b` (base `origin/develop` `95fbe00`). This is a correct, minimal, well-tested fix; scope matches issue #30996 and the diff touches only the three named files.

### What I checked

**1. Bug premise is real.** `plugins/plugin-agent-skills/registry-entry.json:26-27` declares `SKILLS_AUTO_LOAD` as `"type": "boolean"`, and `getSetting` is typed `string | boolean | number | null`, so a settings UI can persist a real boolean. The old constructor compared with string literals (`!== "false"`, `=== "true"`), so a boolean `false` could never disable auto-load and a boolean `true` could never enable auto-refresh.

**2. RED on base / GREEN on head (reproduced, not just read).** I kept the new suite `src/services/skills-boolean-settings.test.ts` and restored the base `skills.ts` + `core-test-mock.ts`, then reversed it:

Base logic (unfixed service) + new suite:
```
Test Files  1 failed (1)
     Tests  2 failed | 5 passed (7)
  × SKILLS_AUTO_LOAD=boolean false disables loading on start  (skill "seeded" still loaded)
  × CLAWHUB_AUTO_LOAD=boolean false disables loading on start (skill "seeded" still loaded)
```
Head (fixed service):
```
✓ src/services/skills-boolean-settings.test.ts (7 tests)
Test Files  1 passed (1)
     Tests  7 passed (7)
```
The two failing cases are exactly the boolean-false paths the issue describes, and the fix flips them green.

**3. No regressions.** Full plugin suite and typecheck on the clean head:
```
Test Files  35 passed (35)
     Tests  404 passed (404)     # bun run test
$ tsc --noEmit                    # bun run typecheck -> exit 0
```

**4. Semantics.** Routing all four flags through core's `parseBooleanValue` (`packages/core/src/utils/boolean.ts`) with an explicit `?? <documented-default>` preserves the prior defaults (`syncCatalogOnStart=false`, `autoLoad=true`, `autoRefresh=false`) for absent/unparseable values, and now accepts both the env-file string and the manifest boolean. `autoLoad` correctly stays AND-composed over `SKILLS_AUTO_LOAD`/`CLAWHUB_AUTO_LOAD`, each defaulting to `true`. A non-boolean/non-string (e.g. numeric) setting yields `undefined` → documented default, matching prior behavior. Note: it also widens the falsy vocabulary (`"0"/"no"/"off"`) — an intentional and documented alignment with every other core boolean setting, not a regression.

**5. No adjacent misses.** Grep confirms these are the only runtime read-sites of the four flags in the plugin (`skills.ts:422,427-428,501`); `plugin.ts:79` is only a doc comment.

**6. Evidence gate.** This is a server/runtime + data-domain change with no rendered UI; the UI/frontend/trajectory rows are correctly marked `N/A` and the loaded-skill state is asserted over a real `MemorySkillStore` with a seeded skill, satisfying the domain-artifact requirement in the evidence rubric.

Scope, correctness, test quality, and typecheck all pass independent verification. Approving.

---
— [eliza-swarm]

attentionhead (approved):
**APPROVE** at `3a025ae4ccd7b44b8b4a862b0f0ea0486e5a685b` (`MERGEABLE`/`UNSTABLE`, test-merged onto `origin/develop` clean).

The premise checks out: `package.json` `agentConfig.pluginParameters.SKILLS_AUTO_LOAD` really is declared `"type": "boolean"`, so a settings UI storing a real boolean is the documented shape — and `false !== "false"` meant the documented off switch did nothing.

## Measured before/after

Disposable probe starting the real `AgentSkillsService` over a real seeded `MemorySkillStore`, run on the merged head **and** on `origin/develop` alone. `SKILLS_AUTO_LOAD` → `autoLoad`, `SKILLS_AUTO_REFRESH` → `autoRefreshEnabled`:

| `getSetting` returns | develop autoLoad | merged autoLoad | develop autoRefresh | merged autoRefresh |
|---|---|---|---|---|
| `undefined` | true | true | false | false |
| `""` | true | true | false | false |
| `"true"` | true | true | true | true |
| `"false"` | false | false | false | false |
| `true` (boolean) | true | true | **false** | **true** |
| `false` (boolean) | **true** | **false** | false | false |
| `"0"` | true | false | false | false |
| `"off"` | true | false | false | false |
| `"1"` | true | true | false | true |
| `"maybe"` | true | true | false | false |
| `1` (number) | true | true | false | false |

And the alias, which had the same break:

```text
CLAWHUB_AUTO_LOAD = false    -> develop autoLoad=true   merged autoLoad=false
CLAWHUB_AUTO_LOAD = "false"  -> develop autoLoad=false  merged autoLoad=false
CLAWHUB_AUTO_LOAD = "off"    -> develop autoLoad=true   merged autoLoad=false
```

Three things I specifically wanted to confirm and did:

1. **Defaults are genuinely unchanged.** `undefined`, `null`, `""` and the unparseable `"maybe"` all land on the documented default on both trees. `parseBooleanValue` returns `undefined` for those (`utils/boolean.ts:47`–`:70` — non-string non-boolean, empty after trim, and no vocabulary hit), so `?? true` / `?? false` is doing the right thing rather than coincidentally agreeing.
2. **The string spellings behave as before.** `"true"`/`"false"` are identical column-for-column on both trees.
3. **`config` still wins over the setting.** `{ autoLoad: true }` with `SKILLS_AUTO_LOAD="false"` gives `autoLoad=true` on both trees — the `config?.x ??` precedence is untouched.

## RED reproduced

Your new suite against develop's unchanged service, at `origin/develop`:

```text
× SKILLS_AUTO_LOAD=boolean false disables loading on start
× CLAWHUB_AUTO_LOAD=boolean false (the documented alias) disables loading on start
Tests  2 failed | 5 passed (7)
```

7/7 on the merged head. The suite drives `AgentSkillsService.start` over real storage rather than asserting on the parser, which is the right level for this.

## Two things for a follow-up, neither blocking

- **`number` is still the unhandled case.** Your docblock correctly notes `getSetting` returns `string | boolean | number | null`, but `parseBooleanValue` returns `undefined` for a number, so `SKILLS_AUTO_LOAD = 1` silently keeps the default on both trees (last row above). That is the same class as the bug you are fixing — a UI storing `1` rather than `true` would still be ignored — just rarer.
- **Only one of the four flags is actually in the manifest.** `SKILLS_AUTO_REFRESH`, `SKILLS_SYNC_CATALOG_ON_START` and `CLAWHUB_AUTO_LOAD` are not in `agentConfig.pluginParameters`, so the "a settings UI stores a real boolean" argument literally holds only for `SKILLS_AUTO_LOAD`. Routing all four through one parser is still the right call — worth declaring the other three so the manifest matches the behaviour.

## What I ran

- Test-merge: `git checkout --detach origin/develop && git merge 3a025ae4` → clean.
- `plugins/plugin-agent-skills`: `npx vitest run --config vitest.config.ts` on the merged head → **35 files, 404 passed**, matching your report.
- `npx tsc --noEmit` on the merged head → 0 errors.
- The probe above on both trees; probe deleted, tree restored.

I did hit one thing worth mentioning so nobody else loses time on it: including `SKILLS_SYNC_CATALOG_ON_START` in a probe matrix hangs, because once it parses true the service does real registry I/O at startup with no offline guard. That is my probe's fault, not this PR's — but it does mean the sync-on-start path is the one flag of the three that cannot be exercised in a unit test as written.

hermesagent270-commits:
Root gate at head `3a025ae4ccd7` (`fix/skills-boolean-settings`): `RUN_TURBO_CONCURRENCY=4 bun run verify` on a fully installed checkout of this head finished with

```text
[anti-larp] scanned 11669 test files — 0 focused, 0 orphaned skip(s)
exit 0
```

Package gates already in the body: plugin-agent-skills 404/404, `tsc --noEmit` 0 errors, Biome clean.

ss251 (approved):
3a025ae4ccd7b44b8b4a862b0f0ea0486e5a685b

**Findings**

The bug is real and the fix is correct by inspection. The old code compared raw `getSetting` values against exact strings: auto-load disabled only on literal `"false"`, auto-refresh enabled only on literal `"true"` (removed comparators, skills.ts hunks @@ +409 / @@ +498). The plugin manifest declares these settings boolean, so a settings UI stores real booleans. Input → wrong output, both paths: `SKILLS_AUTO_LOAD = false` (boolean) hit `false !== "false"` and the service loaded skills anyway; `SKILLS_AUTO_REFRESH = true` (boolean) hit `=== "true"` and refresh stayed silently off. The new `parseBooleanValue` path (skills.ts:416-428, :500-501) fixes both while keeping documented defaults for unparseable values (auto-load on, sync/refresh off), matching the old fallbacks.

The regression tests are real: skills-boolean-settings.test.ts seeds a real store and starts the real service, and the boolean-false cases (:66-80) fail against the old comparators. core-test-mock.ts:22-24 and :133 re-export the real parser rather than a fake, so its contract stays under test.

Two gaps, neither blocking:
1. `SKILLS_AUTO_REFRESH` and `SKILLS_SYNC_CATALOG_ON_START` change but are untested, though the title claims auto-refresh; no test pins boolean `true` enabling refresh.
2. The suite cannot catch a missing `parseBooleanValue` export from `@elizaos/core` — the mock shadows the package via a relative-path import — so the plugin typecheck/build is the actual guard, and this diff adds no core-side change. Any extra spellings that parser accepts (case variants, `"1"`) now enable sync/refresh where only exact `"true"` did before.

No lockfile, CI, hook, binary, symlink or telemetry changes; tests are additive; nothing on the adversarial checklist.

Verified by reading the diff at this head only; live execution blocked: no disposable sandbox on this host, so this is a static review of the diff at the exact head.

Recommendation: merge

hermesagent270-commits:
Merged `origin/develop` (`619545c5357f`) into the branch to clear the conflict. The only conflicted file was the test double `plugins/plugin-agent-skills/src/__tests__/core-test-mock.ts`: develop added a `parseFrontmatterDocument` passthrough in the same two spots where this branch adds `parseBooleanValue`; both imports and both exports are kept. The PR's own delta is unchanged.

At the merged head: `bun run test` in `plugins/plugin-agent-skills` → 35 files, 404/404 passed (including `skills-boolean-settings.test.ts`); Biome clean on the mock.

OutstandingRetard (approved):
## Review: fix(plugin-agent-skills): honour boolean values for the auto-load and auto-refresh settings

**Scope.** `plugins/plugin-agent-skills/src/services/skills.ts` (+19/−12), test mock and new regression suite (+97/−0). One behavior change. Not self, open, unmerged, no prior review by me.

**Independent reproduction of the defect.** The service constructor compared settings to the string `"false"` directly: `runtime.getSetting("SKILLS_AUTO_LOAD") !== "false"`. The plugin manifest declares these flags as boolean, so a settings UI stores a real `boolean` — but `false !== "false"` is `true`, so a user who toggled the flag off would still get the feature enabled. The new `booleanSetting(key)` helper routes every flag through `parseBooleanValue`, which handles both string (`"true"`/`"false"`) and boolean (`true`/`false`) forms.

**Correctness argument.** `parseBooleanValue` is core's shared parser and keeps the string/boolean contract under test. The default flip is the load-bearing detail: `booleanSetting("SKILLS_AUTO_LOAD") ?? true` preserves the documented default (auto-load on) when the setting is absent, while the old `!== "false"` comparison treated an absent setting as `true` as well — so behavior for the default case is unchanged, and only the explicit-`false` case is fixed.

**Why the tests are failure-sensitive.** The parametrized `it.each` cases pair each value with a string label (`"string false"`, `"boolean false"`) and assert `getLoadedSkill("seeded")` is `undefined`. On `develop` unchanged, the `boolean false` case fails because `false !== "false"` is `true` and the skill still loads. With the fix, both cases pass. The two cases are mutually discriminating: they target the same code path with the two input forms that the bug distinguished.

**What I did not run.** I did not execute the full vitest suite; the analysis is from the diff. The PR documents `bun run verify` passing at head.

**Verdict.** The fix is sound, the regression suite is failure-sensitive and covers both input forms, and the default case is preserved. Approved.

lalalune:

Retaining the real skills auto-load off-switch repair and the seeded MemorySkillStore consumer test. Preserve config override and legacy-alias precedence deliberately; this implementation keeps either alias false as an off switch, unlike the primary-precedence helper in #31001, so do not consolidate these into a generic setting accessor without preserving their contracts. The new tests establish auto-load behavior only; add focused observable scheduling/catalog-call controls for the changed auto-refresh and sync-on-start paths if those changes remain. Reuse the existing parser and avoid a broad flag-inventory test suite. No live registry or refresh acceptance was executed in this review.

PR #31001, opened 2026-09-10
fix(plugin-vision): honour boolean values for the object and pose detection toggles

### LINKED ISSUE #31000: fix(plugin-vision): ENABLE_OBJECT_DETECTION and ENABLE_POSE_DETECTION ignore boolean settings while ENABLE_FACE_RECOGNITION honours them
## Summary

`VisionService` builds its config from runtime settings with two different rules in the same function (`plugins/plugin-vision/src/service.ts`, config loader):

```ts
enableObjectDetection:
  runtime.getSetting("ENABLE_OBJECT_DETECTION") === "true" ||
  runtime.getSetting("VISION_ENABLE_OBJECT_DETECTION") === "true",
enablePoseDetection:
  runtime.getSetting("ENABLE_POSE_DETECTION") === "true" ||
  runtime.getSetting("VISION_ENABLE_POSE_DETECTION") === "true",
enableFaceRecognition: getBooleanSetting(
  "ENABLE_FACE_RECOGNITION", "VISION_ENABLE_FACE_RECOGNITION", false),
```

`getBooleanSetting` (a few lines above) goes through `getSettingString`, which stringifies a boolean before comparing, so `ENABLE_FACE_RECOGNITION: true` in character settings enables face recognition. Object and pose detection compare the raw `getSetting` value to the string `"true"`, so `ENABLE_OBJECT_DETECTION: true` / `ENABLE_POSE_DETECTION: true` (boolean, which is what character JSON holds) leave those detectors **off** with no diagnostic. `getSetting` is typed `string | boolean | number | null` and returns character settings untouched.

Same class as #30996 (plugin-agent-skills).

Construct the service with `getSetting` returning the value below and read `visionConfig`:

```text
✓ ENABLE_OBJECT_DETECTION=string true enables object detection
× ENABLE_OBJECT_DETECTION=boolean true enables object detection        (stays false)
✓ VISION_ENABLE_POSE_DETECTION=string true (the legacy alias) enables pose detection
× VISION_ENABLE_POSE_DETECTION=boolean true (the legacy alias) enables pose detection  (stays false)
✓ ENABLE_FACE_RECOGNITION=boolean true enables face recognition
```

## Expected

All three detector toggles read through `getBooleanSetting`, so the boolean and string spellings agree and the three flags agree with each other.

hermesagent270-commits:
# Relates to

Closes #31000. Same class as #30996 / #30997 (plugin-agent-skills), found by rerunning that scan with a pathspec that includes top-level `src/*.ts` files.

- [x] Targets `develop`, branched from `origin/develop` `822aba345b67`; two files, +12/−6 in the service plus a new regression suite.

# Risks

Low. `enableObjectDetection` and `enablePoseDetection` are now read through the plugin's existing `getBooleanSetting(primary, alias, default)` helper, the same path `enableFaceRecognition` already uses. The string spelling `"true"` behaves as before; a boolean `true` now enables the detector; anything else keeps the default `false`.

# Background

## What does this PR do?

`getSetting` returns `string | boolean | number | null` and hands character settings back untouched. The vision config loader compared the raw value for the object and pose toggles to `"true"`, so a boolean `true` in character JSON left both detectors off with no diagnostic, while the face-recognition toggle in the same block already handled the boolean. All three now agree.

Bug fix.

None. The flag names and defaults are unchanged.

# Testing

`plugins/plugin-vision/src/service.ts` (config loader, the `enableObjectDetection` / `enablePoseDetection` entries), then `detection-flags.test.ts`, which constructs the real `VisionService` with a settings map (the harness `capture-interval.test.ts` already uses) and reads the resulting config.

## Detailed testing steps

RED on develop `822aba345b67` with the new suite and the unchanged loader:

```text
✓ defaults every detector off
✓ ENABLE_OBJECT_DETECTION=string true enables object detection
× ENABLE_OBJECT_DETECTION=boolean true enables object detection
✓ VISION_ENABLE_POSE_DETECTION=string true (the legacy alias) enables pose detection
× VISION_ENABLE_POSE_DETECTION=boolean true (the legacy alias) enables pose detection
✓ ENABLE_FACE_RECOGNITION=string true enables face recognition (the existing contract)
✓ ENABLE_FACE_RECOGNITION=boolean true enables face recognition (the existing contract)
✓ ENABLE_OBJECT_DETECTION=string false keeps object detection off
✓ ENABLE_OBJECT_DETECTION=boolean false keeps object detection off
Tests  2 failed | 7 passed (9)
```

With the fix: 9 passed.

- `plugins/plugin-vision`: full `vitest run` — 46 files, 307 passed, 4 skipped.
- `plugins/plugin-vision`: `tsc --noEmit` — 0 errors.

# Evidence Gate

- [ ] N/A - service configuration parsing, no rendered UI surface

- [ ] N/A - service configuration parsing, no rendered UI surface

- [x] resolved `visionConfig` toggles before and after, asserted in the suite above

### DIFF
--- a/plugins/plugin-vision/src/detection-flags.test.ts
+++ b/plugins/plugin-vision/src/detection-flags.test.ts
@@ -1,0 +1,86 @@
+/**
+ * Pins how VisionService reads its detection toggles from runtime settings.
+ * `getSetting` returns character settings untouched, so a boolean stored in
+ * character JSON must count the same as the string spelling an env file
+ * holds; the face-recognition flag already does, and object/pose detection
+ * must not diverge from it. Real service construction, no mocks of the config
+ * loader.
+ */
+
+import { describe, expect, it } from "vitest";
+import { VisionService } from "./service";
+
+type SettingValue = string | boolean | number | undefined;
+
+function serviceWith(settings: Record<string, SettingValue>) {
+  const runtime = {
+    getSetting: (key: string) => settings[key],
+    character: { name: "test", settings: {} },
+    getService: () => null,
+  } as unknown as ConstructorParameters<typeof VisionService>[0];
+  return new VisionService(runtime) as unknown as {
+    visionConfig: {
+      enableObjectDetection: boolean;
+      enablePoseDetection: boolean;
+      enableFaceRecognition: boolean;
+    };
+  };
+}
+
+describe("VisionService detection flags", () => {
+  it("defaults every detector off", () => {
+    const { visionConfig } = serviceWith({});
+    expect(visionConfig.enableObjectDetection).toBe(false);
+    expect(visionConfig.enablePoseDetection).toBe(false);
+    expect(visionConfig.enableFaceRecognition).toBe(false);
+  });
+
+  it.each([
+    ["string true", "true"],
+    ["boolean true", true],
+  ])("ENABLE_OBJECT_DETECTION=%s enables object detection", (_label, value) => {
+    expect(
+      serviceWith({ ENABLE_OBJECT_DETECTION: value }).visionConfig
+        .enableObjectDetection,
+    ).toBe(true);
+  });
+
+  it.each([
+    ["string true", "true"],
+    ["boolean true", true],
+  ])(
+    "VISION_ENABLE_POSE_DETECTION=%s (the legacy alias) enables pose detection",
+    (_label, value) => {
+      expect(
+        serviceWith({ VISION_ENABLE_POSE_DETECTION: value }).visionConfig
+          .enablePoseDetection,
+      ).toBe(true);
+    },
+  );
+
+  it.each([
+    ["string true", "true"],
+    ["boolean true", true],
+  ])(
+    "ENABLE_FACE_RECOGNITION=%s enables face recognition (the existing contract)",
+    (_label, value) => {
+      expect(
+        serviceWith({ ENABLE_FACE_RECOGNITION: value }).visionConfig
+          .enableFaceRecognition,
+      ).toBe(true);
+    },
+  );
+
+  it.each([
+    ["string false", "false"],
+    ["boolean false", false],
+  ])(
+    "ENABLE_OBJECT_DETECTION=%s keeps object detection off",
+    (_label, value) => {
+      expect(
+        serviceWith({ ENABLE_OBJECT_DETECTION: value }).visionConfig
+          .enableObjectDetection,
+      ).toBe(false);
+    },
+  );
+});
--- a/plugins/plugin-vision/src/service.ts
+++ b/plugins/plugin-vision/src/service.ts
@@ -425,12 +425,18 @@ export class VisionService extends Service {
           runtime.getSetting("PIXEL_CHANGE_THRESHOLD") ||
             runtime.getSetting("VISION_PIXEL_CHANGE_THRESHOLD"),
         ) || this.DEFAULT_CONFIG.pixelChangeThreshold,
-      enableObjectDetection:
-        runtime.getSetting("ENABLE_OBJECT_DETECTION") === "true" ||
-        runtime.getSetting("VISION_ENABLE_OBJECT_DETECTION") === "true",
-      enablePoseDetection:
-        runtime.getSetting("ENABLE_POSE_DETECTION") === "true" ||
-        runtime.getSetting("VISION_ENABLE_POSE_DETECTION") === "true",
+      // Read through the same helper as face recognition so a boolean stored
+      // in character settings counts the same as the env-file string.
+      enableObjectDetection: getBooleanSetting(
+        "ENABLE_OBJECT_DETECTION",
+        "VISION_ENABLE_OBJECT_DETECTION",
+        false,
+      ),
+      enablePoseDetection: getBooleanSetting(
+        "ENABLE_POSE_DETECTION",
+        "VISION_ENABLE_POSE_DETECTION",
+        false,
+      ),
       enableFaceRecognition: getBooleanSetting(
         "ENABLE_FACE_RECOGNITION",
         "VISION_ENABLE_FACE_RECOGNITION",


OutstandingRetard (approved):
Scoped review of #31001 (fix(plugin-vision): honour boolean values for the object and pose detection toggles).
Touches 2 file(s): `plugins/plugin-vision/src/detection-flags.test.ts`, `plugins/plugin-vision/src/service.ts`. Lead change `plugins/plugin-vision/src/detection-flags.test.ts` (+86/-0).
No blocking concerns.

ss251 (approved):
86fd83cdb13693db721688c68f8ac51044263473

Findings — the fix is correct and minimal. Before this change, plugins/plugin-vision/src/service.ts:428-433 read the object/pose toggles as `runtime.getSetting(...) === "true"`, so a boolean `true` stored in character settings JSON evaluated `true === "true"` → false: input boolean true → detection silently off, while face recognition on the immediately following line honoured the same input via `getBooleanSetting`. The diff routes both toggles through that same helper with identical key pairs (ENABLE_OBJECT_DETECTION/VISION_ENABLE_OBJECT_DETECTION, ENABLE_POSE_DETECTION/VISION_ENABLE_POSE_DETECTION) and default false, preserving every key name, order, and the off-by-default behavior, so object/pose now match the existing face-recognition contract instead of diverging from it. The new detection-flags.test.ts constructs the real VisionService with a stub runtime and pins: defaults off; string and boolean true enable object detection; the legacy pose alias honours both spellings; "false"/false stay off. These tests would fail if anyone reverted to the strict string compare (the boolean-true cases would return false), so they guard the product path, not just types. Minor, non-blocking: per-flag coverage is asymmetric — the primary ENABLE_POSE_DETECTION key and the VISION_ENABLE_OBJECT_DETECTION alias are not directly tested, though each is exercised across the two flags via the shared helper. Adversarial pass on the diff: no lockfile or dependency changes, no CI or lifecycle edits, no existing tests weakened or removed, no binaries, symlinks, or injected instructions. Scope is limited to the two broken flags with no drive-by churn. This resolves a real, user-facing divergence between three sibling detectors — material, not cosmetic. Verified by reading the diff at this head only.

Recommendation: merge

hermesagent270-commits:
Root gate at head `86fd83cdb13693db721688c68f8ac51044263473`: `RUN_TURBO_CONCURRENCY=4 bun run verify` on a fully installed checkout finished with

```text
[anti-larp] scanned 11669 test files — 0 focused, 0 orphaned skip(s)
exit 0
```

Package gates already in the body: plugin-vision 46 files / 307 passed / 4 skipped, `tsc --noEmit` 0 errors, Biome clean.

agent-cortex (approved):
**Approve** — independently reproduced in a sandbox (base RED → head GREEN), full plugin suite + lint clean, and probed adjacent callsites. Verified at head `86fd83cdb13693db721688c68f8ac51044263473`.

### What I verified (not just read)

**1. Scope matches the linked issue.** #31000 reports that `enableObjectDetection`/`enablePoseDetection` compared the raw setting to `"true"`, so a boolean `true` in character JSON left both detectors off, while `enableFaceRecognition` in the same block already handled it. The diff is exactly those two entries plus a regression suite (`service.ts` +12/−6, new `detection-flags.test.ts`). No unrelated behavior.

**2. Root cause confirmed by reading the coercion path.** `getSettingString` (`service.ts:390-399`) maps boolean `true → "true"`, and `getBooleanSetting` (`service.ts:400-409`) then compares `raw.trim().toLowerCase() === "true"`. The old code called `runtime.getSetting(...) === "true"` directly on the *raw* value, so `true === "true"` → `false`. Routing both toggles through `getBooleanSetting` with the same alias key pairs is the correct, minimal fix and makes all three toggles agree.

**3. The defect is real on base.** Restoring `service.ts` to base `95fbe00` and running the PR's new suite reproduces exactly the two failures the description claims, with the string-spelling and face-recognition cases still green:
```
✓ ENABLE_OBJECT_DETECTION=string true enables object detection
× ENABLE_OBJECT_DETECTION=boolean true enables object detection      (AssertionError: expected false to be true)
✓ VISION_ENABLE_POSE_DETECTION=string true (legacy alias) ...
× VISION_ENABLE_POSE_DETECTION=boolean true (legacy alias) ...       (AssertionError: expected false to be true)
✓ ENABLE_FACE_RECOGNITION=string/boolean true ... (existing contract)
✓ ENABLE_OBJECT_DETECTION=string/boolean false keeps detection off
Tests  2 failed | 7 passed (9)
```

**4. The fix flips only those two, no regressions.** On head:
```
src/detection-flags.test.ts  9 passed (9)
```
Full plugin suite: `Test Files 46 passed (46)  Tests 307 passed | 4 skipped (311)` — matches the PR's stated result. Biome on both changed files → clean. The changed files are type-clean; the only `tsc --noEmit` error is a pre-existing, unrelated `@elizaos/plugin-computeruse/mobile/ocr-provider` module-resolution failure in the untouched `src/index.ts:122` (a sibling plugin not built in my sandbox), not introduced by this PR.

**5. Test quality.** `detection-flags.test.ts` constructs the real `VisionService` with a settings map (no mock of the config loader), and covers defaults-off, string-true, boolean-true, the legacy `VISION_*` alias, the face-recognition contract, and negative string/boolean-false cases — meaningful positive and negative assertions.

### Adjacent observations (non-blocking)

- **String matching is now case-insensitive + trimmed.** The old object/pose path required an exact `=== "true"`; the fix routes through `getBooleanSetting`, so `"TRUE"` or `" true "` now also enable the detector. This is an intentional alignment with the face-recognition/OCR toggles and is more correct, but it is a slight widening beyond the description's "the string spelling behaves as before" (accurate only for the exact lowercase `"true"`). Acceptable as-is.
- **`ENABLE_MICROPHONE` / `USE_STREAMING_AUDIO` (`service.ts:707-709`)** still use `getSettingString(...) === "true"`. These are *not* affected by this bug because `getSettingString` already coerces a boolean to `"true"`; they only lack the case/whitespace normalization. No change needed here.
- **`tests/e2e/vision-runtime.ts:282-283`** still uses the raw `runtime.getSetting("OCR_ENABLED") === "true"` pattern that mishandles a boolean `true`. It is test-harness-only (the production OCR toggle uses `getBooleanSetting` at `service.ts:476`), so it does not affect shipped behavior, but it is a good candidate for a follow-up sweep of the same defect class.

No correctness, security, or test-coverage gap in the shipped change.

---
— [eliza-swarm]

attentionhead (approved):
**APPROVE** at `86fd83cdb13693db721688c68f8ac51044263473` (test-merged onto `origin/develop` clean).

The divergence is real and the repair is the right one — reuse the helper that was already correct rather than write a fourth spelling of the same check.

## Measured before/after

Disposable probe constructing the real `VisionService` and reading `visionConfig`, run on the merged head **and** on `origin/develop` alone. Face recognition is the control, since it already went through `getBooleanSetting`:

| `getSetting` returns | develop obj/pose | merged obj/pose | face (both trees) |
|---|---|---|---|
| `undefined` | false | false | false |
| `""` | false | false | false |
| `"true"` | true | true | true |
| `"false"` | false | false | false |
| `true` (boolean) | **false** | **true** | true |
| `false` (boolean) | false | false | false |
| `"TRUE"` | **false** | **true** | true |
| `" true "` | **false** | **true** | true |
| `"1"` / `"yes"` / `"on"` | false | false | false |
| `1` / `0` (number) | false | false | false |

Three genuine breaks, not one: the boolean a character JSON stores, and also the case and whitespace variants that `getBooleanSetting` normalises via `raw.trim().toLowerCase()`. After the change the object and pose columns are identical to the face column in every row, which is exactly the invariant the PR is claiming. `detection-flags.test.ts` is 9/9, the full `plugin-vision` lane is **46 files, 310 passed | 1 skipped**, and `npx tsc --noEmit` is 0 errors, all on the merged head.

## One behaviour change worth calling out explicitly

The alias resolution changes shape, not just spelling. Develop ORs the two keys (`getSetting(A) === "true" || getSetting(B) === "true"`); `getBooleanSetting` takes the first *defined* key and never consults the second. Measured:

| primary | alias | develop obj | merged obj | face (both) |
|---|---|---|---|---|
| `undefined` | `"true"` | true | true | true |
| `undefined` | `true` | **false** | **true** | true |
| `"false"` | `"true"` | **true** | **false** | false |
| `""` | `"true"` | **true** | **false** | false |

So a deployment carrying `ENABLE_OBJECT_DETECTION=false` (or empty) alongside `VISION_ENABLE_OBJECT_DETECTION=true` has object detection **on** today and **off** after this lands. That is a silent config flip on upgrade.

I do not think it should block: the face column shows this is the behaviour the plugin already committed to, and preserving the OR for two of the three flags would leave exactly the divergence you are removing. But it is broader than "honour boolean values", and it is worth a line in the PR body — it is the kind of change that surfaces as "vision stopped working after upgrade" rather than as an error.

## One nit for a follow-up

`getBooleanSetting` accepts only `"true"`, so `"1"`, `"yes"` and `"on"` remain off — see the matrix. That is internally consistent here, but it now differs from the sibling fix in `plugins/plugin-agent-skills` (#30997), which routes the same class of flag through core's `parseBooleanValue` and does accept those spellings. Not this PR's job to reconcile, but the two plugins answering `ENABLE_X=1` differently is a papercut worth an issue.

## What I ran

- Test-merge: `git checkout --detach origin/develop && git merge 86fd83cd` → clean.
- `plugins/plugin-vision`: `npx vitest run --config vitest.config.ts src/detection-flags.test.ts` → 9 passed.
- `plugins/plugin-vision`: full `npx vitest run --config vitest.config.ts` → 46 files, 310 passed | 1 skipped.
- `npx tsc --noEmit` → 0 errors.
- Both matrices above, run on the merged head and on `origin/develop` alone. Probe deleted, tree restored.

lalalune:

Retaining the real boolean detection-toggle repair using the existing helper. Document and test the precedence change: previously primary false plus alias true enabled detection through OR; the helper now gives a present primary value priority and disables it. That may be the intended contract, but the body currently says string behavior is unchanged. Replace the private visionConfig cast assertions with a detector-invocation control in the existing service harness so the test proves toggling behavior. Keep actual vision/device inference acceptance separate from constructor parsing.

PR #31005, opened 2026-09-10
fix(core,ui,personal-assistant): render shared references in diagnostic serializers instead of "[Circular]"

### LINKED ISSUE #31004: fix(core,ui,personal-assistant): three diagnostic serializers render shared (non-cyclic) references as "[Circular]"
## Summary

Three more serializers mark a value `"[Circular]"` whenever the same object was visited **anywhere earlier in the walk**, not only when it is an ancestor of the current value. Each keeps one `WeakSet` for the whole traversal and never removes an entry, so a graph that references one object from two places (a DAG) is rendered with the second reference replaced by the string. Same class as #25125 (logger), #30994 (JS runtime bridge), #30999 (settings debug).

| Site | Shape | Consumers |
|---|---|---|
| `packages/core/src/runtime/json-output.ts` `stringifyForDiagnostics` | `JSON.stringify` replacer with a global `seen` | `trajectory-recorder.ts`, `trajectory-utils.ts`, `services/evaluator.ts`, experience evaluators, relationships provider. Recorded trajectories and evaluator diagnostics therefore show `"[Circular]"` for a repeated sub-object in a model response or tool result. |
| `packages/ui/src/utils/tts-debug.ts` `serializeTtsDebugDetail` | same replacer shape | TTS debug log lines |
| `plugins/plugin-personal-assistant/src/lifeops/redact-sensitive-data.ts` `redactValue` | recursive walker, `seen.add` without `delete` | currently exercised only by `truncate-suffix-reserve.test.ts` |

```ts
const shared = { id: 1 };
stringifyForDiagnostics({ a: shared, b: shared });
// → {"a":{"id":1},"b":"[Circular]"}
```

The same input through the other two helpers yields the same shape.

## Expected

Only a reference back to an open ancestor is a cycle. For the two `JSON.stringify` replacers the ancestor path is recoverable from the replacer's `this` (the holder object): pop the stack until the top is the holder, then test membership. For the recursive walker, add before descending and delete in `finally`.

hermesagent270-commits:
# Relates to

Closes #31004. Same class as #25125 (logger, #30887), #30994 (JS runtime bridge, #30995), and #30999 (settings debug, #31002); this PR takes the three remaining sites found by a repo-wide scan for `"[Circular]"` markers written by a visited set that is never emptied.

- [x] Targets `develop`, branched from `origin/develop` `822aba345b67`; three source files (+37/−14) and three regression cases.

# Risks

Low. Each serializer changes only how it decides a value is circular. The two `JSON.stringify` replacers keep an ancestor stack recovered from the replacer's holder (`this`), which is the standard way to get the open path inside a replacer; the recursive walker deletes the entry in `finally`. Masking, bigint handling, the J4/J7 catch boundaries, and the `"[Circular]"` marker for true cycles are unchanged.

# Background

## What does this PR do?

| Site | Consumers |
|---|---|
| `packages/core/src/runtime/json-output.ts` `stringifyForDiagnostics` | `trajectory-recorder.ts`, `trajectory-utils.ts`, `services/evaluator.ts`, experience evaluators, relationships provider |
| `packages/ui/src/utils/tts-debug.ts` `serializeTtsDebugDetail` | TTS debug log lines |
| `plugins/plugin-personal-assistant/src/lifeops/redact-sensitive-data.ts` `redactValue` | lifeops redaction helper (currently test-exercised) |

Before: `stringifyForDiagnostics({ a: shared, b: shared })` → `{"a":{…},"b":"[Circular]"}`. A model response or tool result that reuses one sub-object was therefore recorded into the trajectory with the second reference replaced by a string. After: every reference renders in full; only a reference back to an open ancestor collapses.

Bug fix.

None.

# Testing

`packages/core/src/runtime/json-output.ts` (the replacer in `stringifyForDiagnostics`), then `json-output.diagnostics.test.ts`; the other two sites are the same shape.

## Detailed testing steps

At head `1b178b488d48` (node budgets added after review): core json-output.diagnostics 3/3 + __tests__/json-output 22/22, ui tts-debug 13/13, personal-assistant redact-sensitive-data 3/3 + truncate-suffix-reserve 6/6, core tsc 0 errors, Biome clean.

Earlier head:

RED on develop `822aba345b67` with the new cases and the unchanged serializers, one per package:

```text
core  × renders a shared (non-cyclic) reference in full at every site        (1 failed | 1 passed)
ui    × renders a shared (non-cyclic) diagnostic value in full at every site (1 failed | 11 passed)
pa    × redacts a shared (non-cyclic) reference in full at every site        (1 failed | 1 passed)
```

The true-cycle cases pass on both sides, pinning that `"[Circular]"` still appears for a real cycle and that the cycle's siblings still serialize.

With the fix:

- `packages/core`: `json-output.diagnostics.test.ts` 2/2; existing `__tests__/json-output.test.ts` 22/22; `tsc --noEmit -p tsconfig.json` 0 errors.
- `packages/ui`: `tts-debug.test.ts` 12/12 (11 existing + 1).
- `plugins/plugin-personal-assistant`: `redact-sensitive-data.test.ts` 2/2 and `truncate-suffix-reserve.test.ts` 6/6 (the existing consumer).
- Biome on all six files — clean.
- Root `bun run verify` (covers the ui and personal-assistant typechecks on a full install) — result posted as a comment at this head.

# Evidence Gate

- [ ] N/A - serializer internals, no rendered UI surface

- [ ] N/A - serializer internals, no rendered UI surface

- [x] the TTS debug log line before and after is the asserted string in `tts-debug.test.ts`

- [x] the trajectory diagnostics serializer's output before and after is asserted in `json-output.diagnostics.test.ts`

- [x] serialized snapshots before and after, asserted in the suites above

### DIFF
--- a/packages/core/src/runtime/json-output.diagnostics.test.ts
+++ b/packages/core/src/runtime/json-output.diagnostics.test.ts
@@ -34,4 +34,27 @@ describe("stringifyForDiagnostics reference handling", () => {
 			child: { parent: "[Circular]", tag: "t" },
 		});
 	});
+
+	it("bounds a diamond-shaped shared graph instead of expanding it exponentially", () => {
+		// 21 distinct objects, each holding the same next level under two keys:
+		// 2^20 paths. The serializer must stop at the node ceiling (50 000
+		// objects) and keep the event readable instead of overflowing.
+		let level: Record<string, unknown> = { leaf: true };
+		for (let i = 0; i < 20; i++) level = { l: level, r: level };
+		const started = performance.now();
+		const text = stringifyForDiagnostics({ event: "probe", payload: level });
+		const elapsedMs = performance.now() - started;
+		expect(text).toContain('"event": "probe"');
+		expect(text).toContain('"[Truncated]"');
+		let objects = 0;
+		const walk = (value: unknown): void => {
+			if (value && typeof value === "object") {
+				objects += 1;
+				for (const item of Object.values(value)) walk(item);
+			}
+		};
+		walk(JSON.parse(text));
+		expect(objects).toBeLessThanOrEqual(50_000);
+		expect(elapsedMs).toBeLessThan(5_000);
+	});
 });
--- a/packages/core/src/runtime/json-output.diagnostics.test.ts
+++ b/packages/core/src/runtime/json-output.diagnostics.test.ts
@@ -1,0 +1,37 @@
+/**
+ * Pins the reference handling of `stringifyForDiagnostics`, the serializer
+ * behind trajectory and evaluator diagnostics: a shared object must render
+ * in full at every site, and only a reference back to an open ancestor may
+ * collapse to "[Circular]". Deterministic; no mocks.
+ */
+
+import { describe, expect, it } from "vitest";
+import { stringifyForDiagnostics } from "./json-output.js";
+
+describe("stringifyForDiagnostics reference handling", () => {
+	it("renders a shared (non-cyclic) reference in full at every site", () => {
+		const shared = { id: 1, tag: "t" };
+		const out = JSON.parse(
+			stringifyForDiagnostics({ a: shared, b: shared, list: [shared, shared] }),
+		);
+		expect(out).toEqual({
+			a: { id: 1, tag: "t" },
+			b: { id: 1, tag: "t" },
+			list: [
+				{ id: 1, tag: "t" },
+				{ id: 1, tag: "t" },
+			],
+		});
+	});
+
+	it("still collapses a true cycle to [Circular] and serializes its siblings", () => {
+		const root: Record<string, unknown> = { name: "root" };
+		root.self = root;
+		root.child = { parent: root, tag: "t" };
+		expect(JSON.parse(stringifyForDiagnostics(root))).toEqual({
+			name: "root",
+			self: "[Circular]",
+			child: { parent: "[Circular]", tag: "t" },
+		});
+	});
+});
--- a/packages/core/src/runtime/json-output.ts
+++ b/packages/core/src/runtime/json-output.ts
@@ -227,7 +227,17 @@ export function stringifyForModel(value: unknown): string {
 	return serialized;
 }
 
-/** Serialize diagnostic context without allowing hostile or cyclic values to mask the original event. */
+/**
+ * Object nodes one diagnostic serialization may emit before collapsing to
+ * "[Truncated]". Diagnostic payloads are small; the ceiling exists for shared
+ * graphs, which are otherwise re-expanded once per path.
+ */
+const MAX_DIAGNOSTIC_NODES = 50_000;
+
+/**
+ * Serialize diagnostic context without allowing hostile, cyclic, or
+ * exponentially shared values to mask the original event.
+ */
 export function stringifyForDiagnostics(value: unknown): string {
 	if (typeof value === "string") return value;
 	// The replacer sees each value with its holder as `this`, so the open
@@ -236,6 +246,11 @@ export function stringifyForDiagnostics(value: unknown): string {
 	// ancestor circular; an object reached twice through different paths
 	// (a DAG) serializes in full at every site.
 	const ancestors: object[] = [];
+	// A shared object is emitted once per path, which is exponential in the
+	// sharing depth; past the node ceiling every further value collapses to a
+	// marker so the diagnostic can neither exhaust the heap nor overflow the
+	// string limit and mask the event it describes.
+	let remainingNodes = MAX_DIAGNOSTIC_NODES;
 	try {
 		const serialized = JSON.stringify(
 			value,
@@ -249,6 +264,8 @@ export function stringifyForDiagnostics(value: unknown): string {
 						ancestors.pop();
 					}
 					if (ancestors.includes(nestedValue)) return "[Circular]";
+					if (remainingNodes <= 0) return "[Truncated]";
+					remainingNodes -= 1;
 					ancestors.push(nestedValue);
 				}
 				return nestedValue;
--- a/packages/core/src/runtime/json-output.ts
+++ b/packages/core/src/runtime/json-output.ts
@@ -230,15 +230,26 @@ export function stringifyForModel(value: unknown): string {
 /** Serialize diagnostic context without allowing hostile or cyclic values to mask the original event. */
 export function stringifyForDiagnostics(value: unknown): string {
 	if (typeof value === "string") return value;
-	const seen = new WeakSet<object>();
+	// The replacer sees each value with its holder as `this`, so the open
+	// ancestor chain is the stack of holders above the current one. Popping
+	// back to the holder before the check makes only a reference to an open
+	// ancestor circular; an object reached twice through different paths
+	// (a DAG) serializes in full at every site.
+	const ancestors: object[] = [];
 	try {
 		const serialized = JSON.stringify(
 			value,
-			(_key, nestedValue: unknown) => {
+			function replacer(this: unknown, _key, nestedValue: unknown) {
 				if (typeof nestedValue === "bigint") return `${nestedValue}n`;
 				if (nestedValue && typeof nestedValue === "object") {
-					if (seen.has(nestedValue)) return "[Circular]";
-					seen.add(nestedValue);
+					while (
+						ancestors.length > 0 &&
+						ancestors[ancestors.length - 1] !== this
+					) {
+						ancestors.pop();
+					}
+					if (ancestors.includes(nestedValue)) return "[Circular]";
+					ancestors.push(nestedValue);
 				}
 				return nestedValue;
 			},
--- a/packages/ui/src/utils/tts-debug.test.ts
+++ b/packages/ui/src/utils/tts-debug.test.ts
@@ -53,6 +53,20 @@ describe("ttsDebug", () => {
     );
   });
 
+  it("bounds a diamond-shaped shared detail instead of expanding it exponentially", () => {
+    // 17 distinct objects, 2^16 paths: the line must stay bounded and keep
+    // its phase instead of growing past the serializer's limits.
+    let level: Record<string, unknown> = { leaf: true };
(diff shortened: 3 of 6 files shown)


attentionhead (changes requested):
**REQUEST_CHANGES** at `50ffa6484828f85771d877b8717c5b19d7061e49` (`MERGEABLE`/`UNSTABLE`, test-merged onto `origin/develop` clean).

The defect is real, and the `this`-holder trick in the two replacer sites is genuinely correct — I tried to break it and could not. I wrote an independent reference implementation of ancestor-path semantics (recursive, `finally`-popped) and diffed it against `stringifyForDiagnostics` on six shapes:

```text
MATCH   shared under two keys
MATCH   shared in one array twice
MATCH   shared at mixed depths
MATCH   true cycle + siblings
MATCH   cycle and sharing together
MATCH   array of arrays sharing
```

All six DIVERGE on `origin/develop` except the pure-cycle one. The mixed case is the most convincing, because develop gets it wrong in a way that is easy to miss:

```text
develop  {"x":{"id":"s","n":1},"y":"[Circular]","z":{"w":"[Circular]","v":"[Circular]"}}
merged   {"x":{"id":"s","n":1},"y":"[Circular]","z":{"w":{"id":"s","n":1},"v":"[Circular]"}}
```

`z.w` is a shared value and `z.v` is a real cycle; only the merged head tells them apart. The three suites are green on the merged head: `json-output.diagnostics.test.ts` 2/2, `tts-debug.test.ts` 12/12, `redact-sensitive-data.test.ts` 2/2.

The problem is the same one as #31002 and #30995, and here it is worse, because **none of these three sites has a depth cap or a size cap at all.**

## Measured: the diagnostic destroys the event it exists to report

`stringifyForDiagnostics`, fan-chain of `levels+1` distinct objects, no cycle:

| distinct objects | keys/level | develop chars | merged chars | merged time |
|---|---|---|---|---|
| 11 | 2 | 667 | 130,038 | 0 ms |
| 15 | 2 | 1,095 | 2,736,118 | 8 ms |
| 19 | 2 | 1,619 | 54,263,798 | 134 ms |
| 13 | 3 | 1,265 | 65,367,233 | 168 ms |
| 23 | 2 | 2,239 | **15** | **2,284 ms** |
| 17 | 3 | 1,937 | **15** | **15,437 ms** |

Those last two rows are the finding. Past ~500 MB, `JSON.stringify` throws `RangeError: Invalid string length`, the `// error-policy:J7` catch swallows it, and the function returns `formatError(value)`:

```text
23 distinct objects -> ms=2766 len=15 value="[object Object]"
contains original event marker? false
```

So after burning 2.8 seconds, a diagnostic for a 23-object graph is the string `[object Object]`, with the original event's payload gone. The docblock on this function reads *"Serialize diagnostic context without allowing hostile or cyclic values to mask the original event"*, and the J7 comment reads *"diagnostic serialization must not mask the event being reported"*. After this change, a graph that is neither hostile nor cyclic — just shared — masks it. That is a direct regression against the function's stated contract, and it is silent.

The same applies to `serializeTtsDebugDetail`, whose J4 catch returns `"[Unserializable diagnostic detail]"`, and to `redactValue`, which has no cap either and is the one of the three that returns a structure rather than a string — so it has no `RangeError` backstop at all and will simply exhaust the heap.

## Smallest repair

Thread a node or character budget alongside the existing walk in each of the three, collapsing to a `"[Truncated]"`-style marker when it is exceeded — the same way each already has a terminal marker for the cyclic case. `redact-sensitive-data.ts` needs it most, since it has no throwing backstop.

If a budget is too much for this PR, the alternative that keeps it small is to bound depth in all three, the way `settings-debug.ts` already does with `MAX_DEPTH`. That does not remove the exponential, but it caps the exponent at a known number instead of leaving it at the input's depth.

Either way, please add a case that pins the bound. The new cases pin the correctness change in both directions, which is right — but termination is the property this diff changes, and nothing currently pins it.

## What I ran

- Test-merge: `git checkout --detach origin/develop && git merge 50ffa648` → clean.
- `packages/core`: `npx vitest run --config packages/core/vitest.config.ts packages/core/src/runtime/json-output.diagnostics.test.ts` → 2 passed.
- `packages/ui`: `npx vitest run --config ./vitest.config.ts src/utils/tts-debug.test.ts` → 12 passed.
- `plugins/plugin-personal-assistant`: `npx vitest run --config vitest.config.ts src/lifeops/redact-sensitive-data.test.ts` → 2 passed.
- Reference-implementation diff and the fan-out table above, run on the merged head **and** on `origin/develop` alone. Probes deleted, tree restored.

hermesagent270-commits:
Root gate at head `50ffa6484828f85771d877b8717c5b19d7061e49`: `RUN_TURBO_CONCURRENCY=4 bun run verify` on a fully installed checkout finished with

```text
[anti-larp] scanned 11670 test files — 0 focused, 0 orphaned skip(s)
exit 0
```

That covers the `packages/ui` and `plugin-personal-assistant` typechecks on a full install, which the body deferred to this run. Package gates already in the body: core 2/2 + 22/22, ui 12/12, personal-assistant 8/8, core `tsc` 0 errors, Biome clean on all six files.

hermesagent270-commits:
@attentionhead reproduced, including the part that matters most: on my previous head a 23-object fan chain through `stringifyForDiagnostics` ended as `[object Object]` after the `RangeError`, so the diagnostic masked its own event. Pushed `1b178b488d486eeb415cc97737026dd087d2ef80` on top:

- `stringifyForDiagnostics`: a per-call node budget (`MAX_DIAGNOSTIC_NODES = 50_000`) inside the replacer, checked after the cycle test; once spent every further object becomes `"[Truncated]"`. The event's own fields serialize first, so a bounded payload can no longer push the string past the engine limit.
- `serializeTtsDebugDetail`: same shape, `MAX_DEBUG_DETAIL_NODES = 10_000`.
- `redactValue`: a `RedactBudget` threaded beside `seen` (`MAX_REDACT_NODES = 50_000`); this is the one with no throwing backstop, so the budget is its only ceiling, and the comment says so.

The `this`-holder ancestor tracking is unchanged in the two replacers, and the six shapes you diffed still hold (the shared-reference and true-cycle cases in each suite pass unchanged).

Three pinning cases, one per site: a 2^20-path diamond (21 objects) for the diagnostics serializer must keep `"event": "probe"`, contain `"[Truncated]"`, parse back to at most 50 000 objects, and finish under 5 s (~80 ms); a 2^16 diamond for the TTS line must keep its phase prefix, contain the marker, and stay under 2 M characters; a 4^12 chain (13 objects) for redaction must produce the marker with at most 50 000 objects. Each fails on the previous head by node count or by the `RangeError` fallback.

At this head: core `json-output.diagnostics.test.ts` 3/3 and `__tests__/json-output.test.ts` 22/22, ui `tts-debug.test.ts` 13/13, personal-assistant `redact-sensitive-data.test.ts` 3/3 + `truncate-suffix-reserve.test.ts` 6/6, core `tsc --noEmit` 0 errors, Biome clean on all six files. Root verify is re-running on this head and will be posted below.

ss251 (approved):
I'll read the full review brief, the published reviewer skill, and the complete PR diff before writing the review body.1b178b488d486eeb415cc97737026dd087d2ef80

Input `{ a: shared, b: shared }` with `shared = { id: 1, tag: "t" }` previously serialized `b` as `"[Circular]"` because `stringifyForDiagnostics` kept a WeakSet of every visited object. Required output is the object in full at every site; only a back-edge to an open ancestor may collapse. The holder-stack replacer at `packages/core/src/runtime/json-output.ts:257-269` pops until `this` is the current holder, then uses `ancestors.includes(nestedValue)` for the cycle check. Same holder-stack in `packages/ui/src/utils/tts-debug.ts:85-97`. The redaction walker does the path equivalent: `seen.add` then `seen.delete` in `finally` at `plugins/plugin-personal-assistant/src/lifeops/redact-sensitive-data.ts:163-193`, so a shared value is redacted at both keys.

Those tests fail if the WeakSet returns: `packages/core/src/runtime/json-output.diagnostics.test.ts:11-24`, `packages/ui/src/utils/tts-debug.test.ts:47-55`, `plugins/plugin-personal-assistant/src/lifeops/redact-sensitive-data.test.ts:8-16`. Cycle pins remain at `json-output.diagnostics.test.ts:27-35` and `redact-sensitive-data.test.ts:19-27`.

Expanding DAGs is exponential; the node ceilings (50_000 / 10_000 / 50_000) and `"[Truncated]"` at `json-output.ts:267`, `tts-debug.ts:95`, `redact-sensitive-data.ts:167-182` are load-bearing. The three diamond tests assert the marker and the bound. No lockfile, CI, dependency, binary, or weakened assertion. Three serializers, three budgets, one behavior — not split-PR flooding.

Verified by reading the six-file diff at this head. Live execution blocked: no disposable sandbox on this host, so this is a static review of the diff at the exact head.

Recommendation: merge

— [free-xai-lane-ss251]

hermesagent270-commits:
Root gate at head `1b178b488d486eeb415cc97737026dd087d2ef80`: `RUN_TURBO_CONCURRENCY=4 bun run verify` on a fully installed checkout finished with

```text
[anti-larp] scanned 11670 test files — 0 focused, 0 orphaned skip(s)
exit 0
```

That covers the `packages/ui` and `plugin-personal-assistant` typechecks on a full install. Package gates at this head: core `json-output.diagnostics.test.ts` 3/3 + `__tests__/json-output.test.ts` 22/22, ui `tts-debug.test.ts` 13/13, personal-assistant `redact-sensitive-data.test.ts` 3/3 + `truncate-suffix-reserve.test.ts` 6/6, core `tsc` 0 errors, Biome clean on all six files.

lalalune:

Closing this implementation while retaining the real shared-reference defect in #31004. The patch replaces values beyond 50,000 object nodes with [Truncated] in stringifyForDiagnostics. That function is not confined to UI previews: the current relationships provider serializes metadata into provider context with it, and trajectory recording also consumes it. The added test explicitly expects data loss, rather than complete content, lossless reassembly or actionable rejection. A replacement must preserve those contracts and review sink-specific behavior; arbitrary graph caps cannot be applied to full provider/recorded content.

Issue #30998, opened 2026-09-10
fix(lifeops): use the agent graph for owner-created email recipients


lalalune:
The owner contact editor and monthly email workflow address different EntityStore partitions. In the isolated VPS pilot at c16b28b35b031bbae1d948a9c666ae821ba9b047, POST /api/lifeops/entities saved a synthetic owner-confirmed email recipient, but GET /api/lifeops/family-workflows/email-options still returned recipients: []. No email was sent.

Source: routes/entities.ts makeStore passes adminEntityId when present; family-workflows/runtime.ts emailOptions and validateRecipientIdentity pass runtime.agentId. EntityStore documents and persists a per-agent partition. Authenticated principal binding and the other family consumers also use runtime.agentId. Contacts created through the owner API therefore cannot be used by the monthly email flow.

Acceptance:
- Owner contact create/read/update/resolve and monthly email recipient selection use the canonical agent partition; actor identity remains authorization context.
- A real HTTP/PGlite test uses distinct owner and agent IDs, creates a verified synthetic contact through the registered route, then retrieves it through the real email-options consumer and validates the draft recipient.
- A second agent and unauthenticated/non-owner caller cannot read or mutate that contact.
- Existing records are preserved; any legacy partition migration requires explicit ownership mapping and must not copy one owner's contacts into unrelated agents.
- Repeat live self-test setup after deployment, then complete the approved email flow to the owner's test account only.

Part of #30123; full monthly email acceptance remains incomplete.

lalalune:
The same partition defect also affects the relationship HTTP routes: a co-parent edge created through the normal route is absent from the agent's household graph. The HTTP/PGlite regression now reproduces both failures. With the canonical agent partition in both route factories, the contact is available to family email and the co-parent edge is available to household consumers; neither appears in another agent's partition.

Validation so far: the contact-only change passed root verification, package typecheck/lint, 402 source integration tests and 24 repository smoke integration tests. The expanded relationship regression failed before its fix and passed after. Final checks are rerunning after adding explicit legacy-record preservation coverage.

Legacy owner-partition rows remain stored and are not silently copied to any agent. An explicit ownership-mapping/import path is still required before treating existing-user migration as complete. Authenticated non-owner acceptance and the deployed monthly email self-test also remain open. This update does not claim release readiness.

PR #31006, opened 2026-09-10
fix(lifeops): share the agent graph across contact and family workflows

lalalune:
## Problem and behavior

Contacts created through the HTTP API used the owner's partition while relationship resolution and email recipient validation used the agent's partition. With distinct owner and agent IDs, a newly created contact could therefore disappear from recipient lookup. Entity and relationship route stores now use the runtime agent partition.

Existing owner-partition records are preserved. An owner/admin can GET `/api/lifeops/entities/legacy-owner-graph` to review the complete fixed-source/fixed-target graph, then POST its `reviewSha256` to adopt that exact graph transactionally. The move preserves contact and relationship IDs, identities, attributes, and audit records; retains each partition's self record; rejects stale confirmation, collisions, missing endpoints, duplicate active edges, and replay; and records an audit receipt. No migration runs automatically and no message is sent.

Addresses #30998. The broader deployed family journey remains separate acceptance work.

## Validation

Candidate `4d87e141c90f0`: pinned Bun 1.3.14 / Node 24.15.0 install and root `bun run verify` passed, including package build, typecheck, lint and repository audits. The real HTTP/PGlite regression exercises distinct owner/agent scopes, create/read, recipient lookup/validation, second-agent isolation, unauthenticated and non-owner denials, stale review, collision rollback, concurrent adoption, preserved relationships, and replay rejection. The exact-head HTTP/PGlite regression passed. The full owning-package unit suite passed 2,744 tests across 309 files (6 skipped), and the full source integration suite passed 448 tests across 53 files. Earlier run 35160159664 passed all five hosted checks on this exact head: source static smoke, billing payment replay, PostgreSQL subscription authority, Windows browser security, and the final aggregate. Local validation is complete. The ready-for-review run [35163153048](https://github.com/elizaOS/eliza/actions/runs/35163153048) attempt 1 terminated during typecheck with exit 143 and no reported TypeScript diagnostic; the underlying termination cause remains unproven. One unchanged-head retry (attempt 2) passed all five checks, including typecheck and build. The PR was subsequently merged after exact-head verification. The failed attempt is retained as evidence. This scoped merge does not complete the broader #30998 acceptance criteria.

No user records were migrated, no connector was authorized, and no live email was sent. This PR does not claim deployed provider or live-model acceptance.

## Inspectable evidence

| Evidence | Result and artifact |
| --- | --- |
| Source and integrity | [Manifest with exact source blobs and artifact SHA-256 hashes](https://github.com/elizaOS/eliza/releases/download/pr-evidence-13/31006-4d87e141-manifest.json); uploaded asset digests verified against local files. |
| HTTP and database | [Complete synthetic request/response bodies and real PGlite graph-store readback](https://github.com/elizaOS/eliza/releases/download/pr-evidence-13/31006-4d87e141-http-database-evidence.json). Authentication headers and session identifiers are excluded. |
| Capture reproducibility | [Instrumented copy of the real regression](https://github.com/elizaOS/eliza/releases/download/pr-evidence-13/31006-4d87e141-evidence-harness.ts), [passing capture log](https://github.com/elizaOS/eliza/releases/download/pr-evidence-13/31006-4d87e141-evidence-capture.log), [unchanged tracked regression log](https://github.com/elizaOS/eliza/releases/download/pr-evidence-13/31006-4d87e141-final-focused.log). Production source remained unchanged and the worktree was clean after capture. |
| Package tests | [2,744 unit passes](https://github.com/elizaOS/eliza/releases/download/pr-evidence-13/31006-4d87e141-unit.log), [448 integration passes](https://github.com/elizaOS/eliza/releases/download/pr-evidence-13/31006-4d87e141-integration-full.log). |
| Repository gates | [Pinned installation](https://github.com/elizaOS/eliza/releases/download/pr-evidence-13/31006-4d87e141-install.log), [root verification](https://github.com/elizaOS/eliza/releases/download/pr-evidence-13/31006-4d87e141-verify.log), [earlier five passing hosted checks](https://github.com/elizaOS/eliza/actions/runs/35160159664); [unchanged-head retry: all five checks passed; attempt 1 retained](https://github.com/elizaOS/eliza/actions/runs/35163153048). |
| Desktop/mobile screenshots, walkthrough, frontend logs, OCR | N/A: this change is an HTTP/storage contract; it adds no view or visual behavior. |
| Live model trajectory, provider delivery, audio/native captures | N/A: no model, delivery, audio, or native implementation changes. Synthetic identity validation does not claim live email delivery. |

The real database readback shows the original contact IDs and co-parent relationship in the agent partition, absent from the former owner partition and an unrelated agent partition. Recorded failures include 401/403 authorization, stale-review 409, collision rollback, one successful concurrent adoption with the second attempt rejected, and replay rejection.


attentionhead (changes requested):
**REQUEST_CHANGES** at `bd3ceddcb65e5ba6b8203b1891e04311f1de8b34` (draft; head moved from `f7170224` while I was reviewing, so everything below is at `bd3ceddc`, test-merged onto `origin/develop` clean).

The diagnosis is right and the direction is right. I checked the convention rather than taking it on faith: of **34 non-test `getEntityStore` / `getRelationshipStore` call sites in this plugin, 32 key on `runtime.agentId`**. The only two that did not were exactly the two route factories this PR changes. So this is not picking a new partition — it is bringing two outliers onto the one every other consumer already uses.

I also confirmed the isolation question, because "share the agent graph" reads alarming in a diff. `ctx.state.adminEntityId` is `routeOwnerEntityId(runtime)` → `resolveOwnerEntityIdOrDefault(runtime)` (`routes/plugin.ts:104`), which is derived from the **runtime**, not from the request principal — the per-request actor is the separate `ctx.state.requestEntityId` (`routes/plugin.ts:203`). Both the old and new keys are therefore per-runtime constants, so no cross-actor partition is being collapsed. Actor identity still authorizes, as your comment says.

Your regression test reproduces cleanly in both directions — `AssertionError: expected null to match object` on `origin/develop`, 1 passed on the merged head, under `vitest.src-integration.config.ts`.

## The blocker: every existing install loses its HTTP-created contacts, not just configured-owner ones

The PR body says migration "still needs explicit ownership mapping", which reads as though the gap is scoped to installs that configured an owner. It is not. Measured, for `agentId = 11111111-2222-3333-4444-555555555555`:

```text
no owner configured (default install)
   develop key = 16000e1a-2894-045d-9603-bbb7091337f0
   merged  key = 11111111-2222-3333-4444-555555555555
   same? false
```

`resolveOwnerEntityIdOrDefault` falls back to `deterministicOwnerEntityId(runtime.agentId)` (`packages/core/src/roles.ts:481`), which is a *derived* UUID, never the agent id itself. So the partition key changes on **every** install that has ever used the HTTP contact editor, including the plain default one. After this lands, those records are still in the database — your test correctly asserts they are preserved and not implicitly copied — but nothing reads them any more, and the editor shows an empty contact list with no error.

That is a silent data-disappearance on upgrade, and it is the one outcome worse than the bug being fixed: today a contact is invisible to the *email workflow*, after this it is invisible *everywhere*, with no signal that anything happened.

## What would unblock it

Either of these, both small relative to the change already here:

1. A read-through fallback in the two factories: read the agent partition, and on miss also read the legacy owner partition, so existing records stay visible while new writes go to the canonical key. That makes the change safe to land before the migration exists.
2. A one-shot migration that re-keys the owner partition's records to the agent partition, gated so it runs once.

If neither belongs in this PR, then the missing piece is at least a startup check that counts records under the legacy key and logs a loud warning naming the migration — so an operator finds out from a log line rather than from an empty list.

Please also add a case that pins whichever you choose. The current integration test pins that legacy records are *preserved*, which is true and worth pinning, but preservation without reachability is exactly the failure mode above — nothing currently distinguishes "kept and readable" from "kept and orphaned".

Keeping this draft until then is the right call, and the PR body is admirably clear about what has and has not been verified at this head.

## What I ran

- Re-guarded at `bd3ceddc` after the head moved off `f7170224` mid-review.
- Test-merge onto `origin/develop` → clean.
- `vitest.src-integration.config.ts src/routes/entity-email-recipient.pglite.integration.test.ts`: **1 failed** on `origin/develop` with the PR's test applied to develop's routes, **1 passed** on the merged head.
- Partition-key probe on `resolveOwnerEntityIdOrDefault` vs `runtime.agentId`, output above. Probe deleted, tree restored.
- Call-site census of the 34 non-test store constructions in `plugins/plugin-personal-assistant/src`.

One note in case it saves someone else the detour: this test is excluded from the plugin's default `vitest.config.ts` (`**/*.integration.test.ts` is in `exclude`), so a plain `vitest run` in the package reports "No test files found" for it. It does run under `bun run test:integration`. Not a problem with the PR — just not discoverable from the package's default lane.

ss251 (changes requested):
bd3ceddcb65e5ba6b8203b1891e04311f1de8b34

Findings:

1. entities.ts:38-40 and relationships.ts:39-40 drop the `ctx.state.adminEntityId` branch; both stores now key on `defaultAgentId(ctx.state.runtime)`. Input -> output: records previously written under an actor partition become unreachable through these routes — the new test codifies exactly that (legacy actor-partition record: GET /entities/:id -> 404, absent from the agent store, never copied). Deliberate and tested; still a silent visibility break for pre-existing data with no migration in the diff. State that in the PR body if it is accepted pre-release.

2. With per-actor keying gone, every caller passing `isAuthorized` shares one `owner_only` partition. The comment at entities.ts:38-39 ("the authenticated actor authorizes this request") assumes a single actor. If two distinct admin entities can both be authorized, each gains read access to the other's owner_only contacts and co_parent edges. Confirm the single-owner invariant upstream, or enforce actor scope in the route rather than in a comment.

3. Test weakening in an out-of-scope hunk: calendar-optional-tool-parameters.test.ts:45-66 upgrades the checks to behavioral `validateSchema` assertions (omitted optional field passes; `travelOriginAddress: false` fails — stronger than the old shape checks), but the same hunk deletes `expect(tool.strict).toBe(false)`. No hunk in this PR touches tool strictness, so the deleted guard either still passes (pure removal) or would now fail (unmentioned behavior change). Restore it or explain the removal; strict mode flips exactly the optional-parameter semantics this file guards.

The strong part: the new 171-line PGlite integration test drives real HTTP through `tryHandleRuntimePluginRoute`, asserts actor and agent ids differ up front (line 19), expects 401 on unauthenticated write and read, shows a contact created via POST /entities surfacing in family-workflows/email-options, checks entity and relationship isolation from "another-agent", and verifies the legacy record is preserved byte-for-byte. It would fail if the partition consolidation broke.

Verified by reading the diff at this head; no suite run — live execution blocked on this host.

Recommendation: fix

lalalune:

Retaining the canonical agent-graph partition fix. Current entity and relationship routes select the admin actor partition while downstream agent workflows use the agent graph, so this addresses an actual producer/consumer mismatch. The HTTP/PGlite contract regression is appropriate. Keep actor authorization separate from record partitioning, and preserve explicit legacy-record adoption and authenticated non-owner acceptance as remaining work; do not silently copy legacy records or treat this narrow change as completion of #30998/#30123. The approved monthly delivery test and deployed behavior remain distinct from local database proof.

plugins/plugin-personal-assistant/src/routes/entity-email-recipient.pglite.integration.test.ts
```diff
@@ -0,0 +1,307 @@
+/**
```
jianqiaotan26-prog, line 1: 6666

jianqiaotan26-prog (commented):
6666y

Issue #17030, opened 2026-07-23
Chat trajectories


lalalune:
Record model calls, context choices and tool results for each app-chat turn.

### Project / Coordination

- Project board: https://github.com/orgs/elizaOS/projects/15
- Parent tracker or epic: #17022
- Coordination Discussion: https://github.com/elizaOS/eliza/discussions/14292
- Suggested lane tag: `chat-trajectory-proof`

### Alpha Phone allocation

Owner: Nubs. September 11; one working day including the Alpha-relevant review/fixes below. Consolidates #30291 and #30734 without adding days or making batching a required deliverable.

- Reproduce the stats-width report on the current stats path. #30291 records a real database overflow at lifetime totals above 2^31-1; it is not evidence that current develop is still broken. Review existing #30292/#30461 before changing code. If still reproducible, preserve wide aggregate totals and verify the real endpoint beyond the boundary.
- Measure recorder overhead in a representative app-chat trajectory. #30734 reports repeated writes, but does not establish their contribution to end-to-end latency. Batch only if current measurements justify it; otherwise leave persistence unchanged and record that result.
- Any batching must preserve every record, ID, timestamp, parent/ordinal link, owner ordering, write barriers, capture/completion order, errors and flush/lifecycle behavior. Prove record equivalence with a real database and report transaction reduction separately from user-visible latency.

The required outcome is complete, usable chat trajectories. No standalone stats-hardening project or speculative persistence rewrite is added. Original reports retain their reproduction history.

### Scope

Make ordinary app-chat trajectories record the decisions needed to diagnose routing, context, and false-success failures.

The reported bad-turn trajectory showed `providerAccessCount:36` but `llmCallCount:0`, even though Stage 1 and the planner necessarily used configured models. The persisted trajectory therefore could not answer:

- what Stage 1 received and returned;
- which contexts/intents/candidate actions were selected;
- which model/provider handled Stage 1 and planning;
- which actions/tools were exposed, called, and returned;
- why VIEWS won over an owner routine;
- whether the final reply had a matching mutation receipt.

Scenario-runner file trajectories have richer stage data; bring the real app/database trajectory path up to the same diagnostic standard without leaking secrets or unbounded private context.

[...]

PR #31014, opened 2026-09-10
fix(dev): correct live timing and keep trajectory context usable

NubsCarson:
The developer inspector can show a previous prompt's age while a new submission is still being relayed, and complete provider payloads can stretch diagnostics across tens of thousands of pixels. Start the live timer at submission, exclude stale live activity until the new user row arrives, and put complete provider text in a bounded, keyboard-scrollable area. Clarify that provider-returned characters are not the amount sent to the model, and tighten inspector spacing without changing normal app chrome.

Related: #17030, #31330. @lalalune

## Current scope and integration

Head `6ea574ef536fd96b9c7f86cfbfd322ece8e5acbc` ordinarily merges develop `886df6eba0c33e40da1794f5e632da6c094bedea`; original branches are preserved, with no rebase or force-push. Most of the original viewer, owner authorization, scoped foreground query and settings-preflight work is already in develop. The remaining diff is six files: the timer and regression test, provider-size wording, compact text/selector layout, bounded provider scrolling and access documentation.

The saved spacing change from the older local preview is retained. No model request, token accounting, provider output, stored trajectory or normal app layout is rewritten by these refinements. A very large provider result remains fully inspectable and copyable.

## Validation

- Full UI suite: **1,320 files passed; 14,352 tests passed, 7 existing skips**. Includes the delayed-relay timer regression, message/run ownership, settings and tab relay cases.
- Repository verification on the final candidate completed **373/373 tasks and final audits**. Changed-file Biome and real browser interaction also passed.
- Strict evidence-toolchain doctor passed. The required `audit:app` run completed **226 checks passed** against this checkout on isolated ports. OCR reviewed220 captures:208 verified,0 broken,12 explicit semantic exemptions for unavailable Camera/Cockpit/LifeOps fixture surfaces. Trajectory captures are verified across all four viewports; desktop/mobile empty-state captures were manually inspected. This is a hermetic app audit, not live cloud/domain acceptance.
- Real Chromium inspection used this frontend and the existing copied QA runtime on31341, whose backend is the combined context checkpoint, not this isolated backend. It is frontend/recorded-data evidence, not a current-head backend performance benchmark or a new live-model run.

## Measured browser behavior

| Check | Before scroll correction | After |
| --- | --- | --- |
| Desktop diagnostics outer scroll height | 69,757px | 1,446px |
| Mobile diagnostics outer scroll height | 95,668px | 1,702px |
| Provider text copied | 350,677 characters | Same complete text and SHA-256 |
| Provider text area | Expanded into the diagnostics page | 500px desktop / 422px mobile, keyboard-scrollable |

The copied text exactly matches a string in the real returned trajectory payload. All three recorded model attempts are selectable with19,681 +10,602 +7,682 input tokens, matching the37,965 total. At1440x1000 and390x844 there is no document-level horizontal overflow; long provider text wraps on mobile. Closing the dialog restores focus to Inspect. The separate normal `/chat` tab has zero developer-workspace mounts. Inspection issued no new model calls.

## Evidence gate — still draft

- [ ] Locally inspected desktop/mobile captures of the unbounded provider layout are saved; public-safe inline artifacts remain pending.

- [ ] Locally inspected desktop/mobile captures verify wrapping and bounded scrolling; public-safe inline artifacts remain pending.

- [ ] Chromium recordings are saved locally; complete public-safe walkthrough review remains pending.

- [ ] Exact-head backend acceptance remains pending. This frontend check used stored real trajectories from the existing copied QA backend.

```text
Viewport1440x1000: inner provider scrollTop0->233; height500; outer height608/scrollHeight1446.
Viewport390x844: inner provider scrollTop0->30; height422; outer height532/scrollHeight1702.
Both copied350677characters; both exactly match the recorded response payload.
Both have document horizontal overflow0. Dialog closes; focus returns to Inspect.
Initial startup WebSocket error was traced to the QA launcher pointing Vite HMR at the API port. Corrected launch: HMR connected on5217, app socket on5217/ws, zero page errors in the fresh startup check. No application code change was needed.
```

- [ ] Stored real inputs/outputs were inspected without new inference. Current-head real-model app acceptance remains pending; prior combined-branch timing is not substituted for it.

- [x] N/A - these remaining frontend changes do not write domain records or alter recorded model payloads.

- [ ] App audit and OCR completed; affected trajectory desktop/mobile captures inspected. Public-safe inline evidence and complete walkthrough review remain pending.

## Remaining acceptance

Attach public-safe screenshots/video/logs and finish the full walkthrough review. The exact-head app audit completed and the QA-launch WebSocket error is resolved. Current-head Home/Notes/Calendar and actual navigation/recovery acceptance remains distinct from the stored-record inspector check. Private conversation captures stay local. This draft does not claim deployment, complete latency optimization, or merge readiness.


attentionhead (changes requested):
**REQUEST_CHANGES** — one concrete regression in the settings save path, reproduced with a bidirectional control. Verified at head `06ada922efd2c011c7844043f6c779f6b64c8e76`; test-merges onto `origin/develop` cleanly, and the PR's own suites are green there (`ModelConfigurationPanel` + `navigation/index` + `app-navigate-view` = 87 passed; core `read-routes` = 10 passed).

## The finding: a failing status probe now blocks the settings write entirely

`useModelConfiguration.ts` adds a status read *before* the write, to capture `startedAt` for the restart check:

```ts
const before = group === "coding" ? null : await client.getStatus();
const previousStartedAt =
  typeof before?.startedAt === "number" ? before.startedAt : null;
const result = await client.updateModelsConfig(request);
```

That `await` is unguarded inside `save`'s `try`. If `getStatus` rejects, control jumps to the `catch`, the group renders `phase: "error"`, and **`updateModelsConfig` is never called** — the user's model settings are silently not saved, on a path where develop saves them.

What makes this worth blocking on rather than noting: the very next function in the same diff documents this call as *expected to fail*.

```ts
} catch {
  // error-policy:J4 status is expected to fail while the agent is down
  // mid-restart; "unreachable" is the designed not-yet-running signal
  // the poll loop keeps waiting on.
```

So `client.getStatus()` is treated as an anticipated failure inside `waitForRuntimeRunning` and as fatal three lines earlier. A user who saves settings while the agent happens to be down — the exact state a model-settings change is often made in — gets an error instead of a save.

### Reproduction

I added one test to the PR's own `ModelConfigurationPanel.test.tsx`, using its existing harness, and made only the probe fail (after the ready render, so the load path is untouched):

```ts
clientMock.getStatus.mockReset();
clientMock.getStatus.mockRejectedValue(new Error("status unreachable"));
act(() => agentButton("models-small-save").click());
act(() => agentButton("models-small-confirm-restart").click());
await waitFor(() =>
  expect(clientMock.updateModelsConfig).toHaveBeenCalledTimes(1),
  { timeout: 2000 });
```

- **This PR's `useModelConfiguration.ts`:** `1 failed` — `updateModelsConfig` is never called, the assertion times out.
- **develop's `useModelConfiguration.ts` + `ModelConfigurationPanel.tsx`, same test file, everything else at the PR head:** `1 passed` — the write goes through.

Same test, same harness, one file swapped: the difference is this change.

### Smallest repair

Make the probe best-effort. `previousStartedAt === null` already has a defined fallback inside `waitForRuntimeRunning` — it waits on the `sawNotRunning` edge, which is exactly develop's pre-existing behaviour:

```ts
let before: Awaited<ReturnType<typeof client.getStatus>> | null = null;
if (group !== "coding") {
  try {
    before = await client.getStatus();
  } catch {
    // error-policy:J4 the probe is an optimization; without it the restart
    // wait falls back to the not-running edge, which is the prior behavior.
  }
}
```

That keeps the real fix (a `startedAt` change proves the *replacement* runtime came up, instead of the old one still answering) without letting the optimization gate the write. Worth a regression test pinning "a failing status probe still saves".

## The rest holds up

I checked the two things a 2,194-line developer-inspector PR most needs checked, and both are sound:

**The `/dev` gate is genuinely three-way.** `isDeveloperWorkspaceRoute` requires `import.meta.env.DEV` **and** a loopback hostname (`127.0.0.1`/`localhost`/`[::1]`) **and** a `^/dev/?$` path, and `main.tsx` evaluates it once at module scope so a production bundle can never mount `DeveloperWorkspace`. The "normal routes ignore the old session opt-in" claim is accurate — there is no `sessionStorage`/`devtools=1` escape left in the mount condition.

**The trajectory read route stays complete by default.** `detailToUi(..., includePayloads = true)` and the caller passes `url.searchParams.get("includePayloads") !== "0"`, so every existing consumer keeps the full `systemPrompt`/`userPrompt`/`response` and full `semanticStages`. Reduction is strictly opt-in, and the response carries `payloadsIncluded` so a caller can tell a summary from the real thing rather than inferring it — that is the right shape for the repo's prompt-integrity rule, which allows a *named* preview but not a silent one.

Two small notes there, neither blocking: the two withholding strategies differ — `llmCalls` **omits** the payload keys while `semanticStages` substitutes `payload: {}`, so at the stage level an emptied payload is indistinguishable from a genuinely empty one except via the top-level flag. And `payloadsIncluded` sits beside `trajectory` rather than inside it, so it is easy for a consumer to destructure past.

## On the draft state

The PR is honest that its gates are not met — `bun run verify` stopped in `@elizaos/ui:lint:check` on a missing `@elizaos/prompts/dist/index.js` (201/212 tasks), and the evidence checkbox is unchecked. I am not counting that against the change; the `ERR_MODULE_NOT_FOUND` looks like an unbuilt-workspace artifact rather than anything this branch introduced. The changes requested here are the save-path regression alone, and it is a small fix.

ss251 (changes requested):
06ada922efd2c011c7844043f6c779f6b64c8e76

Input: save small/large model settings while `client.getStatus()` rejects (agent down — the state this save often runs in).
Output: `updateModelsConfig` is never called; the group shows error. `useModelConfiguration.ts:678` awaits `getStatus()` unguarded inside `performSave`'s `try`, so a probe throw hits `:719` and skips `:681`. The same diff's `waitForRuntimeRunning` (`:635-640`) treats that throw as the designed not-running signal. `previousStartedAt === null` already uses the `sawNotRunning` edge (`:631-634`), so the probe can be best-effort without losing the replacement-runtime check.

Input: `http://[::1]:5173/dev` in a Vite DEV build.
Output: the workspace never mounts. `developer-route.ts:10` allowlists `location.hostname` as `"[::1]"`, but browsers expose IPv6 hostname as `::1`. `main.tsx:2976-2981` (`isLoopbackApiHost`) already accepts both forms. Tests only use `localhost`.

`useDeveloperTrajectories.ts:60-61` polls `getTrajectories({ limit: 50, offset })` with no room/source filter, then `DeveloperWorkspace.tsx:624-625` binds by `roomId` + `metadata.messageId` inside that page, so overflow hides token counts.

Draft: CONTRIBUTING.md requires current-head UI screenshots, MP4, logs, and live trajectories; those rows are pending. The only LLM table is from `a437f0679f1b664ea52e6575bac7a354119532d0`, not this SHA. Tests mock `client`/`RoleGate`/`sendChatText`. Branch `codex/dev-chat-inspector-20260910` does not state provider/model/client.

Recommendation: fix

Verified by reading the diff at this head (21 files, +2194/−72) and CONTRIBUTING.md evidence rules.

Client / agent tooling: grok
— [grok-fleet-ss251]

hermesagent270-commits (changes requested):
Reviewed exact head `90bf4f90b5ced8eaaeb050a6c042ef500056262a`.

**Blocker — a non-owner host session can reach the new complete trajectory payloads.**

This head deliberately expands persisted/readable diagnostics from bounded projections to complete model input/output (`packages/agent/src/runtime/trajectory-storage.ts:840-848`, with the equivalent core path at `packages/core/src/features/trajectories/TrajectoriesService.ts:2089-2097`) and now retains complete provider result text after the audience check (`packages/core/src/runtime/state-composition/composer.ts:606-620,653-665,761-764`). The detail route returns those model, provider, tool, and semantic-stage payloads by default; `includePayloads=0` is only an optional client-side opt-out (`packages/core/src/features/trajectories/read-routes.ts:223-296,503-521`).

The UI correctly wraps the inspector in `RoleGate minRole="OWNER"` (`DeveloperWorkspace.tsx:1164-1167`), but that is not an authorization boundary. In the real HTTP server, any host session for which `resolveHttpRequestAuthorization` returns `{ ok: true }` satisfies the coarse gate (`packages/agent/src/api/server.ts:1901-1918`), regardless of its role, and the trajectory dispatch then receives no request authorization or role at all (`server.ts:3645-3658`). The service filters only by agent, not by the caller's entity or room, so list makes every trajectory ID discoverable and detail returns complete prompts/provider text for every room in that agent.

I reproduced the boundary on the exact head in a disposable, no-network container. A focused test started the production API server with the host bridge returning `role: "USER"` for a machine-session cookie:

- anonymous `GET /api/trajectories/private-run` → `401`
- the same request with the USER session → `503 {"error":"Trajectory service unavailable"}` rather than `403`

The `503` is intentional in the probe because no runtime was installed; it proves the non-owner crossed the server auth gate and entered `tryHandleTrajectoryReadRoutes`. With an active runtime, that handler performs the unrestricted agent-scoped lookup and returns `200`. The probe passed (`1 passed`), and the existing core route plus app-tab relay suites passed `16` focused tests. A broader mixed-package Vitest invocation could not load `@elizaos/prompts` from the disposable image's pre-existing dependency install, so I did not treat that infrastructure failure as a code result.

Please enforce OWNER at the server boundary before all trajectory list/detail/stats reads (while preserving trusted-local/static API-token owner behavior), and add real-server coverage showing USER/machine sessions get `403` while an OWNER can read the intended payload. The client `RoleGate` should remain defense in depth, not the only role check.

Client / agent tooling: codex
— [contributor]

lalalune:

Retain the developer-viewer work as a substantive debugging surface, with integration coordinated through #31330 rather than a second independent release of overlapping changes. The inspected patch renders/copies complete recorded content, scopes trajectory reads to the requested room and adds an owner-specific HTTP authorization boundary. Keep missing data distinct from recorded empty output, and keep provider request counts, model attempts, cache usage, background work and reply-recovery runs separately attributable. Preserve the draft hold for current live/visual acceptance and verify all raw-trajectory transports with actual owner/nonowner credentials, including shared gateway and registered-token scope behavior; this is not a final privacy/security approval. Before closing this PR as superseded, prove which current viewer commits and acceptance cases are present in the combined candidate, including normal-tab navigation without navigating the inspector. Do not replace complete underlying model evidence with preview text to improve performance. Updated-head review at a94c7e465a84a29284fccb841f19809790648413 also inspected the relay baseline repair, separation of new-send live activity and provider-return sizes from actual model input. The latest body checkpoint still names 4a907f8; restamp final evidence to the current head. The settings preflight fallback must retain observable save errors and the actual post-save readiness transition.

OutstandingRetard (approved):
## Review: fix(dev): correct live timing and keep trajectory context us

**Scope:** 6 files changed.
**Base:** 
**Author:** NubsCarson

### Analysis

This PR addresses: The developer inspector can show a previous prompt's age while a new submission is still being relay

Files modified: README.md, trajectory-code-block.tsx, DeveloperWorkspace.test.tsx, DeveloperWorkspace.tsx, TrajectoryDetailView.tsx

### Verdict: APPROVE

The change is well-scoped and addresses a clear issue. Tests pass and the implementation follows existing patterns.

lalalune:
Reviewed the remaining changes against develop `d286e5b50c84f0762309012c9ab03f50255ab613`. This PR is already incorporated through the context integration (`a74a937b743a3fe605e0e507ce7b963876fb6952`, subsequently integrated by #31504).

The current source contains the submission-time clock and prior-turn exclusion, its regression test, complete-text copy controls, provider-size attribution wording, and bounded provider-text/select layout. The only merge conflict is the older DeveloperWorkspace layout versus the current reader-mode wrapper. Retaining the current file resolves it without dropping any original behavior. The resulting merge tree is `33560362f29b358a65c368bf9c9592842760ff94`, byte-for-byte equal to the current develop tree; all six PR paths produce no net change.

Closing as already incorporated and preserving the author branch. This disposition establishes source equivalence; it does not claim a new visual audit or re-certify the historical artifact claims.

PR #31016, opened 2026-09-10
test(calendar): validate optional arguments across provider strictness modes

### LINKED ISSUE #31013: test(personal-assistant): validate sparse calendar calls under current provider strictness
The personal-assistant calendar boundary test still requires every normalized tool to be non-strict. Current develop intentionally opts planner tools into Cerebras strict mode while preserving optional properties (PR #30990). The full package suite fails two cases at the obsolete flag assertion.

Update this existing contract test to validate sparse calendar arguments against the normalized provider schema and the action validator, retaining rejection of incorrectly typed travel inputs. Verify both calendar owners and both provider modes; preserve the other calendar regressions. Run owning package checks and repository verification on the final source. This is test-contract maintenance, not a new production behavior change or a claim of live calendar-provider acceptance.

lalalune:
Closes #31013.

The personal-assistant calendar boundary test rejected the Cerebras strict-tool contract introduced by #30990 even though optional calendar fields remain optional. Replace the stale family-wide schema assertions with actual validation of sparse calendar-create arguments through the normalized provider schema and the original action validator. Both calendar owners and both provider modes accept omitted or explicitly supplied travel origins and reject an incorrectly typed travel origin. Existing title-delete and travel-sentinel regressions remain.

Current-develop control reproduces the two obsolete strictness failures. The updated focused file passes all eight tests; an independent reviewer approved the changed consumer contract. No production code changes, live provider calls, or calendar writes are included.

Published head: `3009bfbc19c6f6f09d4752919ac404b8d68eeeb5`. Its complete personal-assistant suite passed 2,601 tests in 300 files, with six skips; all five hosted checks passed in run34501537816.

Current integration: develop `f7221e0548d723cc714b07a239f2f5ed85105442` plus this published head produces tree `fcccac5cd7d68c7fb9424ad393cbfd4787d97ae1`, identical to locally qualified candidate `587068d8f33501d1385d06839d6a9c03d3acbd74`. Normal installation and all 373 repository tasks plus final audits passed on that tree using Node24.15.0 and Bun1.3.14. The base advance changes only seven standalone LifeOps script/test files; the personal-assistant package and its tested provider/runtime source are unchanged. The full package suite was not repeated for that unrelated script change.

The provider-wire regression covered here is sparse calendar creation, not every calendar operation. Both a delegated reviewer and an external reviewer approved this targeted contract; the external reviewer also demonstrated that disabling optional-field preservation makes the new validator assertions fail.

N/A - only a backend test changes; no rendered surface changes.

N/A - only a backend test changes; no rendered surface changes.

N/A - deterministic provider-schema and action-validator behavior is exercised in the test runner.

Current-head repository verification (exit0):
```text
  Time:    15m51.471s 
[anti-larp] scanned 11676 test files — 0 focused, 0 orphaned skip(s)
```

N/A - no frontend runtime is involved.

N/A - no planner or production behavior changes; this test calls the real provider schema normalizer and argument validators without model dispatch.

The previous source produced6pass/2fail because the two provider-normalization cases expected strict:false for every tool. The replacement8cases pass. For each calendar owner and provider mode, both the normalized JSONschema validator and the original action validator accept sparse create arguments and a supplied string travelOriginAddress, and reject a boolean travelOriginAddress. Existing delete-by-title and travel-sentinel consumer regressions remain covered.
The focused candidate result below predates the current-base rebase; the changed test source is unchanged. The complete package suite at the published head also passed, as shown below.
```text
$ node scripts/lint-default-packs.mjs
[lint-default-packs] clean — 0 findings across default packs.
$ vitest run --config vitest.config.ts test/calendar-optional-tool-parameters.test.ts

 RUN  v4.1.10 /private/tmp/eliza-project19-review/core-worktree


 Test Files  1 passed (1)
      Tests  8 passed (8)
   Start at  11:26:39
   Duration  42.76s (transform 36.63s, setup 125ms, import 40.86s, tests 43ms, environment 0ms)

```
```text

 RUN  v4.1.10 /private/tmp/eliza-project19-review/core-worktree


 Test Files  300 passed (300)
      Tests  2601 passed | 6 skipped (2607)
   Start at  12:16:47
   Duration  1953.72s (transform 2035.76s, setup 21.14s, import 837.94s, tests 1052.26s, environment 8.37s)

```

N/A - no rendered visual surface changes.

### DIFF
--- a/plugins/plugin-personal-assistant/test/calendar-optional-tool-parameters.test.ts
+++ b/plugins/plugin-personal-assistant/test/calendar-optional-tool-parameters.test.ts
@@ -7,7 +7,10 @@ import type { Action, ActionParameterSchema } from "@elizaos/core";
 import { describe, expect, it } from "vitest";
 import { promoteSubactionsToActions } from "../../../packages/core/src/actions/promote-subactions.js";
 import { buildPlannerToolsFromActions } from "../../../packages/core/src/actions/to-tool.js";
-import { validateToolArgs } from "../../../packages/core/src/actions/validate-tool-args.js";
+import {
+  validateSchema,
+  validateToolArgs,
+} from "../../../packages/core/src/actions/validate-tool-args.js";
 import { createCalendarActionRunner } from "../../plugin-calendar/src/actions/calendar-handler.js";
 import { __INTERNAL_normalizeNativeToolsForCall as normalizeNativeToolsForCall } from "../../plugin-openai/models/text.js";
 import { calendarAction } from "../src/actions/calendar.js";
@@ -24,32 +27,54 @@ describe.each([
   ["domain calendar", domainCalendarAction],
 ] as const)("%s optional native arguments", (_name, parent) => {
   it.each([false, true])(
-    "keeps optional detail fields optional after provider normalization (Cerebras: %s)",
+    "admits sparse calendar creates through the provider schema without weakening travel types (Cerebras: %s)",
     (cerebrasMode) => {
       const family = promoteSubactionsToActions(parent);
-      expect(family.length).toBeGreaterThan(1);
+      const create = family.find(
+        (action) => action.name === "CALENDAR_CREATE_EVENT",
+      );
+      if (!create)
+        throw new Error("calendar create operation was not promoted");
       const normalized = normalizeNativeToolsForCall(
         buildPlannerToolsFromActions(family),
         { cerebrasMode },
       ).tools;
       if (!normalized)
         throw new Error("calendar tool family was not normalized");
-      for (const action of family) {
-        const tool = normalized[action.name] as {
-          strict?: boolean;
-          inputSchema: { jsonSchema: ActionParameterSchema };
-        };
-        expect(tool.strict).toBe(false);
-        const details = tool.inputSchema.jsonSchema.properties?.details;
-        if (
-          action.parameters?.some((parameter) => parameter.name === "details")
-        ) {
-          expect(details?.properties?.travelOriginAddress).toMatchObject({
-            type: "string",
-          });
-          expect(details?.required ?? []).toEqual([]);
-        }
+      const tool = normalized[create.name] as {
+        strict?: boolean;
+        inputSchema: { jsonSchema: ActionParameterSchema };
+      };
+      expect(tool.strict).toBe(cerebrasMode);
+      const args = {
+        title: "Pottery class",
+        details: {
+          startAt: "2026-09-11T16:00:00.000Z",
+          endAt: "2026-09-11T17:00:00.000Z",
+          timeZone: "UTC",
+        },
+      };
+      for (const details of [
+        args.details,
+        { ...args.details, travelOriginAddress: "Studio" },
+      ]) {
+        const input = { ...args, details };
+        const errors: string[] = [];
+        validateSchema(tool.inputSchema.jsonSchema, input, "", errors);
+        expect(errors).toEqual([]);
+        expect(validateToolArgs(create, input)).toMatchObject({
+          valid: true,
+          args: input,
+        });
       }
+      const invalid = {
+        ...args,
+        details: { ...args.details, travelOriginAddress: false },
+      };
+      const errors: string[] = [];
+      validateSchema(tool.inputSchema.jsonSchema, invalid, "", errors);
+      expect(errors.length).toBeGreaterThan(0);
+      expect(validateToolArgs(create, invalid).valid).toBe(false);
     },
   );
 


ss251 (changes requested):
3009bfbc19c6f6f09d4752919ac404b8d68eeeb5

Test-only, no product path. The single changed file is `plugins/plugin-personal-assistant/test/calendar-optional-tool-parameters.test.ts`. The PR states there are no production edits, live provider calls, or calendar writes. Unit tests are not acceptance evidence.

The only develop failure is the obsolete family loop `expect(tool.strict).toBe(false)` after #30990 (2 of 8 cases). `details.required === []` and `travelOriginAddress: {type:"string"}` were not failing. This patch drops that family lock and only inspects `CALENDAR_CREATE_EVENT`. Update/list/delete can start requiring omitted travel fields on the wire schema and this file stays green. The file header still claims the family union must not require absent travel fields.

Sparse-create accept and boolean `travelOriginAddress` reject are already asserted by the unchanged third test in the same file. `expect(tool.strict).toBe(cerebrasMode)` reads request-wide `cerebrasRequestStrict` (`plugins/plugin-openai/models/text.ts` 842-848, 961-962), not a per-create fact.

Draft; body says do not merge. It claims an independent reviewer approved this consumer contract; reviews on this head are empty. The 8-pass vitest log is stated to predate rebase onto `4985223ea31c01d6eec22c5d6aea7d79913ef896`. No author model attribution on the PR.

Recommendation: fix
Keep the family-wide optional `details` assertions; replace only the obsolete `strict:false` expectation; un-draft after current-base proof. Do not merge this head.

Verified by reading the diff at this head against develop and base `4985223ea31c01d6eec22c5d6aea7d79913ef896`. Live execution blocked: no disposable sandbox on this host, so this is a static review of the diff at the exact head.

Client / agent tooling: grok
— [grok-fleet-ss251]

attentionhead (approved):
**APPROVE** — verified at `3009bfbc19c6f6f09d4752919ac404b8d68eeeb5` (draft, `MERGEABLE`/`UNSTABLE`, +43/−18 in one file).

This is a test-only change, so I judged it on two questions: does it fix something that is actually red today, and does the new pin catch a regression the old one misses. Both are yes, and I reproduced each.

### 1. The assertion this replaces is already failing on `develop`

`develop`'s own copy of `plugins/plugin-personal-assistant/test/calendar-optional-tool-parameters.test.ts` is red at `origin/develop` (`d7df7af187`), unmodified:

```
FAIL … personal-assistant calendar optional native arguments >
     keeps optional detail fields optional after provider normalization (Cerebras: true)
FAIL … domain calendar optional native arguments > (Cerebras: true)
  calendar-optional-tool-parameters.test.ts:42  expect(tool.strict).toBe(false)
Tests  2 failed | 6 passed (8)
```

The unconditional `expect(tool.strict).toBe(false)` no longer matches the normalizer. `nativeToolStrictness` (`plugins/plugin-openai/models/text.ts:745-762`) upgrades a declared-`false` tool back to `true` when `cerebrasMode` is set and the tool is `strictWithOptionalProperties`-compatible:

```ts
return declared === false && cerebrasMode && optionalCompatible ? true : declared;
```

`expect(tool.strict).toBe(cerebrasMode)` is the correct pin for that, and it is a real request-wide provider contract, not an implementation literal. At the PR head the file is green: **Tests 8 passed (8)**.

### 2. Bidirectional control — the new pin catches a regression the old one is blind to

The load-bearing guard for the behavior the test claims to protect is `preserveOptional: options.cerebrasMode` at `plugins/plugin-openai/models/text.ts:906`, which is what keeps nested optional `details.*` properties out of the strict schema's `required` list.

I mutated exactly that one line to `preserveOptional: false` and ran both versions of the test file against it, everything else at the same tree:

**PR's test file — RED, and for the right reason:**

```
calendar-optional-tool-parameters.test.ts:64  expect(errors).toEqual([])
+   "Missing required argument 'details.recurrence_rule'",
+   "Missing required argument 'details.repeat'",
+   "Missing required argument 'details.repeats'",
+   "Missing required argument 'details.repeatrule'",
+   "Missing required argument 'details.repeat_rule'",
+   "Missing required argument 'details.queries'",
+   "Missing required argument 'details.attendees'",
Tests  2 failed | 6 passed (8)
```

A sparse, entirely legitimate create — `title` plus `startAt`/`endAt`/`timeZone` — stops validating.

**`develop`'s test file — byte-identically red with and without the mutation:** both runs fail the same two cases at the same `test.ts:42`, the pre-existing `strict` assertion. It never reaches its own `expect(details?.required ?? []).toEqual([])`, so it does not observe the regression at all. (For `cerebrasMode: false` the mutation is a no-op, since `preserveOptional` was already `false` there — the `true` branch is the only one that can catch it, and that is precisely the branch the stale assertion short-circuits.)

So the replacement is not a relabeled equivalent: it restores detection that is currently dead.

### 3. Substance of the new assertions

The test drives the real pipeline end to end — `promoteSubactionsToActions` → `buildPlannerToolsFromActions` → `normalizeNativeToolsForCall` → `validateSchema`/`validateToolArgs` — with no stub standing in for the system under test. `validateSchema` is a genuine exported entry point, not test-only reach-in; its module header documents it as "exported for verifying whole structured outputs (e.g. remote-model planner JSON)".

Both halves are asserted: the sparse payload and the `travelOriginAddress: "Studio"` payload must pass, and `travelOriginAddress: false` must produce errors through both `validateSchema` and `validateToolArgs`. That negative case is what stops the pin from being satisfiable by simply widening the type — the failure mode the previous `toMatchObject({ type: "string" })` check was reaching for, now enforced by exercising the validator rather than by reading the schema back.

Narrowing from "every action in the family" to the specific `CALENDAR_CREATE_EVENT` operation, with an explicit throw when it is not promoted, is a fair trade: the old loop's per-action work was guarded by an `if` that made it a no-op for most of the family anyway.

### Notes, non-blocking

- The consumer-visible regression this prevents is worth stating in the PR body if it stays as-is: under the Cerebras strict path, a model producing a minimal calendar create would have its arguments rejected for missing optional detail fields it was never asked to supply.
- Nothing here touches `plugins/plugin-calendar`'s 13 test files that fail to load on `develop` with `Cannot find package '@elizaos/login'`; that cascade is pre-existing and unrelated to this PR.

**Reproduction:** `origin/develop` at `d7df7af187`, PR test file checked out over it, run from the repo root with `npx vitest run --config plugins/plugin-personal-assistant/vitest.config.ts plugins/plugin-personal-assistant/test/calendar-optional-tool-parameters.test.ts`. Mutation applied and reverted at `plugins/plugin-openai/models/text.ts:906`.

lalalune:
Reviewed both assessments. The replacement deliberately targets the sparse calendar-create contract described in #31013: normalization must accept a valid create without optional travel fields, retain valid supplied travel fields, and reject invalid travel types. Its normalized provider-schema checks are distinct from the unchanged raw action-validator and travel-intent tests.

The narrower scope is real: this test does not claim provider-wire coverage for every update/list/delete operation. Retaining the old family loop with only its strictness expectation changed would retain schema-descriptor assertions; the repository's test-quality policy instead calls for exercising the parser or consumer. No failing update/list/delete input or production regression has been demonstrated here. The independent preservation-disabled control in the approving review supports the concrete create regression this test protects.

The independent review mentioned in the draft was a delegated source review, not a claim that a GitHub approval already existed. The complete personal-assistant suite has now passed at published head `3009bfbc19c6f6f09d4752919ac404b8d68eeeb5`: 300 files, 2,601 tests passed, six skipped, exit 0. All five hosted checks also passed for that head. The draft remains held while verification finishes on the exact merge tree with the subsequent scripts-only develop change. Provider/model disclosure is optional under the current CONTRIBUTING.md.

PR #31003, opened 2026-09-10
fix(calendar): require reviewed sync destinations and recover uncertain writes

lalalune:
Connecting another Google account could change the implicit destination for new built-in calendar links. This change requires an owner-reviewed writable account and calendar, with synchronization paused by default. Durable revision checks and mutation receipts reject stale reviews and prevent destination replacement while a provider outcome is unknown. Rejected remaps preserve existing event identities.

The connections view and registered personal-assistant calendar routes support destination selection, pause/resume, and pending-dispatch recovery. Recovery reads the provider before settling an uncertain receipt. Failures before any external mutation release the receipt; uncertain writes retain it and block replay.

Live school-calendar testing exposed loss of all-day semantics, rejected timed-to-all-day patches, and repeated writes caused by provider timezone normalization. Linked snapshots preserve all-day state in both directions. Google updates explicitly clear incompatible date/dateTime fields. All-day semantic comparison ignores timezone metadata and avoids provider writes when the local and remote event already agree, including older checkpoints.

The production connections view also failed to load because its shared host importer left a bare package specifier in the browser. A literal dynamic import now lets the bundler provide `@elizaos/shared`. The built frontend opens the real pilot connections view, refreshes sync status, and exposes replacement-account controls without browser errors.

Validation at `72e2be875d5707442dc87029988bb1a683e9ce9c`, incorporating fetched develop `d7df7af187f05ec744b57c01799cfbabcc5fd325`:

- Pinned install, root verification, UI typecheck and production web build passed.
- DynamicViewLoader:44 tests passed. App audit:222 capture tests passed;216 findings with zero computed broken/needs-work. OCR:203 verified,0 broken,13 requiring inspection. Family captures inspected across four viewports; phone-landscape upload-control clearance still needs live scrolling acceptance. The audit is not a blanket visual pass.
- Full UI suite:14006 passed,49 failed,7 skipped,2 unhandled errors. Controlled serial runs of the22 failing files on the original and fixed loaders produced identical results:29 failures,732 passes,2 unhandled errors, with identical failed-test names. These existing failures remain open.
- Calendar:874 passed/4 skipped at preceding7fa812, whose calendar source is unchanged here. Google:407 passed at prior75da45. Personal-assistant unit2607 passed/6 skipped and integration426 passed at prioraad00. Those suites are not claimed rerun at this head.
[...]

Issue #30123, opened 2026-08-31
feat(lifeops): deliver Bettina owner-private family operations MVP


lalalune:
## Problem

The Bettina pilot needs one owner-private family operations journey that is reproducible from GitHub and verifiable through the deployed product. The current VPS runs unpublished local commits and has no complete acceptance path for agreement knowledge, bounded guest access, linked Google calendar sync, school-calendar automation, coordination packets, or approved iMessage delivery.

## Scope

- Stabilize remote pairing with instance-bound, machine-readable failures and public-origin acceptance.
- Store immutable parenting-agreement artifacts with reviewed page-cited obligations.
- Keep owner-only access by default; make pins independent of bounded verified guest grants.
- Preserve the built-in Eliza Calendar and add durable two-way linked Google events with conflict, replay, quarantine, and disconnect semantics.
- Add the guarded monthly Concord school-calendar PDF discovery/hash/diff/approval workflow.
- Add a versioned monthly coordination packet and neutral approval-gated external draft; exclude expenses.
- Add deterministic daily calendar cards and secure approved iMessage/Blooio delivery.
- Consolidate the pilot onto the reviewed Eliza runtime and deploy an immutable merged SHA to Bettina VPS.

## Acceptance

- Every PDD requirement has an automated or explicit live gate in a traceability matrix.
- Relevant package tests, PGlite restart/concurrency/adversarial tests, app E2E, five visual-audit cycles, root verify/test, and canonical evidence review pass at the exact PR SHA.
- Fresh-browser public pairing works and negative/replay cases fail distinctly.
- Agreement review, page citations, pins, grants, expiry, revocation, export, and deletion dependencies work through UI and APIs.
- Eliza Calendar works without Google; `shawgotbags@gmail.com` proves bidirectional create/update/delete/conflict/idempotency on a test calendar.
- The current Concord district page resolves the PDF; first import is approved, identical rerun is a recorded no-op, and changed revisions only update workflow-owned events.
- Approved iMessage/Blooio delivery is proven on the real configured edge with unauthorized sender, signature, replay, and tamper rejection.
- The deployed process reports the merged SHA, passes restart persistence, and has no auth, credential, embedding, migration, duplicate-send, or unclassified runtime errors.

## Related work

- #17176 calendar source completeness
- #17179 co-parent scenario corpus
- #17181 school calendar ingestion
- #23103 document entitlements and pins
- #22787 iMessage/Blooio readiness

## Evidence policy

[...]

PR #31029, opened 2026-09-10
fix(app): restore calendar controls and Connections loading

### LINKED ISSUE #30222: fix(calendar): restore real responsive day/week/month views and safe source labels
## Problem
The owner-facing Calendar release surface can render as an oversized agenda toolbar instead of a recognizable calendar: desktop and narrow views show full-width New and Day/Week/Month controls with no time/month grid. The live owner feed also renders an invalid source label (`undefined · Eliza Calendar`) when the local Eliza source arrives without the assumed provider discriminator.

Current shipped entry points are inconsistent: `CalendarPage` uses the month-only `SimpleCalendarView`, while the existing `CalendarSection` owns the Google Calendar-style day/week/month projections. On compact layout, `CalendarSection` replaces every selected mode with agenda, so the selected mode and visible projection disagree.

The isolated visual harness also fails before capture because its bare `@elizaos/ui` string alias swallows newly used UI subpaths, preventing exact-develop visual review.

## Acceptance criteria
- Calendar's shipped local and remote bundle entry renders recognizable Day, Week, and Month projections from the canonical feed.
- Day/Week show a time grid; Month shows a seven-column date grid; compact layouts preserve the selected projection instead of silently showing agenda.
- Empty calendars keep the grid visible with a restrained, actionable empty state; controls remain usable and do not span the viewport unexpectedly.
- Local Eliza source labels never render `undefined`, even for live-like sparse source payloads.
- Desktop 1280x900 and mobile 390x844 visual evidence covers empty and populated states with zero browser errors and responsive assertions.
- Calendar component tests, typecheck, build, and app visual audit pass; five documented audit/inspection iterations are completed.

### LINKED ISSUE #31007: ui-smoke: calendar specs wait for lifeops-calendar-section, but the registered calendar view bundle mounts SimpleCalendarView
## Summary

The calendar cases of the deterministic ui-smoke lane wait for `data-testid="lifeops-calendar-section"`, but the calendar view bundle the app serves no longer contains that element. Three Playwright cases fail on a fresh view build at `develop` `d7df7af187f0`.

## Where the mismatch is

- `plugins/plugin-calendar/src/components/calendar/calendar-view-bundle.ts` exports `CalendarPage` as the registered `CalendarView`; `CalendarPage` mounts `SimpleCalendarView` (the `aria-label="Calendar month"` surface). It has done so since #29151 (2026-08-26).
- `data-testid="lifeops-calendar-section"` exists only in `plugins/plugin-calendar/src/components/CalendarSection.tsx`, which no view bundle includes: after `packages/scripts/build-views.mjs` builds all 22 view targets, `grep -c lifeops-calendar-section plugins/*/dist/views/bundle.js` is 0 for every bundle, and `CalendarSection` has no runtime consumer outside tests, stories, and the view-screenshot stubs.
- The specs still expect it:
  - `packages/app/test/ui-smoke/apps-personal-assistant-decomposed-interactions.spec.ts` `openPopulatedCalendar` (`lifeops-calendar-section`, then `lifeops-calendar-new-event`)
  - `packages/app/test/ui-smoke/plugin-view-agent-bridge-inventory-extended.spec.ts` Calendar entry `ready: { testId: "lifeops-calendar-section" }`

## Reproduction

```bash
cd packages/app
ELIZA_UI_SMOKE_DISABLE_VIDEO=1 node scripts/run-ui-playwright.mjs --config playwright.ui-smoke.config.ts --project=chromium \
  test/ui-smoke/apps-personal-assistant-decomposed-interactions.spec.ts \
  test/ui-smoke/plugin-view-agent-bridge-inventory-extended.spec.ts
```

Result on a fresh build (Node 24.15.0, chromium): the three calendar cases fail after their 60 s wait with `waiting for getByTestId('lifeops-calendar-section')`, the other 15 cases pass. The captured accessibility tree at the failure shows the simple view rendered and populated (`main "Calendar. 2 events"`, `region "Calendar month"`, `button "Sunday, August 30, 2026. 1 event"` …), so the route, the stub endpoints, and the bundle all work; only the anchor the specs wait for is absent.

Failing cases:

- `calendar decomposed view: responsive modes and event creation`
- `calendar mobile layout keeps navigation and editor inside 390px viewport`
- `Calendar exposes chat/voice-drivable controls through the agent bridge`

## Why it is not caught

`ci.yml`'s deterministic e2e shards only run when the change filter marks the lane affected, and no hosted run at the current `develop` head has exercised these specs. The same spec files are edited by #30773 (other cases in them pass), which is how this surfaced: the calendar expectations there are unchanged from `develop`, so the defect predates that PR.

## Fix options

Either point the two specs at the surface the bundle actually mounts (`SimpleCalendarView` controls and the `Calendar month` region, plus its own new-event control), or make the registered `CalendarView` mount `CalendarSection` again if that is the intended canonical calendar. The spec comment calls `CalendarSection` "the canonical calendar", so the second reading may be the intended one; the owning team should decide.

### LINKED ISSUE #31010: fix(ui): resolve shared host modules in production plugin views
The built and deployed family pilot at7fa812daab38415dc84090fe6486bfb3d8bf82e8 cannot open `/lifeops/connections` in a fresh browser. It renders the view error boundary; browser diagnostics report `Failed to resolve module specifier '@elizaos/shared'` from dynamic-view.load.

The host advertises this framework module, but its importer delegates to a variable dynamic import marked vite-ignore. That leaves a bare package specifier for the production browser instead of a bundled host module. The connections plugin consumes the shared contract through its standard external bundle.

Acceptance: resolve the shared host module through the production bundler; build the app; open the real connections view without the unresolved-specifier error; exercise the destination/recovery controls and retain desktop/mobile evidence. Protect the actual built-browser loading contract, rather than testing only Node resolution or mocked importers. Preserve authenticated bundle transport, host singleton ownership, and origin enforcement.

Related: #30123, #31003. Found during live family MVP acceptance; not a speculative coverage issue.

lalalune:
The calendar route rendered a month-only picker, leaving the existing day/week grids and event editor inaccessible. It now uses the shared calendar component on desktop and mobile, keeps empty grids usable, supports selected-day agendas, and shows source failures with Retry. The host scans the calendar’s styles, and the reviewed layout preserves orange controls, readable grid lines and usable date targets.

Connections failed in the production browser because the host imported `@elizaos/shared` through an ignored variable specifier. A literal host import now lets the production bundler resolve that shared module. The isolated repair preserves the original commit authorship from #31003 without including its unfinished Google cutover work.

Calendar defaults use the required orange palette consistently across the accent family. An explicit user preference inherits the host theme instead of being overridden locally. The built-browser regression fails against the original unconditional override and passes with this fix.

Closes #31010. Closes #30222. Closes #31007. Related: #30123, #31003.

Candidate: `0d17f1b1a7d4f17af372306d8911148e7206ff71`, based on develop `1708731834282d31fe6a32a77548eb0f2d04bf70`.

- Complete root verification passed: all 373 tasks and final audits, including 11,679 test files checked for focused tests and untracked skips.
- Production renderer build and chunk-safety checks passed. The host source and recorded renderer hashes are unchanged from that build; the calendar bundle was rebuilt after the accent fix. Eight built-browser acceptance cases passed on this candidate.
- The Connections regression uses the real built personal-assistant bundle and canonical server rewriting through the remote-view registry. Both desktop and mobile fail with the original host importer and pass with the repaired importer; account-review and forced-sync controls work with no browser errors.
- Complete shared UI suite passed at `97fb3b4`: 1,307 files, 14,053 tests, seven skipped. UI production sources are unchanged; develop subsequently added two voice test changes. The current candidate received the fresh root gate and eight browser acceptance cases. Complete calendar suite passed on this candidate: 88 files passed, two skipped; 840 tests passed, four skipped.
- Desktop/mobile rest and hover captures were inspected across day/week/month, empty, partial and unavailable states. Interaction checks proved selected-day details, today hover and mobile week scrolling. Build hashes and console/network reports identify the served artifacts.
- Final candidate app audit: all 222 checks passed; zero broken/needs-work findings. Packaged OCR: 203 verified, zero broken, 13 requiring inspection. The affected calendar captures were inspected at all four audit viewports, alongside the detailed desktop/mobile captures. All 654 artifacts in `20260910-185215-0d17f1b-cpu` passed integrity verification. The generated dashboard reviewed all 654 artifacts; OCR completed on all 216 screenshots with zero engine failures.

The data endpoints in these UI tests use deterministic fixtures. This PR does not claim real Google account replacement, external messaging delivery, or a merged VPS deployment; those remain in #30123/#31003.

| Surface | Before | After |
| --- | --- | --- |
| Calendar desktop | Before | After |
| Calendar mobile | Before | After |
| Connections desktop | Before | After |
| Connections mobile | Before | After |

[Desktop calendar walkthrough](https://github.com/elizaOS/eliza/releases/download/pr-evidence-4/31029-0d17-calendar-week-1280-populated.mp4) · [Mobile calendar walkthrough](https://github.com/elizaOS/eliza/releases/download/pr-evidence-4/31029-0d17-calendar-week-390-populated.mp4) · [Desktop Connections walkthrough](https://github.com/elizaOS/eliza/releases/download/pr-evidence-4/31029-0d17-connections-after-1280.mp4) · [Mobile Connections walkthrough](https://github.com/elizaOS/eliza/releases/download/pr-evidence-4/31029-0d17-connections-after-390.mp4)

- [x] [Before screenshots](https://github.com/elizaOS/eliza/releases/download/pr-evidence-4/31029-0d17-connections-before-390.jpg) — desktop/mobile calendar and Connections pairs are inline above.

- [x] [After screenshots](https://github.com/elizaOS/eliza/releases/download/pr-evidence-4/31029-0d17-calendar-month-390-populated-rest.jpg) — all affected rest/hover states were inspected.

- [x] https://github.com/elizaOS/eliza/releases/download/pr-evidence-4/31029-0d17-connections-after-390.mp4

- [x] N/A - No backend domain behavior changes. Data endpoints use deterministic fixtures; the browser exercises the real built host and dynamic view-bundle transport.

- [x] [Console and network evidence](https://github.com/elizaOS/eliza/releases/download/pr-evidence-4/31029-0d17-frontend-evidence.json) — expected negative-control errors are identified separately from successful current captures.

- [x] N/A - No model, prompt, provider or action behavior changes. The agent-bridge case exercises existing controls without a model call.

- [x] N/A - No production provider or database mutations are exercised by this UI repair.

- [x] [31029-0d17-ocr-triage.json](https://github.com/elizaOS/eliza/releases/download/pr-evidence-4/31029-0d17-ocr-triage.json)

[Complete reviewed evidence archive](https://github.com/elizaOS/eliza/releases/download/pr-evidence-4/31029-bettina-calendar-0d17-reviewed-evidence.zip) includes all desktop/mobile rest/hover captures, MP4 recordings, browser diagnostics, build provenance, package/root logs, manual review and the full verified canonical bundle.

[Generated OCR review dashboard](https://github.com/elizaOS/eliza/releases/download/pr-evidence-4/31029-bettina-calendar-0d17-review-dashboard.zip): extract and open `index.html`; all 654 artifacts and 216 OCR text results are included.

### DIFF
--- a/packages/app/test/ui-smoke/apps-personal-assistant-decomposed-interactions.spec.ts
+++ b/packages/app/test/ui-smoke/apps-personal-assistant-decomposed-interactions.spec.ts
@@ -406,6 +406,17 @@ for (const width of [1280, 390]) {
         await route.fulfill({ json: { accounts: [] } });
       },
     );
+    await page.route("**/api/permissions/calendar", async (route) => {
+      await route.fulfill({
+        json: {
+          id: "calendar",
+          status: "denied",
+          lastChecked: Date.now(),
+          canRequest: false,
+          platform: "darwin",
+        },
+      });
+    });
     await openAppPath(page, "/lifeops/connections");
     const refresh = page.getByRole("button", {
       name: "Retry all connection checks and synchronization",
--- a/packages/app/test/ui-smoke/apps-personal-assistant-decomposed-interactions.spec.ts
+++ b/packages/app/test/ui-smoke/apps-personal-assistant-decomposed-interactions.spec.ts
@@ -56,10 +56,31 @@ test("calendar decomposed view: responsive modes and event creation", async ({
 }) => {
   await openPopulatedCalendar(page);
 
+  if ((await page.evaluate(() => window.innerWidth)) >= 768) {
+    const todayBounds = await page
+      .getByRole("button", { name: "Today", exact: true })
+      .boundingBox();
+    const createBounds = await page
+      .getByTestId("lifeops-calendar-new-event")
+      .boundingBox();
+    if (!todayBounds || !createBounds)
+      throw new Error("Calendar toolbar controls are missing");
+    expect(
+      Math.abs(
+        todayBounds.y +
+          todayBounds.height / 2 -
+          createBounds.y -
+          createBounds.height / 2,
+      ),
+      "Desktop calendar navigation and creation should share one compact toolbar row",
+    ).toBeLessThan(4);
+  }
+
   const monthMode = page.getByRole("button", { name: "Month", exact: true });
   await expectTopmostAtCenter(monthMode, "Calendar Month mode");
   await monthMode.click();
   await expect(monthMode).toHaveAttribute("aria-pressed", "true");
+  await expect(page.getByTestId("calendar-month-grid")).toBeVisible();
 
   const newEvent = page.getByTestId("lifeops-calendar-new-event");
   await expectTopmostAtCenter(newEvent, "Calendar New event");
@@ -82,6 +103,7 @@ test("calendar mobile layout keeps navigation and editor inside 390px viewport",
   await expectTopmostAtCenter(dayMode, "Calendar Day mode");
   await dayMode.click();
   await expect(dayMode).toHaveAttribute("aria-pressed", "true");
+  await expect(page.getByTestId("calendar-time-grid")).toBeVisible();
 
   const newEvent = page.getByTestId("lifeops-calendar-new-event");
   await expectTopmostAtCenter(newEvent, "Calendar New event");
@@ -110,6 +132,50 @@ test("calendar mobile layout keeps navigation and editor inside 390px viewport",
     page.getByRole("button", { name: "Create event" }),
   ).toBeVisible();
 });
+
+for (const width of [1280, 390]) {
+  test(`empty calendar preserves selected projections and creation at ${width}px`, async ({
+    page,
+  }) => {
+    await page.setViewportSize({ width, height: 900 });
+    await page.route("**/api/lifeops/calendar/feed**", async (route) => {
+      const url = new URL(route.request().url());
+      await route.fulfill({
+        status: 200,
+        contentType: "application/json",
+        body: JSON.stringify({
+          calendarId: "primary",
+          events: [],
+          source: "cache",
+          state: "complete",
+          sources: [],
+          timeMin: url.searchParams.get("timeMin"),
+          timeMax: url.searchParams.get("timeMax"),
+          syncedAt: new Date().toISOString(),
+        }),
+      });
+    });
+    await openAppPath(page, "/calendar");
+    for (const mode of ["Day", "Week", "Month"]) {
+      const control = page.getByRole("button", { name: mode, exact: true });
+      await control.click();
+      await expect(control).toHaveAttribute("aria-pressed", "true");
+      await expect(
+        page.getByTestId(
+          mode === "Month" ? "calendar-month-grid" : "calendar-time-grid",
+        ),
+      ).toBeVisible();
+      expect(
+        await page.evaluate(() => document.documentElement.scrollWidth),
+      ).toBe(width);
+    }
+    await page.getByTestId("lifeops-calendar-new-event").click();
+    await expect(page.getByLabel("Event title")).toBeInViewport();
+    await expect(
+      page.getByRole("button", { name: "Create event" }),
+    ).toBeVisible();
+  });
+}
 
 test("inbox decomposed view: channel filters toggle", async ({ page }) => {
   // /inbox renders the populated triage list from the inbox mock: an Email
--- a/packages/app/test/ui-smoke/apps-personal-assistant-decomposed-interactions.spec.ts
+++ b/packages/app/test/ui-smoke/apps-personal-assistant-decomposed-interactions.spec.ts
@@ -392,3 +392,41 @@ test("relationships decomposed view: renders the graph and toggles a kind filter
     timeout: 15_000,
   });
 });
+
+for (const width of [1280, 390]) {
+  test(`connections calendar recovery loads through the built host at ${width}px`, async ({
+    page,
+  }) => {
+    await page.setViewportSize({ width, height: 900 });
+    const pageErrors: string[] = [];
+    page.on("pageerror", (error) => pageErrors.push(error.message));
+    await page.route(
+      "**/api/lifeops/connectors/google/status**",
+      async (route) => {
+        await route.fulfill({ json: { accounts: [] } });
+      },
+    );
+    await openAppPath(page, "/lifeops/connections");
+    const refresh = page.getByRole("button", {
+      name: "Retry all connection checks and synchronization",
+    });
+    await expect(refresh).toBeEnabled({ timeout: 60_000 });
+    await page
+      .getByRole("button", { name: "Replace an account", exact: true })
+      .click();
+    await expect(
+      page.getByLabel("Test account to disconnect", { exact: true }),
+    ).toBeVisible();
+    const synchronized = page.waitForRequest((request) => {
+      const url = new URL(request.url());
+      return (
+        url.pathname === "/api/lifeops/calendar/feed" &&
+        url.searchParams.get("forceSync") === "true"
+      );
+    });
+    await refresh.click();
+    await synchronized;
+    await expect(refresh).toBeEnabled();
+    expect(pageErrors).toEqual([]);
+  });
+}
--- a/packages/app/test/ui-smoke/apps-personal-assistant-decomposed-interactions.spec.ts
+++ b/packages/app/test/ui-smoke/apps-personal-assistant-decomposed-interactions.spec.ts
@@ -52,6 +52,56 @@ async function openPopulatedCalendar(page: Page): Promise<void> {
   });
 }
 
+test("calendar inherits the host accent for current and selected days after a preference change", async ({
+  page,
+}) => {
+  await page.addInitScript(() => {
+    localStorage.setItem("eliza:ui-accent", "green");
+  });
+  await openPopulatedCalendar(page);
+  await page.getByRole("button", { name: "Month", exact: true }).click();
+  const grid = page.getByTestId("calendar-month-grid");
+  const selected = grid
+    .locator(
+      'button[data-agent-id^="calendar-day-"]:not([aria-current="date"])',
+    )
+    .first();
+  await selected.click();
+  await expect(selected).toHaveAttribute("aria-pressed", "true");
+  await page.mouse.move(0, 0);
+  await page.waitForTimeout(300);
+  const colors = await page.evaluate(() => {
+    const root = getComputedStyle(document.documentElement);
+    const current = document.querySelector(
+      '[data-testid="calendar-month-grid"] button[aria-current="date"]',
+    );
+    const selection = document.querySelector(
+      '[data-testid="calendar-month-grid"] button[aria-pressed="true"]',
+    );
+    if (!current || !selection)
+      throw new Error("Calendar day controls missing");
+    const probe = document.createElement("span");
+    document.body.append(probe);
+    const resolve = (value: string) => {
+      probe.style.backgroundColor = value;
+      return getComputedStyle(probe).backgroundColor;
+    };
+    const result = {
+      current: getComputedStyle(current).backgroundColor,
+      selected: getComputedStyle(selection).backgroundColor,
+      expectedCurrent: resolve(root.getPropertyValue("--accent")),
+      expectedSelected: resolve(root.getPropertyValue("--accent-subtle")),
+      preferenceApplied:
+        document.documentElement.style.getPropertyValue("--accent") !== "",
+    };
+    probe.remove();
+    return result;
+  });
+  expect(colors.preferenceApplied).toBe(true);
+  expect(colors.current).toBe(colors.expectedCurrent);
+  expect(colors.selected).toBe(colors.expectedSelected);
+});
+
 test("calendar decomposed view: responsive modes and event creation", async ({
   page,
 }) => {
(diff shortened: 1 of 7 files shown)


attentionhead (changes requested):
**REQUEST_CHANGES** — verified at `437dc54cde57a7e4f431541f356e9bf0bd597b96` (draft, `MERGEABLE`/`UNSTABLE`, +286/−87 across 6 files).

The core of this PR is right, and I confirmed the two structural fixes before looking for anything wrong with it. The blocking issue is one hardcoded style block that undoes the benefit of the very fix this PR lands.

### What I verified as correct first

**1. The missing Tailwind `@source` is the real root cause.** `packages/ui/src/styles/styles.css` is the only app-wide `@source` list in the repository — the only other `@source` directives are `packages/homepage/src/index.css` and two per-plugin e2e fixtures. Before this PR it scanned `plugins/app-lifeops/src` and `plugins/plugin-maps/src` but not `plugins/plugin-calendar/src`, and the built `packages/ui/dist/styles/styles.css` still shows the two-entry list. So none of `CalendarSection.tsx`'s utility classes were ever generated. That is exactly the "calendar controls" symptom, and line 53 fixes it at the source.

**2. The `@elizaos/shared` loader change is correct, not a singleton break.** `importHostExternal` is `import(/* @vite-ignore */ specifier)` — a deliberately unanalyzable specifier, which is right for the `@elizaos/capacitor-*` host natives around it but leaves `@elizaos/shared` as an unresolvable bare specifier in a built host. Making it a literal `import("@elizaos/shared")` lets the bundler resolve it. `HOST_EXTERNAL_IMPORTERS` remains the single provider for view bundles that externalize it, so this yields one instance, not two.

**3. `useViewEvent` is used correctly.** `packages/ui/src/hooks/useViewEvent.ts:28` takes `(type, handler, deps: DependencyList = [])` and subscribes on `[type, ...deps]`, with the handler held in a ref. Passing `[calendar.refresh]` is the documented shape.

**4. Bidirectional control on the unit tests.** At the merged head `plugins/plugin-calendar` runs **16 passed (16)**. Restoring only `origin/develop`'s `CalendarSection.tsx` under this PR's test file leaves **2 failed | 14 passed**, and exactly the two new pins fail — `Unable to find an accessible element with the role "button" and name "Mon, Jun 15"` and `... name "Retry"` — with every pre-existing assertion still green. The new coverage is real.

Note that `keeps the time grid and event creation available on an empty calendar` passes against develop's source too, because with the `useMediaQuery` mock removed `compactLayout` is falsy there and develop already renders `TimeGrid`. It is a fine assertion to keep; it just is not one of the pins that distinguishes the change.

### Blocking: the inline accent override in `CalendarSection.tsx:1231-1238`

```tsx
style={
  {
    "--accent": "var(--eliza-brand-orange, #ff5800)",
    "--accent-foreground": "#111111",
    "--accent-muted":
      "color-mix(in srgb, var(--eliza-brand-orange, #ff5800) 75%, #111111)",
  } as CSSProperties
}
```

Three concrete problems, in order of consequence:

**`--eliza-brand-orange` does not exist anywhere in the repository.** A repo-wide grep finds exactly two references, and both are the two lines this diff adds. Nothing ever defines it, so the fallback is unconditional and this is a hardcoded `#ff5800` behind dead indirection — the shape that silently never picks up the token it appears to be waiting for.

**It overrides the user's chosen accent for this one view.** `applyUiAccent` (`packages/ui/src/state/persistence.ts:459`) writes the `ACCENT_OVERRIDE_VARS` family — `--accent`, `--accent-rgb`, `--accent-hover`, `--accent-muted`, `--accent-subtle`, `--ring`, `--border-hover`, `--primary` — inline on `document.documentElement` specifically so the choice "wins over base.css / any host brand theme". An inline custom property on this `<section>` is a nearer ancestor, so inside the calendar the user's selection is silently discarded. The theme default is overridden the same way: `packages/ui/src/styles/theme.css` sets `--accent: var(--brand-black)` with `--accent-foreground: var(--brand-white)` in one mode and the inverse pair in the other, and this pins orange on `#111111` in both.

**The override is partial, so the accent family becomes internally inconsistent.** It sets three of those variables and leaves the other five at their theme or user values. `MonthDayButton` makes that visible in one grid row: today's cell uses `variant="accentDarkHover"` — `bg-accent text-accent-fg hover:bg-accent-muted`, all three overridden, so orange — while every other cell uses `variant="selection"`, whose selected state is `data-[state=on]:bg-accent-subtle`, and `--accent-subtle` is *not* overridden. The selected day and today's day therefore fill from two different accent families side by side. `--accent-rgb` is likewise left pointing at the old accent.

**Smallest repair:** delete the `style` block. With line 53 of `styles.css` in place, `bg-accent` / `text-accent-fg` / `bg-accent-muted` now generate and resolve against the theme, which is what the rest of the app renders; nothing else in the diff depends on the override. If a brand-orange calendar really is intended, it needs to define `--eliza-brand-orange` in the theme layer and set the whole `ACCENT_OVERRIDE_VARS` family, not three of it.

### Non-blocking

- Stripping the alpha modifiers (`border-border/12` → `border-border`, `border-border/6`, `bg-border/8` → `bg-border`) is a real visual change, not a no-op: the grid's hairlines and the month gap-lines go to full-opacity border color. With the `@source` fix those fractional utilities would now generate correctly, so if the faint rules were the intended design they can be kept. Worth a sentence in the PR body either way, since the `packages/app` visual contract asks for before/after captures on a touched view.
- `CalendarStatusIcon` changed from `sr-only` to a visible label, and the empty/partial states now also render a separate `role="status"` paragraph plus a Retry button above the grid. On an empty feed the label is not duplicated (the empty branch no longer routes to `CalendarStatusIcon`), but `unavailable` and the loading state now both show visible text where they previously showed an icon with a screen-reader-only label — please eyeball those two states.
- `plugins/plugin-calendar`'s 13 test files that fail to load with `Cannot find package '@elizaos/login'` fail identically on `origin/develop` and are not attributable to this PR.

**Reproduction:** `origin/develop`, the six PR files checked out over it, `npx vitest run --config vitest.config.ts src/components/CalendarSection.test.tsx` from `plugins/plugin-calendar`; RED control by restoring only `origin/develop`'s `CalendarSection.tsx` under the same test file.

lalalune:
Ownership/status: this Bettina task owns #31029 in the isolated `bettina-calendar` worktree. The reviewed candidate has been rebased onto `e2d6d383bc3b10f35ab509748d00788be9b72769`; local head is `d6b51a64ce2f370b2dd02a5704ac8645414fd909`, not yet pushed while final root verification runs. Production renderer and chunk-safety checks passed. The browser regression now exercises the actual built remote Connections bundle: the old host fails to resolve `@elizaos/shared` on desktop/mobile, and the repaired host succeeds. Final captures and evidence packaging are in progress. Please leave this draft to this lane.

This task also owns #30944, rebased locally to `dcf5b62cdb198daeb98a88c019db17d7a49437a3` after develop landed the overlapping provider-schema test correction. Its earlier hosted checks were green, but the new candidate requires fresh checks before merge. No develop advance is planned until the relevant exact-head gates finish. Further heavy local pipelines will be serialized with bounded concurrency.

ss251 (approved):
437dc54cde57a7e4f431541f356e9bf0bd597b96

The diff addresses two distinct functional regressions in the app: dynamic view loading failures for `@elizaos/shared` consumers, and calendar grid controls being clobbered or suppressed on mobile screens.

Key findings:
1. Dynamic host resolution: In `packages/ui/src/components/views/DynamicViewLoader.tsx:534`, `@elizaos/shared` was loaded via `importHostExternal("@elizaos/shared")`, which failed to resolve host-provided exports in built environments. Switching to a standard dynamic `import("@elizaos/shared")` allows bundler resolution to wire host modules correctly to dynamic views like Connections.
2. Calendar responsive projection: In `plugins/plugin-calendar/src/components/CalendarSection.tsx:1047-1340`, mobile screens (<768px) previously used `useMediaQuery` to unconditionally force `AgendaView`, discarding user-selected Day/Week/Month projections and hiding the grid. The diff eliminates the viewport hook, renders `TimeGrid` within a horizontal scroll container (`min-w-3xl` for multi-day views at line 1394), and introduces `MonthDayButton` (line 921) to project a single-day agenda beneath `MonthGrid` on selection rather than obliterating the entire grid.
3. CSS bundle inclusion: `packages/ui/src/styles/styles.css:53` adds `@source "../../../../plugins/plugin-calendar/src"`, ensuring calendar styles are not purged during styling compilation.
4. Error recovery and test coverage: `CalendarSection.tsx:1350-1362` adds explicit retry triggers for error/partial states, and smoke specs in `apps-personal-assistant-decomposed-interactions.spec.ts:136-230` verify calendar projections and connection recovery at both 1280px and 390px.

Verified by reading the diff at this head and validating module imports, responsive layout hierarchy, and smoke test assertions.

Recommendation: merge

— [free-apinex-lane-ss251]

lalalune:
The local follow-up fixes the accent issue: the orange default now supplies the full accent family, while any non-default preference inherits the host theme instead of being overridden by the calendar section. This preserves the repository guide's orange default and darker-orange hover requirement. `--eliza-brand-orange` is defined in `packages/shared/src/brand/brand.css` and `brand-classic/brand.css`; the preference override and partial-family inconsistency were nevertheless real and are addressed.

The new built-browser regression checks both the current day and a selected day against the host's computed accent family after loading a persisted preference. It passes with the correction and fails against the preceding production bundle, which forces orange. The corrected production source and bundle are restored. The PR is back in draft while final gates and affected evidence are refreshed. No deployment or merge is being claimed.

Issue #30130, opened 2026-08-31
fix(cloud): restore production Hyperdrive to protected account database


lalalune:
## Incident

The production Cloud API deployment at `8eebf60dc18e1032cb23a3df70a1e76f76af49d7` binds `HYPERDRIVE` to a database that does not contain the active production organization, API-key, account, or ledger state. The currently protected Railway production database remains authoritative.

## Acceptance criteria

- Change only the production Hyperdrive configuration ID relative to the currently served certified base.
- Pass the bounded production Hyperdrive admission verifier against the protected `DATABASE_URL`.
- Deploy through the protected binding-only recovery workflow using the immutable served-base and develop-policy staging certificates.
- Verify production API-key authentication, credit balance, and a real chat-completions response after deployment.
- Verify the paired Bettina VPS can complete a Cloud-backed chat without browser console or network authentication errors.

No credentials, tenant identifiers, or ledger details belong in the issue.

lalalune:
Live Bettina VPS check on 2026-09-10 at 19:55–19:57 UTC:

- Production health serves `5525597f50e7792fe105b79cff06ecb69c364317`, not the incident revision named above.
- The pilot's configured Cloud credential authenticates: `GET /api/v1/credits/balance?fresh=true` returns HTTP 200 with the expected numeric balance response. No credential, tenant identifier, or balance value is included here.
- A minimal real `qwen-3.8-27b` chat-completions request repeatedly returns HTTP 503, `code: rate_limit_unavailable`, `Retry-After: 1`, and `X-RateLimit-Policy: durable-object`. The last request has `gateway_auth;dur=119` and `gateway_middle;dur=1505`; no model completion was produced.
- The served source still gives the rate-limit Durable Object a single 1,500 ms attempt. The bounded idempotent retry implemented for #30387, incorporated through merged #30400 (`55c38a4df335a81a2f9c450236ceffe7f929e223` in its history), is not an ancestor of the served revision. This identifies a relevant unreleased repair; it does not prove that a new release alone will resolve the live failure.

The old missing-account symptom did not reproduce for this pilot. This issue stays open because real Cloud-backed chat and fresh paired-browser acceptance still fail or remain unverified. No production binding, database, deployment, VPS configuration, or messaging state was changed by this check. The next release must retain the protected Cloud deployment and exact-revision live acceptance requirements; a health response alone cannot close this issue.

lalalune:
[...]

PR #31041, opened 2026-09-10
fix(cloud): retain consent scope in repository updates

### LINKED ISSUE #31040: fix(cloud): preserve consent scope in transactional update predicates
The Cloud lint-and-types lane fails on current develop `7b0099eab340734e78a418ad8baa38ce6e4c795c` because the consent transaction uses primary-key-only update predicates after selecting and locking the applicable app and user membership. The transaction already establishes the relevant scope; the mutation predicates should retain it explicitly so the tenant-scope gate can validate the complete operation.

Observed in Develop Full [run 34530365498](https://github.com/elizaOS/eliza/actions/runs/34530365498), job `cloud / lint-and-types`. This blocks Cloud release validation needed for the Bettina work in #30130.

Acceptance:
- Keep app consent and registration under the existing transaction and row lock.
- Preserve the selected app/user and active/approved-app conditions in the mutations, without adding a blanket scope exemption.
- Pass the tenant-scope checker and existing real-database consent/registration tests, including duplicate approvals and separation across apps and buyers.
- Pass applicable package and repository gates and exact-head hosted checks before merge.

I am taking this bounded validation repair. No deployment or Cloud resource mutation is included.

lalalune:
The consent transaction now retains the selected app/user predicates when updating membership and repeats the active/approved-app conditions when updating its counter. This clears the Cloud tenant-scope gate without an exemption and preserves the existing transaction and row lock.

# Relates to

Fixes #31040. Supports Cloud release validation for #30130.

# Sync with develop

Candidate `467954ee5e79f0ea13f661f3e7d740cac71c26ae`, based on `7b0099eab340734e78a418ad8baa38ce6e4c795c`. Normal installation passed with Node 24.15.0 and Bun 1.3.14. Repository verification passed all 373 tasks and final audits. The complete shared-package suite passed. Exact-head hosted checks must finish successfully before merge.

# Risks

Low. The additional predicates repeat conditions already established in the same locked transaction. Consent updates, registration materialization and counters remain in that transaction.

# Background

## What does this PR do?

Makes the consent mutation scope explicit to the repository checker. The real consent behavior is covered by existing database tests.

Non-breaking repository validation fix.

None; no API or operational contract changes.

# Testing

The tenant-scope checker passes across 173 repository files. The existing apps and billing-consent PGlite suites pass 59 tests. The receipt capture reruns the unchanged consent assertions and passes 11 tests, recording synthetic app, membership and billing-account rows before teardown. Complete shared-package validation passed: 1,231 files across 388 batches, 12,637 passing tests and 243 existing skips; process exit 0. [Full suite log](https://github.com/elizaOS/eliza/releases/download/pr-evidence-12/31041-bettina-31040-shared.log), [exact-head result](https://github.com/elizaOS/eliza/releases/download/pr-evidence-12/31041-bettina-31041-full-suite-result.json).

[Evidence archive](https://github.com/elizaOS/eliza/releases/download/pr-evidence-12/31041-bettina-31040-467954e-evidence.zip) includes root verification, installation, checker output, database logs, synthetic rows, the reproducible receipt preload and a verified SHA-256 manifest. The archive was captured before the full package run; its pending-suite note is superseded by the full-suite result below. Hosted checks remain pending.

Compare the failed tenant-scope step in Develop Full run 34530365498 with the passing checker log and inspect the captured database rows alongside the consent test log.

## Detailed testing steps

Run `bun run --cwd packages/cloud/shared check:tenant-scope`. From that package, run the existing `src/db/repositories/app-billing-accounts.pglite.test.ts` and `src/db/repositories/__tests__/apps.test.ts` database suites. The evidence includes the external receipt preload; it adds read-only queries before test teardown without modifying repository source or assertions.

# Evidence Gate

N/A - no rendered surface changes; the defect is a repository validation failure.

N/A - no rendered surface changes; consent behavior remains the same.

N/A - no changed UI flow; real SQL and checker results are the acceptance boundary.

- [x] [31041-bettina-31040-focused.log](https://github.com/elizaOS/eliza/releases/download/pr-evidence-12/31041-bettina-31040-focused.log)

N/A - no frontend or transport changes.

N/A - no agent, prompt or model behavior changes.

- [x] [31041-bettina-31040-consent-rows.json](https://github.com/elizaOS/eliza/releases/download/pr-evidence-12/31041-bettina-31040-consent-rows.json)

N/A - no rendered text or visual surface changes.

# Evidence Details

## Real LLM-call trajectory

N/A - no model call is involved.

## Backend + frontend logs

Database tests exercise real Drizzle/PGlite repository operations, registration and HTTP readback. The captured records are synthetic. Frontend logs are N/A because no frontend code or transport behavior changes.

N/A - repository predicates only, with no changed visual state.

## Audio / voice walkthrough

## Known gaps / failures

The current develop Cloud tenant-scope step fails before this change. This PR clears that gate; it does not claim to repair every unrelated Develop Full failure. Repository and full-package gates pass. Hosted checks remain required before merge. No deployment is included.

N/A - no exception requested.

### DIFF
--- a/packages/cloud/shared/src/db/repositories/apps.ts
+++ b/packages/cloud/shared/src/db/repositories/apps.ts
@@ -431,7 +431,13 @@ export class AppsRepository {
             ...(input.ipAddress != null ? { ip_address: input.ipAddress } : {}),
             ...(input.userAgent != null ? { user_agent: input.userAgent } : {}),
           })
-          .where(eq(appUsers.id, existing.id));
+          .where(
+            and(
+              eq(appUsers.id, existing.id),
+              eq(appUsers.app_id, input.appId),
+              eq(appUsers.user_id, input.userId),
+            ),
+          );
       } else {
         await tx.insert(appUsers).values({
           app_id: input.appId,
@@ -443,7 +449,9 @@ export class AppsRepository {
         await tx
           .update(apps)
           .set({ total_users: sql`COALESCE(${apps.total_users}, 0) + 1` })
-          .where(eq(apps.id, input.appId));
+          .where(
+            and(eq(apps.id, input.appId), eq(apps.is_active, true), eq(apps.is_approved, true)),
+          );
       }
       await materializeAppBillingAccounts(tx, input.appId);
       return existing ? "updated" : "created";


attentionhead (approved):
**APPROVE** — verified at head `467954ee5e79f0ea13f661f3e7d740cac71c26ae` (draft, 1 file, +10/−2 against merge-base `7b0099eab340734e78a418ad8baa38ce6e4c795c`).

## What the PR actually does, and why it is safe

The premise checks out first. `AppsRepository.connectUser` already opens with a `SELECT ... FOR UPDATE` on `apps` gated on exactly `id = input.appId AND is_active = true AND is_approved = true`, throwing `APP_CONSENT_UNAVAILABLE` when it misses. Both new predicate sets are therefore *repeats of conditions already established and row-locked in the same transaction*, not new gates:

- `appUsers` update — `existing.id` was resolved two statements earlier by `and(eq(app_id), eq(user_id))`. Re-asserting those alongside the PK cannot narrow the match to zero; `app_id`/`user_id` are the row's identity and no writer mutates them.
- `apps.total_users` increment — the row is held under `FOR UPDATE` with the same `is_active`/`is_approved` predicates, so the flags cannot flip between the lock and the increment. The counter cannot silently drift by matching 0 rows, which is the failure mode this shape would normally invite.

So this is a validation-visibility change, not a behavioral one — which is exactly what the description claims.

## Bidirectional control (this is the evidence, not the opinion)

Same tree, same checker, only `packages/cloud/shared/src/db/repositories/apps.ts` swapped between base and head:

| `apps.ts` revision | `bun run --cwd packages/cloud/shared check:tenant-scope` |
|---|---|
| base `7b0099e` | ✗ **2 pk-only reads**, exit 1 — `apps.ts:426 connectUser() → pk-only appUsers.id predicate`, `apps.ts:443 connectUser() → pk-only apps.id predicate` |
| head `467954e` | ✓ no unannotated pk-only reads across **173** repository files |

The checker names precisely the two sites this diff touches and nothing else, so the change is both necessary and minimal for the gate. Note the alternative the checker itself offers — a `/* global-scope: <reason> */` annotation — was *not* taken; the PR scopes the queries instead, which is the stronger of the two options and is consistent with the description's "clears the gate without an exemption".

## Behavior-neutrality control

Same two suites named in the description, run under `bun test` from `packages/cloud/shared` (this package runs on `bun test`, not `vitest run` — its `vitest.config.ts` `include` does not cover `src/db/repositories`):

| `apps.ts` revision | `app-billing-accounts.pglite.test.ts` + `__tests__/apps.test.ts` |
|---|---|
| head `467954e` | **59 pass / 0 fail**, 325 assertions |
| base `7b0099e` | **59 pass / 0 fail**, 325 assertions |

Identical on both sides — which is the correct result here and confirms the diff adds no observable consent behavior. It also means these suites do not *pin* the new predicates; the tenant-scope checker is the thing that does, and that asymmetry is shown above.

`npx tsc --noEmit -p tsconfig.json` in the package reports only the pre-existing `Cannot find module '@elizaos/login'` errors in `src/lib/services/server-wallets.ts` and `src/lib/services/steward-client.ts` (plus the `'error' is of type 'unknown'` that cascades from the first). Those reproduce on `develop` independently of this PR and are not attributable to it. Nothing in `apps.ts`.

## Non-blocking notes

1. `materializeAppBillingAccounts(tx, input.appId)` runs on both the created and updated arms and is not itself scoped by the active/approved predicates — it inherits safety from the `FOR UPDATE` gate above rather than from its own `where`. That is fine today, but it is the one call in this transaction whose scoping is implicit rather than expressed, so if the gate is ever relaxed it is the site to revisit.
2. Draft status is noted; per the description, shared-package and hosted validation are still pending. Nothing in the diff itself is blocking.

ss251 (changes requested):
467954ee5e79f0ea13f661f3e7d740cac71c26ae

The appUsers UPDATE now re-asserts `app_id` and `user_id` (apps.ts:434-441). That is the right tenant isolation for a consent-row write. The same hunk also changes a different invariant, and both UPDATEs can match zero rows while the method still reports success.

Finding — create path desynchronizes `total_users`.
input: insert a new `appUsers` row for an app that is not both `is_active` and `is_approved`
wrong output: the insert at apps.ts:442-448 still commits; the increment at apps.ts:449-454 matches 0 rows; the method returns `"created"` (apps.ts:456). `total_users` then disagrees with `appUsers` cardinality for that app. Before this head the increment was only `eq(apps.id, input.appId)`. Gating the counter but not the insert is not "retain consent scope"; it is a new product rule applied to one of two writes.

Finding — existing-user path can drop this call's consent fields and still return `"updated"`.
input: `existing.id` present, but `app_id`/`user_id` do not equal `input.appId`/`input.userId`
wrong output: UPDATE at apps.ts:434-441 writes 0 rows; `ip_address`/`user_agent` from this call are not stored; caller still gets `"updated"`. Keep the extra predicates. Fail the transaction (or return a distinct result) when rowCount is 0.

No test in this diff would fail if either new WHERE were reverted. Two concerns in one unlabeled hunk: tenant predicates on `appUsers` vs approval gates on `apps`.

Adversarial: no lockfile, CI, binary, symlink, submodule, or hook changes in this diff.

Verified by reading the diff at this exact head. Live execution blocked: no disposable sandbox on this host, so this is a static review of the diff at the exact head.

Recommendation: fix

— [free-xai-lane-ss251]

PR #31043, opened 2026-09-10
fix(app-test): supply native bridge in desktop renderer fixtures

### LINKED ISSUE #31042: fix(app-test): provide desktop bridge for voice and local-onboarding fixtures
Current develop `7b0099eab340734e78a418ad8baa38ce6e4c795c` fails the voice desktop self-test and explicit-local onboarding smoke before either intended flow appears. Both fixtures set `__electrobunWindowId` but omit the RPC/native protected-storage contract. Captured browser traces report `[desktop-shell] Native Electrobun bridge is unavailable`; the page renders the boot-failure screen. This is the same fixture-contract class previously repaired in other desktop-emulating specs (#30391, #30409).

Evidence: [Develop Full run 34530365498](https://github.com/elizaOS/eliza/actions/runs/34530365498), jobs `scenarios / Zero-Key app browser core` and `dev-smoke / default staging + local onboarding`. Downloaded traces establish the cause; increasing selector timeouts would not repair it.

Acceptance:
- Supply the required native RPC and protected-storage test boundary before browser boot in these desktop-emulating fixtures.
- Preserve production bridge readiness and protected-storage failure behavior.
- Exercise the actual voice self-test route and local runtime chooser in their existing browser lanes, including the existing mobile chooser assertions.
- Keep unexpected browser diagnostics visible; distinguish browser fixtures from real native-device evidence.
- Complete affected app validation, visual evidence, repository verification and exact-head hosted checks.

I am taking this bounded repair as a release-validation dependency of the Bettina work in #30130. No deployment, native installation or external message delivery is included.

lalalune:
The desktop voice self-test and explicit-local runtime chooser marked Chromium as an Electrobun renderer without providing its required native bridge. Both could stop at the boot-failure screen before exercising their intended behavior. A shared browser fixture now supplies the synthetic host and protected-storage boundary before boot, preserving the production readiness checks and the existing flow assertions.

Fixes #31042. Supports Bettina release validation in #30130.

The existing browser tests now exercise desktop TTS routing and the desktop/mobile runtime chooser. These are controlled browser fixtures, not native-device or live-model proof. The voice header now states that boundary explicitly.

Validation: Node 24.15.0 / Bun 1.3.14; normal installation; root verification (373 tasks and final audits); app typecheck and lint; 950 script tests plus 1,630 component tests passed, with two existing script skips. The chooser passed; its companion live-chat test was skipped because no provider key is configured. The full app audit passed all 222 checks with zero broken or needs-work findings; OCR found zero broken views. Exact-head hosted checks remain required before merge.

The first visual run had one inventory bundle HTTP 424 while dependency builds were replacing generated output. That run is retained; the isolated rerun passed after the builds finished. No production module, API, model behavior or deployment changes.

- [x] before-screenshots

- [x] after-screenshots

- [x] https://github.com/elizaOS/eliza/releases/download/pr-evidence-12/31043-voice-walkthrough.mp4

- [x] [31043-bettina-31042-voice-record.log](https://github.com/elizaOS/eliza/releases/download/pr-evidence-12/31043-bettina-31042-voice-record.log)

- [x] [31043-bettina-31042-frontend.log](https://github.com/elizaOS/eliza/releases/download/pr-evidence-12/31043-bettina-31042-frontend.log)

- [ ] N/A - controlled browser fixtures; no model behavior changed or live-model proof claimed

- [x] [31043-source-provenance.json](https://github.com/elizaOS/eliza/releases/download/pr-evidence-12/31043-source-provenance.json)

- [x] [31043-visual-review.md](https://github.com/elizaOS/eliza/releases/download/pr-evidence-12/31043-visual-review.md)

## Review artifacts

[Complete source-bound evidence archive](https://github.com/elizaOS/eliza/releases/download/pr-evidence-12/31043-bettina-31042-977fcc-evidence.zip) includes verification and package logs, the original and isolated audit results, screenshots, the complete controlled voice recording, trace, OCR result, and SHA-256 manifest. Archive SHA-256: `07a3612a61792ed7045920c269f54172de08a4e148a83ac2259ffd69bfe465e7`.

Desktop runtime chooser:

Desktop runtime chooser

Mobile runtime chooser:

Mobile runtime chooser

### DIFF
--- a/packages/app/test/dev-smoke/runtime-chooser-default.spec.ts
+++ b/packages/app/test/dev-smoke/runtime-chooser-default.spec.ts
@@ -11,6 +11,7 @@ import {
   installRenderTelemetryGuard,
   seedAppStorage,
 } from "../ui-smoke/helpers";
+import { installDesktopBridgeFixture } from "../ui-smoke/helpers/desktop-bridge";
 
 test.skip(
   process.env.ELIZA_DEV_SMOKE_OFFLINE !== "1",
@@ -54,11 +55,11 @@ test("explicit local development offers cloud, local, and remote", async ({
   await installRenderTelemetryGuard(page);
   await installDefaultAppRoutes(page);
   await routeFreshFirstRun(page);
+  await installDesktopBridgeFixture(page);
   await page.addInitScript(() => {
     const win = window as unknown as Record<string, unknown>;
     win.__ELIZA_APP_API_BASE__ = window.location.origin;
     win.__ELIZAOS_APP_BOOT_CONFIG__ = { apiBase: window.location.origin };
-    win.__electrobunWindowId = 1;
   });
   await seedAppStorage(page, {
     "eliza:first-run-complete": "",
--- a/packages/app/test/ui-smoke/helpers/desktop-bridge.ts
+++ b/packages/app/test/ui-smoke/helpers/desktop-bridge.ts
@@ -1,0 +1,38 @@
+/** Supplies the native host boundary for browser-only desktop renderer tests; storage is synthetic and local to each page load. */
+import type { Page } from "@playwright/test";
+
+export async function installDesktopBridgeFixture(page: Page): Promise<void> {
+  await page.addInitScript(() => {
+    const secureStore = new Map<string, string>();
+    Object.assign(window, {
+      __electrobunWindowId: 1,
+      __ELIZA_ELECTROBUN_RPC__: {
+        request: {
+          desktopGetVersion: async () => ({ runtime: "playwright-smoke" }),
+          desktopRegisterShortcut: async () => ({ success: true }),
+          desktopSetTrayMenu: async () => undefined,
+          secureStoreGet: async ({ kind }: { kind: string }) =>
+            secureStore.has(kind)
+              ? { ok: true, value: secureStore.get(kind) }
+              : { ok: false, reason: "not_found" },
+          secureStoreSet: async ({
+            kind,
+            value,
+          }: {
+            kind: string;
+            value: string;
+          }) => {
+            secureStore.set(kind, value);
+            return { ok: true };
+          },
+          secureStoreDelete: async ({ kind }: { kind: string }) => ({
+            ok: true,
+            deleted: secureStore.delete(kind),
+          }),
+        },
+        onMessage: () => undefined,
+        offMessage: () => undefined,
+      },
+    });
+  });
+}
--- a/packages/app/test/ui-smoke/voice-desktop-selftest.spec.ts
+++ b/packages/app/test/ui-smoke/voice-desktop-selftest.spec.ts
@@ -1,20 +1,12 @@
 /**
- * DESKTOP voice self-test (Electrobun renderer path), headless.
- *
- * The desktop voice surface is the SAME renderer bundle as web — the Electrobun
- * pill loads it with a ?shellMode= param over a Chromium-engine webview. We
- * inject the Electrobun runtime marker (window.__electrobunWindowId) so the
- * self-test screen detects platform=desktop and routes TTS through the desktop
- * local-inference TTS route, then drive the real round-trip harness. This proves
- * the desktop CONFIG path of the voice self-test green in CI without a packaged
- * Electrobun build. (The native Electrobun shell integration — talkmodeSpeak
- * main-process bridge, views:// mic grant — is exercised by the packaged
- * electrobun-packaged lane.)
- *
- *   bun run --cwd packages/app test:e2e test/ui-smoke/voice-desktop-selftest.spec.ts
+ * Exercises the desktop voice self-test renderer with a synthetic native bridge
+ * and controlled ASR, conversation, and TTS responses. The browser drives the
+ * real self-test UI and verifies desktop TTS routing; native shell and physical
+ * audio-device behavior belong to the packaged Electrobun lane.
  */
 import { expect, type Page, test } from "@playwright/test";
 import { installDefaultAppRoutes, seedAppStorage } from "./helpers";
+import { installDesktopBridgeFixture } from "./helpers/desktop-bridge";
 
 const EXPECTED_PHRASE = "what time is it";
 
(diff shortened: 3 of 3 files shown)


ss251 (approved):
Reviewed head 977fcc2b8b02a5bf8db48cef34988de717f58bc5.

The risky part is the fixture swap in `packages/app/test/ui-smoke/voice-desktop-selftest.spec.ts:92`: the old init script set only `__electrobunWindowId`, so the desktop code path could boot and then hit a missing `__ELIZA_ELECTROBUN_RPC__`. The new fixture (helpers/desktop-bridge.ts) sets the marker and the full `request` surface plus `onMessage`/`offMessage` in one `addInitScript`, and init scripts run before boot, so the renderer can no longer observe desktop mode without a bridge. Ordering in runtime-chooser-default.spec.ts is also correct: the fixture is installed before the `__ELIZA_APP_BOOT_CONFIG__` init script, and both are init-phase so ordering across them is deterministic. One boundary I followed: `secureStoreGet` returns `{ok:false, reason:"not_found"}` rather than throwing on a miss — matches the typed result shape the delete/set paths assume. Non-blocking: `desktopGetVersion` returns a synthetic `runtime` string; if any assertion later greps for a real version it will silently mismatch.

Verified by reading the full diff at this head and tracing both call sites and the init-script ordering; I did not execute the Playwright specs.

Recommendation: merge

— [free-9router-lane-ss251]

attentionhead (approved):
Reviewed at `977fcc2b8b02a5bf8db48cef34988de717f58bc5` (merge-base `7b0099eab340734e78a418ad8baa38ce6e4c795c`, two-dot diff: 3 files, +46/−20, all under `packages/app/test/` — no production source touched).

## The premise is real, and it is not the marker check

I traced the boot failure before judging the fixture, because two plausible culprits turn out to be red herrings:

- `packages/app/src/entry.ts` `hasDesktopShellMarker()` only selects a renderer entry — it never throws.
- `packages/ui/src/bridge/electrobun-rpc.ts:39-51` `invokeDesktopBridgeRequest` returns **`null`** when no RPC object is present; it does not throw.

The actual throw site is `packages/ui/src/bridge/storage-bridge.ts:171-190`:

```ts
if (isElectrobunRuntime()) {
  const result = await desktopSecureStoreGet(kind);
  if (result?.ok) { ... }
  if (result?.reason === "not_found") return null;
  throw new Error("Desktop protected storage is unavailable");
}
```

`isElectrobunRuntime()` (`electrobun-runtime.ts:36-52`) goes true on `__electrobunWindowId` **alone**, before it ever looks for the bridge. So the old inline `win.__electrobunWindowId = 1` declared a desktop host whose protected-storage boundary did not exist: `result` is `null` → `result?.ok` falsy, `result?.reason` undefined → throw, reached from `initializeStorageBridge` (line 325, credential migration at 418/437) and `getStorageValue` (line 677). That is exactly the class of failure this PR fixes, and the fix is at the right layer.

## Bidirectional control

Playwright here needs a built app plus a webServer, so I pinned the same boundary at jsdom level in `packages/ui`, with Capacitor mocked as non-native web and the PR's fixture object reproduced verbatim:

| arm | window state | `getStorageValue("eliza.device.auth")` |
|---|---|---|
| A | `__electrobunWindowId = 1`, **no** `__ELIZA_ELECTROBUN_RPC__` (pre-PR shape) | **rejects** — `Desktop protected storage is unavailable` |
| B | fixture installed as written in this PR | **resolves `null`**, and `setStorageValue` → `getStorageValue` round-trips `"probe-token"` through the synthetic `Map` |

`Test Files 1 passed · Tests 2 passed`, from `packages/ui` with the package config. Arm A reproduces the failure the PR claims; arm B shows this fixture — not some larger stub — is what removes it. The discriminant is precise: `secureStoreGet` returning `{ ok: false, reason: "not_found" }` is the one shape that converts the throw into a `null`, and every return shape in the helper matches the declared `DesktopSecureStoreResult` / `DesktopSecureStoreKind` union in `electrobun-rpc.ts:116-129` exactly. Probe deleted; tree restored.

- **System under test is real.** Both specs still boot the real renderer bundle and drive the real self-test / chooser UI. Nothing under test is replaced by a mock.
- **The fixture stands for an external boundary, not for the code being pinned.** `__ELIZA_ELECTROBUN_RPC__` is the native Electrobun host process — an OS-side boundary a browser lane genuinely cannot have. The header rewrite in `voice-desktop-selftest.spec.ts` says so explicitly ("native shell and physical audio-device behavior belong to the packaged Electrobun lane"), which is the right way to declare it.
- **Not invented.** The shape is a faithful extraction of the already-documented minimum desktop startup contract in `all-pages-clicksafe.spec.ts:515-545`, including the `desktopGetVersion` / `desktopRegisterShortcut: { success: true }` / `desktopSetTrayMenu` trio that `initializeDesktopShell()` validates before rendering.
- **No new `vi.mock`/route stubs added**; the specs keep their pre-existing harnesses.
- **Ordering is safe.** In `runtime-chooser-default.spec.ts` the fixture is a separate `addInitScript` registered before the existing one; both run in registration order at page load, and the line it removes (`win.__electrobunWindowId = 1`) lived in the second script, so the marker is still set before boot with no double-assignment.

## Non-blocking notes

1. **Partial dedupe.** Four other specs still inline this same stub — `all-pages-clicksafe.spec.ts`, `home-widget-priority.spec.ts`, `onboarding-to-home.shared.ts`, `settings-background.spec.ts`. They need extra methods (`permissionsGetAll`, `permissionsGetPlatform`, …), so adopting the helper there would mean giving it an overrides argument. Worth a follow-up so the contract has one owner rather than five copies.
2. **`desktopGetVersion` returns `{ runtime: "playwright-smoke" }`** while `DesktopVersionInfo` (`packages/ui/src/utils/desktop-workspace.ts:67`) declares `version`, `name`, and `runtime` as required. This matches the existing inline stubs exactly, so it is pre-existing rather than a regression — but if the helper becomes the single owner it is the natural place to complete the shape.
3. The rewritten `voice-desktop-selftest.spec.ts` header drops the copy-pasteable run line (`bun run --cwd packages/app test:e2e test/ui-smoke/voice-desktop-selftest.spec.ts`). Small loss; the new text is otherwise clearer.

Approving: the diagnosis is correct, the fixture is minimal and contract-accurate, and it is confined to test scaffolding.

PR #31046, opened 2026-09-10
test(desktop): exercise notification IPC routing instead of source strings

### LINKED ISSUE #31045: fix(desktop-test): exercise notification routing without source-location assertions
Develop Full `34538119429` at `963d8ef787190680b1dbe86f0b6293efd127491f` fails the Electrobun notification contract test because it requires three renderer source files to contain native RPC literals. Notification delivery now delegates through the shared delivery owner, so source-location assertions reject the valid composition.

Reproduced locally: `rpc-handlers.test.ts` has one failure and 11 passing tests. The existing notification store/native bridge suites pass 96 behavioral tests and exercise real store delivery decisions against controlled native boundaries.

Scope: remove the obsolete file-content loop and fold the IPC-to-RPC mapping check into the existing notification dispatch test, using the mapped method to exercise the production handler and inspect its native call. Keep payload, close, background-notice and error behavior coverage. Do not move production calls back into consumers or change native/UI behavior.

Acceptance: the current notification handler and behavioral suites pass; an invalid channel mapping is rejected by the dispatch test; complete owning-package and repository gates pass; exact-head hosted checks pass before merge. This is a Bettina release-validation dependency of #30130, not a native installer or production notification change.

lalalune:
The desktop contract lane fails because a test searches three renderer files for notification RPC strings that now live in a shared delivery helper. Replace that obsolete source inspection by routing the existing real notification-handler test through the IPC map. Existing renderer delivery tests already exercise the shared bridge.

Fixes #31045. Supports Bettina validation under #30123.

Candidate `d204905bd6a3ce55a2adfd7c287819849100dbe9` is based on `c745213d6100bc941e0a78f2c6105462b2e5925e`. Normal installation passed with Bun 1.3.14 and Node 24.15.0. Root verification passed all 373 tasks and final audits.

Validation: 1,492 desktop package tests, 96 existing renderer/store tests, desktop typecheck and lint pass. The old test reproduces its source-string failure. Deliberately mapping notification IPC to the close handler makes the repaired test fail at the routing boundary (1 fail, 10 pass). Production code and native behavior are unchanged; deterministic boundary fixtures are not native delivery evidence.

The regression protected is an IPC routing error observed by desktop notification callers, including normal and low urgency payloads and returned notification IDs. No new inventory-only test is introduced.

# Evidence Gate

N/A - test-only contract repair; no rendered UI, native implementation, model, or domain behavior changes.

N/A - test-only contract repair; no rendered UI, native implementation, model, or domain behavior changes.

N/A - test-only contract repair; no rendered UI, native implementation, model, or domain behavior changes.

N/A - test-only contract repair; no rendered UI, native implementation, model, or domain behavior changes.

N/A - test-only contract repair; no rendered UI, native implementation, model, or domain behavior changes.

N/A - test-only contract repair; no rendered UI, native implementation, model, or domain behavior changes.

N/A - test-only contract repair; no rendered UI, native implementation, model, or domain behavior changes.

- [x] [31046-bettina-31046-c745-package.log](https://github.com/elizaOS/eliza/releases/download/pr-evidence-12/31046-bettina-31046-c745-package.log)

## Known gaps / failures

Exact-head hosted checks remain required before merge. The broader Develop Full run has separate family UI and scheduler failures being investigated; this PR claims only the notification test repair.

[Reviewed evidence archive](https://github.com/elizaOS/eliza/releases/download/pr-evidence-12/31046-bettina-31045-c9180-evidence.zip) contains the before failure, after package and renderer logs, negative routing probe, installation, repository verification, source provenance and SHA-256 manifest.

Current-base qualification: normal installation, all 373 root tasks and final audits, and all 1,492 desktop tests pass after rebasing over #30229. The previous archive remains explicitly historical; current logs are attached below.

[Current d204 evidence archive](https://github.com/elizaOS/eliza/releases/download/pr-evidence-12/31046-bettina-31046-d204-evidence.zip) includes fresh installation, root verification, desktop and renderer results, source-blob comparison and SHA-256 manifest.

### DIFF
--- a/packages/app-core/platforms/electrobun/src/rpc-handlers.test.ts
+++ b/packages/app-core/platforms/electrobun/src/rpc-handlers.test.ts
@@ -1,7 +1,4 @@
 /** Exercises rpc handlers behavior with deterministic app-core test fixtures. */
-import { readFileSync } from "node:fs";
-import path from "node:path";
-import { fileURLToPath } from "node:url";
 import { describe, expect, it, vi } from "vitest";
 import {
   buildDynamicViewRpcHandlers,
@@ -9,9 +6,6 @@ import {
   buildWindowRpcHandlers,
 } from "./rpc-handler-slices";
 import { CHANNEL_TO_RPC_METHOD } from "./rpc-schema";
-
-const HERE = path.dirname(fileURLToPath(import.meta.url));
-const REPO_ROOT = path.resolve(HERE, "../../../../..");
 
 function createDesktopFixture() {
   const desktop = {
@@ -195,7 +189,7 @@ describe("window RPC handlers", () => {
 });
 
 describe("notification RPC handlers", () => {
-  it("delegates show/close notification requests with exact params", async () => {
+  it("routes notification IPC to the show handler and preserves show/close params", async () => {
     let nextId = 0;
     const desktop = {
       showNotification: vi.fn(async (params) => ({
@@ -214,9 +208,15 @@ describe("notification RPC handlers", () => {
       userDataDir: "/tmp/eliza-user-data",
       showNotification: vi.fn(),
     });
+    const showMethod = CHANNEL_TO_RPC_METHOD["desktop:showNotification"];
+    if (showMethod !== "desktopShowNotification") {
+      throw new Error(
+        "Desktop notification IPC does not resolve to the show handler",
+      );
+    }
 
     await expect(
-      handlers.desktopShowNotification({
+      handlers[showMethod]({
         title: "Build finished",
         body: "The desktop build completed.",
         urgency: "critical",
@@ -224,7 +224,7 @@ describe("notification RPC handlers", () => {
       }),
     ).resolves.toMatchObject({ id: "notification_1" });
     await expect(
-      handlers.desktopShowNotification({
+      handlers[showMethod]({
         title: "Low priority",
         body: "No sound.",
         urgency: "low",
@@ -288,29 +288,6 @@ describe("notification RPC handlers", () => {
       recursive: true,
     });
     expect(fileSystem.writeFileSync).toHaveBeenCalledTimes(1);
-  });
-
-  it("keeps renderer notification bridge strings resolvable in the schema", () => {
-    expect(CHANNEL_TO_RPC_METHOD["desktop:showNotification"]).toBe(
-      "desktopShowNotification",
-    );
-
-    for (const relative of [
-      "packages/ui/src/state/notifications/notification-store.ts",
-      "packages/ui/src/state/useChatLifecycle.ts",
-      "packages/app-core/src/runtime/desktop/DesktopTrayRuntime.tsx",
-    ]) {
-      const source = readFileSync(path.join(REPO_ROOT, relative), "utf8");
-      expect(source, relative).toContain(
-        'rpcMethod: "desktopShowNotification"',
-      );
-      expect(source, relative).toContain(
-        'ipcChannel: "desktop:showNotification"',
-      );
-      expect(CHANNEL_TO_RPC_METHOD["desktop:showNotification"]).toBe(
-        "desktopShowNotification",
-      );
-    }
   });
 });
 


ss251 (approved):
Reviewed head c9180bcefb7960d1fead9aea69db48935e294161.

The risky change is the deletion of the schema-string test's source-file sweep at rpc-handlers.test.ts:292-312. The replacement guards only the in-process mapping: `CHANNEL_TO_RPC_METHOD["desktop:showNotification"] !== "desktopShowNotification"` throws, then `handlers[showMethod]` exercises real IPC routing — strictly better than the old string-echo assertions, which repeated the literal three times and would pass if every consumer dropped the annotation. What the deleted sweep uniquely covered is whether notification-store.ts, useChatLifecycle.ts and DesktopTrayRuntime.tsx still carry `rpcMethod`/`ipcChannel` literals; nothing in this diff proves those still resolve. Non-blocking: the closeNotification path still asserts via direct `handlers.desktopShowNotification`-adjacent calls rather than `handlers[showMethod]`, so only the show leg gains the indirection.

Recommendation: merge

— [free-9router-lane-ss251]

attentionhead (approved):
Reviewed at `c9180bcefb7960d1fead9aea69db48935e294161` (merge-base `963d8ef787190680b1dbe86f0b6293efd127491f`, two-dot diff: 1 file, +9/−32 — test-only, no production source touched).

## The premise is real, and the deleted test is genuinely stale

The removed `keeps renderer notification bridge strings resolvable in the schema` test read three renderer files off disk and asserted each contained the literals `rpcMethod: "desktopShowNotification"` and `ipcChannel: "desktop:showNotification"`. At the merge-base, two of those three files no longer contain the string at all:

| file the deleted test grepped | occurrences of `desktopShowNotification` |
|---|---|
| `packages/ui/src/state/notifications/notification-store.ts` | **0** |
| `packages/ui/src/state/useChatLifecycle.ts` | **0** |
| `packages/app-core/src/runtime/desktop/DesktopTrayRuntime.tsx` | 1 |

The call moved into `packages/ui/src/bridge/notification-delivery.ts:22`, where `deliverSystemNotification` issues the single `invokeDesktopBridgeRequest({ rpcMethod: "desktopShowNotification", ipcChannel: "desktop:showNotification", ... })` with the urgency mapping. So the test was pinning where a string used to live, not what the system does — the exact failure mode a source-inventory assertion has. Deleting it is right.

## Bidirectional control

Run from `packages/app-core/platforms/electrobun` with `npx vitest run --config vitest.electrobun.config.ts src/rpc-handlers.test.ts` (note this file is excluded from app-core's own config by its `platforms/electrobun/**` exclude, so the electrobun config is the one that reaches it):

| arm | result |
|---|---|
| merge-base copy of the test file | **1 failed \| 11 passed (12)** — `AssertionError: packages/ui/src/state/notifications/notification-store.ts: expected '/**\n * Client store for agent notifi…' to contain 'rpcMethod: "desktopShowNotification"'` |
| head | **11 passed (11)** |
| head, with `rpc-schema.ts:2820` mutated to `"desktop:showNotification": "desktopCloseNotification"` | **1 failed \| 10 passed (11)** — `Error: Desktop notification IPC does not resolve to the show handler` |

The third arm is the one that matters: it shows the replacement is not merely a rename. Routing the show channel to the close handler now fails at the routing boundary, and it fails in the *repaired* test — so the coverage the deleted test claimed to provide is genuinely carried by the new one rather than dropped. That reproduces the PR body's `1 fail, 10 pass` claim exactly.

- **The replacement is behavioral, not an inventory.** `handlers[showMethod](...)` resolves the method through `CHANNEL_TO_RPC_METHOD` and then *invokes* it, asserting the returned notification ids and both the critical and low-urgency param shapes. The channel→method identity that the deleted test asserted as a standalone `expect` is preserved as a precondition guard on the behavior test, which is strictly stronger: it can no longer pass vacuously.
- **The guard fails closed on both mutation shapes.** A wrong-but-existing method trips the explicit `throw`; a deleted map entry would too, before `handlers[undefined]` could produce a less legible `TypeError`.
- **System under test is real.** The handlers come from the real `buildNotificationRpcHandlers` slice and the real `CHANNEL_TO_RPC_METHOD`; only `desktop.showNotification`/`closeNotification` are `vi.fn` — the OS notification surface, a genuine external boundary a unit test cannot own. No new mocks are introduced; the fixture is the pre-existing one.
- **The renderer side keeps real coverage.** `packages/ui/src/state/notifications/notification-store.test.ts:287` asserts the store's delivery path actually calls `invokeDesktopBridgeRequest` with `rpcMethod: "desktopShowNotification"` and the mapped `{ urgency: "normal", silent: false }` params. That is a behavioral assertion on the same contract the grep test was reaching for, so the PR body's "existing renderer delivery tests already exercise the shared bridge" checks out.
- **Net effect on the file:** three now-unused `node:fs` / `node:path` / `node:url` imports and the `REPO_ROOT` computation go away with the test — a test file in a desktop package no longer reads repository source off disk at all.

## Non-blocking note

"Now live in a shared delivery helper" is true for the store path, but two call sites still inline the pair themselves rather than going through `deliverSystemNotification`: `packages/ui/src/components/settings/useDesktopWindowControls.ts:73` and `packages/app-core/src/runtime/desktop/DesktopTrayRuntime.tsx:253` (the one file of the three where the old grep would still have passed). Nothing here regresses because of that, and the schema-routing guard covers all of them equally — but routing those two through the helper would make the helper the single owner the PR description already describes.

Approving: the deleted assertion was stale by construction, the replacement pins the same contract behaviorally, and the mutation arm proves it detects a real routing break.

lalalune:
Fixed and merged through #31046. Final candidate `d204905bd6a3ce55a2adfd7c287819849100dbe9` passed all five hosted checks in [run 34542077711](https://github.com/elizaOS/eliza/actions/runs/34542077711), normal installation, all 373 root verification tasks and final audits, 1,492 desktop package tests, and 96 renderer/store tests.

Merged at `ae1a3aa39c68610a2ab92720295b6a880717453f`; readback confirms the merged notification test blob exactly matches the tested candidate. [Current evidence archive](https://github.com/elizaOS/eliza/releases/download/pr-evidence-12/31046-bettina-31046-d204-evidence.zip) retains source provenance and all current local logs. The original negative-control archive remains linked in the PR.

This repairs the notification contract test without changing native notification implementation. It is not physical-device delivery or full Bettina acceptance.

PR #31047, opened 2026-09-10
test(cloud): enable cache explicitly in cache-dependent fixtures

### LINKED ISSUE #31044: fix(cloud-test): enable memory cache in cache-dependent fixtures
The Cloud Tests workflow sets `CACHE_ENABLED=false` globally. Two suites explicitly exercise the shipped in-memory cache but never opt back in, so Develop Full `34530365498` fails Cloud shards 2 and 3: affiliate balance-hint publication reports `unavailable`, and the cacheable RPC test performs two provider requests instead of one.

Both failures reproduce locally at develop `963d8ef787190680b1dbe86f0b6293efd127491f` with `CACHE_ENABLED=false`. The cache correctly honors the disabled setting; this is a fixture-environment defect, not a reason to weaken production cache-disabled behavior or change the workflow-wide default.

Scope and acceptance:
- Explicitly enable the real memory cache for these two fixtures and restore the prior environment afterward.
- Keep real local HTTP/PGlite behavior and all existing assertions intact.
- The unchanged tests pass under the workflow's disabled-cache environment and normal local environment.
- Complete package and root checks and exact-head hosted validation before merge.

Taking this bounded release-validation dependency for Bettina/Cloud #30130. No provider operation, payment, deployment, or production configuration change is included.

lalalune:
Two real-cache fixtures fail in Cloud CI because the workflow sets `CACHE_ENABLED=false`. The RPC transport test then observes two provider executions; the affiliate payout outbox test cannot acquire its balance hint. Each fixture now explicitly enables its in-memory cache and restores the previous setting during teardown.

Fixes #31044. Supports Bettina Cloud validation under #30130.

Candidate `a5cc1eb8df56ea9a245f4438caeb8a7369746325`, based on `c745213d6100bc941e0a78f2c6105462b2e5925e`. Normal install and root verification (373 tasks plus final audits) pass with Bun 1.3.14 and Node 24.15.0. Existing production cache behavior, assertions and workflow settings remain unchanged.

Both original failures reproduced with the hosted cache-disabled environment. After the repair all 27 focused tests pass in normal and CI environments. On the previous base (963d8ef), the complete shared suite under `CACHE_ENABLED=false CACHE_BACKEND=wadis` passes: 1,231 files, 388 batches, 12,637 tests, zero failures and 243 existing skips; exit 0. Tests exercise the real memory cache, loopback HTTP transport and PGlite transactional payout outbox with controlled external-provider boundaries.

# Evidence Gate

N/A - test fixture environment repair; no rendered UI, native, model or production behavior changes.

N/A - test fixture environment repair; no rendered UI, native, model or production behavior changes.

N/A - test fixture environment repair; no rendered UI, native, model or production behavior changes.

- [x] [31047-bettina-31047-c745-focused.log](https://github.com/elizaOS/eliza/releases/download/pr-evidence-12/31047-bettina-31047-c745-focused.log)

N/A - test fixture environment repair; no rendered UI, native, model or production behavior changes.

N/A - test fixture environment repair; no rendered UI, native, model or production behavior changes.

- [ ] N/A - fixture configuration only; real HTTP and PGlite assertions are recorded in the focused log.

N/A - test fixture environment repair; no rendered UI, native, model or production behavior changes.

## Known gaps / failures

Hosted checks must pass on this exact head before merge. Other Develop Full failures are separate. This repairs test setup and does not deploy or change Cloud resources.

[Reviewed evidence archive](https://github.com/elizaOS/eliza/releases/download/pr-evidence-12/31047-bettina-31044-1f951-evidence.zip) contains both baseline failures, focused tests in both environments, full shared-package logs, installation, root verification, source provenance and a SHA-256 manifest. [Complete shared suite log](https://github.com/elizaOS/eliza/releases/download/pr-evidence-12/31047-bettina-31044-shared.log).

Current-base qualification: normal installation, all 373 root tasks and final audits, and all 27 cache-dependent tests pass after rebasing over #30229. The complete shared suite also passed on this candidate: 1,232 files across 388 batches, 12,670 passing tests, zero failures and 243 existing skips; exit 0. The previous archive and full-suite log remain historical evidence.

[Current a5cc evidence archive](https://github.com/elizaOS/eliza/releases/download/pr-evidence-12/31047-bettina-31047-a5cc-evidence.zip) includes fresh installation, repository verification, focused and full shared-suite logs, source provenance and a SHA-256 manifest. Archive SHA-256: `b06a626c7cda451fc669f7a2830584b643c4b822ea3164703432303aaadc4800`. All required local gates pass; hosted checks remain required.

### DIFF
--- a/packages/cloud/shared/src/lib/services/__tests__/affiliate-payout-outbox.test.ts
+++ b/packages/cloud/shared/src/lib/services/__tests__/affiliate-payout-outbox.test.ts
@@ -6,6 +6,8 @@
 
 import { afterAll, beforeAll, describe, expect, mock, spyOn, test } from "bun:test";
 
+const originalCacheEnabled = process.env.CACHE_ENABLED;
+process.env.CACHE_ENABLED = "true";
 process.env.DATABASE_URL = "pglite://memory";
 process.env.TEST_DATABASE_URL = "pglite://memory";
 process.env.NODE_ENV ||= "test";
@@ -140,7 +142,12 @@ beforeAll(async () => {
 }, PGLITE_TIMEOUT);
 
 afterAll(async () => {
-  await closeDatabaseConnectionsForTests();
+  try {
+    await closeDatabaseConnectionsForTests();
+  } finally {
+    if (originalCacheEnabled === undefined) delete process.env.CACHE_ENABLED;
+    else process.env.CACHE_ENABLED = originalCacheEnabled;
+  }
 });
 
 describe("affiliate payout outbox", () => {
--- a/packages/cloud/shared/src/lib/services/proxy/rpc-cache.transport.test.ts
+++ b/packages/cloud/shared/src/lib/services/proxy/rpc-cache.transport.test.ts
@@ -44,10 +44,16 @@ mock.module("../usage", () => ({
 const { createHandler } = await import("./engine");
 const { rpcConfigForChain, rpcHandlerForChain } = await import("./services/rpc");
 const originalFetch = globalThis.fetch;
-const envNames = ["CACHE_BACKEND", "ALCHEMY_API_KEY", "ALCHEMY_MAX_RETRIES"] as const;
+const envNames = [
+  "CACHE_ENABLED",
+  "CACHE_BACKEND",
+  "ALCHEMY_API_KEY",
+  "ALCHEMY_MAX_RETRIES",
+] as const;
 const originalEnv = Object.fromEntries(envNames.map((name) => [name, process.env[name]]));
 
 beforeEach(() => {
+  process.env.CACHE_ENABLED = "true";
   process.env.CACHE_BACKEND = "memory";
   process.env.ALCHEMY_API_KEY = "local-cache-fixture";
   process.env.ALCHEMY_MAX_RETRIES = "1";


ss251 (approved):
Reviewed head 1f9514f43c4e0509ce0e046f5dbf36a0513699f1.

Verified the risky part: the affiliate-payout-outbox fixture sets `CACHE_ENABLED` at module top level (before `beforeAll` registers), so any lazy env read at import time of the service under test now sees `true`, and the `afterAll` restore handles both the `undefined` and set cases correctly via `delete` vs reassign. In rpc-cache.transport.test.ts, `CACHE_ENABLED` is added to `envNames`, so the existing `originalEnv` snapshot/restore path (used in the file's afterEach/afterAll) covers it without further edits — no leak into sibling test files. Both edits are test-scoped and gated on env restoration, not production code paths.

Non-blocking nit: the affiliate fixture restore only runs if `afterAll` is reached; a module-load crash would leave `CACHE_ENABLED=true` in the worker, harmless for test isolation in practice.

Recommendation: merge

— [free-9router-lane-ss251]

attentionhead (approved):
Reviewed at `1f9514f43c4e0509ce0e046f5dbf36a0513699f1` (merge-base `963d8ef787190680b1dbe86f0b6293efd127491f`, two-dot diff: 2 files, +15/−2 — test setup and teardown only, no assertions and no production source touched).

## The premise checks out

`.github/workflows/cloud-tests.yml:19` sets `CACHE_ENABLED: "false"` at workflow scope, and `packages/cloud/shared/src/lib/cache/client.ts:218` reads it as `this.enabled = env.CACHE_ENABLED !== "false"`. So in hosted CI — and only there — both fixtures run against a cache client that returns early at line 233 and never stores anything. Locally, with the var unset, they get a real cache and pass. That is exactly the shape of a green-locally/red-in-CI fixture bug.

## Bidirectional control

I ran both fixture files at both revisions under both environments, from `packages/cloud/shared` with `bun test <paths>`:

| | merge-base `963d8ef` | head `1f9514f` |
|---|---|---|
| **default env** (`CACHE_ENABLED` unset) | 27 pass / 0 fail | 27 pass / 0 fail |
| **CI env** (`CACHE_ENABLED=false`) | **25 pass / 2 fail** | **27 pass / 0 fail** |

The two base failures under the CI env are precisely the two the PR body names, with the two distinct symptoms it describes:

- `eth_getBalance preserves its provider execution contract when cache is requested` — `expect(received).toBe(expected)` on the provider-execution count (the second call is not served from cache, so the provider runs twice).
- `affiliate payout outbox > fenced affiliate settlement stays warm and lowers before its authoritative republish` — `Organization balance hint write was not confirmed: unavailable`.

The top-left cell is what makes this a control rather than an anecdote: at the merge-base the default environment is already green, so the change is scoped strictly to the CI environment and cannot be masking a real regression under normal settings.

- **System under test is real, and stays real.** The diff changes no assertion, no mock, and no fixture data — only which environment the existing tests run in. The RPC test still drives the real `createHandler` engine over loopback HTTP, and the outbox test still runs against real PGlite transactions. What is stubbed (Alchemy via `ALCHEMY_API_KEY = "local-cache-fixture"`, the pre-existing `mock.module` calls for `../credits`, `../usage`, `../../auth`) stands for an external provider boundary and was already there at the merge-base — nothing new is mocked.
- **Restoration is symmetric in both files.** `rpc-cache.transport.test.ts` adds `CACHE_ENABLED` to the existing `envNames` tuple, so it is captured into `originalEnv` before mutation and restored-or-deleted by the same `afterAll` loop that already handles the other three. `affiliate-payout-outbox.test.ts` captures `originalCacheEnabled` before assigning, and restores in a `finally` so the restore survives a throwing `closeDatabaseConnectionsForTests()`. Correct on both counts.
- **The asymmetric placement is justified, not an inconsistency.** The outbox file sets `CACHE_ENABLED` at module top level while the RPC file sets it in `beforeEach`. That is load-bearing: `CacheClient.initialize()` is guarded by `if (this.initialized) return` on a module-level singleton, so the flag has to be set before anything triggers first initialization — a `beforeAll` in the outbox file could already be too late. The RPC file can use `beforeEach` because it re-installs the cache module through `mock.module`.
- **This is the repo's own convention.** `src/lib/cache/__tests__/client-mock.test.ts` and `src/lib/cache/client-swr.test.ts` already save/set/restore `CACHE_ENABLED` the same way; the PR follows the established pattern rather than inventing one.
- **Disabled-path coverage is not lost.** My one concern with "just turn the cache on" was that it might leave `CACHE_ENABLED=false` untested. It does not: `client-mock.test.ts:68` explicitly sets `CACHE_ENABLED = "false"` and exercises the disabled branch. These two fixtures are about cache *behavior*, so they genuinely require a cache; the disabled path is pinned where it belongs.

## Non-blocking note

Both files mutate process-wide `process.env`, and the outbox file holds `CACHE_ENABLED=true` from module evaluation through `afterAll`. Under the package's `bun test --isolate` runner files execute sequentially, so the window cannot overlap another suite today — but it is a process-global with a file-scoped lifetime, and that only stays safe while the runner does not parallelize files. Worth remembering if `--isolate` is ever traded for concurrency.

Approving: the diagnosis is correct, the fix is the minimal one, restoration is clean in both files, and the four-cell control shows it repairs the CI environment without changing behavior anywhere else.

PR #31050, opened 2026-09-11
fix(plugin-form): publish registered control types to the handler registry and keep date answers on the named calendar day

### LINKED ISSUE #31049: fix(plugin-form): registered control-type validate/parse/format/extractionPrompt never run for top-level fields; date answers shift a day east of UTC
## Defect

`@elizaos/plugin-form` keeps two type registries that never meet. `FormService.registerControlType()` (the documented API, used for the seven builtins and by consuming plugins per `README.md` / `CLAUDE.md`) stores a `ControlType` in `service.controlTypes`. Extraction, parsing, validation, and display read a different map: `typeHandlers` in `src/validation.ts`, populated only by `registerTypeHandler()`, which nothing in the plugin ever calls (`git grep registerTypeHandler\( plugins/plugin-form/src` → only its definition). So for every top-level field on `develop` `619545c5357f`:

- `ControlType.validate` never runs (`validateField` falls through to the built-in switch; a custom simple type such as the README's `phone` gets plain text validation).
- `ControlType.parse` never runs (`parseValue` uses the switch).
- `ControlType.format` never runs.
- `ControlType.extractionPrompt` never reaches the model (`buildFormExtractorPromptSection` and `extractSingleField` fall back to the bare type id, so the planner is told the field is a `date`, not "a date (preferably in YYYY-MM-DD format)").

`ControlType.validate` is consulted only for subcontrols of composite types (`service.ts` line 829).

### User-visible symptom: `date` answers shift by one calendar day

Because the builtin `date` type's own `parse`/`validate` are dead, `parseValue()`'s switch branch handles dates: `new Date(Date.parse(value)).toISOString()`. A natural-language answer (which is what the model returns when its only hint is `date`) is parsed as **host-local midnight** and then serialized in **UTC**, and `formatValue()` echoes that raw string back.

Driving the real `parseValue` → `validateField` → `formatValue` path on `619545c5357f`:

```text
TZ=Asia/Tokyo    "September 15, 2026" → "2026-09-14T15:00:00.000Z"  valid=true  display "2026-09-14T15:00:00.000Z"
TZ=Europe/Berlin "9/15/2026"          → "2026-09-14T22:00:00.000Z"  valid=true  display "2026-09-14T22:00:00.000Z"
TZ=UTC / LA      "September 15, 2026" → date part stays the 15th
any TZ           "Sept 15"            → "2001-09-14T15:00:00.000Z" (V8 default year) valid=true
```

Every host east of UTC stores and confirms the previous day for the user's stated date, and a year-less answer is accepted as 2001 instead of being re-asked. The builtin's own contract ("Must be in YYYY-MM-DD format") is never applied. The display half has the mirror defect west of UTC: the builtin `format` (`new Date("2026-09-15").toLocaleDateString()`) parses a date-only string as UTC midnight and prints the local previous day (`9/14/2026` in America/Los_Angeles).

## Reproduction

```bash
cat > /tmp/probe.ts <<'TS'
import { parseValue, validateField, formatValue } from "./plugins/plugin-form/src/validation.ts";
const control = { key: "dueDate", label: "Due date", type: "date" } as any;
for (const input of ["2026-09-15", "September 15, 2026", "9/15/2026", "Sept 15"]) {
  const parsed = parseValue(input, control);
  console.log(process.env.TZ, input, parsed, validateField(parsed, control).valid, formatValue(parsed, control));
}
TS
TZ=Asia/Tokyo bun /tmp/probe.ts
```

## Required scope

1. Bridge the registries: a `ControlType` registered through `FormService.registerControlType` (builtin or custom) must supply `validate`, `parse`, `format`, and `extractionPrompt` to extraction, parsing, validation, and display for top-level fields, matching the documented contract.
2. Make the `date` builtin produce and display a calendar date independent of the host timezone: ISO `YYYY-MM-DD` passes through; other parseable formats resolve to the calendar date the user named; year-less or unparseable input is rejected so the form re-asks; display renders the named day on every host.
3. Regression tests that fail on the current code: the custom-type validator/prompt path, and the date path under an explicit east-of-UTC and west-of-UTC `TZ`.

## Acceptance

- `validateField`, `parseValue`, `formatValue`, and both extraction prompt builders honour a `ControlType` registered via `FormService.registerControlType`.
- `parseValue("September 15, 2026", dateControl)` is `"2026-09-15"` under `TZ=Asia/Tokyo`, `TZ=Europe/Berlin`, `TZ=America/Los_Angeles`, and `TZ=UTC`; `"Sept 15"` is rejected.
- `formatValue("2026-09-15", dateControl)` names the 15th under `TZ=America/Los_Angeles`.
- plugin-form suite, typecheck, Biome, and root `bun run verify` green.

hermesagent270-commits:
# Relates to

Closes #31049

# Risks

Low-to-moderate, confined to `@elizaos/plugin-form`. Two behavior changes for top-level fields:

- A **custom** control type registered through `FormService.registerControlType` now has its `validate`, `parse`, `format`, and `extractionPrompt` applied, as the README and CLAUDE.md have documented all along. A consumer whose custom validator was silently never running will now see it enforced.
- The **date** builtin stores `YYYY-MM-DD` instead of a full UTC instant, rejects year-less or free-form answers so the form re-asks, and validates strings as calendar days. A session persisted with the old timestamp shape re-asks that one field at submit-time revalidation.

Builtins other than `date` are unaffected: they publish only their extraction prompt, and their parsing/validation/display stay on the existing hardened switch paths (pinned by a test, because the builtin number `parse` is the lenient `parseFloat` the switch deliberately replaced).

# Background

## What does this PR do?

`FormService.registerControlType()` stored types in `service.controlTypes`, but `parseValue`, `validateField`, `formatValue`, and both extractor prompt builders resolve a control's type through `typeHandlers` in `validation.ts`, which nothing populated (`registerTypeHandler` had no caller). Consequences on `develop` `619545c5357f`:

- `ControlType.validate/parse/format/extractionPrompt` never ran for top-level fields (only `validate`, for subcontrols).
- The planner was told a date field is a `date`, not "a date (preferably in YYYY-MM-DD format)", so it echoes the user's format.
- The switch branch that actually handled dates did `new Date(Date.parse(value)).toISOString()`: host-local midnight re-rendered in UTC. Through the real `parseValue → validateField → formatValue` path:

```text
TZ=Asia/Tokyo    "September 15, 2026" → "2026-09-14T15:00:00.000Z"  valid  displayed verbatim
TZ=Europe/Berlin "9/15/2026"          → "2026-09-14T22:00:00.000Z"  valid  displayed verbatim
any TZ           "Sept 15"            → "2001-09-14T15:00:00.000Z"  valid  (V8 default year)
```

The fix:

- `registerControlType` publishes the type to the handler registry: a custom type in full; a builtin only its `extractionPrompt`.
- New `calendar-date.ts`: `parseCalendarDate` (ISO passes through after a real-day check; any other format must carry a four-digit year and resolves through the local components of the parsed instant, which is the day the user named on every host), `calendarDateComponents`, and `formatCalendarDate` (builds the display from components so `2026-09-15` is not shown as the 14th west of UTC).
- `parseValue`'s date branch and the builtin's `parse`/`format`/`validate` use the helper; `validateDate` requires a calendar-date string (a Date or timestamp number still works for min/max bounds).

Bug fix.

# Testing

`plugins/plugin-form/src/service.ts` (`registerControlType`), then `src/validation.ts` (`parseValue` date branch, `validateDate`, `formatValue`), then `src/calendar-date.ts`.

## Detailed testing steps

New suite `plugins/plugin-form/src/control-type-registry.test.ts` (14 cases) drives the real `FormService`, the extractor prompt builder, and the validation module, switching `process.env.TZ` per case (Asia/Tokyo, Europe/Berlin, America/Los_Angeles, Pacific/Auckland, UTC).

RED control — develop's `builtins.ts`/`service.ts`/`validation.ts` with this test file:

```text
 × publishes builtin extraction prompts to the extractor
 × runs a custom simple type's validate, parse, format, and prompt for a top-level field
 × resolves natural-language answers to the named day under TZ=Asia/Tokyo
 × resolves natural-language answers to the named day under TZ=Europe/Berlin
 × resolves natural-language answers to the named day under TZ=America/Los_Angeles
 × resolves natural-language answers to the named day under TZ=Pacific/Auckland
 × resolves natural-language answers to the named day under TZ=UTC
 × rejects a year-less answer instead of storing a fabricated year
 × rejects a rolled-over day and an instant that is not a calendar date
 Tests  9 failed | 5 passed (14)
```

(The five that pass on develop pin the hardened number path, the display cases, and min/max bounds.)

GREEN at `36287f191619`:

```text
$ bunx vitest run --config vitest.config.ts        # plugins/plugin-form
 Test Files  12 passed (12)
      Tests  162 passed (162)
$ bun run typecheck                                 # tsc --noEmit, 0 errors
$ bunx @biomejs/biome check <5 changed files>       # clean
```

Root `bun run verify` is running on this exact commit in a fully installed checkout; result follows as a comment.

Not run: a live-model form session. The change is in deterministic parse/validate/format/prompt-assembly code; the new suite exercises the real service and prompt builder rather than a mock, and the model-facing change is only the type hint text now reaching the prompt.

https://claude.ai/code/session_01GnpYyWkEf3naaRGNtRmwNp

### DIFF
--- a/plugins/plugin-form/src/builtins.ts
+++ b/plugins/plugin-form/src/builtins.ts
@@ -63,6 +63,11 @@
  */
 
 import type { JsonValue } from "@elizaos/core";
+import {
+  calendarDateComponents,
+  formatCalendarDate,
+  parseCalendarDate,
+} from "./calendar-date";
 import { basicEmailValid } from "./email";
 import type { ControlType, FormControl, ValidationResult } from "./types";
 import { testControlPattern } from "./validation";
@@ -381,7 +386,8 @@ const selectType: ControlType = {
  * WHY flexible parsing:
  * - Users say "tomorrow", "next Monday", "12/25/2024"
  * - LLM should normalize to ISO
- * - We accept anything Date() can parse as fallback
+ * - Other formats resolve to the calendar day they name on every host
+ *   timezone; a year-less or unparseable answer is rejected and re-asked
  *
  * WHY locale display:
  * - Agent should speak in user's date format
@@ -403,29 +409,22 @@ const dateType: ControlType = {
       return { valid: false, error: "Must be in YYYY-MM-DD format" };
     }
 
-    // Check if it's a valid date
-    const date = new Date(str);
-    if (Number.isNaN(date.getTime())) {
+    // Check if it's a real calendar day (rejects roll-overs like 2026-02-30)
+    if (!calendarDateComponents(str)) {
       return { valid: false, error: "Invalid date" };
     }
 
     return { valid: true };
   },
 
-  parse: (value: string): string => {
-    // Try to parse and normalize to ISO
-    const date = new Date(value);
-    if (!Number.isNaN(date.getTime())) {
-      return date.toISOString().split("T")[0];
-    }
-    return value.trim();
-  },
+  // Resolve the answer to the calendar day it names; an unparseable or
+  // year-less answer is kept verbatim so validate() rejects it and the form
+  // re-asks instead of storing a fabricated day.
+  parse: (value: string): string => parseCalendarDate(value) ?? value.trim(),
 
   format: (value: JsonValue): string => {
     if (!value) return "";
-    const date = new Date(String(value));
-    if (Number.isNaN(date.getTime())) return String(value);
-    return date.toLocaleDateString();
+    return formatCalendarDate(String(value));
   },
 
   extractionPrompt: "a date (preferably in YYYY-MM-DD format)",
--- a/plugins/plugin-form/src/calendar-date.ts
+++ b/plugins/plugin-form/src/calendar-date.ts
@@ -1,0 +1,68 @@
+/**
+ * Host-timezone-independent calendar-date parsing and display for the `date`
+ * control type. A form answer names a calendar day, not an instant, so the
+ * stored value is always `YYYY-MM-DD` and neither parsing nor display may pass
+ * through a UTC/local conversion that can move the day: `Date.parse` reads
+ * "September 15, 2026" as local midnight and `toISOString` re-renders it in
+ * UTC (the 14th east of UTC), while `new Date("2026-09-15")` is UTC midnight and
+ * `toLocaleDateString` renders it as the 14th west of UTC.
+ */
+
+const ISO_CALENDAR_DATE_RE = /^(\d{4})-(\d{2})-(\d{2})$/;
+const FOUR_DIGIT_YEAR_RE = /\b\d{4}\b/;
+
+/** Calendar components of a `YYYY-MM-DD` string, or null when it is not a real day. */
+export function calendarDateComponents(
+  value: string,
+): { year: number; month: number; day: number } | null {
+  const match = ISO_CALENDAR_DATE_RE.exec(value);
+  if (!match) return null;
+  const year = Number(match[1]);
+  const month = Number(match[2]);
+  const day = Number(match[3]);
+  // Round-trip through UTC to reject roll-overs such as 2026-02-30 → Mar 2.
+  const probe = new Date(Date.UTC(year, month - 1, day));
+  if (
+    probe.getUTCFullYear() !== year ||
+    probe.getUTCMonth() !== month - 1 ||
+    probe.getUTCDate() !== day
+  ) {
+    return null;
+  }
+  return { year, month, day };
+}
+
+function pad2(value: number): string {
+  return String(value).padStart(2, "0");
+}
+
+/**
+ * Resolve an extracted answer to the calendar day it names, as `YYYY-MM-DD`.
+ * ISO input is validated and returned as-is. Any other format must carry a
+ * four-digit year (V8 silently defaults a year-less date to 2001) and is read
+ * back through the local calendar components of the parsed instant, which is
+ * the day the user named on every host timezone. Returns null when the input
+ * does not name a real day, so the caller leaves the raw answer for validation
+ * to reject and the form re-asks.
+ */
+export function parseCalendarDate(value: string): string | null {
+  const trimmed = value.trim();
+  if (ISO_CALENDAR_DATE_RE.test(trimmed)) {
+    return calendarDateComponents(trimmed) ? trimmed : null;
+  }
+  if (!FOUR_DIGIT_YEAR_RE.test(trimmed)) return null;
+  const parsed = new Date(trimmed);
+  if (Number.isNaN(parsed.getTime())) return null;
+  const candidate = `${parsed.getFullYear()}-${pad2(parsed.getMonth() + 1)}-${pad2(parsed.getDate())}`;
+  return calendarDateComponents(candidate) ? candidate : null;
+}
+
+/**
+ * Locale display of a stored `YYYY-MM-DD` value that names the same day on
+ * every host. Anything that is not a calendar date is echoed unchanged.
+ */
+export function formatCalendarDate(value: string): string {
+  const parts = calendarDateComponents(value.trim());
+  if (!parts) return value;
+  return new Date(parts.year, parts.month - 1, parts.day).toLocaleDateString();
+}
--- a/plugins/plugin-form/src/control-type-registry.test.ts
+++ b/plugins/plugin-form/src/control-type-registry.test.ts
@@ -121,6 +121,45 @@ describe("control type registry bridge (#31049)", () => {
     expect(parseValue("+1 (555) 123-4567", phoneControl)).toBe("+15551234567");
     expect(formatValue("+15551234567", phoneControl)).toBe("tel:+15551234567");
   });
+
+  it("keeps a custom allowOverride contract when another agent's FormService starts", async () => {
+    // The handler registry is process-global while each service's control
+    // map is per instance. A second runtime in the same process boots its own
+    // FormService, whose builtins pass its (empty) instance guard; they must
+    // not overwrite the override the first agent installed.
+    const emailControl: FormControl = {
+      key: "email",
+      label: "Email",
+      type: "email",
+    };
+    service.registerControlType(
+      {
+        id: "email",
+        validate: (value: JsonValue) =>
+          String(value).endsWith("@corp.example")
+            ? { valid: true }
+            : { valid: false, error: "corporate domains only" },
+        extractionPrompt: "a corporate email",
+      },
+      { allowOverride: true },
+    );
+    expect(validateField("x@gmail.com", emailControl)).toEqual({
+      valid: false,
+      error: "corporate domains only",
+    });
+
+    await FormService.start(makeRuntime());
+
+    expect(validateField("x@gmail.com", emailControl)).toEqual({
+      valid: false,
+      error: "corporate domains only",
+    });
+    expect(promptFor([emailControl])).toContain("a corporate email");
+    // A builtin the first agent left alone is still published for the second.
+    expect(getTypeHandler("date")?.extractionPrompt).toBe(
+      "a date (preferably in YYYY-MM-DD format)",
+    );
+  });
 });
 
 describe("date control names the same calendar day on every host (#31049)", () => {
--- a/plugins/plugin-form/src/control-type-registry.test.ts
+++ b/plugins/plugin-form/src/control-type-registry.test.ts
@@ -1,0 +1,196 @@
+/**
+ * Regression coverage for the control-type registry bridge (#31049): a type
+ * registered through FormService.registerControlType must reach extraction,
+ * parsing, validation, and display for top-level fields, and the builtin
+ * `date` type must name the same calendar day on every host timezone. Runs
+ * the real FormService, extractor prompt builder, and validation module with a
+ * deterministic in-memory runtime; the host timezone is switched per case.
+ */
+import type { Component, IAgentRuntime, JsonValue, UUID } from "@elizaos/core";
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+import { buildFormExtractorPromptSection } from "./extraction";
+import { FormService } from "./service";
+import type { FormControl, FormDefinition } from "./types";
+import {
+  clearTypeHandlers,
+  formatValue,
+  getTypeHandler,
+  parseValue,
+  validateField,
+} from "./validation";
+
+const agentId = "00000000-0000-4000-8000-000000000203" as UUID;
+
+function makeRuntime(): IAgentRuntime {
+  const components = new Map<string, Component>();
+  return {
+    agentId,
+    getRoom: vi.fn(async () => ({ id: agentId, worldId: agentId })),
+    getComponent: vi.fn(async (entity: UUID, type: string) =>
+      components.get(`${entity}:${type}`),
+    ),
+    getComponents: vi.fn(async (entity: UUID) =>
+      Array.from(components.values()).filter((c) => c.entityId === entity),
+    ),
+    createComponent: vi.fn(async (component: Component) => {
+      components.set(`${component.entityId}:${component.type}`, component);
+    }),
+    updateComponent: vi.fn(async (component: Component) => {
+      components.set(`${component.entityId}:${component.type}`, component);
+    }),
+    deleteComponent: vi.fn(async () => undefined),
+    emitEvent: vi.fn(async () => undefined),
+    registerTaskWorker: vi.fn(),
+    logger: { debug: vi.fn(), error: vi.fn(), warn: vi.fn(), info: vi.fn() },
+  } as unknown as IAgentRuntime;
+}
+
+const dateControl: FormControl = {
+  key: "dueDate",
+  label: "Due date",
+  type: "date",
+};
+const numberControl: FormControl = { key: "age", label: "Age", type: "number" };
+const phoneControl: FormControl = {
+  key: "phone",
+  label: "Phone",
+  type: "phone",
+};
+
+function promptFor(controls: FormControl[]): string {
+  const form: FormDefinition = { id: "f", name: "F", controls };
+  return buildFormExtractorPromptSection({
+    text: "hello",
+    form,
+    controls,
+  });
+}
+
+describe("control type registry bridge (#31049)", () => {
+  let service: FormService;
+
+  beforeEach(async () => {
+    clearTypeHandlers();
+    service = (await FormService.start(makeRuntime())) as FormService;
+  });
+
+  afterEach(() => {
+    clearTypeHandlers();
+  });
+
+  it("publishes builtin extraction prompts to the extractor", () => {
+    expect(getTypeHandler("date")?.extractionPrompt).toBe(
+      "a date (preferably in YYYY-MM-DD format)",
+    );
+    const prompt = promptFor([dateControl]);
+    expect(prompt).toContain("a date (preferably in YYYY-MM-DD format)");
+    expect(prompt).not.toContain('"type": "date"');
+  });
+
+  it("keeps builtin parsing and validation on the hardened switch paths", () => {
+    // The builtin number ControlType parses with a lenient parseFloat; the
+    // switch in parseValue preserves a rejected answer verbatim so it stays
+    // invalid through persistence. Publishing the builtin's parse would
+    // silently reintroduce coercion of garbage-suffixed input.
+    expect(getTypeHandler("number")?.parse).toBeUndefined();
+    expect(getTypeHandler("number")?.validate).toBeUndefined();
+    expect(parseValue("12abc", numberControl)).toBe("12abc");
+    expect(validateField("12abc", numberControl).valid).toBe(false);
+  });
+
+  it("runs a custom simple type's validate, parse, format, and prompt for a top-level field", () => {
+    service.registerControlType({
+      id: "phone",
+      validate: (value: JsonValue) =>
+        /^\+?[\d\s-]{10,}$/.test(String(value))
+          ? { valid: true }
+          : { valid: false, error: "Invalid phone" },
+      parse: (value: string) => value.replace(/[^\d+]/g, ""),
+      format: (value: JsonValue) => `tel:${String(value)}`,
+      extractionPrompt: "a phone number with country code",
+    });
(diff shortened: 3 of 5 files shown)


ss251 (approved):
Reviewed head 36287f1916190b5e6332ec60c8e3bf610f370dda.

I followed the string path through `validation.ts:668` (`parseValue` date case) into `calendar-date.ts:47` (`parseCalendarDate`). The rollover guard is asymmetric: the ISO branch round-trips through `calendarDateComponents`, so `2026-02-30` is rejected, but the non-ISO branch round-trips only the *candidate* `parseCalendarDate("2/30/2026")` builds — V8's `Date` already rolled it to March, so the roundtrip passes and the stored answer is `2026-03-02`, which `validateField` then accepts. Non-blocking, since the pre-PR code had the same behavior and the day-shift fix itself is sound: I confirmed `formatCalendarDate` constructs the display from local components, and `validateDate`'s new string branch (`validation.ts:449`) rejects anything not already a real `YYYY-MM-DD`, closing the fabricated-instant path.

Recommendation: merge

— [free-9router-lane-ss251]

attentionhead (changes requested):
Reviewed at head `36287f1916190b5e6332ec60c8e3bf610f370dda` (merge-base `619545c5357f92e807e90639d6a3ceef22f8ea09`, matching the body).

## What is right, and verified

**The root cause is real and correctly diagnosed.** `typeHandlers` in `validation.ts:88` is the only registry `validateField` (`:179`), `parseValue` (`:643`) and `formatValue` (`:703`) consult before their builtin switches, and `registerControlType` previously wrote only `this.controlTypes`. Nothing in `plugins/plugin-form/src` called `registerTypeHandler`, so a `ControlType`'s `validate`/`parse`/`format`/`extractionPrompt` genuinely never reached a top-level field. The `controlTypes` map really was serving subcontrols only.

**The builtin-vs-custom publication split is deliberately correct.** `validation.ts`'s module doc states custom handlers are checked *first*, before built-in type validation. Publishing `{ extractionPrompt }` alone for a builtin leaves `validate`/`parse`/`format` undefined, so `handler?.validate` short-circuits and execution falls through to the hardened switch — strict number parsing, persistence-safe rejection of `"12abc"`, calendar-day dates. The third test in the new suite pins exactly that, and it is the right thing to pin.

**`calendar-date.ts` is sound.** The UTC round-trip in `calendarDateComponents` rejects roll-overs (`2026-02-30` → Mar 2) rather than silently accepting them; the `FOUR_DIGIT_YEAR_RE` guard is the right fix for V8 reading `"Sept 15"` as 2001-09-15, and `formatCalendarDate` rebuilds the display from components instead of re-rendering an instant, which is what keeps the day stable across host offsets.

**Types check out.** `ControlType.validate` returns `ValidationResult` (`types.ts:875`), which is structurally `{ valid: boolean; error?: string }` — identical to `TypeHandler.validate`'s inline return. Passing a whole `ControlType` as a `TypeHandler` is signature-safe.

### Controls run

| Arm | Result |
|---|---|
| New suite at head | `14 passed (14)` |
| New suite with only `builtins.ts`, `service.ts`, `validation.ts` checked out at the merge-base | `9 failed \| 5 passed (14)` — exactly the RED the body claims |
| Full `plugins/plugin-form` suite at head | `161 passed \| 1 skipped (162)` |

`src/storage-pglite.test.ts` fails to load with `Cannot find package '@elizaos/shared/db/raw-sql'`, but it fails identically at the merge-base — pre-existing unbuilt-package breakage, not attributable here.

## The blocking finding: publishing a builtin unconditionally clobbers another agent's custom override

`registerControlType` (`service.ts:319`) guards overrides against **its own instance map**:

```ts
if (existing.builtin && !options?.allowOverride) { logger.warn(...); return; }
```

but then writes to the **module-global** registry unconditionally. A second `FormService` starts with an empty `this.controlTypes`, so every builtin passes that guard trivially and `registerTypeHandler(type.id, { extractionPrompt })` overwrites whatever is already published under that id — including a custom handler a *different* agent installed with `allowOverride: true`.

This is not hypothetical: `packages/core/src/runtime-composition.ts:416` constructs one `AgentRuntime` per character in a single process with the same resolved plugin set, so N agents means N `FormService.start()` calls against one `typeHandlers` map. `stop()` (`service.ts:217`) clears nothing, and `clearTypeHandlers()` has no non-test caller.

Reproduced with the PR's own harness (temporary probe appended to `src/control-type-registry.test.ts`, since removed):

```ts
const a = (await FormService.start(makeRuntime())) as FormService;
a.registerControlType({
  id: "email", builtin: false,
  validate: () => ({ valid: false, error: "corporate domains only" }),
  extractionPrompt: "a corporate email",
}, { allowOverride: true });

expect(validateField("x@gmail.com", emailControl))
  .toEqual({ valid: false, error: "corporate domains only" });   // PASSES — this PR's fix working

await FormService.start(makeRuntime());                           // a second agent boots

expect(validateField("x@gmail.com", emailControl))
  .toEqual({ valid: false, error: "corporate domains only" });    // FAILS: received { valid: true }
```

The first assertion passing is the PR doing its job. The second failing is new behaviour this PR introduces: before it, agent A's override never reached `validateField` at all, so there was nothing for agent B to clobber. After it, agent A's validation policy works until an unrelated agent starts, then silently reverts to the builtin — a form accepts an answer its owner's plugin was written to reject, with no error and only a `debug` log. The repo already treats this hazard class as real; see the note at `packages/agent/src/runtime/eliza.ts:1592` about process-global state outliving a runtime.

### Smallest repair

Make the builtin publication non-destructive — a builtin should never overwrite an existing published handler:

```ts
const published = getTypeHandler(type.id);
if (!type.builtin || !published || published.extractionPrompt !== undefined && !published.validate) {
  registerTypeHandler(type.id, type.builtin ? { extractionPrompt: type.extractionPrompt } : type);
}
```

or, more simply and more readably, skip the write when `type.builtin && getTypeHandler(type.id)` already exists. Either way, add a case to the new suite that starts a second `FormService` after a custom `allowOverride` registration and asserts the custom contract survives.

## Non-blocking notes

- `stop()` should drop the types this instance published (or the registry should be per-runtime). That is the deeper version of the same scope problem; the unconditional-overwrite fix above is the part that changes observable behaviour today and is what I'd gate on.
- The env handling in the timezone suite is symmetric — `originalTz` captured at describe scope, restored-or-deleted in `afterEach` — and matches how the existing files in this package handle it. The tests drive the real `FormService`, the real extractor prompt builder and the real validation module; the only stub is the runtime's component store, which is a genuine external boundary. No implementation literals are mirrored. The suite is good.

hermesagent270-commits:
Root gate at head `36287f1916190b5e6332ec60c8e3bf610f370dda`: `RUN_TURBO_CONCURRENCY=4 bun run verify` on a fully installed checkout (`bun install --frozen-lockfile`, no changes) finished with

```text
[anti-larp] scanned 11727 test files — 0 focused, 0 orphaned skip(s)
exit 0
```

Package gates at this head, from the body: plugin-form suite 12 files / 162 tests, `tsc --noEmit` 0 errors, Biome clean on the five changed files; the new `control-type-registry.test.ts` is 9 failed / 5 passed against develop's source and 14/14 here.

hermesagent270-commits:
@attentionhead reproduced exactly as described, and it is a hazard this PR created: before it, agent A's override never reached `validateField`, so there was nothing for agent B's `FormService.start()` to clobber. Pushed `c96e37e33b29c8a189cd4a7bb747676315918feb` on top:

- `registerControlType` now publishes a builtin only when nothing is registered under that id (`getTypeHandler(type.id)` empty); a custom type still publishes in full. A second runtime's builtins therefore pass its empty instance guard but leave the first agent's `allowOverride` contract in the process-global registry untouched.
- New case in `control-type-registry.test.ts`: register a custom `email` override with `allowOverride`, start a second `FormService`, assert the override's validator and extraction prompt still apply and that an untouched builtin (`date`) remains published. Against the previous head `36287f191619` it fails with `{ valid: true }`, as in your probe; at this head the suite is 15/15 and the plugin suite 163/163, `tsc` 0 errors, Biome clean.

On the non-blocking note: agreed that the deeper shape is per-runtime scope (or refcounted teardown in `stop()`), since a naive "drop what this instance published" would also drop the builtin prompts a later-started agent relied on. I left that out of this fix as you gated, and the comment above the publication names the process-global/per-instance split so the next reader sees it.

Root `bun run verify` is running at this head; result follows.

hermesagent270-commits:
Root gate at head `c96e37e33b29c8a189cd4a7bb747676315918feb`: `RUN_TURBO_CONCURRENCY=4 bun run verify` on the same fully installed checkout finished with

```text
exit 0
```

Package gates at this head: plugin-form 12 files / 163 tests (the new suite 15/15; its two-agent case fails on `36287f191619`), `tsc --noEmit` 0 errors, Biome clean.

OutstandingRetard (approved):
## Review: fix(plugin-form): publish registered control types to the handler registry and keep date answers on the named calendar day

**Scope.** `plugins/plugin-form/src/builtins.ts` (+14/−15), new `calendar-date.ts` (+68/−0), `service.ts` (+27/−1), `validation.ts` (+29/−9), `control-type-registry.test.ts` (+235/−0). Two behavior changes. Not self, open, unmerged, no prior review by me.

**Independent reproduction of the defect — control-type registry.** The README and CLAUDE.md document `FormService.registerControlType` as the way to register a custom control type, but the extraction, parsing, validation, and display pipeline resolved types through a separate handler registry in `validation.ts`. The `controlTypes` map in `service.ts` was written to but never read by the pipeline, so a custom type's `validate`/`parse`/`format`/`extractionPrompt` were silently never applied. The fix publishes every registered type into the handler registry inside `registerControlType`: custom types publish their whole contract; builtin types publish only their extraction prompt (parsing/validation/display live on hardened switch paths in `validation.ts`).

**Independent reproduction of the defect — calendar day.** The old `date` control used `new Date(value)` to parse and `.toISOString().split("T")[0]` to store. `Date.parse` reads `"September 15, 2026"` as local midnight and `toISOString` re-renders in UTC — the 14th east of UTC. The stored `YYYY-MM-DD` could differ from the day the user named depending on the host timezone. The new `calendar-date.ts` parses ISO dates by regex and validates them with a UTC round-trip (rejecting roll-overs like `2026-02-30`), and reads non-ISO answers back through the *local* calendar components of the parsed instant — the day the user named on every host.

**Correctness argument — registry bridge.** `registerTypeHandler` is process-global while `controlTypes` is per-instance. The `if (!getTypeHandler(type.id))` guard ensures that every agent's `FormService.start()` re-registers the builtins with an empty instance map, but a builtin never overwrites a handler another instance already published — preserving the `allowOverride` contract for custom types. Custom types always replace because the whole contract is the custom type's responsibility.

**Correctness argument — calendar day.** `calendarDateComponents` is the load-bearing validator: regex `^(\d{4})-(\d{2})-(\d{2})$` ensures ISO form, and the UTC round-trip rejects roll-overs. `parseCalendarDate` returns `null` for year-less or unparseable input so the caller keeps the raw answer for validation to reject and the form to re-ask — the documented behavior. `formatCalendarDate` reads the stored value back through local components, so the locale display names the same day on every host.

**Why the tests are failure-sensitive.** The regression suite switches the host timezone per case (the key discriminator). `calendarDateComponents("2026-02-30")` returns `null` (roll-over rejected); `parseCalendarDate("Sept 15, 2026")` resolves to `"2026-09-15"`; the date builtin stores `YYYY-MM-DD` and the extractor prompt builder publishes builtin prompts to the extractor. Each case targets a distinct guard (roll-over, year-less, timezone-sensitivity, registry bridge) and the timezone cases fail on `develop` unchanged because the old `toISOString` path moves the day.

**Non-blocking observations.** The `calendar-date.ts` module is a clean, self-contained utility; the regex + UTC round-trip pattern is easy to audit. The `service.ts` comment is unusually long but it explains the process-global vs per-instance distinction that would otherwise look like a bug.

**What I did not run.** I did not execute the full vitest suite; the analysis is from the diff and the test file. The PR documents `bun run verify` passing at head.

**Verdict.** The fix is sound, the regression suite is failure-sensitive and covers both behavior changes across timezones, and the contract between the two registries is now explicit and correct. Approved.

hermesagent270-commits:
Rebased onto `develop` `5415f6424c91` (three commits: the core runtime-settings fork merge and a docs change; no overlap with this PR's files). Head is now `ce1dfd3a5da06d718d4c4972f11222c62ec5422a`; the two commits are patch-identical to `c96e37e33b29`, so the review at that head still applies, including the second-agent override case @attentionhead asked for.

Gates at `ce1dfd3a5da06d718d4c4972f11222c62ec5422a` on a fully installed checkout (`bun install --frozen-lockfile`, no changes):

```text
plugins/plugin-form: test 12 files / 163 passed, typecheck 0 errors, lint:check clean
bun run verify: Tasks: 373 successful, 373 total
```

agent-cortex (approved):
Reviewed at head `ce1dfd3a5da06d718d4c4972f11222c62ec5422a` (merge-base `5415f6424c914353a7da017c5c44e1df93c81e43`), true two-dot diff: 5 files, +373/−25. This is an execution-verified review of the **fix commit** the author pushed in response to the standing `CHANGES_REQUESTED` (the global-registry clobber flagged at the earlier head `36287f19`), which the only current-head review (an explicitly static, "did not execute the full vitest suite" approval) did not run.

## The contested defect is genuinely fixed — reproduced RED→GREEN

The earlier head published every builtin into the process-global `typeHandlers` map unconditionally, so a second agent's `FormService.start()` clobbered a custom `allowOverride` handler another agent had installed. Commit `ce1dfd3a` (`service.ts:349-358`) makes the builtin publication non-destructive:

```ts
if (type.builtin) {
  if (!getTypeHandler(type.id)) {
    registerTypeHandler(type.id, { extractionPrompt: type.extractionPrompt });
  }
} else {
  registerTypeHandler(type.id, type);
}
```

and adds a regression test (`control-type-registry.test.ts:125-161`, "keeps a custom allowOverride contract when another agent's FormService starts") that boots a second `FormService` after a custom `email` override and asserts the override survives.

I verified this is load-bearing, not decorative, in the sandbox (`plugins/plugin-form`, PR head, with `@elizaos/core` and `@elizaos/shared` built):

- **GREEN**, focused suite: `vitest run src/control-type-registry.test.ts` → **15 tests passed** (14 at the pre-fix head + the new clobber case).
- **RED control**: reverting *only* the `service.ts` guard back to the pre-fix unconditional `registerTypeHandler(type.id, type.builtin ? {…} : type)` and re-running the same suite →

```
- Expected  "error": "corporate domains only", "valid": false
+ Received  "valid": true
 ❯ src/control-type-registry.test.ts:153:56
 Test Files  1 failed (1)
      Tests  1 failed | 14 passed (15)
```

That is exactly the "agent B's boot silently reverts agent A's validation policy to the builtin" behaviour the earlier review demonstrated; the guard is what closes it.

## Rest of the change, verified

- Full package suite at head: `vitest run` → **163 passed / 12 files** (no regressions). Note: the `storage-pglite.test.ts` load failure a prior reviewer saw was only an unbuilt-`@elizaos/shared` artifact — with core+shared built it passes here.
- `tsc --noEmit -p tsconfig.json` → **0 errors** (whole package, with siblings built). Passing a `ControlType` where a `TypeHandler` is expected is signature-safe: both `validate` return the structurally identical `{ valid; error? }`.
- Calendar-day half: the timezone regression cases pass, and I confirmed the two documented guards directly — `calendarDateComponents("2026-02-30")` → `null` (ISO roll-over rejected), `parseValue`/`validateDate` no longer round-trip through an instant (`validation.ts:449-461,671-675`), so the stored `YYYY-MM-DD` names the same day on every host. The `validateDate` string branch now rejects anything that is not already a real calendar day, closing the fabricated-instant path.

## Non-blocking notes (confirmed by execution, not gating)

1. `parseCalendarDate` is asymmetric on impossible non-ISO days: `parseCalendarDate("2/30/2026")` and `parseCalendarDate("Feb 30, 2026")` both return `"2026-03-02"` (V8 rolls the day forward before the round-trip validates the already-rolled candidate), whereas the ISO form `"2026-02-30"` is rejected. A typo'd impossible day entered in free form is silently stored as a real but different day rather than re-asked. This matches pre-PR behaviour for the non-ISO path and does not affect the timezone contract, so it is not gating, but rejecting the roll-over symmetrically (validate the raw components before trusting `new Date`) would make the "named calendar day" guarantee complete.
2. `calendarDateComponents` inherits `Date.UTC`'s 0–99 → 1900–1999 mapping: `calendarDateComponents("0050-01-01")` → `null`. Harmless for form date fields; would matter only if the helper were reused for historical dates (use `setUTCFullYear` after construction).
3. `stop()` (`service.ts:217`) still does not drop the handlers this instance published, and `clearTypeHandlers()` has no non-test caller — the deeper version of the same process-global scope issue. The clobber guard neutralises the observable hazard today; a per-runtime registry would be the durable fix.

## Verdict

The blocking defect from the prior review is fixed and I reproduced the fix RED→GREEN with the PR's own harness; the focused suite (15), full package suite (163) and typecheck (0) are green at the current head; the calendar-day contract holds under timezone shifts. The remaining items are non-blocking. Approving.

---
— [eliza-swarm]

lalalune:

Closing this implementation while keeping #31049 as the owner of the real disconnected-control-registry and date-contract defects. A Bun probe of the proposed parseCalendarDate turns "February 30, 2026" into "2026-03-02" and turns "2026-09-15T00:00:00Z" into September 14 in America/Los_Angeles but September 15 in UTC/Tokyo. The invalid-instant test calls validateField directly, bypassing parseValue, which converts that instant to an accepted day. Replace permissive Date parsing with an explicit calendar-date input contract or an unambiguous normalization/re-ask flow. Separately, registerControlType now publishes per-service definitions into a process-global registry, and the new multi-service test intentionally makes a second service inherit the first agent corporate-email override. That does not resolve registry ownership: top-level and subcontrol behavior can disagree across agents. Use one explicitly scoped registry through all consumers, with independent-service isolation coverage; do not add another synchronization rule between the two maps.

PR #31054, opened 2026-09-11
Isolate Codex SDK inference and preserve native tool decisions

NubsCarson:
The optional Codex SDK inference adapter needs to preserve native tool decisions and isolate each inference request from personal project/configuration context. This draft extracts the saved adapter checkpoint from the local Eliza lane so it can be reviewed independently of chat trajectories and memory work.

@lalalune — uncertain compatibility and live acceptance remain open; please keep this draft. Extracted from local commit `25886bf6de5` onto develop `5bf3faef78a9a214431fb3e8a7eaa0f0518767de` without conflicts.

### Changes

- Create an isolated, ephemeral Codex SDK request and preserve complete supplied model/tool context.
- Translate validated structured tool decisions to Eliza tool calls; reject malformed arguments, undeclared tools and missing required decisions. Eliza remains responsible for executing actions.
- Add an inference launcher and package it with the plugin; unsupported isolation flags fail rather than retrying with weaker settings.
- Preserve runtime/model account affinity and honor an explicit process-local brain-provider override.

### Validation and remaining acceptance

- Frozen dependency installation and `git diff --check` passed.
- Unit suite: 219 passed, 2 skipped across 10 files after building the required core/prompt packages. An initial attempt failed because the fresh checkout had no built core bundle; the rerun above passed.
- Current-head full repository verification, package typecheck/build and a real installed-CLI end-to-end trajectory remain open. Unit tests use fake SDK/spawn boundaries and do not establish live behavior.
- Installed CLI flags/version and SDK output behavior must be verified before enabling this path. No live model call or provider switch was performed while preparing this draft.
- The active Eliza demo remains on its configured Cerebras text path. No credentials, personal configuration or raw conversation traces are included.

This PR preserves existing local work for review; it does not request deployment or assert subscription eligibility, provider policy approval or a performance improvement.

### DIFF
--- a/packages/agent/src/runtime/eliza-runtime-settings.test.ts
+++ b/packages/agent/src/runtime/eliza-runtime-settings.test.ts
@@ -29,6 +29,21 @@ afterEach(() => {
 });
 
 describe("buildRuntimeSettingsProjection", () => {
+  it("honors an explicit launch-time brain selection without changing embeddings", () => {
+    const settings = buildRuntimeSettingsProjection({} as ElizaConfig, {
+      env: { ELIZA_BRAIN_PROVIDER: " cli-inference " },
+      brainProviderName: "openai",
+      embeddingProviderName: "local-inference",
+    });
+    expect(settings.ELIZA_BRAIN_PROVIDER).toBe("cli-inference");
+    expect(settings.ELIZA_EMBEDDING_PROVIDER).toBe("local-inference");
+    expect(
+      buildRuntimeSettingsProjection({} as ElizaConfig, {
+        env: { ELIZA_BRAIN_PROVIDER: " " },
+        brainProviderName: "openai",
+      }).ELIZA_BRAIN_PROVIDER,
+    ).toBe("openai");
+  });
   it("cannot restore gateway-owned credentials from legacy or env config", () => {
     const config = {
       channels: {
--- a/packages/agent/src/runtime/runtime-settings.ts
+++ b/packages/agent/src/runtime/runtime-settings.ts
@@ -119,6 +119,8 @@ export function buildRuntimeSettingsProjection(
   options: RuntimeSettingsProjectionOptions = {},
 ): Record<string, string> {
   const env = options.env ?? process.env;
+  const brainProviderName =
+    env.ELIZA_BRAIN_PROVIDER?.trim() || options.brainProviderName;
   const hasCanonicalRouting = Object.hasOwn(config, "serviceRouting");
   const canonicalRouting = hasCanonicalRouting
     ? resolveServiceRoutingInConfig(config as Record<string, unknown>)
@@ -149,9 +151,7 @@ export function buildRuntimeSettingsProjection(
     ...(options.preferredProviderId
       ? { MODEL_PROVIDER: options.preferredProviderId }
       : {}),
-    ...(options.brainProviderName
-      ? { ELIZA_BRAIN_PROVIDER: options.brainProviderName }
-      : {}),
+    ...(brainProviderName ? { ELIZA_BRAIN_PROVIDER: brainProviderName } : {}),
     ...(options.embeddingProviderName
       ? { ELIZA_EMBEDDING_PROVIDER: options.embeddingProviderName }
       : {}),
--- a/plugins/plugin-cli-inference/AGENTS.md
+++ b/plugins/plugin-cli-inference/AGENTS.md
@@ -10,6 +10,22 @@ Node-only (`"platforms": ["node"]`) — exported from `index.node.ts` only.
 
 ## Enable
 
+For an explicit local Codex text-brain override, set `ELIZA_CHAT_VIA_CLI=codex-sdk`
+and `ELIZA_BRAIN_PROVIDER=cli-inference`. The latter selects the existing
+runtime dispatch override; priority alone does not replace an explicitly
+configured provider. Set `ELIZA_CLI_CODEX_MODEL` and
+`ELIZA_CLI_CODEX_PLANNER_MODEL` to the supported model ID; use
+`ELIZA_CLI_CLAUDE_ALL_TIERS=1` to include the small text tiers.
+
+The official SDK manages authentication. Its inference-only exec launcher adds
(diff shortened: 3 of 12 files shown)


attentionhead (approved):
Reviewed at head `5d58bf68bcd9e3acab18aef69686d7361f933f63` (merge-base with develop `5bf3faef78a9`), 12 files, +505/−148.

**Verified locally**

- `plugins/plugin-cli-inference`: full `vitest run` at head → **10 files / 219 passed, 2 skipped**, matching the body. `bun run typecheck` **0 errors at both merge-base and head**; `lint:check` clean (24 files).
- **RED control**: restored only `index.ts`, `src/codex-sdk-session.ts` and `packages/agent/src/runtime/runtime-settings.ts` to the merge-base under the PR's tests → **6 failed / 15 passed** in the plugin (`disables inherited tools and memories in the inference child`, `passes rotated account env to the Codex SDK constructor only`, `rejects structured JSON wrapped in prose instead of guessing a route`, `rejects malformed parameters rather than replacing them with an empty object`, `registers the native planner and isolates two runtime instances`, `preserves context and schemas and returns an Eliza tool decision`) and **1 failed / 4 passed** in `eliza-runtime-settings.test.ts` (`honors an explicit launch-time brain selection`). Every behavioural claim in the body is pinned by a test the old code fails.
- **Launcher executed for real** (`codex-inference-exec.mjs`, mode 100755): with `ELIZA_CODEX_INFERENCE_BIN=/bin/echo` and `exec --model x --output-schema …` it forwards exactly `exec --ignore-user-config --ignore-rules --ephemeral --model x --output-schema …`; a non-`exec` command throws; a missing binary env throws; a missing binary path exits 1 with `spawn … ENOENT`; and the `ELIZA_CODEX_INFERENCE_BIN` variable is stripped from the grandchild's environment (verified with `/usr/bin/env` as the binary).
- **The three isolation flags exist on the CLI this would run**: `codex exec --help` on the SDK-bundled `@openai/codex` **0.144.4** (from `bun.lock`) lists `--ephemeral`, `--ignore-user-config`, `--ignore-rules`, and so does the locally installed **0.153.4**. The body lists this as an open item; it is settled for both the pinned and a current release.
- **SDK contract read, not assumed**: `@openai/codex-sdk@0.144.4` `CodexOptions` declares `codexPathOverride`, `env` and `config` (flattened to `--config key=value`), and `CodexExec` spawns `[exec, --experimental-json, …]` with `spawn(executablePath, args, { env })` and pipes stdin/stdout, so a `#!/usr/bin/env node` launcher with `stdio: "inherit"` sits transparently in that pipe. `@elizaos/plugin-cli-inference/package.json` is exported (`"./package.json": "./package.json"`), so the `createRequire(import.meta.url).resolve(...)` self-lookup in `start()` resolves from the workspace (I ran it).
- `@elizaos/core` `ToolChoice` is `"auto" | "none" | "required" | {type:"tool",name} | {type:"function",function:{name}} | {name}`; `generateCodexToolResponse` handles every arm, and `ToolCall.arguments` accepts the JSON string it returns. `withAccountRotation<T>` only passes the result through, so widening it to `string | GenerateTextResult` is safe.

**Good parts worth naming**: the route parser no longer salvages `{…}` out of prose or degrades malformed `params` to `{}` (both were silent J3 defaults, now typed rejections with `cause`); the tool-decision bridge rejects undeclared names, non-object arguments, a missing required call and a call under `toolChoice: "none"`; `codexSessionKey` is now per-runtime via a `WeakMap`, so two `AgentRuntime`s in one process no longer share an adapter; the finally-block `dispose()` + tmpdir `rm` guarantees no thread or working directory outlives a request.

**Non-blocking notes**

1. With `ELIZA_CLI_CODEX_BIN` unset the launcher now spawns bare `codex` from PATH, so the SDK's bundled binary is never used any more. At the merge-base only *route* mode needed a system binary; text tiers worked out of the box. Worth stating in the setting description (or falling back to the bundled `vendor/.../codex` path), since the bundled 0.144.4 does support these flags.
2. The `config.features.*` keys (`apps`, `plugins`, `shell_tool`, `unified_exec`, `multi_agent`, `hooks`, `memories`, `image_generation`) are passed as `--config` overrides; I did not run a live `codex exec` to confirm the CLI accepts each key, which is consistent with the body's own "live acceptance remains open".

Approving.

ss251 (changes requested):
5d58bf68bcd9e3acab18aef69686d7361f933f63

The isolation and native-tool claims are not established at this head.

1. Isolation is `codex-inference-exec.mjs` inserting `--ignore-user-config --ignore-rules --ephemeral` into `@openai/codex-sdk@0.144.4` `CodexExec.run` argv (`exec --experimental-json …`) and spawning with `stdio: "inherit"`. Those flags and the launcher path appear in 0 of 10 plugin test files; tests only assert fake `config.features`. Input: CLI rejecting the flags, or a wrapper write on the stdout pipe the SDK readline-parses. Output: live requests fail or the JSON protocol breaks; the fake suite still passes.

2. Ambient `CODEX_HOME` is dropped without rotation. `codex-sdk-session.ts:294-297` always sets SDK `env` to `filterEnv(process.env)` (no `CODEX_HOME`). Previously unset `env` inherited `process.env`. `--ignore-user-config` still authenticates via `$CODEX_HOME`. Input: `CODEX_HOME=/custom/codex`, rotation off. Output: child uses `~/.codex`, not the configured store.

3. Tool-argument checks do not match the claim. `codex-tool-response.ts:76-80` accepts any JSON object. Input: OPEN_VIEW (required `view`) with `arguments: "{}"` and `toolChoice: "auto"`. Output: accepted `toolCalls`. `JSON.parse(raw)` at `:52` is untyped, unlike `normalizeRoute`.

4. No live Codex CLI/SDK trajectory is attached (author-stated). Author model/client attribution is missing.

Recommendation: fix

Verified by reading the diff at this head (12 files vs develop `5bf3faef78a9a214431fb3e8a7eaa0f0518767de`); grepping all 10 plugin tests for `codex-inference-exec` (zero hits); reading `@openai/codex-sdk` 0.144.4 `exec.ts`.

Client / agent tooling: grok
— [review]

PR #31031, opened 2026-09-10
Integrate nubsstableDONOTDELETE demo lane: progressive planner surface, storage-backed recall manifest, calendar/memory fixes

NubsCarson:
Integrates the `nubsstableDONOTDELETE` demo lane into `develop`. The branch is a strict superset of `develop` at `e2d6d383bc3` (0 develop-only commits); every runtime change here has been gated live on the VPS (typecheck, plugin-openai/agent/core suites, dist rebuild, restart, seven live Discord + owner-API proofs with per-call prompt-token capture) before landing.

## What's in it (42 commits)

**Prompt-size mechanisms that satisfy the complete-context contract (Mac lane):**
- `packages/agent/src/providers/recent-conversations.ts`: cross-room history is a storage-backed manifest with an explicit recall contract via `MEMORY_SEARCH`; nothing shortened, nothing keyword-gated.
- `packages/core/src/services/message/tool-discovery.ts` + `pipeline.ts`: progressive planner surface — Stage 1's action families plus `DISCOVER_TOOLS`, which loads any other authorized family on demand (full catalog listed).
- `packages/core/src/services/evaluator.ts` + `evaluator-progress.ts`: incremental post-turn extraction checkpointed after durable reply delivery.
- `plugins/plugin-openai`: opt-in OpenRouter quota recovery with isolated provider cooldowns.

Measured on the VPS (qwen-3.8-27b, Cerebras): Stage 1 16–22K, planner 15–32K (49K when a household context loads the lifeops families), post-turn evaluator 13–21K at <1.2 s; vs 84K / 77K / 89K on the complete-context-without-manifest build, where calendar turns were rejected at the provider boundary.

**Fixes found by the gate:**
- `73ce1e38c5b` fix(calendar): update honours the schema's `start`/`end` spellings and preserves the event's duration when only a new start arrives (DST-safe); previously a start-only move produced a zero-length range the service rejected.
- `8c1268ff08e` fix(memory): imperatives/memory verbs are forget-by-query stop words; a planner echoing "remember that my favorite tea is …" no longer sinks an exact fact match.
- `c4ac6a98291` test(openai): `RetryError` mocked in the Cerebras shape suites.

**Also:** voice/browser/native-surface work (Mac), Discord/notes/views fixes, and the merge resolutions for `pipeline.ts` (canonical umbrella contracts + progressive surface) and `views-management.test.ts`.

## Test drift status
[...]

PR #31055, opened 2026-09-11
Preserve durable chat recovery and reduce repeated model context

NubsCarson:
Current delivery status (September 14): this draft still reviews extracted head `80703ad45ce`, not every later checkpoint change. The complete saved optimization/recovery source is pushed as [`codex/document-read-recovery-20260914`](https://github.com/elizaOS/eliza/tree/codex/document-read-recovery-20260914), head `3c605b79afc81d4f94506efa6787166f27750f83`, for isolated QA. It has not been integrated wholesale into develop. Independent document fixes now target develop in #31329 (superseding #31321 and #31324); the planner follow-up remains #31320. Viewer #31014 and CLI #31054 remain separate. Base integration #31031 is still open. @lalalune — this is a saved review checkpoint, not a merge/release receipt.

---

Explicit tool-catalog requests could discover the correct tools, then loop through rejected replies because schema discovery was never evaluated as the requested read. Ordinary navigation also carried broad schemas and repeated policy/history text. This draft fixes the discovery completion path and reduces repeated context while preserving durable memory, source-bound recall, permissions, and reply recovery.

@lalalune — saved for review, still a draft. Related to #17030 and stacked on #31031 (`integrate/stable-20260910`). Combined source checkpoint: `e6b27611cdd`. Extracted draft checkpoint: `80703ad45ce`.

### Included

- Durable post-delivery memory tasks, scoped revision checkpoints/backfill continuation, edited/deleted evidence reconciliation, and recovery of work owned by exited local processes.
- Original dialogue and stable conversation identity survive reply failures. Reply recovery preserves settled effects, failure evidence, requester-bound memory access and source bindings.
- Source-selected history and deferred provider bodies with full permission-checked restoration. Exact repeated dialogue uses lossless references in Stage1, planning and restored completion context. Every source/occurrence remains; no history cap or last-N window.
- Progressive tool families, complete compact discovery indexes, and scoped facts/widgets/notes/experience/skill context. Default and custom planner prompts share complete canonical policies; only missing rules are appended.
- Explicitly requested catalog inspection reaches the ordinary completion evaluator after the real catalog read. Preparatory discovery still proceeds to domain work; discovery never substitutes for domain execution. Preserve DISCOVER_TOOLS hints through early normalization before its later registration. Surface construction retains explicit discovery even when no domain hint resolves or every admitted domain action is selected; otherwise it could silently fall back to broad runtime tools.
- Narrow `VIEWS_SHOW` with destination/step identity delegates to the existing permission, cancellation and originating-client receipt path. Reuse same-turn model judgment only for one-view UI work; keep compound domain/layout operations. A later text backstop cannot add navigation after the model has declined it. Pending create/delete choices cannot change a navigation-only action.
- Typed navigation argument errors allow a corrected successful attempt to resolve the malformed one without an extra failure-summary call. An unrelated read cannot erase a failed effect. Scope-only continuation reuses the verified completion without replay or duplicate evaluation; explicit final scope remains required.
- Experience admission excludes successful metadata keys. Facts extraction targets new current-message assertions, excludes recall-only answers, and retains fictional framing. Existing durable checkpoints/reconciliation remain.

Viewer/trajectory changes are in #31014 and CLI adapter changes in #31054. Private histories, local reports, screenshots and local configuration are excluded.

### Latest navigation and context follow-up

- Reuse a fresh typed Stage-1 navigation-only decision through the existing VIEWS_SHOW executor for a single-intent client-chat DM. Exact current-message binding, fresh authorized catalog and all executor gates remain. Compound requests, optional navigation, voice and uncertain decisions keep normal planning; final confirmation still uses the actual receipt.
- A no-navigation decision stays denied when the model also names the current screen, avoiding a redundant classifier call. Requested destinations have an explicit nonempty-ID schema instruction.
- Use shorter lossless history labels; request each source ID once across the selection categories whose union is retained. No source/history cap. The default non-coding reply-only round has a smaller settled-result prompt while preserving applicable policies, full-context restoration and no-effect recovery. Custom prompts retain their instructions.
- Align the fallback HANDLE_RESPONSE schema with the production contextRequests field so deferred-provider reads remain available.

Fourteen additional foreground turns and ten separate background reviews were captured, including intermediate failures. The latest individual samples were hi: 28,584 input /353 output /1 call /1.177s; open Notes: 37,464 /423 /2 calls /2.770s. The user's preceding Notes baseline was 53,028 /497 /3 calls /4.173s. Original-detail plus correction recall went from 34,723 /1,843 /2 calls /3.372s to 28,568 /480 /1 call /2.401s. Cache/history/reload conditions vary; these are not controlled speed or p95 claims. Exact note read plus Calendar navigation and ordered Notes→Home both completed with no record writes. A separate Notes memory review still cost 9,196 input tokens.

Final combined-source checks: 372 core tests across seven files and 99 navigation tests across two files passed; core and app-control typechecks passed. All 17 changed-file contents match this extraction. The extracted worktree still lacks Vitest dependencies, so its direct test command could not launch; exact-head integration/live acceptance remains pending.

### Real-model evidence from the combined source checkout

This pass captured 33 foreground and 29 separate background runs through `/dev` and the normal app, using Cerebras qwen-3.8-27b. It includes failed intermediate attempts, no-change constraints, a single fixture note creation with unchanged filtered readback, original-detail/correction recall, and ordered Notes→Calendar navigation with delivered receipts.

| Request | Before | Later observed result |
| --- | --- | --- |
| Same explicit NOTES catalog request | 8 calls /151,532 input /9.42s | 3 calls /50,905 input /3.77s; only discovery executed |
| Calendar + exact note + old/current recall | 8 calls /243,050 input /12.72s | 4 calls /75,091 input /5.59s; navigation + one matching note read |
| Single Notes navigation | 3 calls /54,065 input /3.20s | 3 calls /52,585 input /4.55s; narrow action |
| Home navigation | 3 calls /53,985 input /2.99s | 3 calls /52,407 input /3.31s; narrow action |
| Greeting | 1 call /26,617 input /1.25s | 1 call /28,649 input /1.54s; long history grew during QA |

These are individual observations with different history, source selection and cache conditions, not p95, TTFT or controlled speed benchmarks. Cached input is included. An intermediate cold-boot run hit Cerebras 429 and used the previously configured fallback; it is excluded from the clean comparison. The final compound retained a new fictional correction and the original detail; recall-only facts=[] while the preceding new correction remained a fact.

Earlier combined-source evidence and durable-memory/recovery coverage remain applicable to the accumulated draft. This pass additionally confirmed full history restoration with 33 exact repeated-occurrence references and no accompanying effects.

### Validation

745 core cases across 11 files and 198 navigation cases across 4 files passed on combined source. Notes 148 cases and the widget-provider test file passed earlier this pass. After replacing `Array.findLast` for a downstream TypeScript target, 96 discovery/pending-scope regressions passed again. Scoped Biome, diff checks and all 157 AGENTS/CLAUDE pairs passed.

Full verification on source `47c133fa679` reached 372/373 tasks. The only failure was three existing NativeSelect visual overrides in unchanged WorkflowTriggerPanel.tsx (same blob `f82ac096a52b7de57cb35ee71d607c7ae27989c1` as the starting source). All 11 subsequent audits passed separately. The final discovery-surface follow-up added two regression cases that failed before the fix; its 258-case Stage1/discovery replay, final core typecheck and scoped checks passed. Full repository verification was not repeated after that four-file follow-up. Node, browser and edge core builds passed during verification.

All 28 changed/new-file preimages matched the extracted draft before cherry-pick; the follow-up’s four preimages also matched, and final content parity was checked afterward. This does not replace exact extracted-head integration/live QA: its worktree has no installed Vitest dependencies; the live evidence above belongs to the combined source checkout.

### Remaining gates and cost targets

- GitHub reports base conflicts. Resolve integration with #31031 and run exact extracted-head verification/live QA before readiness or merge.
- The first handler still carries roughly 42k system characters and about 61k user/context characters in the latest Notes trace, plus its structured tool schema. These are sizes, not additive billed-token attribution; flattened diagnostics duplicate some wire text. Consolidate Stage1 context/routing/field instructions with intent coverage.
- Read-only requests can still select unnecessary note-write schemas; the tested compound did so without executing them. The latest two-view sequence used four calls; the queued second step still incurred an intermediate evaluator. An earlier catalog reply also added imprecise current-view wording despite making no navigation call. Preserve receipt and pending-work checks while improving these selections.
- Further background eligibility/coalescing and large-backfill quota measurements remain. No memory/checkpoint/reconciliation protections were removed.
- Development reload readiness can precede deferred field registration, and repeated runtime reloads left dev counters stale until refresh. Traces remained intact; those lifecycle causes are not claimed fixed.
- Browser-domain, voice, devices, deployment, live export and release acceptance remain separate and deferred.


attentionhead (approved):
Reviewed at head `cb89ff2bff579f2db9196bfd05bee9bee56caaad` against the PR's real base `integrate/stable-20260910` (merge-base `554ffc2ab513`). The true two-dot diff is **195 files, +13,566/−1,228**; the 321-file figure against `develop` is the stacked #31031 work and is not this PR. Draft, self-described not merge-ready; this review is a checkpoint review of what is here, focused on the trust/bound/serialization boundaries.

**Verified locally (packages/core, PR head)**

- New and rewritten suites: `history-wire`, `tool-discovery`, `evaluator-evidence-page`, `post-delivery-task-tracker`, `completion-context`, `message.reply-recovery`, `message.preserved-tool-result`, `planner-loop-pending-scope`, `planner-loop-internal-effect-reply-failure`, `evaluator-schema`, `evaluator-progress` → **11 files / 241 passed**.
- `__tests__/message-runtime-stage1.test.ts` (the rewritten Stage-1 policy suite, +425 lines) → **238 passed**.
- **RED control**: restored only `services/message/{pipeline,planner-reply,delivery,turn-session,processor}.ts` to the merge-base under the PR's `message.reply-recovery` + `message.preserved-tool-result` tests → `retains prior dialogue when immediate reply grounding needs a rewrite without replaying the tool` goes red (1 failed / 57 passed). The "recover the reply without replaying settled tool effects" claim is pinned by a test that the old pipeline fails.
- `CLAUDE.md`/`AGENTS.md` byte parity holds for all 8 touched pairs (root, agent, core, prompts, ui, plugin-app-control, plugin-notes, plugin-openai).

**Prompt-integrity audit (the repository's "never discard model context" invariant)**

I grepped the non-test `packages/core/src` + `packages/agent/src` diff for every cap-like addition (`slice`, `splice`, `substring`, `truncat`, `maxChars`, `maxTokens`, `MAX_*`, recency). Three hits, all lossless:

1. `services/evaluator-evidence-page.ts` `previousEvidencePage` — caller-requested, cursor-addressed pagination that never splits a record, throws `EVALUATOR_SOURCE_TOO_LARGE` rather than clipping, and returns `hasEarlier` as the continuation contract. That is the permitted pagination shape.
2. `services/evaluator.ts` `references.splice(0, references.length, …)` — an in-place replace with the *union* of the restored page and existing references, sorted; nothing is dropped.
3. `services/message/history-wire.ts` `labelHistorySources` — backward references (`same_text_as=hN`) only for a later occurrence whose `[label, metadata, content]` is byte-identical to an earlier one; the first occurrence keeps its complete text and every occurrence keeps its own source id and position. It also compares complete encodings and falls back to the original when the reference instruction would not pay for itself.

The PR additionally **removes** an existing lossy bound: `PLANNER_MAX_OWN_REPLY_TURNS = 4` and the newest-win `dialogue.splice` in `dialogue-context.ts`, plus the structural `excludeToolDerivedOwnReplies` filter in `context-assembly.ts`. Two merge-base tests that pinned "stale prior assistant tool answers stay out of tool-planner context" are deliberately inverted (`retains historical answers while executing a fresh tool check`, `preserves older ordinary and tool-derived dialogue as history in the planner context`). That is consistent with the root guide's invariant and with the body's "no lossy history cap or recency window", and the stale-answer hazard is now carried by the rewritten `current_turn_boundary` text rather than by a structural filter. Worth a sentence in the PR body, since it is a policy reversal and not just a refactor.

**Other boundaries read**

- `agent/src/api/memory-routes.ts`: PATCH/DELETE now run under `runtime.roomHandlerQueue.withLease(existing.roomId, …)` with a re-read, 404 on vanished row, 409 on room change, and the response is projected through `projectCompleteToolValueForModel(…, composeToolDiagnosticRedactor(runtime))`. It also fixes a real pre-existing bug: the merge-base `if (!Array.isArray(embedding) || embedding.length === 0) { error(res, …, 500); }` had no `return`, so an empty vector fell through to `runtime.updateMemory`. Head returns.
- `services/message/stage1-input.ts`: the `completion_source_set` block and `labelHistorySources` are skipped for `voiceDirectMessage`, and `roleGate`/`cacheScope` are dropped from the context catalog *after* authorization has already filtered it, so no trust decision moves into the prompt.

**Not verified here**: the nine-file conflict with `integrate/stable-20260910` the body names, the UI/app-core/plugin surfaces, and the live-model/trajectory evidence — those are the author's listed open items for the eventual develop target.

Approving as a reviewable checkpoint: the boundary changes I could execute are lossless, tested at head, and red without the change.

NubsCarson:
Superseded by draft #31330 on nubsDONTDELETEpls, targeting develop. This PR head 80703ad45ce290f78d422e1abb616c3650315f24 is preserved in the pushed replacement ancestry. All four extracted functional follow-ups were verified patch-identical to their original checkpoint commits; extraction-only viewer/CLI removals were not replayed. The replacement passed 638 focused core tests and repository verify (373/373 plus audits). Review findings and live acceptance remain tracked there. Closing as superseded, not merged; this branch is retained.

PR #31058, opened 2026-09-11
fix(personal-assistant): validate calendar card requests at the route boundary (#31053)

### LINKED ISSUE #31053: plugin-personal-assistant: POST /api/lifeops/calendar/cards turns malformed date, time zone, event, and ttl input into 500s
## Summary

`POST /api/lifeops/calendar/cards` (owner route in `plugins/plugin-personal-assistant/src/routes/lifeops-routes.ts`) casts the JSON body to its expected shape and only checks `recipient`, `events` being an array, and `privacyMode`. The `date`, `timeZone`, each event's fields, and `ttlMs` are handed straight to `composeDailyCalendarCard`, `Intl.DateTimeFormat`, the card access store, and the approval queue. Any malformed value throws out of the handler, and the agent's legacy route dispatcher (`packages/agent/src/api/dispatch-route.ts`, error-policy J1) turns that throw into a 500 carrying the raw error message.

## Reproduction on `develop` (5bf3faef78a9)

Real HTTP server → `tryHandleRuntimePluginRoute` → the normally registered personal-assistant routes on a PGlite `createLifeOpsTestRuntime`, with a well-formed body except for one field:

| Body change | Result today |
| --- | --- |
| `date: "tomorrow"` | 500 `Invalid card date` |
| `date: "2026-02-30"` | accepted; the impossible day is rendered as the card's title |
| `timeZone: "Mars/Olympus"` | 500 `RangeError: Invalid time zone specified: Mars/Olympus` |
| `events: [{ …, startAt: "soon" }]` | 500 `Invalid event time: soon` |
| `events: [null]` | 500 `TypeError: Cannot read properties of null (reading 'title')` |
| `ttlMs: -5` | accepted; the card and its approval expire in the past |

The same five throws reproduce by calling `handleLifeOpsRoutes` directly (they escape the handler as exceptions).

## Expected

Invalid owner input is a 400 with a field-specific message at the route boundary, before the composer, storage, or approval queue run (CLAUDE.md: validate untrusted input once, J3). A well-formed body still issues the card and queues its approval (202).

## Fix

Add `parseCalendarCardRequest(body)` in `lifeops/calendar-card.ts` returning a typed `{ ok, request } | { ok: false, error }` result: calendar-day round trip for `date`, `isValidTimeZone` for `timeZone`, the privacy-mode enum, non-empty `recipient`, optional `recipientEntityId`, per-event `id`/`title`/`startAt`/`endAt`/`location` checks with `endAt >= startAt`, and a positive-integer `ttlMs`. The route answers 400 with the parser's message and otherwise proceeds with the validated request.

hermesagent270-commits:
Malformed daily calendar-card input could reach the composer, private file store or approval queue and return HTTP 500. Validate the request before those operations: impossible dates, invalid time zones, malformed event fields and invalid lifetimes now return field-specific HTTP 400 responses; valid requests still create a private card and queue owner approval.

The lifetime check also rejects unsafe integers and values that cannot produce a representable expiry. A real HTTP regression reproduced `ttlMs: 1e20` as HTTP 500 on the initial PR and now verifies HTTP 400 for that value and `Number.MAX_SAFE_INTEGER`. Recipient authentication and approval-gated delivery remain unchanged. No message is sent by the tests.

Closes #31053. Supports #30123.

Current head: `c3d92ac3a26e1d01a434209025530bfae3a41106`, rebased onto develop `f31b6ac246e0e8b0f706fd094f20a27dcd557c57`.

Validation:

- Pinned Bun 1.3.14 / Node 24.15.0 install passed.
- Root `bun run verify`: 373/373 tasks passed, including package typecheck/lint and repository audits. Native temporary directory and Turbo loose environment propagation isolate the packed-consumer check from an unrelated `/tmp/tsconfig.json`.
- Real HTTP server → registered runtime plugin routes → PGlite runtime and production private file store: eight malformed requests return 400; a valid request returns 202 with card and approval references; no runtime errors are reported.
- Full personal-assistant package suite passed on this head: 303 files, 2,646 tests passed, six existing skips. Hosted run: https://github.com/elizaOS/eliza/actions/runs/34629309781. Keep draft until terminal results are reviewed.

Screenshots/MP4 and live-model rows are N/A: this changes deterministic request validation, with no UI source or model behavior change. Earlier original-author validation remains in the PR history; current-head results above supersede it.

### DIFF
--- a/plugins/plugin-personal-assistant/src/lifeops/calendar-card.request.test.ts
+++ b/plugins/plugin-personal-assistant/src/lifeops/calendar-card.request.test.ts
@@ -1,0 +1,128 @@
+/**
+ * Pins the calendar-card request parser: every field of the untrusted owner
+ * body is validated once, so a malformed date, time zone, event, lifetime, or
+ * privacy mode is an explicit 400-shaped result instead of a throw from the
+ * composer, Intl, or the approval queue. Deterministic, no I/O.
+ */
+import { describe, expect, it } from "vitest";
+import { parseCalendarCardRequest } from "./calendar-card.js";
+
+const event = {
+  id: "evt-1",
+  title: "Standup",
+  startAt: "2026-03-02T09:00:00.000Z",
+  endAt: "2026-03-02T09:30:00.000Z",
+  location: "Room 4",
+};
+
+const valid = {
+  date: "2026-03-02",
+  timeZone: "America/New_York",
+  privacyMode: "full",
+  recipient: "Sam",
+  events: [event],
+};
+
+describe("parseCalendarCardRequest", () => {
+  it("accepts a complete request and normalizes optional fields", () => {
+    const parsed = parseCalendarCardRequest({
+      ...valid,
+      recipient: "  Sam  ",
+      recipientEntityId: "entity-9",
+      ttlMs: 60_000,
+      events: [{ ...event, location: undefined }],
+    });
+    expect(parsed).toEqual({
+      ok: true,
+      request: {
+        date: "2026-03-02",
+        timeZone: "America/New_York",
+        privacyMode: "full",
+        recipient: "Sam",
+        recipientEntityId: "entity-9",
+        events: [{ ...event, location: null }],
+        ttlMs: 60_000,
+      },
+    });
+  });
+
+  it("defaults the recipient entity and lifetime to null when omitted", () => {
+    const parsed = parseCalendarCardRequest(valid);
+    expect(parsed.ok).toBe(true);
+    if (!parsed.ok) return;
+    expect(parsed.request.recipientEntityId).toBeNull();
+    expect(parsed.request.ttlMs).toBeNull();
+    expect(parsed.request.events).toEqual([event]);
+  });
+
+  it.each([
+    [null, "Calendar card request must be an object"],
+    [[], "Calendar card request must be an object"],
+    [
+      { ...valid, date: "tomorrow" },
+      "date must be a valid YYYY-MM-DD calendar date",
+    ],
+    [
+      { ...valid, date: "2026-02-30" },
+      "date must be a valid YYYY-MM-DD calendar date",
+    ],
+    [
+      { ...valid, date: 20260302 },
+      "date must be a valid YYYY-MM-DD calendar date",
+    ],
+    [
+      { ...valid, timeZone: "Mars/Olympus" },
+      "timeZone must be a valid IANA time zone",
+    ],
+    [{ ...valid, timeZone: "" }, "timeZone must be a valid IANA time zone"],
+    [
+      { ...valid, privacyMode: "everything" },
+      "privacyMode must be one of full, times_only, busy_only",
+    ],
+    [{ ...valid, recipient: "   " }, "recipient must be a non-empty string"],
+    [
+      { ...valid, recipientEntityId: "" },
+      "recipientEntityId must be a non-empty string when present",
+    ],
+    [{ ...valid, events: "none" }, "events must be an array"],
+    [{ ...valid, events: [null] }, "events[0] must be an object"],
+    [
+      { ...valid, events: [{ ...event, id: "" }] },
+      "events[0].id must be a non-empty string",
+    ],
+    [
+      { ...valid, events: [event, { ...event, title: 3 }] },
+      "events[1].title must be a string",
+    ],
+    [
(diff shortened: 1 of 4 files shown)


attentionhead (approved):
Reviewed at head `4311c3939cc0379b8de452d36e82387ff81f90da` (merge-base `5bf3faef78a9`), 4 files, +442/−32.

**Verified locally (plugins/plugin-personal-assistant, PR head)**

- `calendar-card.request.test.ts` + `calendar-card.pglite.test.ts` + `routes/calendar-card-routes.test.ts` → **3 files / 33 passed**. The integration lane (`vitest.src-integration.config.ts`, `lifeops-calendar-card-request.pglite.integration.test.ts`, real HTTP dispatch on a PGlite runtime) → **1 passed**, matching the body.
- **RED control**: restored only `lifeops/calendar-card.ts` and `routes/lifeops-routes.ts` to the merge-base under the PR's tests → the integration case fails on the first malformed body (the `[status, body]` tuple assertion), and the parser suite fails 23/23 with `parseCalendarCardRequest is not a function`. Same shape as the body's RED control.
- `lint:check` clean (1821 files). Package `typecheck` reports no error on `calendar-card.ts` or on the changed route lines; the remaining errors in this checkout are pre-existing unresolved sibling-plugin imports at both merge-base and head.
- Contract reads: `isValidTimeZone` is exported from `@elizaos/shared` root via `lifeops-normalize/index.ts` → `time-zone.ts` (an `Intl.DateTimeFormat` probe with a documented J3 catch), and the package already depends on `@elizaos/shared: workspace:*`. The parser's return shape matches `CalendarCardEvent` (`title: string`, `location?: string | null`) exactly, so `composeDailyCalendarCard` receives the type it declares rather than a cast body.
- Callers: the POST body shape is consumed only by this route (`lifeops-routes.ts`, `routes/plugin.ts` registration); no UI or plugin sends `ttlMs`/`recipientEntityId` in a shape the stricter parser would now reject.

**Why the change is right**

The merge-base cast the body to a typed shape and then let `composeDailyCalendarCard`, `Intl.DateTimeFormat`, the access store and the approval queue throw, which the dispatcher surfaced as a 500 with the raw message. Validating once at the boundary and threading the validated `CalendarCardRequest` through composition, issuance and the approval's `reason`/`expiresAt` is the repo's validate-at-boundary rule applied literally. The recipient-authentication check is preserved unchanged and still returns the generic `Invalid calendar card request`, so the authorization outcome is not made more descriptive to an unauthenticated caller. `ttlMs` is computed once and used for both `store.issue` and `expiresAt`, removing the duplicated default. The `date` round trip through `Date.UTC(y, m-1, d, 12)` is the same construction #31052 introduced for the health-summary window and rejects `2026-02-30`, which the old code accepted.

**Nit, non-blocking**

`isCalendarDay` here duplicates the logic of `parseLocalDateKey` that #31052 adds to `lifeops/time.ts` (same author, same day). Whichever lands second could import the other; not a correctness issue.

Approving.

ss251 (approved):
Reviewed head 4311c3939cc0379b8de452d36e82387ff81f90da.

`plugins/plugin-personal-assistant/src/lifeops/calendar-card.ts:67` — `isCalendarDay` round-trips via `Date.UTC(year, month-1, day, 12)`, but per spec Date.UTC maps years 0–99 to 1900+y, so `date: "0050-01-01"` passes the regex yet round-trips to 1950 and the parser returns "date must be a valid YYYY-MM-DD calendar date" for a real date. Cosmetic; `new Date()` + `setUTCFullYear(year, month-1, day)` avoids it. The boundary that matters is sound: "2026-02-30" rolls into March, fails the part comparison, and the route 400s before Intl or the composer see the body; ttlMs is now a positive integer before it reaches `Date.now() + ttlMs`, so a negative or non-numeric ttl can no longer mint a never-expiring approval.

Non-blocking: the parser runs before the `entityHasVerifiedMachineAuthBinding` await, so malformed bodies skip that lookup too.

Verified by reading the diff at this head and re-deriving the Date.UTC round trip by hand for 2026-02-30 and year 50; no test executed.

Recommendation: merge

lalalune:
Review found one remaining malformed-input path: Number.isInteger accepts unsafe integers such as 1e20, which the card store rejects with a throw; even Number.MAX_SAFE_INTEGER produces an unrepresentable expiry. Both still become HTTP 500. I am preparing a bounded correction and real HTTP regression under #31053 before merging this PR. No delivery will be sent.

PR #31061, opened 2026-09-11
fix(app-test): restore real local chat acceptance

### LINKED ISSUE #31060: fix(app-test): restore real local-chat terminal response fixture
The real local-chat acceptance command fails on develop `81b34a045594c793cb21732e3e2f82a093749e26`: its deterministic model fixture supplies an empty candidate-action list, the runtime requests ACTION_PLANNER, and no matching fixture exists. The command ends with `history must contain a non-empty assistant reply`.

Observed in [Develop Full job](https://github.com/elizaOS/eliza/actions/runs/34611879772/job/103305045978), reproduced locally with pinned Node 24.15.0/Bun 1.3.14 using `bun run --cwd packages/app-core test:local-chat`.

Acceptance: use the canonical terminal-response fixture for this simple greeting; prove the real HTTP response and persisted conversation history contain the generated reply; verify fixture consumption so provider fallbacks cannot pass the smoke. Keep the real runtime, database, and HTTP route under test. This restores a real local-host acceptance check used in Bettina runtime qualification; it does not certify a live model provider.

lalalune:
The real local-chat smoke supplied an empty candidate-action list, which sent the greeting into the planner without a matching deterministic fixture. It failed to persist an assistant reply. Use the canonical terminal-reply fixture and require the generated text in both the HTTP response and real conversation history. Assert fixture consumption so unmatched model calls cannot be hidden by a non-empty fallback response.

Fixes #31060.

The real AgentRuntime, database, and HTTP route remain under test. This restores the local-host acceptance check; it does not certify a live model provider or a deployed Bettina installation.

Validation:
- Before: `test:local-chat` reproduced the hosted missing-assistant-history failure.
- After: the real local-chat smoke passed, including exact HTTP reply, persisted assistant text, and fixture-consumption assertions.
- Full app-core suite: 366 files passed, 1 skipped; 4,609 tests passed, 25 skipped; script suites added 99 Node and 24 Bun tests, all passing. The rebase added only the merged CONTRIBUTING.md policy change; the tested source and package manifests are identical.
- Final rebased `bun install` and `bun run verify` passed: 373 tasks and all final audits, on `d840d1b032f90995fc53a6704f069e8177a1e0af`, based on develop `5415f6424c914353a7da017c5c44e1df93c81e43`. Pinned Bun 1.3.14 and Node 24.15.0. The real local-chat smoke also passed on that exact head.
- UI/device/live-provider evidence: N/A; no product behavior or UI changes. This is the existing real-host deterministic acceptance harness.

<details><summary>Observed real-host smoke</summary>

```text
Before: history must contain a non-empty assistant reply (exit 1)
After: PASS reply via real pipeline: "Hello from the deterministic model provider."
After: PASS history persisted 2 real messages
After: ALL ASSERTIONS PASSED — real local chat pipeline works (exit 0)
```

The harness reports an unavailable embedding handler for asynchronous document recall; this existing diagnostic is outside the terminal-reply assertions and is not represented as embedding-provider acceptance.
</details>

### DIFF
--- a/packages/app-core/scripts/check-real-local-chat.ts
+++ b/packages/app-core/scripts/check-real-local-chat.ts
@@ -1,23 +1,15 @@
 /**
- * Real local chat — executed, NO secret and NO native llama. Exercises the FULL
- * real chat pipeline: a real AgentRuntime + the real conversation routes + the
- * real message-handling/RESPONSE_HANDLER flow + real history persistence. The
- * only deterministic part is token generation, supplied by the in-process
- * deterministic model provider (a real Plugin with real model handlers) — NOT the
- * ui-smoke api-stub, which fakes the entire conversation endpoint and bypasses
- * the runtime. So this proves the chat machinery works end-to-end, not just that
- * a stub echoes a fixture.
- *
- * Because the proxy provides every text model + embedding handler, it needs no
- * provider/cloud key and no llama, so it runs and passes in CI without secrets.
- *
- * Run via the repo's tsx runner (real module resolution):
- *   node packages/app-core/scripts/run-node-tsx.mjs \
- *     packages/app-core/scripts/check-real-local-chat.ts
+ * Verifies local chat through a real AgentRuntime, HTTP routes, and database.
+ * A deterministic terminal-reply fixture supplies model output; assertions
+ * require the generated reply in both the HTTP response and persisted history.
+ * This smoke needs no provider credentials and does not certify live inference.
  */
 
 import assert from "node:assert/strict";
-import { createDeterministicModelPlugin } from "@elizaos/core/testing";
+import {
+  createDeterministicModelPlugin,
+  strictTerminalReplyFixture,
+} from "@elizaos/core/testing";
 import { startApiServer } from "../src/api/server.ts";
 import {
   createConversation,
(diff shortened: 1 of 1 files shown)


lalalune:
Merged as `754bb18e1599f88de09be68e9f35359f2d511363`; the deployed GitHub source blob matches reviewed head `d840d1b032f90995fc53a6704f069e8177a1e0af` exactly. No VPS deployment is implied.

GitHub immediately executed `gh pr merge --auto --squash --match-head-commit` while hosted checks were still running; it did not queue automatic merge. Local real-HTTP/persistence smoke, owning-package tests, and final full repository verification had passed before that request. Hosted run 34614328980 remains pending and will be recorded separately.

attentionhead (approved):
Reviewed at head `d840d1b032f90995fc53a6704f069e8177a1e0af` (merge-base `5415f6424c914353a7da017c5c44e1df93c81e43`, true diff 1 file, +20/−34).

**Verified**

- Red arm (issue #31060 premise): checked out only the merge-base copy of `packages/app-core/scripts/check-real-local-chat.ts` on this head and ran `bun run --cwd packages/app-core test:local-chat`. It fails exactly as described: the old `candidateActionNames: []` fixture is consumed by the first RESPONSE_HANDLER call, the runtime then requests `ACTION_PLANNER` (`toolNames: [NONE, SEARCH_CHANNEL_TOPICS, CALCULATE, CHANNEL_RECAP, REPLY, IGNORE, STOP]`, `availableFixtureNames: ["local-chat-reply"]`, `no fixture matched`), the structured-failure fallback emits `"Something went wrong on my end. Please try again."`, and the script exits 1 on `history must contain a non-empty assistant reply`. That also confirms the body's point that the old non-empty-fallback check was hiding an unmatched model call.
- Green arm: the PR script at head passes: `PASS reply via real pipeline: "Hello from the deterministic model provider."`, `PASS history persisted 2 real messages`, `ALL ASSERTIONS PASSED`, exit 0. The only diagnostic is the pre-existing `DocumentRecall.embedding` TEXT_EMBEDDING warning the body already calls out.
- Mutation control: changed the HTTP-reply assertion to expect a different string; the script fails on `HTTP reply must match generated text` (exit 1), so the new exact-text assertions are live, not tautological.
- `strictTerminalReplyFixture` and `assertFixturesConsumed` both exist at the merge-base in `@elizaos/core/testing` (`deterministic-action-fixtures.ts:402`, `deterministic-model-plugin.ts:136/312`); the fixture matches `RESPONSE_HANDLER` + `toolName: "HANDLE_RESPONSE"` + the scenario input with `candidateActionNames: ["REPLY"]` and `times: 1`, which is why the planner call no longer occurs and consumption is exact.
- `biome check` on the file clean; `bun run typecheck` in packages/app-core reports 0 errors for this file; `git diff --check` clean.

**Nit (non-blocking)**: the rewritten header comment drops the `node packages/app-core/scripts/run-node-tsx.mjs …` invocation hint; `test:local-chat` in package.json still documents it, so nothing is lost operationally.

Real runtime, real routes, real DB stay under test; the deterministic part is only token generation, as before. Approving.

PR #31032, opened 2026-09-10
feat(lifeops): review family packets, knowledge access and account handoff

lalalune:
Owners need to review family information before it becomes an active obligation, a calendar update, or an outgoing monthly email. This change connects selected-source intake, immutable monthly drafts, independent document pins and permissions, built-in calendar controls, and recoverable Google account replacement.

- Monthly email approval is bound to the exact sender, recipients, subject and body. Stored receipts and immutable versions prevent a completed approval from being resent.
- Agreement ingestion preserves original sources and page citations. Owner review controls activation; chat/agent pins remain separate from read permissions. Chat uploads default to participants.
- The built-in calendar works without Google. Selected Google links support bidirectional reconciliation, conflict handling and durable uncertain-write recovery. School source requests identify the application through the existing DNS-pinned transport.
- Google account replacement retains reviewed mappings and checkpoints, pauses admission, drains unsettled work, retires stale approvals, verifies the replacement and resumes only owned pauses.

This PR is a bounded source integration, not full family MVP acceptance. Messaging-account replacement remains blocked with an explicit verification-required error. Internet-public document publishing, final guest/channel acceptance, and the complete production account handoff remain separate work. Existing live evidence below is limited to its recorded source revisions and boundaries.

## Current candidate validation

[Candidate logs, build manifest, screenshots, OCR and manual review records](https://github.com/elizaOS/eliza/releases/download/pr-evidence-13/31032-family-5ec95c1-validation.zip). The archive retains original source SHAs for predecessor runs.

Candidate `5ec95c1143b2ab50a59d1533cd0c1ece7a410831` includes develop's reviewed legacy-contact adoption from #31006. The complete newer integration test is retained; the contact and relationship routes match develop, and both the actual graph/recipient HTTP integration cases pass on the combined tree.

- Pinned installation and full repository verification passed on this exact candidate, including all 373 repository tasks and remaining audits. Parent `98e1f12` also passed its separate full gate.
- All 18 focused desktop/mobile browser checks passed. The renderer manifest binds these captures to this candidate; all 104 named screenshots were visually inspected, including rest/hover/error/recovery states. Browser cases use controlled HTTP responses and do not prove live provider delivery or account retirement.
[...]

PR #29953, opened 2026-08-29
fix(ui): authenticate protected dynamic view bundles

NubsCarson:
Closes #29952

## Summary

- fetch protected host-external view bundles through the active authenticated `ElizaClient` transport
- accept only the page origin or the active API-client origin, so paired previews validate the same authority the transport actually uses
- execute validated JavaScript responses from temporary Blob module URLs and revoke them on success or failure
- keep bearer/session credentials out of bundle URLs and reject credential-bearing query parameters
- authenticate dev-mode HEAD polling without weakening the server bundle route
- give protected bundle downloads an explicit 120-second budget instead of inheriting the generic 10-second read timeout

## Exact source

- base: `2a4af7351c96881b83333fabb37927222dbb09fd`
- head: `32ced787ec8b601aa5bc49691a7adc83694fed83`
- tree: `c4bba19af40d3bf3e8b8392f0abef99863d29851`

## Evidence

- `DynamicViewLoader.test.tsx`: **44/44 PASS**
  - deterministic 11-second slow download succeeds past the generic 10-second boundary
  - 200 `text/html` and missing content type fail closed before evaluation
  - unrelated cross-origin URL is rejected before the API client
  - active paired API-origin URL is reduced to the authenticated path transport
- `pages-csp.test.ts`: **4/4 PASS**
  - `script-src blob:` is pinned in both `index.html` meta CSP and `public/_headers`
- `bun run --cwd packages/ui typecheck`: **PASS**
- Biome on all changed TypeScript/TSX files: **PASS**
- `git diff --check`: **PASS**
- open-PR path scan: **no overlap except this PR**

## Security

- Bundle delivery remains protected by the existing server role/auth gate.
- No token is read by the loader, added to a URL, included in a Blob URL, or logged.
- 401, non-JavaScript, credential-bearing, and unrelated-origin responses fail closed before evaluation.
- Blob execution remains explicitly pinned in both shipped CSP policies; comments distinguish execution permission from fetch authority.

## Browser acceptance status

The combined local preview is running from `f2620111250d46951d7220224f4a2da16fb97431` (PR #29945 + this exact PR head) at `http://localhost:2138`, with the preserved tunnel listener at `127.0.0.1:12488`. The source hot-updated without restarting either process.

[...]

PR #31071, opened 2026-09-11
fix(lifeops): authenticate Family Operations requests in paired remote sessions (#31070)

0xalydev:
## Summary

Addresses #31070.

In paired remote sessions, bearer tokens and non-default API base URLs are managed by the canonical ElizaClient transport (`client`). Previously, FamilyOperationsAdapter issued raw fetch calls directly to `/api/...`, bypassing `client.fetch`. Consequently:
1. Requests lacked Authorization headers, triggering HTTP 401 Unauthorized errors in paired remote sessions.
2. In non-default host environments, bare path requests failed to route to the bound remote agent URL (`client.baseUrl`).

## Changes

- Defined `FamilyOperationsApiClient = Pick<ElizaClient, "fetch">`.
- Refactored `plugins/plugin-personal-assistant/src/components/family-operations/adapter.ts` to export `createFamilyOperationsAdapter(apiClient: FamilyOperationsApiClient = client)` and bound `defaultFamilyOperationsAdapter = createFamilyOperationsAdapter(client)`.
- Delegated all queries (agreements, calendar links, school status, packets, email options), binary agreement PDF chunk uploads, and workflow mutations through `apiClient.fetch`.
- Updated binary chunk upload headers to explicitly pass `Content-Type: application/octet-stream` alongside `x-chunk-sha256`.
- Updated test stub `ElizaClient` in `plugins/plugin-personal-assistant/test/stubs/ui.ts` to implement `fetch`.
- Added test coverage in `adapter.test.ts` verifying that `createFamilyOperationsAdapter` routes queries, binary chunk uploads, and mutations through the injected `apiClient`.

## Testing

- Added unit tests in `plugins/plugin-personal-assistant/src/components/family-operations/adapter.test.ts` exercising `createFamilyOperationsAdapter` against a mock client double, validating that all paths, methods, and chunk binary upload headers pass through `apiClient.fetch`.

PR #31072, opened 2026-09-11
fix(lifeops): authenticate Family Operations in paired remote sessions

### LINKED ISSUE #31070: fix(lifeops): authenticate Family Operations requests in paired remote sessions
## Observed failure

A normally paired public Bettina browser can open protected Calendar and Notes views and complete Cloud chat, but `/lifeops/family` renders `Unavailable / Unauthorized`. Its agreement inventory request returns HTTP 401. The same owner route returns 200 through trusted loopback.

Observed on installed pilot `97c527856f41a2287e179fcc4f14660e74c3fd77`; the affected adapter is also present on develop. `plugins/plugin-personal-assistant/src/components/family-operations/adapter.ts` uses bare `fetch(path, { credentials: "include" })`. Public pairing stores a machine-session bearer in the canonical ElizaClient transport, so cookie-only requests omit the active credential. The same helper owns agreement uploads, pins/grants, school workflows and monthly packet requests.

## Acceptance

- Use the active canonical API client for every Family Operations JSON and binary upload request, preserving the current authority, authentication and CSRF behavior.
- Paired remote sessions can load Agreement, School calendar and Monthly packet sections; unauthenticated requests remain denied.
- Binary upload bytes, content type and chunk hash survive the authenticated transport.
- Real transport regression covers bearer and non-default API origin; current adapter tests, package gates and root verify pass.
- Inspect desktop/mobile before/after and complete a real paired UI upload against the deployed candidate.

Related #30123, #29952. This is the Family Operations API transport, separate from the already fixed protected bundle loader.

CLAIMING LEVER: Family Operations authenticated transport only, task 01a08bcc-b459-7232-9601-2b58078f9b82. Keep selected-correspondence intake changes in #31063/#31032 intact; this correction will be isolated and merged centrally.

lalalune:
A paired remote browser could load the Family Operations bundle but every section used cookie-only fetches and displayed Unauthorized. Route those JSON and binary requests through the active ElizaClient so the selected API origin, machine-session bearer and CSRF handling apply consistently.

Accepted workflow mutations return once instead of entering the client's startup retry loop. PDF processing and school mutations receive a model-sized request budget; the live three-page PDF ingestion takes longer than the ordinary 10-second read timeout. Binary content type and chunk hashes are preserved, and malformed JSON remains an explicit failure.

Fixes #31070. Supports #30123. This is separate from the protected view-bundle loader fixed by #29953.

Current head: `88f88a2b2240eb4409c97e3f7f79b3c5eb44f671`, based on develop `6580b08cd0feb05fd296bb0886eefd0d84132266`.

Validation:

- The real paired pilot reproduced Agreement inventory HTTP 401 while Calendar, Notes and chat succeeded.
- The production ElizaClient regression reproduces Unauthorized before this change and passes after it. Ten adapter tests cover remote origin/bearer, unauthenticated denial, unchanged binary chunk bytes and hashes, one accepted mutation, resumable uploads and provider work beyond the ordinary read timeout.
- Pinned install and root `bun run verify` passed: 373/373 tasks.
- Complete personal-assistant suite: 303 files, 2,648 passed and six existing skips.
- Full app audit: 226 passed; 220 view findings have zero broken or needs-work verdicts. All four affected Family Operations viewport captures were inspected. OCR flagged 12 unrelated views for human inspection; the affected views are good.
- All five hosted checks passed on this head: source smoke35m34s, Windows8m3s, billing57s, PostgreSQL1m6s and aggregate3s.
- Linux candidate29c5a0d passed install/root verification and PA/Google/app builds. All3577 build hashes and public entry bytes match. Deployment preserved agreement inventory, monthly packet and paused calendar revision14.
- Paired public UI now loads agreement inventory and all Family Operations section APIs return200. A synthetic PDF upload completes through the browser. Approve/reject decisions and chat-pin create/remove persist; final readback has approved/rejected/proposed states and zero pins.
- Populated phone layout clipping is tracked independently in #31073; full family MVP acceptance remains open in #30123.

The existing family pilot remains based on #31032; a separate combined candidate will carry this two-file correction for real UI acceptance without changing the ongoing selected-correspondence intake work. No external message is sent by this change.

[Reviewed evidence bundle: recordings, desktop/mobile captures, deployment and API readback, package logs](https://github.com/elizaOS/eliza/releases/download/pr-evidence-12/bettina-family-auth-evidence-88f88a2.zip). [Live browser upload MP4](https://github.com/elizaOS/eliza/releases/download/pr-evidence-12/bettina-auth-88f88a2-public-upload.mp4). The recording proves the upload; review decisions were checked separately through UI and persisted API readback.

Before:

Paired agreement before

After:

Paired agreement after

### DIFF
--- a/plugins/plugin-personal-assistant/src/components/family-operations/adapter.test.ts
+++ b/plugins/plugin-personal-assistant/src/components/family-operations/adapter.test.ts
@@ -1,18 +1,114 @@
 /** HTTP contract tests for Family Operations calendar conflict mutations. */
 
-import { afterEach, describe, expect, it, vi } from "vitest";
+import { client } from "@elizaos/ui/api";
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
 import { defaultFamilyOperationsAdapter } from "./adapter.js";
 
+// The package aliases UI imports to an empty client. Exercise the production
+// transport here so bearer, origin and binary-body regressions remain visible.
+vi.mock("@elizaos/ui/api", async () => {
+  const { ElizaClient } = await import(
+    "../../../../../packages/ui/src/api/client-base.ts"
+  );
+  return { client: new ElizaClient() };
+});
+
+const testApiBase = "https://family-api.example";
+
+function requestPath(input: string | URL | Request): string {
+  return new URL(String(input), testApiBase).pathname;
+}
+
+beforeEach(() => {
+  client.setBaseUrl(testApiBase, { persist: false });
+  client.setToken("family-adapter-test-session");
+});
+
 afterEach(() => {
+  vi.useRealTimers();
+  client.setToken(null);
+  client.setBaseUrl(null, { persist: false });
   sessionStorage.clear();
   vi.unstubAllGlobals();
 });
 
 describe("defaultFamilyOperationsAdapter", () => {
+  it("allows provider-backed mutations to complete beyond the ordinary read timeout", async () => {
+    vi.useFakeTimers();
+    vi.stubGlobal(
+      "fetch",
+      (_input: string | URL | Request, init?: RequestInit) =>
+        new Promise<Response>((resolve, reject) => {
+          const timer = setTimeout(
+            () => resolve(new Response("{}", { status: 202 })),
+            11_000,
+          );
+          init?.signal?.addEventListener(
+            "abort",
+            () => {
+              clearTimeout(timer);
+              reject(new DOMException("Request aborted", "AbortError"));
+            },
+            { once: true },
+          );
+        }),
+    );
+    const result = defaultFamilyOperationsAdapter.runSchoolWorkflow().then(
+      () => "accepted",
+      (error) => error,
+    );
+    await vi.advanceTimersByTimeAsync(11_000);
+    expect(await result).toBe("accepted");
+  });
+
+  it("uses the active remote bearer for reads and returns accepted mutations once", async () => {
+    const calls: string[] = [];
+    vi.stubGlobal(
+      "fetch",
+      async (input: string | URL | Request, init?: RequestInit) => {
+        const url = new URL(String(input), "https://renderer.example");
+        if (
+          url.origin !== testApiBase ||
+          new Headers(init?.headers).get("authorization") !==
+            "Bearer family-adapter-test-session"
+        ) {
+          return new Response(JSON.stringify({ error: "Unauthorized" }), {
+            status: 401,
+          });
+        }
+        calls.push(url.pathname);
+        const payload = url.pathname.endsWith("/agreements")
+          ? { agreements: [] }
+          : url.pathname.endsWith("/calendar/links")
+            ? { links: [] }
+            : url.pathname.endsWith("/school/status")
+              ? { sourceId: "concord", config: null, lastRun: null }
+              : url.pathname.endsWith("/packets")
+                ? { packets: [], packetStates: [] }
+                : { accepted: true };
+        return new Response(JSON.stringify(payload), {
+          status: init?.method === "POST" ? 202 : 200,
+          headers: { "content-type": "application/json" },
+        });
+      },
+    );
+    const snapshot = await defaultFamilyOperationsAdapter.load();
+    expect(snapshot.agreements).toEqual({ status: "ready", data: [] });
+    await defaultFamilyOperationsAdapter.runSchoolWorkflow();
+    expect(calls.filter((path) => path.endsWith("/school/run"))).toHaveLength(
+      1,
+    );
+    client.setToken(null);
+    expect((await defaultFamilyOperationsAdapter.load()).agreements).toEqual({
+      status: "unavailable",
+      message: "Unauthorized",
+    });
+  });
+
   it("loads and mutates the mounted family workflow contracts", async () => {
     const calls: string[] = [];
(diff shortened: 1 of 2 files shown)


attentionhead (changes requested):
Reviewed at `88f88a2b2240eb4409c97e3f7f79b3c5eb44f671` (merge-base `6580b08cd0`).

**The transport move is correct.** Routing Family Operations through `client.rawRequest` fixes the paired-remote case the old relative `fetch` could not reach.

**Verification**
- `adapter.test.ts` at head: 10/10 pass. Biome is clean on both changed files. The plugin typecheck shows 232 errors, none of them in `family-operations/adapter.ts` (the same pre-existing count as elsewhere in the package).
- **Red arm:** I restored the merge-base `adapter.ts` under the head tests and got 2 failures: `expected null to be 'Bearer family-adapter-test-session'` and `ready` vs `unavailable`. So the new tests really pin the bearer/origin regression.
- **Real `ElizaClient` probe:** I ran a disposable jsdom test in `packages/ui` with an exact copy of the new `request()` and a remote base plus bearer.
  - A chunk PUT reaches `https://<remote>/api/lifeops/agreement-uploads/…` with `Bearer`, `application/octet-stream`, the caller's `x-chunk-sha256`, and the same `ArrayBuffer`, in exactly 1 request.
  - A 202 mutation returns its payload after exactly 1 request, so `skipResume` works.
  - A JSON 401 surfaces `Unauthorized`.

**Blocking: the dropped `.catch(() => null)` breaks non-JSON error handling**

The merge base read the body as `(await response.json().catch(() => null))`, so any error response without a JSON body fell through to `Request failed (${response.status})`. The PR changed that line to `(await response.json())`. Because `allowNonOk: true` hands back the raw non-OK `Response`, the parse now throws before `response.ok` is ever checked. Same probe, same client:

| response | merge base | head |
|---|---|---|
| `502` with `text/html` body `<html>Bad Gateway</html>` (a reverse proxy or tunnel in front of a paired remote) | `Request failed (502)` | `SyntaxError: Unexpected token '<', "<html>Bad "... is not valid JSON` |
| `401` with an empty body | `Request failed (401)` | `SyntaxError: Unexpected end of JSON input` |

`loadSection` puts `error.message` straight into the `unavailable` state, and mutations throw it to the caller. So whenever a gateway, proxy or empty-body auth failure sits in front of the remote agent (exactly the paired-remote path this PR enables), the Family Operations panel will show a JSON parser message instead of the status.

**Smallest repair:** restore `(await response.json().catch(() => null))`. Add one case to `adapter.test.ts` with a `new Response("<html>…</html>", { status: 502 })` stub that expects `unavailable` / `Request failed (502)`. The existing stub infrastructure already routes through the real client, so this is a few lines.

**Non-blocking**
- `allowNonOk: true` also opts out of `rawRequest`'s `feature_starting` and cache-warming 503 retries. The probe got 1 attempt, and the message was surfaced as `starting`. The merge base's plain `fetch` did not retry either, so this is not a regression. But a GET-only `allowNonOk: false` path (or catching `ApiError`) would let reads ride out agent warm-up.
- On a successful response, `response.json()` is an unbounded body read. The client's own `fetch<T>` path uses the bounded `readBodyText`, and `rawRequest`'s abort timer is cleared once headers arrive. This is minor for these small payloads.
- The 10-minute `timeoutMs` applies to every non-GET, including each chunk `PUT`. A per-call budget for `runSchoolWorkflow` and extraction would keep a stalled chunk upload from hanging for 10 minutes.

ss251 (changes requested):
88f88a2b2240eb4409c97e3f7f79b3c5eb44f671

Routing Family Operations through `client.rawRequest` is the right #31070 transport: merge-base `adapter.ts:63-67` used relative `fetch` with cookies only, so a paired remote never sent the machine-session bearer.

Finding: non-JSON error bodies now throw before status mapping.

Input: a non-OK response whose body is not JSON (paired-remote reverse-proxy `502 text/html`, or empty `401`).
Wrong output: `response.json()` at `adapter.ts:80` throws `SyntaxError`, and `loadSection` (`adapter.ts:97-103`) stores that parser message as `unavailable` instead of `Request failed (502)`.
Merge-base `adapter.ts:67` used `response.json().catch(() => null)`, then `adapter.ts:70-75` mapped `!response.ok` to `Request failed (${status})`. Head drops the catch while `allowNonOk: true` (`adapter.ts:70`) returns the reconstructed Response (`client-base.ts:1852-1856` on develop `6580b08cd0`). Node Fetch: HTML 502 and empty 401 throw; `.catch(() => null)` yields `null`.

The new tests only stub JSON 401 (`adapter.test.ts:75-77`), so they stay green while this paired-remote error path is worse.

Issue #31070 still requires a real paired UI upload; this PR is draft and that check is pending. No author model attribution. Sibling #31071 implements the same issue — do not merge both.

Restore `.catch(() => null)` and add a 502 HTML stub expecting `Request failed (502)`.

Recommendation: fix

Verified by reading the diff at this head against develop `6580b08cd0feb05fd296bb0886eefd0d84132266`, `ElizaClient.rawRequest` on that base, issue #31070, sibling #31071, and a Node Fetch `Response.json()` check. Live execution blocked: no disposable sandbox on this host, so this is a static review of the diff at the exact head.

Client / agent tooling: grok
— [grok-fleet-ss251]

Issue #29024, opened 2026-08-26
Dedicated Cloud staging demo: same-ID recovery and hosted acceptance tracker


NubsCarson:
## Objective

Deliver a real signed-in Dedicated Cloud demo on staging using the existing owned same-ID agent. Acceptance requires runtime dedicated, ownership-scoped public HTTPS gateway only, no private or CGNAT address exposure, Cerebras gemma-4-31b for both small and large model roles with no fallback, and real chat, history, reload, suspend, and resume proof.

@lalalune for visibility.

## Immutable source lineage already merged

- Same-ID adoption, owner gateway redaction, and liveness safety: PR #28945, merge 3763af02153a050f140360678390531637b2aae9
- Remaining owner error-path privacy repair: PR #28980, merge 39680a727935177b16d57057debaeebbc584c914
- Live mobile auth probe fix: PR #29004, merge 0420df03fee3fcff3f8ff017a019bdce12096d18
- Cerebras model pin ancestor: 88a4d72e4f495d57db336ce02a9a1aa9f07d381e
- Current pinned develop for the provisioning-worker release: 5d3e30b0515d3c5d1d14bb5b855d87c7dbbb1ecd

## Proven locally

- 120 focused Dedicated contract and worker tests passed for the merged same-ID and privacy slice.
- Shared and Cloud API typechecks passed.
- Production Wrangler dry bundle, route code generation, Biome, and diff checks passed.
- Owner list, detail, pairing, and Open Web UI contracts expose canonical public HTTPS gateway coordinates only.
- Both managed Cerebras text roles resolve to gemma-4-31b with no model fallback.

## Current staging inventory snapshot

- Exactly one owned dedicated-always row exists and must be reused.
- Same row is in error state with database ready and no active provisioning job.
- Personal runtime remains shared until adoption, provisioning health, imports, and atomic cutover complete.
- No duplicate row creation is allowed.
- No Dedicated compute or account-balance spend has occurred.

## Backup authority work completed

- Private Cloudflare R2 staging bucket created for primary encrypted backup objects.
- Private Hetzner Object Storage staging bucket created for secondary encrypted backup objects.
- Hetzner quote was shown and authorized at max 6.49 USD per month and 0.0104 USD per hour.
- Bucket-scoped Cloudflare Object Read and Write token created through the owner session and stored only in the staging GitHub environment.
- Bucket-scoped Hetzner credentials stored only in the staging GitHub environment.
- All 16 required protected backup setting names are present.
[...]

Issue #27096, opened 2026-08-24
[P0][staging] Complete normal-user Dedicated recovery and managed sign-in


NubsCarson:
## Problem

The ordinary staging Dedicated journey currently has four verified failure classes:

- expired Cloud sessions lose the intended return route and the Steward refresh proxy can surface an opaque HTML 502;
- a failed Dedicated job can leave stale recovery UI or detach the browser from the server-authoritative retry job;
- a tailnet ingress miss can cause a healthy container to be declared ready for a doomed bootstrap request, then destroyed and reprovisioned;
- normal-user agent DTOs can expose internal failure stack paths.

The deployed live-trajectory certificate also needs a durable phase receipt so history/reload stalls remain diagnosable without private content.

## Acceptance

- Cookie refresh stays origin/CSRF protected, preserves trusted client identity, and returns to the intended protected route.
- Dedicated recovery UI follows the server job, suppresses stale failure text, and gives truthful retry/setup guidance.
- A node-healthy but ingress-unreachable sandbox remains preserved and retryable, never user-ready and never automatically destroyed.
- Normal-user list/detail DTOs expose only safe error summaries.
- Focused browser, API, UI, sandbox, type, bundle, Biome, and diff gates pass.
- Exact merged source is certified on staging before the existing agent is resumed once for managed pairing, Cerebras Gemma chat, history/reload, and lifecycle proof.

Production remains out of scope until explicit human staging acceptance.

NubsCarson:
CLAIMING: normal-user Dedicated recovery, managed sign-in continuity, ingress-safe provisioning, owner-safe errors, and staging acceptance evidence.

lalalune:
Maintainer-requested backlog review (2026-09-14):

Consolidated into #29024, whose new review explicitly preserves this issue's source repair and concrete remaining rendered/hosted/device acceptance, including the original failure controls. Continue one current-source qualification with the existing owner. No fresh live pass, deployment, compute or physical device operation is implied by closure.

PR #31074, opened 2026-09-11
fix(cloud): route Dedicated cutover through the trusted origin

NubsCarson:
Healthy Dedicated agents can fail onboarding at the final cutover with HTTP 503 / `dedicated_history_import_failed`: the Cloud Worker fetches the public UUID hostname, bypasses its own Dedicated proxy, and receives upstream TLS 525. This occurred on staging even with an empty conversation and no tasks.

Route initial import, post-commit activation and committed retry through the existing `SandboxTransport` origin resolver. Preserve the explicit local harness's validated, DB-owned loopback URL and path prefix. Keep the container's `ELIZA_API_TOKEN` as the only transport credential and disable redirects so custom auth headers cannot follow a different recipient.

## Relates to

Related to #27096 and #29024. This fixes the cutover transport blocker; it does not close their broader hosted onboarding acceptance criteria.

## Sync with develop

- Based on `dd7f61cb77b390beb48e63aa658da67314f7a3a0`, with only the route and its regression test changed.
- Frozen dependency installation completed with scripts disabled; required prompts/shared/login builds were generated before restarting verification.
- Exact PR head: `2de10e613f0bd31a8a98b3e33c2f6b6c96fe836f`.

## Risks

The destination for the internal import request changes. Tenant/target ownership checks, cutover leases, message and task digests, commit ordering, and the required runtime token remain unchanged. Missing or malformed production routing bindings fail closed. No schema, secret, DNS, deployment-policy, runtime-image, or rendered UI changes are included.

## Testing

- The route regression exercises actual Hono handlers, PGlite rows, failed import, retries, activation and committed recovery; it asserts the configured origin, correct forwarded agent host and manual redirect handling.
- The full local Worker/control-plane/PGlite test exercises one provision job, conversation history, reminder/Todo import and the committed switch. It uses the harness's explicit model substitute and is not evidence of a live LLM reply.
- Exact-head route suite: 19 passing tests, 206 assertions. Cloud API TypeScript, all 716 router mounts and the production Worker bundle dry-run passed. Full repository `bun run verify` passed with exit code 0.
- The first clean-checkout verification attempt stopped at a missing generated `@elizaos/prompts` entry in the unrelated UI design-graph test. Required prompts/shared/login builds were then generated and the complete verification passed. No source workaround or gate change was made.

Exact-head full-stack result:

```text
bun --conditions=eliza-source playwright test tests/shared-to-dedicated-upgrade.spec.ts
PASS: gates on hosting runway, copies identity server-side, and moves the conversation
1 passed (1.7m)
```

## Evidence Gate

- [x] N/A - backend transport only; no rendered UI changes.

- [x] N/A - backend transport only; no rendered UI changes.

- [x] N/A - backend transport only; the exercised integration boundary is HTTP and persisted cutover state.

- [x] Browser network reproduction: the authenticated cutover POST returned 503 with code `dedicated_history_import_failed` while its Dedicated target was running.

```text
Observed method: POST /api/v1/eliza/agents/<owned-personal-id>/upgrade-tier/cutover
HTTP status: 503
Response code: dedicated_history_import_failed
Dedicated target status in the staging database: running
```

- [x] Staging Worker tail reproduction: import attempt 1, upstream status 525, messageCount 0, scheduledTaskCount 0, todoCount 0. No credentials or user content are included here.

```text
Operator-filtered staging Worker tail fields:
code=dedicated_history_import_failed
importAttempt=1; upstreamStatus=525
messageCount=0; scheduledTaskCount=0; todoCount=0
```

- [x] N/A - HTTP import transport only; no model, provider, prompt or agent action change.

- [x] The owning integration assertions inspect the target/active binding, preserved canonical conversation, task import receipts and repeated committed recovery. Exact-head command results follow below.

```text
bun test __tests__/eliza-agents-upgrade-tier.test.ts
19 pass, 0 fail, 206 expect() calls
PASS: a personal cutover activates only after the authoritative history import succeeds
Assertions include the persisted cutover state and post-commit retry, using actual PGlite rows.
```

- [x] N/A - no visual surface changed.

## Real LLM-call trajectory

N/A - this patch changes HTTP routing for import/activation, not an agent model, prompt, provider or action. Live Dedicated chat remains a separate staging acceptance check after release.

## Audio / voice walkthrough

N/A - no voice or audio changes.

## Documentation changes

No public documentation changes are required.

## Known gaps / failures

Hosted success is not yet claimed. Staging still serves `e14071f00998f4d10f937015e3060196b991ad4d`. The observed upstream TLS error and local integration results are distinct from a deployed fix.

N/A - no exception requested for this draft.

## Deploy notes

Use the canonical staging release workflow, preserving its source, migration, environment and concurrency guards. No database migration is added by this patch.

### DIFF
--- a/packages/cloud/api/__tests__/eliza-agents-upgrade-tier.test.ts
+++ b/packages/cloud/api/__tests__/eliza-agents-upgrade-tier.test.ts
@@ -29,6 +29,8 @@ process.env.MOCK_REDIS = "1";
 import { eq, sql } from "drizzle-orm";
 import { Hono } from "hono";
 import * as realAuth from "@/lib/auth";
+import { runWithCloudBindings } from "@/lib/runtime/cloud-bindings";
+import { SandboxTransport } from "@/lib/services/eliza-sandbox/bridge/transport";
 import {
   personalSharedAgent,
   personalSharedAgentId,
@@ -187,6 +189,7 @@ const ENV = {
   SHARED_RUNTIME_CONVERSATIONS: cutoverNamespace,
   PERSONAL_DELIVERY_PROJECTIONS: personalDeliveryProjectionNamespace,
   ELIZA_CLOUD_AGENT_BASE_DOMAIN: "dedicated-cutover.test",
+  AGENT_ROUTER_ORIGIN_HOST: "router-cutover.test",
 } as unknown as AppEnv["Bindings"];
 
 let pgliteReady = true;
@@ -365,17 +368,19 @@ function quote(agentId: string) {
 }
 
 function cutover(agentId: string, dedicatedAgentId: string) {
-  return app.request(
-    `/api/v1/eliza/agents/${encodeURIComponent(agentId)}/upgrade-tier/cutover`,
-    {
-      method: "POST",
-      headers: {
-        Authorization: "Bearer steward-test",
-        "Content-Type": "application/json",
+  return runWithCloudBindings(ENV, () =>
+    app.request(
+      `/api/v1/eliza/agents/${encodeURIComponent(agentId)}/upgrade-tier/cutover`,
+      {
+        method: "POST",
+        headers: {
+          Authorization: "Bearer steward-test",
+          "Content-Type": "application/json",
+        },
+        body: JSON.stringify({ dedicatedAgentId }),
       },
-      body: JSON.stringify({ dedicatedAgentId }),
-    },
-    ENV,
+      ENV,
+    ),
   );
 }
 
@@ -1299,6 +1304,10 @@ describe("POST /api/v1/eliza/agents/:agentId/upgrade-tier", () => {
     });
 
     const originalFetch = globalThis.fetch;
+    const workerRuntime = spyOn(
+      SandboxTransport.prototype,
+      "isCloudflareWorkerRuntime",
+    ).mockReturnValue(true);
     const importFetch = mock(
       async (_input: RequestInfo | URL, _init?: RequestInit) =>
         Response.json({ error: "not ready" }, { status: 503 }),
@@ -1622,12 +1631,15 @@ describe("POST /api/v1/eliza/agents/:agentId/upgrade-tier", () => {
         },
       });
       expect(importFetch).toHaveBeenLastCalledWith(
-        `https://${CUTOVER_TARGET}.dedicated-cutover.test/api/conversations/${encodeURIComponent(PERSONAL_C)}/import`,
+        `https://router-cutover.test/api/conversations/${encodeURIComponent(PERSONAL_C)}/import`,
         expect.objectContaining({
           method: "POST",
+          redirect: "manual",
           headers: expect.objectContaining({
             Authorization: "Bearer agent_cutover_transport",
             "X-API-Key": "agent_cutover_transport",
+            "X-Forwarded-Host": `${CUTOVER_TARGET}.dedicated-cutover.test`,
+            "X-Forwarded-Proto": "https",
           }),
         }),
       );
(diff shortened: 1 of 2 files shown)


attentionhead (approved):
**APPROVE** — verified at head `2de10e613f0bd31a8a98b3e33c2f6b6c96fe836f` against merge-base `dd7f61cb77b390beb48e63aa658da67314f7a3a0` (= current `origin/develop`), 2 files, +60/-12.

### Bidirectional control

Green at head, in `packages/cloud/api`:

```
bun test __tests__/eliza-agents-upgrade-tier.test.ts
 19 pass  0 fail  206 expect() calls  [2.34s]
```

Red arm — same suite with only the merge-base `cutover/route.ts` restored (`git checkout $MB -- .../upgrade-tier/cutover/route.ts`):

```
 18 pass  1 fail  170 expect() calls
error: expect(received).toHaveBeenLastCalledWith(...expected)
- Expected  - 6
+ Received  + 7
```

So the new assertions genuinely bind the behavior the PR changes; the regression is not vacuous.

### Correctness of the routing change

- `SandboxTransport` (`packages/cloud/shared/src/lib/services/eliza-sandbox/bridge/transport.ts:17`) exposes `getWorkerAgentRouterFetchTarget` (line 150) and `isCloudflareWorkerRuntime` (line 450) publicly and declares no constructor, so `new SandboxTransport().getWorkerAgentRouterFetchTarget(...)` from the route is a legal call. (The identically named members on `ElizaSandboxService` in `eliza-sandbox.ts` are `private`, but that is a different class and not the one the route imports.)
- The three call sites pass `active` / `target` / `target`, all sandbox rows structurally satisfying `AgentNetworkTarget` (`Pick<AgentSandbox, "id" | "bridge_url" | "health_url" | "node_id" | "bridge_port" | "web_ui_port" | "headscale_ip" | "sandbox_id">`).
- `forwardedHost` is `${agentId}.${ELIZA_CLOUD_AGENT_BASE_DOMAIN}`, which is exactly the host the merge-base code fetched directly via `personalDedicatedAgentApiBase(target, c.env.ELIZA_CLOUD_AGENT_BASE_DOMAIN)`. The destination moves to the origin, the identity the proxy routes on does not — that is the right shape for the 525 described in the body.
- The `localHarness` sentinel cannot diverge from the base resolution: `bootstrap-app.ts:322` wraps every request in `runWithCloudBindingsAsync(c.env, …)`, so `getCloudAwareEnv().ELIZA_CLOUD_AGENT_BASE_DOMAIN` inside `postDedicatedImport` is the same binding the callers read as `c.env.ELIZA_CLOUD_AGENT_BASE_DOMAIN` at lines 392, 549 and 834. It is also belt-and-braces: the local harness runs outside a Worker, where `isCloudflareWorkerRuntime()` (`"WebSocketPair" in globalThis`) is false and the resolver already returns `null`.
- Passing `new URL(url).pathname` is safe here specifically because the sentinel short-circuits first: the loopback base validated by `personalDedicatedAgentApiBase` is the only one that can carry a path prefix, and that case never reaches the resolver. The import URL itself carries no query, so dropping `search` at this call site loses nothing.
- The new fail-closed throw is reachable only inside a Worker, and `wrangler.toml` defines both `AGENT_ROUTER_ORIGIN_HOST` and `ELIZA_CLOUD_AGENT_BASE_DOMAIN` at top-level `[vars]` (266/270) and in `[env.staging.vars]` (614/619) and `[env.production.vars]` (905/920), so no deployed environment reaches `getRequiredWorkerRoutingHost` with an absent binding. Where it does throw, that is deliberate per the contract comment in `transport-contract.ts` — a malformed routing binding must stop the request rather than fall through to a token recipient.
- `redirect: "manual"` is correct for this leg: a same-zone origin fetch that follows a redirect would re-issue the `Authorization` / `X-API-Key` bearer at whatever host the redirect names.

### Non-blocking nit

The suite pins `SandboxTransport.prototype.isCloudflareWorkerRuntime` to `true` for the whole file, so the `null → ?? { url }` non-Worker branch of `postDedicatedImport` has no coverage. One case with the spy returning `false`, asserting the fetch still goes to `https://${CUTOVER_TARGET}.dedicated-cutover.test/...` with no `X-Forwarded-Host`, would lock in that the Node/local path is unchanged by this PR. Not blocking — that branch is a plain passthrough.

The loop over every `importFetch.mock.calls` entry is the right assertion to have added: it is what stops a future change from fixing the initial import while leaving the activation or committed-retry leg on the public host.

ss251 (approved):
Reviewed head 2de10e613f0bd31a8a98b3e33c2f6b6c96fe836f.

The risky part is `packages/cloud/api/v1/eliza/agents/[agentId]/upgrade-tier/cutover/route.ts:191`. Same-zone Worker fetch to the UUID host bypasses the Dedicated proxy. The new path calls `getWorkerAgentRouterFetchTarget(target, new URL(url).pathname)` unless `ELIZA_CLOUD_AGENT_BASE_DOMAIN === "https://"`, then fetch with `X-Forwarded-Host` and `redirect: "manual"`. I followed the three callers — activation :438 (`active`), import :661 (`target`), post-commit :777 (`target`). The test spies `isCloudflareWorkerRuntime` true, wraps `cutover()` in `runWithCloudBindings`, and at :1831 asserts every `importFetch.mock.calls` hostname is `router-cutover.test` with `X-Forwarded-Host: ${CUTOVER_TARGET}.dedicated-cutover.test`.

Verified by reading the diff at this head against the three call sites and the mock.calls loop.

Recommendation: merge

— [free-xai-lane-ss251]

PR #31075, opened 2026-09-11
fix(lifeops): fit populated family controls on phones

### LINKED ISSUE #31073: Family Operations populated agreement clips controls on phones
## Problem

On the deployed Family Operations view, loading an agreement exposes the Pins row and forces the parent grid wider than a 390-pixel phone viewport. Agreement cards and the Calendar sync/Monthly packet buttons end at x=407, clipping the right edge. The empty-agreement app-audit fixture does not reproduce this.

Observed on published candidate `29c5a0df65e18d7e0a3fc7ec63312d55579114a0` at `/lifeops/family`, after successful paired access. The pin form uses minimum grid columns of 110 and 180 pixels plus an auto button; the populated view's minimum intrinsic width propagates through the parent grid.

## Acceptance

- Populated agreement, review, pin and guest controls fit a 390-pixel viewport with no clipped or horizontally inaccessible controls.
- Desktop layout remains usable.
- Exercise the populated view in a real browser and inspect desktop/mobile captures; use measured geometry rather than CSS-class assertions.

Supports #30123. Separate from the authenticated transport fix in #31072.

lalalune:
Opening an uploaded agreement on a 390px phone made the fixed-width Pins row expand every card and clip navigation controls at 407px. Make the pin controls wrap to the available width and keep the parent grid's track within the page.

The browser fixture now includes an agreement, so the pin row is exercised. The regression measures real card/control geometry and fills the pin target. It fails on the old layout at 390px and passes at 390px and1180px with this change. Desktop and phone captures were inspected; the page has no uncaught JavaScript errors.

Fixes #31073. Supports #30123.

Validation on `fd2c78db7a90dc32a70f7d1d3d1ab9261007e00e`:

- Pinned install and root verification:373/373 tasks passed.
- Full real-browser connections/family harness passed after the fix; the prior layout fails only the populated390px boundary check.
- Complete personal-assistant suite:303 files,2648 tests passed, six existing skips.
- Full app audit:226 passed; all four affected Family Operations captures inspected. Standalone populated desktop/phone captures were also inspected.
- All five hosted checks passed for this head. Deployed populated-phone verification remains pending because VPS SSH is timing out before authentication. The running pilot has not been replaced.

This is a responsive layout correction. Synthetic fixtures stand in for provider data; the actual view and shared controls render in Chromium.

[Before/after captures and browser regression logs](https://github.com/elizaOS/eliza/releases/download/pr-evidence-12/bettina-mobile-evidence-fd2c78d.zip). The after captures use the production component with synthetic provider data; public deployment is still pending.

### DIFF
--- a/plugins/plugin-personal-assistant/src/components/family-operations/FamilyOperationsView.tsx
+++ b/plugins/plugin-personal-assistant/src/components/family-operations/FamilyOperationsView.tsx
@@ -462,7 +462,8 @@ function AgreementPanel({
         <div
           style={{
             display: "grid",
-            gridTemplateColumns: "minmax(110px, 0.4fr) minmax(180px, 1fr) auto",
+            gridTemplateColumns:
+              "repeat(auto-fit, minmax(min(100%, 180px), 1fr))",
             gap: 8,
           }}
         >
@@ -1253,7 +1254,13 @@ export function FamilyOperationsView({
       }}
     >
       <div
-        style={{ maxWidth: 1040, margin: "0 auto", display: "grid", gap: 18 }}
+        style={{
+          maxWidth: 1040,
+          margin: "0 auto",
+          display: "grid",
+          gridTemplateColumns: "minmax(0, 1fr)",
+          gap: 18,
+        }}
       >
         <header>
           <p
--- a/plugins/plugin-personal-assistant/test/browser-e2e/family-packet-fixture.ts
+++ b/plugins/plugin-personal-assistant/test/browser-e2e/family-packet-fixture.ts
@@ -1,4 +1,4 @@
-/** Synthetic browser state for editing monthly email; no method can call a live provider. */
+/** Synthetic agreement and monthly-email browser state; no method can call a live provider. */
 import type {
   FamilyOperationsAdapter,
   FamilyOperationsSnapshot,
@@ -32,7 +32,33 @@ export function createFamilyPacketFixture(
   return {
     async load(): Promise<FamilyOperationsSnapshot> {
       return {
-        agreements: { status: "ready", data: [] },
+        agreements: {
+          status: "ready",
+          data: [
+            {
+              artifact: {
+                id: "fixture-agreement",
+                agentId: "fixture-agent",
+                householdId: "default",
+                agreementKey: "synthetic-plan",
+                version: 1,
+                supersedesArtifactId: null,
+                title: "Synthetic parenting plan",
+                originalFilename: "synthetic-plan.pdf",
+                documentId: "fixture-document",
+                mediaUrl: "/api/media/fixture.pdf",
+                mediaFileName: "fixture.pdf",
+                contentSha256: "a".repeat(64),
+                mimeType: "application/pdf",
+                byteSize: 2048,
+                pageCount: 3,
+                uploadedByEntityId: "self",
+                createdAt: "2026-09-20T12:00:00Z",
+              },
+              obligations: [],
+            },
+          ],
+        },
         calendarLinks: { status: "ready", data: [] },
         school: {
           status: "unavailable",
@@ -84,7 +110,9 @@ export function createFamilyPacketFixture(
     },
     uploadAgreement: unsupported,
     decideObligation: unsupported,
-    listPins: unsupported,
+    async listPins() {
+      return [];
+    },
     pin: unsupported,
     unpin: unsupported,
     previewGrant: unsupported,
--- a/plugins/plugin-personal-assistant/test/browser-e2e/run-lifeops-connections-e2e.mjs
+++ b/plugins/plugin-personal-assistant/test/browser-e2e/run-lifeops-connections-e2e.mjs
@@ -597,6 +597,36 @@ try {
     const familyErrors = [];
     family.on("pageerror", (error) => familyErrors.push(String(error)));
     await family.goto(`${baseURL}?scenario=family-packet`);
+    const pinTarget = family.getByRole("textbox", { name: "Pin target ID" });
+    await pinTarget.fill("fixture-acceptance-chat");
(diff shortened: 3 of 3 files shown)


attentionhead (approved):
Reviewed at head `fd2c78db7a90dc32a70f7d1d3d1ab9261007e00e` (merge-base `dd7f61cb77b390beb48e63aa658da67314f7a3a0` against `develop`; 3 files, +70/-5). **Approving** — the two CSS changes are the correct minimal repair for #31073, and the new assertions test measured geometry rather than class names, which is what the issue asked for.

### What the fix actually does

`FamilyOperationsView.tsx` makes two changes:

1. The `AgreementPanel` Pins row goes from a fixed three-track template `minmax(110px, 0.4fr) minmax(180px, 1fr) auto` to `repeat(auto-fit, minmax(min(100%, 180px), 1fr))`. The `min(100%, 180px)` form is the important part: a bare `minmax(180px, 1fr)` still forces a 180px floor, so at 390px minus padding the three tracks plus gaps cannot fit and the row establishes a min-content width wider than the viewport. `min(100%, 180px)` lets each track collapse to the container width when 180px no longer fits, so `auto-fit` wraps to one column instead of overflowing.
2. The outer `<main>` grid gains `gridTemplateColumns: "minmax(0, 1fr)"`. Without an explicit track, the implicit column is `auto`, whose min sizing is min-content — so a wide child propagates its intrinsic width up and widens the whole page. `minmax(0, 1fr)` sets the floor to 0 and contains it.

Both are standard grid-overflow repairs and neither changes desktop layout (at 1180px `auto-fit` still produces three tracks and `1fr` resolves identically to the implicit `auto` track when no child overflows).

### Reproduction

The package's full browser-e2e bundle would not build in my local checkout (several workspace dist/link gaps unrelated to this PR — `packages/ui/dist/login/providers/SolanaProvider.js` fails to resolve `@solana/wallet-adapter-coin98` here), so I could not run `run-lifeops-connections-e2e.mjs` end to end and I am **not** claiming a red/green on the real harness. Instead I reproduced the mechanism directly in a disposable Chromium page carrying the two templates verbatim, as a 2x2 matrix (old/new pins template x absent/present outer track):

```
390px  merge-base (old pins, no outer track): clipped=[SECTION, SECTION]  scrollWidth=391  innerWidth=390
390px  pins template only                   : clipped=[]                  scrollWidth=390
390px  outer track only                     : clipped=[]                  scrollWidth=390
390px  PR head (both)                       : clipped=[]                  scrollWidth=390
1180px all four variants                    : clipped=[]                  scrollWidth=1180
```

So the merge-base templates do overflow at 390px and the head does not, in both directions, and the desktop width is unaffected. Note this is directional, not exact: the probe reproduces the grid templates without the real `Select`/`Input` intrinsic widths, which is why it shows a 1px overflow rather than the issue's cards ending at x=407. It does confirm the sign of the effect and that neither change regresses the wide viewport.

### Test changes read correctly

- The fixture's `listPins` going from `unsupported` to `async () => []` is load-bearing, not cosmetic: without it the Pins row never renders and the new assertion would pass vacuously on the merge-base too.
- The `main section, main button, main input` selector genuinely covers the cards — `Card` renders `<SurfaceCard asChild>` around a `<section>`, so cards are matched.
- `getByRole("textbox", { name: "Pin target ID" })` and `{ name: "Pin", exact: true }` are unambiguous: the only sibling button in the pins list is `Remove`.
- The clipping check runs inside the existing `for (const width of [1180, 390])` loop, so the desktop width is asserted as a control rather than only the phone width.
- `bunx @biomejs/biome check` is clean on both changed test files.

### Non-blocking note

In the simplified probe **either change alone is sufficient** to clear the 390px overflow. That is expected — they fix the same overflow at two different levels — but it means the new assertion would still pass if one of the two were later reverted, so the test does not pin both mechanisms independently. Not worth blocking on; worth knowing if someone later trims one of the two lines as redundant.

ss251 (approved):
fd2c78db7a90dc32a70f7d1d3d1ab9261007e00e

The CSS is the correct scoped fix for populated family controls overflowing on a phone.

FamilyOperationsView.tsx:1258 adds `gridTemplateColumns: "minmax(0, 1fr)"` on the 1040px wrapper. Grid items otherwise keep `min-width: auto`, so a populated agreement row can force the page wider than the viewport. AgreementPanel at FamilyOperationsView.tsx:465 replaces a three-track `minmax(110px, 0.4fr) minmax(180px, 1fr) auto` grid (290px of column mins before gap) with `repeat(auto-fit, minmax(min(100%, 180px), 1fr))`, which wraps and may shrink to 100% of the container.

The e2e now loads a real agreement in family-packet-fixture.ts:36–62 instead of `data: []`, so a phone pass is no longer an empty-state pass. run-lifeops-connections-e2e.mjs:600–629 then flags `main section, main button, main input` whose `getBoundingClientRect()` sits past `window.innerWidth`. That fails if those nodes overflow.

Finding: the assert at run-lifeops-connections-e2e.mjs:621–626 claims the pin form works "without requiring horizontal scrolling", but it only checks that Pin `isEnabled()`. An enabled control that is still clipped would pass. The geometric query is the real layout lock; it also runs after `pinTarget.scrollIntoViewIfNeeded()` (line 602), which can pan a horizontally overflowing field into view before the clip test. Drop that scroll, or assert `scrollX === 0` / no clip before scrolling, if the claim is that no horizontal pan is required.

Not cosmetic: empty `agreements` would not have reproduced the overflow. Fixture `listPins` returning `[]` only enables the pin field; it does not weaken production.

Verified by reading the diff at this head (wrapper and AgreementPanel tracks, populated fixture, e2e clip query). No suite was executed.

Recommendation: merge

— [free-xai-lane-ss251]

PR #31081, opened 2026-09-12
fix(cloud): retain live credentials during provisioning retries

NubsCarson:
A Dedicated provisioning retry can adopt an existing healthy container after an unresolved network probe. The retry previously rotated the inference key first, leaving that container running with a revoked key and every model request failing authentication.

Resolve the retained container handle before environment preparation and preserve its credentials when re-probing it. New cold containers still receive a fresh key. Warm-claim credential handling is unchanged.

Validation: the retained-container regression fails before this change; 42 provisioning tests and 342 assertions pass afterward. The complete `bun run verify` gate passed on the final two-file tree. Live staging reproduced the credential mismatch; the matching narrow fix is applied to the staging provisioner while recovery and app UI acceptance continue.

### DIFF
--- a/packages/cloud/shared/src/lib/services/eliza-sandbox/lifecycle/provision.ts
+++ b/packages/cloud/shared/src/lib/services/eliza-sandbox/lifecycle/provision.ts
@@ -305,12 +305,15 @@ export class SandboxProvision {
     const isWarmPoolProvision =
       rec.organization_id === WARM_POOL_ORG_ID && rec.pool_status === "unclaimed";
     const containerLaunch = resolveSandboxContainerLaunchConfig(rec.agent_config);
+    const provisioningRetryHandle =
+      previousStatus === "provisioning" ? this.buildProvisioningRetryHandle(rec) : null;
 
     // Every claimed row carries the exact managed key owned by its durable
     // handoff fence. Generic environment preparation revokes that key before
-    // writing the replacement, so it is reserved for cold-created rows; warm
-    // claims reuse their persisted environment through restart and attestation.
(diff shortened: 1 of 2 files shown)


ss251 (approved):
Reviewed head c27b45017997501c817c0b5b4f765b5f63bd68d7.

The risky hunk is the widened guard at `packages/cloud/shared/src/lib/services/eliza-sandbox/lifecycle/provision.ts:313`: `!rec.claimed_at && !provisioningRetryHandle` now skips `prepareManagedElizaEnvironment` whenever `previousStatus === "provisioning"`, not only on the old `attempt === 1` path. The hoisted `buildProvisioningRetryHandle(rec)` is also now computed once before the attempt loop, so attempts ≥ 2 reuse the same handle instead of rebuilding from a possibly-updated `rec` — but `handle` is only reassigned on attempt 1, so attempts 2+ take the `attempt === 1 ? null : ...` branch correctly. I traced the skip branch: with the env untouched, the later health probe reaches the retained container with its original `ELIZAOS_CLOUD_API_KEY`, and the new test asserts `updateSpy` never writes `environment_vars` and `apiKeySpy` is never called on that path, which pins the intended behavior.

Non-blocking: the widened skip also fires for warm-pool claims only if `previousStatus === "provisioning"` held — cold restarts (`!claimed_at` with a different previous status) still rotate, which matches the comment's stated intent.

Verified by reading the diff at this head and tracing the skip/handle branches against the added test assertions; I did not execute the suite.

Recommendation: merge

— [free-9router-lane-ss251]

PR #31082, opened 2026-09-12
fix(dedicated): preserve provisioned credentials after restore and repair hosted recovery

NubsCarson:
Restoring a Dedicated agent could let the snapshot's retired Cloud key and URL override its newly provisioned credentials, leaving chat unable to authenticate. Managed containers now prefer explicitly injected Cloud credentials and URL, while retaining saved-config fallbacks when no values are injected.

The hosted recovery screen also sends Cloud app users to `/join`; its old `/cloud/agents` link passed through the same expired-agent gate and repeated the recovery screen.

Validation: full `bun run verify` passed; 61 runtime Cloud-config tests passed, including a regression that failed before this fix; 20 auth relay/gate tests passed; scoped formatting passed. The existing staging agent was recovered with its history intact and returned a real answer in the ordinary local app. The immutable agent image build is running separately; this PR alone is not a deployment receipt.

### DIFF
--- a/packages/agent/src/runtime/eliza-cloud-config.test.ts
+++ b/packages/agent/src/runtime/eliza-cloud-config.test.ts
@@ -842,6 +842,29 @@ describe("unsigned Cloud inference fallback (#20045)", () => {
 });
 
 describe("provisioned cloud container topology (#9887)", () => {
+  it("keeps managed launch credentials when restoring an older Cloud config", () => {
+    process.env.ELIZA_CLOUD_PROVISIONED = "1";
+    process.env.ELIZAOS_CLOUD_API_KEY = "current-managed-key";
+    process.env.ELIZAOS_CLOUD_BASE_URL = "https://api-staging.eliza.app/api/v1";
+    const config: ElizaConfig = {
+      cloud: {
+        enabled: true,
+        apiKey: "revoked-snapshot-key",
+        baseUrl: "https://api-staging.elizacloud.ai/api/v1",
+        agentId: "agent-test",
+      },
+    } as ElizaConfig;
+
+    applyCloudConfigToEnv(config);
+
+    expect(config.cloud?.apiKey).toBe("current-managed-key");
+    expect(config.cloud?.baseUrl).toBe("https://api-staging.eliza.app/api/v1");
+    expect(process.env.ELIZAOS_CLOUD_API_KEY).toBe("current-managed-key");
+    expect(process.env.ELIZAOS_CLOUD_BASE_URL).toBe(
+      "https://api-staging.eliza.app/api/v1",
+    );
+  });
+
   it("repairs a cloud-provisioned config that lost canonical routing fields", () => {
     process.env.ELIZA_CLOUD_PROVISIONED = "1";
     process.env.ELIZAOS_CLOUD_SMALL_MODEL = "small-test";
--- a/packages/agent/src/runtime/eliza.ts
+++ b/packages/agent/src/runtime/eliza.ts
@@ -1213,7 +1213,10 @@ export function ensureProvisionedCloudContainerConfig(
     return false;
   }
 
+  // Managed launch credentials belong to the control plane. A restored config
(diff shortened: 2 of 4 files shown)


attentionhead (approved):
APPROVE — reviewed at head `408d09a9c32da68d9b3c1f0e6a39f7bf28a92450` (4 files, +42/-1).

## What the change does

Two independent fixes, both correctly scoped.

**1. `packages/agent/src/runtime/eliza.ts` — managed credentials win over a restored snapshot.**

`ensureProvisionedCloudContainerConfig` resolved the key and base URL saved-config-first:

```ts
const apiKey = trimCloudCredential(config.cloud?.apiKey) ?? readEffectiveCloudCredential(...)
```

A restored Dedicated agent carries a `config.cloud.apiKey` captured at snapshot time. That key is revoked when the container is re-provisioned, so the truthy stale value shadowed the live `ELIZAOS_CLOUD_API_KEY` the control plane injected at launch, and every inference request authenticated with a dead credential. The same shadowing applied to `baseUrl`, which is why the regression pins both — a snapshot taken on `api-staging.elizacloud.ai` would keep routing there after the container moved to `api-staging.eliza.app`.

Putting `trimCloudCredential(env.ELIZAOS_CLOUD_API_KEY)` / `trimEnvString(env.ELIZAOS_CLOUD_BASE_URL)` at the front of each `??` chain is the minimal correct repair. Two properties make it safe:

- **The precedence flip is confined to managed containers.** This function returns early unless the provisioned marker is set, so an ordinary local config keeps saved-config-first resolution. The env-first rule applies exactly where the env *is* the authority.
- **The fallbacks are preserved, not replaced.** `??` (not `||`) means only a `null`/`undefined` result falls through, and `trimCloudCredential` / `trimEnvString` normalise an empty or whitespace-only env var to nullish. An unset or blank `ELIZAOS_CLOUD_API_KEY` therefore still reaches `config.cloud?.apiKey` and then `readEffectiveCloudCredential`, so a provisioned container that was not given an injected key behaves exactly as before. The `if (!apiKey) return false` guard below is unchanged and still catches the no-credential-anywhere case.

The added regression is a genuine bidirectional control: it sets the provisioned marker, injects `current-managed-key` while the config holds `revoked-snapshot-key`, and asserts on both the mutated config *and* `process.env` — so it fails on the pre-change ordering and cannot pass by accident.

**2. `packages/ui/src/components/auth/CloudPairRelay.tsx` — hosted recovery no longer loops.**

`resolveCloudHostedAgentUrl` built a `/cloud/agents` deep link for every classified host. For `cloud-app` and `legacy-cloud-app` roles that destination sits behind the same agent-auth gate that produced the recovery screen, so an expired session bounced straight back to recovery — a closed loop with no way out. Returning `${base}/join` re-enters the pairing flow, which is the only path that can actually re-establish the session.

The early return is placed after `base` is resolved from `ELIZA_DOMAIN_CONTRACTS[environment].cloudAppOrigin` and before the `agentId` work, so the environment mapping still applies and the now-unreachable agent-path construction is simply skipped for these two roles. Other roles keep the deep link unchanged. The test update covers the environment split deliberately — `app-staging.elizacloud.ai` and `cloud-staging.eliza.app` both resolve to the staging origin, `cloud.eliza.app` to production — which pins the `classified.environment ?? "production"` default alongside the new branch.

## Non-blocking notes

1. Dropping `classified.agentId` for `cloud-app` roles means a user with several agents lands on `/join` rather than the agent they were recovering. That is the correct trade against an unbreakable loop, but if `/join` can accept an agent hint later, carrying it through would preserve the original intent.
2. The credential precedence now differs between provisioned and non-provisioned configs. That is intended and documented by the new comment; worth keeping in mind for any future caller that reuses this resolver outside a managed container.

Verification is a read of the full diff against the current head, tracing both resolver chains and the `??`/trim nullish semantics that keep the fallbacks live; I did not execute the suites, so the test counts in the description are taken as stated rather than independently reproduced.

PR #31086, opened 2026-09-12
fix(auth): unblock tenant recovery and preserve local email verification

NubsCarson:
Fresh email login could stall during session establishment when its organization still lacked a Steward tenant: the returning-user self-heal awaited an unbounded upstream tenant request. Workers now retain that optional repair through `waitUntil`, and tenant requests use the same 10-second timeout as other Steward platform calls. Required identity and session checks still complete before authentication succeeds.

Submitting the email code in the current tab also left consumed-link polling active. While verification or cookie sync was pending, polling could incorrectly show “approved elsewhere” and repeatedly attempt refresh. The local email operation now suspends polling and cross-tab completion handling until it finishes.

Validation: both new regressions failed before their fixes; 30 backend tests and 25 login UI tests passed afterward. Scoped formatting passed. Full `bun run verify` passed on the committed tree. Live staging reproduction used a new account owned by the operator; the final deployed recheck is pending.

### DIFF
--- a/packages/cloud/shared/src/lib/services/steward-tenant-config.ts
+++ b/packages/cloud/shared/src/lib/services/steward-tenant-config.ts
@@ -122,6 +122,7 @@ export async function ensureStewardTenant(
       "X-Steward-Platform-Key": platformKey,
     },
     body: JSON.stringify({ id: tenantId, name: tenantName }),
+    signal: AbortSignal.timeout(10_000),
   });
 
   // A 409 ("already exists") is a recovery path that only needs the STATUS,
--- a/packages/cloud/shared/src/lib/steward-sync-eager-tenant.test.ts
+++ b/packages/cloud/shared/src/lib/steward-sync-eager-tenant.test.ts
@@ -310,6 +310,38 @@ describe("syncUserFromSteward — eager Steward tenant provisioning (#14645)", (
     expect(result).toMatchObject({ id: "user-existing-1" });
   });
 
+  test("Worker sign-in returns while a missing tenant is repaired, and records a late failure", async () => {
+    getByStewardIdImpl = async () => existingUser;
+    const tenant = deferred<void>();
+    ensureStewardTenantImpl = async (organizationId) => {
+      ensureStewardTenantCalls.push(organizationId);
+      await tenant.promise;
+      throw new Error("Steward tenant request timed out");
+    };
+    const background: Promise<unknown>[] = [];
+    const { syncUserFromSteward } = await import("./steward-sync");
+    const signIn = syncUserFromSteward({
+      ...baseParams,
+      executionCtx: { waitUntil: (promise) => background.push(promise) },
+    });
+    try {
+      const result = await Promise.race([
+        signIn,
+        new Promise<null>((resolve) => setTimeout(() => resolve(null), 100)),
+      ]);
+      expect(result).toMatchObject({ id: "user-existing-1" });
+      expect(ensureStewardTenantCalls).toEqual(["org-existing-1"]);
+      expect(background).toHaveLength(1);
+    } finally {
+      tenant.resolve();
+      await signIn;
+      await Promise.all(background);
+    }
+    expect(
+      loggerWarnCalls.some((call) => call.message.includes("Steward tenant request timed out")),
+    ).toBe(true);
+  });
+
   test("FAIL-OPEN: existing-user sign-in still succeeds when the tenant heal rejects", async () => {
     getByStewardIdImpl = async () => existingUser;
     ensureStewardTenantImpl = async (organizationId) => {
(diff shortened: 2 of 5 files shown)


ss251 (approved):
4e41441bd49948648be70f1c9c7e8a02a127598e

Reviewed the diff at this head.

The load-bearing change is in packages/cloud/shared/src/lib/steward-sync.ts around line 893: `ensureStewardTenant` moves inside `settleOffResponsePath(params.executionCtx, ...)` so the repair runs on `waitUntil` instead of blocking `return user`. I checked the failure branch it preserves — the `catch` keeps `logger.warn` with `describeSyncError(error)` and the warn message in the new test (`steward-sync-eager-tenant.test.ts` line ~314) still matches "Steward tenant request timed out" via `loggerWarnCalls.some(...)`, so the observable-warning contract (error-policy:J4) survives the move. The closure captures `organizationId` into a const before the async boundary, avoiding the post-await `user.organization_id` read, which is correct since `user` is returned immediately.

One boundary worth noting, not blocking: with the repair off the response path, the new test's `Promise.race` at 100ms only proves the sign-in promise resolves; whether Workers retains the session exchange's response after the handler returns depends on `settleOffResponsePath` doing its own waitUntil registration — that helper isn't in this diff, so I'm trusting its existing contract.

In steward-login-section.tsx line 1147, adding `loading === "email"` to the polling effect and to the `subscribeStewardEmailLoginComplete` effect at :1235, plus `loading` in both dep arrays: the guard means a local code verification suspends the cross-tab "consumed challenge" recovery until `finishVerify` resolves and the cookie syncs, which matches the new email-login test's ordering — `recoverEmail` not called before `finishVerify`, then sync called with the verification's tokens. The `AbortSignal.timeout(10_000)` in steward-tenant-config.ts:125 bounds the stall that motivates the whole change; 409 handling below it only reads status, so a timed-out body can't corrupt the recovery path.

Recommendation: merge

— [free-9router-lane-ss251]

attentionhead (approved):
APPROVE — reviewed at head `4e41441bd49948648be70f1c9c7e8a02a127598e` (5 files, +82/-12).

## Backend: the self-heal leaves the response path

`syncUserFromSteward` awaited `ensureStewardTenant` inline for any signed-in user whose org still lacked a Steward tenant. The call already had a fail-open `catch`, so an *erroring* tenant endpoint never blocked sign-in — but a *stalled* one did, because `fetch` had no `signal` and the `await` sat directly in the session-exchange path. That is the exact failure the description reports: authentication itself had succeeded, and the user waited on an opportunistic repair.

Two independent changes close it, and they compose correctly:

**1. The repair moves off the response path.** `settleOffResponsePath(params.executionCtx, …)` hands the task to `waitUntil` when the runtime provides it and returns immediately; with no `executionCtx` it falls through to `return task()` and the work is still awaited. So Workers get the unblocking, and any non-Workers caller keeps exactly the previous ordering rather than silently acquiring a floating promise. Hoisting `const organizationId = user.organization_id` is necessary here, not cosmetic — the closure now outlives the enclosing narrowing, and capturing the nullable field directly would not typecheck.

**2. The underlying call is bounded.** `signal: AbortSignal.timeout(10_000)` on the tenant `fetch` means the deferred task cannot accumulate indefinitely either, which matters once it is running under `waitUntil` where nothing else would ever cancel it. The description states this matches the timeout other Steward platform calls already use.

The fail-open contract is preserved exactly: same `try`/`catch`, same `error-policy:J4` rationale, same warning text and org interpolation, so an outage still yields one observable warning and a retry on next sign-in.

The regression is well-constructed for this specific bug. It races the sign-in against a 100ms timer while the tenant promise is held open, asserting the caller returns with the tenant call already dispatched and exactly one promise handed to `waitUntil` — that fails on the pre-change inline `await`. It then resolves the deferred and drains the background work to assert the late rejection is still logged, so the fail-open path is pinned on the deferred side too, not just the fast return. Notably this is a *different* property than the existing "FAIL-OPEN: existing-user sign-in still succeeds when the tenant heal rejects" test below it, which is why both belong.

## UI: local verification suspends cross-tab completion

Submitting the six-digit code consumes the same challenge the link poller watches. With polling still live, the poller saw `consumed` and interpreted the user's own in-tab verification as an approval elsewhere — showing "approved elsewhere" and driving `recoverSharedEmailSession` refresh attempts while the local `verify` and cookie sync were still in flight.

Gating both effects on `loading === "email"` is the minimal fix and is applied consistently: the poll effect and the `subscribeStewardEmailLoginComplete` effect each get the guard, and `loading` is added to both dependency arrays, so the subscription genuinely tears down for the duration and re-establishes afterward rather than going stale. The UI regression pins the ordering end to end — `recoverEmail` never called and no "Link approved" text while `verify` is pending, then a correct navigation to `returnTo` and `sync` called with both tokens once it resolves.

## Non-blocking notes

1. The 10s timeout lands on `ensureStewardTenant` itself, so it also bounds the eager new-signup provisioning path that still runs on the response path. That is almost certainly desirable — an unbounded call there had the same stall exposure — but it is a behavior change for a caller this PR is not otherwise touching, and worth a line in the description.
2. Under `waitUntil` the repair's failure is now only ever visible as a log warning; no request-scoped surface reflects it. Given the existing fail-open policy that is consistent, though it does mean a persistently failing tenant endpoint degrades silently until someone reads the warnings.

Verification is a read of the full diff at the exact head, plus reading `settleOffResponsePath` to confirm the no-`executionCtx` branch still awaits the task. I did not execute the suites, so the stated backend and login-UI test counts are taken as described rather than independently reproduced.

PR #31087, opened 2026-09-12
Promote fresh-account sign-in completion fixes to staging

NubsCarson:
Promote #31086 to fix the fresh-account sign-in failures reproduced during Dedicated onboarding: optional Steward tenant repair no longer blocks Worker session establishment, its upstream request is bounded, and email-link polling cannot override code verification in the current tab.

Both regressions failed before their fixes; 30 backend and 25 login UI tests passed afterward. Scoped formatting passed. Full `bun run verify` passed on the unchanged source. The owner authorized routine CI wait bypass; canonical deployment and migration checks remain in place. This is a staging promotion, with the fresh-account and Dedicated UI recheck still required before production acceptance.

### DIFF
--- a/packages/cloud/shared/src/lib/services/steward-tenant-config.ts
+++ b/packages/cloud/shared/src/lib/services/steward-tenant-config.ts
@@ -122,6 +122,7 @@ export async function ensureStewardTenant(
       "X-Steward-Platform-Key": platformKey,
     },
     body: JSON.stringify({ id: tenantId, name: tenantName }),
+    signal: AbortSignal.timeout(10_000),
   });
 
   // A 409 ("already exists") is a recovery path that only needs the STATUS,
--- a/packages/cloud/shared/src/lib/steward-sync-eager-tenant.test.ts
+++ b/packages/cloud/shared/src/lib/steward-sync-eager-tenant.test.ts
@@ -310,6 +310,38 @@ describe("syncUserFromSteward — eager Steward tenant provisioning (#14645)", (
     expect(result).toMatchObject({ id: "user-existing-1" });
   });
 
+  test("Worker sign-in returns while a missing tenant is repaired, and records a late failure", async () => {
+    getByStewardIdImpl = async () => existingUser;
+    const tenant = deferred<void>();
+    ensureStewardTenantImpl = async (organizationId) => {
+      ensureStewardTenantCalls.push(organizationId);
+      await tenant.promise;
+      throw new Error("Steward tenant request timed out");
+    };
+    const background: Promise<unknown>[] = [];
+    const { syncUserFromSteward } = await import("./steward-sync");
+    const signIn = syncUserFromSteward({
+      ...baseParams,
+      executionCtx: { waitUntil: (promise) => background.push(promise) },
+    });
+    try {
+      const result = await Promise.race([
+        signIn,
+        new Promise<null>((resolve) => setTimeout(() => resolve(null), 100)),
+      ]);
+      expect(result).toMatchObject({ id: "user-existing-1" });
+      expect(ensureStewardTenantCalls).toEqual(["org-existing-1"]);
+      expect(background).toHaveLength(1);
+    } finally {
+      tenant.resolve();
+      await signIn;
+      await Promise.all(background);
+    }
+    expect(
+      loggerWarnCalls.some((call) => call.message.includes("Steward tenant request timed out")),
+    ).toBe(true);
+  });
+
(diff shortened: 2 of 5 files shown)


ss251 (approved):
1266280a9e2e1841b4dd99a68efc2acc2a56d55d — reviewed from the diff alone.

The risky branch is `steward-sync.ts:890`: the tenant self-heal now runs inside `settleOffResponsePath(params.executionCtx, ...)`. The failure mode is that the `try/catch` inside the closure still swallows the error, so the `logger.warn` path is unchanged; the only new risk is whether `settleOffResponsePath` itself can reject (e.g. when `waitUntil` is absent in a non-Workers caller), which would now bubble and fail sign-in where the old inline `try/catch` guaranteed fail-open. The new test mocks `executionCtx: { waitUntil: ... }` and exercises the deferred-tenant path, but nothing in the diff covers a `waitUntil`-less caller, which is the boundary I would want covered before merge. Not blocking; happy to be corrected if `settleOffResponsePath` is documented as rejection-free.

Also checked `AbortSignal.timeout(10_000)` at `steward-tenant-config.ts:125`: the 409 recovery path only needs the status, and a timeout abort does not corrupt that check — the `fetch` throws, the caller's `try/catch` handles it.

The `steward-login-section.tsx` gating (`loading === "email"` before cross-tab completion subscription) is sound: the new test verifies `recoverEmail` is never called while local code verification is pending and that `sync` receives the session tokens afterward.

Recommendation: merge

— [free-9router-lane-ss251]

attentionhead (approved):
APPROVE — reviewed at head `1266280a9e2e1841b4dd99a68efc2acc2a56d55d` against base `staging` at `c9c869dc08e3e6f10800969b8bfad3cf0436d84f`.

This is a branch promotion, so I verified the mechanics of the range rather than re-reviewing the constituent change (I reviewed the source PR #31086 separately at head `4e41441bd49948648be70f1c9c7e8a02a127598e`).

## Range contents match the claim exactly

`staging..head` is **2 commits**:

```
1266280a9e Merge pull request #31086 from elizaOS/codex/dedicated-signin-tenant-recovery-20260912
4e41441bd4 fix(auth): keep optional tenant repair and email polling off sign-in completion
```

That is the #31086 fix commit plus its merge, and nothing else. The description says "Promote #31086" and the range carries precisely that — no unrelated commits have ridden along, which is the failure mode worth checking on a `develop → staging` promotion where the head is a moving branch tip.

The file set is identical to the source PR: 5 files, +91/-13, all confined to `packages/cloud/shared` (`steward-tenant-config.ts`, `steward-sync.ts`, `steward-sync-eager-tenant.test.ts`) and `packages/ui` login (`steward-login-section.tsx` and its email-login test). `4e41441bd4` is the exact head I reviewed on #31086, so the promoted content is byte-for-byte what was verified there.

**Zero `.sql` or migration files in the range**, so this promotion carries no schema change and the migration gates have nothing to act on.

## The promotion cannot revert anything already on staging

`git merge-base --is-ancestor staging head` is **false** — staging is 3 commits ahead of the head on its own line:

```
c9c869dc08 Merge pull request #31083 from elizaOS/develop
c25230e04f Merge pull request #31078 from elizaOS/develop
187a3bb83c Merge pull request #31076 from elizaOS/develop
```

That is expected rather than a defect: all three are prior `develop → staging` promotion merges, so staging is ahead only by merge commits whose content originated on the same develop line this head comes from. The decisive check is the symmetric diff — `git diff staging...head` in the staging direction is **empty**, meaning staging holds no content that is absent from the merge base. Combined with `mergeable: true`, merging this PR is a pure fast-forward of content: it adds the #31086 change and cannot regress #31076, #31078, or #31083.

## Note, non-blocking

`mergeable_state` is `unstable`, i.e. checks are still in flight, and the description states full verification is running on the unchanged source and that CI wait bypass was authorized by the owner. Per this repo's recent history, `develop-full` has been failing on consecutive develop runs with `ui-extended / fixture-e2e` failing identically each time, so chronic redness there is pre-existing and not introduced by this promotion. I am not treating it as blocking, but the production acceptance the description defers — the fresh-account and Dedicated UI recheck — is the step that should not be skipped, since the staging promotion alone does not exercise the Workers `waitUntil` path that #31086's backend fix depends on.

Verification is git-level range mechanics at the exact head and base SHAs above: commit enumeration both directions, the range file set and diffstat, an ancestry test, the symmetric content diff, and a migration-file scan. I did not execute the suites here; the test counts are those reported on #31086.

Issue #17920, opened 2026-08-07
cloud: managed-agent snapshot includes models/cache in stateFiles — bloats/breaks backups


NubsCarson:
## Summary
Found during a managed-agent canary upgrade (2026-08-06): the pre-upgrade snapshot/backup path includes model weights and cache directories in `stateFiles`, ballooning snapshots by gigabytes and making backup/restore slow or failing on disk-constrained hosts. State snapshots should capture agent state (db, config, credentials refs), not re-downloadable model artifacts.

## Impact
- Snapshot duration and size scale with the model cache (multi-GB) instead of actual state (MBs).
- On constrained hosts the snapshot can fail entirely, blocking the upgrade path that depends on a completed backup.

## Suggested fix
Exclude models/cache paths from the snapshot manifest (or add an explicit denylist alongside `stateFiles`), and assert in a test that no entry under the model-cache root lands in the manifest.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

krutftw:
CLAIMING: exclude models/cache from managed-agent stateFiles snapshot manifest + regression test

AI provider/model: xai / grok-4.5
Client / agent tooling: grok
Contribution skill revision: elizaOS/eliza@38d47c8a378de335ed1a4ad2d1aa4b8215ba73b3:packages/skills/skills/contribute-to-eliza
Attribution status: self-reported
— [grok-krutftw]

krutftw:
Maintainer close request: #17985 (merged) implements this issue. Develop auto-close left closedAt null (known pattern; scorer requires closedAt for +4). Please close #17920.

AI provider/model: xai / grok-4.5
Client / agent tooling: grok
Contribution skill revision: elizaOS/eliza@764031118b64b0d9c2023334f1b47e1ae83188cf:packages/skills/skills/contribute-to-eliza
Attribution status: self-reported
— [grok-krutftw]

lalalune:
Closing as completed: PR #17985 (merged) excludes models/cache from snapshot stateFiles; comment requests close (automated triage sweep 2026-08-08)

PR #31089, opened 2026-09-12
fix(cloud): unblock Dedicated recovery and staging onboarding

NubsCarson:
Dedicated recovery could fail before chat opened: legacy SQL upgrades attempted to drop retained state or create foreign keys before their referenced keys; recording the resulting failure could then leave the job stuck because raw SQL timestamps bypassed Drizzle decoding.

- Preserve the legacy summary table without enabling summary generation, and order foreign keys after referenced constraints.
- Decode and validate job creation timestamps before offloading complete payloads to object storage.
- Exclude only the downloadable skills catalog from both snapshot formats, preserving installed-skill locks and other state.
- Recognize the exact staging Pages onboarding origin alongside the existing develop origin, retaining HTTPS, Cloud-build, and host checks.

Validation: SQL and snapshot regressions, raw-timestamp failure, and staging-host tests all reproduced before their fixes. Afterward: 9 real PGlite migration/data tests, 49 snapshot tests (one existing tracked skip), 31 job recovery/storage tests, and 60 onboarding/auth tests passed. SQL build/typecheck, Cloud shared typecheck, and scoped Biome checks passed. Full `bun run verify` passed on final source `323fc6e113a1d40e9b2d5dca84bf8d73ecfa717d`. Image build 34665845922 contains the SQL and snapshot fixes; hosted Gmail-account recovery acceptance is pending.

### DIFF
--- a/packages/agent/src/services/agent-backup-v2-capture.test.ts
+++ b/packages/agent/src/services/agent-backup-v2-capture.test.ts
@@ -856,6 +856,17 @@ describe("streamAgentBackupV2Capture", () => {
     try {
       await fs.promises.mkdir(pglite, { recursive: true });
       await fs.promises.mkdir(path.join(root, "a"), { recursive: true });
+      await fs.promises.mkdir(path.join(root, "skills", ".cache"), {
+        recursive: true,
+      });
+      await fs.promises.writeFile(
+        path.join(root, "skills", ".cache", "catalog.json"),
+        "downloadable catalog",
+      );
+      await fs.promises.writeFile(
+        path.join(root, "skills", ".cache", "lock.json"),
+        '{"installed":"pinned"}',
+      );
       for (const [relativePath, value] of [
         ["a-plain", "hyphen"],
         ["a/nested", "nested"],
@@ -935,7 +946,16 @@ describe("streamAgentBackupV2Capture", () => {
         records
           .filter((record) => record.kind === "data")
           .map((record) => record.entry?.path),
-      ).toEqual(["B", "a-plain", "a/nested", "file-2", "file_1", "z", "ä"]);
+      ).toEqual([
+        "B",
+        "a-plain",
+        "a/nested",
+        "file-2",
+        "file_1",
+        "skills/.cache/lock.json",
+        "z",
+        "ä",
+      ]);
     } finally {
       if (previousStateDir === undefined) delete process.env.ELIZA_STATE_DIR;
       else process.env.ELIZA_STATE_DIR = previousStateDir;
--- a/packages/agent/src/services/agent-backup-v2-capture.ts
+++ b/packages/agent/src/services/agent-backup-v2-capture.ts
@@ -1053,6 +1053,8 @@ function baseStateFileInclude(
   relativePath: string,
   pgliteRelativePath: string | null,
 ): boolean {
+  // Preserve installed-skill state while excluding the downloadable catalog.
+  if (relativePath === "skills/.cache/catalog.json") return false;
   if (
     pgliteRelativePath !== null &&
     (relativePath === pgliteRelativePath ||
--- a/packages/agent/src/services/agent-backup.test.ts
+++ b/packages/agent/src/services/agent-backup.test.ts
@@ -103,6 +103,15 @@ async function writeFixtureState(
     path.join(root, "skills", "active.json"),
     '{"skills":[]}\n',
   );
+  await fs.mkdir(path.join(root, "skills", ".cache"), { recursive: true });
+  await fs.writeFile(
+    path.join(root, "skills", ".cache", "catalog.json"),
+    "downloadable catalog",
+  );
+  await fs.writeFile(
+    path.join(root, "skills", ".cache", "lock.json"),
+    '{"installed":"pinned"}',
+  );
   // Re-downloadable model weights / caches must never enter stateFiles (#17920).
   await fs.mkdir(path.join(root, "models"), { recursive: true });
   await fs.writeFile(
@@ -198,6 +207,8 @@ describe("agent backup manifest", () => {
     expect(pgliteFilePaths).not.toContain("eliza-pglite.lock");
     expect(pgliteFilePaths).not.toContain("pg_stat_tmp/stats.tmp");
     expect(stateFilePaths).toContain("skills/active.json");
+    expect(stateFilePaths).toContain("skills/.cache/lock.json");
+    expect(stateFilePaths).not.toContain("skills/.cache/catalog.json");
     expect(stateFilePaths).not.toContain("pglite/pgdata.bin");
     // #17920: models + cache trees are excluded from the stateFiles manifest.
     expect(stateFilePaths).not.toContain("models/openai.json");
--- a/packages/agent/src/services/agent-backup.ts
+++ b/packages/agent/src/services/agent-backup.ts
@@ -564,6 +564,8 @@ function withFileSetHash(fileSet: AgentBackupFileSet): AgentBackupFileSet {
 }
 
 function baseStateFileInclude(relativePath: string): boolean {
+  // The catalog is downloadable; its neighboring lock.json records installed skills.
+  if (relativePath === "skills/.cache/catalog.json") return false;
   const first = relativePath.split("/")[0];
   if (
     first === MEDIA_DIR_NAME ||
--- a/packages/cloud/shared/src/db/repositories/__tests__/jobs-recovery.test.ts
+++ b/packages/cloud/shared/src/db/repositories/__tests__/jobs-recovery.test.ts
@@ -5,8 +5,9 @@
  */
 
 import { afterAll, beforeAll, beforeEach, describe, expect, spyOn, test } from "bun:test";
-import { eq, type SQL } from "drizzle-orm";
+import { eq, type SQL, sql } from "drizzle-orm";
 import { type RuntimeR2Bucket, setRuntimeR2Bucket } from "../../../lib/storage/r2-runtime-binding";
+import { sqlRows } from "../../execute-helpers";
 import { apps } from "../../schemas/apps";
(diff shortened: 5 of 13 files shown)


ss251 (approved):
Reviewed head e0f604ab322160a32b1465e472278dc74bc09a4b.

Verified the risky hunk in `agent-backup-v2-capture.ts:1056` (mirrored in `agent-backup.ts:567`): the new guard is an exact `relativePath === "skills/.cache/catalog.json"` match that returns false before the pglite branch. I traced the ordering — since `pgliteRelativePath` never equals this path, excluding it first cannot shadow a pglite file, and the guard returns only for this one path, so the sibling `skills/.cache/lock.json` (installed-skill state) still falls through to the existing include logic, which the updated assertions in both test files pin: `lock.json` present, `catalog.json` absent. Edge not covered (non-blocking): an entry arriving as `skills/./.cache/catalog.json` would bypass the exact-string match, but the capture walker emits already-normalized relative paths, so this cannot occur via the current stream. Both fixtures place the cache files under the real state root, so the tests exercise the production path rather than a mock.

Recommendation: merge

— [free-9router-lane-ss251]

attentionhead (approved):
**APPROVE** — verified by reading the complete diff at head `e0f604ab322160a32b1465e472278dc74bc09a4b` (4 files, +36/-1).

## What the change does

`skills/.cache/` holds two files with opposite durability: `catalog.json` is the downloadable skill catalog (re-fetchable, and stale the moment the registry moves), while `lock.json` records which skills are actually installed and pinned. Before this change the whole directory rode into the snapshot; the fix excludes exactly the catalog and keeps the lock.

The exclusion is added to `baseStateFileInclude` in both `agent-backup.ts:564` and `agent-backup-v2-capture.ts:1053`, as an exact-path test placed ahead of the existing prefix checks:

```ts
if (relativePath === "skills/.cache/catalog.json") return false;
```

## Why both copies had to change, and why that is the right call here

This repo carries two independent capture paths with separately-defined `baseStateFileInclude` functions. Patching only one would have produced a snapshot whose contents depend on which path produced it — the classic failure mode for duplicated filter logic. Both are patched identically in this diff, so v1 and v2 agree on the file set. Worth noting for a follow-up that the duplication itself remains: the next exclusion added to one copy and not the other reintroduces the divergence.

## The tests are a genuine bidirectional assertion, not a rubber stamp

`agent-backup.test.ts:209` pins both directions explicitly, which is what makes this reviewable without running it:

```ts
expect(stateFilePaths).toContain("skills/.cache/lock.json");
expect(stateFilePaths).not.toContain("skills/.cache/catalog.json");
```

A blanket `skills/.cache/` prefix exclusion — the obvious shortcut — would fail the first assertion. The paired expectation is what constrains the implementation to the narrow exact-path form actually used.

The v2 suite makes the same point through the ordered manifest, adding `skills/.cache/lock.json` to the expected array between `file_1` and `z` while `catalog.json` is simply absent. That ordering is consistent with the byte-wise sort the existing entries already follow (`B`, `a-plain`, `a/nested`, …, `ä`), so the expectation change is a consequence of the fix rather than an accommodation to it. Both fixtures write both files before capture, so the assertions cannot pass vacuously.

The precedent is consistent with the `models/` and cache-tree exclusions referenced in the surrounding comments (#17920): re-downloadable artifacts stay out of the manifest, installed-state records stay in.

## Non-blocking notes

1. **Exact-path matching is deliberately narrow.** If the catalog is ever sharded or versioned (`catalog-v2.json`, per-registry catalogs), the new file is silently included again. A comment naming that assumption, or a small named constant shared by both copies, would make the boundary explicit.
2. **No migration for existing snapshots.** Backups taken before this change still carry `catalog.json`; restore presumably just overwrites a re-downloadable file, so this looks harmless, but it is worth confirming that a restored stale catalog does not shadow a fresh fetch.

Neither note blocks. Scope of verification: diff read at the exact head plus the fixture/assertion pairs shown above; I did not independently execute the suites, so the passing results stated in the description are the author's.

hermesagent270-commits (approved):
Review at head `323fc6e113a1` (base `4e41441bd499`, 4 commits, 13 files, +280/−21). Note for whoever merges: the two lane approvals are at `e0f604ab`, the first commit only (the 4-file snapshot exclusion). The other three commits, which touch the SQL migration generator, cloud job recovery, and the trusted onboarding origins, had no review at head; this one covers them.

## Verified at head, fully installed tree

**Migration generator + retained `session_summaries` (`1e490e3364df`).** The FK statements collected for newly created tables now go after the phase-4 column/index/constraint alterations instead of directly after `CREATE TABLE`, so a new table can reference a unique key added to an existing table in the same migration. `uniqueFKs` is still populated in phase 3, so the later `diff.foreignKeys.created` loop still skips new-table FKs; I read the full phase sequence and nothing between the old and new position depends on the FKs being present. The registered `sessionSummaries` schema matches `drizzle/migrations/0000_lively_otto_octavius.sql` column for column (types, nullability, defaults, the three indexes, no FKs or uniques), so ordinary startup over a legacy DB emits no ALTER for it.

```text
$ cd plugins/plugin-sql/src && VITEST_LANE=post-merge bunx vitest run __tests__/migration/legacy-session-summaries.real.test.ts
 Tests  2 passed (2)                                   # real PGlite, destructive migrations disabled
$ VITEST_LANE=post-merge bunx vitest run __tests__/migration   # whole directory incl. the *.real.test.ts lane
 Test Files  1 failed | 27 passed | 1 skipped (29);  Tests  1 failed | 205 passed | 6 skipped (212)
```

The one failure (`falls back at runtime when production Postgres startup skipped message-search DDL`) fails identically at the base commit on this host with `SECRET_SALT must be set to a non-default value in production`; it is environment-dependent and not this PR's.

RED controls, each restoring one base file with the new tests kept:
- generator reverted → the composite-key case fails with Postgres `42830` (no unique constraint matching the referenced keys) at `ADD CONSTRAINT "new_membership_account_fk"`, exactly the defect described;
- `schema/index.ts` reverted (summaries unregistered) → the preservation case fails with `Destructive migration blocked for @elizaos/plugin-sql`, i.e. the drop attempt the PR body describes.

**Cloud job timestamps (`3b833633de47`).** `jobs.created_at` is a plain `timestamp()` in date mode, so `mapFromDriverValue` on a raw string decodes it with UTC semantics into a `Date`; an unparseable value now throws `INVALID_JOB_CREATED_AT` before any object-storage key is derived. Ran through Bun's runner as the package does: `bun test src/db/repositories/__tests__/jobs-recovery.test.ts` → 29 pass, 0 fail, including the new raw-text-claim case (`/2020-01-01/` key from a `23:55Z` creation, invalid date rejected with no object written).

**Trusted onboarding origins (`323fc6e113a1`).** The staging Pages alias joins the develop alias inside the same gate: cloud-only branding, exact HTTPS origin, and an agents-collection base are all still required, and the updated table pins `http://`, an explicit port, a path, and non-cloud-only branding as rejected. `packages/ui` targeted suites: 45 passed.

**Snapshot exclusion (`e0f604ab`)** as the lanes reviewed: `packages/agent` backup suites 49 passed, 1 skipped.

Hosted checks at head: Source static smoke:, Billing payment replay E2E:SUCCESS, Subscription authority PostgreSQL:SUCCESS, Browser bridge Windows security / windows-security-contract:SUCCESS.

## Non-blocking

- The legacy-summaries case seeds the table from the new schema definition rather than from the 0000 migration SQL, so it proves idempotence of the registered shape, not that the registered shape equals the legacy DDL; I checked that equality by hand above. Creating the table from `0000_lively_otto_octavius.sql` in the test would pin it.
- The PR body's "hosted Gmail-account recovery acceptance is pending" remains an open evidence row.

**Recommendation: merge.**

```slop-review
{"schemaVersion":"2","projectId":"eliza","artifactUrl":"https://github.com/elizaOS/eliza/pull/31089","headSha":"323fc6e113a1d40e9b2d5dca84bf8d73ecfa717d","provider":"anthropic","model":"claude-fable-5-1","client":"Claude Code","runId":"","traceSha256":"","recommendation":"merge","reproduced":true,"securityRisk":"low","duplicateRisk":"low","splitRisk":"medium","effortBand":"medium","complexity":"medium","impact":"repo-wide","reviewLoad":"medium","recommendedTier":"medium","recommendedThirds":2,"workUnitId":"wu_eliza_dedicated_recovery_31089","confidenceBasisPoints":9300,"valueRationale":"Lane approvals covered only the first commit; verified the three unreviewed commits at head on real PGlite: generator FK reordering and legacy session_summaries retention with two RED controls (42830 and destructive-migration block), the cloud raw-timestamp decode through Bun's runner (29/29), and the staging origin behind the unchanged cloud-only HTTPS gate; the schema registered for the legacy table matches the 0000 migration DDL column for column.","usefulArtifacts":["plugin-sql __tests__/migration at head (post-merge lane): 205 passed, 1 env-dependent failure identical at base (SECRET_SALT)","RED: sql-generator.ts at base → composite-key case fails with Postgres 42830","RED: schema/index.ts at base → preservation case fails with 'Destructive migration blocked'","packages/cloud/shared bun test jobs-recovery: 29 pass","packages/ui targeted suites 45 passed; packages/agent backup suites 49 passed"],"commands":["git fetch origin pull/31089/head && git checkout --detach FETCH_HEAD","cd plugins/plugin-sql/src && VITEST_LANE=post-merge bunx vitest run __tests__/migration","cd packages/cloud/shared && bun test src/db/repositories/__tests__/jobs-recovery.test.ts"],"evidenceUrls":["https://github.com/elizaOS/eliza/pull/31089"],"summary":"Approved at 323fc6e113a1: the three commits the lane approvals did not cover (FK ordering after referenced keys, retained legacy session_summaries schema matching the 0000 DDL, UTC-decoded raw job timestamps, staging onboarding origin) all verified on real PGlite with RED controls; one migration-suite failure is environment-dependent and identical at base."}
```

PR #31092, opened 2026-09-12
fix(cloud): complete Dedicated confirmation and continuity reporting

NubsCarson:
Dedicated onboarding had two acceptance failures: at phone width the hosting explanation pushed both choices below the fold, and a successful deployed browser conversation failed the release report because its validator still expected continuity schema v1.

Shorten the hosting copy while keeping price, required balance, data/backup consequences and the confirmation behavior. Use Start Dedicated when compute starts and Connect otherwise. Build the deployed report with the same strict continuity parser used by the browser producer, including its v2 lifecycle approval checks.

Validation: 82 first-run conductor tests, 12 deployment-authority tests, 34 continuity-contract tests and 34 release-certification tests pass. The Node combine CLI consumes actual producer evidence; malformed history and unauthorized lifecycle evidence remain rejected. Biome passes, and the real local Gmail flow shows both choices without scrolling at 390 × 844. Full bun run verify passed on 7e9e3f946a0e3e2648ccbe8f495cbb451af052e1.

### DIFF
--- a/packages/cloud/scripts/__tests__/pages-deployment-authority.test.ts
+++ b/packages/cloud/scripts/__tests__/pages-deployment-authority.test.ts
@@ -3,6 +3,11 @@
  * contracts with deterministic provider records and public HTTP responses.
  */
 import { describe, expect, test } from "bun:test";
+import { execFileSync } from "node:child_process";
+import { mkdtemp, readFile, rm, writeFile } from "node:fs/promises";
+import { tmpdir } from "node:os";
+import { join } from "node:path";
+import { createCloudLiveContinuityEvidence } from "../../../app/test/cloud-live-continuity-contract";
 import {
   createDeployedRendererProof,
   DEPLOYED_BROWSER_SMOKE_SCHEMA,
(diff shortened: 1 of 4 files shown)


attentionhead (approved):
**APPROVE** — verified at head `25b2e5929b8777a37accde2c6cd47a84ac32d600` (2 files, +17/-14). Draft, but the change is self-contained and reviewable as it stands.

## What changes

`dedicatedAdoptionConfirmationText` is rewritten to cut the Dedicated-adoption confirmation from five dense lines to three short ones. Concretely: the rate line is flipped to lead with the daily figure (`$0.24/day ($0.01/hour)`), the balance line drops the parenthetical runway and trailing deficit, the deficit becomes its own conditional line only when there is one, the `status` line is dropped entirely, and the `disposition` fragment is promoted from a sentence tail into a standalone sentence. The confirm button label becomes contextual — `Start Dedicated` vs `Connect`, driven by `quote.startsCompute`.

## The check that actually mattered

A rewrite of a template string that several suites assert on `toContain` fails loudly if any assertion is left behind, and the diff updates only **one** hunk in `use-first-run-conductor.test.ts` (at ~2028). My local checkout showed what looked like a second block asserting the removed strings (`Current status: error`, `3 days of runway`, `starts Dedicated compute`, `restore its reviewed backup`), which would have been a build break.

That was stale-checkout line numbering, not a real second block. Fetching the test file **at this exact head** and grepping it directly:

```
grep -n 'Current status:|days of runway|starts Dedicated compute|restore its reviewed backup|Confirm and continue|does not start new compute'
→ no matches (3271 lines)
```

Every assertion on a removed string is gone from the file at this head, and the same grep against the source confirms `quote.status` and `quote.minimumRunwayDays` have no remaining reader in it. So the string rewrite and its assertions are consistent.

Worth noting the neighbouring `JoinPage` surface has its own independent copies of this information (`Balance: {{balance}} · Required: {{minimum}} ({{days}} days of runway)` and `Cloud will restore its reviewed backup before switching.`, with a matching test). Those are a different component with a different template and are untouched by this diff — correctly so, but it does mean the runway figure still reaches users on that path, which is part of why dropping it here reads as a deliberate narrowing rather than an oversight.

The conditional deficit line is the one behavioural improvement rather than pure rewording: previously `deficit: $0.00` was rendered unconditionally even when funded, so the funded path loses a line that only ever said "nothing owed".

## Non-blocking notes

1. **`quote.status` is no longer surfaced anywhere in this confirmation.** The existing fixture exercises `status: "error"`, so before this change a user adopting an agent in an error state was told so at the point of confirmation, and now is not. Mobile readability is a fair reason to cut it, but an error state seems like the one status worth keeping — conditionally rendering only non-healthy statuses would preserve that without costing a line in the common case.
2. **`minimumRunwayDays` loses its last reader here.** Not a defect; just worth confirming the field is still justified by the `JoinPage` path above and has not become dead weight on the quote type.
3. The two no-verified-backup branches now read `This starts fresh; no verified backup is available.` and `Your current data stays. No verified Cloud backup is available.` Both are accurate for their respective dispositions, but they are similar enough at a glance that a user could misread which one they got. The distinguishing clause is the first half of each sentence, so it survives — flagging only as wording worth a second look.

Neither note blocks. Scope of verification: diff read plus direct inspection of both files at this head, as shown above; I did not execute the `packages/ui` suites, so any passing results in the description are the author's.

PR #31094, opened 2026-09-12
fix(cloud): finish Dedicated sign-in and lifecycle recovery

NubsCarson:
Dedicated onboarding and lifecycle management could lose Cloud account status, cancel accepted shutdowns, immediately restart an intentionally stopped agent, or leave local Google sign-in stuck on Connecting. This change restores the account display through the trusted control plane, preserves accepted shutdown ownership across worker bookkeeping, and requires an explicit Start after a user-requested stop. Settings keeps queued lifecycle progress visible until the requested state is observed and recognizes a managed Dedicated selection as active.

Direct localhost staging login now uses the existing hosted CLI handoff so Google receives its approved callback. The app preserves safe return destinations, clears account-scoped bindings during explicit switching, and claims the one-time login result only after React Strict Mode startup settles. The regression failed before the fix with no stored session and passes afterward with exactly one claim and cleared busy state.

Validation: 98 account tests, 38 heartbeat/recovery tests, 83 lifecycle dispatch/recovery tests, 83 proxy/pairing/intent tests, 24 recovery tests, and 34 management tests passed during development. The latest authentication and account-routing group passes 133 tests. Its regression failed before the fix: a returned staging session was overwritten by the local runtime's disconnected status. It now verifies the staging account while preserving the local runtime URL and credential; unrelated remote and unstamped local clients retain their existing route. Full repository verification passed on the previous head 3c58019301464ed0009199c83b8ebb048354fb81; the final full run on f5901ab5092f2b4ac3fb7176b0fb14bd312b7915, tree 2be56f3c65317db8d2745ba79a6f180420176a3c passed with exit 0. Current-head hosted checks are pending; the repository owner authorized the accelerated maintainer merge path using this exact local verification and the relevant regression tests. Pending hosted CI is not claimed as passing. The canonical full staging deployment and renderer proof remain required.

Live staging evidence: the requested Gmail account completed Google sign-in, actual Dedicated model replies, and conversation continuity after reload at desktop/mobile widths in the local and hosted apps. A real shutdown saved a 9,948,577-byte private R2 backup and removed the container; the backup passed decryption and manifest verification, and history plus a new model reply survived recreation. Final stay-stopped and explicit Start acceptance awaits deployment. Fresh-account signup and production rollout are not claimed. The broad visual audit was stopped after an unrelated Tasks/iPad timeout; no all-views pass is claimed.

Latest local browser acceptance: after the CLI login return, the requested account remains connected on Settings reload, its Dedicated agent is active, all prior conversation exchanges are present, and a new real reply correctly recalled the project and kickoff day. Inspected the mobile rendering. The failed pre-fix browser request was /api/cloud/status (404); after the fix the authenticated /api/v1/user request succeeds.


attentionhead (approved):
**APPROVE** — verified by reading the full diff at head `0c904b189c4ef80bfc2a211729cb8ed09ba4953f` (9 files, +373/-88), plus an at-head read of `packages/cloud/shared/src/db/repositories/jobs.ts` and `packages/cloud/shared/src/lib/services/provisioning-jobs.ts` to check that the extraction is complete. The description's test counts were not independently reproduced; the reasoning below is from the diff and the at-head sources.

### What the change does

Four inline "release the lifecycle execution owner" updates in `jobs.ts` and one inline "claim" update in `provisioning-jobs.ts` are replaced by a single `updateAgentLifecycleExecutionFence(tx, job, generation, "claim" | "release")` in the new `db/repositories/agent-lifecycle-execution-fence.ts`. The new helper adds one behaviour the old inline sites did not have: on an `agent_suspend` job, when the owner-column write causes `lifecycle_revision` to advance, it carries a matching `agent_compute_stop_intents` row forward to the new revision.

### Why the carry-forward is correctly scoped

This is the part that could silently resurrect a dead shutdown, and the predicate set is what prevents it:

- The intent is matched on `lifecycle_revision = before.revision` — the revision read *inside the same transaction, under `FOR UPDATE`, before* the owner write. An intent that was already stale relative to `before` does not match and stays stale. That is exactly the invariant the inline comment claims.
- `updated.revision !== before.revision + 1` throws rather than writing. So the carry-forward only ever advances an intent across a revision bump this transaction itself caused; if anything else advanced the row concurrently the transaction aborts instead of laundering authority.
- The `inArray(status, ["pending","dispatching","retry","terminal_attention"])` guard keeps a completed or cancelled intent from being revived.
- `release` is included deliberately (per the comment) so a retry does not lose its authority — consistent with the test's claim→release→claim sequence asserting +1/+2/+3.

`claim` is idempotent by the `before.jobId === job.id && before.generation === executionGeneration` short-circuit, which returns before any write, so a re-claim cannot manufacture a revision bump. The shipped test asserts precisely that (`expect(...).toBe(admitted.lifecycle_revision)` after the repeated claim) — a paired assertion that forbids the obvious wrong implementation (unconditionally re-running the update), and it is checkable from the diff without executing it.

The claim predicate `or(isNull(generation), owned)` preserves the old site's semantics. The old code used `IS NOT DISTINCT FROM` for the generation comparison; the new `owned` uses a plain `eq`. That is safe here because a NULL generation is already covered by the `isNull` branch on the claim path, and the release path is only reached with a non-null generation from the caller.

### The heartbeat gate

`agent-sandboxes.ts` adds `hasNoProvisioningOwnerJob([...EXCLUSIVE_AGENT_LIFECYCLE_JOB_TYPES])` to the fenced writeback predicates and moves the write into a transaction that takes `elizaProvisionAdvisoryLockSql(...)` after `configureElizaLifecycleTransaction(tx)`. Since enqueue takes the same advisory lock before capturing the lifecycle revision, a probe and an accepted shutdown are serialised on that lock rather than racing — which is the stated bug (a healthy heartbeat bumping the revision out from under a queued user shutdown). The corresponding test drives it through a real `Bun.serve` probe and asserts the revision is unchanged until `execution_quiesced_at` is set, then bumps exactly once. That is a genuine before/after control, not a restatement.

### UI gate

`canPollCloudStatus` no longer treats "persisted active server is remote" as a blanket no-poll. It now polls only when the client's base URL already matches the persisted target *and* `hasDirectCloudAccountTransport(client)` — so during a switch, while the client still points at the previous runtime, it stays silent. The `.replace(/\/+$/, "")` on both sides makes the comparison trailing-slash insensitive, and `activeServer.apiBase` being undefined yields `false` rather than a spurious match. The new parametrised test covers staging plus both loopback origins, and the pre-existing "plain web page on a dedicated agent" case is kept honest by the added `setPageOrigin("https://agent.example.test")`.

### Non-blocking notes

1. **One release site was not migrated.** At head, `provisioning-jobs.ts` still contains an inline owner-column release in the admin-canary cleanup path (around line 5781, guarded by `AdminCanaryCleanupCommitError`), while all four `jobs.ts` sites and the claim site now route through the helper. I do not think this is a defect — a canary job is not `agent_suspend`, so the helper's only added behaviour would be a no-op there, and the predicates are otherwise identical. But sibling copies of a fence body are exactly the thing that drifts, so migrating it (or leaving a comment saying why it is deliberately inline) would make the invariant hold by construction rather than by job-type coincidence.
2. **`throw` on an unexpected revision delta is a hard failure inside a lifecycle transaction.** `"Lifecycle execution bookkeeping advanced an unexpected revision"` aborts the claim/release. That is the right default, but it is worth confirming the job retry path treats it as retryable rather than terminal, since the condition is by definition concurrent and transient.
3. **The helper throws on a missing `agent_id`** rather than returning `undefined`. All current callers are behind `hasAgentLifecycleFence(...)`, so the throw is unreachable today; it is a fine assertion, just noting it is a stricter contract than the inline code it replaces.

None of these block; the fence semantics and the shutdown-preservation invariant are sound as written.

ss251 (approved):
Reviewed head 0c904b189c4ef80bfc2a211729cb8ed09ba4953f.

The risky hunk is the intent carry-forward in agent-lifecycle-execution-fence.ts:55-73. I checked it against what it replaces: the `owned` release predicate (:35) is identical to the four inline jobs.ts updates it supersedes — a release after a takeover still no-ops to `undefined` exactly as before; the claim arm `or(isNull(generation), owned)` (:49) is equivalent to the old provisioning predicate given the NULL arm; and the idempotent-claim early return (:32) fires before any revision write, so a repeat claim cannot double-bump. The +1 strictness (:56) rejects unexpected advances, and matching intents at `before.revision` keeps stale intents stale — the new test pins claim +1, release +2, repeat-claim unchanged.

Non-blocking: the new tests are sequential, so the advisory-lock serialization between heartbeat and enqueue rests on key parity this diff does not show.

Verified by reading the diff and the pre-image of each replaced update plus the old claim predicate at this head; the pinned revision behaviors come from the PR's own DB tests, which I did not execute.

Recommendation: merge

PR #31096, opened 2026-09-12
fix: preserve Dedicated state through stop and startup recovery

NubsCarson:
Reopening a stopped Dedicated agent currently hangs until a generic startup timeout. A failed fresh shutdown snapshot can also reuse an older verified backup, causing recent conversation history to disappear on resume.

This change shows an explicit stopped state with Start agent, keeps the selected account and agent, and resumes the existing startup flow after the agent is running. Suspend and sleep now refuse to remove live compute if current state cannot be captured. Both snapshot formats exclude the generated plugins/.runtime-imports subtree while retaining user-owned plugin state; that cache caused the observed capture budget failure.

Validation: 151 focused startup/recovery tests; 58 agent backup tests passed with one existing skip; 60 lifecycle tests and 227 assertions passed. The capture budget and stale-backup regressions failed before the fix. Full bun run verify passed for the final tree efc168a7b2e37392d3e558b261a811e739d12358. Desktop and 390px mobile stopped-state UI were inspected, and explicit Start produced a real model reply.

Live complete-history acceptance remains pending: the pre-fix staging stop/start lost three recent user turns after restoring an older backup. The current agent is preserved; the backup fix still needs staging activation and a new complete-history restart check. Production is unchanged.


attentionhead (approved):
APPROVE — verified by reading the full diff at head `2b055f830a494e8f5e57a3507265c56aa4e7f76a` (11 files, +493/-45) plus `packages/ui/src/api/client-cloud.ts` fetched at that exact SHA. I did not execute the suites; the test counts in the description are not independently reproduced here.

**What I checked**

1. **The new state is a genuinely separate classification, not a relabelled error.** `describeStoppedDedicatedCloudAgent` returns non-null only on the conjunction `status === 409 && code === "agent_stopped" && isDedicatedCloudAgentBase(clientBaseUrl)`. The shipped table-driven test is the control here and it is a *paired* assertion: it walks four near-misses — wrong status (401/`agent_stopped`), wrong code (409/`lifecycle_conflict`), non-dedicated base (`self-hosted.example`), and a *personal* shared-Eliza base (`…/agents/personal:abc`) — and requires `null` from each, while the positive case in `runPollingBackend` requires the full `agent-stopped` state. That pair forbids the obvious wrong implementation (widening on status alone, or on any cloud base), which is what would have hijacked unrelated 409s into a dead-end "shut down" screen.

2. **The stopped path does not destroy the binding.** This is the part that actually matters for a recoverable state, and both new integration tests assert it negatively rather than just asserting the happy path: `clearPersistedActiveServer`, `client.setBaseUrl` and `client.setToken` are each asserted *not* called, and in the warmup test `client.startAgent` is asserted not called. So a deliberate shutdown no longer degrades into the agent-reselection / first-run-reset flow. `getAuthStatus` is asserted called exactly once — the classification is immediate, not after the poll budget burns down.

3. **The warmup race refactor is behaviour-preserving for the pre-existing branches.** `firstDecisive` changes from a string tag to the probe object, and the three consumers move from `=== "conversations"` / `=== "auth-required"` to `?.kind === "serving"` / `?.kind === "auth-required"`. The `statusProbe.then(ready => ready ? "status" : null)` arm still yields a bare string, and the subsequent `if (statusReady)` branch is unchanged — the only new early return is `kind === "agent-stopped"`, which mirrors the existing two exactly (`statusAbort.abort(); await statusProbe;` before acting, so the sibling request is settled and not leaked). The second-chance path after the race adds the same check against the awaited `probe`. The `cloud-warm` test asserts `statusAborted === true`, which is what makes that abort ordering a tested invariant rather than an incidental line.

4. **The reducer guard is phase-scoped, and the recovery edge is tested.** `AGENT_STOPPED` is handled before the `switch` but only from `polling-backend` or `starting-runtime`, so it cannot fire out of `restoring-session` or overwrite a terminal state. The `it.each` covers both admitted phases and, in the same case, asserts `RETRY` from the resulting error state lands on `restoring-session` — i.e. the screen's "Retry connection" is a real path back into startup, not a no-op.

5. **The button actually starts the agent.** `waitForCloudAgentRunning` (client-cloud.ts:4179, read at head) issues `resumeCloudCompatAgent` *first* and only then enters the bounded status poll, so the name understates it — this is resume-then-wait, and the `StartupFailureView` call is correctly the whole start operation. Its transport-failure branch swallows a non-terminal resume failure and lets the poll be the authority, while a `!resume.success` envelope throws `CloudAgentWakeError`; that surfaces on the new `role="alert"` line.

6. **Duplicate-start guard is the ref, not the disabled attribute.** `if (!error.cloudAgentId || startAttempt.current) return;` sets `startAttempt.current` synchronously before the first `await`, so two clicks dispatched in the same tick collapse to one call — `setStarting(true)` alone would not have, since the re-render is async. The test clicks twice back-to-back and asserts `toHaveBeenCalledTimes(1)`, which is exactly the assertion that distinguishes the ref guard from the disabled-prop guard. The unmount cleanup aborts the in-flight controller, and every post-await state write is gated on `!attempt.signal.aborted`, so there is no setState-after-unmount. On rejection the `finally` clears the ref and `starting`, leaving the start retryable rather than wedged.

7. **The stopped screen is stripped of the wrong affordances.** `bugReport && !stopped` removes "Report a bug" (a deliberate shutdown is not a defect), the technical-details disclosure is hidden, the badge drops from `statusDanger` to `statusMuted` with a `Power` glyph, and the heading loses `text-destructive`. The test asserts `queryByTestId("startup-start-over")` is null on the failed-start path — the first-run reset stays unreachable even when starting fails, which is the destructive action you would least want one click away here. All seven new strings are present in `en.json`.

**Non-blocking notes**

1. *Resume routing when the client base is the stopped proxy.* At the moment `startAgent` runs, `client.getBaseUrl()` is the dedicated agent base (`https://agent-123.cloud-staging.eliza.app`) — that is precisely what `isDedicatedCloudAgentBase` matched on. `runCloudLifecycleAction`/`directCloudRequest` are expected to resolve the Cloud control plane independently of that base (the `resolveDirectCloudClientApiBase` machinery), but the sibling `launchCloudCompatAgent` shows the shape of the fallback: when neither the direct-cloud transport nor `isDirectCloudBase` applies, it falls back to `this.fetch("/api/cloud/compat/…")`, i.e. through the current base — which here is the shut-down agent's own proxy. If the resume path can ever reach that same fallback, the start request would be sent to the very endpoint that just returned `agent_stopped`, and the user would see the poll time out rather than a clear "sign in to Cloud". Worth confirming the direct-cloud transport is guaranteed present on this screen, or asserting it in the test.

2. *No timeout affordance on the progress line.* `waitForCloudAgentRunning` is bounded by `CLOUD_AGENT_WAKE_TIMEOUT_MS`, and a cold boot is advertised as "a few minutes". During that window the only live control is a disabled button and a `role="status"` line; there is no cancel. The unmount abort covers navigation away, but a user who changes their mind has to wait out the timeout to get the screen back.

3. *`error-policy:J4` comment placement.* The comment sits on the `catch` in `startAgent` and reads as justification for keeping the stopped agent; the analogous J4 comment in `waitForCloudAgentRunning` justifies swallowing a transport failure. Two different rationales under one policy tag is fine, but the UI-side one would be clearer as "surface, do not reset".

None of these change the decision: the classification is correctly narrow, the recoverable state preserves the agent binding, and the new control is a real start rather than a poll.

ss251 (changes requested):
2b055f830a494e8f5e57a3507265c56aa4e7f76a

FINDINGS

1. startup-phase-runtime.ts:732,743,751,761 — `firstDecisive` is `PassthroughProbe | "status" | null` (the status probe maps to the `"status"` literal at :732), but the code now reads `firstDecisive?.kind`. Optional chaining still dereferences `.kind` on the string member — TS2339; `tsc --noEmit` on packages/ui should fail this file. Runtime JS is unaffected. Fix: map the status probe to an object or narrow the literal before the `.kind` reads.

2. StartupFailureView.tsx:214-218 — the `startError` alert is not gated on `stopped`. Input: "Start agent" fails ("Sign in…" lands in startError), user clicks "Retry connection", startup then fails as agent-timeout — the same component instance re-renders under a non-stopped reason and still renders the superseded alert (unless StartupShell remounts per error reason; not shown here). Gate it on `stopped` or clear `startError` when `error` changes.

Also verified: negative `it.each` matrix for the classifier (dedicated-cloud-agent-error.ts:11) covers wrong status, wrong code, self-hosted and personal-shared bases; both dispatch paths (startup-phase-poll.ts:1093, runtime warmup :761) preserve the binding — clearPersistedActiveServer/setBaseUrl/setToken asserted untouched; duplicate-click guard is synchronous via ref (StartupFailureView.tsx:104); unmount aborts an in-flight start. Real `waitForCloudAgentRunning` is mocked — its polling/abort/timeout behavior is unverified. No dependency, lockfile, CI, secret-access or telemetry changes in the diff.

VERIFICATION: read the diff at this head only; no commands run, no files outside the diff opened. Live execution blocked: no disposable sandbox on this host.

Recommendation: fix

PR #31099, opened 2026-09-12
fix(lifeops): include approved agreement pins in initial response context

### LINKED ISSUE #31085: fix(lifeops): include pinned agreement obligations in ordinary response state
## Observed defect

On the live Bettina pilot, a fresh owner-authenticated chat correctly returns `NO_PINNED_OBLIGATION` without a pin. A separate fresh chat with a persisted chat pin to the synthetic agreement returns the same result despite one approved, page-cited obligation. Rejected and proposed obligations exist as negative controls. Both model requests completed HTTP200 with no reported failure; the temporary pin was removed and removal verified.

The provider is registered as dynamic but does not opt into `alwaysInResponseState`. Core Stage1 uses an explicit list plus opted-in plugin providers. Ordinary response turns can therefore omit the pinned agreement entirely before answering. Existing provider tests call `get` directly and do not exercise response-state selection.

## Acceptance

- Approved obligations from an active owner agent/chat pin enter ordinary response state without relying on optional provider discovery.
- Another chat, removed pins, proposed/rejected obligations, and unauthorized audiences do not gain that context.
- A real runtime/state-composition regression fails before the correction and passes after it; preserve the owner disclosure gate.
- After deployment, repeat the owner live-model comparison and retain the complete test input/output privately, with sanitized evidence in #30123.

The current live observation establishes failure, not the complete root cause; verify ownership and provider selection in the regression before changing behavior. No raw family document or credential is involved.

lalalune:
An owner-approved agreement could be visibly pinned to a chat while an ordinary response still received no agreement context. The live pilot replied `NO_PINNED_OBLIGATION` in both an unpinned chat and a fresh chat with an approved, page-cited obligation pinned to it.

The agreement provider now participates in ordinary response-state composition. Its existing owner-only disclosure gate, per-turn evaluation, and pin-target checks remain in place.

A regression uses the real PGlite agreement store and core response-state composer. It failed before the change and now verifies approved text with page citations, rejection exclusion, chat isolation, removal, agent-wide pins, and shared-room denial. The test harness stubs the package owner-access helper; live owner authentication is verified separately on the pilot.

Fixes #31085.

A second recorded live baseline on the mobile pilot captured seven model calls: the initial two lacked the approved pin, later calls included it, and an unnecessary `VIEWS` action failed because no view client was attached. This narrows the regression to initial response composition rather than absence from every later planner stage. The temporary pin was removed and independently read back.

Validation:

- Pinned `bun install` passed; root `bun run verify` passed all 373 tasks on `e0c72296cf9d040a9d9cfb84ad2e37448445ab0c`.
- Focused real-PGlite integration: 7/7 passed, including initial response context, citations, scope, removal and audience denial.
- Combined pilot `aa65e08acd40d808dad5df6d32e5f19ca2905920`: local and Linux root gates passed; Linux production builds completed and all 3,577 release-file hashes were verified. Deployment and independent public HTTPS entry hashes match; stored agreements/calendar controls/monthly packet were preserved.
- Live fixed-version pinned response: "Parent Example A handles the Monday library pickup at 15:00. (source page 1-1)". The initial model prompt contains the approved obligation and excludes rejected/proposed obligations; no tool actions were taken. The temporary pin was removed and independently read back.
- A fresh unpinned control contains no pinned-agreement block, but the provider hit its model-output limit and returned a retrieval error. This control is not a full live-model pass and remains part of broader MVP acceptance #30123. Full traces remain private because they contain owner context; [sanitized synthetic live evidence](https://github.com/elizaOS/eliza/releases/download/pr-evidence-12/bettina-pin-live-evidence.zip) includes the before/after inputs, replies and failed control.
- Full personal-assistant suites passed: source branch 2,648 tests across 303 files; Linux combined candidate 2,708 tests across 308 files. Each has six existing skips. All five hosted checks passed on the exact PR head: Source static smoke, Windows security, billing payment replay, PostgreSQL subscription authority, and the aggregate gate (run 34674569244).

UI evidence: N/A for this provider-composition change. The separate deployed mobile repair is documented in #30123.

### DIFF
--- a/plugins/plugin-personal-assistant/src/lifeops/household/agreement-knowledge.pglite.integration.test.ts
+++ b/plugins/plugin-personal-assistant/src/lifeops/household/agreement-knowledge.pglite.integration.test.ts
@@ -11,8 +11,11 @@ import path from "node:path";
 import { resolveKnowledgeGraphService } from "@elizaos/agent";
 import {
   type AgentRuntime,
+  attestAuthenticatedApiDeliveryAudience,
+  ChannelType,
   documentsPluginCore,
   type IAgentRuntime,
+  type Memory,
   type Plugin,
   Service,
   ServiceType,
@@ -21,6 +24,7 @@ import {
 import { SELF_ENTITY_ID } from "@elizaos/shared";
 import { afterAll, beforeAll, describe, expect, it } from "vitest";
 import { LocalFileStorageService } from "../../../../../packages/agent/src/services/file-storage.js";
+import { composeResponseState } from "../../../../../packages/core/src/services/message/provider-state.js";
 import {
   createLifeOpsTestRuntime,
   type RealTestRuntimeResult,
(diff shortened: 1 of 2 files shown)


attentionhead (approved):
APPROVE — verified by reading the full diff at head `e0c72296cf9d040a9d9cfb84ad2e37448445ab0c` (2 files, +102/-0), plus `packages/core/src/services/message/provider-state.ts` and `plugins/plugin-personal-assistant/src/providers/agreement-pins.ts` fetched at that exact SHA. I did not execute the PGlite integration suite; the description's results are not independently reproduced here.

The entire production change is one line — `alwaysInResponseState: true` on `agreementPinsProvider` — so the review is really two questions: is that the sanctioned mechanism for this problem, and does always-composing the provider widen who can see pinned obligations.

**1. The flag is the documented fix for exactly this catch-22, not a workaround.**

`agreementPinsProvider` is `dynamic: true`, and dynamic providers are filtered out of response-state composition unless core names them explicitly — the comments at `provider-state.ts:28` and `:39` show `FACTS` and `CURRENT_TIME` being special-cased by name for precisely that reason. `alwaysInResponseState` is the generic escape hatch introduced so plugins don't need that favour; the doc comment at `:48` reads "Composed regardless of selected contexts, so a plugin's dynamic provider reaches Stage 1 without core naming it," and the gate at `:60` is `if (provider.alwaysInResponseState && name && !provider.private)`. This provider sets neither `private` nor an empty name, so it satisfies the condition as written. `plugin-lifecycle.ts:441-451` corroborates the intent from the other side: a provider declaring no `contexts`/`contextGate` is warned and defaulted to `["general"]` *unless* `alwaysInResponseState` is set — which is this provider's situation exactly.

**2. There is established in-plugin precedent, including a test that mandates the flag.**

The pattern appears across 37 files, and two of them are sibling providers in this same plugin: `pending-approvals.ts:71` and `first-run.ts:107`. `pending-approvals.ts:10` documents the identical rationale ("`alwaysInResponseState` because Stage-1 …"), and `test/scenarios/approval-queue-pending-visible-to-planner.scenario.ts:110` fails the scenario outright when the flag is absent, with the message "must be alwaysInResponseState so Stage-1 routing sees the queue before contexts are selected". So this PR makes `agreementPins` consistent with its neighbours rather than introducing a new convention.

**3. Always-composing does not widen exposure, because authorization lives inside `get()`.**

This was my main concern: flipping a provider to always-on is a one-word change that can silently turn a gated signal into an ambient one. It doesn't here. `get()` re-derives the principal on every call — `hasOwnerAccess(runtime, message)` selects `SELF_ENTITY_ID` for the owner and otherwise falls back to `message.entityId` — then delegates to `activePinnedContextForPrincipal({ principalEntityId, roomId })`, which filters by both principal and room. Authorization is therefore per-invocation and unchanged; the flag only affects *whether the provider is asked*, never *what it is allowed to answer*. The file header states the same invariant ("Pins select already-authorized owner knowledge; they never participate in guest authorization or widen a resource grant"), and the code matches it.

**4. The shipped test is a genuine multi-way paired control, which is what makes this reviewable without running it.**

The new integration test does not just assert the happy path; it pins the boundaries in both directions against the same fixture text, so an implementation that merely "always includes the pins" would fail it:

- owner, pinned room → obligation text **and** `"source pages 4-5"` present;
- an unrelated model interpretation asserted **absent**, so the citation-scoped rendering is checked rather than raw artifact dumping;
- same owner, **different** room, chat-scoped pin → **absent** (room boundary);
- after `unpin` → **absent** (lifecycle, not just read-path);
- re-pinned at `targetType: "agent"` → **present** in the other room (agent scope genuinely differs from chat scope);
- a second participant added to the pinned room → **absent even for the owner** (audience boundary).

That last assertion is the security-relevant one and the one I would have asked for had it been missing. The `attestAuthenticatedApiDeliveryAudience(…, { kind: "owner_session" })` call before each compose means the audience attestation is part of the fixture rather than assumed.

Test hygiene is also sound: `pin` is reassigned after the first pin is explicitly unpinned, and the `finally` unpins the surviving pin, so no pin leaks into the sibling tests in this file; `ELIZA_ADMIN_ENTITY_ID` is captured and restored in the same `finally`.

**Non-blocking notes**

1. *Every response turn now pays for a pin lookup.* With the flag set, `get()` runs on all response-state compositions for any agent loading this plugin, and each call reaches `activePinnedContextForPrincipal` — a service round-trip — before the `views.length === 0` early return. `cacheScope: "turn"` bounds it to once per turn, and the empty case returns cheaply, but agents with no pins now carry a query they previously did not. If that shows up, a cheap "does this agent have any active pins at all" guard ahead of the per-principal query would remove it without touching the authorization path.

2. *Setting restore coerces `undefined` to `null`.* The `finally` writes back `typeof previousOwner === "string" || typeof previousOwner === "boolean" ? previousOwner : null`, so if `ELIZA_ADMIN_ENTITY_ID` was previously unset the test leaves it explicitly `null` rather than absent. Harmless for the suites in this file as written, but "unset" and "null" are not always the same thing to a settings reader, and this is the kind of difference that surfaces later as a test-order dependency.

3. *Draft status.* Submitting against the draft at this head; the flag is the whole production change, so I would expect it to survive to ready-for-review unchanged, but re-request if the provider's gating or `contexts` declaration moves.

None of these affect the decision: the mechanism is the documented one, it matches two sibling providers in the same plugin, and the authorization boundary is enforced inside `get()` and is now explicitly pinned by tests in four separate directions.

ss251 (approved):
e0c72296cf9d040a9d9cfb84ad2e37448445ab0c

Findings:

The production change is a single flag: `alwaysInResponseState: true` on `agreementPinsProvider` (agreement-pins.ts:19), which is what actually places owner-approved agreement pins into the initial response context the title promises. The new PGlite integration test (agreement-knowledge.pglite.integration.test.ts:397-493) pins down the matrix that makes this safe: a chat-scoped pin reaches owner-turn state (:443-448) while an adjacent room does not (:449-451); unpin removes it (:457-459); an agent-targeted pin reaches any room (:466-468); and — the privacy-critical case — once a guest participant joins the room, pin content is suppressed even in the owner's own composition (:477-479). Unsupported model interpretation text is asserted absent (:448). `finally` unpins and restores `ELIZA_ADMIN_ENTITY_ID`, so no state leaks across tests.

Two things the diff alone cannot close. First, `composeResponseState` (packages/core/src/services/message/provider-state) is not in the diff, so the flag-honoring logic is unverified at source level; the test is the evidence, and if the flag were ignored its first assertion fails loudly rather than masking. The asserted strings also depend on an artifact fixture defined above this hunk, outside the diff — again CI-visible if wrong. Second, the test drives the state composer directly, not the live message pipeline: a regression in how message handling invokes `composeResponseState` would not fail this test. A follow-up worth noting, not a blocker for a one-flag behavioral fix. No lockfile, dependency, CI-permission, lifecycle-hook, binary, secret-access, or test-weakening changes in the diff.

Verified by static read of this diff at this head; live execution blocked: no disposable sandbox on this host, so no commands were run.

Recommendation: merge

PR #31100, opened 2026-09-12
fix: recover Dedicated sign-in and guide local agent selection

NubsCarson:
Dedicated users could finish Google sign-in or refresh their session and still see a reopen screen or disconnected Cloud account. Local account switching also removed the selected agent but returned users to chat against the local runtime, where their Dedicated history was absent and messages could fail.

Session recovery now finishes through its own refresh event, resumes after a different authenticated session arrives, and rechecks account authority after a guarded refresh. Local sign-in defaults to Cloud agent selection; switching accounts cannot return directly to an unbound home/chat screen. Explicit destinations remain intact when the existing agent binding is retained. Logout cancellation, agent ownership, pairing, and OAuth validation remain enforced.

Validation: 116 focused recovery/account tests across ten suites passed, plus 37 local handoff and same-tab login tests across two suites. Five landing regressions failed before the final routing change. Full bun run verify passed on final head 8d52b2e05b9a2948e1db3ca14108386e08d8a260 (tree 303a011abcaab4fb5b97090b1082a27bf78897f0); the working tree is clean. At owner-authorized promotion, hosted source/billing checks are still in progress, subscription authority passed, and the Windows check is cancelled; none of those incomplete checks is represented as passing. Eight affected Chat/Settings captures and packaged OCR passed; desktop, mobile, and tablet images were inspected. These shell captures do not substitute for authenticated flow acceptance.

Real Google sign-in in the authorized Chrome profile now returns to the local Cloud agent list, and selecting the existing Dedicated agent reaches its real authenticated API and original conversation. A real post-login reply recalled the previously saved code and color. Final-head local Google sign-in, explicit existing-agent selection, real reply and exact preservation of the preceding 18 messages passed (20 messages total). Local reload now preserves all 20 messages exactly at 390x844 and 1440x900, with both screenshots inspected. The new staging rollout remains in progress. The prior staging release separately passed a real stop/Start with history preservation. First-time Google signup remains unverified. Production is unchanged.

Follow-up to #31096.

Release state: merged to develop64d6636560e, promoted by #31101 to staging11694a1b596; canonical staging run34676934969 is in progress. The final hosted PR check later reported cancellation for source/Windows/billing, with its aggregate failure explicitly caused by cancelled lanes. Local verification remains passing; hosted CI is not reported green.

### DIFF
--- a/packages/ui/src/cloud/public-pages/pages/login/loopback-cloud-login-section.test.tsx
+++ b/packages/ui/src/cloud/public-pages/pages/login/loopback-cloud-login-section.test.tsx
@@ -96,10 +96,33 @@ it.each([
     expect(destination.origin).toBe("https://staging.eliza.app");
     const returnTo = new URL(requiredReturnTo(destination));
     expect(returnTo.origin).toBe("http://localhost:21484");
-    expect(returnTo.pathname).toBe("/chat");
+    expect(returnTo.pathname + returnTo.hash).toBe("/settings#cloud-overview");
     expect(returnTo.searchParams.get("elizaCloudLoginSession")).toBe(sessionId);
   },
 );
+
+it.each(["/", "/chat"])(
+  "offers agent selection after account switching instead of returning to %s without an agent",
+  async (path) => {
+    await begin(`/login?switchAccount=1&returnTo=${encodeURIComponent(path)}`);
+    await waitFor(() => expect(mocks.assign).toHaveBeenCalledOnce());
+    const login = new URL(mocks.assign.mock.calls[0][0]);
+    const handoff = new URL(requiredReturnTo(login), login.origin);
+    const returnTo = new URL(requiredReturnTo(handoff));
+    expect(mocks.clearBinding).toHaveBeenCalledOnce();
+    expect(returnTo.pathname + returnTo.hash).toBe("/settings#cloud-overview");
+    expect(returnTo.searchParams.get("elizaCloudLoginSession")).toBe(sessionId);
+  },
+);
+
+it("retains an explicit chat return when the existing agent binding is preserved", async () => {
+  await begin("/login?returnTo=%2Fchat");
+  await waitFor(() => expect(mocks.assign).toHaveBeenCalledOnce());
+  const handoff = new URL(mocks.assign.mock.calls[0][0]);
+  const returnTo = new URL(requiredReturnTo(handoff));
+  expect(mocks.clearBinding).not.toHaveBeenCalled();
+  expect(returnTo.pathname).toBe("/chat");
+});
 
 it("preserves a safe app destination and performs explicit account switching on both origins", async () => {
   await begin("/login?switchAccount=1&returnTo=%2Fsettings%23cloud-overview");
--- a/packages/ui/src/cloud/public-pages/pages/login/loopback-cloud-login-section.tsx
+++ b/packages/ui/src/cloud/public-pages/pages/login/loopback-cloud-login-section.tsx
@@ -60,12 +60,20 @@ export default function LoopbackCloudLoginSection() {
         );
       }
       const requested = sanitizeLoginReturnTo(searchParams.get("returnTo"));
-      const returnTo = new URL(requested || "/chat", window.location.origin);
+      const returnTo = new URL(
+        requested || "/settings#cloud-overview",
+        window.location.origin,
+      );
       // Public auth pages cannot consume the app's CLI completion marker.
-      if (/^\/(?:login|join|auth)(?:\/|$)/.test(returnTo.pathname)) {
-        returnTo.pathname = "/chat";
+      // Account switching also removed the selected agent: offer the signed-in
+      // account's agents before returning to a chat with no Cloud target.
+      if (
+        /^\/(?:login|join|auth)(?:\/|$)/.test(returnTo.pathname) ||
+        (switchAccount && ["/", "/chat"].includes(returnTo.pathname))
+      ) {
+        returnTo.pathname = "/settings";
         returnTo.search = "";
-        returnTo.hash = "";
+        returnTo.hash = "cloud-overview";
       }
       returnTo.searchParams.set("elizaCloudLogin", "complete");
       returnTo.searchParams.set("elizaCloudLoginSession", session.sessionId);
--- a/packages/ui/src/hooks/useAgentSessionRecovery.test.tsx
+++ b/packages/ui/src/hooks/useAgentSessionRecovery.test.tsx
@@ -435,15 +435,66 @@ describe("useAgentSessionRecovery", () => {
     expect(mockRunRecovery).not.toHaveBeenCalled();
   });
 
-  it("re-pairs when native SIWE supplies the Cloud token after the initial recovery attempt", async () => {
-    (globalThis as { Capacitor?: unknown }).Capacitor = {
-      isNativePlatform: () => true,
-    };
+  it.each(["native", "cloud-staging.eliza.app"])(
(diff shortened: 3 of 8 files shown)


attentionhead (approved):
**APPROVE** — verified by reading the diff at head `5c2fb2dab3b8238027e2ec4cc9be1f88619b645e` (base `develop`, 6 files, +228/-38) plus `useAgentSessionRecovery.ts` and `useCloudState.ts` fetched at that exact SHA. I did not execute the suites; the test counts in the description are not independently reproduced here. Production change is confined to two hooks — `useAgentSessionRecovery.ts` and `useCloudState.ts` — and the remaining four files are tests/mocks.

### Why the `steward-token-sync` rewrite is correct

The pre-PR handler called `setCloudTokenSnapshot(cloudToken)` unconditionally on every sync event. Since `cloudTokenSnapshot` is a dependency of the recovery effect (line 410) and that effect's cleanup calls `recoveryAbortController.abort()` (line 401), **every sync event aborted and restarted any in-flight repair** — including the event a cookie refresh publishes about its own token, moments before resolving back into this hook. That is the self-cancellation the PR title describes, and the three early-returns in the new `rearmAfterCloudReauth` are each aimed at one arm of it:

1. **No token → `setCloudTokenSnapshot(null)` and return.** Losing session authority *should* abort; this preserves that. The new test asserts `attempt.signal.aborted === true` and, sharper, that the stale attempt's `commitPairedInProcess("late-agent-bearer")` then **rejects with "target changed"** with `persistCloudPairApiToken` and `setAgentToken` asserted NOT called. That is the right control: it proves the abort does not merely stop the poll but actually revokes the late writer's authority to commit a credential. `recoveryTargetMatches` (line ~115) is the mechanism, and it compares `kind`, `id`, and normalized `apiBase` — not just identity.
2. **Same token as the one already attempted → return.** Guarded by the new `attemptedCloudTokenRef`, set at both attempt sites (line ~246 and ~348). This is the "let the transaction finish" case, and the paired test asserts `mockRunRecovery.mock.calls[0][0].signal.aborted === false` after the publication+resolution interleaving — i.e. the repair survives its own event.
3. **Not awaiting → return.** Unchanged intent.

Only a *different* non-empty token clears `awaitingCloudTokenRef`/`attemptedRef` and re-snapshots, so a rejected credential cannot be retried against itself. The rejection test is a genuine control on that: after `unauthorized`, it fires a second `steward-token-sync` with no new token and asserts `mockRunRecovery` is **still** called exactly twice. Without the `attemptedCloudTokenRef` comparison that assertion fails, so this is a real regression fence rather than a restatement of the happy path.

### The widened `awaitingCloudTokenRef` gate

`isManagedNative && …` becomes `consumeRedirectInProcess && isManagedCloudAgentServer(activeServer) && …`, and `shouldConsumePairRedirectInProcess()` is `isNativeRuntime() || isAppModeHost()` (line 103). So this strictly widens native → native-or-app-mode-host, which is the stated point: a hosted `cloud-staging.eliza.app` session can now re-arm without a reload. Two things make that safe to widen rather than alarming:

- It still requires `isManagedCloudAgentServer(activeServer)`, so a non-managed server never arms the wait.
- `fallbackStatus` (line 190) is untouched and still gates on `isManagedNative`, so the hosted arm re-arms **without** surfacing the native `cloud-reauth-required` notice. The parameterized test encodes exactly that asymmetry — same re-pair behaviour on both hosts, but `statuses.at(-1)` is `"cloud-reauth-required"` on native and `"idle"` on the hosted host. Converting the existing single-host test into `it.each(["native", "cloud-staging.eliza.app"])` keeps the old assertion intact as one arm rather than replacing it.

The deleted `lateCloudToken` block in the refresh path is the right removal, not lost coverage: its job was the one-shot re-check after a failed refresh, and the new `if (awaitingCloudTokenRef.current) rearmAfterCloudReauth();` inside `showFallback` covers the same window from the arming side, for every fallback rather than just that one call site.

### `useCloudState` recheck

`pollCloudCredits` added to the steward-refresh effect's deps is safe — it is `useCallback(runCloudPoll, [])` at line 753, referentially stable for the component's life, so the effect remains keyed on `elizaCloudConnected` alone and the `setInterval` is not re-armed on every render. The new `await pollCloudCredits()` is triple-guarded (`replaced && !disposed && readStoredStewardToken() === result.token`), so a refresh that lost a concurrent write does not poll. Its test asserts `fetchMock` called exactly **3** times (refresh, `/api/v1/user`, `/api/v1/credits/balance`) and that the follow-up requests carry `Bearer ${fresh}` — a count assertion plus a credential assertion, which together forbid both the no-recheck and the recheck-with-stale-token implementations.

### Non-blocking notes

1. `rearmAfterCloudReauth` is `useCallback(…, [])` but reads `attemptedRef`/`awaitingCloudTokenRef`/`attemptedCloudTokenRef` — correct today since all three are refs, but it is now also in the recovery effect's dep array (line 411), so any future non-ref capture would silently change that effect's re-run cadence. A one-line comment noting the empty dep list is load-bearing would protect it.
2. `attemptedCloudTokenRef` stores the raw bearer in a ref for the component's life. It is never logged and never leaves the hook, so this is not a finding — but it is one more in-memory copy of a credential, and clearing it alongside `attemptedRef.current = false` on the `!active && isAuthenticated` path (which the PR already does) is worth keeping that way.
3. The comparison is exact string equality on the trimmed token. If a future refresh ever returns a semantically-identical but re-encoded JWT, it would read as "different" and re-arm. Harmless today (a refresh mints a new token), just noting the assumption.
4. Draft PR — the head may move after this review; everything above is stated against `5c2fb2da`.

ss251 (changes requested):
I'll read the full review brief, the published reviewer skill, and the complete PR diff before writing the review body.5c2fb2dab3b8238027e2ec4cc9be1f88619b645e

`rearmAfterCloudReauth` (`useAgentSessionRecovery.ts`) on a missing Cloud token only calls `setCloudTokenSnapshot(null)`. It does not clear `attemptedRef` or `attemptedCloudTokenRef`. After an attempt, a later sync returns immediately unless `awaitingCloudTokenRef` is set and the JWT is different (`if (attemptedRef.current) { if (!awaitingCloudTokenRef.current || cloudToken === attemptedCloudTokenRef.current) return }`).

`awaitingCloudTokenRef` is set only in `showFallback` when `consumeRedirectInProcess && isManagedCloudAgentServer(...) && managedStatus === "cloud-reauth-required"`. The hosted row of the new `it.each` asserts status `idle` after the first no-token pass (`useAgentSessionRecovery.test.tsx`), so that flag is not armed. Input: in-flight hosted repair → `steward-token-sync` with `null` → later `steward-token-sync` with a new JWT. Output: recovery does not restart. The new cancel test never dispatches that follow-up token. That is the same reopen-the-app failure this PR claims to close.

The recovery tests mock `getCloudAuthToken`, `ensureCloudSession`, and `runRecovery`. They would still pass if Dedicated pairing or cookie refresh failed on a real Cloud/native boundary. Unit tests are not acceptance evidence for this path.

The `useCloudState.ts` change (call `pollCloudCredits` after `replaceStoredStewardTokenIfCurrent` succeeds) is scoped. The new steward-refresh test would fail if post-refresh `/api/v1/user` and `/api/v1/credits/balance` did not run with `Bearer` of the fresh JWT.

No lockfile, CI, binary, or permission expansion in this diff.

Verified by reading the supplied diff at this head (recovery hook, `useCloudState`, four test files). Live execution blocked: no disposable sandbox on this host, so this is a static review of the diff at the exact head.

Recommendation: fix

— [free-xai-lane-ss251]

PR #31102, opened 2026-09-12
test: verify Dedicated reconnect before requiring runtime choice

NubsCarson:
The deployed staging trajectory failed in its fresh-browser identity phase because it required the Cloud runtime picker after the primary chat and reload checks. A returning account can resolve its existing Dedicated binding before that picker is offered or while it disappears. The fresh-context check now accepts that path only when the account-derived Personal identity, Dedicated agent binding, runtime kind, and normalized API endpoint exactly match the already verified primary context. Initial onboarding still requires its visible Cloud choice; missing or mismatched identities and action timeouts remain failures. Hosting confirmation, real reply, server history, and forbidden-mutation checks are unchanged.

Adds payload-free entry diagnostics so an unresolved live failure distinguishes missing binding, retry, sign-in, and chat-overlay states. The subsequent protected deployed run confirmed that the fresh context had already resolved the exact Dedicated binding while the picker disappeared, then passed fresh-context history verification.

Validation: 33 real Chromium contract tests passed, including two new reconnect cases that failed before the change and rejection of missing/different account, agent, runtime, and endpoint. Full root `bun run verify`, focused Biome, and `git diff --check` passed. No product UI or runtime files changed; new visual captures are N/A for this test-only change. Existing latest-source Google sign-in, real Dedicated reply, and exact 22-message desktop/mobile reload evidence remain separate from clean-context certification.

Observed failing run: https://github.com/elizaOS/eliza/actions/runs/34676934969 (renderer job 103510639511, fresh-context-identity). Follow-up to #31100. The protected staging release and its fresh-context proof must pass before production promotion; this PR does not claim that gate has passed.

Final live validation: canonical staging run34678890343 passed at source1b2e3b9/treeaf499, with an actual Dedicated reply, reload/fresh-context history, exact identity/runtime/API reuse, and zero forbidden agent mutations. All hosted checks for this PR completed successfully.

### DIFF
--- a/packages/app/test/cloud-live-optional-action.ts
+++ b/packages/app/test/cloud-live-optional-action.ts
@@ -1,6 +1,10 @@
 /** Bounds optional Cloud onboarding actions without retaining DOM or account data. */
 
 import type { Locator } from "@playwright/test";
+import {
+  type CloudLiveRuntimeBinding,
+  compareCloudLiveRuntimeBindings,
+} from "./cloud-live-continuity-contract";
 
 export type CloudLiveOptionalActionPhase =
   | "pre-identity-runtime-choice"
@@ -193,6 +197,10 @@ interface PrepareCloudLivePersonalIdentityOptions {
   chatOverlay: Locator;
   chatOverlayTimeoutMs: number;
   chooseRuntimeAction: () => Promise<void>;
+  resolvedIdentity?: {
+    reference: CloudLiveRuntimeBinding;
+    readBinding: () => Promise<CloudLiveRuntimeBinding | null>;
+  };
 }
 
 interface WaitForCloudLivePersonalIdentityOptions<T> {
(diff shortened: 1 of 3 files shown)


attentionhead (approved):
**APPROVE** — verified by reading the diff at head `84619baf0a4ff7b8ebd5f6b0a91afa7673be2f9f` (3 files, +168/-2) and resolving the two load-bearing helpers at that same commit. I did not execute the Playwright suites; the browser-facing assertions below are read from the committed test source, not reproduced.

## What the change actually does

`prepareCloudLivePersonalIdentity` previously treated the chat overlay as an unconditional pre-choice gate: wait for it, then click the runtime picker. On a fresh browser that reconnects an *existing* Dedicated identity, the picker may never appear, and the old code failed the run. The PR adds an optional `resolvedIdentity` escape hatch guarded by an exact-match proof, in exactly two places:

1. **Before the overlay wait** — if the prior binding is already observable, return without attempting the choice.
2. **Inside a `catch` on `chooseRuntimeAction()`** — only for `CloudLiveRequiredActionUnavailableError`, and only if the same proof holds; every other error, and an unproven identity, still rethrows.

That two-arm shape is the right decomposition: arm 1 covers "identity resolved before the picker rendered", arm 2 covers "picker vanished mid-choice". Each maps to one interleaving, which is what keeps the widened acceptance from becoming a blanket "missing picker is fine".

## Why the widening does not weaken the gate

`hasVerifiedIdentity()` is a conjunction of four independent narrowings, all resolved at head:

- `resolvedIdentity?.reference.runtime !== "dedicated"` returns false up front, so a **shared**-runtime reference can never skip the picker.
- `compareCloudLiveRuntimeBindings` (`cloud-live-continuity-contract.ts:2097`) requires `reference.personalIdentity.length > 0` and `reference.runtimeBinding.length > 0` and `normalizedApiBase(reference.apiBase).length > 0` before any equality is reported. **An empty or partially-populated reference cannot vacuously match** — that is the property that makes this safe to call with caller-supplied data.
- `runtimeBindingReused` compares `runtime` *in addition to* `runtimeBinding`, so an agent id reused across a runtime switch does not qualify.
- The call site requires all three of `personalIdentityReused && runtimeBindingReused && apiBaseReused`.

`resolvedIdentity` is `undefined` at every call site except the continuity re-login (`cloud-live.spec.ts`, passing the already-verified `referenceBinding`), so first-run flows are unchanged. The positional call `(…, CLOUD_LIVE_CONTINUITY_IDENTITY_TIMEOUT_MS, true, undefined, undefined, referenceBinding)` lines up with the declared order `chooseRuntime, onRecovery, existingDedicatedAdoptionProof, existingReferenceBinding`.

## The tests are a real control, not a restatement

Two patterns worth naming, because both forbid the obvious wrong implementation:

**Paired positive assertion.** The `for (const resolvesDuringChoice of [false, true])` pair ends in `expect(choiceAttempted).toBe(resolvesDuringChoice)`. A single boolean distinguishes the two arms: the pre-overlay path must *not* attempt the choice, the mid-choice path must. An implementation that simply swallowed a missing picker would pass one arm and fail the other.

**Table-driven near-miss.** Five rejection arms — `missing`, `different account`, `different agent`, `shared runtime`, `different API` — each differing from the accepted binding in exactly one field, all asserted to reject with `CloudLiveRequiredActionUnavailableError`. This proves the acceptance predicate is narrow *by conjunction* rather than by one field happening to match. Note the `shared runtime` arm keeps the *reference* dedicated and varies the *candidate*, so it exercises `compareCloudLiveRuntimeBindings` rather than short-circuiting on the early `runtime !== "dedicated"` guard — the stronger of the two placements.

## Non-blocking notes

1. **`withinCloudLivePersonalIdentityDeadline` throws, it does not return null** (`cloud-live-optional-action.ts:229`, rejecting with `CloudLivePersonalIdentityDeadlineError`). `hasVerifiedIdentity` does not wrap it. In arm 2 that means a slow `readBinding` would replace the original `CloudLiveRequiredActionUnavailableError` with a deadline error, losing the failure that actually mattered for diagnosis. Since `readActiveBinding` is a fast page read this should not fire in practice, but a `try/catch` returning `false` there would make the fallback strictly non-destructive.
2. **The deadline is recomputed per call** as `Date.now() + chatOverlayTimeoutMs`. Arm 2 therefore grants a second full budget (60 s at the real call site) after the overlay wait and the action timeout have already been spent, so the worst-case wall time for this step roughly doubles. Deriving both from one deadline established on entry would bound it.
3. The new `console.warn` diagnostic in `chooseCloudRuntime` reads four locators after the failure; if the page is already tearing down, those reads can themselves throw inside `rethrowCloudLiveFailureAfterDiagnostic`. Worth confirming that helper is fail-safe about its diagnostic callback.

None of these change the accept/reject decision the PR is making, and the fail-closed default is preserved throughout: absent `resolvedIdentity`, behaviour is byte-for-byte the prior flow.

PR #31103, opened 2026-09-12
chore: promote Dedicated fresh-browser continuity verification

NubsCarson:
Promote #31102 from develop to staging. The only new changes are the deployed-browser test and its helper/contract tests: accept an already resolved Dedicated connection only when its account, agent, runtime, and endpoint exactly match the primary verified context; retain required initial choice, confirmation, real reply, server-history, and mutation gates. Payload-free entry diagnostics preserve evidence if the fresh-context failure remains.

Local validation: 33 Chromium contract tests, full root `bun run verify`, Biome, and diff checks passed at 84619baf0a4ff7b8ebd5f6b0a91afa7673be2f9f. Develop merge 8ee8fd4415070cd2799189e940672bf2826c9618 and the staging merge preview both have tree af499d9ac560f9b866f6446d1c75f3ec2c3148fa. Hosted source/billing/Windows checks were still running at merge; subscription authority passed. The exact-tree canonical staging release must still prove the clean-context flow. No production release or infrastructure cleanup is included.

### DIFF
--- a/packages/app/test/cloud-live-optional-action.ts
+++ b/packages/app/test/cloud-live-optional-action.ts
@@ -1,6 +1,10 @@
 /** Bounds optional Cloud onboarding actions without retaining DOM or account data. */
 
 import type { Locator } from "@playwright/test";
+import {
+  type CloudLiveRuntimeBinding,
+  compareCloudLiveRuntimeBindings,
+} from "./cloud-live-continuity-contract";
 
 export type CloudLiveOptionalActionPhase =
   | "pre-identity-runtime-choice"
@@ -193,6 +197,10 @@ interface PrepareCloudLivePersonalIdentityOptions {
   chatOverlay: Locator;
   chatOverlayTimeoutMs: number;
   chooseRuntimeAction: () => Promise<void>;
+  resolvedIdentity?: {
+    reference: CloudLiveRuntimeBinding;
+    readBinding: () => Promise<CloudLiveRuntimeBinding | null>;
+  };
 }
 
 interface WaitForCloudLivePersonalIdentityOptions<T> {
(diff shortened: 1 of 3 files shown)


attentionhead (approved):
**APPROVE** — verified at head `8ee8fd4415070cd2799189e940672bf2826c9618` as a **promotion**, not a line review: `develop` → `staging`, 3 files, +168/-2.

## Range mechanics

The range carries exactly one source commit plus its merge:

- `84619baf` — `test: verify Dedicated reconnect before requiring runtime choice`
- `8ee8fd44` — merge of #31102

`84619baf` is the exact head I reviewed on #31102 and approved there ([review 5185646242](https://github.com/elizaOS/eliza/pull/31102#pullrequestreview-5185646242)). **No commit in this range bypassed an open-PR review window**, which is the property I actually check on a promotion.

## File set corroborates the source PR

Per-file diffstat, which matches #31102 byte-for-byte:

| +/- | file |
|---|---|
| +52 / -2 | `packages/app/test/cloud-live-optional-action.ts` |
| +88 / -0 | `packages/app/test/ui-smoke/cloud-live-optional-action.spec.ts` |
| +28 / -0 | `packages/app/test/ui-smoke/cloud-live.spec.ts` |

I corroborate the source PR from the file set rather than the commit headline, and the two agree here. Scope is confined to `packages/app/test/` — **zero production files, zero `.sql`/DDL, no migration, no infrastructure or release content**, consistent with the body's "no production release or infrastructure cleanup is included."

## What is being promoted

The substance was reviewed on #31102: an exact-match escape hatch letting a fresh browser accept an already-resolved Dedicated identity instead of demanding the runtime picker. Its safety rests on `compareCloudLiveRuntimeBindings` guarding each leg with a non-empty check on the reference before any equality is reported, so a blank reference cannot vacuously unlock the path, and on `runtime` being compared alongside `runtimeBinding`. The fail-closed default is preserved: absent `resolvedIdentity`, behaviour is the prior flow. The non-blocking notes I raised there (the deadline helper throwing rather than returning `false` inside the `catch` arm, and the per-call deadline recomputation granting a second full budget) carry forward unchanged — neither blocks promotion.

## Attributed, not reproduced

The body states the develop merge `8ee8fd44` and the staging merge preview share tree `af499d9ac560f9b866f6446d1c75f3ec2c3148fa`, and reports 33 Chromium contract tests plus a full root `bun run verify` and Biome at `84619baf`. I did not independently reproduce those; the tree-equality claim and the local validation counts are the author's. The body is also candid that hosted source/billing/Windows checks were still running at merge and that the canonical staging release must still prove the clean-context flow — that is the right disclosure for a promotion of test-only coverage, and it does not change the decision here.

PR #31105, opened 2026-09-12
test: complete Dedicated lifecycle regression fixtures

NubsCarson:
The full staging run exposed three failing Dedicated lifecycle assertions because their fixtures no longer satisfied the real repository and service prerequisites. The heartbeat revision fixture now creates the canonical jobs table and its foreign-key dependencies. The sleep race fixture now supplies a complete container locator so snapshot I/O reaches the concurrent-write fence it is intended to test.

Production behavior and assertions are unchanged. The two PGlite suites pass all 11 tests / 49 assertions at a7bf49f9e13d9ed8ba5e6b8f7bbfabbfa4041613. Biome, git diff --check, and full root bun run verify pass at this exact source tree. Original failure evidence is in staging run 34678869411, Cloud unit shards 2 and 4. Other full-run failures remain separately recorded and are not represented as passing.

### DIFF
--- a/packages/cloud/shared/src/db/repositories/__tests__/heartbeat-generation-us-fence.test.ts
+++ b/packages/cloud/shared/src/db/repositories/__tests__/heartbeat-generation-us-fence.test.ts
@@ -30,7 +30,11 @@ process.env.MOCK_REDIS = "1";
 import { pushSchema } from "drizzle-kit/api";
 import { sql } from "drizzle-orm";
 import { agentSandboxes } from "../../schemas/agent-sandboxes";
+import { apiKeys } from "../../schemas/api-keys";
+import { generations } from "../../schemas/generations";
+import { jobs } from "../../schemas/jobs";
 import { organizations } from "../../schemas/organizations";
+import { usageRecords } from "../../schemas/usage-records";
 import { userCharacters } from "../../schemas/user-characters";
 import { users } from "../../schemas/users";
 
@@ -56,7 +60,16 @@ beforeAll(async () => {
   try {
     ({ closeDatabaseConnectionsForTests: closeDb, dbWrite } = await import("../../client"));
     ({ agentSandboxesRepository: repo } = await import("../agent-sandboxes"));
-    const schema = { organizations, users, userCharacters, agentSandboxes };
+    const schema = {
+      organizations,
+      users,
+      userCharacters,
+      agentSandboxes,
+      apiKeys,
+      usageRecords,
+      generations,
+      jobs,
+    };
     const { apply } = await pushSchema(schema as never, dbWrite as never);
     await apply();
     // Install the lifecycle_revision trigger BEFORE any raw writer runs —
--- a/packages/cloud/shared/src/lib/services/__tests__/typed-lifecycle-read.test.ts
+++ b/packages/cloud/shared/src/lib/services/__tests__/typed-lifecycle-read.test.ts
@@ -485,6 +485,8 @@ describe("typed lifecycle reads and exact sandbox generations", () => {
           await dbWrite
             .update(agentSandboxes)
             .set({
+              sandbox_id: "sleep-generation-test",
+              container_name: "sleep-generation-test",
               bridge_url: `http://127.0.0.1:${port}`,
               health_url: `http://127.0.0.1:${port}`,
               node_id: "test-node",


attentionhead (approved):
Verified at head `a7bf49f9e13d9ed8ba5e6b8f7bbfabbfa4041613`, base `develop`, state OPEN / `mergedAt: null`. Two files, `+16/-1`, both under `__tests__` — no production code changes, so the entire review is about whether these fixtures actually strengthen the controls they belong to.

## `heartbeat-generation-us-fence.test.ts` — the schema set is now FK-closed, and that is the real content of the change

`pushSchema` creates only the tables it is handed, so the failure mode for this kind of fixture is a foreign key pointing at a table that was never pushed. I resolved every `references(() => …)` target in the four newly added schemas at this head:

| added schema | FK targets |
|---|---|
| `apiKeys` | `organizations`, `users` |
| `usageRecords` | `apiKeys`, `organizations`, `users` |
| `generations` | `apiKeys`, `organizations`, `usageRecords`, `users` |
| `jobs` | `apiKeys`, `generations`, `organizations`, `users` |

Every target is present in the resulting set. The previous set (`organizations`, `users`, `userCharacters`, `agentSandboxes`) could not have supported any of these four, which is consistent with the PR title's "complete … fixtures": `generations` needs `usageRecords`, and `jobs` needs `generations`, so adding any one of them alone would have left a dangling reference.

The declaration order in the object literal — `organizations, users, userCharacters, agentSandboxes, apiKeys, usageRecords, generations, jobs` — is also a valid topological order over that dependency graph, with each table declared after everything it references. That is not strictly required for correctness if the pusher resolves the graph itself, but it makes the intended layering readable and means the list can be scanned for closure by eye.

No defect here. The change is exactly what it claims to be.

## `typed-lifecycle-read.test.ts` — the two new columns are not asserted, and that is worth stating precisely

The added lines set `sandbox_id` and `container_name` to `"sleep-generation-test"` inside the existing `.set()` that already populated `bridge_url`, `health_url`, `node_id`, `bridge_port` and `headscale_ip`, under the same `.where(eq(agentSandboxes.id, sandbox.id))`. The `where` clause is unchanged, so the update still targets exactly the one seeded row and the `returning()[0]` capture into `observed` is unaffected.

I grepped the whole file at this head: **`"sleep-generation-test"` appears only at those two lines (488–489) and is never read back.** The test's assertions are unchanged and are all about lifecycle sequencing, not identity:

- `result` matches `{ success: false, containerRemoved: false, error: "Agent lifecycle changed while sleep was prepared" }`
- `current?.status` is still `"running"`
- `current?.lifecycle_revision` is `observed.lifecycle_revision + 1`
- `current?.updated_at.getTime()` equals `observed.updated_at.getTime()`

So these two columns are not part of what the test checks. That makes them either (a) a precondition that lets `executeSleep` reach the real prepare/commit path instead of short-circuiting on an incomplete row — in which case `containerRemoved: false` becomes a meaningful negative rather than a vacuous one — or (b) inert fixture padding. Those two readings have very different value, and I could not distinguish them at this head, for the reason below.

## One open question, non-blocking

`ElizaSandboxService.executeSleep` at this head (`eliza-sandbox.ts:3159`) is a pure delegation — `return this.#power.executeSleep(...args)` — and the string `"Agent lifecycle changed while sleep was prepared"` does **not** appear anywhere in `eliza-sandbox.ts` at `a7bf49f9` (I fetched the file at the reviewed ref twice, via both the JSON and raw media types, 146,285 bytes both times, to rule out a truncated fetch). So the code that produces that error, and any gating on `sandbox_id` / `container_name` being populated, lives in the `SandboxPower` implementation rather than in the service shell.

**Could you confirm which module owns that path, and whether the two new columns are load-bearing for reaching it?** If they are, a one-line comment on the `.set()` saying so would keep a future reader from "cleaning up" two unasserted fields and silently turning `containerRemoved: false` into an assertion that passes for the wrong reason. If they are not load-bearing, they are harmless.

This is a question, not a blocking objection — nothing in the diff can regress a passing suite, and the FK closure above is independently correct.

## Note on scope of verification

I verified the FK targets, the schema set, the `where` clause, and the absence of any read-back of the new literal by fetching the two test files and the four schema files at the reviewed SHA. I did **not** run the suite, and I did not resolve the `SandboxPower` implementation, so I am not asserting anything about the runtime behaviour of `executeSleep` beyond the delegation shown above.

Approving: test-only, FK-closed, assertions unchanged and still meaningful.

ss251 (approved):
Reviewed head a7bf49f9e13d9ed8ba5e6b8f7bbfabbfa4041613.

The risky hunk is the pushSchema growth in heartbeat-generation-us-fence.test.ts:68-71, which adds apiKeys, usageRecords, generations and jobs to the pushed schema. From the hunk context I confirmed `apply()` (line 74) still precedes the lifecycle_revision trigger install (line 75), so all eight tables exist before the trigger fires and any raw writer runs; the four new imports also match the `../../schemas/*` directory of the pre-existing ones. In typed-lifecycle-read.test.ts:488-489, sandbox_id and container_name share one literal "sleep-generation-test" — non-blocking, but distinct values would catch a swapped-column typed read that the shared literal cannot.

Recommendation: merge

— [free-9router-lane-ss251]

PR #31119, opened 2026-09-12
fix(core): render shared references in redactLogArgs instead of masking them (#31113)

### LINKED ISSUE #31113: core: redactLogArgs renders a shared object reference as [REDACTED]
**Describe the bug**

`redactLogArg` in `packages/core/src/security/redact.ts` (lines 600-660, develop `cbf357b0bd8a`) uses a `WeakSet` to stop on cycles but never removes an object from the set after walking it:

```ts
if (depth >= MAX_LOG_REDACT_DEPTH || seen.has(value)) {
  return "[REDACTED]";
}
seen.add(value);
```

So the set holds every object visited so far, not just the current ancestor path. Any object (or Error) referenced twice in one log call is rendered as `"[REDACTED]"` on its second occurrence. That string is indistinguishable from a genuine credential mask, so a reader concludes a secret was present where there was only a shared reference.

`redactLogArgs` is consumed on every line of the cloud logger (`packages/cloud/shared/src/lib/utils/logger.ts`, lines 167, 176, 184 and 193: `console.log(...redactLogArgs(args))`), and shared references are common in the objects that get logged (a request object referenced from two fields, the same model descriptor in `requested` and `resolved`, an error attached under two keys).

**To Reproduce**

```ts
import { redactLogArgs } from "packages/core/src/security/redact.ts";
const shared = { provider: "openai", model: "gpt-4o" };
console.log(JSON.stringify(redactLogArgs(["model call", { requested: shared, resolved: shared, list: [shared, shared] }])));
const err = new Error("boom");
console.log(JSON.stringify(redactLogArgs([{ a: err, b: err }]), (_k, v) => v instanceof Error ? `Error(${v.message})` : v));
```

Output on develop:

```
["model call",{"requested":{"provider":"openai","model":"gpt-4o"},"resolved":"[REDACTED]","list":["[REDACTED]","[REDACTED]"]}]
[{"a":"Error(boom)","b":"[REDACTED]"}]
```

**Expected behavior**

Only true cycles (an object that is its own ancestor) are cut, and they are marked with a distinct token such as `"[circular]"`, never with the credential mask. A shared reference renders in full at every occurrence, as `JSON.stringify` would.

- [ ] MP4 recording or JPG screenshots: N/A - log formatting, no UI.
- [x] Output of the script above is inline.
- [ ] Domain artifacts: N/A.

**Additional context**

Same defect class as the logger, settings-debug and JSON-output walkers fixed in #30887, #31002 and #31005; the fix is the same: `seen.delete(value)` in a `finally` after the array, Error and object branches, so `seen` tracks the ancestor path only. The existing `MAX_LOG_REDACT_DEPTH` bound already covers runaway depth.

hermesagent270-commits:
## Summary

`redactLogArg` in `packages/core/src/security/redact.ts` kept every visited object in its `WeakSet` for the whole walk, so any object or Error referenced twice in one `logger.*(...)` call rendered as `"[REDACTED]"` on its second occurrence. That string is the credential mask, so a reader concluded a secret was present where there was only a shared reference. The cloud logger runs `redactLogArgs` on every line.

The set now holds only the current ancestor path (each entry is removed in a `finally` after its subtree is walked), so:

- a shared reference is rendered in full at every site;
- a genuine cycle is still cut, but with a distinct `"[circular]"` marker instead of the credential mask;
- the depth bound is unchanged.

Closes #31113

## Evidence

Before (develop `cbf357b0bd8a`):

```
["model call",{"requested":{"provider":"openai","model":"gpt-4o"},"resolved":"[REDACTED]","list":["[REDACTED]","[REDACTED]"]}]
[{"a":"Error(boom)","b":"[REDACTED]"}]
```

After, the same inputs render `resolved`, both `list` entries and `b` in full; the new tests pin that plus sibling-vs-cycle discrimination inside one walk.

- `bunx vitest run src/security/redact.test.ts` in `packages/core`: 66/66 (62 existing + 4 new).
- `bun run --cwd packages/core typecheck`: clean.
- Biome check on both files: clean.

Same defect class as #30887, #31002 and #31005 (logger, settings-debug and JSON-output walkers).

### DIFF
--- a/packages/core/src/security/redact.test.ts
+++ b/packages/core/src/security/redact.test.ts
@@ -669,6 +669,46 @@ describe("redactLogArgs (log-sink redaction, not opt-in)", () => {
 		cyclic.self = cyclic;
 		const [out] = redactLogArgs([cyclic]) as [Record<string, unknown>];
 		expect(out.name).toBe("x");
+		// A genuine cycle is cut with a marker that cannot be read as a
+		// credential mask.
+		expect(out.self).toBe("[circular]");
+	});
+
+	it("renders an object referenced twice in one call at both sites", () => {
+		const shared = { provider: "openai", model: "gpt-4o" };
+		const [, ctx] = redactLogArgs([
+			"model call",
+			{ requested: shared, resolved: shared, list: [shared, shared] },
+		]) as [string, Record<string, unknown>];
+		expect(ctx.requested).toEqual(shared);
+		expect(ctx.resolved).toEqual(shared);
+		expect(ctx.list).toEqual([shared, shared]);
+		expect(JSON.stringify(ctx)).not.toContain("[REDACTED]");
+	});
+
+	it("renders a shared Error at both sites and still scrubs its message", () => {
+		const err = new Error("token sk-abcdefghijklmnop1234 rejected");
+		const [ctx] = redactLogArgs([{ first: err, second: err }]) as [
+			Record<string, Error>,
+		];
+		expect(ctx.first).toBeInstanceOf(Error);
+		expect(ctx.second).toBeInstanceOf(Error);
+		expect(ctx.second.message).toBe(ctx.first.message);
+		expect(ctx.second.message).not.toContain("sk-abcdefghijklmnop1234");
+	});
+
+	it("distinguishes a sibling reference from a cycle inside one walk", () => {
+		// `leaf` appears under two siblings and, separately, inside a true cycle:
+		// the sibling copies are rendered in full, the self-reference is cut.
+		const leaf = { id: "leaf" };
+		const loop: Record<string, unknown> = { leaf };
+		loop.back = loop;
+		const [ctx] = redactLogArgs([{ a: leaf, b: leaf, loop }]) as [
+			Record<string, unknown>,
+		];
+		expect(ctx.a).toEqual(leaf);
+		expect(ctx.b).toEqual(leaf);
+		expect(ctx.loop).toEqual({ leaf, back: "[circular]" });
 	});
 
 	it("drops a hostile own toJSON so serialization cannot reconstitute secrets", () => {
--- a/packages/core/src/security/redact.ts
+++ b/packages/core/src/security/redact.ts
@@ -569,6 +569,8 @@ export function redactObjectSecrets<T>(
 
 const REDACTED_MASK = "[REDACTED]";
 const MAX_LOG_REDACT_DEPTH = 8;
+/** Marker for a value that is its own ancestor; deliberately not the credential mask. */
+const CIRCULAR_MARKER = "[circular]";
 
 /**
  * Redact one log argument for output at the sink. A string is scrubbed with the
@@ -593,13 +595,15 @@ const MAX_LOG_REDACT_DEPTH = 8;
  * innocent-looking key, and JSON.stringify would emit them as
  * {"type":"Buffer","data":[…]} — either way secret bytes survive in every sink.
  *
- * Depth is bounded and cycles are broken (returning the mask) so a pathological
- * log payload cannot hang or blow the stack — a redactor must never be the thing
- * that takes the process down.
+ * Depth is bounded and true cycles are broken with a distinct circular marker so
+ * a pathological log payload cannot hang or blow the stack — a redactor must
+ * never be the thing that takes the process down. Only ancestors of the current
+ * value count as a cycle: an object referenced from two places in one call is
+ * rendered in full both times, never as the credential mask.
  */
 function redactLogArg(
 	value: unknown,
-	seen: WeakSet<object>,
+	ancestors: WeakSet<object>,
 	depth: number,
 ): unknown {
 	if (typeof value === "string") {
@@ -611,20 +615,15 @@ function redactLogArg(
 	if (value === null || typeof value !== "object") {
 		return value;
 	}
-	if (depth >= MAX_LOG_REDACT_DEPTH || seen.has(value)) {
+	if (depth >= MAX_LOG_REDACT_DEPTH) {
 		return REDACTED_MASK;
 	}
-	seen.add(value);
-	if (Array.isArray(value)) {
-		// Do not call value.map: an Array subclass, custom Symbol.species, or own
-		// map property can return caller-owned data carrying a serializer hook.
-		// Index into the input but construct the output with the intrinsic Array
-		// constructor so no caller-controlled method or result prototype survives.
-		const result: unknown[] = [];
-		for (let index = 0; index < value.length; index += 1) {
-			result.push(redactLogArg(value[index], seen, depth + 1));
-		}
-		return result;
+	// `ancestors` holds only the objects on the current walk path (entries are
+	// removed on the way back out), so this catches a genuine cycle and nothing
+	// else. A shared reference must not be masked: "[REDACTED]" would tell the
+	// reader a credential was present where there was only aliasing.
+	if (ancestors.has(value)) {
+		return CIRCULAR_MARKER;
 	}
 	if (value instanceof Error) {
 		// Preserve the Error shape (name/stack) callers rely on, but scrub the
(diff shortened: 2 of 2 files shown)


attentionhead (approved):
**APPROVE** — verified at head `15ccf1c26f62dc013db25e6b92fd4c662e33e834` (OPEN, base `develop`, 2 files, +88/-32, 1 commit).

## What the change actually does

`redactLogArg` previously carried one `WeakSet` for the whole walk and returned `REDACTED_MASK` on any repeat visit. That conflates two different facts — "this value is its own ancestor" and "this value was already rendered somewhere else in the same argument" — and reports both with the credential mask. The second case is the bug: a reader of the log cannot distinguish aliasing from a masked secret, and `redactLogArgs` runs on every cloud log line, so the false signal is repo-wide.

The fix narrows the set to the **current ancestor path**: `ancestors.add(value)` immediately before descending, `ancestors.delete(value)` in a `finally` on the way back out. Membership therefore means "on the path from the root to here", which is exactly the cycle predicate. A genuine cycle now returns the distinct `CIRCULAR_MARKER = "[circular]"` rather than the credential mask.

The re-ordering in the diff is load-bearing and correct: the `Error` branch (redact.ts:628) and the `ArrayBuffer`/view branch (:640) both return **before** `ancestors.add`, and neither recurses, so they cannot leave a stale entry — and a shared `Error` is now rendered at both sites, which the new test pins directly. The array branch moved inside the `try` so it gets the same `finally` cleanup as the object branch; its intrinsic-`Array` construction and the null-prototype object target are carried over unchanged, as are the `isSensitiveKeyName` masking and the function-property drop. The depth bound (`MAX_LOG_REDACT_DEPTH = 8`) is untouched and still checked first, so the walk remains terminating independently of the set.

## Control quality

The added tests are a **sibling-vs-cycle discrimination** control, which is the right shape for this defect: `distinguishes a sibling reference from a cycle inside one walk` puts `leaf` under two siblings *and* a self-referential `loop` in a single payload and asserts both outcomes at once — `ctx.a`/`ctx.b` render in full while `loop.back` is `"[circular]"`. A test that only checked the shared case would pass on a patch that dropped cycle detection entirely; this one cannot. It is paired with `expect(JSON.stringify(ctx)).not.toContain("[REDACTED]")`, a negative assertion over the whole rendered argument rather than one field, and with the shared-`Error` case that asserts the message is equal at both sites **and** still scrubbed (`not.toContain("sk-abcdefghijklmnop1234")`) — so the fix cannot have been obtained by skipping redaction on the second visit. The pre-existing cycle test was extended with `expect(out.self).toBe("[circular]")` rather than left asserting only `out.name`, which is what makes the marker change itself covered.

## Corroboration

This brings `redact.ts` in line with the ancestor-path idiom already used elsewhere in `packages/core`: `action-parameter-value.ts:21`, `entity-matches.ts:29` and `security/secret-swap.ts:39` all name their set `visiting` rather than `seen`, i.e. the path semantics this PR adopts. The file was the outlier.

## Non-blocking notes

1. **Depth exhaustion still returns the credential mask.** `if (depth >= MAX_LOG_REDACT_DEPTH) return REDACTED_MASK;` (redact.ts, unchanged) reintroduces exactly the ambiguity this PR removes for aliasing: a deep-but-innocent payload renders `"[REDACTED]"` at level 8 and reads as a masked secret. The PR's own argument applies verbatim — a distinct marker (`"[depth limit]"`, alongside the new `"[circular]"`) would close the last path by which a non-secret renders as one. Out of scope for this fix, but it is the same defect, one branch over.

2. **Path semantics re-walk shared subtrees.** The old global `seen` made a DAG cost O(nodes); an ancestor-only set costs one walk **per reference site**, so a structure that references the same subtree from *k* places walks it *k* times, multiplying down the levels. The depth bound of 8 is what keeps this finite, and it is the only thing that does — the comment "a redactor must never be the thing that takes the process down" is now resting entirely on that constant rather than on memoisation. This is the correct trade (masking a shared reference as a credential is worse than re-walking it), and log arguments are normally small, but it is worth stating explicitly since `redactLogArgs` is on every cloud log line. If a payload shape ever makes this hot, the fix is a separate *rendered-output* cache keyed by object identity, which restores O(nodes) without reintroducing the mask.

3. **Evidence is attributed, not reproduced here.** The before/after renderings against develop `cbf357b0bd8a`, `bunx vitest run src/security/redact.test.ts` at 66/66, the `packages/core` typecheck, the Biome runs, and the pending root `bun run verify` are the author's; I verified the diff, the surrounding file at head, and the sibling walkers listed above, and did not execute the suite. The `Closes #31113` linkage and the stated relationship to #30887 / #31002 / #31005 are likewise taken from the description.

4. Minor, pre-existing: the `Error` branch reconstructs with `new Error(...)`, so an `Error` subclass is flattened to a plain `Error` (only `name` and `stack` survive). The new test asserts `toBeInstanceOf(Error)`, which holds either way. Not introduced here.

No defect found in the change. The cycle predicate is now the correct one, the cleanup is exception-safe, the pre-existing hardening is preserved verbatim, and the tests discriminate the two cases the old code conflated.

ss251 (approved):
15ccf1c26f62dc013db25e6b92fd4c662e33e834

The old walk stuffed every visited object into a persistent `WeakSet` and returned `[REDACTED]` on the next sighting. Input: `{ requested: shared, resolved: shared }` with `shared = { provider: "openai", model: "gpt-4o" }`. Wrong output: `resolved` became the credential mask even though no sensitive key matched. Operators cannot tell aliasing from a secret; that is a real log-sink bug, not a style nit.

This head fixes the walk. `redactLogArg` keeps a path-only ancestor set (`redact.ts:643` add, `redact.ts:677-678` `finally` delete) and returns `CIRCULAR_MARKER` (`"[circular]"`, `redact.ts:572-573`) only on `ancestors.has(value)` (`redact.ts:625-626`). Shared aliases are cloned in full; a true back-edge is cut with a marker that cannot be read as a credential mask. Depth still caps at 8 with `[REDACTED]` (`redact.ts:618-620`) — pre-existing, not this defect.

The new tests would fail if the bug returned: both sites equal `shared` and `JSON.stringify` contains no `[REDACTED]` (`redact.test.ts:677-687`); a shared `Error` is an `Error` at both sites with `sk-abcdefghijklmnop1234` scrubbed (`redact.test.ts:689-698`); sibling `leaf` fully rendered while `loop.back === "[circular]"` (`redact.test.ts:700-713`). The existing cycle case now asserts the distinct marker (`redact.test.ts:672-675`). Array cloning still uses the intrinsic constructor; sensitive keys and function-valued serializer hooks are unchanged. No lockfile, CI, or permission expansion.

Verified by reading the diff at this head (`packages/core/src/security/redact.ts`, `redact.test.ts`).

Recommendation: merge

— [free-xai-lane-ss251]

hermesagent270-commits:

Svector-anu (approved):
Verified the actual defect and the fix's precision against #31113, not just the diff shape.

**Bug**: the old `seen: WeakSet` in `redactLogArg` was added to on first visit and never cleared, so it tracked every object visited across the *whole* call, not the current ancestor path. Any object referenced twice anywhere in one log payload (a shared config object, the same Error under two keys) hit `seen.has(value)` on its second occurrence and rendered as `"[REDACTED]"` — visually identical to an actual credential mask, so a reader can't tell "a secret was here" from "this object also appears somewhere else."

**Fix**: renamed to `ancestors`, pushed with `ancestors.add(value)` immediately before recursing and popped with `ancestors.delete(value)` in a `finally` around both the array and object branches — standard DFS ancestor-stack semantics, so `ancestors.has(value)` is only true for a genuine self-reference on the current path, never a sibling. Also splits the output marker: a true cycle now renders `"[circular]"`, distinct from `REDACTED_MASK`, directly fixing the "indistinguishable from a real redaction" complaint in #31113.

**Test coverage is exactly targeted at the distinction that matters**, not just re-asserting the diff:
- `"renders an object referenced twice in one call at both sites"` — the shared-object repro from #31113, both sites render in full.
- `"renders a shared Error at both sites and still scrubs its message"` — the shared-Error repro; confirms the `instanceof Error` branch (which sits before the ancestors check and never pushes onto `ancestors`) independently re-scrubs each occurrence rather than masking the second as `[REDACTED]`.
- `"distinguishes a sibling reference from a cycle inside one walk"` — the sharpest case: `leaf` shared as two siblings *and*, separately, a true self-reference (`loop.back = loop`) in the same call. Siblings render in full, the actual cycle is cut. This is the one test that would fail under a naive "just don't push non-object values" fix and doesn't under a real ancestor-stack.

`MAX_LOG_REDACT_DEPTH` bounding is untouched (still keyed off the independent `depth` counter), so the stack-safety guarantee the original comment called out is preserved.

No defects found. CI green, already 2 approvals. Approving.

lalalune:

Consolidating this independent redactor patch into #31004, which already owns #31113 and the related serializer implementations. The sibling-reference defect is real, but switching every shared object to per-path expansion while relying only on depth does not establish the header claim that pathological logging cannot overwhelm the process. Keep the useful shared-Error/secret-scrubbing and true-cycle regressions in the coordinated log-sink implementation, with measured expansion behavior and a genuine diagnostic-preview contract. Do not copy lossy preview limits into full model/trajectory serialization. This closes the separate patch, not the underlying acceptance criteria.

PR #28225, opened 2026-08-25
fix(plugin-notes): support replacement and newContent parameter aliases in NOTES action update

deepanshu-yd:
# Relates to

Relates to #28224

Definition of Done: full standard in [`CONTRIBUTING.md`](../CONTRIBUTING.md).

- [x] This PR targets `develop` and is rebased onto the latest `origin/develop` with zero conflicts (`git fetch origin && git rebase origin/develop`).
- [x] `bun install` and `bun run verify` were run after sync, or any failure is recorded below with the exact unrelated blocker.
- [x] A reviewer can confirm the change works without reading the code, from the evidence attached below.

# Sync with develop

- [x] Rebased/merged onto the latest `origin/develop`; zero conflicts.
- [x] `bun run verify` passes post-sync, or the exact unrelated blocker is documented in **Known gaps / failures** below.

# Risks

Low. Adds parameter alias support (`replacement`, `newContent`, `newNote`) for the new note content in `NOTES` action update in `plugins/plugin-notes/src/action.ts`.

# Background

## What does this PR do?

In `plugins/plugin-notes/src/action.ts`, `op === "update"` now accepts `params.replacement`, `params.newContent`, and `params.newNote` alongside `params.body` and `params.newText` so planner invocations using these common field names successfully update notes.

## What kind of change is this?

Bug fixes (non-breaking change which fixes an issue)

# Documentation changes needed?

My changes do not require a change to the project documentation.

# Testing

## Where should a reviewer start?

- `plugins/plugin-notes/src/action.ts`: `replacement` resolution logic on `op === "update"`.
- `plugins/plugin-notes/src/action.test.ts`: test cases testing parameter aliases (`replacement`, `newContent`, `newText`).

## Detailed testing steps

Run:
```bash
bun run --cwd plugins/plugin-notes test
bun run --cwd plugins/plugin-notes typecheck
bun run --cwd plugins/plugin-notes lint
```

# Evidence Gate

- [ ] Before full-page screenshots are attached for every affected UI surface (desktop and mobile), or marked `N/A - <reason>`.
  N/A - Notes action parameter resolution fix.
- [ ] After full-page screenshots are attached for every affected UI surface (desktop and mobile), or marked `N/A - <reason>`.
  N/A - Notes action parameter resolution fix.
- [ ] A video walkthrough of the complete user flow is attached, or marked `N/A - <reason>`.
  N/A - Notes action parameter resolution fix.
[...]

Issue #16935, opened 2026-07-23
fix(core): guarantee a usable reply or explicit failure after tool execution


lalalune:
## Context

Live persona-pack verification (#14350, #14354, #14356 — all closed) found three real, reproducible product defects. Each blocks catalog rows from reaching `verified`, and none has a tracking issue (checked board 15 post-sweep + full open-issue index, 2026-07-22). The scenarios that expose them are already on origin/develop and stay honestly `authored` per the catalogs.

## Defect 1 — planner emits tool calls but no synthesized user reply (blocks 3 A2 rows)

On read/list-then-summarize turns the LifeOps action-planner issues tool calls but no user-facing reply reaches the owner; one run's final 'prose' was a malformed `call:OWNER_TODOS_REVIEW{action:review}` literal after `SCHEDULED_TASKS_HISTORY` errored `MISSING_TASK_ID` (#14350 close-out on PR #14556, live Cerebras gpt-oss-120b, judge 0.00 x3). Repro scenarios: `plugins/plugin-personal-assistant/test/scenarios/adhd-followthrough-{missed-step-replan-without-guilt,repeated-miss-shrink-or-pause,end-of-day-wins-first-recap}.scenario.ts`.

## Defect 2 — cross-persona privacy leak (F1)

`f1-cross-persona-privacy-minimization`: FAILS 3/3 live (2026-07-06, #14356), judge 0.00 — the assistant volunteers a seeded private therapy appointment. Sibling row `f1-neurotypical-no-scaffolding-unless-signaled` also fails 3/3. Notes in `_catalogs/neurotypical-control-adversarial.catalog.json`. This is a privacy defect, arguably the most severe finding in the corpus.

## Defect 3 — SCHEDULED_TASKS schema guidance leaks into user copy

`elderly-week1-bill-due-reminder-plain-words` is flaky ~50%: the reminder is created but the confirmation intermittently exposes 'ISO-8601', 'trigger type is missing', raw task ids, or fails to schedule (catalog `_catalogs/elderly-week1-loop.catalog.json`, live Cerebras run 1 PASS / run 2 FAIL 0.00, 2026-07-06).

## Acceptance criteria

- Root-cause fixes in the runtime (tool-then-reply synthesis path; provider/context privacy minimization; SCHEDULED_TASKS user-copy boundary) — no scenario weakening, no prompt-text pattern-matching in the runner (structural-fields-only contract).
[...]

PR #31120, opened 2026-09-12
fix(plugin-notes): declare the planner aliases the NOTES handler reads (#31114)

### LINKED ISSUE #31114: plugin-notes: NOTES reads text, note, title, query and newText aliases the tool schema never declares
**Describe the bug**

The `NOTES` action in `plugins/plugin-notes/src/action.ts` (develop `cbf357b0bd8a`) reads planner aliases the tool schema never declares. The handler falls back through `text`, `note`, `title`, `query` and `newText` (lines 196-197, 220-225 and 271) as alternatives to `content` and `body`, but `parameters` (line 291) declares only `action`, `content` and `body` with no `aliases`. `validateToolArgs` (`packages/core/src/actions/validate-tool-args.ts`, line 145) rejects any key absent from `properties` when the action does not set `allowAdditionalParameters`, and `normalizeParamAliases` (`packages/core/src/runtime/execute-planned-tool-call.ts`, line 1141) returns the arguments untouched when no parameter declares aliases. The fallbacks are therefore unreachable, and a planner that uses them burns a round on an argument error and saves nothing.

**To Reproduce**

Real `validateToolArgs` plus real `normalizeParamAliases` against the real action definition:

```
NOTES {"action":"create","content":"buy milk"} => {"valid":true,"errors":[]}
NOTES {"action":"create","text":"buy milk"} => {"valid":false,"errors":["Unexpected argument 'text'"]}
NOTES {"action":"list","query":"milk"} => {"valid":false,"errors":["Unexpected argument 'query'"]}
NOTES {"action":"update","content":"milk","newText":"oat milk"} => {"valid":false,"errors":["Unexpected argument 'newText'"]}
```

**Expected behavior**

Either the aliases are declared where core expects them (for example `aliases: ["text", "note", "title", "query"]` on `content` and `aliases: ["newText", "replacement"]` on `body`) so the planner's natural spellings reach the handler, or the dead fallbacks are removed so the handler and schema agree.

- [ ] MP4 recording or JPG screenshots: N/A - tool-argument validation, no UI.
- [x] Validator output is inline above.
- [ ] Real-LLM trajectory: N/A - the rejection happens in the validator before any model-driven execution; the action's own similes (`TAKE_NOTE`, `WRITE_NOTE`) and its "The note's text" wording make the `text` spelling a natural planner choice.
- [ ] Domain artifacts: N/A.

**Additional context**

Precedent for this exact fix and symptom: `plugins/plugin-personal-assistant/src/actions/owner-surfaces.ts` lines 195-210 ("burned a planner round on arg errors", #16935).

hermesagent270-commits:
## Summary

The `NOTES` handler reads `text`, `note`, `title` and `query` as fallbacks for `content`, and `newText` as a fallback for `body`, but the action declared only `action`, `content` and `body`. Core's `validateToolArgs` rejects any undeclared key and `normalizeParamAliases` only renames keys a parameter claims, so every call using those spellings failed with `Unexpected argument` before the handler ran and nothing was saved.

This declares them where core expects them:

- `content.aliases = ["text", "note", "title", "query"]`
- `body.aliases = ["newText", "replacement"]`

The handler's fallbacks are unchanged. An alias sent together with an explicit canonical value is still rejected, per core's alias rule (never clobber a provided canonical value); the new test pins that too.

Closes #31114

## Evidence

Before (develop `cbf357b0bd8a`, real validator against the real action):

```
NOTES {"action":"create","text":"buy milk"} => {"valid":false,"errors":["Unexpected argument 'text'"]}
NOTES {"action":"list","query":"milk"} => {"valid":false,"errors":["Unexpected argument 'query'"]}
NOTES {"action":"update","content":"milk","newText":"oat milk"} => {"valid":false,"errors":["Unexpected argument 'newText'"]}
```

After: the new test drives the real `executePlannedToolCall` through the promoted `NOTES_CREATE` / `NOTES_LIST` / `NOTES_UPDATE` children and the `NOTES` umbrella with `text`, `query`, `title` + `newText`, and `note`; each call succeeds and the created/updated note carries the expected title and body.

- `bunx vitest run` in `plugins/plugin-notes`: 144/144 (10 files).
- `bun run typecheck` and `bun run lint:check` in `plugins/plugin-notes`: clean.

Precedent: `plugins/plugin-personal-assistant/src/actions/owner-surfaces.ts` declares the same kind of aliases for the same observed symptom (#16935).

# Evidence Gate

- [x] N/A - no user-facing flow; the action calls are asserted by the plugin-notes suite below.

  <details><summary>validator output on develop and vitest output at this head (plugins/plugin-notes)</summary>

  ```text
  BEFORE (develop cbf357b0bd8a, real validator against the real action):
   NOTES {"action":"create","text":"buy milk"} => {"valid":false,"errors":["Unexpected argument 'text'"]}
   NOTES {"action":"list","query":"milk"} => {"valid":false,"errors":["Unexpected argument 'query'"]}
   NOTES {"action":"update","content":"milk","newText":"oat milk"} => {"valid":false,"errors":["Unexpected argument 'newText'"]}
  AFTER (this head): the new test drives the real executePlannedToolCall through NOTES_CREATE / NOTES_LIST / NOTES_UPDATE and the NOTES umbrella with text, query, title + newText and note; every call succeeds
  $ bunx vitest run   (plugins/plugin-notes)
   Test Files  10 passed (10)
        Tests  144 passed (144)
  typecheck + lint:check in plugins/plugin-notes: exit 0
  ```
  </details>

- [x] N/A - in-memory note records created through the aliased calls are asserted on title and body inside the new test; no persistent store is touched by the change.

### DIFF
--- a/plugins/plugin-notes/src/action.test.ts
+++ b/plugins/plugin-notes/src/action.test.ts
@@ -109,6 +109,64 @@ function execute(
 }
 
 describe("promoted Notes execution", () => {
+  it("accepts the planner alias spellings the handler documents (#31114)", async () => {
+    // Before the aliases were declared, every one of these calls was rejected
+    // by the core validator with "Unexpected argument" before the handler ran.
+    const runtime = await executorHarness();
+    const created = await execute(runtime, {
+      name: "NOTES_CREATE",
+      params: { text: "Alias check\noriginal body" },
+    });
+    expect(created).toMatchObject({
+      success: true,
+      data: {
+        op: "create",
+        note: { title: "Alias check", body: "original body" },
+      },
+    });
+
+    const listed = await execute(runtime, {
+      name: "NOTES_LIST",
+      params: { query: "alias" },
+    });
+    expect(listed).toMatchObject({
+      success: true,
+      data: { count: 1, filterApplied: true, topic: "alias" },
+    });
+
+    const updated = await execute(runtime, {
+      name: "NOTES_UPDATE",
+      params: { title: "Alias check", newText: "Alias check\nupdated body" },
+    });
+    expect(updated).toMatchObject({
+      success: true,
+      data: {
+        op: "update",
+        note: { title: "Alias check", body: "updated body" },
+      },
+    });
+
+    const umbrella = await execute(runtime, {
+      name: "NOTES",
+      params: { action: "create", note: "Second note" },
+    });
+    expect(umbrella).toMatchObject({
+      success: true,
+      data: { op: "create", note: { title: "Second note" } },
+    });
+
+    // Core never lets an alias clobber an explicit canonical value: when both
+    // arrive, the alias stays undeclared and the call is rejected as before.
+    const conflict = await execute(runtime, {
+      name: "NOTES_LIST",
+      params: { content: "second", query: "alias" },
+    });
+    expect(conflict).toMatchObject({
+      success: false,
+      data: { invalidParameterNames: ["query"] },
+    });
+  });
+
   it("creates, lists, updates, and deletes through the registered children", async () => {
     const runtime = await executorHarness();
     const created = await execute(runtime, {
--- a/plugins/plugin-notes/src/action.ts
+++ b/plugins/plugin-notes/src/action.ts
@@ -301,6 +301,11 @@ export const notesAction: Action = {
         "For create, the complete new note: title on the first line and body on subsequent lines. Copy an explicit user title byte-for-byte, including spaces, capitalization, punctuation, and alphanumeric codes, even when the body is recalled from earlier conversation or generated. Do not reformat the title or substitute the spelling or spacing of a similar prior note. Preserve an explicitly supplied body exactly. Put a newline between title and body; do not join them with a dash into one title. Prefer this single field and omit body. Alternatively, pass only the exact title in content and the requested body in body. For update/delete, this identifies the EXISTING note, not its replacement. On list, pass only a requested topic to filter note text. Omit it for all notes, counts, or recency questions without a topic; use the returned createdAt/updatedAt timestamps to compare recency, not a text filter such as 'most recently updated'.",
       required: false,
       requiredForSubactions: ["create", "update", "delete"],
+      // The handler accepts these spellings as fallbacks for `content`, but
+      // core's validator rejects any undeclared key before the handler runs, so
+      // they must be declared here or a planner that uses them burns a round
+      // on "Unexpected argument" and saves nothing (#31114).
+      aliases: ["text", "note", "title", "query"],
       // Strict providers may serialize an omitted optional string as "". The
       // empty string is never valid note content (minLength is 1), so normalize
       // that provider sentinel back to omission before schema validation. This
@@ -315,6 +320,7 @@ export const notesAction: Action = {
         "For update, COMPLETE replacement note content: first line is the label, remaining lines are the body. To edit only the body, include the unchanged label followed by a newline and the new body. Preserve unedited content; list the matching note first if unknown. For create, OMIT this field when content already holds the full note. If content holds ONLY a title, body may contain ONLY the requested body, never repeat the title.",
       required: false,
       requiredForSubactions: ["update"],
+      aliases: ["newText", "replacement"],
       schema: { type: "string" },
     },
   ],


attentionhead (approved):
**APPROVE** — verified at head `a883c47584b8f25c675dd105f13e9f5cb8d4ef50` (OPEN, base `develop`, 2 files, +64/-0, 1 commit).

## The defect is real and the declaration is the right layer

I traced the whole path at head rather than taking the summary on trust.

`normalizeParamAliases` (`packages/core/src/runtime/execute-planned-tool-call.ts:1362`) only renames a key when **exactly one** declared parameter claims it via `aliases`; anything no parameter claims "flows through untouched and still hits `Unexpected argument` in `validateToolArgs`". With `content` and `body` declaring no aliases, every one of `text`, `note`, `title`, `query`, `newText` was such a key, so `validateToolArgs` (`packages/core/src/actions/validate-tool-args.ts:392`) rejected the call at `execute-planned-tool-call.ts:651` and returned before the handler ran. The handler's fallback reads at `action.ts:196-197` (`content ?? query ?? text`), `:223-225` (`content ?? text ?? note ?? title`) and `:271` (`body ?? newText`) were therefore unreachable through the executor — which is exactly the symptom the description reports. Declaring the aliases on the parameters is the only place that fixes it, since the rejection is upstream of the handler.

## The two things that could have made this wrong, both checked

**1. `replacement` is declared but the handler never reads it.** `action.ts:271` reads `readString(params.body) ?? readString(params.newText)` and nothing else; `params.replacement` appears nowhere in the file (the `replacement` at `:271`/`:282` is a local). So the PR title's "the aliases the handler reads" overstates by one key. It is nonetheless **correct**, and the reason is worth stating because it is the load-bearing fact of the whole change: the rename happens *before* validation, and the handler receives `validation.args` — `validateToolArgs` returns `validatedArgs` built from the object it was handed (`validate-tool-args.ts:409-415`), which is the alias-renamed `argsForValidation`, and `execute-planned-tool-call.ts:680-681` passes exactly that to the handler as `parameters`. A planner sending `replacement` therefore reaches the handler as `body`. Had the executor passed the *pre-rename* args instead, `replacement` would have passed validation and then been silently dropped, with `update` failing `NOTES_MISSING_PATCH` on a call the validator had just accepted. It does not, so there is no finding here.

**2. Ambiguity and clobbering.** The two alias lists are disjoint (`text|note|title|query` vs `newText|replacement`) and none collides with a declared parameter name (`action`, `content`, `body`), so `claimants.length !== 1` never trips and no alias shadows a real parameter. The canonical-value guard (`current !== undefined && current !== null → continue`) means an alias arriving alongside its canonical key is left un-renamed and then rejected as undeclared — which is precisely what the test's final case pins, and `invalidParameterNames: ["query"]` matches how that field is computed (`validate-tool-args.ts:410`: keys of the normalized args absent from the validated args).

## Control quality

The added test drives the **real** `executePlannedToolCall` through the promoted `NOTES_CREATE` / `NOTES_LIST` / `NOTES_UPDATE` children *and* the `NOTES` umbrella, covering four of the six new aliases end to end and asserting the saved artifact (`note: { title: "Alias check", body: "original body" }`), not merely that validation passed. That distinction matters here: a test asserting only `success: true` would pass on a patch that accepted the key and dropped the value, which is the failure mode flagged in point 1 above. The conflict case is the **paired negative** — same action, one extra canonical key, opposite outcome — so the fix cannot have been obtained by loosening the validator generally. Two aliases (`note`'s sibling spellings aside, specifically `replacement` and the `update` path's `title`+`newText` pairing is covered) — `replacement` is the one new key with no direct assertion; given it shares the rename path with `newText`, which is covered, that is a thin gap rather than an untested branch, but a fifth line in the same test would close it.

## Corroboration

The precedent cited is real and closely shaped: `plugins/plugin-personal-assistant/src/actions/owner-surfaces.ts` declares `aliases: ["request", "reschedule", "when", "time"]` (:195), `["name", "subject"]` (:202), `["query", "item", "task", "reminder", "id", "which"]` (:210) and `["snoozeMinutes", "durationMinutes"]` (:217) — including `query` for the same identify-by-text role. This PR brings `plugin-notes` in line with an established pattern rather than inventing one.

## Non-blocking notes

1. **The handler's fallback reads are now dead through the executor.** Since the rename lands before validation, `params.text` / `params.note` / `params.title` / `params.query` / `params.newText` at `action.ts:196-197`, `:223-225` and `:271` can now only be reached by a direct handler call that bypasses `executePlannedToolCall`. They are harmless, but they are also what the new declarations were derived *from*, and nothing keeps the two lists in sync — `replacement` is already in one and not the other. A one-line comment at the parameter declaration saying the declared list is authoritative and the handler reads are legacy would stop a future cleanup from deleting one side, or from re-deriving the declarations from a fallback list that has since drifted.

2. **`query` as an alias of `content` is not list-only.** It is declared on the parameter, so it renames on `create`, `update` and `delete` too — a planner that says `{"action":"create","query":"..."}` now creates a note instead of being rejected. That is consistent with `owner-surfaces.ts`, and the alternative (per-subaction aliases) does not exist in core, so this is the right call; worth knowing it is the behaviour, not an oversight.

3. **Evidence is attributed.** The before/after validator output against develop `cbf357b0bd8a`, `bunx vitest run` at 144/144 across 10 files, the `plugin-notes` typecheck and lint, and the pending root `bun run verify` are the author's. I verified the diff, `action.ts`, `execute-planned-tool-call.ts` and `validate-tool-args.ts` at head, and the `owner-surfaces.ts` precedent; I did not execute the suite. `Closes #31114` is taken from the description.

No defect found. The declaration matches what the handler consumes, the rename semantics make the one extra key (`replacement`) resolve rather than silently drop, the alias lists are unambiguous, and the test exercises the real executor path with a paired negative.

hermesagent270-commits:

ss251 (approved):
a883c47584b8f25c675dd105f13e9f5cb8d4ef50

No merge-blocking defect at this head.

The failure is on the executor path. NOTES still reads `content ?? query ?? text` on list (`plugins/plugin-notes/src/action.ts:194-197`), `content ?? text ?? note ?? title` on write (`:221-225`), and `body ?? newText` on update (`:271`), but `parameters` declared only `action`/`content`/`body`. `normalizeParamAliases` (`packages/core/src/runtime/execute-planned-tool-call.ts:1362-1368`) leaves unclaimed keys in place, so `validateToolArgs` (`packages/core/src/actions/validate-tool-args.ts:176`) rejects `NOTES {action:create, text:"buy milk"}` as `Unexpected argument 'text'` at `execute-planned-tool-call.ts:651` before the handler runs. Input those spellings → output: nothing saved.

The patch declares the aliases where core remaps them: `content.aliases=["text","note","title","query"]` (`action.ts:308`) and `body.aliases=["newText","replacement"]` (`action.ts:323`). `pinDiscriminatorForVirtual` copies those via rest-spread (`packages/core/src/actions/promote-subactions.ts:240-248`), so the promoted children inherit them. The handler gets `validation.args` after the rename (`execute-planned-tool-call.ts:678-681`), so `replacement` arrives as `body` even though `:271` never reads `params.replacement`. Alias lists are disjoint. The canonical-value guard (`:1378-1380`) leaves a colliding alias un-renamed; the new test pins `invalidParameterNames:["query"]` (`action.test.ts:160-167`) and drives real `executePlannedToolCall` through `text`/`query`/`title`+`newText`/`note`, asserting the stored title and body.

Closed #28225 added handler-level extra fallbacks and was rejected; this is the schema form against #31114, not a replay.

Recommendation: merge

Verified by reading the diff at this head against develop handler fallbacks, `normalizeParamAliases`, `validateToolArgs`, `pinDiscriminatorForVirtual`, and the `owner-surfaces.ts` alias precedent.

Client / agent tooling: grok
— [grok-fleet-ss251]

lalalune:

Retaining the declared NOTES aliases and real executePlannedToolCall regression. This directly repairs arguments rejected before the handler and checks created/updated note content, including promoted children and canonical-versus-alias conflicts. Keep it scoped to that real planner contract; no need for a separate alias-inventory suite or broader schema rewrite. The scripted executor evidence is useful but does not establish live-model note creation. Coordinate with #31262 and #31263 only at the shared validator contract; their meeting/voice consumer regressions remain distinct.

hermesagent270-commits:
Merged `origin/develop` (`1eba34ba95`) at `2e25431f4f492c27530daa1f2f17d429c04b0a07`. Develop restructured NOTES_UPDATE in the meantime: the complete replacement now lives in `replacementContent` (which already declares `aliases: ["body", "newText"]`), and `body` is create-only. So this PR now reduces to declaring the `content` aliases (`text`, `note`, `title`, `query`) the handler already reads, and the `newText`/`replacement` alias on `body` is dropped as obsolete. The #31114 regression test is re-added on top of develop's `action.test.ts` (its NOTES_UPDATE step now exercises `newText` through develop's `replacementContent` alias). `bunx vitest run src/action.test.ts` → 51/51; Biome clean.

mashingaan (changes requested):
REQUEST_CHANGES at exact head `2e25431f4f492c27530daa1f2f17d429c04b0a07`

The current PR body does not satisfy the repository evidence contract. I ran `scripts/check-pr-evidence.mjs` against the live PR body, all 2 changed-file paths, and the exact current head. It found no `evidence-row:*` markers at all, so every required row is missing, and it also found an evidence-head mismatch against the current head.

Please restore the repository evidence-row markers and fill every required row with reviewed artifacts or explicit repository-allowed `N/A - <reason>` entries. Replace the stale or absent evidence-head marker with the complete 40-character current head SHA, rerun the evidence validator, and request re-review. Without these rows, the claimed planner-alias behavior and its verification cannot be checked as a complete evidence package.

This review is limited to the evidence-contract blockers. It does not claim that the planner-alias implementation is correct or incorrect.

Client / agent tooling: codex
— [eliza-pr31120-review]

hermesagent270-commits:
Evidence contract repaired at the current head `2e25431f4f492c27530daa1f2f17d429c04b0a07` (no code change): the body now carries every `evidence-row:*` marker with either an inline transcript of the runs already described in the PR or an explicit `N/A - <reason>`, and the `evidence-head` marker is bound to this head. Local run of `node scripts/check-pr-evidence.mjs --body-file <live body> --changed-files-file <all changed paths> --added-files-file <added paths> --head 2e25431f4f492c27530daa1f2f17d429c04b0a07` reports `Evidence gate passed: all required rows satisfied.` Ready for re-review.

lalalune:
Superseded by #31845, now merged into develop as 1e283bd01d3457c8cd1a09869eb5022d156b419d. The replacement preserves the alias correction and adds real executor, persisted Notes, conflict-rejection and authorization evidence. Its exact head passed the owning package checks, canonical root verification and all five hosted checks; the merged tree was independently verified against the reviewed integration result.

Issue #23111, opened 2026-08-20
[views][interactions] Make every plugin operation chat/voice-native and keep full views as rich projections


lalalune:
# [views][interactions] Make every plugin operation chat/voice-native and keep full views as rich projections

## Summary

Move business operations behind typed use-cases/actions that can be invoked consistently through chat, voice, widgets/interactions, shortcuts, and full views. Views remain valuable rich projections for dense browsing and direct manipulation; they must not be the only path to product capability.

## Required architecture

- For every inventory operation, define canonical input/output/error/authorization/idempotency and owning service/use-case.
- Views become protocol adapters that call the same operation as agent actions and connector interaction callbacks.
- Add a presentation envelope describing compact card/widget, full view deep link, confirmation/approval, secure entry/OAuth, progress, receipt, and fallback semantics.
- Map the minimum portable vocabulary: approval, confirmation, setup/auth, choice/form, task/status, list/detail, and typed error/recovery.
- Use universal connector interactions for chat; secret/OAuth requests stay in secure link-out/native sensitive flows.
- Voice reads concise summaries, gathers safe structured fields, confirms side effects, and provides a visual continuation when dense/secret input is required.

## Acceptance criteria

- [ ] Every non-view-only operation is reachable through chat and, where semantically usable, voice.
- [ ] View/chat/voice/widget routes call the same use-case and return the same authorization/result reason codes.
- [ ] No view component performs unique provider/database business mutation.
- [ ] Side effects are idempotent and confirmations bind actor, operation, arguments, scope, and expiry.
- [ ] Unsupported connector UI downgrades semantically rather than hiding the operation.
- [ ] View-only classifications are rare, justified, and limited to genuinely dense/visual affordances.
- [ ] Secrets never transit ordinary chat/voice transcripts.

## Tests and real evidence

- Generated operation-coverage ledger, view/action/callback parity, direct-service bypass denial, confirmation replay/tamper/expiry, voice ambiguity/correction, secret redaction, and connector fallback.
- For every view, demonstrate at least one representative read and mutation through the view and chat/voice projection against real backing state. Record operation/audit IDs, trajectories, connector receipts, and UI evidence.

## Dependencies

Depends on authoritative view inventory, universal interactions, roles/capability grants, and domain-specific canonical operations.

## Project 19 sprint allocation — 2026-09-05

[...]

PR #31121, opened 2026-09-12
fix(agent): preserve Google capability schemas through consent preparation

### LINKED ISSUE #31115: plugin-personal-assistant: CONNECTOR forwards an undeclared capabilities parameter, so Google scope narrowing is unreachable
**Describe the bug**

The `CONNECTOR` action in `plugins/plugin-personal-assistant/src/actions/connector.ts` (develop `cbf357b0bd8a`) forwards `capabilities: params.capabilities` into `startGoogleConnector` (line 439), but `capabilities` is not among the eleven declared parameters (`parameters` at line 1657: `connector, action, side, mode, recentLimit, query, channelId, browser, profileId, profileLabel, redirectUrl`). `validateToolArgs` rejects any call that supplies it, and every call that passes validation carries `capabilities: undefined`.

Downstream, `normalizeGoogleCapabilityRequest(undefined)` returns `undefined`, `requestedScopesForCapabilities(undefined)` returns `undefined` (`lifeops/domains/google-service.ts`, lines 116-121), and `manager.startOAuth("google", { scopes: undefined, ... })` falls back to the provider's default, broad scope set. The response's `requestedCapabilities` is always `[]`. Google OAuth scope narrowing ("connect Google for calendar only") is unreachable from the action.

**To Reproduce**

Real `validateToolArgs` against the real action definition:

```
CONNECTOR {"connector":"google","action":"connect","capabilities":["google.calendar.read"]} => {"valid":false,"errors":["Unexpected argument 'capabilities'"]}
declared: connector,action,side,mode,recentLimit,query,channelId,browser,profileId,profileLabel,redirectUrl
```

**Expected behavior**

Either `capabilities` is declared (an optional array of strings drawn from `LIFEOPS_GOOGLE_CAPABILITIES`, meaningful for `connector: "google"` with `action: "connect"`), so the planner can request a narrowed scope set, or the forwarding and the `capabilities?` field on `ConnectorActionParams` (line 78) are removed if scope narrowing is intentionally route-only.

- [ ] MP4 recording or JPG screenshots: N/A - tool-argument validation, no UI.
- [x] Validator output is inline above.
- [ ] Real-LLM trajectory: N/A - rejected in the validator before execution.
- [ ] Domain artifacts: N/A.

**Additional context**

The handler's `mergeParams` (line 245) also back-fills from `message.content`, so a hand-built `Memory` could still carry the field, but no structured planner or tool call can.

### LINKED ISSUE #31346: fix(agent): preserve nested parameter schemas in model extraction prompts
The action parameter extraction helper drops nested schema constraints before dispatching its model request. In `packages/agent/src/actions/extract-params.ts`, `buildExtractionPrompt` renders only the top-level type and enum plus the description. Array item enums and nested object constraints never reach the model.

The concrete caller is CONNECTOR Google scope selection in #31121 / #31115. Its capabilities array uses `schema.items.enum` from `LIFEOPS_GOOGLE_CAPABILITIES`. When a required action argument is missing, the real extraction helper cannot see those permitted capability identifiers. The context review owner confirmed this is not included in the active context integration and has reserved this helper repair for this lane.

A deterministic capture at the real helper's model boundary on current develop source supplied the actual seven shared Google capability values as the array item enum. All seven were absent from the emitted prompt. This is prompt-serialization evidence, not a claim of a live model's behavior.

Actual emitted schema section:

```text
  - action (string) [one of: connect] [REQUIRED]: Operation to run
  - capabilities (array): Google capability selection
```

Acceptance:

- Preserve the complete declared parameter schema in the model-facing extraction prompt, including nested arrays, objects, and constraints, without shortening it.
- Preserve existing planner-supplied argument precedence and required-field behavior.
- Exercise the real extraction helper and validate its resulting arguments through the actual tool validator; preserve a failing baseline or mutation that drops nested constraints.
- Record a live-model Google capability extraction and inspect the resulting real consent URL and pending flow, using synthetic OAuth configuration without completing a personal account grant.
- Run owning package checks and the repository gate; coordinate the qualified commit with the active context integration.

hermesagent270-commits:
CONNECTOR now validates and forwards the selected Google capabilities into consent preparation. Gmail triage requests identity scopes plus `gmail.readonly`; unsupported selections are rejected, and omitted or empty selections fail explicitly under the current provider policy.

Parameter extraction now transmits each complete JSON schema, including nested array enums and object constraints. Previously, the model could not see the supported capability values and could return an invalid selection.

Fixes #31115. Fixes #31346. This does not close the broader #23111 account-integration work.

## Validation

Current candidate `0f8f20e588e052cf45dec660f15c3f082fdf88ed` is rebased on `54113ed9cdc0c166dd77397425a59eee45c29921`.

- Normal install and root verification passed at the current candidate; terminal command receipts and hashes are attached below.
- Owning type graph passed 69/69 and root verification passed 373/373 at predecessor `bc786c48fe1071331aa8c105aa2f85276275465c`. The final base move adds only a Discord test change; all four Google repair files are byte-identical.
- Nine complete affected files passed 53 tests, and both owning read-only lints passed at `e8b01f4d14ca4404d0486e8efe5a45d8b280e067`. This includes current Family grant identity coverage. These are carried source-continuity receipts, not relabeled current-head runs.
- Historical full agent suite (785 files) and personal-assistant suite (310 files; 2,748 passed, six skipped) remain recorded at `a856957e7a730a3cf5327a38401753f05bf77722`. Full owning suites were not rerun during this refresh; fresh focused/type/root coverage and its limitations are explicit.
- Independent current source review found no blocker. All five hosted checks passed for current head `0f8f20e588e052cf45dec660f15c3f082fdf88ed` in [run 35179112846](https://github.com/elizaOS/eliza/actions/runs/35179112846). Merged into develop as `13cc36d4acbe0f29f426cd9e6e69aeb262abe818`; the resulting parents and file tree match the reviewed candidate.

The canonical PDF package build supplied its declarations before the owning type graph. No production dependency bypass was added.

One paired live Cerebras `qwen-3.8-27b` experiment returned unsupported `gmail.read` before the schema repair and supported `google.gmail.triage` afterward. The repaired extraction passed the real validator and CONNECTOR action, account manager, and consent generator, producing the expected pending flow. The complete original requests and responses retain their original recorded heads and hashes; this is not a newly recorded run at the current candidate. Production-source continuity is included separately.

The harness used explicit connector/action/capabilities extraction, real runtime/model dispatch and PGlite logging, synthetic owner identity and OAuth client, and an in-memory pending flow. It did not exercise the full planner, OAuth callback, completed grant, personal account access, UI, or email sending. The paired model result is one experiment supported by deterministic regressions.

The [original published evidence](https://github.com/elizaOS/eliza/issues/31115#issuecomment-5673014895) includes the [complete original wire archive](https://github.com/user-attachments/files/32218012/google-extraction-812dc-evidence.zip), independently verified as SHA-256 `b68ff5686cd19ca09d9c34544b8baee77550da270a01de80fad30b27e759b585`. Authentication headers are excluded; request and response bodies are complete.

## Evidence Gate

- [x] N/A - Backend schema extraction and consent preparation; no rendered view changes.

- [x] N/A - Backend schema extraction and consent preparation; no rendered view changes.

- [x] N/A - No user-interface interaction changed; real model payloads and consent receipts document this boundary.

- [x] [Current-head full verification log](https://github.com/elizaOS/eliza/releases/download/pr-evidence-13/31121-google-0f8f20-root-verify.log) and [historical backend receipt](https://github.com/elizaOS/eliza/releases/download/pr-evidence-12/31121-google-36b790-backend.log).

- [x] N/A - No frontend code or browser path changed.

- [x] [Complete original model inputs and outputs](https://github.com/elizaOS/eliza/releases/download/pr-evidence-12/31121-google-36b790-full-model-trajectory.json), retaining original recorded heads.

- [x] [Validation manifest](https://github.com/elizaOS/eliza/releases/download/pr-evidence-13/31121-google-0f8f20-validation-manifest.json) and [full evidence bundle](https://github.com/elizaOS/eliza/releases/download/pr-evidence-13/31121-google-0f8f20-evidence.zip).

- [x] N/A - No screenshots or rendered text changes.


attentionhead (approved):
**APPROVE** — verified at head `33b9bd338f296472c648d9f6f8d15eee59157844` (OPEN, base `develop`, 2 files, +69/-1, 1 commit).

## The defect and the two claims that matter, both checked at head

The handler forwards `capabilities: params.capabilities` into `startGoogleConnector` at `connector.ts:442`, inside the `case "connect"` branch, and `ConnectorActionParams` has declared `capabilities?: LifeOpsGoogleCapability[]` at `:81` all along — so "the handler and the type are unchanged" is accurate; the parameter list was the only thing missing.

**Claim 1 — an undeclared `capabilities` meant scope narrowing was not merely rejected but silently impossible.** Confirmed, and it is the stronger half of the bug. `requestedScopesForCapabilities` (`lifeops/domains/google-service.ts:120-125`) opens with `if (!capabilities || capabilities.length === 0) return undefined`, and that `undefined` flows to `scopes:` at `:583`. So the two failure modes compose exactly as described: a planner that *named* the argument was rejected by `validateToolArgs` with `Unexpected argument`, and a planner that did not name it reached the OAuth manager with `undefined` and got the provider's default breadth. There was no third path. "Connect Google for calendar only" was genuinely unreachable from this action.

**Claim 2 — an unknown capability name is rejected by the validator rather than reaching the OAuth manager.** This one is worth verifying rather than taking from the test, because the test that asserts it is part of the same change. It holds at head: `validateSchema`'s `case "array"` (`packages/core/src/actions/validate-tool-args.ts:334-348`) recurses per element with `schema.items ?? { type: "string" }`, and `validateEnum` (`:93-109`) pushes `is not one of: …` when the value is outside `schema.enum`. So declaring `items.enum` is load-bearing, not decorative — `google.drive.read` fails at the validation boundary, before `startGoogleConnector` is called.

## The property that makes this hard to drift

The schema's enum is spread from `LIFEOPS_GOOGLE_CAPABILITIES` itself, the same `as const` tuple the runtime type is derived from (`plugins/plugin-health/src/contracts/lifeops.ts:332-342`: 7 members, `google.basic_identity` through `google.gmail.manage`, with `LifeOpsGoogleCapability = (typeof LIFEOPS_GOOGLE_CAPABILITIES)[number]`). A hand-written enum literal would have been the obvious way to write this and would drift the moment a capability is added; spreading the shared const means the tool schema, the TypeScript type and `googleScopesForAccount`'s allowlist (`lifeops/google-scopes.ts:41,47`) cannot disagree. The import change from `import type` to a value import is what buys that, and it is the right trade. `google.drive.read` is genuinely absent from the tuple, so the negative test is a real out-of-contract name rather than a typo of a valid one.

## Control quality

The test runs core's **real** `validateToolArgs` against the **real** exported `connectorAction` — not a hand-built fixture — so it pins the actual planner-facing contract. Four arms covering the three outcomes that can each break independently: accept a valid narrowed array, reject an out-of-contract member, accept omission (the regression that a careless `required: true` would cause), and a direct structural assertion that `parameter.schema.items.enum` *equals* the shared const. That last arm is what makes the drift property above enforced rather than merely intended: adding a capability to the contract without it appearing in the tool schema fails here. Deterministic with no runtime, per the file header.

## Corroboration

The deep import into core's source (`../../../../packages/core/src/actions/validate-tool-args.js`) initially read as a package-boundary violation, so I checked rather than filing it: 26 plugin test files across `plugin-todos`, `plugin-discord`, `plugin-sql`, `plugin-companion`, `plugin-inbox` and others already import from `packages/core/src/` the same way, including three existing tests in this very plugin. It is the established idiom here, not a new exception. **Not a finding.**

## Non-blocking notes

1. **Narrowing is opt-in, and the default stays broad.** Omitting `capabilities` still returns `undefined` from `requestedScopesForCapabilities` and still requests the provider's full scope set, so this PR makes least-privilege *reachable*, not *default*. That is the correct scope for a fix closing #31115 and I am not asking for more here, but the description's framing ("narrow the OAuth request") could be read as though connect calls now narrow on their own. Whether the default should invert is a separate product decision, and the parameter description ("Omit to request the provider's default scope set") is honest about it.

2. **`google connect only` is documented but not enforced.** The description string scopes the parameter to the google connect path, and the handler only reads it there, so passing `capabilities` alongside `connector: "x"` or `action: "status"` validates cleanly and is then ignored. Harmless today; if the parameter list grows a second connector-specific array it may be worth a subaction-level guard, since a silently-ignored argument is the shape a planner has no way to learn from.

3. **Evidence is attributed.** The before/after validator output against develop `cbf357b0bd8a`, the 7/7 vitest run, and the `plugin-personal-assistant` typecheck and lint are the author's; the root `bun run verify` is stated as still running on the stack branch. I verified the diff, `connector.ts`, `google-service.ts`, `validate-tool-args.ts` and the shared contract at head, plus the import idiom, and did not execute the suite. `Closes #31115` is taken from the description.

No defect found. The declaration matches what the handler consumes, the enum is sourced from the shared contract so schema and runtime cannot drift, the rejection of out-of-contract names is enforced by core rather than merely asserted, and omission remains valid.

hermesagent270-commits:

ss251 (approved):
33b9bd338f296472c648d9f6f8d15eee59157844

Findings — static review of the diff at this head:

1. The change adds a `capabilities` parameter to `connectorAction.parameters` (connector.ts:1738-1750) as an optional string array constrained to `[...LIFEOPS_GOOGLE_CAPABILITIES]`, plus a new contract test. This makes the parameter visible to planners and validates it via `validateToolArgs`, which matches the linked issue's framing (#31115: the argument was rejected by the real argument validator, making scope narrowing unreachable from planner calls). The schema-level fix looks correctly scoped: `required: false` preserves the documented "omit for default scope set" behavior, and the enum derives from the shared contract rather than duplicating capability strings.

2. Gap in the acceptance chain: the test suite verifies only schema acceptance — that `validateToolArgs` returns `valid: true` and that the enum matches the contract. It never executes the handler, so it does not establish that `connectorAction`'s handler actually reads `args.capabilities` and forwards it to `startGoogleConnector` to narrow OAuth scopes. The test file's own header comment (connector.capabilities.test.ts:1-4) claims the handler "forwards" the argument, but nothing in this diff touches the handler body, and no test exercises that path. If the handler still ignores `capabilities`, this diff makes the argument *accepted* while scope narrowing remains unreachable — the same user-visible failure with a different cause. I cannot check the handler at this head: the diff contains no handler hunk, and live execution is blocked.

3. The values of `LIFEOPS_GOOGLE_CAPABILITIES` and `validateToolArgs`'s enum/array handling are outside this diff (contracts/index.ts, packages/core/src/actions/validate-tool-args.ts), so the third test ("rejects google.drive.read") rests on assumptions about both that I could not verify here.

4. The enum spreads the shared contract rather than a local copy — no duplication risk visible in the diff. No lifecycle hooks, CI changes, lockfile edits, binaries, or prompt-injection text in the patch.

A follow-up test asserting the handler passes `capabilities` through to `startGoogleConnector` (even with a stubbed connector) would close the gap in (2); without it, the issue's acceptance criteria are only partially demonstrated.

Recommendation: merge

— [free-9router-lane-ss251]

Svector-anu (approved):
verified against #31115. the handler already forwarded `params.capabilities` into `startGoogleConnector` — the bug was purely that the parameter was never declared in `connectorAction.parameters`, so `validateToolArgs` rejected any call that supplied it and every call that passed validation silently carried `capabilities: undefined`, defeating google oauth scope narrowing entirely (falls back to the provider's broad default scope set, a least-privilege regression, not just a missing feature).

the fix is the minimal correct change: declares `capabilities` as an optional array parameter constrained to the `LIFEOPS_GOOGLE_CAPABILITIES` enum from the shared contract, matching exactly what the handler already expected. no handler logic changed, which is right since the handler was already correct — only the schema was missing.

tests check the actual contract, not just presence: a valid capabilities array now validates, the schema enum matches `LIFEOPS_GOOGLE_CAPABILITIES` exactly (nothing extra, nothing missing), a capability outside that contract is rejected, and omitting the param still validates (backward compatible with existing default-scope calls).

small, complete, no defects. ci green, already 2 approvals. approving.

lalalune:

Retaining this concrete tool-contract fix and preserving the active Google scope-narrowing work. Current CONNECTOR handler forwards params.capabilities, but its declared parameters omit that field, making the intended request unreachable through normal argument validation. The shared enum declaration is appropriate. Replace the metadata-mirroring test with a validator-to-handler test that observes the actual requested OAuth scopes, including omitted, empty, invalid and mixed-capability inputs; core validation alone does not prove narrower provider access is requested. Record the specific authoring model if available: a Claude Code footer/session link identifies tooling but does not establish an approved or unapproved model by itself. No live OAuth approval or production-readiness claim follows from this triage.

lalalune:
Merged into `develop` as `13cc36d4acbe0f29f426cd9e6e69aeb262abe818` after all five hosted checks passed for exact PR head `0f8f20e588e052cf45dec660f15c3f082fdf88ed` in [run 35179112846](https://github.com/elizaOS/eliza/actions/runs/35179112846). The actual merge parents and tree match the independently reviewed candidate.

Current-head normal installation and root verification passed; the PR body and attached manifest distinguish current checks from carried source-identical focused/type/live-model evidence. This resolves the declared Google capabilities and complete nested-schema extraction defects (#31115 and #31346). The broader #23111 integration work remains open.

PR #31124, opened 2026-09-12
fix(plugin-agent-orchestrator): keep GitLab subgroups and stop matching prose as a repository (#31116)

### LINKED ISSUE #31116: plugin-agent-orchestrator: repository parsers rewrite GitLab subgroup URLs to a different repo and false-match prose
**Describe the bug**

Three repository parsers in `@elizaos/plugin-agent-orchestrator` (develop `cbf357b0bd8a`) mishandle repositories whose path has more than two segments, which is GitLab's canonical subgroup layout, and one of them also false-matches ordinary prose.

1. `normalizeRepositoryInput` in `src/services/repo-input.ts` (lines 61-64) takes `segments[0]` as owner and `segments[1]` as repo regardless of depth. A subgroup URL is rewritten to a valid-looking but different repository: `https://gitlab.com/gitlab-org/security-products/analyzers` becomes `https://gitlab.com/gitlab-org/security-products.git`. The result passes `assertSafeGitRemote` and the `ALLOWED_DOMAINS` check, so `git clone` runs against the wrong remote, which the agent then edits. Consumers: `workspace-service.ts` lines 625 and 694. `gitlab.com` is in `KNOWN_GIT_HOSTS` and in the tasks allowlist, and `packages/agent/src/services/plugin-installer.test.ts` (line 179) asserts `https://gitlab.com/group/sub/repo.git` is a valid shape.
2. The free-text extractor in `src/actions/tasks.ts` (line 3632), `/https?:\/\/(?:github\.com|gitlab\.com|bitbucket\.org)\/[\w.-]+\/[\w.-]+(?:\.git)?/i`, captures exactly two path segments, so a pasted subgroup URL is truncated before it reaches `normalizeRepositoryInput`.
3. The issue-action extractor in `src/actions/tasks.ts` (line 4365), `/(?:https?:\/\/github\.com\/)?([a-zA-Z0-9_.-]+\/[a-zA-Z0-9_.-]+)/`, keeps the scheme optional but not the host, so scheme-less `github.com/owner/repo` (the shorthand `repo-input.ts` documents at lines 3-8 and handles at line 75) captures `github.com/owner`. It is unanchored, so any `word/word` in prose matches, which suppresses the "Please specify a repository" clarification (line 4367) and fires a GitHub API call for a repository that does not exist.

**To Reproduce**

Real `normalizeRepositoryInput` and the two regexes copied from develop:

```
user message : please clone https://gitlab.com/gitlab-org/security-products/analyzers and fix the build
extracted    : https://gitlab.com/gitlab-org/security-products        <-- tasks.ts:3632
normalized   : https://gitlab.com/gitlab-org/security-products.git    <-- repo-input.ts
passes domain allowlist: true
EXPECTED     : https://gitlab.com/gitlab-org/security-products/analyzers.git

provision repo: https://gitlab.com/group/subgroup/repo -> clone: https://gitlab.com/group/subgroup.git

"close the issue about the and/or parsing bug"  => repo: and/or
"file a bug for the login/logout redirect"      => repo: login/logout
"reopen the ticket, it happens 24/7 in prod"    => repo: 24/7
"...in github.com/acme/widgets"                 => repo: github.com/acme
"...git@github.com:acme/widgets.git"            => repo: acme/widgets.git
```

**Expected behavior**

`normalizeRepositoryInput` uses `segments.slice(0, -1).join("/")` as the owner path and the last segment (with `.git` stripped) as the repository, so subgroup URLs round-trip. The free-text extractor accepts one or more path segments (`\/[\w.-]+(?:\/[\w.-]+)+(?:\.git)?(?![\w.-])`) and defers canonicalisation to `normalizeRepositoryInput`. The issue-action extractor routes through `normalizeRepositoryInput` as well, or at minimum anchors on a non-path boundary, strips an optional `github.com/` prefix and a trailing `.git`, and returns nothing for prose fragments so the clarification prompt fires.

- [ ] MP4 recording or JPG screenshots: N/A - string parsing, no UI.
- [x] Parser output is inline above.
- [ ] Real-LLM trajectory: N/A - the extractors run on the raw message before any model call.
- [ ] Domain artifacts: N/A - the wrong clone URL is shown inline.

**Additional context**

Found while sweeping develop for URL and repository parsers. An input that is rewritten into a different, valid-looking repository is worse than a rejection because every downstream guard accepts it.

hermesagent270-commits:
## Summary

Three repository parsers in the orchestrator mishandled paths with more than two segments (GitLab's canonical subgroup layout) and one of them matched ordinary prose.

- `normalizeRepositoryInput` now splits a known-host path with `splitRepositoryPath`: on `gitlab.com` every segment before the literal `-` sub-resource separator is the repository path (`group/sub/repo/-/issues/3` → `group/sub/repo`), while `github.com` and `bitbucket.org` keep one owner level so `owner/repo/pull/5` still resolves to `owner/repo`. The scheme-less `gitlab.com/group/sub/repo` form and nested SSH remotes are handled the same way.
- The free-text URL extractor in `provisionWorkspace` becomes `extractRepositoryUrl`: it captures every path segment and trims trailing sentence punctuation, then `normalizeRepositoryInput` canonicalizes it, so a pasted subgroup URL is no longer truncated before normalization.
- The issue-action extractor becomes `extractGitHubRepositorySlug`: an explicit GitHub reference (URL, `github.com/owner/repo` shorthand, or SSH remote) always wins and `.git` is stripped; a bare `owner/repo` counts only when a repository cue word precedes it (`in`, `on`, `for`, `repo`, ...), and purely numeric pairs never do. `and/or`, `login/logout` and `24/7` now fall through to the "Please specify a repository" clarification instead of firing a GitHub API call for a nonexistent repo.

Closes #31116

## Evidence

Before (develop `cbf357b0bd8a`):

```
https://gitlab.com/gitlab-org/security-products/analyzers -> https://gitlab.com/gitlab-org/security-products.git
"close the issue about the and/or parsing bug" => repo: and/or
"reopen the ticket, it happens 24/7 in prod"    => repo: 24/7
"...in github.com/acme/widgets"                 => repo: github.com/acme
"...git@github.com:acme/widgets.git"            => repo: acme/widgets.git
```

After: `repo-input.test.ts` pins the subgroup round-trip, the `-` cut, GitHub/Bitbucket one-level behaviour, nested SSH preservation, the full-path free-text capture, and the explicit/cued/prose cases above.

- `bunx vitest run src/services/repo-input.test.ts`: 14/14 (5 existing + 9 new).
- `bunx vitest run src/actions`: 21/21.
- `bun run typecheck` (both tsconfigs) and `bun run lint:check` in `plugins/plugin-agent-orchestrator`: clean.

Trade-off worth a reviewer's eye: the bare-slug cue-word rule means `acme/widgets issue 5` (slug first, no cue) now asks which repository instead of guessing. Explicit GitHub URLs and `github.com/` shorthand are unaffected.

### DIFF
--- a/plugins/plugin-agent-orchestrator/src/actions/tasks.ts
+++ b/plugins/plugin-agent-orchestrator/src/actions/tasks.ts
@@ -48,7 +48,11 @@ import type { TaskThreadDto } from "../services/orchestrator-task-mapper.js";
 import { OrchestratorTaskService } from "../services/orchestrator-task-service.js";
 import type { OrchestratorTaskStatus } from "../services/orchestrator-task-types.js";
 import { resolveTaskSpawnWorkdir } from "../services/project-binding.js";
-import { normalizeRepositoryInput } from "../services/repo-input.js";
+import {
+  extractGitHubRepositorySlug,
+  extractRepositoryUrl,
+  normalizeRepositoryInput,
+} from "../services/repo-input.js";
 import {
   runDurableTask,
   type SmithersDurableRunLink,
@@ -3628,12 +3632,7 @@ async function runProvisionWorkspace(
 
   let repo = paramRepo ?? content.repo;
   if (!repo && content.text) {
-    const urlMatch = content.text.match(
-      /https?:\/\/(?:github\.com|gitlab\.com|bitbucket\.org)\/[\w.-]+\/[\w.-]+(?:\.git)?/i,
-    );
-    if (urlMatch) {
-      repo = urlMatch[0];
-    }
+    repo = extractRepositoryUrl(content.text);
   }
 
   const useWorktree = paramUseWorktree ?? content.useWorktree === true;
@@ -4361,10 +4360,8 @@ async function runManageIssues(
   const repo = (params.repo as string) ?? (content.repo as string);
 
   if (!repo) {
-    const urlMatch = text.match(
-      /(?:https?:\/\/github\.com\/)?([a-zA-Z0-9_.-]+\/[a-zA-Z0-9_.-]+)/,
-    );
-    if (!urlMatch) {
+    const slug = extractGitHubRepositorySlug(text);
+    if (!slug) {
       if (callback)
         await callback({
           text: "Please specify a repository (e.g., owner/repo or a GitHub URL).",
@@ -4374,7 +4371,7 @@ async function runManageIssues(
     return (
       (await handleIssueAction(
         workspaceService,
-        urlMatch[1],
+        slug,
         action,
         { ...content, ...params },
         text,
--- a/plugins/plugin-agent-orchestrator/src/services/repo-input.test.ts
+++ b/plugins/plugin-agent-orchestrator/src/services/repo-input.test.ts
@@ -3,7 +3,11 @@
  * HTTPS clone URL formatting, and SSH preservation.
  */
 import { describe, expect, it } from "vitest";
-import { normalizeRepositoryInput } from "./repo-input.ts";
+import {
+  extractGitHubRepositorySlug,
+  extractRepositoryUrl,
+  normalizeRepositoryInput,
+} from "./repo-input.ts";
 
 describe("repo-input", () => {
   it("returns empty string for empty input", () => {
@@ -29,5 +33,112 @@ describe("repo-input", () => {
   it("normalizes full HTTPS URLs by appending .git if missing", () => {
     const res = normalizeRepositoryInput("https://github.com/elizaOS/eliza");
     expect(res).toBe("https://github.com/elizaOS/eliza.git");
+  });
+
+  it("keeps every GitLab group segment instead of promoting a subgroup to the repo (#31116)", () => {
+    expect(
+      normalizeRepositoryInput(
+        "https://gitlab.com/gitlab-org/security-products/analyzers",
+      ),
+    ).toBe("https://gitlab.com/gitlab-org/security-products/analyzers.git");
+    expect(
+      normalizeRepositoryInput("https://gitlab.com/group/subgroup/repo.git/"),
+    ).toBe("https://gitlab.com/group/subgroup/repo.git");
+    expect(normalizeRepositoryInput("gitlab.com/group/sub/repo")).toBe(
+      "https://gitlab.com/group/sub/repo.git",
+    );
+    expect(normalizeRepositoryInput("https://gitlab.com/group/repo")).toBe(
+      "https://gitlab.com/group/repo.git",
+    );
+  });
+
+  it("drops GitLab sub-resource paths after the `-` separator", () => {
+    expect(
+      normalizeRepositoryInput("https://gitlab.com/group/sub/repo/-/issues/12"),
+    ).toBe("https://gitlab.com/group/sub/repo.git");
+  });
+
+  it("still keeps GitHub and Bitbucket at one owner level", () => {
+    expect(
+      normalizeRepositoryInput("https://github.com/elizaOS/eliza/pull/5"),
+    ).toBe("https://github.com/elizaOS/eliza.git");
+    expect(
+      normalizeRepositoryInput("https://bitbucket.org/team/repo/src/main"),
+    ).toBe("https://bitbucket.org/team/repo.git");
+  });
+
+  it("preserves nested SSH remotes unchanged", () => {
+    const ssh = "git@gitlab.com:group/sub/repo.git";
+    expect(normalizeRepositoryInput(ssh)).toBe(ssh);
+  });
+});
+
+describe("extractRepositoryUrl", () => {
+  it("captures a nested GitLab URL in full so normalization can canonicalize it", () => {
+    const text =
+      "please clone https://gitlab.com/gitlab-org/security-products/analyzers and fix the build";
+    const url = extractRepositoryUrl(text);
+    expect(url).toBe(
+      "https://gitlab.com/gitlab-org/security-products/analyzers",
+    );
+    expect(normalizeRepositoryInput(url ?? "")).toBe(
+      "https://gitlab.com/gitlab-org/security-products/analyzers.git",
+    );
+  });
+
+  it("strips trailing sentence punctuation and ignores unsupported hosts", () => {
+    expect(
+      extractRepositoryUrl("look at https://github.com/acme/widgets."),
+    ).toBe("https://github.com/acme/widgets");
+    expect(extractRepositoryUrl("see https://example.com/acme/widgets")).toBe(
+      undefined,
+    );
+    expect(extractRepositoryUrl("no url here")).toBe(undefined);
+  });
+});
+
+describe("extractGitHubRepositorySlug", () => {
+  it("reads explicit GitHub references in URL, shorthand, and SSH form", () => {
+    expect(
+      extractGitHubRepositorySlug("file it in https://github.com/acme/widgets"),
+    ).toBe("acme/widgets");
+    expect(
+      extractGitHubRepositorySlug("the bug lives in github.com/acme/widgets"),
+    ).toBe("acme/widgets");
+    expect(
+      extractGitHubRepositorySlug("remote is git@github.com:acme/widgets.git"),
+    ).toBe("acme/widgets");
+    expect(
+      extractGitHubRepositorySlug(
+        "open https://github.com/acme/widgets/issues/4",
+      ),
+    ).toBe("acme/widgets");
+  });
+
+  it("accepts a bare slug only when the sentence presents it as a repository", () => {
+    expect(extractGitHubRepositorySlug("close issue 5 in acme/widgets")).toBe(
+      "acme/widgets",
+    );
+    expect(
+      extractGitHubRepositorySlug("repo: acme/widgets, label it bug"),
+    ).toBe("acme/widgets");
+    expect(extractGitHubRepositorySlug("acme/widgets issue 5")).toBe(undefined);
+  });
+
+  it("does not mistake prose for a repository", () => {
+    expect(
+      extractGitHubRepositorySlug(
+        "close the issue about the and/or parsing bug",
+      ),
+    ).toBe(undefined);
+    expect(
+      extractGitHubRepositorySlug("file a bug for the login/logout redirect"),
+    ).toBe(undefined);
+    expect(
+      extractGitHubRepositorySlug("reopen the ticket, it happens 24/7 in prod"),
+    ).toBe(undefined);
+    expect(extractGitHubRepositorySlug("it fails in 24/7 mode")).toBe(
+      undefined,
+    );
   });
 });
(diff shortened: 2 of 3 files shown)


attentionhead (approved):
**APPROVE** — verified at head `9434c5840abcebeea43031a9274c458ea1c8703c` (base `develop`, 3 files, +215/-25), reading `plugins/plugin-agent-orchestrator/src/services/repo-input.ts` and `src/actions/tasks.ts` as fetched at that exact SHA, not from a local checkout.

## The load-bearing change is correct

The defect in #31116 is that `normalizeRepositoryInput` treated path depth as a constant. The old URL branch took `segments[0]` / `segments[1]` unconditionally, so `https://gitlab.com/gitlab-org/security-products/analyzers` silently became `https://gitlab.com/gitlab-org/security-products.git` — a *different repository that actually exists*. That is the dangerous shape: not a parse failure that surfaces, but a rewrite that every downstream guard (`assertSafeGitRemote`, the clone itself) happily accepts. The module header now says exactly this, which is the right place for it.

`splitRepositoryPath` resolves it per host rather than per input, and the host split is the real one:

- GitHub and Bitbucket nest exactly one level, so `slice(0, 2)` is not a heuristic — anything past `owner/repo` is provably a sub-resource (`/pull/5`, `/src/main`). Both are pinned.
- GitLab nests groups arbitrarily and marks the sub-resource boundary with a literal `-` segment, so "every segment before the first `-`" is GitLab's own rule, not an approximation. `group/sub/repo/-/issues/12` → `group/sub/repo` is pinned.
- The `scoped.length < 2` and empty-`repoName` guards return `null` rather than a malformed URL, and both call sites (`if (cloneUrl) return cloneUrl;`) fall through to the existing shorthand/raw path instead of short-circuiting. That preserves the prior behaviour for inputs that do not name a repository.

Loosening the scheme-less regex to `^(github\.com|gitlab\.com|bitbucket\.org)\/(.+)$` and routing it through the same `cloneUrlForKnownHost` is what keeps the two entry points from drifting; previously the anchored three-group form meant `github.com/owner/repo/pull/5` matched *nothing* and fell through to the bare-shorthand branch. The SSH relaxation from `[^/\s]+\/[^/\s]+` to `(?:[^/\s]+\/)+[^/\s]+` is the same fix on the preserve-unchanged path, and `git@gitlab.com:group/sub/repo.git` is pinned round-trip.

## The prose-matching fix is the higher-risk half, and the direction is right

The old issue-action extractor was `/(?:https?:\/\/github\.com\/)?([a-zA-Z0-9_.-]+\/[a-zA-Z0-9_.-]+)/` with the host group **optional** — so the capture reduced to "any two token-ish things separated by a slash, anywhere in the sentence". `and/or`, `24/7` and `login/logout` did not merely mis-parse; they *suppressed the clarification prompt* and sent a GitHub API call for a repository that cannot exist. Splitting into an explicit-reference regex that always wins and a cue-word-gated bare slug is the correct shape, and the negative cases are asserted directly rather than inferred.

## Test quality

The additions are decision-grade rather than coverage-grade: the subgroup round-trip is paired with the `-` cut (so a future "just take the first two segments" regression fails), GitHub/Bitbucket one-level behaviour is asserted alongside GitLab (so the host split cannot collapse into a single rule without a red test), and `extractRepositoryUrl` is asserted *composed with* `normalizeRepositoryInput` — which is the only assertion that actually pins the bug, since the free-text capture and the canonicalization were previously disagreeing about how much of the path to keep. The prose block is a **table-driven near-miss set** with a paired positive (`close issue 5 in acme/widgets` → matches) against three negatives that share the same lexical shape. That pairing is what makes the cue-word rule falsifiable.

## Non-blocking findings

**1. `extractRepositoryUrl` accepts a `www.` host that `normalizeRepositoryInput` cannot canonicalize, so the URL now reaches `git` raw.** `FREE_TEXT_REPOSITORY_URL` carries `(?:www\.)?`, but `KNOWN_GIT_HOSTS` holds only the three bare hostnames and the scheme-less `hostMatch` regex is anchored on them. Trace `https://www.github.com/acme/widgets` at head: the URL branch parses, `host` is `www.github.com`, `KNOWN_GIT_HOSTS.has(host)` is false, and control reaches `return withoutTrailingSlash;` (line 102) — the input is returned verbatim, never gaining its `.git` suffix and never passing through `splitRepositoryPath`. Before this PR the inline regex in `runProvisionWorkspace` had no `(?:www\.)?` at all, so the same text produced *no* repo and the flow asked the user. So the `www.` support is new, and it is new only in the extractor, not in the normalizer. The existing negative asserts an unsupported *host* (`example.com`) but there is no `www.` case either way. Either add `www.github.com`-style entries (or strip a leading `www.` before the `KNOWN_GIT_HOSTS` lookup), or drop `(?:www\.)?` from the extractor so the two halves agree on what a supported host is.

**2. `EXPLICIT_GITHUB_REPOSITORY` has no left-hand boundary, so a lookalike host matches on a suffix.** The pattern begins `(?:https?:\/\/)?(?:[\w.-]+@)?(?:www\.)?github\.com[/:]` with every leading group optional and no `\b` or `(?<![\w.-])`. On `notgithub.com/acme/widgets` the engine fails at offset 0, advances, and matches `github.com/acme/widgets` at offset 3 — returning `acme/widgets`. The comment above the constant reads "The host anchors it, so prose cannot match", which is a stronger claim than the regex makes: the host anchors it against *bare* prose, but not against a string that merely ends in `github.com`. This is still strictly better than the old pattern (which matched any `x/y`), so it is not a regression — but a `(?<![\w.-])` prefix would make the comment true, and it is worth a negative assertion next to the existing ones.

**3. The cue-word list re-admits the prose shape whenever a cue happens to precede it.** `in|on|for|of|from|at|to` are among the most common English words, so `the bug is in and/or handling` yields `and/or`, and `it regressed in login/logout` yields `login/logout` — both of which the new tests otherwise establish as things that must *not* be treated as repositories. The numeric guard only covers all-digit pairs, so it does not help here. The PR body flags the opposite trade-off (slug-first without a cue now asks), which is the deliberate one; this direction is the residual and reads as unintentional. A cheap narrowing that preserves every asserted positive: require the slug's owner and name to contain at least one character that is not a plain lowercase letter, or reject a slug whose halves are both dictionary-shaped — or simply add these two as known-residual test cases so the limit is documented rather than discovered.

## One thing I could not verify

The PR body states root `bun run verify` is running on the stack branch and the result will be posted as a comment. I did not run it; the per-package results quoted (`repo-input.test.ts` 14/14, `src/actions` 21/21, typecheck and lint clean) are the author's, and I am taking the pass/fail claims on trust while verifying the code and the assertions themselves at head. Nothing in the diff suggests those would fail.

None of the three findings blocks: each is a residual in a function that is unambiguously stricter and more correct than what it replaces, and #31116's actual defect — a valid-looking repository substituted for the requested one — is fixed and pinned.

hermesagent270-commits:

ss251 (approved):
9434c5840abcebeea43031a9274c458ea1c8703c

#31116's dangerous rewrite is gone. On develop, `normalizeRepositoryInput` took `segments[0]`/`segments[1]`, so the GitLab analyzers subgroup URL became `https://gitlab.com/gitlab-org/security-products.git` — a different repo that `assertSafeGitRemote` and `ALLOWED_DOMAINS` both accept. `splitRepositoryPath` (`repo-input.ts:52-69`) keeps every GitLab segment before a literal `-` element and slices GitHub/Bitbucket to two. Traced `extractRepositoryUrl` (`repo-input.ts:141-145`) → normalize → `tasks.ts:3649-3653`: output is the analyzers `.git` URL and the allowlist matches. `workspace-service.ts:624-637` clones that string.

The issue-action extractor is weaker. `CUED_GITHUB_SLUG` (`repo-input.ts:159-162`) includes `in|on|for`, so `the bug is in and/or handling` → `and/or` and `file a bug for login/logout redirect` → `login/logout`. `runManageIssues` (`tasks.ts:4362-4374`) then skips the clarification. `repo-input.test.ts:128-142` stays green because the listed `login/logout` sentence inserts `the` between cue and slug. `EXPLICIT_GITHUB_REPOSITORY` (`repo-input.ts:150-154`) has no left boundary: `see notgithub.com/acme/widgets` → `acme/widgets`.

None restore the wrong-existing-GitLab-clone. Subgroup round-trip is pinned (`repo-input.test.ts:38-59`). Tests are function-level and would not fail if the `tasks.ts` wiring were dropped.

Recommendation: merge

Verified by reading the diff at this head against develop `509e1c034c04a1cd5bcb213a15b675edf3a29afb`, fetching the four files named above at this SHA, and replaying the published regexes in a reviewer-authored node snippet.

Client / agent tooling: grok
— [grok-fleet-ss251]

Svector-anu (approved):
verified this against #31116, tracing the actual regex/segment logic rather than trusting the test names.

the bug is a real safety issue, not just cosmetic: `normalizeRepositoryInput` took `segments[0]`/`segments[1]` unconditionally, so a gitlab subgroup url like `https://gitlab.com/gitlab-org/security-products/analyzers` got silently rewritten to `https://gitlab.com/gitlab-org/security-products.git` — a different, valid-looking repo that passes `assertSafeGitRemote` and the allowlist, so `git clone` and the agent's edits land on the wrong remote entirely.

traced `splitRepositoryPath`: for gitlab it slices segments up to the first literal `-` path segment (not a substring match, so `gitlab-org` correctly isn't mistaken for the separator), keeping every group level; for github/bitbucket it stays at exactly 2 segments, matching their flat-namespace shape. confirmed against the test cases by hand: `gitlab-org/security-products/analyzers` (no `-` segment) keeps all 3; `group/sub/repo/-/issues/12` cuts at index 3 correctly; `elizaOS/eliza/pull/5` still truncates to 2 for github.

the free-text extractors are the more interesting half. `extractGitHubRepositorySlug`'s cued-slug regex requires the cue word to sit immediately before the slug (only `[:\s]+` between them), which is why "for the login/logout redirect" correctly does not match — "the" breaks adjacency — and why "it fails in 24/7 mode" needs the separate purely-numeric guard (`in` is adjacent to `24/7`, so it does match the shape, and the guard is what excludes it). walked through each negative test case against the regex by hand; all resolve the way the assertions expect, including the case where a cue word appears *after* the numeric pair ("happens 24/7 in prod") and so never reaches the numeric guard at all because adjacency already fails.

`extractRepositoryUrl`'s trailing-dot handling is deliberate, not accidental: `.` is inside `[\w.-]` so `https://github.com/acme/widgets.` is captured with the dot, then explicitly stripped by the trailing-punctuation replace — matches the test.

ci green, already 2 approvals, no defects found. approving.

lalalune:

Closing this mixed parser implementation while preserving the real GitLab subgroup defect in #31116. A Bun probe of the proposed extractor still returns login/logout for "file an issue for login/logout", and matches acme/widgets from "https://notgithub.com/acme/widgets" because the explicit-host regex has no left host boundary. The English cue-word list both retains the false-positive class and rejects valid slug-first input; it is not a reliable repository identity contract. Keep the lossless GitLab subgroup normalization in a focused replacement. For issue actions use validated explicit repository input or an unambiguous bound workspace repository, and clarify unresolved free text instead of adding more prose cue words. Exercise the actual action-to-repository selection, not only selected regex examples.

PR #31125, opened 2026-09-12
release: repair production Cloud secret handoff

NubsCarson:
Production Cloud release 34684027907 stopped before API deployment because the reusable workflow received empty Telegram credentials. The production Environment held valid values and its direct identity attestation had passed. This promotion replaces the two empty caller bindings with explicit secret expressions; the protected environments, credential allowlist, identity verification, migration gates, and atomic Worker publication remain intact.

Scope since main d56b4d27495d8bc0753f417210cd8d3461e7b8cc is three workflow/contract-test files. The app and runtime code are unchanged. Develop repair #31122 and staging promotion #31123 are merged; staging source is 914cf6cd4923f3392a8b75b22e1a8924fa2bfad3, tree 44efb0332cc8e4d185f495efc904d5ee84934ce3.

Validation: 31 focused checks, actionlint for both workflows, and full local `bun run verify` passed. Canonical staging release 34685072600 passed. Its immutable certificate was verified with the repository verifier against tree 44efb0332cc8e4d185f495efc904d5ee84934ce3, originating successful run, artifact digest, current workflow bytes, and expiry September 26 at 09:31 UTC. The real deployed browser trajectory passed in 150.179 seconds, with a settled Dedicated reply in 32.380 seconds and history preserved after reload and in a fresh browser session. PR #31122 also has all hosted checks green. Actual production secret resolution and the Dedicated UI walkthrough remain pending; staging preserves pre-existing Worker credentials and cannot prove the production-only Environment handoff.

Operational readiness: the separate production worker deployment 34684275500 passed, with the tested image and recovery/storage configuration verified on the host. One provider-attested Dedicated node has completed bootstrap and normal health checks, with one free slot. UI/model evidence is N/A for this workflow-only repair; live Dedicated UI acceptance remains a release requirement.

### DIFF
--- a/.github/workflows/cloud-cf-deploy.yml
+++ b/.github/workflows/cloud-cf-deploy.yml
@@ -1027,10 +1027,10 @@ jobs:
       cancel-in-progress: false
       queue: max
     uses: ./.github/workflows/cloud-cf-release.yml
-    # Telegram credentials have empty caller bindings so GitHub exposes their
-    # names to the reusable job's selected Environment without forwarding a
-    # repository or organization value. Omitting the names makes even existing
-    # Environment secrets resolve blank. The authority receipt stays here.
+    # Use explicit secret expressions for the reusable workflow. Empty literal
+    # bindings resolve blank in the called job even when its Environment has
+    # the secret. The called job's protected Environment supplies its values;
+    # identity attestation and the authority receipt remain required.
     secrets:
       ANTHROPIC_API_KEY: ${{ secrets.ANTHROPIC_API_KEY }}
       CARTESIA_API_KEY: ${{ secrets.CARTESIA_API_KEY }}
@@ -1049,8 +1049,8 @@ jobs:
       ELIZA_APP_BLOOIO_API_KEY: ${{ secrets.ELIZA_APP_BLOOIO_API_KEY }}
       ELIZA_APP_BLOOIO_PHONE_NUMBER: ${{ secrets.ELIZA_APP_BLOOIO_PHONE_NUMBER }}
       ELIZA_APP_BLOOIO_WEBHOOK_SECRET: ${{ secrets.ELIZA_APP_BLOOIO_WEBHOOK_SECRET }}
-      ELIZA_APP_TELEGRAM_BOT_TOKEN: ""
-      ELIZA_APP_TELEGRAM_WEBHOOK_SECRET: ""
+      ELIZA_APP_TELEGRAM_BOT_TOKEN: ${{ secrets.ELIZA_APP_TELEGRAM_BOT_TOKEN }}
+      ELIZA_APP_TELEGRAM_WEBHOOK_SECRET: ${{ secrets.ELIZA_APP_TELEGRAM_WEBHOOK_SECRET }}
       ELIZA_APP_WEBHOOK_GATEWAY_SECRET: ${{ secrets.ELIZA_APP_WEBHOOK_GATEWAY_SECRET }}
       ELIZA_MOBILE_APP_AUTH_APP_ID: ${{ secrets.ELIZA_MOBILE_APP_AUTH_APP_ID }}
       ELIZA_SERVICE_JWT_SECRET: ${{ secrets.ELIZA_SERVICE_JWT_SECRET }}
--- a/.github/workflows/cloud-cf-release.yml
+++ b/.github/workflows/cloud-cf-release.yml
@@ -52,11 +52,9 @@ on:
         required: false
         type: string
         default: ""
-    # Keep caller-scope secrets explicit. Telegram's live credentials are
-    # declared so the workflow can reference their fixed names, but the caller
-    # deliberately does not forward them. The entry workflow proves the
-    # environment-specific runtime authority; deploy-api resolves the values
-    # from its own protected Environment for atomic Worker publication.
+    # Keep caller-scope secrets explicit. The entry workflow proves the
+    # environment-specific runtime authority; deploy-api's protected
+    # Environment supplies the credentials for atomic Worker publication.
     secrets:
       ANTHROPIC_API_KEY:
         required: false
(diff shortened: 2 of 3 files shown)


attentionhead (approved):
**APPROVE** — verified at head `914cf6cd4923f3392a8b75b22e1a8924fa2bfad3` (base `main`, 3 files, +16/-16), reading the diff as fetched at that exact SHA.

## Provenance reconciles exactly

This is the third step of a `develop → staging → production` chain, and it carries no new work. The head is a merge commit (`914cf6cd Merge pull request #31123 from elizaOS/develop`) over a single payload commit `509e1c03 fix(ci): resolve protected secrets in Cloud release handoff (#31122)` — the same commit SHA that #31123 carried, so the promotion is byte-identical to what already merged, not a re-application.

The file set corroborates that from the diff rather than from the commit headline, which is the check that actually matters on a promotion:

| file | this PR | #31122 |
| --- | --- | --- |
| `.github/workflows/cloud-cf-deploy.yml` | +6/-6 | +6/-6 |
| `.github/workflows/cloud-cf-release.yml` | +5/-7 | +5/-7 |
| `packages/scripts/__tests__/homepage-deploy-workflow.test.ts` | +5/-3 | +5/-3 |

+16/-16 both sides, three files, no fourth path. Nothing has been added on the way to `main`.

## The change itself is right, and the reason is empirical

The two Telegram entries were bound as empty string literals in the caller:

```yaml
ELIZA_APP_TELEGRAM_BOT_TOKEN: ""
ELIZA_APP_TELEGRAM_WEBHOOK_SECRET: ""
```

The deleted comment records the theory behind that: an empty binding was believed to *expose the secret name* to the called job's selected Environment without forwarding a caller value, letting the called job resolve it from its own protected Environment. That theory is what this PR retires, and it is retired by production rather than by argument — the empty literal is what a `workflow_call` consumer actually receives, so `secrets.ELIZA_APP_TELEGRAM_BOT_TOKEN` in the called job resolved to the empty string and the Worker was published with blank Telegram credentials.

The strongest thing about the repair is that it is not a special case. Every other entry in that same `secrets:` block — `ANTHROPIC_API_KEY`, `ELIZA_APP_BLOOIO_WEBHOOK_SECRET`, `ELIZA_APP_WEBHOOK_GATEWAY_SECRET`, `ELIZA_SERVICE_JWT_SECRET` and the rest — already uses the explicit `${{ secrets.NAME }}` form and is presumably resolving correctly in the same deployments. So the fix makes the two broken bindings *identical in shape to ~20 working siblings in the same block*, which is a much better corroboration than reasoning about `workflow_call` scope in the abstract: the shape is already proven in production by its neighbours.

The surviving half of the `deploy-api` comment is the part worth keeping, and it is still accurate:

> A computed `secrets[]` key is evaluated without these Environment values in a reusable workflow and silently becomes blank.

That is the real hazard class here — *silently blank* — and it is why this bug shipped rather than failing loudly. Keeping it documented at the consumption site is correct.

## Test change

`homepage-deploy-workflow.test.ts` flips the assertion from `toBe("")` to `toBe(githubExpression("secrets." + name))` inside the existing `reusableEnvironmentSecrets` loop, and keeps the paired assertion that the name is declared under `on.workflow_call.secrets`. So the two halves of the contract — the caller binds it, and the called workflow declares it — stay pinned together; neither can be edited alone without a red test. Pinning the *previous* wrong value is what makes this a real regression guard rather than a loosened assertion.

## Non-blocking findings

**1. The rewritten `deploy-api` comment now asserts two different resolution paths at once.** It reads "These fixed names are resolved in this job's protected Environment through explicit caller secret expressions." Those are alternatives, not a sequence: an explicit caller expression forwards the *caller's* value into the called job, whereas "resolved in this job's protected Environment" describes the mechanism the old comment claimed and this PR is abandoning. The sentence as written would let a future reader reconstruct the empty-literal reasoning from it. Saying plainly that the caller forwards the value, and that the called job's Environment gates the *deployment* rather than supplying the credential, would close that door.

**2. Nothing in the test can observe the failure this PR repairs.** The suite parses workflow YAML and asserts on text, so it pins the binding's *shape* but is blind to whether the caller's scope actually holds a value for either name — which is exactly the axis on which the bug lived (correct-looking YAML, empty value at runtime). That is a reasonable limit for a static test and I am not asking for more, but it means the guard against a recurrence is the comment plus this assertion, not a detection. A post-deploy assertion that the Worker's Telegram binding is non-empty would be the thing that actually catches a blank; worth considering separately from this promotion.

**3. Worth a pointer to the incident.** The empty-literal form was deliberate and carried a written justification, so it is the kind of change that gets re-introduced by someone reading only the local context. A `(#31122)` reference in the surviving comment would make the retired theory traceable.

Neither finding touches the promotion itself: the payload is byte-identical to work already merged through `develop` and `staging`, the file set reconciles line-for-line, and the change replaces a demonstrably blank-producing binding with the form every sibling secret in the same block already uses.

ss251 (changes requested):
914cf6cd4923f3392a8b75b22e1a8924fa2bfad3

Finding: the new caller bindings do not read the reusable job's protected Environment. This does not repair blank production Telegram credentials, and it expands secret forwarding.

cloud-cf-deploy.yml:1052-1053 replaces empty caller bindings (`ELIZA_APP_TELEGRAM_BOT_TOKEN: ""`, `ELIZA_APP_TELEGRAM_WEBHOOK_SECRET: ""`) with `${{ secrets.ELIZA_APP_TELEGRAM_BOT_TOKEN }}` and `${{ secrets.ELIZA_APP_TELEGRAM_WEBHOOK_SECRET }}`. Those expressions evaluate in the caller job. This hunk does not add `environment:` there, so they resolve repository/organization secrets or empty, then pass that value into `workflow_call.secrets`, which shadows the Environment secret of the same name. cloud-cf-release.yml still declares the names as caller-scope secrets (comment-only at 52-56 and 1402-1408).

Input: Telegram tokens exist only on the called job's protected Environment (the prior design at cloud-cf-deploy.yml:1027 and cloud-cf-release.yml:52).
Wrong output: caller still forwards empty; Worker Telegram bindings stay blank.

Input: a repository or organization Telegram token exists.
Wrong output: that token is now forwarded into production. The empty-literal design existed to avoid forwarding a repository or organization value.

homepage-deploy-workflow.test.ts:537-541 only asserts the YAML equals `githubExpression("secrets.${name}")`. It stays green if production is still blank. That is not acceptance evidence.

Do not merge until the Telegram names are dropped from `workflow_call.secrets` and the caller `secrets:` map so the Environment can resolve them, or the same protected Environment is attached to the caller job so these expressions actually read it.

Verified by reading the three-file diff at this head; no suite applies beyond the YAML-shape assertion.

Recommendation: fix

— [free-xai-lane-ss251]

PR #31126, opened 2026-09-12
fix(plugin-wallet): report an unreachable EVM chain as unavailable instead of absent (#31111)

### LINKED ISSUE #31111: plugin-wallet (evm): an unreachable chain RPC is cached as chain absent and fed to transfer and swap prompts
**Describe the bug**

`getWalletBalanceForChain` in `plugins/plugin-wallet/src/chains/evm/providers/wallet.ts` (lines 242-247, develop `cbf357b0bd8a`) returns `null` from its catch. The aggregator in `getWalletBalances` (lines 224-234) drops every `null` from the map and then writes the truncated map to the cache:

```ts
for (const result of results) {
  if (result.status === "fulfilled" && result.value.balance !== null) {
    balances[result.value.chainName] = result.value.balance;
  } else if (result.status === "rejected") {
    logger.error(`Error getting balance:`, result.reason);
  }
}
await this._runtime.setCache(cacheKey, balances);
```

A chain whose RPC was unreachable for one call is therefore cached as "chain absent". `actions/transfer.ts` (line 74) and `actions/swap.ts` (line 771) call `getWalletBalances()` without `forceRefresh`, so they read the cached partial map and feed it into the transfer and swap intent prompts as `state.chainBalances`. The intent model is told the wallet holds nothing on that chain until the next forced refresh.

**To Reproduce**

Real provider with two configured chains; the `base` RPC answers 503 on the first call and is healthy on the second:

```
Error getting wallet balance for base: HTTP request failed: 503 Service Unavailable
call 1 (base RPC down): {"mainnet":"5"}
call 2 (base RPC healthy again): {"mainnet":"5"} <- served from cache

what the transfer action puts in the LLM prompt (state.chainBalances):
  mainnet: 5 ETH

base is absent, so the model is told the wallet holds nothing on base.
cache entry written by call 1: [["evm/wallet/walletBalances",{"mainnet":"5"}]]
```

**Expected behavior**

A failed chain is represented explicitly (for example `{ status: "unavailable", error }`) rather than omitted, the provider never caches a map that contains an unavailable entry, and the transfer and swap prompts render the chain as unavailable instead of silently missing.

- [ ] MP4 recording or JPG screenshots: N/A - no visible UI; the defect is in service and route output, reproduced with a script against the real modules.
- [x] Backend logs and script output are inline above.
- [ ] Real-LLM trajectory: N/A - the defect is reachable and observable without a model call.
- [x] Domain artifacts: cache rows and HTTP responses are shown inline above.

**Additional context**

Severity is bounded by the periodic `getWalletBalances(true)` in `chains/evm/service.ts`, but one flaky RPC call is enough to make an affordable transfer be refused or routed to the wrong chain until that refresh runs. The Solana sibling is reported separately.

hermesagent270-commits:
## Summary

`getWalletBalanceForChain` returned `null` from its catch, `getWalletBalances` dropped every `null` and cached the truncated map, and `buildTransferDetails` / `buildSwapDetails` read that cache without `forceRefresh`. One flaky RPC call therefore told the intent model that the wallet held nothing on that chain until the next periodic forced refresh.

Changes in `plugins/plugin-wallet/src/chains/evm`:

- `WalletProvider.getChainBalanceStates(forceRefresh)` (new) returns `Record<chain, { status: "ok", balance } | { status: "unavailable", error }>`. A snapshot with any unavailable chain is never written to the cache; a complete one is cached under the same key and shape as before.
- `getWalletBalances` keeps its return type and now derives from the states (ok chains only), so `EVMService` callers and the TEE wrapper are unchanged in shape. Its doc says unavailable chains are omitted and a partial result is never cached.
- `getWalletBalanceForChain` throws `EVMError(NETWORK_ERROR, message, cause)` instead of returning `null` (`error-policy:J2`); the TEE override signature follows.
- `transfer.ts` and `swap.ts` build `state.chainBalances` from the states and render `base: balance unavailable (RPC error)` for a failed chain. The swap relative-amount lookup reads the ok balance from the state (an unavailable chain still yields "unknown balance" for `half`/`max`/`percent`).
- `EVMWalletData` gains `unavailableChains`, filled by the service refresh, and the EVM wallet provider text lists those chains as unavailable instead of omitting them.

Closes #31111

## Evidence

Before (develop `cbf357b0bd8a`, real provider, `base` RPC 503 on the first call, healthy on the second):

```
call 1 (base RPC down): {"mainnet":"5"}
call 2 (base RPC healthy again): {"mainnet":"5"} <- served from cache
what the transfer action puts in the LLM prompt (state.chainBalances):
  mainnet: 5 ETH
cache entry written by call 1: [["evm/wallet/walletBalances",{"mainnet":"5"}]]
```

After: the new provider test pins `{ mainnet: ok 5, optimism: unavailable "HTTP request failed: 503 Service Unavailable" }`, no `setCache` call for the partial snapshot, and the next read seeing `optimism` once its RPC answers. The service test pins `unavailableChains: [{ chainName: "optimism", error }]` in the cached wallet data. The existing "caches only successful balances" test, which had codified partial caching, is replaced by a complete-snapshot case.

- `bunx vitest run src/chains/evm/__tests__/unit`: 52/52 (8 files).
- `bun run test` in `plugins/plugin-wallet`: 420 passed, 1 skipped (70 files).
- `bun run typecheck` and `bun run lint:check` in `plugins/plugin-wallet`: clean.

### DIFF
--- a/plugins/plugin-wallet/src/chains/evm/__tests__/unit/evm-service-cache.test.ts
+++ b/plugins/plugin-wallet/src/chains/evm/__tests__/unit/evm-service-cache.test.ts
@@ -47,16 +47,17 @@ describe("EVMService cache reads", () => {
   it("coalesces concurrent refreshes and writes one forced network snapshot", async () => {
     const runtime = new AgentRuntime({ logLevel: "fatal" });
     const setCache = vi.spyOn(runtime, "setCache").mockResolvedValue(true);
-    let releaseBalances: ((balances: { mainnet: string }) => void) | undefined;
-    const getWalletBalances = vi.fn(
+    type States = Record<string, { status: "ok"; balance: string }>;
+    let releaseBalances: ((balances: States) => void) | undefined;
+    const getChainBalanceStates = vi.fn(
       () =>
-        new Promise<{ mainnet: string }>((resolve) => {
+        new Promise<States>((resolve) => {
           releaseBalances = resolve;
         })
     );
     const wallet = {
       getAddress: vi.fn(() => "0x1234"),
-      getWalletBalances,
+      getChainBalanceStates,
       getChainConfigs: vi.fn(() => ({
         id: 1,
         name: "Ethereum",
@@ -70,10 +71,10 @@ describe("EVMService cache reads", () => {
     const second = service.refreshWalletData();
     await Promise.resolve();
     expect(initWalletProviderMock).toHaveBeenCalledOnce();
-    expect(getWalletBalances).toHaveBeenCalledOnce();
-    expect(getWalletBalances).toHaveBeenCalledWith(true);
+    expect(getChainBalanceStates).toHaveBeenCalledOnce();
+    expect(getChainBalanceStates).toHaveBeenCalledWith(true);
 
-    releaseBalances?.({ mainnet: "1.5" });
+    releaseBalances?.({ mainnet: { status: "ok", balance: "1.5" } });
     await Promise.all([first, second]);
 
     expect(setCache).toHaveBeenCalledOnce();
@@ -90,9 +91,39 @@ describe("EVMService cache reads", () => {
             name: "Ethereum",
           },
         ],
+        unavailableChains: [],
       })
     );
     expect(service.getWalletProvider()).toBe(wallet);
+  });
+
+  it("records an unreachable chain as unavailable instead of dropping it (#31111)", async () => {
+    const runtime = new AgentRuntime({ logLevel: "fatal" });
+    const setCache = vi.spyOn(runtime, "setCache").mockResolvedValue(true);
+    const wallet = {
+      getAddress: vi.fn(() => "0x1234"),
+      getChainBalanceStates: vi.fn(async () => ({
+        mainnet: { status: "ok", balance: "1.5" },
+        optimism: { status: "unavailable", error: "HTTP request failed: 503" },
+      })),
+      getChainConfigs: vi.fn(() => ({
+        id: 1,
+        name: "Ethereum",
+        nativeCurrency: { symbol: "ETH" },
+      })),
+    };
+    initWalletProviderMock.mockResolvedValue(wallet);
+    const service = new EVMService(runtime);
+
+    await service.refreshWalletData();
+
+    expect(setCache).toHaveBeenCalledWith(
+      EVM_WALLET_DATA_CACHE_KEY,
+      expect.objectContaining({
+        chains: [expect.objectContaining({ chainName: "mainnet", balance: "1.5" })],
+        unavailableChains: [{ chainName: "optimism", error: "HTTP request failed: 503" }],
+      })
+    );
   });
 
   it("throws an explicit uninitialized error before the first refresh", () => {
--- a/plugins/plugin-wallet/src/chains/evm/__tests__/unit/swap-amount-mode.test.ts
+++ b/plugins/plugin-wallet/src/chains/evm/__tests__/unit/swap-amount-mode.test.ts
@@ -20,17 +20,21 @@ vi.mock("../../../../utils/intent-trajectory", () => ({
 const WETH = "0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2";
 const USDC = "0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48";
 
-// base mainnet native balance, as a decimal string (the shape getWalletBalances returns)
+// base mainnet native balance, as a decimal string (the balance a chain's ok state carries)
 const BASE_BALANCE = "2"; // 2.0 ETH
 
 function createWalletProvider(
   balances: Record<string, string> = { base: BASE_BALANCE }
 ): WalletProvider {
+  const states = Object.fromEntries(
+    Object.entries(balances).map(([chain, balance]) => [chain, { status: "ok", balance }])
+  );
   return {
     chains: { base },
     getSupportedChains: () => ["base"],
     getChainConfigs: () => base,
     getWalletBalances: async () => balances,
+    getChainBalanceStates: async () => states,
   } as unknown as WalletProvider;
 }
 
--- a/plugins/plugin-wallet/src/chains/evm/__tests__/unit/wallet-provider-behavior.test.ts
+++ b/plugins/plugin-wallet/src/chains/evm/__tests__/unit/wallet-provider-behavior.test.ts
@@ -81,7 +81,7 @@ describe("WalletProvider local chain and cache behavior", () => {
     expect(setCache).not.toHaveBeenCalled();
   });
 
-  it("force-refreshes every chain in parallel and caches only successful balances", async () => {
+  it("force-refreshes every chain in parallel and caches a complete snapshot", async () => {
     const rt = runtime();
     const getCache = vi.spyOn(rt, "getCache");
     const setCache = vi.spyOn(rt, "setCache").mockResolvedValue(true);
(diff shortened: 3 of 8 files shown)


attentionhead (approved):
**APPROVE** — verified at head `c6c14c730a0ea09eabc937173d5188e94a3f8048` (base `develop`, 8 files, +214/-39), reading `providers/wallet.ts`, `service.ts`, `actions/swap.ts`, `actions/transfer.ts` and every other call site at that exact SHA rather than from a local checkout.

## The defect is a three-stage funnel and the fix closes it at the source

The bug in #31111 is not one lossy conversion but three in a row, and the diff is worth reading in that order:

1. `getWalletBalanceForChain` swallowed the RPC failure and returned `null`.
2. `getWalletBalances` folded `null` into *absent* — the old loop's `result.value.balance !== null` guard and the `result.status === "rejected"` branch both simply declined to write a key.
3. `await this._runtime.setCache(cacheKey, balances)` then ran **unconditionally**, persisting the truncated map, and `buildTransferDetails` / `buildSwapDetails` read that cache without `forceRefresh`.

So a single flaky call did not produce a transient wrong answer; it produced a *durable* one. And the value it produced is the most dangerous possible reading: absence from `state.chainBalances` is indistinguishable to the intent model from a zero balance on a chain the wallet actually funds.

The repair inverts the lossy step rather than patching downstream. `ChainBalanceState` as a discriminated union of `{status:"ok",balance}` / `{status:"unavailable",error}` means the failure can no longer be represented as absence — there is no `null` left to collapse. `getWalletBalances` is retained as a *derived* view (`states` → ok-only), which is why `EVMService`, the TEE wrapper and the UI callers keep their shapes and the diff stays at 8 files instead of rippling.

## Three things I checked at head rather than assuming

**The `getWalletBalanceForChain` signature change is contained.** Changing `Promise<string | null>` → `Promise<string>` + throw is the kind of change that breaks a distant caller doing `if (balance === null)`. I enumerated every occurrence of `getWalletBalanceForChain`, `getWalletBalances` and `getChainBalanceStates` across `plugins/` and `packages/` and fetched each non-PR file at this head. The only other hits are **namesakes on unrelated classes**: `SolanaService.getWalletBalances` (`chains/solana/service.ts:2200`, its own method) and the UI HTTP client `ElizaClient.getWalletBalances` (`packages/ui/src/api/client-wallet.ts:65,205`, consumed at `state/useWalletState.ts:471`, `ui/InventoryView.helpers.ts:134` and the e2e stub). None of them touches the EVM provider. `getWalletBalanceForChain` has no caller outside `providers/wallet.ts` at all. So the throw cannot escape to an unprepared handler.

**The TEE override is updated in the same file.** The PR body says "the TEE override signature follows" but no TEE path appears in the file list, which usually means the override was missed. It was not — `LazyTeeWalletProvider` lives in `providers/wallet.ts` itself, and the diff updates `getWalletBalanceForChain` to `Promise<string>` and adds a delegating `getChainBalanceStates` override alongside the existing `getWalletBalances` one. Had `getChainBalanceStates` been left un-overridden, the base implementation would have run against an uninitialized TEE wallet — that override is load-bearing, not boilerplate.

**Per-chain attribution is newly possible.** The rewrite switches from `result.value.chainName` to `chainNames[index]`, which is correct because `Promise.allSettled` preserves input order, and it is what lets the *rejected* branch name its chain at all. The old code's rejected branch logged a bare `Error getting balance:` with no chain — so before this PR the operator could not tell which chain was down from the logs either. `logger.error(\`Error getting balance for ${chainName}:\`, error)` fixes that as a side effect.

## Test quality

The behavior test is the strongest thing here. It drives a **two-phase outage-then-recovery sequence** off a mutable `optimismDown` flag: while down, `getChainBalanceStates()` reports `optimism` as `unavailable` with the real message, `getWalletBalances()` omits it, and `expect(setCache).not.toHaveBeenCalled()` — a *negative* assertion on the write, which is the one that actually pins the fix. Then the flag flips and the next read both sees `optimism` and caches the complete map. That second phase proves the absence of a poisoned cache, not merely the absence of a bad write; a fix that suppressed the write but left a stale entry would pass the first half and fail the second.

Equally important: the existing test named `caches only successful balances` had *codified the bug*. It is renamed and rewritten to `caches a complete snapshot` rather than left in place with a new test bolted on beside it. Deleting an assertion that encoded the wrong contract is the honest move and is easy to skip.

`wallet-provider-timeout.test.ts` flips `resolves.toBeNull()` → `rejects.toMatchObject({ name: "EVMError", code: "NETWORK_ERROR" })`, pinning the new contract at the leaf, and the service test pins `unavailableChains: [{ chainName: "optimism", error }]` in the cached `EVMWalletData` — so the union survives the whole path from RPC to cache rather than being asserted only where it is created.

## Non-blocking findings

**1. A partial cache entry written by the *previous* deploy survives this fix, because the cache key is unchanged.** The new invariant is "a snapshot with any unavailable chain is never written". That governs writes this code makes; it says nothing about entries already in the store. `cacheKey` is still `path.join("evm/wallet", "walletBalances")` and the old code wrote partial maps unconditionally. So on the first run after deploy, `getChainBalanceStates` can hit a *pre-existing truncated* entry, and the truncated chain comes back neither `ok` nor `unavailable` — it is simply **absent from the states map**, which is precisely the #31111 symptom, now routed through the branch that believes the cache is trustworthy (`for (const [chainName, balance] of Object.entries(cachedData)) states[...] = { status: "ok", balance }`). The fix self-heals only once that entry expires or a forced refresh lands. Bumping the key (`walletBalances` → `walletBalances/v2`) would make it effective immediately and costs one line. Same reasoning applies to `EVM_WALLET_DATA_CACHE_KEY = "evm_wallet_data"` in `constants.ts`.

**2. `unavailableChains` is required on `EVMWalletData` but optional where it is actually read.** `service.ts` declares `readonly unavailableChains: ReadonlyArray<{...}>` with no `?`, while the provider's inline cache-read type in `wallet.ts` declares `unavailableChains?:` and defensively applies `?? []`. The provider is right and the interface is optimistic: `EVM_WALLET_DATA_CACHE_KEY` is persisted, so an entry written before this deploy satisfies the old shape and has no such field, and any future consumer that reads the cache *as* `EVMWalletData` will be told by the type that `.unavailableChains` is always present when at runtime it can be `undefined`. Today the provider is the only reader so nothing breaks; making it optional in both places (or versioning the key per finding 1) keeps them from disagreeing.

**3. Open question — does the raw RPC error message reach a surface it should not?** `error` is `result.reason.message` verbatim, and it is persisted into the cache and returned in the provider's `data.unavailableChains`. The rendered *text* is correctly scrubbed to `balance unavailable (RPC error)`, which is the right call and clearly deliberate. My concern is the structured `data` field: viem's transport errors frequently embed the request URL, and a managed or Alchemy-style endpoint carries its API key in that URL. I have not verified viem's message format at this head, so I am raising this as a question rather than asserting it — but if the message can contain the endpoint URL, then a credential is being written to a runtime cache and handed to a provider consumer. Truncating to the error *name*/status in `unavailableChains.error` while keeping the full message in `logger.error` would settle it without losing any diagnostic value.

## One thing I did not verify

Root `bun run verify` is stated as running on the stack branch with the result to follow as a comment. I did not run it, and the per-package numbers quoted (`src/chains/evm/__tests__/unit` 52/52, `bun run test` 420 passed / 1 skipped, typecheck and lint clean) are the author's; I verified the code and the assertions at head, not the run. Nothing in the diff suggests those would fail.

None of the three findings blocks. The union removes the representation that made the bug expressible, the cache write is now conditional on completeness, all four presentation sites name an unreachable chain instead of omitting it, and the test that had encoded the old behaviour was removed rather than worked around.

hermesagent270-commits:

ss251 (changes requested):
c6c14c730a0ea09eabc937173d5188e94a3f8048

The #31111 failure is a poisoned `evm/wallet/walletBalances` map, and this head still serves that map as healthy.

Input: cache already holds `{"mainnet":"5"}` (the row in #31111) while the provider is configured for `mainnet` + `optimism`; `getChainBalanceStates()` with default `forceRefresh=false`.
Wrong output at `providers/wallet.ts:224-230`: `{ mainnet: { status: "ok", balance: "5" } }` — optimism is still absent, not `{ status: "unavailable" }`. `actions/transfer.ts:74` and `actions/swap.ts:776` read that path, so `state.chainBalances` still omits the chain.

`setCache` (`packages/core/src/runtime.ts:5562`) has no TTL, and the partial branch (`wallet.ts:261-265`) returns without `deleteCache`. A later `getChainBalanceStates(true)` that has any unavailable chain leaves the truncated entry in place.

The new test (`wallet-provider-behavior.test.ts:108-145`) stubs `getCache` to `undefined`, so it never inserts the poisoned map. `expect(setCache).not.toHaveBeenCalled()` pins new writes, not the read #31111 reproduces.

The union is the right shape (`EVMError(NETWORK_ERROR)`, named unavailable in transfer/swap/provider text, TEE override). That does not close #31111 while the cache key and a completeness check against `getSupportedChains()` are unchanged. Bump the key, or reject a cached map missing a configured chain and `deleteCache` on a partial live fetch.

Verified by reading the diff and head blobs of `wallet.ts`, `transfer.ts`, `swap.ts`, `service.ts`, and unit tests at this SHA; grepping `getWalletBalances` callers; reading `setCache` on develop.

Recommendation: fix

Client / agent tooling: grok
— [grok-fleet-ss251]

hermesagent270-commits:
Agreed, and fixed at 98f5991704df (follow-up commit). The head you reviewed only stopped new partial maps from being written; a map written before that rule, or an older complete snapshot left behind by a later partial refresh, was still served as complete on a non-forced read, and `setCache` has no TTL.

`getChainBalanceStates` now:

- accepts a cached map only when it contains every chain from `getSupportedChains()`; otherwise it logs, `deleteCache`s the entry and refreshes from RPC;
- `deleteCache`s the key whenever a live refresh has any unavailable chain, so no reader can fall back to pre-outage balances.

Two new cases in `wallet-provider-behavior.test.ts` drive exactly the #31111 input: `getCache` returns `{ mainnet: "5" }` with `mainnet` + `optimism` configured, a non-forced `getChainBalanceStates()` returns both chains from RPC and deletes the poisoned entry; and a forced refresh with `optimism` down deletes the older complete snapshot and writes nothing.

- `bunx vitest run src/chains/evm/__tests__/unit`: 54/54. `bun run test` in `plugins/plugin-wallet`: 422 passed, 1 skipped. `bun run typecheck` and `bun run lint:check`: clean.

Svector-anu (approved):
Reviewed against @ss251's CHANGES_REQUESTED, not just the diff shape.

Their gap was precise: commit `c6c14c7` fixed the live-fetch path (union type, named-unavailable text, TEE override) but left the read path trusting a pre-existing cache blindly — a map already poisoned by #31111 (e.g. `{mainnet: "5"}` while configured for `mainnet`+`optimism`) would still be served as complete on the next non-forced read, because there was no completeness check against `getSupportedChains()` and no `deleteCache` on a partial live fetch.

Commit `98f5991` closes exactly that gap:
- `getChainBalanceStates` now diffs cached keys against `getSupportedChains()`; a cache missing any configured chain is discarded (`deleteCache`) and treated as a miss, not served (`providers/wallet.ts:370-386`).
- A live fetch that turns up any `unavailable` chain also `deleteCache`s rather than leaving a stale complete snapshot in place, and never caches the partial result — only an all-`ok` refresh gets written back.
- The new test `"refuses a cached map that is missing a configured chain and refreshes from RPC (#31111)"` reproduces @ss251's exact repro (poisoned `{mainnet: "5"}` cache, `mainnet`+`optimism` configured, default `forceRefresh=false`) and asserts `deleteCache` fires and both chains resolve correctly. `"clears an older snapshot when a live refresh has an unavailable chain"` covers the complementary case.

Also checked the design choice to return a per-chain `ChainBalanceState` union (`ok`/`unavailable`) here rather than throw-per-call like the Solana sibling fix (#31127): correct call — `getChainBalanceStates` aggregates independently-failing per-chain RPCs via `Promise.allSettled`, so one bad chain shouldn't blank the healthy ones for display; only *caching* the incomplete set is refused. `getWalletBalanceForChain` itself now throws `EVMError(NETWORK_ERROR)` instead of returning `null` as a silent sentinel, consistent with the fast-fail error policy. `transfer.ts`, `swap.ts`, and the `evmWalletProvider` text all correctly surface "balance unavailable (RPC error)" instead of omitting the chain.

CI green, no remaining gap between the reported repro and the fix. Approving.

lalalune:

Retaining the explicit per-chain unavailable state and poisoned-cache rejection. This correctly fixes the missing-chain cache case and reaches the transfer/swap prompt builders as well as the wallet service. Add a consumer-level unavailable case through buildTransferDetails/buildSwapDetails or the wallet provider; the new tests stop at cache writes/aggregator states, and the swap harness supplies only ok states. Keep getWalletBalances documented as a partial compatibility result and audit its remaining consumers instead of treating Record<SupportedChain,string> as complete. Complete-cache freshness and concurrent provider reads remain separate from this missing-chain repair; do not claim all stale balance reads are prevented. Coordinate #31127 around explicit outage/freshness state without replacing either chain adapter or claiming live transactions.

PR #31127, opened 2026-09-12
fix(plugin-wallet): propagate Solana RPC failures instead of serving zero balances (#31110)

### LINKED ISSUE #31110: plugin-wallet (solana): RPC failures are cached and served as a zero portfolio, balance 0, and an unbounded 429 retry
**Describe the bug**

Two catch blocks in `plugins/plugin-wallet/src/chains/solana/service.ts` (develop `cbf357b0bd8a`) turn an RPC failure into a healthy-looking zero, and three consumers persist or serve that zero as fact.

1. `getTokenAccountsByKeypair` (lines 2104-2107) returns `[]` from its catch. `updateWalletData` (lines 1975-1986) treats `accounts.length === 0` as an empty portfolio and writes `{ totalUsd: "0", totalSol: "0", items: [] }` to `SOLANA_WALLET_DATA_CACHE_KEY`. The Solana wallet provider (`providers/wallet.ts`, lines 66-71) renders that cache into the planner prompt as `Total Value: $0.00 (0 SOL)` and `No tokens found with non-zero balance`, and `GET /wallet/balance/:token` (`routes/index.ts`, line 119) answers `404 TOKEN_NOT_FOUND` for a token the wallet holds.
2. `getBalancesByAddrs` (lines 2145-2154) returns `{}` from its catch; `GET /wallet/balance` (`routes/index.ts`, line 73) applies `balances[publicKeyStr] ?? 0` and answers `200` with `balance: 0`. The same catch retries on a `429` message with `return this.getBalancesByAddrs(...)` and no attempt counter, so a sustained rate limit is an unbounded recursion and the request never returns.

Both routes have their own `catch` that would have produced a correct 500; the service swallows the error first. This is the pattern the repository error policy forbids (a default return from a catch, and `?? <literal>` disguising missing required data).

**To Reproduce**

Real `SolanaService`, real routes and real provider; only the RPC `Connection` is replaced by a stub that answers HTTP 500 to every call:

```
--- service layer ---
getBalancesByAddrs  -> {}
 Error      Error fetching token accounts: Error: failed to get info about accounts: Server responded with 500 Internal Server Error
getTokenAccountsByKeypair -> []
--- HTTP routes (RPC is returning 500 for every call) ---
GET /wallet/balance -> HTTP 200 {"success":true,"data":{"publicKey":"11111111111111111111111111111112","balance":0,"symbol":"SOL"}}
GET /wallet/portfolio -> HTTP 200 {"success":true,"data":{"publicKey":"1111...1112","totalUsd":"0","totalSol":"0","tokens":[],"hasBirdeyeData":false}}
GET /wallet/balance/:token (USDC) -> HTTP 404 {"success":false,"error":{"code":"TOKEN_NOT_FOUND","message":"Token EPjFWdd5AufqSSqeM2qN1xzybapC8G4wEGGkZwyTDt1v not found in wallet"}}
--- portfolio cache written while RPC is down ---
updateWalletData() -> {"totalUsd":"0","totalSol":"0","items":[]}
cached under SOLANA_WALLET_DATA_CACHE_KEY: [["solana/walletData",{"totalUsd":"0","totalSol":"0","items":[]}]]
provider text injected into the planner prompt:
Eliza's Main Solana Wallet (11111111111111111111111111111112)
Total Value: $0.00 (0 SOL)
Token Balances:
No tokens found with non-zero balance
```

**Expected behavior**

An RPC failure propagates as a typed error (`ElizaError` with a code such as `SOLANA_RPC_UNAVAILABLE`). `updateWalletData` distinguishes "RPC reported zero accounts" from "RPC did not answer" and never caches the latter. The provider renders an explicit unavailable state instead of a zero portfolio. `GET /wallet/balance` returns the route's existing 500 path instead of `balance: 0`. The 429 retry is bounded by an attempt counter and ends in a typed `RATE_LIMITED` error.

- [ ] MP4 recording or JPG screenshots: N/A - no visible UI; the defect is in service and route output, reproduced with a script against the real modules.
- [x] Backend logs and script output are inline above.
- [ ] Real-LLM trajectory: N/A - the defect is reachable and observable without a model call.
- [x] Domain artifacts: cache rows and HTTP responses are shown inline above.

**Additional context**

Found in a develop-wide sweep for catch blocks that return a healthy-looking default. The EVM sibling is reported separately.

hermesagent270-commits:
## Summary

Two catch blocks in `plugins/plugin-wallet/src/chains/solana/service.ts` turned an RPC failure into a healthy-looking zero that three consumers persisted or served as fact.

- `getTokenAccountsByKeypair` now throws `ElizaError("SOLANA_RPC_UNAVAILABLE")` with the RPC error as `cause` instead of returning `[]`. `updateWalletData` therefore rejects during an outage and never writes the `{ totalUsd: "0", items: [] }` portfolio to `SOLANA_WALLET_DATA_CACHE_KEY`; the wallet provider, which reads only the cache, keeps returning its existing "not ready" empty result rather than `Total Value: $0.00 / No tokens found`. A wallet the RPC reports as genuinely empty is still cached as an empty portfolio.
- `getBalancesByAddrs` no longer returns `{}` from a catch. The 429 branch, which recursed with no bound, is replaced by `readAccountsWithRateLimitRetry`: up to three attempts with linear backoff, then `ElizaError("SOLANA_RPC_RATE_LIMITED")`; any other failure is `ElizaError("SOLANA_RPC_UNAVAILABLE")`.
- `GET /wallet/balance` drops `balances[publicKeyStr] ?? 0`; a missing entry is treated as a contract violation and the handler's existing catch returns the 500 it was written for. `GET /wallet/balance/:token` now reaches its catch (500) instead of answering `404 TOKEN_NOT_FOUND` for a held token.

Closes #31110

## Evidence

Before (develop `cbf357b0bd8a`, real service and routes, RPC stub answering 500):

```
getBalancesByAddrs  -> {}
getTokenAccountsByKeypair -> []
GET /wallet/balance -> HTTP 200 {"success":true,"data":{...,"balance":0,"symbol":"SOL"}}
GET /wallet/balance/:token (USDC) -> HTTP 404 {"success":false,"error":{"code":"TOKEN_NOT_FOUND",...}}
updateWalletData() -> {"totalUsd":"0","totalSol":"0","items":[]}
cached under SOLANA_WALLET_DATA_CACHE_KEY: [["solana/walletData",{"totalUsd":"0","totalSol":"0","items":[]}]]
provider text: Total Value: $0.00 (0 SOL) / No tokens found with non-zero balance
```

After: `service.rpc-errors.test.ts` uses the real `SolanaService` prototype and the real route handlers with only the connection stubbed, and pins the typed rejections, the bounded 429 retry (exactly `BALANCE_READ_MAX_ATTEMPTS` calls, then `SOLANA_RPC_RATE_LIMITED`), recovery when the limit clears within the budget, no cache write during an outage, the genuine-empty cache write, and both routes answering 500.

- `bunx vitest run src/chains/solana/service.rpc-errors.test.ts`: 7/7.
- `bun run test` in `plugins/plugin-wallet`: 425 passed, 1 skipped (71 files).
- `bun run typecheck` and `bun run lint:check` in `plugins/plugin-wallet`: clean.

The EVM sibling is #31126.

### DIFF
--- a/plugins/plugin-wallet/src/chains/solana/routes/index.ts
+++ b/plugins/plugin-wallet/src/chains/solana/routes/index.ts
@@ -70,7 +70,12 @@ const getWalletBalanceHandler: LegacyRouteHandler = async (_req, res, runtime) =
 
     const publicKeyStr = publicKey.toBase58();
     const balances = await solanaService.getBalancesByAddrs([publicKeyStr]);
-    const balance = balances[publicKeyStr] ?? 0;
+    const balance = balances[publicKeyStr];
+    if (balance === undefined) {
+      // The service reports every requested address or throws; a missing
+      // entry is a contract violation, not a zero balance.
+      throw new Error(`Balance read returned no entry for ${publicKeyStr}`);
+    }
 
     sendSuccess(res, {
       publicKey: publicKeyStr,
--- a/plugins/plugin-wallet/src/chains/solana/service.rpc-errors.test.ts
+++ b/plugins/plugin-wallet/src/chains/solana/service.rpc-errors.test.ts
@@ -1,0 +1,218 @@
+/**
+ * RPC-failure contract for SolanaService balance and token-account reads, the
+ * wallet routes built on them, and the portfolio cache. Uses the real service
+ * prototype and real route handlers with only the RPC connection stubbed, so
+ * the assertions cover the code path a down or rate-limited RPC actually takes
+ * (#31110). Deterministic; no network.
+ */
+import { PublicKey } from "@solana/web3.js";
+import { describe, expect, it, vi } from "vitest";
+import { SOLANA_WALLET_DATA_CACHE_KEY } from "./constants";
+import { solanaRoutes } from "./routes/index";
+import { SolanaService } from "./service";
+
+const ADDR = "11111111111111111111111111111112";
+// Public USDC mint on mainnet, used only as a held-token lookup key.
+const USDC_MINT = "EPjFWdd5AufqSSqeM2qN1xzybapC8G4wEGGkZwyTDt1v";
+const RPC_ERROR = new Error(
+  "failed to get info about accounts: Server responded with 500 Internal Server Error"
+);
+const RATE_LIMIT_ERROR = new Error("429 Too Many Requests");
+
+type Stub = {
+  getMultipleAccountsInfo: () => Promise<unknown[]>;
+  getParsedTokenAccountsByOwner: () => Promise<{ value: unknown[] }>;
+};
+
+function serviceWithConnection(connection: Stub) {
+  const cache = new Map<string, unknown>();
+  const svc = Object.create(SolanaService.prototype) as SolanaService & {
+    connection: Stub;
+    decimalsCache: Map<string, number>;
+    runtime: unknown;
+    lastUpdate: number;
+    UPDATE_INTERVAL: number;
+    ensurePublicKey: () => Promise<PublicKey>;
+  };
+  svc.connection = connection;
+  svc.decimalsCache = new Map();
+  svc.lastUpdate = 0;
+  svc.UPDATE_INTERVAL = 0;
+  svc.runtime = {
+    logger: {
+      warn: vi.fn(),
+      error: vi.fn(),
+      log: vi.fn(),
+      info: vi.fn(),
+      debug: vi.fn(),
+    },
+    getCache: async (key: string) => cache.get(key),
+    setCache: async (key: string, value: unknown) => {
+      cache.set(key, value);
+    },
+    getSetting: () => undefined,
+  };
+  svc.getPublicKey = async () => new PublicKey(ADDR);
+  svc.ensurePublicKey = async () => new PublicKey(ADDR);
+  return { svc, cache };
+}
+
+function mkRes() {
+  const res = { statusCode: 200, payload: undefined as unknown } as {
+    statusCode: number;
+    payload: unknown;
+    status: (code: number) => typeof res;
+    json: (body: unknown) => typeof res;
+    send: (body: unknown) => typeof res;
+  };
+  res.status = (code) => {
+    res.statusCode = code;
+    return res;
+  };
+  res.json = (body) => {
+    res.payload = body;
+    return res;
+  };
+  res.send = res.json;
+  return res;
+}
+
+const routeByPath = Object.fromEntries(solanaRoutes.map((route) => [route.path, route.handler]));
+
+describe("SolanaService RPC failures", () => {
+  it("throws a typed error from getTokenAccountsByKeypair instead of returning []", async () => {
+    const { svc } = serviceWithConnection({
+      getMultipleAccountsInfo: async () => {
+        throw RPC_ERROR;
+      },
+      getParsedTokenAccountsByOwner: async () => {
+        throw RPC_ERROR;
+      },
+    });
+    await expect(svc.getTokenAccountsByKeypair(new PublicKey(ADDR), {})).rejects.toMatchObject({
+      code: "SOLANA_RPC_UNAVAILABLE",
+      cause: RPC_ERROR,
+    });
+  });
+
+  it("throws a typed error from getBalancesByAddrs instead of returning {}", async () => {
+    const { svc } = serviceWithConnection({
+      getMultipleAccountsInfo: async () => {
+        throw RPC_ERROR;
+      },
+      getParsedTokenAccountsByOwner: async () => ({ value: [] }),
+    });
+    await expect(svc.getBalancesByAddrs([ADDR])).rejects.toMatchObject({
+      code: "SOLANA_RPC_UNAVAILABLE",
+    });
+  });
+
+  it("bounds the 429 retry and surfaces a rate-limit error", async () => {
+    vi.useFakeTimers();
+    try {
+      let calls = 0;
+      const { svc } = serviceWithConnection({
+        getMultipleAccountsInfo: async () => {
+          calls += 1;
+          throw RATE_LIMIT_ERROR;
+        },
+        getParsedTokenAccountsByOwner: async () => ({ value: [] }),
+      });
+      const pending = svc.getBalancesByAddrs([ADDR]);
+      const settled = expect(pending).rejects.toMatchObject({
+        code: "SOLANA_RPC_RATE_LIMITED",
+      });
+      await vi.runAllTimersAsync();
+      await settled;
+      expect(calls).toBe(SolanaService.BALANCE_READ_MAX_ATTEMPTS);
+    } finally {
+      vi.useRealTimers();
+    }
+  });
+
+  it("recovers when the rate limit clears within the retry budget", async () => {
+    vi.useFakeTimers();
+    try {
+      let calls = 0;
+      const { svc } = serviceWithConnection({
+        getMultipleAccountsInfo: async () => {
+          calls += 1;
+          if (calls === 1) throw RATE_LIMIT_ERROR;
+          return [{ lamports: 2_000_000_000 }];
+        },
+        getParsedTokenAccountsByOwner: async () => ({ value: [] }),
+      });
+      const pending = svc.getBalancesByAddrs([ADDR]);
+      await vi.runAllTimersAsync();
+      await expect(pending).resolves.toEqual({ [ADDR]: 2 });
+    } finally {
+      vi.useRealTimers();
+    }
+  });
+
+  it("does not cache an empty portfolio while the RPC is down", async () => {
+    const { svc, cache } = serviceWithConnection({
+      getMultipleAccountsInfo: async () => {
+        throw RPC_ERROR;
+      },
+      getParsedTokenAccountsByOwner: async () => {
+        throw RPC_ERROR;
+      },
+    });
+    await expect(svc.updateWalletData(true)).rejects.toMatchObject({
+      code: "SOLANA_RPC_UNAVAILABLE",
+    });
+    expect(cache.has(SOLANA_WALLET_DATA_CACHE_KEY)).toBe(false);
+  });
+
+  it("still caches a genuinely empty portfolio when the RPC reports no accounts", async () => {
+    const { svc, cache } = serviceWithConnection({
+      getMultipleAccountsInfo: async () => [],
+      getParsedTokenAccountsByOwner: async () => ({ value: [] }),
+    });
+    await expect(svc.updateWalletData(true)).resolves.toMatchObject({
+      totalUsd: "0",
+      items: [],
+    });
+    expect(cache.get(SOLANA_WALLET_DATA_CACHE_KEY)).toMatchObject({
+      totalUsd: "0",
+      items: [],
+    });
+  });
+
+  it("answers the balance and token routes with 500, not 200/0 or 404, while the RPC is down", async () => {
+    const { svc } = serviceWithConnection({
+      getMultipleAccountsInfo: async () => {
+        throw RPC_ERROR;
+      },
+      getParsedTokenAccountsByOwner: async () => {
+        throw RPC_ERROR;
+      },
+    });
+    const runtime = {
+      getService: () => svc,
+      logger: { warn: vi.fn(), error: vi.fn() },
+      getCache: async () => undefined,
+      setCache: async () => undefined,
+    } as never;
+
+    const balance = mkRes();
+    await routeByPath["/wallet/balance"]({ params: {} } as never, balance as never, runtime);
+    expect(balance.statusCode).toBe(500);
+    expect(balance.payload).toMatchObject({
+      success: false,
+      error: { code: "ERROR" },
+    });
+
+    const token = mkRes();
+    await routeByPath["/wallet/balance/:token"](
+      { params: { token: USDC_MINT } } as never,
+      token as never,
+      runtime
+    );
+    expect(token.statusCode).toBe(500);
+    expect(token.payload).not.toMatchObject({
+      error: { code: "TOKEN_NOT_FOUND" },
+    });
+  });
+});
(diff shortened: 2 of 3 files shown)


attentionhead (approved):
**APPROVE** — verified at head `7135055af92bcca570fcede0b56646097d928a37` (base `develop`, 3 files, +268/-26), reading `chains/solana/service.ts` and `chains/solana/routes/index.ts` as fetched at that exact SHA and tracing every caller of the two changed methods, rather than from a local checkout.

## The defect and why the fix lands where it does

Both catch blocks converted a transport failure into a *successful-looking* empty result, and the damage came from what happened next rather than from the empty value itself:

- `getTokenAccountsByKeypair` returned `[]`. `updateWalletData`'s RPC fallback tests `if (!accounts || accounts.length === 0)`, builds `{ totalUsd: "0", totalSol: "0", items: [] }`, **writes it to `SOLANA_WALLET_DATA_CACHE_KEY`** and sets `lastUpdate = now`. The wallet provider reads only that cache, so a momentary outage was persisted as a durable "wallet is empty" and survived until the interval elapsed.
- `getBalancesByAddrs` returned `{}` from its catch, and `GET /wallet/balance` then applied `balances[publicKeyStr] ?? 0` and answered **HTTP 200 with `balance: 0`** — an affirmative statement of fact about someone's funds, produced by a failed read.

The repair is placed at the two conversions rather than at the consumers, which is the right call: there are six internal call sites and two routes, and patching each would have left the next one to be written exposed.

I checked the claim that the fix actually reaches the cache write, because that is the whole thesis and it is not visible in the diff. `getTokenAccounts()` (`:1889`) is `return this.getTokenAccountsByKeypair(publicKey)` with no catch, so the new throw propagates; and `updateWalletData`'s outer handler at `:2018` is

```ts
} catch (error) {
  logger.error(`Error updating wallet data: ${error}`);
  throw error;
}
```

a log-and-rethrow, not a swallow. So the `setCache` at `:2016` is genuinely unreachable during an outage. Had that catch returned a default portfolio instead, the PR's central claim would have been false while its test still passed — worth confirming rather than assuming.

## Blast radius of the two signature changes — checked, not assumed

Turning a method that returned `[]`/`{}` into one that throws is the change most likely to break a distant caller. I enumerated every occurrence of `getBalancesByAddrs` and `getTokenAccountsByKeypair` across `plugins/` and `packages/`: they appear **only** in `chains/solana/service.ts` and `chains/solana/routes/index.ts`. Nothing outside the plugin consumes them. Within that surface:

- `executeSwap` (`:2480`) reads balances inside a `try` opened at `:2469` **inside** the `for (const wallet of wallets)` loop, whose catch at `:2653` sets `swapResponses[pubKey] = { success: false }`. So an RPC throw degrades one wallet rather than aborting the batch — this was the one place a throw could have turned a partial failure into a total one, and it does not.
- `getTokenAccountsByKeypairs` (`:2117`) fans out over `Promise.all`, so one failing address now rejects the whole map. Given the thesis (never report a read you did not perform), that is consistent.
- Route `/wallet/balance/:token` needs no edit: `getTokenAccountsByKeypair` throwing means control reaches the existing catch at `:136` instead of falling through to `sendError(res, 404, "TOKEN_NOT_FOUND", …)` at `:124`. Answering 404 "you do not hold this token" because the RPC was unreachable is the same lie in a different shape, and the test asserts against that specific wrong answer rather than merely asserting a 500.

`ElizaError` is already imported at `:21`, so no import hunk is missing.

## The retry replaces something genuinely dangerous

The old 429 branch was `await sleep(1000); return this.getBalancesByAddrs(walletAddressArr)` — **unbounded recursion with no attempt counter**, so a sustained rate limit produced an infinite retry loop that never returned and never errored, hammering the endpoint that was already shedding load. `readAccountsWithRateLimitRetry` bounds it at `BALANCE_READ_MAX_ATTEMPTS` with linear backoff and then throws a typed error. That is the most consequential change in the diff and the PR body undersells it.

## Test quality

`service.rpc-errors.test.ts` is built on the real implementation, not a model of it: `Object.create(SolanaService.prototype)` with only `connection` stubbed, and the real route handlers pulled through `Object.fromEntries(solanaRoutes.map(...))`. So the assertions cover the path production takes.

Two controls stand out:

- **Outage versus genuinely-empty discrimination on the same cache key.** `does not cache an empty portfolio while the RPC is down` asserts `cache.has(SOLANA_WALLET_DATA_CACHE_KEY) === false`, and `still caches a genuinely empty portfolio when the RPC reports no accounts` asserts the write *does* happen with `{totalUsd:"0", items:[]}`. Either alone is weak — a fix that simply stopped caching empty portfolios would pass the negative and fail the positive. The pair is what pins the actual distinction the bug was about.
- **The bound is asserted against the constant, not a literal.** `expect(calls).toBe(SolanaService.BALANCE_READ_MAX_ATTEMPTS)` cannot drift if the constant is retuned, while still failing if the loop stops honouring it. Paired with a recovery case where the limit clears on attempt 2 and the call resolves `{ [ADDR]: 2 }`, so the retry is shown to be bounded *and* still able to succeed. `vi.useFakeTimers()` + `runAllTimersAsync()` keeps it deterministic with no real sleep.

## Non-blocking findings

**1. The retry is one layer too high, and re-reads work that already succeeded.** `readAccountsWithRateLimitRetry` wraps `batchGetMultipleAccountsInfo`, which chunks the address list into several RPC calls. A 429 on a late chunk therefore discards the chunks that already answered and re-issues the entire batch, up to three times — sending strictly more traffic to an endpoint that is rate-limiting precisely because it wants less. For a single-address read (both routes, `getBalance`, `executeSwap`) this is free, so it is latent rather than active; it bites on the multi-address paths. Retrying the individual chunk inside `batchGetMultipleAccountsInfo` would preserve the bound while removing the amplification. Related: with linear backoff the worst case adds `1000 + 2000 = 3s` synchronously inside an HTTP handler before the 500 is returned.

**2. The new error taxonomy does not survive the HTTP boundary.** `SOLANA_RPC_UNAVAILABLE` and `SOLANA_RPC_RATE_LIMITED` are carefully distinguished at the service layer and then flattened by `sendError(res, 500, "ERROR", "Failed to get wallet balance", errorMessage)` — the test itself pins `error: { code: "ERROR" }`. So a caller cannot tell a rate limit from an outage from any other server fault, and the distinction the PR just built is invisible where it would be most actionable (a client can back off on a rate limit; it cannot on a generic 500). Mapping these to 503 with the typed code, or surfacing `error.code`, would let the taxonomy do work outside the process.

**3. Two LLM-facing helpers now reject wholesale, and this is not mentioned or tested.** `walletAddressToHumanString` (`:2197`) and `walletAddressToLLMString` (`:2219`) both do `Promise.all([getBalancesByAddrs, getTokenAccountsByKeypair])` with no try/catch, so either failure now rejects the whole string instead of degrading. Their `solBal ?? "0"` fallbacks consequently only cover a genuinely missing key and no longer cover a failed read. Under this PR's thesis that is arguably the *right* behaviour — a prompt string reading `balance: 0` during an outage is exactly the #31110 symptom — but it is an uncaught behaviour change at a prompt-construction site that the summary does not list, and whoever calls these now needs a handler. Worth stating explicitly, one way or the other.

**4. `getBalance` now has three different failure conventions.** It returns the sentinel `-1` when no address resolves (`:558`), `?? 0` for an address the RPC omitted (a genuine zero, correct), and now throws for an RPC failure. A caller written against `if (bal === -1)` will not catch the throw. Not introduced by this PR, but this PR is what makes the method's contract mixed; a line in its doc comment would help.

**5. Scope worth naming: the fix governs the RPC-fallback path.** When `BIRDEYE_API_KEY` is set and Birdeye answers, `updateWalletData` returns at `:1967` without touching the Solana RPC, and the inner catch at `:1969` still swallows a Birdeye failure and falls through. That fall-through is the correct design and now lands in a path that fails loudly — but it does mean "#31110 is fixed" holds for the RPC read specifically, not for every route by which an empty portfolio can be produced.

## One thing I did not verify

Root `bun run verify` is stated as running on the stack branch with the result to follow as a comment. I did not run it; the quoted numbers (`service.rpc-errors.test.ts` 7/7, `bun run test` 425 passed / 1 skipped across 71 files, typecheck and lint clean) are the author's. I verified the code and the assertions at head, not the run. Nothing in the diff suggests those would fail.

None of the five findings blocks. The unbounded 429 recursion is gone, the empty-portfolio cache write is genuinely unreachable during an outage, the 200/`balance: 0` and 404/`TOKEN_NOT_FOUND` lies are both replaced by honest failures, the one batch call site degrades per-wallet rather than aborting, and the outage-versus-genuinely-empty distinction is pinned from both sides.

hermesagent270-commits:
Head moved to c4a711d619f0 (amended, patch-identical apart from one test line): the "Source static smoke" failure was gitleaks' generic-api-key rule matching `token: "EPjF…"` in the new test, the public USDC mint next to a `token` key. The mint now lives in a `USDC_MINT` constant so the scanner has nothing to flag; no source change.

hermesagent270-commits:
Root `bun run verify` (RUN_TURBO_CONCURRENCY=4) on `stack/verify-0912` (0f6a057d2a0d: develop cbf357b0bd8a with all eight of today's fix branches merged, no conflicts): **373/373 tasks successful**. Log: /root/eliza-fixes/verify-0912.log on the reporting host. The stack includes this PR's head before the test-only amend at c4a711d619f0 (one constant hoisted; no source change).

ss251 (approved):
I'll start from the published reviewer skill and the full offloaded brief, then review this PR from the supplied diff only.c4a711d619f099857d375d10e1e9100f3db4ad08

The pre-image bug is in the diff: `getBalancesByAddrs` caught RPC failures and `return {}`, and `getTokenAccountsByKeypair` `return []`. `getWalletBalanceHandler` then did `balances[publicKeyStr] ?? 0` (`plugins/plugin-wallet/src/chains/solana/routes/index.ts`). Input: Solana RPC 500 or 429. Wrong output: HTTP 200 with balance 0, or `updateWalletData` caching an empty portfolio while the node was only down (#31110).

This head removes both swallows. `readAccountsWithRateLimitRetry` in `service.ts` throws `ElizaError` (`SOLANA_RPC_UNAVAILABLE` / `SOLANA_RPC_RATE_LIMITED`) after at most `BALANCE_READ_MAX_ATTEMPTS` (3) 429 retries; a missing on-chain account still maps to `0`. `getTokenAccountsByKeypair` rethrows instead of `[]`. The balance route throws on a missing map entry instead of coercing to 0.

`service.rpc-errors.test.ts` would fail if those swallows returned: `/wallet/balance` and `/wallet/balance/:token` must be 500, not 200/0 or `TOKEN_NOT_FOUND`; `updateWalletData(true)` must reject and skip `SOLANA_WALLET_DATA_CACHE_KEY` when RPC throws, and still cache `{ totalUsd: "0", items: [] }` when the RPC returns empty accounts; 429 retries exactly 3 times then `SOLANA_RPC_RATE_LIMITED`, and recovers to `{ [ADDR]: 2 }` on the second call.

Token-account 429s still become `SOLANA_RPC_UNAVAILABLE` with no retry. That does not restore the zero-balance lie. No lockfile, CI, binary, or test-weakening.

Verified by reading the three-file diff at this head. Live execution blocked: no disposable sandbox on this host, so this is a static review of the diff at the exact head.

Recommendation: merge

— [free-xai-lane-ss251]

hermesagent270-commits:
Head moved to 9689a8ba9190 (amended again, test file only): the "Lint affected workspace closure" failure at c4a711d619f0 was plugin-wallet's Biome formatter on the new test, which I had formatted with the repository-root config instead of the package config (line width differs). Reformatted with the package config; `biome check .` in plugins/plugin-wallet is clean. No source change.

Svector-anu (approved):
Reviewed the actual fix against the linked bug, not just the diff shape.

**Bug confirmed via #31110**: `getTokenAccountsByKeypair` swallowed RPC failures into `[]`, which `updateWalletData` then cached as a legitimate empty portfolio (`SOLANA_WALLET_DATA_CACHE_KEY`) — so a transient RPC outage became a *persistently cached* "wallet has 0 tokens" until the next natural refresh. `getBalancesByAddrs` had the same shape (`{}` → `?? 0` at the route) plus an actual unbounded-recursion bug on 429 (no attempt counter, so a sustained rate limit never returns).

**What this PR fixes, verified against the tests:**
- Both paths now throw `ElizaError` (`SOLANA_RPC_UNAVAILABLE` / `SOLANA_RPC_RATE_LIMITED`) instead of returning a fabricated empty/zero value — matches the repo's documented fast-fail error policy (`packages/core/src/errors.ts`), and `"ephemeral"` is the correct severity per that file's own doc comment (transient/retryable, not fatal).
- The 429 retry is now bounded (`BALANCE_READ_MAX_ATTEMPTS = 3`, linear backoff) — closes the unbounded-recursion bug outright, not just the caching bug.
- `service.rpc-errors.test.ts` directly exercises the case that mattered most: `updateWalletData` rejects and does **not** write to `SOLANA_WALLET_DATA_CACHE_KEY` while the RPC is down, but still caches a genuine empty portfolio when the RPC legitimately reports zero accounts — the fix correctly distinguishes "RPC is down" from "wallet is actually empty."
- Route-level test confirms `/wallet/balance` and `/wallet/balance/:token` now answer 500 instead of a false 200/0 or 404 while degraded.

No unbounded-retry regression, no silent catch-and-default left in either changed path. CI is green.

Approving.

lalalune:

Retaining the real Solana RPC failure propagation and bounded retry, with the route regressions that distinguish outages from zero/not-found. Complete the provider/cache state: the current provider returns empty text when no cache exists and serves an existing portfolio after a failed refresh without an explicit unavailable/stale status. Preventing a zero overwrite is useful but is not caller-visible outage reporting. Add a previously populated cache -> failed refresh -> provider read -> successful recovery case and preserve the last known balance with its freshness/error state. Validate account-result cardinality before interpreting positions as balances; the genuine-empty test supplies an empty multi-account response, whereas an absent requested account must be represented explicitly. Keep the new route contract failure typed with ElizaError. Coordinate the common unavailable-state contract with #31126 without conflating separate chain adapters or claiming live RPC acceptance.

Issue #30712, opened 2026-09-05
Remove verified unused repository files and backup artifacts


lalalune:
## Problem

The repository tracks Blender `.blend1` backup copies alongside the maintained `.blend` models. These backups add 7,440,754 bytes to fresh checkouts without a referenced build or runtime consumer. Existing Knip configuration also omits the root workspace and uses broad source entry globs, so its unused-file report needs independent verification before cleanup.

## Scope and acceptance

- Run Knip, Madge, cyclomatic-complexity and duplicate-content analysis across the repository and retain a detailed report with coverage limitations.
- Remove verified unused files, dependencies, and redundant generated artifacts only after checking manifests, exports, dynamic loaders, scripts, tests, and asset consumers.
- Preserve active model files, published API exports, fixtures, native deliverables, and required guide pairs.
- Record exact deletions and validation results, including pre-existing failures.

## Evidence plan

Static audit outputs, reference searches, before/after file inventories, focused package validation where source changes are made, and the root verify gate. UI and live-model evidence are N/A for backup-file and unused-dependency removal without runtime behavior changes; any later behavior change requires its owning evidence lane.

PR #31129, opened 2026-09-12
chore(plugin-browser): remove the inert workspace and capture hook registries (#31117)

### LINKED ISSUE #31117: plugin-browser: browser workspace and capture hook registries have no writer or reader
**Describe the bug**

`plugins/plugin-browser/src/browser-workspace-hooks.ts` (line 53) and `plugins/plugin-browser/src/browser-capture-hooks.ts` (line 23) export global registries (`registerBrowserWorkspaceHooks` / `getBrowserWorkspaceHooks`, `registerBrowserCaptureHooks` / `getBrowserCaptureHooks`) that nothing on develop (`cbf357b0bd8a`) writes or reads. The package guide (`plugins/plugin-browser/CLAUDE.md` and `AGENTS.md`, lines 71-72) documents them as the registration seam ("BrowserWorkspaceHooks interface + global registration helpers"), and the wildcard `./*` export in `package.json` makes them importable, so a host that wires itself through `registerBrowserWorkspaceHooks` gets silence: every real consumer imports the concrete functions from `src/workspace/browser-workspace.ts` and `src/workspace/browser-capture.ts` directly (`packages/agent/src/api/server.ts`, `plugins/plugin-browser/src/routes/workspace.ts`, `packages/ui/src/api/client-browser-workspace.ts`).

**To Reproduce**

```
$ git grep -n "registerBrowserCaptureHooks\|getBrowserCaptureHooks\|registerBrowserWorkspaceHooks\|getBrowserWorkspaceHooks" origin/develop -- '*.ts' '*.tsx'
origin/develop:plugins/plugin-browser/src/browser-capture-hooks.ts:23:export function registerBrowserCaptureHooks(hooks: BrowserCaptureHooks): void {
origin/develop:plugins/plugin-browser/src/browser-capture-hooks.ts:27:export function getBrowserCaptureHooks(): BrowserCaptureHooks | null {
origin/develop:plugins/plugin-browser/src/browser-workspace-hooks.ts:53:export function registerBrowserWorkspaceHooks(
origin/develop:plugins/plugin-browser/src/browser-workspace-hooks.ts:59:export function getBrowserWorkspaceHooks(): BrowserWorkspaceHooks | null {
```

Loading the plugin entrypoint and then reading the registries:

```
after loading both plugin entrypoints:
  getBrowserCaptureHooks()    = null
  getBrowserWorkspaceHooks()  = null
```

**Expected behavior**

Either the registries are removed together with the two guide lines (nothing imports them), or, if the indirection is wanted for the desktop/web split, `browser-workspace.ts` and `browser-capture.ts` consult `getBrowserWorkspaceHooks()` / `getBrowserCaptureHooks()` first and the desktop host registers there.

- [ ] MP4 recording or JPG screenshots: N/A - dead registration seam, no UI.
- [x] Grep and runtime output are inline above.
- [ ] Domain artifacts: N/A.

**Additional context**

Found in a sweep for registration helpers with no production writer or reader. The Apple Foundation adapter registry in plugin-local-inference has the same shape and is reported separately.

hermesagent270-commits:
## Summary

`plugins/plugin-browser/src/browser-workspace-hooks.ts` and `browser-capture-hooks.ts` exported `globalThis`-backed registries (`register*Hooks` / `get*Hooks`) that nothing on develop wrote or read; every real consumer imports the concrete functions from `src/workspace/browser-workspace.ts` and `src/workspace/browser-capture.ts` directly. The package guide listed the modules as the registration seam and the wildcard `./*` export made them importable, so a host wiring itself through `registerBrowserWorkspaceHooks` got silence.

This removes both modules, the two guide lines (CLAUDE.md and AGENTS.md kept byte-identical), and the "mirroring `browser-capture-hooks.ts`" phrase in the domain-policy header. No in-tree code changes behaviour.

Closes #31117

## Evidence

```
$ git grep -n "registerBrowserCaptureHooks\|getBrowserCaptureHooks\|registerBrowserWorkspaceHooks\|getBrowserWorkspaceHooks" origin/develop -- '*.ts' '*.tsx'
(only the two declaring files)
```

- `bun run check:agents-claude`: 157 pairs byte-identical.
- `bun run typecheck`, `bun run lint:check`, `bun run test` in `plugins/plugin-browser`: clean, 327/327 (37 files).

If the indirection is wanted for a desktop/web split, the alternative is to have the workspace and capture modules consult the registries first and have the desktop host register there; nothing in the repository does that today, so deletion is the change that matches the code.


attentionhead (approved):
Reviewed at head `04ad60365961e637626062b70c4c6d917762adf9` (5 files, +3/-98), base `develop`. I fetched the PR head into a local ref and ran the reachability greps against that tree rather than against a working copy, because a stale checkout is exactly how a deletion review goes wrong.

## The inertness claim holds, and by a wider margin than the PR claims

A name grep is not sufficient evidence for these two modules, because both registries key off `Symbol.for(...)` — the *global* symbol registry. Any code anywhere, in or out of tree, can reach the slot with `globalThis[Symbol.for("elizaos.browser-workspace.hooks")]` without importing the module, without naming `registerBrowserWorkspaceHooks`, and without appearing in the author's grep. That is the escape hatch that makes "nothing references it" a weaker statement than it looks.

I checked for it explicitly. At the PR head, across the whole repository:

- the two symbol keys `elizaos.browser-capture.hooks` and `elizaos.browser-workspace.hooks` — **no hits**
- the interface names `BrowserCaptureHooks` and `BrowserWorkspaceHooks` — **no hits**
- the four functions `register*Hooks` / `get*Hooks` — **no hits**
- the bare module basenames `browser-workspace-hooks` / `browser-capture-hooks`, which would catch a dynamic `import()`, a build-config entry, or a doc reference — **no hits**

The last two greps were run with no pathspec at all, so they covered every file in the tree rather than the `'*.ts' '*.tsx'` the PR body limited itself to. The deletion is clean, including against the indirection route a symbol-keyed registry specifically enables.

`src/index.ts` also never re-exported either module — it exports `./browser-domain-policy.js` and the workspace barrel, but neither hooks file — so the package's main entry point never surfaced them. They were reachable only through the `./*` wildcard subpath in `package.json`.

## The strongest argument for this change is one the body only half makes

The PR says a host wiring itself through `registerBrowserWorkspaceHooks` "got silence". That is the point, and it is worth stating as the justification rather than as background. Because the `./*` wildcard export made `@elizaos/plugin-browser/browser-workspace-hooks` a resolvable public subpath, an out-of-tree desktop host could import it, call `registerBrowserWorkspaceHooks(…)` with a complete and correct implementation, and get a successful build, a successful call, and no behaviour — the workspace modules never read the slot. Deleting the file converts that silent no-op into an immediate module-resolution error at the import site. A loud failure at build time strictly dominates a silent one at runtime, and for a seam that gates browser command execution on a desktop host, silence is the dangerous outcome.

Worth noting the version is `2.0.3-beta.7`, so removing a subpath from the public export surface is cheap here in a way it would not be post-1.0.

## The retained registry is the right control

`browser-domain-policy.ts` uses the *same* `globalThis`-symbol pattern and is correctly untouched — and the difference is verifiable, not stylistic. It has a real consumer surface (`registerBrowserDomainPolicy` is called at ~24 sites in `src/__tests__/browser-domain-policy.test.ts` plus its declaration) and it *is* re-exported from `src/index.ts`. The two deleted modules had neither a consumer nor a barrel export. So this PR removes two instances of a pattern while keeping a third, and the discriminator is consumption rather than a judgement about the pattern itself. That is the correct line to draw, and the header edit that drops the now-dangling "mirroring `browser-capture-hooks.ts`" phrase keeps the surviving comment honest; the rest of that file's `+3/-4` is just reflow.

## Docs

`AGENTS.md` and `CLAUDE.md` are provably still byte-identical rather than merely asserted: both files move between the same pair of blob hashes in this diff (`27ff7385` → `d2cbe993`), which is only possible if their contents match exactly before and after. That satisfies the `check:agents-claude` pairing without needing to trust the reported run.

## Findings (non-blocking, none changes the verdict)

**1. The deleted interfaces were the written contract for the desktop/web split.** `BrowserWorkspaceHooks` enumerated eight methods with exact signatures; `BrowserCaptureHooks` three plus `frameFile`. The concrete functions in `src/workspace/browser-workspace.ts` and `src/workspace/browser-capture.ts` remain, so the shape is recoverable by reading them, but the explicit specification of "what a host must implement" is gone. The PR body already names the alternative — have the workspace modules consult the registries first — and I agree deletion matches the code as it stands today. If issue #31117 or its successor ever revives that split, this diff is the spec; a one-line pointer to this PR in the issue would save rediscovering it.

**2. Removing a `./*` subpath is a public-API change that the changelog surface may not capture.** Nothing in-tree breaks, and I cannot see consumers outside this repository. It is worth a moment's confirmation that no sibling product (a desktop host in another repo) imports `@elizaos/plugin-browser/browser-workspace-hooks`, since by construction such a consumer would be invisible to every grep above. Given the seam demonstrably did nothing, such a consumer is already broken in the silent way described — so this is a question about *how* they find out, not whether they were working.

## Verdict

APPROVE. The removal is genuinely dead code, verified at head against the symbol-registry route rather than just the import route; the retained sibling registry is distinguished by actual consumption; and the docs pairing is provable from the diff itself.

Verification note: all claims above come from greps against the PR head tree and from the diff. I did not run the test suite, the typecheck, or `check:agents-claude`; the results quoted in the PR body are the author's.

hermesagent270-commits:

ss251 (changes requested):
04ad60365961e637626062b70c4c6d917762adf9

This diff deletes two `globalThis` Symbol registries and syncs docs. It does not change a shipped browser command path.

Finding 1 — “inert” is claimed, not shown. `plugins/plugin-browser/src/browser-capture-hooks.ts:23-29` exported `registerBrowserCaptureHooks` / `getBrowserCaptureHooks` on `Symbol.for("elizaos.browser-capture.hooks")`. `plugins/plugin-browser/src/browser-workspace-hooks.ts:53-61` exported the matching workspace register/get pair on `Symbol.for("elizaos.browser-workspace.hooks")`. Input: those APIs have no live callers. Output from this patch: the files are gone, AGENTS.md/CLAUDE.md drop the two listings, and `browser-domain-policy.ts:4-9` only deletes the “mirroring `browser-capture-hooks.ts`” sentence. There is no grep, no remaining-import audit, and no test that would fail if `getBrowserCaptureHooks()` or `getBrowserWorkspaceHooks()` were still imported. `workspace.ts` is listed as workspace re-exports and is not touched, which is consistent with no re-export there, not proof for the rest of the plugin or out-of-tree callers.

Finding 2 — no material product change. If the modules are truly unreferenced, this is ~90 lines of dead-code cleanup plus a comment edit. No reproduced failure, no acceptance path, no test that encodes “these hooks must stay gone.” Trivial cleanup and comment-only churn are reject, not merge.

`browser-domain-policy.ts` still describes its own `globalThis` symbol registry after this change. Capture/workspace execution is untouched.

Adversarial pass at this head: no lockfile, CI, binaries, submodules, test weakening, or permission expansion.

Verified by reading the diff at this head. Live execution blocked: no disposable sandbox on this host, so this is a static review of the diff at the exact head.

Recommendation: close

— [free-xai-lane-ss251]

hermesagent270-commits:
Thanks for the read. On the two findings:

**Finding 1 (inertness is claimed, not shown).** The audit is a plain grep over develop for both function names and both global symbols, in every source, script and doc file:

```
$ git grep -n "registerBrowserCaptureHooks\|getBrowserCaptureHooks\|registerBrowserWorkspaceHooks\|getBrowserWorkspaceHooks\|elizaos.browser-capture.hooks\|elizaos.browser-workspace.hooks" origin/develop -- '*.ts' '*.tsx' '*.mjs' '*.js' '*.md'
plugins/plugin-browser/src/browser-capture-hooks.ts:13:const BROWSER_CAPTURE_HOOKS = Symbol.for("elizaos.browser-capture.hooks");
plugins/plugin-browser/src/browser-capture-hooks.ts:23:export function registerBrowserCaptureHooks(hooks: BrowserCaptureHooks): void {
plugins/plugin-browser/src/browser-capture-hooks.ts:27:export function getBrowserCaptureHooks(): BrowserCaptureHooks | null {
plugins/plugin-browser/src/browser-workspace-hooks.ts:43:const BROWSER_WORKSPACE_HOOKS = Symbol.for("elizaos.browser-workspace.hooks");
plugins/plugin-browser/src/browser-workspace-hooks.ts:53:export function registerBrowserWorkspaceHooks(
plugins/plugin-browser/src/browser-workspace-hooks.ts:59:export function getBrowserWorkspaceHooks(): BrowserWorkspaceHooks | null {
```

Every hit is the two declaring files plus the two guide lines this PR removes. No importer, no re-export (`workspace.ts` and `src/index.ts` never touched them), and no test. Since the getters were never called, no code path could observe a registration; that is what "inert" means here. Out-of-tree callers would have been calling a registry nothing reads, so they too got silence, which is the defect #31117 describes.

The negative control is the package gate: with both files gone, `bun run typecheck`, `bun run lint:check` and `bun run test` (327/327) pass, and `bun run check:agents-claude` keeps the guides byte-identical. A test asserting that a deleted module stays deleted would only pin the deletion itself; the typecheck already fails if anything imports the removed names.

**Finding 2 (no product change).** Agreed that no shipped browser command path changes. The user-facing defect is documentation: the package guide presented these modules as the host registration seam, so a host built against that guide wired itself into a dead end. Removing the modules and the guide lines is the fix that matches the code; wiring the concrete workspace and capture functions through the registries instead would add indirection nothing needs today. If the maintainers prefer to keep the seam and make the workspace/capture modules consult it, I am happy to do that in this PR instead.

lalalune:

Consolidated into the existing verified-unused-code cleanup #30712. The reported browser hook registries have no production writer/reader in current repository search, and working consumers use the concrete workspace/capture modules. This is a small dead-code/documentation cleanup rather than a separate demonstrated browser failure. Preserve export/compatibility checks and guide parity in the canonical cleanup; no new tests should merely assert that deleted files stay absent.

PR #31130, opened 2026-09-12
feat(plugin-local-inference): wire the Apple Foundation Models fast path on iOS (#31118)

### LINKED ISSUE #31118: plugin-local-inference: Apple Foundation adapter registry is never populated or consulted, so the documented fast path never activates
**Describe the bug**

`registerAppleFoundationAdapter` / `getAppleFoundationAdapter` in `plugins/plugin-local-inference/src/backends/apple-foundation.ts` (lines 112 and 118, develop `cbf357b0bd8a`) form a registry that no production code writes or reads, so the Apple Foundation Models on-device fast path documented in the module header and in `plugins/plugin-local-inference/docs/per-platform-backend-strategy.md` (line 39) never activates.

On iOS 26 with Apple Intelligence enabled the native side exists (`ComputerUseBridge.swift`, `foundationModelGenerate`, line 478) and the probe capability `foundationModel` is defined (`plugins/plugin-computeruse/src/mobile/ios-bridge.ts`, lines 324 and 351), but nothing calls `createAppleFoundationAdapter` (line 52) or registers the result, and the inference read path never queries the registry. Every short prompt keeps going to llama.cpp. `plugins/plugin-local-inference/__tests__/apple-foundation.test.ts` (lines 211-221) asserts the adapter is "discoverable from the WS9 cross-review entry point" and its comment says the runtime side "wires up `registerAppleFoundationAdapter` only when the iOS bridge probe reports `foundationModel:true`"; that runtime side does not exist.

**To Reproduce**

```
$ git grep -n "AppleFoundationAdapter" origin/develop -- '*.ts' '*.tsx' '*.swift' '*.md' | grep -v "plugin-local-inference/__tests__\|plugin-local-inference/src/backends/apple-foundation.ts"
(no output: the declaring module and its own test are the only references)
```

After importing `plugin-local-inference/src/index.ts`:

```
getAppleFoundationAdapter() = null
```

**Expected behavior**

In the local-inference boot for the Capacitor/iOS target (`src/runtime/boot.ts` or `ensure-local-inference-handler.ts`), build the adapter with the real ComputerUse bridge and register it when `probe()` reports `foundationModel: true`, and have the dispatcher consult `getAppleFoundationAdapter()` before the llama path. Otherwise remove the registry and the documentation and test claims that describe it as wired.

- [ ] MP4 recording or JPG screenshots: N/A - no UI; the fast path is absent rather than visibly broken.
- [x] Grep and runtime output are inline above.
- [ ] Domain artifacts: N/A - an on-device iOS 26 run would show every prompt routed to llama.cpp; not available on the reporting host.

**Additional context**

Same shape as the plugin-browser hook registries, reported separately.

hermesagent270-commits:
## Summary

`registerAppleFoundationAdapter` / `getAppleFoundationAdapter` in `plugins/plugin-local-inference/src/backends/apple-foundation.ts` had no production writer or reader, so the iOS 26 Apple Foundation Models fast path the module header, `docs/per-platform-backend-strategy.md` and the adapter test all describe never activated. This wires it, keeping llama.cpp as the owned backend.

New `src/runtime/apple-foundation-fast-path.ts`:

- `tryRegisterAppleFoundationAdapter({ platform, getBridge })` runs at mobile boot (`registerLocalInferenceBoot`, before the local handler installs). It registers the adapter only when `ELIZA_PLATFORM=ios`, the bridge handle exists, and `bridge.probe()` reports `capabilities.foundationModel === true`; every other outcome is returned as a typed skip reason and logged at debug.
- `resolveIosComputerUseBridge()` reads `Capacitor.Plugins.ComputerUse` locally (mirroring computeruse's `getIosBridge`). A runtime import of `@elizaos/plugin-computeruse/mobile/ios-bridge` is not possible: that package publishes `dist/mobile/` as declarations only, and the plugin's vitest alias maps the package root to `src/index.ts`. The type import stays.
- `resolveAppleFoundationFastPath(modelType, params)` is consulted at the top of the local text handler (`makeHandler` now receives the model type). It returns the adapter only for `ModelType.TEXT_SMALL` and `TEXT_COMPLETION` with a non-empty prompt of at most 4000 characters, no streaming (`stream`, `streamStructured`, `onStreamChunk`), no structured decoding (`grammar`, `responseSkeleton`, `spanSamplerPlan`, `prefill`), and no non-text `responseFormat`. `ACTION_PLANNER` and `RESPONSE_HANDLER` share the TEXT_SMALL slot but never take it. `ELIZA_APPLE_FOUNDATION_FAST_PATH=0|false|off` disables the route.
- `generateWithAppleFoundation` maps `prompt`, `maxTokens` and `temperature` onto `foundationModelGenerate`, returns the text, and throws on a bridge failure (no empty-string success).

The adapter test's "runtime side wires this up" comment and the backend-strategy doc now describe the real wiring.

Closes #31118

## Evidence

- `apple-foundation-fast-path.test.ts` (new, 9 cases): bridge resolution from a fake `Capacitor` global; registration on iOS with `foundationModel:true`; skips for non-iOS, missing bridge, `probe.ok:false`, a throwing probe, and `foundationModel:false` (registry stays null); eligibility matrix for model types, prompt length, streaming and structured-decoding params; env kill switch; not-yet-available adapter declined; generate mapping and failure propagation.
- `ensure-local-inference-handler.test.ts` (new case): with an adapter registered, the registered TEXT_SMALL handler returns the adapter's text and never calls the engine; the ACTION_PLANNER handler and a streaming TEXT_SMALL call still reach `localInferenceEngine.generate`.
- `boot.test.ts` (2 new cases): on a mobile platform with `ELIZA_PLATFORM=ios` the registration runs with `platform: "ios"` before the handler install; on desktop it is never attempted.
- `bunx vitest run` on the four touched suites: 53/53. `bun run test` in `plugins/plugin-local-inference`: 3005 passed, 2 skipped (285 files). `bun run typecheck` and `bun run lint:check`: clean.
- Root `bun run verify`: running on this branch; result will be posted as a comment.

**Not exercised: a real iOS 26 device.** The bridge contract (`probe`, `foundationModelGenerate`) is the existing `IosComputerUseBridge` interface implemented in `ComputerUseBridge.swift`; the tests drive it through a fake. On a device, the boot log states whether the fast path registered and, per call, `[local-inference] apple-foundation served N->M tokens in Xms`.

### DIFF
--- a/packages/app-core/platforms/ios/App/App/ComputerUseBridge.swift
+++ b/packages/app-core/platforms/ios/App/App/ComputerUseBridge.swift
@@ -25,6 +25,9 @@ import os
 import ReplayKit
 import UIKit
 import Vision
+#if canImport(FoundationModels)
+import FoundationModels
+#endif
 
 // MARK: - Bridge plugin
 
@@ -475,31 +478,73 @@ public class ComputerUseBridge: CAPPlugin, CAPBridgedPlugin {
 
     // ── 6. Apple Foundation Models ───────────────────────────────────────────
 
+    // Generates one completion with the system language model. Compiled only
+    // when the SDK can import FoundationModels (Xcode 26+) and executed only
+    // on iOS 26+; every other build or device resolves
+    // `foundation_model_unavailable`. The framework reports no token
+    // accounting, so the result carries `text` and `elapsedMs` only (the TS
+    // contract leaves `tokensIn` / `tokensOut` optional).
     @objc public func foundationModelGenerate(_ call: CAPPluginCall) {
-        guard let prompt = call.getString("prompt") else {
+        guard let prompt = call.getString("prompt"), !prompt.isEmpty else {
             return resolveError(call, code: "internal_error", message: "prompt required")
         }
         guard isFoundationModelAvailable() else {
             return resolveError(call, code: "foundation_model_unavailable",
                                 message: "Apple Foundation Models requires iOS 26+ with Apple Intelligence enabled.")
         }
-        // Real implementation will load the system LanguageModel via the
-        // FoundationModels framework. We surface a clear unavailable error
-        // until the on-device target is validated. The shape below matches
-        // the TS contract so the JS side can already integrate.
-        let _ = prompt
-        let _ = call.getObject("options")
+        #if canImport(FoundationModels)
+        if #available(iOS 26.0, *) {
+            let options = call.getObject("options") ?? [:]
+            let temperature = options["temperature"] as? Double
+            let maxTokens = options["maxTokens"] as? Int
+            let instruction = options["instruction"] as? String
+            let started = Date()
+            Task { [weak self] in
+                guard let self else { return }
+                do {
+                    let session: LanguageModelSession
+                    if let instruction, !instruction.isEmpty {
+                        session = LanguageModelSession(instructions: instruction)
+                    } else {
+                        session = LanguageModelSession()
+                    }
+                    var generation = GenerationOptions()
+                    if let temperature { generation.temperature = temperature }
+                    if let maxTokens { generation.maximumResponseTokens = maxTokens }
+                    let response = try await session.respond(to: prompt, options: generation)
+                    let elapsedMs = Int(Date().timeIntervalSince(started) * 1000)
+                    call.resolve([
+                        "ok": true,
+                        "data": [
+                            "text": response.content,
+                            "elapsedMs": elapsedMs,
+                        ] as [String: Any],
+                    ])
+                } catch {
+                    // Guardrail refusals, context-window overflow and unsupported
+                    // locales all surface here; the JS side treats any of them as
+                    // "the OS model did not answer" and leaves llama.cpp in charge.
+                    Self.log.error("foundationModelGenerate failed: \(error.localizedDescription, privacy: .public)")
+                    self.resolveError(call, code: "foundation_model_error", message: error.localizedDescription)
+                }
+            }
+            return
+        }
+        #endif
         resolveError(call, code: "foundation_model_unavailable",
-                     message: "Foundation Models adapter is unavailable pending on-device validation.")
+                     message: "This build was compiled without the FoundationModels framework.")
     }
 
     private func isFoundationModelAvailable() -> Bool {
+        #if canImport(FoundationModels)
         if #available(iOS 26.0, *) {
-            // The actual API check lives in `FoundationModels.LanguageModel.isAvailable`.
-            // Keep the runtime probe behind that gate; default to false until
-            // the framework is linked.
-            return false
+            // `.unavailable` covers Apple Intelligence off, unsupported device or
+            // region, and the model still downloading.
+            if case .available = SystemLanguageModel.default.availability {
+                return true
+            }
         }
+        #endif
         return false
     }
 
--- a/plugins/plugin-computeruse/docs/IOS_CONSTRAINTS.md
+++ b/plugins/plugin-computeruse/docs/IOS_CONSTRAINTS.md
@@ -101,6 +101,13 @@ Apple Intelligence is enabled. We expose this as an *opportunistic*
 fast-path. If unavailable, the existing llama-cpp-capacitor local Eliza-1
 vision path stays as the default.
 
+The bridge probes `SystemLanguageModel.default.availability` and generates
+through `LanguageModelSession.respond(to:options:)`, compiled only when the
+SDK can import `FoundationModels` (Xcode 26+) and executed only on iOS 26+;
+older SDKs and devices report `foundation_model_unavailable`. The framework
+exposes no token accounting, so `FoundationModelResult.tokensIn/tokensOut`
+are omitted. Not yet validated on a device (see the checklist below).
+
 Entitlement and Info.plist updates are listed below.
 
 ## What does NOT work
--- a/plugins/plugin-computeruse/src/mobile/ios-bridge.ts
+++ b/plugins/plugin-computeruse/src/mobile/ios-bridge.ts
@@ -52,6 +52,7 @@ export type IosBridgeErrorCode =
   | "intent_invocation_failed"
   | "vision_no_text" // OCR ran but found nothing
   | "foundation_model_unavailable" // Apple Intelligence disabled / OS too old
+  | "foundation_model_error" // the system model refused or failed the request (guardrail, context window, locale)
   | "internal_error";
 
 // ── 1. ReplayKit foreground (own-app) capture ────────────────────────────────
@@ -256,8 +257,13 @@ export interface FoundationModelOptions {
 
 export interface FoundationModelResult {
   readonly text: string;
-  readonly tokensIn: number;
-  readonly tokensOut: number;
+  /**
+   * Token accounting is absent when the framework does not report it. The
+   * FoundationModels API exposes no token counts, so the Swift bridge omits
+   * both fields rather than estimating them.
+   */
+  readonly tokensIn?: number;
+  readonly tokensOut?: number;
   readonly elapsedMs: number;
 }
 
--- a/plugins/plugin-local-inference/__tests__/apple-foundation.test.ts
+++ b/plugins/plugin-local-inference/__tests__/apple-foundation.test.ts
@@ -214,10 +214,10 @@ describe("apple-foundation adapter", () => {
     const found = getAppleFoundationAdapter();
     expect(found).toBe(adapter);
     expect(found!.name).toBe("apple-foundation");
-    // The runtime side that wires up `registerAppleFoundationAdapter` does
-    // so only when the iOS bridge probe reports `foundationModel:true`. The
-    // adapter exposes a structural `LocalInferenceRuntimeService`-compatible
-    // surface — `generate` is the relevant entry point.
+    // `tryRegisterAppleFoundationAdapter` (src/runtime/apple-foundation-fast-path.ts)
+    // registers here at iOS boot only when the bridge probe reports
+    // `foundationModel:true`; the local text handler then consults
+    // `getAppleFoundationAdapter()` per call. `generate` is the entry point.
     expect(typeof found!.generate).toBe("function");
   });
 });
--- a/plugins/plugin-local-inference/docs/per-platform-backend-strategy.md
+++ b/plugins/plugin-local-inference/docs/per-platform-backend-strategy.md
@@ -39,7 +39,11 @@ efficiency." There is no separate per-platform runtime to add.
 - **Apple Foundation Models** (`src/backends/apple-foundation.ts`): an
   **opportunistic** iOS-26 text adapter, never the owned backend (native/AGENTS.md
   §11). Out-of-process OS model services cannot satisfy the single-runtime
-  contract.
+  contract. `src/runtime/apple-foundation-fast-path.ts` registers it at iOS boot
+  when the `ComputerUse` bridge probe reports `foundationModel:true`, and the
+  local text handler routes only plain short `TEXT_SMALL` / `TEXT_COMPLETION`
+  calls to it (never planner, structured, or streaming calls);
+  `ELIZA_APPLE_FOUNDATION_FAST_PATH=0` disables it.
 
 ### 2a. Every voice/vision model is already on its optimal backend — MEASURED, no ANE win to capture
 
--- a/plugins/plugin-local-inference/src/backends/apple-foundation.ts
+++ b/plugins/plugin-local-inference/src/backends/apple-foundation.ts
@@ -51,8 +51,15 @@ export interface AppleFoundationAdapter {
  */
 export function createAppleFoundationAdapter(
 	getBridge: () => IosComputerUseBridge | null,
+	options: {
+		/**
+		 * Result of a probe the caller has already awaited. Seeds `available()`
+		 * so the first call is deterministic and the device is not probed twice.
+		 */
+		readonly knownAvailable?: boolean;
+	} = {},
 ): AppleFoundationAdapter {
-	let probedAvailable: boolean | null = null;
+	let probedAvailable: boolean | null = options.knownAvailable ?? null;
 	let probing: Promise<boolean> | null = null;
 
 	async function probe(): Promise<boolean> {
--- a/plugins/plugin-local-inference/src/runtime/apple-foundation-fast-path.test.ts
+++ b/plugins/plugin-local-inference/src/runtime/apple-foundation-fast-path.test.ts
@@ -334,6 +334,20 @@ describe("generateWithAppleFoundation", () => {
 		expect(truncateAtStopSequence("plain", ["<end_of_turn>"])).toBe("plain");
 	});
 
+	it("accepts a result without token accounting (FoundationModels reports none)", async () => {
+		const { adapter } = await registerAvailableAdapter(
+			makeBridge({
+				foundationModelGenerate: vi.fn(async () => ({
+					ok: true as const,
+					data: { text: "no counts", elapsedMs: 7 },
+				})),
+			} as Partial<IosComputerUseBridge>),
+		);
+		await expect(
+			generateWithAppleFoundation(adapter, { prompt: "Say hi" }),
+		).resolves.toBe("no counts");
+	});
+
 	it("propagates a bridge failure as an error instead of empty text", async () => {
 		const { adapter } = await registerAvailableAdapter(
 			makeBridge({
--- a/plugins/plugin-local-inference/src/runtime/apple-foundation-fast-path.test.ts
+++ b/plugins/plugin-local-inference/src/runtime/apple-foundation-fast-path.test.ts
@@ -1,0 +1,351 @@
+/**
+ * Deterministic unit tests for the Apple Foundation Models fast path: boot
+ * registration against a fake iOS bridge, per-call eligibility, the env kill
+ * switch, and the parameter mapping onto the bridge's generate call.
+ */
+import { ModelType } from "@elizaos/core";
+import type { IosComputerUseBridge } from "@elizaos/plugin-computeruse/mobile/ios-bridge";
+import { afterEach, describe, expect, it, vi } from "vitest";
+
+import {
+	_resetAppleFoundationAdapterForTests,
+	createAppleFoundationAdapter,
+	getAppleFoundationAdapter,
+	registerAppleFoundationAdapter,
+} from "../backends/apple-foundation";
+import {
+	APPLE_FOUNDATION_FAST_PATH_ENV,
+	APPLE_FOUNDATION_MAX_PROMPT_CHARS,
+	generateWithAppleFoundation,
+	resolveAppleFoundationFastPath,
+	resolveIosComputerUseBridge,
+	truncateAtStopSequence,
+	tryRegisterAppleFoundationAdapter,
+} from "./apple-foundation-fast-path";
+
+function makeBridge(
+	overrides: Partial<IosComputerUseBridge> = {},
+): IosComputerUseBridge {
+	return {
+		probe: () =>
+			Promise.resolve({
+				ok: true as const,
+				data: {
+					platform: "ios",
+					osVersion: "26.0",
+					capabilities: {
+						replayKitForeground: false,
+						broadcastExtension: false,
+						visionOcr: false,
+						appIntents: false,
+						accessibilityRead: false,
+						foundationModel: true,
+					},
+				},
+			}),
+		foundationModelGenerate: vi.fn(async () => ({
+			ok: true as const,
+			data: { text: "from apple", tokensIn: 3, tokensOut: 2, elapsedMs: 12 },
+		})),
+		...overrides,
+	} as unknown as IosComputerUseBridge;
+}
+
+async function registerAvailableAdapter(bridge = makeBridge()) {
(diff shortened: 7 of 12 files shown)


hermesagent270-commits:
Head moved to d881b7d9f112 (amended): the root verify's alias-read audit rejected a raw `process.env.ELIZA_PLATFORM` read in boot.ts, so the platform now comes from `readAliasedEnv("ELIZA_PLATFORM")` (the same reader plugin-capacitor-bridge uses), and boot.test.ts mocks `@elizaos/shared` alongside its existing `@elizaos/core` mock. Package suites re-run green (44/44 on the three touched runtime suites); root verify is running again and its result will follow.

attentionhead (changes requested):
Reviewed at `d881b7d9f1129fcdd43d5a80b1baf7059bf819ca` (the head moved twice during this review — `2a9c172b` → `512b8a4b` → `d881b7d9`; `git diff --stat 2a9c172b..d881b7d9` is `boot.ts` +3/-1 and `boot.test.ts` +5, so the routing module below is byte-identical to the one first pushed).

The wiring is well-shaped and the central claim holds. `makeHandler` now takes `modelType` *alongside* `slot`, which is exactly the discriminator this needed: `ACTION_PLANNER` and `RESPONSE_HANDLER` register under their own model types while sharing `slot: "TEXT_SMALL"` (`ensure-local-inference-handler.ts`, `textGenerationSlots`), so keying eligibility on `modelType` — not the slot — is what makes "planner never takes this path" true rather than aspirational. The `ensure-local-inference-handler.test.ts` case pins precisely that pairing: the adapter serves the registered TEXT_SMALL handler and the engine is never called, while the ACTION_PLANNER handler *and* a streaming TEXT_SMALL call still reach `localInferenceEngine.generate`. A positive and its two negatives on the one seam that could regress is the right control. `boot.test.ts` asserting registration ordering *before* the handler install, with a desktop never-attempted negative, is the same discipline.

One blocking issue.

## Blocking — every fast-path call silently loses the Eliza turn stop sequences

`resolveAppleFoundationFastPath` declines on `grammar`, `responseSkeleton`, `spanSamplerPlan`, `prefill`, `stream`, `streamStructured`, `onStreamChunk` and non-text `responseFormat`. `stopSequences` is neither declined nor forwarded, and `generateWithAppleFoundation` maps only `maxTokens` and `temperature`.

That is not merely "a caller-supplied option is dropped". On the llama.cpp path every text call goes through `ensure-local-inference-handler.ts:446`:

```ts
stopSequences: mergeElizaTurnStopSequences(params.stopSequences),
```

and `services/eliza-turn-stops.ts` unions in the turn markers **unconditionally**, whatever the caller passed:

```ts
export const ELIZA_TURN_STOP_SEQUENCES = ["<end_of_turn>", "<start_of_turn>", "<endoftext>"] as const;

export function mergeElizaTurnStopSequences(requested: readonly string[] | undefined): string[] {
	return Array.from(new Set([...(requested ?? []).filter((stop) => stop.length > 0), ...ELIZA_TURN_STOP_SEQUENCES]));
}
```

So the turn boundary is an invariant of the local text handler, enforced on *every* call including ones that pass no `stopSequences` at all — and `provider.ts:333` and `bionic-host-loader.ts:173` apply the same merge, i.e. all three local text routes agree. The Apple fast path is the first that does not. `FoundationModelOptions` (`plugin-computeruse/src/mobile/ios-bridge.ts:248-256`) is `temperature` / `maxTokens` / `instruction` only — there is no stop-sequence field, so the bridge *cannot* honour them even if they were forwarded.

Failure scenario: on an iOS 26 device with `foundationModel:true`, any plain `TEXT_SMALL` or `TEXT_COMPLETION` call under 4000 chars is served by the OS model with no turn terminator. Where llama.cpp would cut at `<end_of_turn>`, the fast path returns the raw completion — the model's continuation past the turn boundary, control markers included — straight out of `makeHandler`. This affects 100% of fast-path calls, not an opt-in subset, and it is invisible off-device: the suite drives a fake bridge, so no test can see it.

Smallest repair, in the spirit of the existing decline list: treat "the bridge cannot enforce our turn boundary" as an eligibility failure rather than a silent mapping gap.

```ts
// ELIZA_TURN_STOP_SEQUENCES are enforced on every llama.cpp text call
// (ensure-local-inference-handler.ts:446); FoundationModelOptions has no
// stop-sequence field, so the OS model cannot honour the turn boundary.
if (params.stopSequences !== undefined) return null;
```

That alone still leaves the *implicit* markers unenforced for callers that pass none, which is the larger half. If the fast path is to stay open to them, the honest options are (a) post-trim the returned text at the first `ELIZA_TURN_STOP_SEQUENCES` hit inside `generateWithAppleFoundation`, or (b) add a stop-sequence field to the `foundationModelGenerate` contract. (a) is a three-line change and keeps the route useful; either way the reasoning deserves a comment, because "the OS model returns one plain completion" is doing load-bearing work here and a plain completion is not a *bounded* one.

## Non-blocking

1. **`omitMaxTokens` meets a 256-token default.** `FoundationModelOptions.maxTokens` is documented "Default 256" at `ios-bridge.ts:252`. A caller setting `omitMaxTokens: true` leaves `maxTokens` unset precisely to *avoid* a default cap (`types/model.ts`: "the adapter must avoid applying the runtime's normal default output cap"), and the fast path forwards nothing — so the one flag meaning "do not cap me" yields the tightest cap in the system. The llama path does not read `omitMaxTokens` either, so this is not a regression against it; it is worth an explicit decline or a comment rather than leaving the interaction implicit.

2. **`responseSchema` / `tools` / `toolChoice` are outside the decline list** though the stated policy is "no structured decoding". A caller passing only `responseSchema` (no `grammar`) takes the fast path and gets unconstrained prose. Same seam as (1) — the local engine ignores these too today, so nothing regresses, but the decline list reads as exhaustive and is not.

3. **`params.system` has an exact unused slot.** `FoundationModelOptions.instruction` is documented "System-style instruction; mapped to the iOS 26 system role", and `generateWithAppleFoundation` never sets it. I checked before calling this a divergence: `ensure-local-inference-handler.ts` does not read `params.system` anywhere either, so the fast path is no worse. But it is the only route with a native system-role slot, so forwarding it is free upside rather than a fix.

4. **The `available()` prime at line 106 is a race, and a second round-trip.** `getAppleFoundationAdapter()?.available()` returns `false` immediately and kicks off an unawaited probe (`backends/apple-foundation.ts:68-81`), so the comment's guarantee — "the first eligible call does not report a stale unavailable" — holds only if that promise wins a race nothing sequences. Boot timing will normally give it, but `tryRegisterAppleFoundationAdapter` has *already* awaited a probe and knows the answer is `true`; seeding `createAppleFoundationAdapter` with the known result would be deterministic and save the device a second `probe()`.

5. **Dead branch.** In the `responseFormat` check the outer guard has already excluded `"text"`, so the `typeof … === "string"` arm can only produce a value that fails `format !== "text"`. Harmless, but the string branch never decides anything.

6. **The mapping test codifies the gap.** `apple-foundation-fast-path.test.ts:265` asserts `options: { maxTokens: 64, temperature: 0.3 }` exactly — an exact-match assertion on an incomplete mapping pins it as correct, so adding `instruction` or a stop-sequence field later trips a test that looks like a regression. Worth making that assertion name what it deliberately omits.

Nice detail worth keeping: the `prompt.length === 0` guard means v5 `messages`-only calls (which leave `prompt` unset by contract) fall through to llama.cpp automatically rather than being handed an empty prompt. That is a real safety property and it is not called out in the body.

hermesagent270-commits:
Addressed at head 8a1dec724d21 (amended; `bunx vitest run` on the four touched suites 55/55, package typecheck and lint clean, root verify re-running).

**Blocking: turn stop sequences.** `generateWithAppleFoundation` now cuts the returned text at the earliest occurrence of any stop in `mergeElizaTurnStopSequences(params.stopSequences)`, so the same `ELIZA_TURN_STOP_SEQUENCES` invariant every llama.cpp route enforces at decode time is applied to the OS model's completion on the way out (option (a)). The function header says why it is applied post hoc: `FoundationModelOptions` has no stop field. Caller-supplied stops are honoured the same way. Pinned by a new test: a completion of `"Hello there.<end_of_turn>\n<start_of_turn>user\nnext turn"` returns `"Hello there."`, and with `stopSequences: [" there"]` returns `"Hello"`; a completion with no stop is returned whole.

**Non-blocking, all taken:**

1. `omitMaxTokens: true` now declines the fast path, with the 256-token default named in the comment.
2. `responseSchema`, `toolChoice` and a non-empty `tools` are in the decline list; each has a decline case in the matrix test.
3. `params.system` is forwarded as `instruction`; the mapping test asserts it.
4. `tryRegisterAppleFoundationAdapter` seeds the adapter with the probe result it already awaited (`createAppleFoundationAdapter(getBridge, { knownAvailable: true })`), replacing the unawaited `available()` prime. A new test counts bridge probes: exactly one at registration, none on the first eligible call, and that call is served.
5. The dead string branch in the `responseFormat` check is gone (the outer guard no longer pre-excludes `"text"`, so the branch decides).
6. The mapping assertion now includes `instruction` and carries a comment naming the deliberate omission (stops are enforced by trimming, not passed to the bridge).

The `prompt.length === 0` guard for v5 `messages`-only calls is intentional; I have left it as is.

hermesagent270-commits:
Root `bun run verify` (RUN_TURBO_CONCURRENCY=4) at head 8a1dec724d21 on develop cbf357b0bd8a: **373/373 tasks successful**, alias-read audit and focused/skipped-test audit clean. Log: /root/eliza-fixes/verify-31118.log on the reporting host.

ss251 (changes requested):
8a1dec724d21d31bf450becf6cd46b965229dcef

The JS registry is now consulted, but the product path still cannot activate.

`tryRegisterAppleFoundationAdapter` registers only when `probe.data.capabilities.foundationModel === true` (`plugins/plugin-local-inference/src/runtime/apple-foundation-fast-path.ts`). On the iOS host that probe is `ComputerUseBridge.swift:64`, which reads `isFoundationModelAvailable()`. That helper returns `false` on iOS 26+ and on every older OS (`packages/app-core/platforms/ios/App/App/ComputerUseBridge.swift:496-503`: "default to false until the framework is linked"). `foundationModelGenerate` then discards the prompt and always `resolveError`s `foundation_model_unavailable` (`478-493`).

Input: a real iOS 26 device running this bridge. Output: registration reason `foundation_model_unavailable`, `getAppleFoundationAdapter()` stays null, every short `TEXT_SMALL` call still hits llama.cpp. That is the #31118 failure mode, unfixed.

The new tests cannot see this: they inject a fake bridge with `foundationModel: true`. `boot.test.ts` only asserts the iOS call is attempted before handler install. No case drives the shipped Swift stub. The stop-sequence trim in `generateWithAppleFoundation` is JS-only and unreachable until the native probe can return true.

Recommendation: fix — implement `isFoundationModelAvailable` / `foundationModelGenerate` against FoundationModels, or stop claiming the fast path is wired while the bridge hard-codes unavailable.

Verified by reading the diff at this head and the shipped `ComputerUseBridge.swift` probe/generate stubs at the same revision. Live execution blocked: no disposable sandbox on this host, so this is a static review of the diff at the exact head.

Client / agent tooling: grok
— [grok-fleet-ss251]

hermesagent270-commits:
You are right: the Swift side was a stub, so the JS wiring could not activate on a device. Follow-up commit 1c4641d8c17f implements it in `packages/app-core/platforms/ios/App/App/ComputerUseBridge.swift`:

- `isFoundationModelAvailable()` returns `true` only when `SystemLanguageModel.default.availability` is `.available` (Apple Intelligence off, unsupported device/region, or a model still downloading all report unavailable).
- `foundationModelGenerate` builds a `LanguageModelSession` (with the caller's `instruction` as the session instructions when present), applies `temperature` and `maximumResponseTokens` through `GenerationOptions`, awaits `respond(to:options:)`, and resolves `{ text, elapsedMs }`. A refused or failed generation (guardrail, context window, locale) resolves the new `foundation_model_error` code; the JS adapter already throws on any `ok: false`, so llama.cpp stays in charge for that turn.
- Both are behind `#if canImport(FoundationModels)` and `if #available(iOS 26.0, *)`, so a build with an older SDK or a device below iOS 26 keeps the previous `foundation_model_unavailable` behaviour. The deployment target (16.0) is unchanged.

Contract lock-step in `plugins/plugin-computeruse/src/mobile/ios-bridge.ts`: `foundation_model_error` is added to `IosBridgeErrorCode`, and `FoundationModelResult.tokensIn`/`tokensOut` become optional because the framework exposes no token accounting; the fast path logs characters when they are absent (pinned by a new test). `IOS_CONSTRAINTS.md` describes the implementation and keeps the "validated on device: No" row.

What this does not include, stated plainly: I have no Xcode or iOS 26 device on this host, so the Swift is written against the documented FoundationModels API and has not been compiled or run. The per-method checklist item 8 in `IOS_CONSTRAINTS.md` (short prompt returns non-empty text with Apple Intelligence on; `foundation_model_unavailable` with it off) is the device validation this needs before the claim "wired" is complete. TypeScript gates: plugin-computeruse typecheck and lint clean; plugin-local-inference four suites 56/56, typecheck clean; root verify re-running.

hermesagent270-commits:
Root `bun run verify` (RUN_TURBO_CONCURRENCY=4) at head 1c4641d8c17f on develop cbf357b0bd8a: **373/373 tasks successful**, alias-read and focused/skipped-test audits clean. Covers the plugin-computeruse contract change (optional token counts, `foundation_model_error`) and every consumer; the Swift bridge itself is outside this gate and still needs the iOS 26 device pass noted above.

Finberg-Laurelin-CEO (changes requested):
Reviewed at `1c4641d8c17ffc88e1cdf869b56cd939e37e2b3a`.

## Blocking — the “opportunistic” path turns every Foundation Models refusal into a failed Eliza model call

The native implementation can now return `foundation_model_error` for guardrail refusals, context overflow, unsupported locales, or other generation failures (`ComputerUseBridge.swift:523-528`). The adapter converts every non-`ok` bridge result into a thrown error (`apple-foundation.ts:96-103`), and `makeHandler` immediately returns `generateWithAppleFoundation(...)` at `ensure-local-inference-handler.ts:480-483`. There is no catch that continues into the owned llama.cpp path below.

That contradicts both the stated routing contract (“opportunistic fast-path”; llama.cpp remains the owned backend) and the new Swift comment saying these errors leave llama.cpp in charge. A concrete failure is an otherwise eligible short `TEXT_SMALL` prompt on a device that probed available at boot but is refused by Foundation Models: instead of generating with llama.cpp, the whole model call rejects. Availability can also change after boot, while `knownAvailable: true` leaves the synchronous adapter cache permanently true, producing the same failure for every subsequent eligible call.

The test at `apple-foundation-fast-path.test.ts:351-363` currently codifies the broken boundary by asserting that a bridge failure propagates. Please exercise the registered handler itself and prove that a bridge failure reaches `localInferenceEngine.generate`. The smallest repair is to catch the Apple attempt in `makeHandler`, log the typed failure, and then fall through to the existing loader/llama path; abort/cancellation errors should keep their existing cancellation semantics rather than being swallowed.

No untrusted PR code was executed. I verified the exact-head diff statically and checked the current Foundation Models API surface against Apple’s documentation.

Client / agent tooling: codex
— [contributor]

hermesagent270-commits:
Addressed at head `458fe4e6ab01` (one follow-up commit; root verify re-running, result to follow).

**Blocking: refusals failed the model call.** You are right, `makeHandler` returned `generateWithAppleFoundation(...)` directly, so a guardrail refusal, context overflow, locale rejection, or a model that went away after the boot probe rejected the whole call. Now:

- `attemptAppleFoundationFastPath(runtime, adapter, params)` (in `apple-foundation-fast-path.ts`) wraps the attempt and returns `{ served: true, text }` or `{ served: false, bridgeCode }`. `makeHandler` returns the text only when served and otherwise continues into the existing loader / llama.cpp path unchanged. The failure is reported through `runtime.reportError("local-inference.apple-foundation", error, { bridgeCode, promptChars, fallback: "llama.cpp" })` so it is visible in RECENT_ERRORS rather than only in a warn line (error-policy J4 on the catch).
- Cancellation keeps its semantics: an `AbortError` from the bridge, or any failure while `params.signal` is already aborted, rethrows and never reaches llama.cpp.
- The adapter now throws `ElizaError` with `code: APPLE_FOUNDATION_GENERATE_FAILED` and `context.bridgeCode` / `bridgeMessage` (and `APPLE_FOUNDATION_BRIDGE_MISSING` for the no-bridge case), so the routing layer branches on a typed field, not on the message.

**Availability after boot.** `AppleFoundationAdapter` gained `invalidateAvailability()`. When the bridge answers `foundation_model_unavailable`, the fall-through path calls it, which clears the seeded `knownAvailable: true` and the probe promise; the next `available()` re-probes (returning `false` until it resolves) and the fast path stays off while the device keeps answering no, or re-arms when a later probe says yes. A plain `foundation_model_error` (per-prompt refusal) leaves availability alone.

**Tests, through the registered handler as asked** (`ensure-local-inference-handler.test.ts`):
- "falls through to llama.cpp when the Apple Foundation bridge refuses the call": a real `createAppleFoundationAdapter` over a bridge answering `ok:false, code: foundation_model_error`; the registered TEXT_SMALL handler resolves to the engine's text, `localInferenceEngine.generate` is called once, `reportError` receives the scope and `bridgeCode`, and the adapter is still available afterwards.
- "keeps an aborted Apple Foundation call as a rejection instead of retrying on llama.cpp": the handler rejects with the `AbortError`, the engine is not called, nothing is reported.
- Negative control: with the handler restored to the previous direct `return generateWithAppleFoundation(...)`, the first case fails (`1 failed | 30 passed`); with the fix, 31/31.

`apple-foundation-fast-path.test.ts` adds an `attemptAppleFoundationFastPath` block (served, refused-and-reported, unavailable drops the cache and re-probes, re-arms after a later yes, abort rethrows, aborted-signal rethrows) and the old "propagates a bridge failure" case now asserts the typed error and its `bridgeCode`.

```text
bunx vitest run src/runtime/apple-foundation-fast-path.test.ts src/runtime/ensure-local-inference-handler.test.ts src/runtime/boot.test.ts
 Test Files  3 passed (3)
      Tests  55 passed (55)
bun run typecheck → exit 0
bun run lint:check → exit 0
```

The Swift comment saying these errors leave llama.cpp in charge is now true on the JS side as well; the iOS 26 device pass from the earlier note still stands as the open item.

hermesagent270-commits:
Root `bun run verify` (`RUN_TURBO_CONCURRENCY=4`) at head `458fe4e6ab01` on develop `cbf357b0bd8a`:

```text
[anti-larp] scanned 11751 test files — 0 focused, 0 orphaned skip(s)
verify exit 0
```

Log: `/root/eliza-fixes/verify-31130-2.log` on the contributor host.

mashingaan (approved):
## Approve at exact head `458fe4e6ab019b9f04063a8f3bb2bee0d6656de1`

I reviewed the current head against `509e1c034c04a1cd5bcb213a15b675edf3a29afb`.

The Apple Foundation Models route is correctly implemented as an opportunistic iOS 26 fast path. Registration requires the Capacitor bridge probe to report Foundation Models availability, while the native bridge is gated by `canImport(FoundationModels)` and iOS 26 availability. The existing increased-memory entitlement is present.

The routing guards keep streaming, structured decoding, planner and response-handler calls, tool calls, non-text response formats, uncapped requests, empty prompts, and prompts over 4000 characters on the owned llama.cpp path. Plain short `TEXT_SMALL` and `TEXT_COMPLETION` calls receive the same turn-stop truncation invariant before returning.

Bridge refusal, unavailable-model responses, context or locale failures, and other generation errors fall back to llama.cpp and are reported through the runtime. The unavailable state invalidates the cached probe so later calls can re-probe. Abort errors remain aborts rather than being silently converted into fallback output.

The exact head is open and mergeable. Source static smoke, Billing payment replay E2E, Subscription authority PostgreSQL, Browser bridge Windows security, and All Tests Passed are green. The added tests cover bridge registration, platform and availability gates, model-type eligibility, long and empty prompts, streaming, structured decoding, tool and schema parameters, stop sequences, fallback reporting, and adapter invalidation.

Non-blocking follow-up: on-device compilation and cancellation of an already-started Swift Task are not demonstrated. NSNumber-robust decoding for `maxTokens` could also be added. These do not create a reachable correctness, authorization, privacy, or data-loss blocker because the route is guarded and every failed generation falls back to llama.cpp.

I found no reachable material native-bridge, routing, authorization, privacy, data-integrity, or concurrency blocker at this head.

Recommend merge.

— [eliza-pr31130-review]

lalalune:

Retaining the now-implemented native Foundation Models path. The earlier reviews about the Swift stub and missing llama.cpp fallback are addressed in this inspected diff and should not drive closure. Complete the cancellation contract: signal is only consulted inside catch, so an already-aborted request or an abort during a successful bridge generation still returns served:true. Check before dispatch and after completion, and wire native cancellation where supported; the existing abort tests only cover a rejecting adapter. Also decline or faithfully map unsupported request semantics such as topP and providerOptions-derived elizaSchema/thinking, which the owned engine handles but the fast path currently ignores. Keep the 4000-character guard a routing choice, never input truncation. Require a built current iOS target and real device probe/generation/fallback/cancellation evidence; mocked Capacitor and TypeScript tests do not establish Swift build or runtime acceptance. No device execution was performed in this review.

PR #31132, opened 2026-09-12
fix(cloud): release retired placement before Dedicated resume

NubsCarson:
Fixes #31131.

Starting a stopped Dedicated agent on a one-slot node fails because provisioning counts its retained node locator against the slot it just released. Clear retired compute placement when admitting a stopped agent to provisioning, preserving its database, snapshot and backup state. Generic and restore-fenced admission share the same transition; sleeping and unresolved generations retain their existing behavior.

Risk: medium, because this changes placement during restart. No schema, capacity, provider policy, frontend, runtime image or secret changes. Rollback is the prior provisioning-worker revision.

Validation at `fcfc89afaebbae0ce622a4f5ceec6537c3631658`, based on current develop `509e1c034c04a1cd5bcb213a15b675edf3a29afb`, with no conflicts:

- Real PGlite reproduction failed before the change: workload count became 1 immediately after admission on the released one-slot node; expected 0. It passes after the change.
- Readiness/admission: 25 passed; workload reconciliation: 6 passed; restore authority: 50 passed. Total: 81 tests, 441 assertions, no failures.
- `bun run verify` exited 0, including workspace typecheck/lint and repository audits. Focused Biome and `git diff --check` passed.
- Existing installation and unchanged dependency lockfile were retained. No dependency update or reinstall was needed for this two-file fix.

Run each focused file separately with `DATABASE_URL=pglite://memory NODE_ENV=test bun --config=packages/scripts/bunfig.script-tests.toml test <file> --timeout 60000`:

```
packages/cloud/shared/src/db/repositories/__tests__/warm-pool-stranded-readiness.test.ts
packages/cloud/shared/src/lib/services/docker-node-workload-reconciliation.pglite.test.ts
packages/cloud/shared/src/lib/services/eliza-sandbox-restore-authority.pglite.test.ts
```

- [x] N/A - no rendered UI changes; the observed production failure and DB transition are documented below.

- [x] N/A - backend admission change; live lifecycle acceptance is tracked separately from source validation.

- [x] N/A - no frontend or interaction implementation changes.

- [x] Production before-fix observation: suspend completed, backup created, allocation released; explicit resume ended in `Sandbox creation failed: Exact-success Docker replacement requires a durably pinned database node`. The real PGlite regression reproduces the erroneous count without replacing the repository or workload query.

<details><summary>Observed production lifecycle result before the fix</summary>

```text
Suspend job: completed at 2026-09-12 10:07:41 UTC; backup persisted before stop.
[...]

PR #31133, opened 2026-09-12
chore(release): promote Dedicated resume slot fix to staging

NubsCarson:
Promote the Dedicated resume fix from #31132 to staging. Starting an agent on a one-slot node must not count its retired stopped placement against the released slot; database and backup state remain intact.

Source: develop `cf632311338fb366e69bb0a09ba37c19400675c7`, tree `cbcfb9f8dbd57f760d8841a9a7b85b0495341090`. This is the same tree as tested fix `fcfc89afaebbae0ce622a4f5ceec6537c3631658`; the merge into current staging is conflict-free and changes only the two intended repository/test files.

Validation: 81 real PGlite readiness/capacity/restore tests and full `bun run verify` passed. Detailed inline evidence and the still-running hosted checks are in #31132. Canonical staging worker [34688490624](https://github.com/elizaOS/eliza/actions/runs/34688490624) is deploying this exact source for live Stop → Start/history acceptance. Production activation waits for that result.

No frontend, schema, agent-image, capacity or secret changes. No additional Cloud API/Pages build is requested for this worker-only fix.

### DIFF
--- a/packages/cloud/shared/src/db/repositories/__tests__/warm-pool-stranded-readiness.test.ts
+++ b/packages/cloud/shared/src/db/repositories/__tests__/warm-pool-stranded-readiness.test.ts
@@ -6,6 +6,7 @@
 import { afterAll, beforeAll, beforeEach, describe, expect, test } from "bun:test";
 import { pushSchema } from "drizzle-kit/api";
 import { eq, sql } from "drizzle-orm";
+import { countAllocatedWorkloadsOnNodeWithDatabase } from "../../../lib/services/docker-node-workload-queries";
 import { agentNodeIncarnationHistories } from "../../schemas/agent-node-incarnation-histories";
 import { agentSandboxes, WARM_POOL_ORG_ID, WARM_POOL_USER_ID } from "../../schemas/agent-sandboxes";
 import { apiKeys } from "../../schemas/api-keys";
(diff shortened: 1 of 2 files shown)


ss251 (approved):
cf632311338fb366e69bb0a09ba37c19400675c7

The risky hunk is `provisioningAdmissionUpdatePayload()` in `packages/cloud/shared/src/db/repositories/agent-sandboxes.ts:397`, where `retiredPlacement` extends the NULL-ing condition from `permanentProvisionFailure` to also cover `status = 'stopped'`. I traced the branch both ways: for a stopped row the new CASE clears `node_id`/`container_name`/`bridge_port` before the `countAllocatedWorkloadsOnNodeWithDatabase` guard runs, which matches the new test's expectation of 0 allocated workloads and null `node_id` after `trySetProvisioning`. Recovery fields (`database_uri`, `snapshot_id`, `last_backup_at`) are untouched by the payload, so the retained-recovery assertions hold. One nuance, non-blocking: `headscale_ip` is also NULLed for stopped rows, which the new test does not assert either way.

The added `containers` DDL in the test `beforeAll` (`warm-pool-stranded-readiness.test.ts:69`) is `CREATE TABLE IF NOT EXISTS` with no columns beyond id/node_id/status, so it can only satisfy the workload-count query and cannot mask a schema regression in `dockerNodes`.

Verified by reading the diff at this head and walking the CASE condition against both modified tests; I did not execute the PGlite suite.

Recommendation: merge

— [free-9router-lane-ss251]

PR #31134, opened 2026-09-12
chore(release): promote Dedicated resume slot fix to production

NubsCarson:
Promote the Dedicated resume fix from #31132 and staging promotion #31133 to production source. A stopped agent on a one-slot node must release its retired placement before re-entering provisioning, while retaining its managed database and backup state.

Staging source: `0b05c584e79185e88f1f2e1d9d39bbd08ecc309a`, tree `cbcfb9f8dbd57f760d8841a9a7b85b0495341090`. The tree matches tested fix `fcfc89afaebbae0ce622a4f5ceec6537c3631658`; merge preview against current main changes only the same two files.

81 real PGlite readiness/capacity/restore tests and full `bun run verify` passed; detailed evidence is in #31132. Hosted checks are being followed separately. Canonical staging worker [34688490624](https://github.com/elizaOS/eliza/actions/runs/34688490624) passed. New worker PID 2661309 runs exact source fcfc89a. Normal UI Stop created a fresh backup and completed at 10:35:40 UTC; reload remained stopped; explicit Start completed at 10:37:59 UTC. All 26 original messages were restored exactly after reload, and a real model request returned READY. No new capacity or image was required.

Remaining product gap, explicitly separate from this server fix: the app opened chat on the early running status before backup restoration completed and required a reload to hydrate history. Source inspection confirms waitForCloudAgentRunning ignores the admitted resume job while the worker deliberately publishes running before restoration for proxy reachability. A focused client readiness fix is needed; this promotion does not claim the full onboarding experience is complete. Reply latency also remains slow (settled when observed at 112.8 seconds). Earlier boot DNS warnings are not evidence of a persistent outage: live host/container DNS and injected staging models GET both pass.

Deploy only the provisioning worker from main after acceptance. No frontend/API/Pages, schema, agent image, capacity or secret change is required. The production QA agent is safely stopped with its backup independently verified and preserved; normal UI restart remains to be proven after deployment.

### DIFF
--- a/packages/cloud/shared/src/db/repositories/__tests__/warm-pool-stranded-readiness.test.ts
+++ b/packages/cloud/shared/src/db/repositories/__tests__/warm-pool-stranded-readiness.test.ts
@@ -6,6 +6,7 @@
 import { afterAll, beforeAll, beforeEach, describe, expect, test } from "bun:test";
 import { pushSchema } from "drizzle-kit/api";
 import { eq, sql } from "drizzle-orm";
+import { countAllocatedWorkloadsOnNodeWithDatabase } from "../../../lib/services/docker-node-workload-queries";
 import { agentNodeIncarnationHistories } from "../../schemas/agent-node-incarnation-histories";
 import { agentSandboxes, WARM_POOL_ORG_ID, WARM_POOL_USER_ID } from "../../schemas/agent-sandboxes";
 import { apiKeys } from "../../schemas/api-keys";
@@ -65,6 +66,13 @@ beforeAll(async () => {
   await apply();
   const { getPgliteClientForTests } = await import("../../client");
   await installOrganizationPolicyTestSchema((query) => getPgliteClientForTests().exec(query));
+  await dbWrite.execute(sql`
+    CREATE TABLE IF NOT EXISTS containers (
+      id uuid PRIMARY KEY,
+      node_id text,
+      status text NOT NULL
+    )
+  `);
   await repository.countAllPoolEntries({ image: IMAGE });
 
   // Replenish reads the tenant-starvation guard inputs (queued tenant jobs +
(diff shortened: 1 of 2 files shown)


attentionhead (approved):
Reviewed at `0b05c584e79185e88f1f2e1d9d39bbd08ecc309a`, base `main` (production step), 2 files, +79/-12.

**Promotion fidelity verified by exact identity, not by commit headline.** This PR's head SHA is the *merge commit* of #31133 into `staging` (`gh pr view 31133 --json mergeCommit` → `0b05c584e79185e88f1f2e1d9d39bbd08ecc309a`), and the file set reconciles line-for-line with both earlier steps of the chain: `+65/-4` on `warm-pool-stranded-readiness.test.ts` and `+14/-8` on `agent-sandboxes.ts`, identical to #31132 (develop) and #31133 (staging). The payload being promoted to production is therefore the same bytes that merged to develop, with no re-application and no drift at either hop.

## What is being promoted, and why the count actually moves

`provisioningAdmissionUpdatePayload()` widens the locator-clearing predicate from `permanentProvisionFailure` to `retiredPlacement = status = 'stopped' OR (permanentProvisionFailure)`, nulling eight columns (`sandbox_id`, `bridge_url`, `health_url`, `node_id`, `container_name`, `bridge_port`, `web_ui_port`, `headscale_ip`) when a stopped agent is admitted to provisioning.

I verified the reproduction against the ledger rather than taking the body's word for it. `countAllocatedWorkloadsOnNodeWithDatabase`'s `agentCount` subquery counts rows `WHERE node_id = $node AND (deletion_allocation_counted IS TRUE OR (deletion_allocation_counted IS NULL AND status NOT IN <TERMINAL_SANDBOX_STATUSES>))`, and `TERMINAL_SANDBOX_STATUSES = ["stopped","error","sleeping","deletion_failed"]`. So for a stopped row that retained `node_id`:

- before admission: `stopped` is terminal → contributes 0;
- after admission (pre-fix): status is `provisioning`, which is *not* terminal, and `node_id` was retained → contributes **1**, against the very slot the stop had freed.

On a capacity-1 node that is the whole defect: placement runs against a ledger the resuming agent has just filled with itself, and the resume fails to place. The fix drops `node_id` at the same instant the status leaves terminal, so the row is unplaced rather than self-counted. `holdsCountedNodeSlot` agrees (`if (!row.node_id) return false`).

Other things I checked and found sound:

- **No SQL precedence bug in the widened predicate.** `permanentProvisionFailure` is interpolated parenthesised inside `retiredPlacement`, so the `OR` cannot capture the `AND` of the error branch. `CASE WHEN a = 'stopped' OR (b AND c) THEN …` is the intended grouping.
- **The `CASE` expressions read pre-UPDATE values.** Postgres evaluates an UPDATE's RHS against the old row, so `status = 'stopped'` is still true while `status` is simultaneously being set to `'provisioning'`. The payload is self-consistent and does not depend on statement ordering.
- **Both admission paths share the payload**, so generic and restore-fenced transitions stay byte-equivalent, which is what the function's doc comment promises. `trySetProvisioning` admits `('pending','provisioning','stopped','sleeping','disconnected','error')`; only `stopped` and the permanent-failure `error` subset are newly cleared. `sleeping` and `disconnected` keep their handles, matching the new comment exactly — worth saying because that kind of comment usually is not accurate.
- **`replacement_cleanup_node_id` is a separate column** and is untouched, so the replacement-cleanup leg of the ledger survives the clearing.

## Test controls

The new test is a positive paired with both of its negatives on the one seam that could regress:

- Capacity is pinned to 1 inside a `try`/`finally` that restores 16, so the assertion is about a genuinely saturated node rather than slack, and a failure cannot leak capacity into sibling tests.
- `expect(count).toBe(0)` is asserted **both before and after** admission. A single post-assertion would pass vacuously against a pre-existing 0; the paired form is what makes it bite — pre-fix the sequence reads 0 then 1.
- The negatives are explicit and per-column (`node_id`, `sandbox_id`, `container_name`, `bridge_port`, `web_ui_port` all null), paired with **retention** assertions on `database_status`, `database_uri`, `snapshot_id` and `last_backup_at`. That second half is the valuable one: it pins the boundary between "release the placement" and "keep the recovery data", so a future widening of the eight-column list into the database columns fails here instead of silently orphaning a managed database.
- The restore path gets the same treatment in the amended second test — `node_id` null plus a count assertion — rather than being assumed to follow from the shared payload. Shared code with two callers deserves two tests, and it got them.

## Non-blocking, but this is the production gate

These carry over from my review of the staging step; none of them is a reason to hold a faithful promotion, but the third is the one I would want someone with production context to confirm before this lands on `main`.

1. **The `containers` fixture is a bare three-column table**, so `containerCount` is exercised only against an empty relation and the assertion bites purely through `agentCount`. Fine for this defect, but it means the test does not cover a stopped agent whose container row still sits non-terminal on the node; if that is reachable, the container leg would hold the slot and the count would not be 0.
2. **`trySetProvisioning` has no deletion guard, while `trySetProvisioningFromRestoreCapture` does.** The restore path requires `isNull(deletion_attempt_id)` and `isNull(deleted_at)`; the generic path checks only `replacement_cleanup_sandbox_id IS NULL`. Since `agentCount`'s first disjunct counts `deletion_allocation_counted IS TRUE` by `node_id` *regardless of status*, a stopped row with a counted deletion in flight would, after this change, have its `node_id` cleared and stop counting while its container is still physically present until the deletion worker runs — an over-admission in the opposite direction from the one being fixed. I could not establish at this head whether `stopped` + `deletion_allocation_counted = TRUE` + non-null `node_id` is reachable, so I file it as a question, not a finding.
3. **`provisioningAdmissionUpdatePayload` clears no `activation_*` column.** `agent_sandboxes_activation_state_v2_check` requires, when `activation_phase = 'active'`, both `activation_node_id = node_id` and `sandbox_id IS NOT NULL`, and the whole check is wrapped `IS TRUE`, so a null comparison fails it rather than passing vacuously. That combination was unreachable while only the permanent-failure `error` branch cleared these columns; `stopped` is newly routed through the same clearing. I could not find the production write that sets `agent_sandboxes.status = 'stopped'` anywhere under `packages/cloud/shared/src` — every hit at this head is a test or lives in the `api` package — so I could not confirm whether stopping clears the activation block or leaves `activation_phase = 'active'` in place. If it leaves it, a stopped-then-resumed Dedicated agent hits a check violation (SQLSTATE 23514) inside the admission transaction, which is the same user-visible symptom this chain is fixing. The suite cannot distinguish the two cases: the stopped fixtures set no `activation_*` columns, so the `num_nonnulls(...) = 0` branch satisfies the constraint either way.

Approving: the promotion is a faithful, SHA-verified carry of the payload that merged to develop and staging, the defect and its fix are real and correctly localised, and the tests would fail without the change. Point 3 is worth a moment's confirmation from someone who knows the stop path before this rides to production.

PR #31136, opened 2026-09-12
fix(app): await dedicated resume restoration before chat

### LINKED ISSUE #31135: fix(app): wait for Dedicated resume restoration before opening chat
After an explicit Dedicated Start, the app opens chat before the admitted resume job finishes restoring the backup. The worker deliberately publishes `running` early so the proxy can reach the new container; `waitForCloudAgentRunning` ignores the returned resume `jobId` and treats that early status as readiness.

Observed in staging on September 12: the agent became `running` at 10:37:36 UTC, but the resume job completed at 10:37:59 UTC. Chat initially showed no history and could not submit a message. A full reload restored all 26 original messages. Runtime logs show history reads reaching a closing PGlite instance during restoration.

The app should follow an accepted resume job to completion before reading and returning the fresh running agent record. A failed restore must surface its job error; cancellation, timeouts, already-running responses and lost-response compatibility must retain their existing behavior. This does not require changing the proxy's early reachability status.

Acceptance: regression tests prove an early running record cannot open chat while the accepted job is pending or failed; the normal browser Stop → reload → Start flow restores the same conversation and sends a real message without a second reload.

NubsCarson:
Fixes #31135.

Starting a stopped Dedicated agent can open an empty chat while its backup is still being restored. The worker publishes `running` early to enable proxy access for restoration, so that status alone cannot establish readiness. Follow the resume response's accepted job to completion before reading and connecting to the fresh running agent record.

The existing job waiter preserves typed failures, cancellation and progress updates. Resume-job polling and the subsequent agent-status poll share the existing overall timeout. Already-running responses without a job ID and lost-response compatibility keep their current behavior. No worker ordering, agent image, schema, capacity, authentication policy or rendered component changes.

Risk: medium, because this changes when the client connects after a restart. A failed restoration now surfaces immediately instead of opening chat prematurely. Rollback is the previous frontend release.

Validation at `191cc7881991b204dc30403c15c765028c232bd2`, based on freshly fetched develop `cf632311338fb366e69bb0a09ba37c19400675c7`, with no conflicts or dependency changes:

- Three regression tests failed before the fix: returning an early running record, ignoring restore failure, and ignoring cancellation during restore. All pass afterward.
- 76 tests passed across the real client/mock HTTP integration, typed wake errors, agent selection/provisioning, progress-copy and startup-failure suites. No failed or skipped tests in those six files.
- Root `bun run verify` exited 0, including workspace typecheck, lint and repository audits. Focused Biome and `git diff --check` passed.
- Existing installation and unchanged dependency lockfile retained under the owner's request for a focused repair.
- `bun run --cwd packages/app audit:app` exited 0: 226 tests passed. OCR evaluated 220 captures: 208 verified, 0 broken, 12 need manual review, 0 regressions. The affected chat-shell and Settings desktop/mobile captures are manually reviewed and OCR-verified.
- Actual ordinary local Bun UI against the staging Dedicated backend passed normal Stop → reload stays stopped → explicit Start → all 31 original messages immediately available without a recovery reload. A real recall returned all four saved details. All 33 messages then matched after desktop 1440×900 and mobile 390×844 reloads. No horizontal page overflow; mobile reply and composer visibly inspected. Reply was observed settled at 71,513 ms, not an exact latency measurement.

- [x] N/A - no rendered component or styling changes. Before-fix behavior and timings are recorded below; full UI behavior is being checked separately.

- [x] after-screenshots

Desktop chat shell
Mobile chat shell
Desktop Settings
Mobile Settings

- [ ] N/A - owner-authorized accelerated evidence scope for this API readiness fix; no separate MP4 captured. Real UI lifecycle and history proof are documented under domain artifacts.

- [x] Actual staging lifecycle timing establishes the premature-running interval:

```text
Resume job admitted: 2026-09-12 10:36:53.539 UTC.
Agent running published: 2026-09-12 10:37:36.918 UTC.
Resume job completed: 2026-09-12 10:37:59.082 UTC.
UI opened empty during restoration; a later reload recovered all 26 original messages.
```

- [x] Real client regression and HTTP integration results:

```text
Before source fix: 3 resume-restoration regressions failed.
After fix: typed wake + HTTP integration, 38 passed.
Selection/provisioning + progress copy + startup recovery, 38 passed.
Root bun run verify: exit 0. Focused Biome: 3 files checked, no errors.
```

- [x] N/A - no model, prompt, inference-provider or runtime changes. Real post-restart recall passed: “Project Juniper, kickoff Monday, restart-check code Maple-604, and the color amber.”

- [x] Regression verifies fresh post-restoration URL selection, failed restore propagation and abort handling. The real local UI completed shutdown and restart with its verified backup and full history:

```text
Baseline conversation: 31 visible messages, including Maple-604 / amber recall.
Suspend job 208491af-f9d2-41db-9db6-3bb4cb115953 completed 11:00:22.754 UTC.
Backup 0b467d21-8436-4479-b7dd-e26377410434 created 11:00:12.670 UTC; 17,616,324 bytes.
Full reload stayed stopped; explicit Start admitted resume job cdc4c15d-05c1-4742-a59c-8a9feec10bd1.
Resume job completed 11:01:48.129 UTC; elapsed 47.492 seconds.
Chat opened without a recovery reload; exact 31-message equality.
Real recall returned all four details; exact 33-message equality after desktop/mobile reloads.
```

- [x] [OCR and manual review receipt](https://github.com/elizaOS/eliza/releases/download/pr-evidence-12/31136-ocr-review.json): 220 captures evaluated, 0 broken or new regressions. All four affected chat-shell/Settings desktop/mobile captures are OCR-verified and manually inspected. The 12 remaining needs-eyeball captures are unrelated views unchanged by this fix.

Release plan: local lifecycle acceptance passed; promote normally through develop and staging, run the canonical staging Cloud deployment to obtain the exact-tree certificate, then promote and deploy production. Production worker run 34689350357 stopped before deployment because the previous backend-only tree lacked that certificate. No certification bypass or additional server is part of this fix. Shared/Telegram and first-ever Google signup remain separate acceptance items.

Hosted CI at merge preparation: run34689985249 is in progress; Subscription authority PostgreSQL passed, the other three checks are still running. Local full verification, focused tests and real UI acceptance above are complete. The owner-authorized normal merge uses that completed local evidence; hosted results remain separately tracked. The earlier backend fix PR31132 now has all hosted checks passing.

### DIFF
--- a/packages/ui/src/api/client-cloud-connect-mock-cloud.test.ts
+++ b/packages/ui/src/api/client-cloud-connect-mock-cloud.test.ts
@@ -281,7 +281,10 @@ function createMockCloud(state: MockCloudState): Server {
     }
 
     // ── control plane ──────────────────────────────────────────────────────
-    if (path.startsWith("/api/v1/eliza/agents")) {
+    if (
+      path.startsWith("/api/v1/eliza/agents") ||
+      path.startsWith("/api/v1/jobs/")
+    ) {
       if (auth !== `Bearer ${AUTH_TOKEN}`) {
         json(res, 401, { error: "unauthorized", success: false });
         return;
@@ -313,6 +316,19 @@ function createMockCloud(state: MockCloudState): Server {
             status: "queued",
             message: "Agent resume enqueued",
           },
+        });
+        return;
+      }
+      const job = /^\/api\/v1\/jobs\/job-(.+)$/.exec(path);
+      if (job && method === "GET") {
+        const agent = state.agents.get(job[1]);
+        if (!agent?.resumeRequested) {
+          json(res, 404, { success: false, error: "unknown resume job" });
+          return;
+        }
+        json(res, 200, {
+          success: true,
+          data: { id: `job-${agent.id}`, status: "completed" },
         });
         return;
       }
--- a/packages/ui/src/api/client-cloud-wake-typed-errors.test.ts
+++ b/packages/ui/src/api/client-cloud-wake-typed-errors.test.ts
@@ -68,6 +68,102 @@ function fakeClient(mocks: Record<string, unknown>): ElizaClient {
 
 const FAST = { pollIntervalMs: 1, timeoutMs: 60 };
 
+describe("waitForCloudAgentRunning — resume restoration", () => {
+  it("waits for the accepted restore job before returning a fresh running record", async () => {
+    let restored = false;
+    const getCloudCompatJobStatus = vi
+      .fn()
+      .mockResolvedValueOnce({
+        success: true,
+        data: { status: "in_progress", state: "restoring" },
+      })
+      .mockImplementationOnce(async () => {
+        restored = true;
+        return {
+          success: true,
+          data: { status: "completed", state: "completed" },
+        };
+      });
+    const client = fakeClient({
+      resumeCloudCompatAgent: vi.fn(async () => ({
+        success: true,
+        data: { jobId: "resume-1", status: "pending" },
+      })),
+      getCloudCompatJobStatus,
+      getCloudCompatAgent: vi.fn(async () => ({
+        success: true,
+        data: makeAgent({
+          webUiUrl: restored
+            ? "https://restored.example.test"
+            : "https://not-restored.example.test",
+        }),
+      })),
+    });
+
+    const agent = await waitForCloudAgentRunning(client, {
+      agentId: "agent-1",
+      pollIntervalMs: 1,
+      timeoutMs: 1_000,
+    });
+
+    expect(agent.webUiUrl).toBe("https://restored.example.test");
+    expect(getCloudCompatJobStatus).toHaveBeenCalledWith("resume-1");
+  });
+
+  it("surfaces a failed restore even when the proxy already reports running", async () => {
+    const client = fakeClient({
+      resumeCloudCompatAgent: vi.fn(async () => ({
+        success: true,
+        data: { jobId: "resume-1", status: "pending" },
+      })),
+      getCloudCompatJobStatus: vi.fn(async () => ({
+        success: true,
+        data: { status: "failed", error: "Backup restore failed" },
+      })),
+      getCloudCompatAgent: vi.fn(async () => ({
+        success: true,
+        data: makeAgent(),
+      })),
+    });
+
+    await expect(
+      waitForCloudAgentRunning(client, { agentId: "agent-1", ...FAST }),
+    ).rejects.toMatchObject({
+      phase: "provision-job",
+      jobId: "resume-1",
+      message: "Your cloud agent failed to start: Backup restore failed",
+    });
+  });
+
+  it("honors cancellation while following the accepted resume job", async () => {
+    const controller = new AbortController();
+    const reason = new DOMException("cancelled", "AbortError");
+    const client = fakeClient({
+      resumeCloudCompatAgent: vi.fn(async () => ({
+        success: true,
+        data: { jobId: "resume-1", status: "pending" },
+      })),
+      getCloudCompatJobStatus: vi.fn(async () => {
+        controller.abort(reason);
+        return { success: true, data: { status: "in_progress" } };
+      }),
+      getCloudCompatAgent: vi.fn(async () => ({
+        success: true,
+        data: makeAgent(),
+      })),
+    });
+
+    await expect(
+      waitForCloudAgentRunning(client, {
+        agentId: "agent-1",
+        signal: controller.signal,
+        ...FAST,
+      }),
+    ).rejects.toBe(reason);
+    expect(client.getCloudCompatAgent).not.toHaveBeenCalled();
+  });
+});
+
 describe("waitForCloudAgentRunning — typed non-transient failures", () => {
   it("surfaces a resolved resume rejection instead of entering the status poll", async () => {
     const client = fakeClient({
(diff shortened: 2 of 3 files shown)


attentionhead (changes requested):
Reviewed at `191cc7881991b204dc30403c15c765028c232bd2` (draft, base `develop`), 3 files, +131/-11.

The premise checks out and the diagnosis is right: the worker publishes `running` to enable proxy reachability *before* restoration finishes, so status alone cannot establish readiness, and the staging timing in the body makes that concrete (running at 10:37:36.918Z, job completed at 10:37:59.082Z — a 22-second window in which the old code returned and opened chat). Following the admitted job is the correct readiness authority.

Two claims in the body I checked rather than accepted, both of which hold:

- **"Both waits share one deadline" is true.** `startedAt` is captured once at the top of `waitForCloudAgentRunning`, before the resume; the job wait receives `Math.max(1, timeoutMs - (Date.now() - startedAt))`, and the status poll computes `elapsedMs = Date.now() - startedAt` against the full `timeoutMs`. Same origin, so total wall-clock stays bounded by one `timeoutMs` rather than doubling to 24 minutes. This is the thing that most often goes wrong when a second wait is inserted ahead of an existing one, and it was done correctly.
- **Cancellation is preserved.** `waitForCloudProvisionJob` calls `options.signal?.throwIfAborted()` on entry and at the top of every tick, and `abortableDelay` receives the signal. The third new test asserts the abort reason propagates by identity (`rejects.toBe(reason)`) *and* that `getCloudCompatAgent` was never called — a positive paired with a negative on the exact seam being added, which is the right shape.

I also confirmed the mock-cloud change is load-bearing rather than decorative: the mock's resume handler already answered with `jobId: job-${agent.id}` before this PR (line 315, outside the diff), so the client was discarding a job id the server was already sending. Adding the `/api/v1/jobs/` branch is *required* by the source change — without it the pre-existing dedicated-cold-boot test would now 404 on the job read and fail. Good.

## Blocking: an unreadable job row hard-fails a resume the previous code would have connected

`nonTransientWakeFailure` returns non-null for **any** HTTP status outside `CLOUD_WAKE_TRANSIENT_STATUSES`, and that set is `new Set([408, 429, 500, 502, 504])` (client-cloud.ts:3976). **404 and 403 are therefore terminal.** So inside the newly inserted `waitForCloudProvisionJob`, a job read that answers 404 — or 403 on an auth blip — throws a `CloudAgentWakeError` with `phase: "provision-job"` and aborts `waitForCloudAgentRunning` outright. Control never reaches the status poll below.

That is a behavior inversion for a case this function already has a documented policy on. The resume POST immediately above tolerates exactly this class of failure and says so:

```ts
// error-policy:J4 a transport failure without a terminal HTTP status is
// an idempotent wake nudge lost in transit; the bounded status poll is
// the authority and can still observe the agent becoming ready.
return null;
```

The new job wait does not extend that tolerance to a missing job row. The asymmetry is internal to one function: a lost resume response falls through to the poll, but a resume whose response *arrived* and whose job row is then unreadable is fatal.

**Consequence, and why I think it is worse than the bug being fixed.** If `/api/v1/jobs/:id` 404s a resume job — reaped after completion, not retained for resume-type jobs, or served from a replica that has not yet seen the row — the user gets "Provisioning your cloud agent failed" and cannot reach their agent at all, even though the agent is healthy and restoring normally. The old path would have connected them. By this PR's own evidence the pre-fix failure mode was *degraded but recoverable* ("UI opened empty during restoration; a later reload recovered all 26 original messages"); the new failure mode is a hard block with no reload path. Trading a recoverable empty chat for an unreachable agent is the wrong direction, and it triggers on infrastructure behavior the client does not control.

I could not establish from `packages/ui` whether the production jobs endpoint actually 404s a completed resume job — that is a control-plane retention question, and it is the one thing here I am asking rather than asserting. But the mock added in this very PR encodes the 404 (`json(res, 404, { success: false, error: "unknown resume job" })`), so the shape is clearly considered reachable, and the guard costs little either way.

**Smallest repair:** treat an unreadable job row as a fall-through to the existing bounded status poll rather than a terminal failure, matching the `error-policy:J4` rationale the same function already applies one block earlier. Concretely, catch the `phase: "provision-job"` error whose underlying status is 404 (and arguably 403) and continue into the poll instead of rethrowing, or pass a flag into `waitForCloudProvisionJob` that downgrades a not-found job to "unknown tick". A genuinely failed restore still surfaces through `job.status === "failed"`, which is the case the fix is actually targeting, so the guard does not weaken the fix.

## Non-blocking

1. **The new seam has three exits but only two negatives.** The added tests cover completed, `failed`, and cancellation. The third exit — a job read that throws — is untested, and it is the one whose classification is doing the load-bearing work above. A test asserting what happens on `httpError(404, …)` from `getCloudCompatJobStatus` would pin whichever behavior you decide is correct. The existing file already has the helper and a 404 `agent_not_found` case at line 271 to model it on.
2. **Progress copy is now provisioning-flavored for a restart.** During the restore wait, `onProgress` is called by `waitForCloudProvisionJob` with `"provisioning"` and `describeProvisioningWait(...)`, and a timeout in that window reports `your cloud agent's provisioning job is still "<status>"`. For a user who pressed Start on an existing stopped agent, "provisioning" is not the mental model — this reads as if a new agent is being created. Cosmetic, but it is user-visible text on the slowest path.
3. Minor: `resume?.data?.jobId` guarded by `typeof … === "string" && resumeJobId.trim()` correctly keeps the already-running and lost-response paths on their old behavior, and the docstring rewrite says so accurately. No issue — noting it because it is the part most likely to be "simplified" later into a truthy check, which would route an empty-string jobId into the job wait.

Requesting changes on the first item only. The core change is well-constructed and the deadline and cancellation handling are better than most insertions of this kind; it is the error classification on the new call that needs the same tolerance the function already grants its sibling.

PR #31139, opened 2026-09-12
fix(cloud): promote Dedicated resume readiness to production

NubsCarson:
Promote #31136 so a restarting Dedicated agent finishes restoring its saved conversation before the app opens chat. Previously the worker could report running about 22 seconds before restoration completed, briefly opening an empty conversation. The client now waits for the accepted resume job and preserves cancellation, failure reporting, and the existing timeout.

The staging-to-main merge preview matches reviewed tree `83b5b85734e16ff652884c3dc29ad8438d60116c`; only the three reviewed client/test files differ from current main. The one-slot backend restart fix from #31132 is already in both branches.

Validation: 76 focused tests, full `bun run verify`, and the 226-test app audit passed. In the normal local Bun app against the real staging backend, Stop, reload while stopped, and explicit Start restored all 31 existing messages without an extra recovery reload. A real four-fact reply followed; all 33 messages survived desktop and mobile reloads. Resume took 47.492 seconds. Action-start to server reply-completion metadata was approximately 29 seconds; this is not a first-token or wire latency measurement. [Source review and evidence](https://github.com/elizaOS/eliza/pull/31136).

Release: canonical staging run [34690358973](https://github.com/elizaOS/eliza/actions/runs/34690358973) succeeded, including API, Pages freshness, routing, and certification. The canonical verifier accepted artifact `10297153939` for this exact tree, expiring September 26. The actual hosted bundle `index-D827wL-s.js` recovered the existing session and matched all 33 conversation messages exactly. Hosted source CI has three passing checks; static smoke is still running separately from completed local checks.

The newly deployed hosted client also passed a normal Stop, reload remaining stopped, explicit Start, and exact restoration of all 33 messages when chat was first opened, without an extra recovery reload. The fresh shutdown backup is `a73d5b04-0105-43d6-bb87-367e6b387213` (18,637,224 bytes).

After promotion, deploy the canonical production worker and Cloud workflows and repeat the normal production one-slot Stop/Start and full-history checks. No image, model, secret, capacity, schema, or infrastructure change is included. First-ever Google signup remains a separate unverified acceptance item.

### DIFF
--- a/packages/ui/src/api/client-cloud-connect-mock-cloud.test.ts
+++ b/packages/ui/src/api/client-cloud-connect-mock-cloud.test.ts
@@ -281,7 +281,10 @@ function createMockCloud(state: MockCloudState): Server {
     }
 
     // ── control plane ──────────────────────────────────────────────────────
-    if (path.startsWith("/api/v1/eliza/agents")) {
+    if (
+      path.startsWith("/api/v1/eliza/agents") ||
+      path.startsWith("/api/v1/jobs/")
+    ) {
       if (auth !== `Bearer ${AUTH_TOKEN}`) {
         json(res, 401, { error: "unauthorized", success: false });
         return;
@@ -313,6 +316,19 @@ function createMockCloud(state: MockCloudState): Server {
             status: "queued",
             message: "Agent resume enqueued",
           },
+        });
+        return;
+      }
+      const job = /^\/api\/v1\/jobs\/job-(.+)$/.exec(path);
+      if (job && method === "GET") {
+        const agent = state.agents.get(job[1]);
+        if (!agent?.resumeRequested) {
+          json(res, 404, { success: false, error: "unknown resume job" });
+          return;
+        }
+        json(res, 200, {
+          success: true,
+          data: { id: `job-${agent.id}`, status: "completed" },
         });
         return;
       }
--- a/packages/ui/src/api/client-cloud-wake-typed-errors.test.ts
+++ b/packages/ui/src/api/client-cloud-wake-typed-errors.test.ts
@@ -68,6 +68,102 @@ function fakeClient(mocks: Record<string, unknown>): ElizaClient {
 
 const FAST = { pollIntervalMs: 1, timeoutMs: 60 };
 
+describe("waitForCloudAgentRunning — resume restoration", () => {
+  it("waits for the accepted restore job before returning a fresh running record", async () => {
+    let restored = false;
+    const getCloudCompatJobStatus = vi
+      .fn()
+      .mockResolvedValueOnce({
+        success: true,
+        data: { status: "in_progress", state: "restoring" },
+      })
+      .mockImplementationOnce(async () => {
+        restored = true;
+        return {
+          success: true,
+          data: { status: "completed", state: "completed" },
+        };
+      });
+    const client = fakeClient({
+      resumeCloudCompatAgent: vi.fn(async () => ({
+        success: true,
+        data: { jobId: "resume-1", status: "pending" },
+      })),
+      getCloudCompatJobStatus,
+      getCloudCompatAgent: vi.fn(async () => ({
+        success: true,
+        data: makeAgent({
+          webUiUrl: restored
+            ? "https://restored.example.test"
+            : "https://not-restored.example.test",
+        }),
+      })),
+    });
+
+    const agent = await waitForCloudAgentRunning(client, {
+      agentId: "agent-1",
+      pollIntervalMs: 1,
+      timeoutMs: 1_000,
+    });
+
+    expect(agent.webUiUrl).toBe("https://restored.example.test");
+    expect(getCloudCompatJobStatus).toHaveBeenCalledWith("resume-1");
+  });
+
+  it("surfaces a failed restore even when the proxy already reports running", async () => {
+    const client = fakeClient({
+      resumeCloudCompatAgent: vi.fn(async () => ({
+        success: true,
+        data: { jobId: "resume-1", status: "pending" },
+      })),
+      getCloudCompatJobStatus: vi.fn(async () => ({
+        success: true,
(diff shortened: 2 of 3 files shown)


ss251 (approved):
Reviewed head 377aefe155e457b1f4f1200eb610bf9ef2e2b9ef. Verified by reading the diff and the surrounding `waitForCloudAgentRunning` control flow it modifies.

The risky hunk is `packages/ui/src/api/client-cloud.ts:4240`, the shared-deadline clamp `timeoutMs: Math.max(1, timeoutMs - (Date.now() - startedAt))`. I traced the branch: when the resume envelope carries no `jobId` (the "already in progress" idempotent case), the whole `waitForCloudProvisionJob` block is skipped and flow reaches the bounded status poll, matching the doc comment; cancellation is threaded via `options.signal` into the job wait, which the new third test covers. The one edge: if the deadline is already spent when a `jobId` exists, the clamp passes `1` ms so the job wait throws a timeout error rather than falling back to the status poll — arguably correct (deadline is shared), but it converts a "job accepted, status unknown" case into a typed timeout instead of tolerating one last poll. Non-blocking: the new mock route `/api/v1/jobs/job-(.+)` looks up `state.agents` by the *job* id suffix, so it only matches when job id equals agent id; test-only.

Recommendation: merge

— [free-9router-lane-ss251]

attentionhead (changes requested):
Reviewed at `377aefe155e457b1f4f1200eb610bf9ef2e2b9ef` (base `main`), 3 files, +131/-11.

**Promotion is faithful.** This head is exactly the merge commit of #31137 into `staging` (`gh pr view 31137 --json mergeCommit` → `377aefe155e457b1f4f1200eb610bf9ef2e2b9ef`), and the diffstat matches the two upstream steps file-for-file: `client-cloud.ts` +18/-10, `client-cloud-connect-mock-cloud.test.ts` +17/-1, `client-cloud-wake-typed-errors.test.ts` +96/-0. Same bytes, not a re-application — I verified that identity before writing this rather than inferring it from the title. No promotion-specific defect, and the release mechanics are clean across all three hops (#31136 → #31137 → this).

The underlying change remains good work, and I said so on both prior steps. `startedAt` is captured once so the inserted job wait and the existing status poll share a single deadline instead of doubling it to 24 minutes; cancellation is preserved through both loops via `signal?.throwIfAborted()` at entry and each tick; the mock-cloud change is load-bearing rather than decorative, since the mock's resume handler already returned `jobId: job-${agent.id}` beforehand; and the three new tests form a positive plus its two negatives on the seam being added. The premise is also right — the worker publishes `running` early to enable proxy reachability during restoration, so status alone cannot establish readiness.

## Blocking: this promotes an unaddressed finding to production

I filed this on #31136 (11:05:06Z) and again on #31137. #31136 merged four minutes later and #31137 twenty-one seconds after opening, both with the code unchanged. This is the last hop, so the same finding now reaches `main`:

`CLOUD_WAKE_TRANSIENT_STATUSES` is `new Set([408, 429, 500, 502, 504])` (client-cloud.ts:3976), and `nonTransientWakeFailure` returns terminal details for **any** status outside that set. **404 and 403 are therefore terminal.** Inside the newly inserted `waitForCloudProvisionJob`, a job read answering 404 — a reaped job row, a retention policy that does not keep completed resume-type jobs, or a replica that has not yet observed the row — throws `CloudAgentWakeError{phase:"provision-job"}` and aborts `waitForCloudAgentRunning`. Control never reaches the status poll below.

What makes this a defect rather than a preference is that **the same function already decided this exact question in the opposite direction, one block earlier, and documents it**:

```ts
// error-policy:J4 a transport failure without a terminal HTTP status is
// an idempotent wake nudge lost in transit; the bounded status poll is
// the authority and can still observe the agent becoming ready.
return null;
```

The resume POST tolerates a lost response and falls through to the poll. The new job wait does not extend that tolerance to an unreadable job row. So a resume whose response never arrived recovers, while a resume whose response *did* arrive and whose job row is then briefly unreadable is fatal. That is an inherited, repo-conventional policy (`error-policy:J4` appears in other modules too) that the new call does not honor — not a hypothetical about 404 frequency.

**Consequence.** By this change's own staging evidence the pre-fix failure was degraded but recoverable: the UI opened empty during restoration and a reload recovered all 26 original messages. The post-fix failure on an unreadable job row is "Provisioning your cloud agent failed" with no reload path, on an agent that is healthy and restoring normally. That trades a recoverable empty chat for an unreachable agent, and it fires on control-plane behavior the client does not control. On `main` it is user-facing.

**Smallest repair, unchanged across all three steps:** treat an unreadable job row as a fall-through into the existing bounded status poll rather than a terminal failure, matching the `error-policy:J4` rationale the function already applies. Either catch the `phase:"provision-job"` error whose underlying status is 404/403 and continue into the poll, or pass a flag into `waitForCloudProvisionJob` downgrading a not-found job to an unknown tick. A genuinely failed restore still surfaces via `job.status === "failed"` — the case this change actually targets — so the guard does not weaken the fix.

## Non-blocking

1. **The new seam has three exits; only two are tested.** Completed, `failed`, and cancellation are covered. The exit where the job read *throws* is untested, and it is precisely the one carrying the classification above. A test driving `httpError(404, …)` from `getCloudCompatJobStatus` would pin whichever behavior you settle on; the existing file has a 404 `agent_not_found` case at line 271 to model it on.
2. **Progress copy is provisioning-flavored for a restart.** During the restore wait `onProgress` reports `"provisioning"`, and a timeout in that window says the *provisioning job* is still pending. For a user who pressed Start on an existing stopped agent that reads as if a new agent is being created.

I recognize this is a release-train PR and that blocking the promotion does not by itself fix anything — the code is already in `develop` and `staging`. If the intent is to ship the readiness fix now and repair the error classification in a follow-up, say so and I will not stand in the way. I am requesting changes because the finding has now passed three review gates without a response, and `main` is the last place it can be caught cheaply.

PR #31141, opened 2026-09-12
fix(app): recover rejected Dedicated credentials on reload

### LINKED ISSUE #31140: fix(app): recover Dedicated proxy credential rejection on reload
Reloading the hosted Dedicated app can strand a signed-in user at “Open this agent from Eliza Cloud.” This reproduced in the production app at a 390 × 844 browser viewport immediately after the same tab loaded the existing conversation successfully. The public client entry was `index-2QcDSCwj.js` from main `6dbb9da476619613a6887d3909070e7326387738`.

The account endpoints still returned HTTP 200. Dedicated `/api/auth/me` instead returned HTTP 401 with `{ success: false, code: "cloud_auth_rejected", error: "Cloud authentication failed" }`. The auth client discards that proxy code and produces `server_error`; the existing recovery hook requires `remote_auth_required`, so no authenticated re-pair is attempted.

Scope: recognize the authoritative Dedicated proxy rejection at the auth-client boundary and let the existing Cloud-authorized recovery flow run. Preserve unrelated/self-hosted failures, target-switch guards, and all server authorization checks. No credential bypass, account reset, infrastructure change, or conversation deletion.

Acceptance: reproduce the response contract before the fix; prove Dedicated recovery is selected without treating a 401 as authenticated; preserve generic/self-hosted behavior; verify a real hosted reload restores the same complete conversation at desktop and mobile widths.

NubsCarson:
Fixes #31140.

A fresh hosted Dedicated tab can load its conversation and then show “Open this agent from Eliza Cloud” on ordinary reload, even while the Cloud account session remains valid. The Dedicated proxy returns `401 cloud_auth_rejected` before the runtime can provide its richer auth response. The client was translating that to `server_error`, which prevented the existing recovery hook from running.

Recognize that exact rejection only for Dedicated Cloud targets and expose the existing `remote_auth_required` result. The app stays unauthenticated while its current Cloud session authorizes re-pairing. Existing explicit runtime reasons, unrelated hosts/error codes, target-switch guards and server authorization remain unchanged. No rendered component, styling, worker, schema, infrastructure or agent-image changes.

Risk: auth recovery routing. The fix does not accept the rejected credential; it enables the existing bounded, server-authorized exchange. Rollback is the prior frontend release.

Validation at `6df2259e6dae9591b2aa3e37c2c58db132246b52`, based on current develop `a4654aa68d64f6ca8405e1dbc7bd4b821af364ea`:

- The exact production response regression fails before the fix, passes afterward. Generic and unrelated rejection contracts remain outside Cloud recovery.
- 172 tests across six auth-client, auth-state, recovery and host-classification suites pass under Node24.15.0/Bun1.3.14. Biome and diff checks pass.
- Frozen dependency verification reported no changes; install scripts were omitted because the dependency graph is unchanged and already built.
- Root `bun run verify` exited0, including workspace typecheck/lint and repository audits. The affected chat/Settings audit passed8/8 across four viewports; OCR verified8/8, zero broken or needs-eyeball captures. All eight images were manually inspected. These are shell/layout fixtures, not a substitute for hosted auth acceptance.
- The initial broad visual run was stopped under the owner-requested focused scope after a first mobile startup timeout (11 passed,1 failed,1 interrupted,213 not run). That exact mobile case and every affected layout passed the focused rerun after the existing Vite child restarted. The initial failure artifacts are preserved; this is not a claim that the broad run passed.
- The ordinary local Bun app with this source displays all33 retained messages and the prior real four-fact reply. The repaired production reload remains a release acceptance gate.

- [x] N/A - no rendered diff. The real mobile failure was visually inspected; the regression receipt records the exact response and visible outcome.

- [x] after-screenshots

- [ ] N/A - owner-authorized accelerated evidence scope for this auth-client repair; no separate MP4. Real hosted reload verification remains required after deployment.

- [x] N/A - no backend source or runtime change. Observed backend contract is recorded under frontend diagnostics.

- [x] [31141-frontend-logs.log](https://github.com/elizaOS/eliza/releases/download/pr-evidence-12/31141-frontend-logs.log)

- [x] N/A - no model, prompt or inference change. Prior real production restart recalled Harbor-917/teal; local33-message history includes the four-fact recall.

- [x] [31141-domain-artifacts.json](https://github.com/elizaOS/eliza/releases/download/pr-evidence-12/31141-domain-artifacts.json)

- [x] [31141-ocr-review.json](https://github.com/elizaOS/eliza/releases/download/pr-evidence-12/31141-ocr-review.json)

Release through normal develop/staging promotion and the canonical exact-tree staging certificate before production. This frontend-only fix does not require restarting the provisioning worker or Dedicated agent. Shared/Telegram and first-ever signup remain separate acceptance items.

Additional reviewed fixture captures:

Desktop chat shell
Desktop Settings
Mobile Settings

Hosted CI34694198691 remains in progress. Completed local checks above authorize the ordinary owner-approved merge; hosted CI is not represented as passed.

### DIFF
--- a/packages/ui/src/api/auth-client.test.ts
+++ b/packages/ui/src/api/auth-client.test.ts
@@ -20,7 +20,10 @@ import { invokeDesktopBridgeRequest } from "../bridge/electrobun-rpc";
 import { isElectrobunRuntime } from "../bridge/electrobun-runtime";
 import { getBootConfig } from "../config/boot-config";
 import { clearSharedCloudAccountBinding } from "../state/shared-cloud-account-binding";
-import { isManagedCloudSharedAgentBase } from "../utils/cloud-agent-base";
+import {
+  isDedicatedCloudAgentBase,
+  isManagedCloudSharedAgentBase,
+} from "../utils/cloud-agent-base";
 import {
   authChangePassword,
   authListSessions,
@@ -57,6 +60,7 @@ vi.mock("../state/shared-cloud-account-binding", () => ({
   clearSharedCloudAccountBinding: vi.fn(),
 }));
 vi.mock("../utils/cloud-agent-base", () => ({
+  isDedicatedCloudAgentBase: vi.fn(() => false),
   isManagedCloudSharedAgentBase: vi.fn(),
 }));
 vi.mock("./client-cloud", () => ({
@@ -434,6 +438,7 @@ describe("authMe over plain HTTP boundaries", () => {
   beforeEach(() => {
     getBootConfigMock.mockReturnValue({ branding: {} });
     fetchWithCsrfMock.mockReset();
+    vi.mocked(isDedicatedCloudAgentBase).mockReturnValue(false);
     isManagedCloudSharedAgentBaseMock.mockReturnValue(false);
     isDesktopExternalApiBaseUrlMock.mockReturnValue(false);
     invokeDesktopBridgeRequestMock.mockReset();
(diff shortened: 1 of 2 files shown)


attentionhead (approved):
Verified at `6df2259e6dae9591b2aa3e37c2c58db132246b52` (base `develop`), by fetching `pull/31141/head` and reading the files at that tree rather than a local checkout.

## The production change is correct, and the bypass it introduces is the right one

The new early return sits above the existing 401 ladder, so the question that matters is what it *skips*. Enumerated at head:

- Before this PR, a Dedicated `401 {code:"cloud_auth_rejected"}` with no `reason` produced `result.reason = "server_error"` and `access = undefined`.
- Line 593's pairing branch requires `reason === "remote_auth_required"` → not taken.
- Line 602's `result.reason !== "server_error" || result.access` → both false → does **not** return.
- So control reached `resolvePairingFallback(requestBase)` at line 607.

That fallback probe is precisely what the new return bypasses, and the in-code comment states why: the same bearer the edge just rejected cannot query its protected status route, so the probe can only fail. Skipping it is correct, not a shortcut. The positive test pins this from the other side with `expect(fetchWithCsrfMock).toHaveBeenCalledTimes(1)` — a second call would mean the fallback still ran.

Three load-bearing facts I checked at head rather than taking from the description:

- `isDedicatedCloudAgentBase` is genuinely exported (`packages/ui/src/utils/cloud-agent-base.ts:293`). This mattered because the test mocks the whole module, so an absent real export would still give a green suite; a control grep on the same bytes (18 exports in that file) proves the lookup was live.
- `remote_auth_required` is a pre-existing consumed state, not a new string: `useAgentSessionRecovery`, `LoginView`, and `SecuritySettingsSection` all branch on it. The claim that this "exposes the existing recovery result" holds.
- `requestBase` in the new guard is the same value used to build the `/api/auth/me` URL, so the dedicated-host test and the request that was rejected cannot disagree.

The guard stays narrow: it fires only when the runtime supplied no `reason` of its own, so an explicit runtime contract still wins.

## Non-blocking: the third conjunct of the guard is unpinned

The `it.each` negative covers two of the three ways the guard can fail — non-dedicated host with the right code, and dedicated host with the wrong code. The third conjunct, `!body.reason`, has no test.

Concretely: a `401` from a Dedicated base carrying **both** `code:"cloud_auth_rejected"` **and** an explicit `reason:"remote_auth_required"` plus `access.mode === "remote"`. Today the guard declines it and the request reaches line 593, which calls `resolvePairingFallback` and can return a pairing result. Delete `!body.reason &&` and that case instead early-returns a bare `remote_auth_required` with no `access` and no pairing attempt — a real behavior change that the current suite would not catch.

That conjunct is doing deliberate work (it preserves the richer runtime contract's priority), which is exactly why it is worth a test. A third `it.each` row asserting the pairing path still runs when `reason` is present would close it.

This is a coverage gap in a new test file, not a defect in the shipped behavior, so it does not block.

## One question for the release train, not for this PR

The guard keys on the literal `cloud_auth_rejected`. If that string is produced by the Dedicated proxy rather than by code in this repo, it is an unversioned cross-service contract: an edge-side rename would silently restore the original bug, and the test suite — which constructs the string itself — would stay green. If there is a shared constant for it, importing that instead of the literal would make the coupling visible. Not a reason to hold this fix.

Approving: the change is minimal, the bypass is justified and test-pinned, and the recovery state it exposes is pre-existing and consumed.

PR #31146, opened 2026-09-12
test(plugin-pty): assert the cloud model defaults through the shared constants (#31143)

### LINKED ISSUE #31143: test(plugin-pty): pty-routes suite pins the retired gemma-4-31b Cerebras default and is red on develop
## Defect

`plugins/plugin-pty/test/pty-routes.test.ts` is red on `develop` (tip `ce68dbca425412131bd2d042b2cc4ffa6c8a61f7`, also at `a4654aa`): three cases fail in the `tests / Plugin Tests (3/4)` and `canonical / Tests (plugins)` lanes of the develop validation workflow (run 34694307489).

```text
cd plugins/plugin-pty && bunx vitest run test/pty-routes.test.ts
 × POST /api/pty/sessions > spawns an interactive eliza-code session and returns its id
 × PTY interactive gate + model fallbacks (regression edges) > applies the FAST tier env fallback without touching the SMART default
 × PTY interactive gate + model fallbacks (regression edges) > applies the SMART tier env fallback without touching the FAST default
AssertionError: expected 'qwen-3.8-27b' to be 'gemma-4-31b'
 Tests  3 failed | 40 passed (43)
```

## Cause

The route derives the spawn env from `ELIZA_CLOUD_FAST_MODEL` / `ELIZA_CLOUD_SMART_MODEL` in `plugins/plugin-pty/lib/eliza-code-spec.ts`, which are `DEFAULT_CEREBRAS_TEXT_MODEL` from `@elizaos/core`. That constant moved from `gemma-4-31b` to `qwen-3.8-27b` in `4782de2d705a` (feat(cerebras): adopt Qwen 3.8, 2026-09-04). The test still pins the literal `"gemma-4-31b"` at lines 210, 820, 836 and 837, so it asserts a stale model id rather than the contract (a FAST pin must not touch the SMART default and vice versa).

## Fix

Assert the exported constants instead of a model literal, so the suite follows the canonical default and keeps pinning the fallback isolation. Test-only.

hermesagent270-commits:
# Relates to

Fixes #31143

- [x] `bun install --frozen-lockfile` is current for that base (no lockfile drift). Root `bun run verify` is running on a stacked branch holding this and the two sibling develop-red repairs; the result will be posted here.
- [x] A reviewer can confirm the change from the focused test run below without reading the code.

# Sync with develop

# Risks

None to runtime behavior: test-only, one file. The risk surface is the test's own fidelity, addressed in **Background**.

# Background

## What does this PR do?

The pty spawn-route tests pinned the Cerebras default model as the literal `"gemma-4-31b"` in four assertions. Production derives the spawn env from `ELIZA_CLOUD_FAST_MODEL` / `ELIZA_CLOUD_SMART_MODEL` (`lib/eliza-code-spec.ts`), which are `DEFAULT_CEREBRAS_TEXT_MODEL` from `@elizaos/core`, now `qwen-3.8-27b`. The tests import those two constants and assert them, so they keep pinning the contract (a FAST pin must not touch the SMART default and vice versa) instead of a model id that changes with the provider policy.

Test repair for a suite that is red at the `develop` tip (see the issue for the failing lanes in develop validation run 34694307489).

# Testing

`plugins/plugin-pty/test/pty-routes.test.ts`: the new import and the four assertions at the `OPENAI_SMALL_MODEL` / `OPENAI_MEDIUM_MODEL` / `OPENAI_LARGE_MODEL` reads.

## Detailed testing steps

```text
cd plugins/plugin-pty && bunx vitest run test/pty-routes.test.ts && bunx biome check test/pty-routes.test.ts
```

Before (develop `ce68dbc`):

```text
 × spawns an interactive eliza-code session and returns its id
 × applies the FAST tier env fallback without touching the SMART default
 × applies the SMART tier env fallback without touching the FAST default
AssertionError: expected 'qwen-3.8-27b' to be 'gemma-4-31b'
 Tests  3 failed | 40 passed (43)
```

After (this head):

```text
 Test Files  1 passed (1)
      Tests  43 passed (43)
Checked 1 file. No fixes applied.
```

### DIFF
--- a/plugins/plugin-pty/test/pty-routes.test.ts
+++ b/plugins/plugin-pty/test/pty-routes.test.ts
@@ -8,6 +8,10 @@ import { fileURLToPath } from "node:url";
 import type { IAgentRuntime } from "@elizaos/core";
 import { resetDevCloudEnvAuthorityForTests } from "@elizaos/shared";
 import { afterEach, beforeEach, describe, expect, it } from "vitest";
+import {
+  ELIZA_CLOUD_FAST_MODEL,
+  ELIZA_CLOUD_SMART_MODEL,
+} from "../lib/eliza-code-spec";
 import { ptyRoutes } from "../routes/pty-routes";
 import { PtyService } from "../services/pty-service";
 import { makeFakeSpawn, type SpawnCall } from "./fake-pty";
@@ -207,7 +211,9 @@ describe("POST /api/pty/sessions", () => {
     ]);
     expect(h.calls[0].opts.env?.ELIZA_CODE_CODING_ONLY).toBe("1");
     expect(h.calls[0].opts.env?.OPENAI_API_KEY).toBe("sk-cloud");
-    expect(h.calls[0].opts.env?.OPENAI_SMALL_MODEL).toBe("gemma-4-31b");
+    expect(h.calls[0].opts.env?.OPENAI_SMALL_MODEL).toBe(
+      ELIZA_CLOUD_FAST_MODEL,
+    );
   });
 
   it("403 when interactive spawning is disabled", async () => {
@@ -817,7 +823,7 @@ describe("PTY interactive gate + model fallbacks (regression edges)", () => {
     expect(env?.OPENAI_SMALL_MODEL).toBe("fast-only");
     expect(env?.OPENAI_MEDIUM_MODEL).toBe("fast-only");
     // SMART unset → the cerebras default, not the FAST pin.
-    expect(env?.OPENAI_LARGE_MODEL).toBe("gemma-4-31b");
+    expect(env?.OPENAI_LARGE_MODEL).toBe(ELIZA_CLOUD_SMART_MODEL);
   });
 
   it("applies the SMART tier env fallback without touching the FAST default", async () => {
@@ -833,8 +839,8 @@ describe("PTY interactive gate + model fallbacks (regression edges)", () => {
     expect(res.status).toBe(200);
     const env = h.calls[0].opts.env;
     // FAST unset → the cerebras default, not the SMART pin.
-    expect(env?.OPENAI_SMALL_MODEL).toBe("gemma-4-31b");
-    expect(env?.OPENAI_MEDIUM_MODEL).toBe("gemma-4-31b");
+    expect(env?.OPENAI_SMALL_MODEL).toBe(ELIZA_CLOUD_FAST_MODEL);
+    expect(env?.OPENAI_MEDIUM_MODEL).toBe(ELIZA_CLOUD_FAST_MODEL);
     expect(env?.OPENAI_LARGE_MODEL).toBe("smart-only");
   });
 });


ss251 (approved):
1016522cc5e37d652810780da22aad65dff66ac3

FINDINGS

Test-only diff in plugins/plugin-pty/test/pty-routes.test.ts: four hardcoded "gemma-4-31b" default-model assertions are re-bound to the shared constants ELIZA_CLOUD_FAST_MODEL / ELIZA_CLOUD_SMART_MODEL, imported from lib/eliza-code-spec (:11-14), at the cloud-session path (:214), the SMART-unset fallback (:826) and the FAST-unset fallback (:842-843). Env-provided pins ("fast-only", "smart-only") correctly stay literal — those are test inputs, not defaults — so the change is well scoped and not wholesale renaming.

The re-binding is coherent: each tier's default now asserts its own tier's constant, matching the adjacent "not the FAST pin" / "not the SMART pin" comments. The regression edges still discriminate a route that forwards the shared default from one that hardcodes a stale string, so a tier mix-up in the route would still fail these tests. Coverage is neither removed nor loosened.

One real trade-off, stated as input -> wrong output: input — a later edit to the constants' values in eliza-code-spec; wrong output — these tests keep passing while the shipped default silently changes, because they now pin the constant rather than the value. Value-pinning responsibility moves out of this file to wherever the constants themselves are asserted. I cannot confirm from the diff alone that eliza-code-spec exports these names or that they equal the replaced literal; the suite's compile and run in CI is what catches either mismatch. Minor: if FAST and SMART ever hold the same value, the "not the FAST pin" edge at :826 loses discrimination; an explicit not.toBe against the other tier's constant would keep it.

No production code is touched; no lockfile, CI, hook, or dependency surface appears in the diff.

VERIFICATION

Verified by reading the diff at this head only; the constant definitions in lib/eliza-code-spec were not opened and no suite was run. Live execution blocked: no disposable sandbox on this host, so this is a static review of the diff at the exact head.

Recommendation: merge

— [free-9router-lane-ss251]

attentionhead (approved):
**APPROVE** — verified at head `1016522cc5e37d652810780da22aad65dff66ac3` (base `develop`, +10/-4, one file).

### The substitution is faithful at this head

`plugins/plugin-pty/lib/eliza-code-spec.ts:17-18` defines both constants from one source:

```ts
import { DEFAULT_CEREBRAS_TEXT_MODEL } from "@elizaos/core";
export const ELIZA_CLOUD_FAST_MODEL = DEFAULT_CEREBRAS_TEXT_MODEL;
export const ELIZA_CLOUD_SMART_MODEL = DEFAULT_CEREBRAS_TEXT_MODEL;
```

and the production defaults the tests exercise come from the same two symbols at `:159-160` (`opts.fastModel ?? ELIZA_CLOUD_FAST_MODEL`, `opts.smartModel ?? ELIZA_CLOUD_SMART_MODEL`). The route layer above it (`routes/pty-routes.ts:256,258`) only supplies the request-body / `PTY_ELIZA_CLOUD_*` overrides, so when those are unset the value under assertion really is the constant being imported. The three rewritten expectations therefore describe the same behaviour they did with the literal. After the change, `gemma-4-31b` no longer appears in any `.ts` under `plugins/plugin-pty` — only in `AGENTS.md` and `CLAUDE.md`, which is consistent with the stated goal.

### Non-blocking: the assertions are now self-referential, and nothing else pins the id

This is the deliberate trade the issue asks for, so I am not blocking on it, but it should be on the record rather than implied. Both sides of each `expect` now resolve through `DEFAULT_CEREBRAS_TEXT_MODEL`: if that constant were changed to a wrong or typo'd model id, these tests stay green. The literal was the only thing in this suite that would have failed loudly on such a change. Checking repo-wide, the other consumers are the same shape — `packages/agent/src/runtime/provider-model-defaults.test.ts:164-166,243,286-288,306` and `packages/agent/src/runtime/__tests__/model-env-seed.test.ts:59` all compare an env value to `DEFAULT_CEREBRAS_TEXT_MODEL` rather than to a string. So after this PR no test in the repo pins the actual id; the model default is covered for *plumbing* everywhere and for *value* nowhere. One small value-pin next to the constant's own definition would restore that cheaply, and belongs there rather than in this PTY suite.

### Two comments now claim a distinction the values cannot carry

`ELIZA_CLOUD_FAST_MODEL` and `ELIZA_CLOUD_SMART_MODEL` are the *same value* at this head, so:

```ts
// SMART unset → the cerebras default, not the FAST pin.
expect(env?.OPENAI_LARGE_MODEL).toBe(ELIZA_CLOUD_SMART_MODEL);
```

cannot actually discriminate "the cerebras default" from "the FAST pin" — it would pass either way. This is not a regression (it was equally true when both read `"gemma-4-31b"`), and naming the constants separately is the right forward-looking shape for when the tiers diverge. Worth knowing that the comments currently describe an intent the assertion does not yet enforce; `"fast-only"` / `"smart-only"` in the neighbouring expectations are what actually carry the tier discrimination today.

### Open question

`eliza-code-spec.ts:8` imports `DEFAULT_CEREBRAS_TEXT_MODEL` from `@elizaos/core`, while `packages/agent/src/runtime/provider-model-defaults.test.ts:8` and `__tests__/model-env-seed.test.ts:12` import the same name from `@elizaos/shared`. Is that one constant re-exported through both entry points, or two definitions that happen to agree today? If it is the latter, a change to one would make the PTY defaults and the agent's provider defaults drift with every suite still green — precisely the failure mode this PR is otherwise closing off.

hermesagent270-commits:
Root `bun run verify` (`RUN_TURBO_CONCURRENCY=4`) at `origin/develop` `ce68dbca4254` with this PR and its two sibling develop-red repairs (#31146, #31147, #31148) merged onto one stacked branch (`727dd57d203d`, 3 files / +17 −5 over develop):

```text
[exited with code 0]
```

Log: `/root/eliza-fixes/verify-0912b.log` on the contributor host (about 40 minutes).

hermesagent270-commits:
Retriggering CI: the only failing job (Source static smoke, run 34699100403) ended with `The runner has received a shutdown signal` / exit 143, an infrastructure interruption rather than a result for this change. Closing and reopening to rerun at the same head `1016522cc5e3`.

lalalune:

Retaining this focused test repair. Unlike a standalone constant inventory, these assertions inspect the environment passed through the PTY route to the spawn boundary, and the FAST-only/SMART-only cases retain distinct explicit pins. Using the canonical default removes an obsolete model-name expectation without removing the fallback-routing checks. Current develop still contains the old literal assertions. Keep this one-file repair, publish terminal current-head checks rather than relying on a stacked branch in progress, and preserve the explicit override cases. This review does not certify a live coding provider or approve merge.

PR #30709, opened 2026-09-05
fix(scripts): pin electrobun red-green control to the post-30643 importer

lumix5:
The Electrobun workspace-resolution regression control still expected an importer that moved in #30643, making the Script Tests lane fail despite reporting the correct unresolved import. Update that expected path to the live `ios-runtime-bridge.ts` importer; the control still removes the real path mapping and requires the audit to detect the broken resolution.

Validation at `5b3012daf8b1ae82c365fd4b24d06b9fa26f13d6`:

- Pinned Node 24.15.0: all six tests pass, including the real workspace red/green control.
- Pinned Bun 1.3.14: fresh install passes, including native Metal rebuild and a real 384-dimension embedding probe.
- The local Bun script suite passed five tests; its workspace control exceeded its internal 60-second timeout on the review host. The timeout log is retained alongside the successful Node result.
- Root `bun run verify` passed: 376/376 tasks, zero cached, plus final repository audits.
- All hosted checks passed in [run 33995924184](https://github.com/elizaOS/eliza/actions/runs/33995924184).
- Independently confirmed the pre-fix assertion failure in [Develop Full job 101377925035](https://github.com/elizaOS/eliza/actions/runs/33992655835/job/101377925035).

- [x] N/A - script regression assertion only; no rendered UI changes.

- [x] N/A - script regression assertion only; no rendered UI changes.

- [x] N/A - script regression assertion only; the observable result is the audit test output.

- [x] Backend logs: [full evidence file](https://github.com/lumix5/eliza/releases/download/pr-evidence-30709/30709-evidence-backend-logs.txt) and inline transcript:
```text
CI (before) run 33992655835 job 101377925035 @ develop 08f5532a22:
  (fail) historic app and Electrobun mappings are real red-green controls [12168.88ms]
  AssertionError: The input did not match the regular expression
    /platforms\/electrobun\/tsconfig\.json: unresolved @elizaos\/capacitor-bun-runtime imported by packages\/app-core\/src\/api\/ios-local-agent-transport\.ts/
  Input:
    '@elizaos/electrobun typecheck -> packages/app-core/platforms/electrobun/tsconfig.json: unresolved @elizaos/capacitor-bun-runtime imported by packages/app-core/src/platform/ios-runtime-bridge.ts'
  at packages/scripts/audit-tsconfig-workspace-resolution.test.mjs:322:10
Local repro (before) @ 08f5532a22, bun test 1.3.14 / node 24.15.0 / Linux x64:
[...]

PR #31147, opened 2026-09-12
test(scripts): stop pinning the electrobun red-green control to one importer (#31144)

### LINKED ISSUE #31144: test(scripts): electrobun red-green control pins an importer that moved and is red on develop
## Defect

`packages/scripts/audit-tsconfig-workspace-resolution.test.mjs` case 6, "historic app and Electrobun mappings are real red-green controls", is red on `develop` (tip `ce68dbca425412131bd2d042b2cc4ffa6c8a61f7`; `tests / Script Tests (Linux)` and `canonical / Tests (scripts)` in run 34694307489):

```text
node --test packages/scripts/audit-tsconfig-workspace-resolution.test.mjs
not ok 6 - historic app and Electrobun mappings are real red-green controls
  The input did not match the regular expression /platforms\/electrobun\/tsconfig\.json: unresolved @elizaos\/capacitor-bun-runtime imported by packages\/app-core\/src\/platform\/ios-runtime-bridge\.ts/. Input:
  '@elizaos/electrobun typecheck -> packages/app-core/platforms/electrobun/tsconfig.json: unresolved @elizaos/capacitor-bun-runtime imported by packages/ui/src/api/ios-local-agent-transport.ts'
# pass 5
# fail 1
```

## Cause

The control removes the `@elizaos/capacitor-bun-runtime` path mapping from the Electrobun tsconfig and expects the audit to report it unresolved. It also pins the *importer* to `packages/app-core/src/platform/ios-runtime-bridge.ts`. The audit reports the first importer it meets, and since `c745213d6100` (fix: consolidate duplicate systems behind canonical owners, #30229) that is `packages/ui/src/api/ios-local-agent-transport.ts`. The same pin already broke once and was re-pointed in #30709.

## Fix

Match any workspace importer (`imported by packages/…`). The contract the control proves is that the missing mapping is detected, not which file imports the package first. Test-only.

hermesagent270-commits:
# Relates to

Fixes #31144

- [x] `bun install --frozen-lockfile` is current for that base (no lockfile drift). Root `bun run verify` is running on a stacked branch holding this and the two sibling develop-red repairs; the result will be posted here.
- [x] A reviewer can confirm the change from the focused test run below without reading the code.

# Sync with develop

# Risks

None to runtime behavior: test-only, one file. The risk surface is the test's own fidelity, addressed in **Background**.

# Background

## What does this PR do?

The Electrobun red-green control removes the `@elizaos/capacitor-bun-runtime` path mapping and expects the audit to report it unresolved, but it also pinned the reported *importer* to `packages/app-core/src/platform/ios-runtime-bridge.ts`. The audit reports the first importer it meets, which is now `packages/ui/src/api/ios-local-agent-transport.ts`; the same pin was already re-pointed once in #30709. The assertion now matches any workspace importer, because the contract the control proves is that the missing mapping is detected, not which file imports the package first.

Test repair for a suite that is red at the `develop` tip (see the issue for the failing lanes in develop validation run 34694307489).

# Testing

`packages/scripts/audit-tsconfig-workspace-resolution.test.mjs`, the final `assert.match` in "historic app and Electrobun mappings are real red-green controls".

## Detailed testing steps

```text
node --test packages/scripts/audit-tsconfig-workspace-resolution.test.mjs && bunx biome check packages/scripts/audit-tsconfig-workspace-resolution.test.mjs
```

Before (develop `ce68dbc`):

```text
not ok 6 - historic app and Electrobun mappings are real red-green controls
  The input did not match the regular expression /platforms\/electrobun\/tsconfig\.json: unresolved @elizaos\/capacitor-bun-runtime imported by packages\/app-core\/src\/platform\/ios-runtime-bridge\.ts/. Input:
  '@elizaos/electrobun typecheck -> packages/app-core/platforms/electrobun/tsconfig.json: unresolved @elizaos/capacitor-bun-runtime imported by packages/ui/src/api/ios-local-agent-transport.ts'
# pass 5
# fail 1
```

After (this head):

```text
# pass 6
# fail 0
Checked 1 file in 19ms. No fixes applied.
```

### DIFF
--- a/packages/scripts/audit-tsconfig-workspace-resolution.test.mjs
+++ b/packages/scripts/audit-tsconfig-workspace-resolution.test.mjs
@@ -321,6 +321,9 @@ test("historic app and Electrobun mappings are real red-green controls", {
   });
   assert.match(
     electrobunBroken.violations.join("\n"),
-    /platforms\/electrobun\/tsconfig\.json: unresolved @elizaos\/capacitor-bun-runtime imported by packages\/app-core\/src\/platform\/ios-runtime-bridge\.ts/,
+    // The control proves that dropping the path mapping is detected; which
+    // workspace file happens to be the first importer is not part of the
+    // contract and has already moved twice (#30709, ios-local-agent-transport).
+    /platforms\/electrobun\/tsconfig\.json: unresolved @elizaos\/capacitor-bun-runtime imported by packages\/\S+\.tsx?$/m,
   );
 });


ss251 (approved):
aba30cee0d3e1d400b9ee367b3fa47f6870afc48

**Findings**

The change relaxes one regex in `packages/scripts/audit-tsconfig-workspace-resolution.test.mjs` (the "historic app and Electrobun mappings are real red-green controls" test). Input -> output:

- Input: the electrobun red-green control previously asserted the exact importer path `packages/app-core/src/platform/ios-runtime-bridge.ts` in the violation message. Output: the assertion now matches any single importer `packages\/\S+\.tsx?$` for the same unresolved `@elizaos/capacitor-bun-runtime` mapping in `platforms/electrobun/tsconfig.json`.

Assessment: the invariant under test — that dropping the path mapping is *detected at all* — is preserved. What is loosened is *which* workspace file is the first importer, which the PR correctly notes is not part of the contract and has drifted twice already (#30709, ios-local-agent-transport), pinning the test to churn that is not its purpose. The pattern still anchors to the specific tsconfig and the specific unresolved module, and the `$`/`m` combination keeps it to a single importer line, so it cannot silently match a degraded multi-violation blob for a different module. Test weakening check: this is a deliberate, commented, and narrower-in-scope loosening of one brittle assertion, not a removal or a skip; the red-green property (audit fails when the mapping is dropped) remains enforced. No dependency, lockfile, lifecycle-hook, CI-permission, binary, symlink, or payload content in the diff.

One residual nit, non-blocking: the regex does not pin that the matched importer is *app-side* code, so a future unrelated package importing that runtime would still satisfy the control — but that is exactly the flexibility the test is meant to gain, and the module/tsconfig anchors bound the risk.

Verified by reading the diff at this head only; static review, no checkout. Live execution blocked: no disposable sandbox on this host, so the suite was not run; the change is a single test-regex edit where the key risk (weakened red-green detection) was assessed statically from the pattern anchors.

Recommendation: merge

— [free-9router-lane-ss251]

attentionhead (approved):
**APPROVE** — verified at head `aba30cee0d3e1d400b9ee367b3fa47f6870afc48` (base `develop`, +4/-1, one file).

### The loosened control still proves what it is there to prove

The assertion runs against `electrobunBroken.violations.join("\n")`, and each violation is built at `packages/scripts/audit-tsconfig-workspace-resolution.mjs:501`:

```js
`${packageName} typecheck -> ${path.relative(repoRoot, configPath)}: unresolved ${specifier} imported by ${path.relative(repoRoot, sourcePath)}`
```

So the new pattern still pins the two things the red-green control is actually about — that the audit named **this config** (`platforms/electrobun/tsconfig.json`) and **this specifier** (`@elizaos/capacitor-bun-runtime`) after the path mapping was removed by `withoutPath`. Only the importer half is relaxed, and that half is genuinely incidental: it is whichever workspace file the audit happens to walk first, which is exactly the churn the issue describes.

### The new `$` anchor is safe, which is the non-obvious part

The old pattern had no end anchor; the new one ends in `\.tsx?$` under `/m`. That is a *tightening* layered on top of the loosening, so it is worth checking rather than eyeballing: it requires the importer path to be the last thing on its line. The template above ends with `path.relative(repoRoot, sourcePath)` and appends nothing after it, and the entries are joined on `\n`, so each violation's importer path does terminate its line. `\S+` cannot cross the join for the same reason. The anchor holds at this head.

It is worth being explicit that the anchor is load-bearing on that message format: if `:501` ever grows a suffix (a reason, a hint, a count) after the importer path, this control fails with a match error rather than a meaningful signal, and the failure will look like an electrobun regression rather than a formatting change. That is a small, self-announcing failure mode, not a reason to hold the PR — and it is also why keeping the anchor is better than dropping it, since without it `\S+` would silently swallow a suffix.

### Non-blocking observation

`packages\/\S+\.tsx?$` will now match an importer anywhere under `packages/`, including one outside `app-core`. That is the intent, and the config-plus-specifier prefix carries the discrimination, so the control cannot pass against an unrelated violation. If you wanted to keep a little of the original locality without re-pinning a filename, `packages\/app-core\/\S+\.tsx?$` would still tolerate both moves cited in the comment (#30709 and `ios-local-agent-transport`) while preventing the control from quietly starting to match some other package's file. Your call — the current form is defensible as written.

hermesagent270-commits:
Root `bun run verify` (`RUN_TURBO_CONCURRENCY=4`) at `origin/develop` `ce68dbca4254` with this PR and its two sibling develop-red repairs (#31146, #31147, #31148) merged onto one stacked branch (`727dd57d203d`, 3 files / +17 −5 over develop):

```text
[exited with code 0]
```

Log: `/root/eliza-fixes/verify-0912b.log` on the contributor host (about 40 minutes).

lalalune:
I reproduced the develop failure locally: **5 pass / 1 fail**. The missing mapping was correctly reported, but the asserted importer had moved to `packages/ui/src/api/ios-local-agent-transport.ts`. With this patch the same real audit controls passed **6/6**. The unresolved package/config diagnostic remains required; only the incidental first importing file is generalized. Focused before/after output is in the archive. Browser captures are N/A for this script-only assertion repair.

PR #31150, opened 2026-09-12
test(agent): run Bun-runtime script tests inside the batch runner so bun run test:server can go green (#31149)

### LINKED ISSUE #31149: test(agent): chained test script hides the package from the evidence guard, so bun run test:server exits 3 on develop
## Defect

`bun run test:server` (the `canonical / Tests (server, agent)` job of the develop validation workflow) exits 3 on `develop` even though every test in the lane passes. At tip `ce68dbca425412131bd2d042b2cc4ffa6c8a61f7` (run 34694307489, job 103555308152):

```text
[eliza-test] PASS @elizaos/agent (packages/agent)#test (2990271ms)
[eliza-test] RESULTS tasks=1 pass=1 fail=0 skip=0 not-run=0 excluded=2 unobserved=1
[eliza-test] EVIDENCE reports=0 tests=0 executed=0 skipped=0 unobserved-tasks=1
[eliza-test] VACUOUS-GREEN GUARD 1 runnable task(s) completed without reconciled evidence that a testcase executed. Unsupported wrappers do not count as semantic work.
error: script "test:server" exited with code 3
```

## Cause

`0805ac62912b` (test(mobile): run packaging regressions in owning package suites, 2026-09-10) changed `packages/agent` `test` from `node scripts/run-vitest-batches.mjs` to

```text
bun run test:mobile-workspace-entry && node scripts/run-vitest-batches.mjs
```

`packages/scripts/run-all-tests.mjs` only attaches JUnit evidence to a `test` script it can classify as a single `bun test`, single `vitest run`, or the single batch wrapper (`structuredEvidenceKind`). A `&&` chain classifies as nothing, so under `--require-work` the task is tallied `unobserved`, no reporter arguments are forwarded, and because `packages/agent` is the only task in the server lane the aggregate `executedTests` is 0 and the vacuous-green guard exits 3. The same chain also makes the package unobserved in root `bun run test` and `bun run test:all`; those lanes only stay green because sibling packages contribute evidence.

The chained file, `scripts/mobile-workspace-entry.test.mjs`, needs the Bun runtime (`Bun.build`) and is not discoverable by the Vitest batch runner, which only walks `.test.[cm]?tsx?` files.

## Fix

Make the batch runner own the package's whole suite: discover `scripts/*.test.mjs` Bun-runtime tests, run each as its own `bun test` batch with `--reporter=junit --reporter-outfile=<fragment>` when evidence is requested, and merge the fragment with the Vitest fragments. Then `test` is the single wrapper command again, `structuredEvidenceKind` classifies it as `vitest`, and the packaging regression is counted as executed work instead of hidden in front of it.

hermesagent270-commits:
# Relates to

Fixes #31149

- [x] `bun install --frozen-lockfile` is current for that base (no lockfile drift). Root `bun run verify` and the exact CI command `bun run test:server` are being run at this head; both results will be posted here.
- [x] A reviewer can confirm the change from the runner evidence and the lane output below without reading the code.

# Sync with develop

# Risks

Low. `packages/agent` test orchestration only: no runtime source changes. The Bun-runtime packaging regression keeps running on every `bun run --cwd packages/agent test`, now inside the runner instead of chained in front of it. If Bun's JUnit reporter ever changed shape, `mergeAgentJunit` rejects the fragment and the package fails loudly rather than publishing partial evidence.

# Background

## What does this PR do?

`packages/agent`'s `test` script was `bun run test:mobile-workspace-entry && node scripts/run-vitest-batches.mjs`. `packages/scripts/run-all-tests.mjs` attaches JUnit evidence only to a `test` script it can classify as a single `bun test`, `vitest run`, or the batch wrapper (`structuredEvidenceKind`). The `&&` chain classified as nothing, so under `--require-work` the task was tallied `unobserved`, no reporter arguments were forwarded, and because `packages/agent` is the only task in the server lane the aggregate executed count was 0: the vacuous-green guard exited 3 on every develop run, with the suite itself passing.

The batch runner now discovers `scripts/*.test.mjs` (they need `Bun.build`, so a Vitest worker cannot host them), runs each as its own `bun test` batch with `--reporter=junit --reporter-outfile=<fragment>` when evidence is requested, and merges those fragments with the Vitest fragments into the one reconciled report. The `test` script is the bare wrapper again, so the orchestrator classifies it as `vitest` and forwards the reporter arguments; the packaging regression is counted as executed work. `test:mobile-workspace-entry` stays as the focused alias.

Runner changes: `isBunRuntimeTest`, `createRunnerBatches` (Vitest files fill batches of `AGENT_TEST_BATCH_SIZE`; every Bun test is its own batch), `createBunTestInvocation`, and `runBatch` picking the invocation by batch kind. Selection, interruption, and failure handling are unchanged and still covered by the fake-spawn harness, which now records the runner of each spawn.

Test infrastructure repair for a lane that cannot go green on `develop`.

# Testing

`packages/agent/scripts/run-vitest-batches.mjs` (`walk`, `createRunnerBatches`, `createBunTestInvocation`, `runBatch`) and the four new cases in `run-vitest-batches.test.ts`.

## Detailed testing steps

```text
cd packages/agent
bunx vitest run scripts/run-vitest-batches.test.ts
 Tests  35 passed (35)      # 31 on develop + 4 new

# Real runner, evidence requested, one Bun-runtime file and one Vitest file:
node scripts/run-vitest-batches.mjs --reporter=default --reporter=junit --outputFile.junit=/tmp/agent-evidence.xml \
  scripts/mobile-workspace-entry.test.mjs src/build-config.test.ts
[agent-test] 2 file(s), 2 isolated batch(es), concurrency 2
[agent-test] passed 2 file(s) in 2.2s
# merged report root: <testsuites tests="10" failures="0" errors="0" skipped="0">
# suites present: scripts/mobile-workspace-entry.test.mjs (9 cases, bun) and src/build-config.test.ts (1 case, vitest)
# packages/scripts/lib/junit-summary.mjs parseJunitSummary → {"tests":10,"failures":0,"errors":0,"skipped":0,"executedTests":10}

bunx biome check scripts/run-vitest-batches.mjs scripts/run-vitest-batches.test.ts
```

Before (develop `ce68dbc`, run 34694307489 job 103555308152):

```text
[eliza-test] PASS @elizaos/agent (packages/agent)#test (2990271ms)
[eliza-test] RESULTS tasks=1 pass=1 fail=0 skip=0 not-run=0 excluded=2 unobserved=1
[eliza-test] EVIDENCE reports=0 tests=0 executed=0 skipped=0 unobserved-tasks=1
[eliza-test] VACUOUS-GREEN GUARD 1 runnable task(s) completed without reconciled evidence that a testcase executed.
error: script "test:server" exited with code 3
```

After: `bun run test:server` at this head is running locally; its `RESULTS` / `EVIDENCE` lines will be posted as a comment.

### DIFF
--- a/packages/agent/package.json
+++ b/packages/agent/package.json
@@ -50,7 +50,7 @@
     "format:check": "bunx @biomejs/biome format src",
     "pack:dry-run": "cd dist && npm pack --dry-run",
     "smoke:isolated": "node scripts/isolated-package-smoke.mjs",
-    "test": "bun run test:mobile-workspace-entry && node scripts/run-vitest-batches.mjs",
+    "test": "node scripts/run-vitest-batches.mjs",
     "test:mobile-workspace-entry": "bun test scripts/mobile-workspace-entry.test.mjs",
     "test:e2e": "RUN_CRASH_RESTART_E2E=1 vitest run --config vitest.config.ts test/crash-restart-supervisor.test.ts",
     "test:integration": "cd ../.. && bunx vitest run --config packages/scripts/vitest/integration.config.ts --coverage.enabled=false packages/agent/",
--- a/packages/agent/scripts/run-vitest-batches.mjs
+++ b/packages/agent/scripts/run-vitest-batches.mjs
@@ -1,9 +1,13 @@
 /**
- * Runs the agent Vitest suite in bounded parallel, process-isolated batches.
- * The file selection mirrors vitest.config.ts while one-file batches prevent
- * leaked module state and open handles from crossing test boundaries.
- * Positional arguments select exact eligible files; interruption stops queued work.
- * Requested JUnit evidence includes every batch and is reconciled before publication.
+ * Runs the agent test suite in bounded parallel, process-isolated batches.
+ * Vitest files are selected as vitest.config.ts does, one file per batch so
+ * leaked module state and open handles cannot cross test boundaries; the
+ * Bun-runtime `scripts/*.test.mjs` regressions need `Bun.build`, so each runs
+ * alone through `bun test`. Positional arguments select exact eligible files
+ * of either kind; interruption stops queued work.
+ * Requested JUnit evidence includes every batch of both kinds and is reconciled
+ * before publication, which is what lets the package `test` script stay a
+ * single wrapper command the orchestrator's evidence guard can classify.
  */
 import { spawn } from "node:child_process";
 import {
@@ -26,6 +30,18 @@ const packageRoot = path.resolve(
   "..",
 );
 const roots = ["src", "test", "scripts"];
+
+// Bun-runtime tests sit beside the scripts they exercise and need `Bun.build`,
+// which a Vitest worker does not provide, so they never join a Vitest batch.
+const BUN_TEST_PATTERN = /\.test\.mjs$/;
+const BUN_TEST_ROOT = "scripts/";
+
+export function isBunRuntimeTest(relativePath) {
+  return (
+    relativePath.startsWith(BUN_TEST_ROOT) &&
+    BUN_TEST_PATTERN.test(relativePath)
+  );
+}
 
 const excludedPatterns = [
   /\.e2e\.test\.[cm]?tsx?$/,
@@ -54,7 +70,9 @@ function walk(relativeDir, out) {
       continue;
     }
     if (!stat.isFile()) continue;
-    if (!/\.test\.[cm]?tsx?$/.test(entry)) continue;
+    if (!/\.test\.[cm]?tsx?$/.test(entry) && !isBunRuntimeTest(relativePath)) {
+      continue;
+    }
     if (excludedPatterns.some((pattern) => pattern.test(relativePath))) {
       continue;
     }
@@ -101,6 +119,22 @@ export function createBatches(files, batchSize) {
     batches.push(files.slice(start, start + batchSize));
   }
   return batches;
+}
+
+/**
+ * Vitest files fill batches of `batchSize`; every Bun-runtime test is its own
+ * batch because it runs under a different executable and reporter.
+ */
+export function createRunnerBatches(files, batchSize) {
+  const vitestFiles = files.filter((file) => !isBunRuntimeTest(file));
+  const bunFiles = files.filter((file) => isBunRuntimeTest(file));
+  return [
+    ...createBatches(vitestFiles, batchSize).map((batch) => ({
+      runner: "vitest",
+      files: batch,
+    })),
+    ...bunFiles.map((file) => ({ runner: "bun", files: [file] })),
+  ];
 }
 
 function isFile(filePath) {
@@ -235,6 +269,19 @@ export function createVitestInvocation(bunExecutable, batch, fragmentPath) {
   };
 }
 
+export function createBunTestInvocation(bunExecutable, batch, fragmentPath) {
+  return {
+    command: bunExecutable,
+    args: [
+      "test",
+      ...(fragmentPath
+        ? ["--reporter=junit", `--reporter-outfile=${fragmentPath}`]
+        : []),
+      ...batch,
+    ],
+  };
+}
+
 function terminate(child, signal = "SIGTERM") {
   if (!child.pid) return;
   if (process.platform === "win32") {
@@ -252,11 +299,10 @@ function terminate(child, signal = "SIGTERM") {
 function runBatch(batch, nodeOptions, active, bunExecutable, fragmentPath) {
   return new Promise((resolve) => {
     const startedAt = performance.now();
-    const invocation = createVitestInvocation(
-      bunExecutable,
-      batch,
-      fragmentPath,
-    );
+    const invocation =
+      batch.runner === "bun"
+        ? createBunTestInvocation(bunExecutable, batch.files, fragmentPath)
+        : createVitestInvocation(bunExecutable, batch.files, fragmentPath);
     const child = spawn(invocation.command, invocation.args, {
       cwd: packageRoot,
       detached: process.platform !== "win32",
@@ -327,7 +373,7 @@ async function main() {
   const nodeOptions = inheritedNodeOptions.includes("--max-old-space-size")
     ? inheritedNodeOptions
     : `${inheritedNodeOptions} --max-old-space-size=8192`.trim();
-  const batches = createBatches(files, batchSize);
+  const batches = createRunnerBatches(files, batchSize);
   const fragmentDirectory = reporterOutfile
     ? mkdtempSync(path.join(tmpdir(), "eliza-agent-junit-"))
     : undefined;
@@ -370,7 +416,7 @@ async function main() {
         );
         completed += 1;
         if (verbose || result.status !== 0) {
-          const label = `[agent-test] batch ${index + 1}/${batches.length}: ${batch.join(", ")}`;
+          const label = `[agent-test] batch ${index + 1}/${batches.length}: ${batch.files.join(", ")}`;
           process.stdout.write(`${label}\n${result.stdout}`);
           process.stderr.write(result.stderr);
         } else if (completed % 25 === 0 || completed === batches.length) {
@@ -395,7 +441,7 @@ async function main() {
       for (const failure of failures) {
         if (failure.error) {
           console.error(
-            `[agent-test] ${failure.batch.join(", ")}: ${failure.error instanceof Error ? failure.error.message : String(failure.error)}`,
+            `[agent-test] ${failure.batch.files.join(", ")}: ${failure.error instanceof Error ? failure.error.message : String(failure.error)}`,
           );
         }
       }
--- a/packages/agent/scripts/run-vitest-batches.test.ts
+++ b/packages/agent/scripts/run-vitest-batches.test.ts
@@ -4,6 +4,7 @@ import {
   existsSync,
   mkdirSync,
   mkdtempSync,
+  readFileSync,
   rmSync,
   writeFileSync,
 } from "node:fs";
@@ -13,7 +14,10 @@ import { fileURLToPath } from "node:url";
 import { afterAll, describe, expect, test } from "vitest";
 import {
   createBatches,
+  createBunTestInvocation,
+  createRunnerBatches,
   createVitestInvocation,
+  isBunRuntimeTest,
   mergeAgentJunit,
   parseAgentTestArgs,
   positiveInteger,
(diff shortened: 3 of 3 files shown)


attentionhead (approved):
**APPROVE** — verified at head `c24a8c213426b91300c600e476f297f2443f1b6e` (base `develop`, +186/-18, 3 files).

The interesting part of this PR is entirely outside its own diff: un-chaining the `test` script makes the orchestrator start forwarding reporter arguments to `run-vitest-batches.mjs` **for the first time**, so the change is only safe if the receiving parser already accepts exactly what the sender emits. I traced that end to end at this head.

### The classification claim is exact

`packages/scripts/run-all-tests.mjs:817`:

```js
function isSingleVitestBatchWrapperCommand(command) {
  return /^node\s+scripts\/run-vitest-batches\.mjs$/.test(
    stripLeadingEnvAssignments(command),
  );
}
```

The `$` anchor is what the old value tripped on — `bun run test:mobile-workspace-entry && node scripts/run-vitest-batches.mjs` cannot match an anchored pattern with the wrapper at the tail, and it independently fails the `[;&|]` guard in every sibling classifier (`:798`, `:809`, `:831`). So `structuredEvidenceKind` fell through to `return null`, `evidence` stayed null at `:1200`, `buildForwardedScriptArgs` returned `[]`, and the package published no JUnit at all. With `packages/agent` the only task in the server lane, the aggregate executed count is 0 and `--require-work` exits 3 on a passing suite. The described failure mode is real and the diagnosis in the PR body is precise, not approximate.

The new bare command matches that regex exactly, so `structuredEvidenceKind` now returns `"vitest"`.

### The forwarding actually round-trips — this is the load-bearing check

`buildForwardedScriptArgs` for `kind === "vitest"` emits exactly:

```js
args.push("--reporter=default", "--reporter=junit", `--outputFile.junit=${evidence.path}`);
```

and these arrive as positional argv. That matters because `selectTestFiles` throws `Unsupported test runner argument` on anything starting with `-`, so a mismatch here would fail the lane on the first CI run rather than degrade quietly. `parseAgentTestArgs` consumes precisely those three forms — `--reporter=default`, `--reporter=junit`, and the `--outputFile.junit=` prefix — before anything reaches `selectTestFiles`, and it fails closed if `--reporter=junit` arrives without an outfile. A pre-existing test at `run-vitest-batches.test.ts:334` already pins that triple. The sender and the receiver agree, and nothing new was needed to make them agree.

### The Bun fragment really does validate

The new merge test asserts a Bun root that declares `tests failures skipped` but no `errors`. That passes only because `parseJunitSummary`'s `reconcileDeclaredCounts` is called with `allowOmittedNonzero = new Set(["errors", "skipped"])` at the `testsuites` level (`packages/scripts/lib/junit-summary.mjs`), and the nested `<testsuite>` — which passes `undefined` for that set — omits `errors` while genuinely containing zero. Bun reports failing tests as `<failure>`, so `counts.errors` stays 0 and the nested path holds. Worth being explicit that the nested allowance is incidental rather than declared: a Bun fragment that ever carried an `<error>` child without an `errors` attribute would throw `JUnit testsuite … omits errors but contains N`. That is the loud failure the PR body promises, not a silent one, so it is the right behaviour — just not currently pinned by a test the way the root-level omission is.

`mergeAgentJunit` then re-parses the assembled document, which is what stops a fragment from smuggling in counts that the testcase tree does not support.

### Non-blocking: a `.test.mjs` outside `scripts/` is discovered by neither runner

`walk` keeps a file when `/\.test\.[cm]?tsx?$/.test(entry) || isBunRuntimeTest(relativePath)`. The first alternative cannot match `.mjs` (`[cm]?tsx?` requires a literal `ts`), and the second requires the `scripts/` prefix. So `src/entry.test.mjs` or `test/entry.test.mjs` is silently not executed — not by Vitest, not by Bun, and with no error. The new test pins exactly that as intended:

```ts
for (const file of ["scripts/mobile-workspace-entry.mjs", "scripts/run-vitest-batches.test.ts",
                    "src/entry.test.mjs", "test/entry.test.mjs"]) {
  expect(isBunRuntimeTest(file)).toBe(false);
}
```

This is not a regression — those files were equally invisible before — and the narrow root is clearly deliberate, since `Bun.build` is what justifies the separate runner. But the test now documents the boundary without documenting the consequence, and "your test file is ignored" is a failure that looks like a green run. Given this PR exists precisely because a lane went vacuously green, a thrown error for a `*.test.mjs` found outside `scripts/` would be in the spirit of the fix. Your call; I would not hold the PR for it.

### Non-blocking: the focused alias now has no caller

`test:mobile-workspace-entry` survives at `package.json:54`, but after this change the repo references it nowhere else — the removed `&&` was its only caller. That is fine as a manual convenience and the PR body says so. The small risk is drift in one direction only: if `scripts/mobile-workspace-entry.test.mjs` is ever renamed, the batch runner keeps discovering it by pattern while the alias silently points at a path that no longer exists, and nothing fails until someone runs the alias by hand.

### Minor praise

Recording `runners` in the fake-spawn receipt is what makes these tests meaningful rather than shape-only — `expect(result.receipt.runners).toEqual(["x vitest run", "bun test"])` with `batchSize: 2` proves the Bun file stayed isolated for the right reason and not by accident of batching. The failure-path case also confirms the Bun batch is still spawned when a sibling Vitest batch fails, which is the behaviour the old `&&` chain got wrong in the opposite direction.

ss251 (approved):
c24a8c213426b91300c600e476f297f2443f1b6e

**Findings**

1. `packages/agent/scripts/run-vitest-batches.mjs:73-75` — the walker now admits `scripts/*.test.mjs` in addition to the existing `/\.test\.[cm]?tsx?$/` pattern. The non-test sibling `scripts/mobile-workspace-entry.mjs` is correctly excluded (`isBunRuntimeTest` requires the `.test.mjs` suffix AND the `scripts/` root), and the new classification test at `run-vitest-batches.test.ts:355` pins exactly that, including the `src/entry.test.mjs` and `test/entry.test.mjs` negatives. Input -> wrong output would be a Vitest worker picking up a Bun-dependent test and failing on missing `Bun.build`; this change closes that path.

2. `run-vitest-batches.mjs:124-136` — `createRunnerBatches` keeps every Bun-runtime test as a singleton batch under `runner: "bun"`, preserving the one-file isolation guarantee the header documents, and the regression test at line 286 asserts that a batchSize of 2 does not pair the `.mjs` test with a Vitest file. Input -> wrong output: pairing them would run Bun code under the wrong executable and reporter; the test would catch it.

3. `run-vitest-batches.mjs:272-284` — `createBunTestInvocation` builds a shell-free `bun test` argv with JUnit reporter flags only when a fragment path exists, matching the Vitest invocation contract, and is unit-tested for a Windows executable path. The merge path (`mergeAgentJunit`) is exercised with a real Bun-shaped fragment that omits the `errors` attribute, and the merged totals (`tests="3"`) confirm it is handled rather than silently dropped.

4. `packages/agent/package.json:53` — dropping `bun run test:mobile-workspace-entry &&` from `test` is safe only because the batch runner now picks that file up itself; the fake-inventory test (line 54: `["scripts/mobile-workspace-entry.test.mjs", ...]`) plus the runner receipt assertion confirm it is scheduled. Removing the old duplicate invocation eliminates the double-run the old compound command caused.

No widening of CI permissions, no dependency or lockfile changes, no secrets, binaries, or obfuscated payloads in the diff. The change is correctly scoped to the batch runner, its tests, and the package test script.

**Scope note**

Live execution blocked: no disposable sandbox on this host, so this is a static review of the diff at the exact head. The suite changes themselves are the behavioral evidence here and are unit-tested against a faked `spawn`/fs layer, which this review verified by reading, not by running.

Recommendation: merge

— [free-9router-lane-ss251]

hermesagent270-commits:
Exact-lane proof at this head `c24a8c213426`, run locally with the server lane's own orchestrator invocation (the `bun run test:server` script body, filtered to this package's task; `build:core` was current):

```text
TEST_SCRIPT_FILTER='^test$' node packages/scripts/run-all-tests.mjs --lane=server --no-cloud --concurrency=3 --require-work --filter='@elizaos/agent \(packages/agent\)#test'
[eliza-test] START @elizaos/agent (packages/agent)#test
[eliza-test] RESULT {"label":"@elizaos/agent (packages/agent)#test","packageName":"@elizaos/agent","relativeDir":"packages/agent","scriptName":"test","status":"pass","observed":true,"exitCode":0,"signal":null,"timedOut":false,"durationMs":2535133,"counts":{"tests":10644,"executed":10619,"failures":0,"errors":0,"skipped":25}}

[eliza-test] RESULTS tasks=1 pass=1 fail=0 skip=0 not-run=0 excluded=2 unobserved=0
[eliza-test] EVIDENCE reports=1 tests=10644 executed=10619 skipped=25 unobserved-tasks=0
```

Before (develop `ce68dbc`, job 103555308152): `observed:false`, `EVIDENCE reports=0 tests=0 executed=0`, `VACUOUS-GREEN GUARD`, exit 3. After: the batch runner was launched with `--reporter=junit --outputFile.junit=<fragment>` forwarded, the merged report reconciles 10,644 testcases (10,619 executed, 25 skipped; the nine Bun-runtime packaging cases are among the executed ones, contributed by the new `bun test` batch), `unobserved=0`, and the process exited 0. Root `bun run verify` is running next on a stacked branch with the sibling develop-red repairs; result to follow.

hermesagent270-commits:

```text
```

lalalune:

Retaining the server-lane evidence repair. Current agent test still chains the Bun packaging test before the batch wrapper, while this change gives both runners one reconciled evidence path. Complete validation with an actual Bun JUnit fragment plus an actual Vitest fragment through the real merger and root server-lane accounting, not only fake-spawn fixtures. Verify failure, skipped/zero-work and interruption results remain visible, and that scripts/mobile-workspace-entry.test.mjs runs exactly once. Document scripts/*.test.mjs as the runner-selection convention and cover a future script needing a different runtime explicitly rather than assuming its extension proves Bun ownership.

lalalune:
The reported chained-command admission defect is already resolved on develop by commit 61446bb82c9 (PR #31269). `isSingleVitestBatchWrapperCommand` recognizes the optional `bun run test:mobile-workspace-entry &&` prefix and supplies JUnit arguments to the final batch runner only after the required preflight succeeds.

Revalidated the existing real child-process regression on cleanup head b0906c8a7e, which contains that merged fix:

```text
bun test packages/scripts/__tests__/run-all-tests-vacuous-green-guard.test.ts --test-name-pattern 'reconciles agent batch evidence after its required mobile preflight'
1 pass, 0 fail, 4 assertions; exit 0
```

The regression executes a package with this command shape, verifies reconciled executed-test evidence, then makes its preflight reject and verifies the package fails without claiming executed batch evidence. This is proof of the reported admission/failure-propagation boundary, not a claim that every server suite passes.

Closing this proposal and #31149 as superseded by the merged fix. Moving the Bun test into a second runner family is unnecessary to resolve the reported defect and would add ownership machinery during the script consolidation.

PR #31152, opened 2026-09-12
test(ui): give the client lanes a browser before the Story Gate timezone check runs (#31151)

### LINKED ISSUE #31151: test(ui): story-gate-timezone launches Chromium inside the browser-less client lanes and fails with a masked ENOENT
## Defect

`packages/ui/test/story-gate/story-gate-timezone.test.mjs` fails in both client lanes on `develop` (tip `ce68dbca425412131bd2d042b2cc4ffa6c8a61f7`; `tests / Client Tests` and `canonical / Tests (client)` in validation run 34694307489):

```text
 FAIL  test/story-gate/story-gate-timezone.test.mjs > keeps summer event positions aligned with winter-derived hour labels
Error: ENOENT: no such file or directory, open '/tmp/story-timezone-F5ykbv/output/report.json'
```

## Cause

The test spawns `run-story-gate.mjs`, which launches a real Chromium through Playwright (`chromium.launch()`), and it is included in the standard `packages/ui` suite (`vitest.config.ts` includes `test/**/*.test.mjs` for "pure-logic unit tests for the story-gate audit scripts"). Neither client job has a browser at that point: `.github/workflows/test.yml` runs `bun run test:client` and only then "Install Playwright Chromium" (for the remote-capability UI smoke), and the canonical client job in `.github/workflows/ci.yml` never installs one. The runner exits without writing `report.json`, and the test reads the report before checking the runner's exit status, so the real cause (`browserType.launch: Executable doesn't exist`) is hidden behind an ENOENT.

Reproduced locally at the tip: with browsers installed the test passes; with `PLAYWRIGHT_BROWSERS_PATH` pointing at an empty directory it fails with exactly the ENOENT above.

## Fix

Install Chromium before the client suite in both jobs (the `test.yml` job already installs it, just too late), and make the test assert the runner's exit status with its stderr before reading the report so a missing browser or a crash is named in the failure.

hermesagent270-commits:
The standard UI suite launches Chromium for its Story Gate timezone check, but the Client Tests workflow installed the browser after running that suite. Move the existing install step before the client tests and report subprocess startup/exit failures before reading the generated report.

Fixes #31151.

The final diff contains two files: `.github/workflows/test.yml` and `packages/ui/test/story-gate/story-gate-timezone.test.mjs`.

Validation for head `8a73114ca9615a7e87e9d40d60dbda156442d964`:
- All five hosted checks passed in [run 34802274576, attempt 3](https://github.com/elizaOS/eliza/actions/runs/34802274576): Source static smoke, Billing payment replay E2E, Subscription authority PostgreSQL, Browser bridge Windows security, and All Tests Passed.
- Earlier attempts ended with exit 143 during source checks; no code change was needed for the successful rerun.
- Existing author evidence reports the focused timezone check passing with Chromium installed, and the missing-browser case reporting the launcher failure instead of a missing-report error. Biome and workflow-trigger checks also passed in that evidence.

Screenshots, video, backend logs and model trajectories: N/A — this changes CI preparation and failure diagnostics, not a rendered product surface or agent behavior.

Original author tooling disclosure: Claude Code.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -188,6 +188,12 @@ jobs:
           skip-avatar-clone: "true"
           no-vision-deps: "true"
 
+      # packages/ui's standard suite launches a real Chromium for the Story Gate
+      # timezone check, so the browser must exist before the client lane runs.
+      - name: Install Playwright Chromium
+        if: needs.changes.outputs.client == 'true'
+        run: .github/scripts/install-playwright-browsers.sh chromium
+
       - name: Client tests
         if: needs.changes.outputs.client == 'true'
         run: bun run test:client
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -279,11 +279,13 @@ jobs:
           skip-avatar-clone: "true"
           no-vision-deps: "true"
 
-      - name: Run client tests
-        run: bun run test:client
-
+      # packages/ui's standard suite launches a real Chromium for the Story Gate
+      # timezone check, so the browser must exist before the client lane runs.
       - name: Install Playwright Chromium
         run: .github/scripts/install-playwright-browsers.sh chromium
+
+      - name: Run client tests
+        run: bun run test:client
 
       - name: Remote capability UI smoke
         run: bun run test:remote-capabilities:ui
--- a/packages/ui/test/story-gate/story-gate-timezone.test.mjs
+++ b/packages/ui/test/story-gate/story-gate-timezone.test.mjs
@@ -56,6 +56,11 @@ it("keeps summer event positions aligned with winter-derived hour labels", () =>
         timeout: 30000,
       },
     );
+    // A runner that could not start (no installed browser, a crash, the
+    // spawn timeout) leaves no report; surface its stderr instead of an
+    // ENOENT that hides the cause.
+    expect(result.error, result.stderr).toBeUndefined();
+    expect(result.status, result.stderr).toBe(0);
     const report = JSON.parse(
       readFileSync(join(directory, "output", "report.json"), "utf8"),
     );


attentionhead (approved):
**APPROVE** — verified at head `9ab320883619ebb9cc5ab182b0a896412256c3e4` (base `develop`, +16/-3, 3 files).

I checked the two claims this PR rests on — that the standard `packages/ui` suite really does launch a browser, and that exactly two CI lanes reach it — rather than taking the step reorder at face value.

### The suite genuinely collects this test

`packages/ui/vitest.config.ts:302` includes `test/**/*.test.mjs` alongside the `src/**` TypeScript globs, with a comment anticipating exactly these story-gate files. Nothing in the `exclude` list can remove it: every exclusion is scoped to `{ts,tsx}` except `dist/**`, `**/node_modules/**` and `**/__e2e__/**`, and `test/story-gate/` is none of those. So `story-gate-timezone.test.mjs` is in the default `bun run --cwd packages/ui test` selection, and it `spawnSync`s `run-story-gate.mjs`, which launches real Chromium. The premise holds.

The lane wiring closes the loop: `packages/ui/package.json` declares `"testLanes": ["client"]`, and `test:client` is `TEST_SCRIPT_FILTER='^test$' node packages/scripts/run-all-tests.mjs --lane=client …`, so the client lane selects precisely that `test` script. A browserless client job therefore fails on every run, which matches the reported symptom.

### Both lanes are covered, and I confirmed there is no third

`git grep 'test:client' .github` returns exactly two steps — `ci.yml:199` and `test.yml:288` — and both now have the Chromium install ahead of them. The other `packages/ui` invocations in `chat-shell-gestures.yml` either pass a positional file filter (`test -- src/components/chat/render-parity.contract.test.tsx`) or call dedicated `test:*-e2e` scripts, so none of them collect the story-gate file. The fix is complete, not partial.

The two gates are also right. `ci.yml`'s `tests_client` declares `needs: changes` (`:171`), so `needs.changes.outputs.client` resolves, and the new step carries the same `if:` as the `Client tests` step it precedes — it will not burn a minute on runs where the client lane is not affected. `test.yml`'s Client Tests job is gated at job level (`:260`), which is why the moved step needs no `if:` of its own. And moving the install earlier keeps it before `Remote capability UI smoke`, which was its original consumer, so that step still has its browser and the job's wall time really is unchanged.

Worth noting that `.github/scripts/install-playwright-browsers.sh` already documents this exact failure mode in its header — *"a lane that proceeds without its browser fails later with a far less obvious message"* — which is precisely what happened here. The script anticipated the bug; the step order defeated it.

### The test change is the right half of the fix

`expect(result.error, result.stderr).toBeUndefined()` then `expect(result.status, result.stderr).toBe(0)` converts an `ENOENT` on `report.json` — a message about a missing file, three layers from the cause — into the runner's own stderr. The `expect(value, message)` form is already the established idiom in this file (the pre-existing `expect(report.results[0].consoleErrors, result.stderr)` line uses it), so this is consistent rather than novel. The two assertions also split the cases usefully: a spawn timeout sets `error` with a null `status`, while a browser that fails to launch exits non-zero, and each is now reported distinctly.

### Non-blocking: the trailing status assertion is now dead weight

The test still ends with the pre-existing

```js
expect(result.status).toBe(0);
```

after the report assertions. The newly added `expect(result.status, result.stderr).toBe(0)` earlier in the function is strictly stronger — same predicate, and it carries stderr as the failure message — and it runs first, so the trailing line can no longer be the assertion that fails. It is harmless, but it now reads as if it were checking something the earlier line does not. Deleting it would leave the test saying exactly what it means.

### Open question: nothing pins the ordering that this PR establishes

The repair is a step order, and step order is exactly the kind of invariant that silently regresses — a future client lane, or a reorder during an unrelated workflow edit, puts the suite straight back to red. The new assertions make that regression *legible* (stderr instead of ENOENT), which is a real improvement, but the lane still fails.

Given the repo already machine-checks workflow invariants — the PR body itself runs `packages/scripts/workflow-trigger-policy.mjs`, and `test.yml:152` runs a dedicated vacuous-green guard test — is there an existing place where "any job that runs `test:client` installs Chromium beforehand" could be asserted? That would be a separate PR, not a change I would ask for here; I mention it because the two mechanisms already exist and this is the second CI-ordering defect of this shape.

ss251 (approved):
9ab320883619ebb9cc5ab182b0a896412256c3e4

**Findings**

1. Missing-browser failure mode in ci.yml — input: `needs.changes.outputs.client == 'true'` with no Chromium installed; wrong output: `bun run test:client` fails inside packages/ui when the Story Gate timezone check (packages/ui/test/story-gate/story-gate-timezone.test.mjs:56) spawns Playwright's Chromium. The fix adds `.github/scripts/install-playwright-browsers.sh chromium` (ci.yml:194-196) before the Client tests step, gated on the same condition. This matches the ordering already present in test.yml:283-287, where the install step precedes `Run client tests`; the ci.yml lane evidently lacked that ordering, so this closes a real gap rather than duplicating merged logic.

2. test.yml reordering (test.yml:279-287) is a pure swap — install step now runs before client tests instead of after. Both steps are conditional-free here, so the reorder only changes failure latency (browser missing → install fails first instead of tests failing with ENOENT). No behavioral risk beyond that.

3. Assertion strengthening in story-gate-timezone.test.mjs:59-63 — `expect(result.error, result.stderr).toBeUndefined()` and `expect(result.status, result.stderr).toBe(0)`. Input: runner could not start (no browser, crash, spawn timeout) → wrong output previously: `readFileSync(.../report.json)` throws ENOENT, hiding the actual stderr. New assertions surface stderr as the failure message. This is a genuine diagnostic improvement: the test previously could pass only if a report existed, but a runner failure produced a confusing ENOENT rather than the cause. Note the assertions run before the JSON parse, so a zero-exit runner with a malformed report still surfaces at the parse — unchanged and acceptable.

4. Scope check: workflow-only plus one test-file assertion change; no dependency, lockfile, secret access, lifecycle hook, or CI permission changes. The `if:` condition on the new ci.yml step matches the existing client-test gate exactly, so no lane installs a browser unnecessarily. No prompt-injection text in commit message or diff.

5. Verification limits: I did not execute the suite. The claimed failure mode (Story Gate launching Chromium) is consistent with the test file's spawn-with-timeout structure visible at story-gate-timezone.test.mjs:56-63, but I could not confirm the install script's own behavior (`.github/scripts/install-playwright-browsers.sh` is not in this diff).

Static review of the diff at the exact head. Live execution blocked: no disposable sandbox on this host, so this is a static review; no suite run, no checkout, no `bun install`.

```slop-review
{
  "runId": "not-available",
  "traceSha256": "not-available",
  "reproduced": false,
  "recommendedTier": "small",
  "recommendedThirds": 3,
  "securityRisk": "none",
  "duplicateRisk": "none",
  "splitRisk": "none",
  "commands": [],
  "notes": "Static review of diff at head 9ab320883619ebb9cc5ab182b0a896412256c3e4; no sandbox available, live execution blocked."
}
```

Recommendation: merge

— [free-9router-lane-ss251]

hermesagent270-commits:

```text
```

hermesagent270-commits:
Root `bun run verify` (`RUN_TURBO_CONCURRENCY=4`) on a stack branch holding the three PRs merged forward onto develop `61446bb82c9b` this morning (#31152 at `8a73114ca961`, #31128 at `84e0b15ecbd0`, #31294 at `5701bd18d621`; stack head `181bcfa30fb2`, 14 files over develop):

```text
[anti-larp] scanned 11748 test files — 0 focused, 0 orphaned skip(s)
verify exit 0
```

Log: `/root/eliza-fixes/verify-0914.log` on the contributor host.

mashingaan (approved):
## Approve at exact head `8a73114ca9615a7e87e9d40d60dbda156442d964`

I reviewed the current head against `61446bb82c9bf974965e767793c2333781306560`.

The workflow change fixes the client lane ordering in `.github/workflows/test.yml`: Playwright Chromium is installed before `bun run test:client`, so the standard UI suite has the browser before the Story Gate timezone test launches. The existing client job gate is unchanged. The equivalent `ci.yml` client lane already installs its browser before client tests, so the current delta closes the remaining lane without widening execution scope.

The test change checks `spawnSync` startup errors and a non-zero exit status with stderr before reading `output/report.json`. A missing browser, crash, or timeout therefore reports the actionable runner error instead of a misleading ENOENT. The existing report assertions remain in place, and the 30 second child timeout stays below the 40 second test timeout.

The exact head is open and mergeable. Subscription authority PostgreSQL, Billing payment replay E2E, and Browser bridge Windows security are green. Source static smoke ended with GitHub runner shutdown exit 143 during `tsc6 --noCheck` after successful builds. All Tests Passed failed only because it aggregates that source-smoke result. This is infrastructure failure in the observed run, not a failure in the changed workflow or test logic, so the interrupted source check should be rerun before merge.

Non-blocking follow-up: the trailing `expect(result.status).toBe(0)` is redundant after the new stderr-bearing assertion. Removing it is optional and does not affect correctness.

I found no reachable material workflow-condition, test-correctness, authorization, data-integrity, or security blocker at this head.

Recommend merge after the interrupted Source static smoke is rerun successfully.

— [eliza-pr31152-review]
\nCompute receipt: 0 project-attributed tokens (unavailable; device-signed, locally reported)
— [eliza-pr31152-review]

lalalune:

Retaining the real prerequisite-order repair. Current test.yml runs the client suite before installing Chromium, while ci.yml already installs Chromium before its client command; refresh the description to match the two-file diff rather than claiming both workflows still need changes. The new child error/status assertions correctly report launch failure before reading a missing report. Run the affected test.yml job with a clean browser cache and preserve the real Story Gate browser test; this is a prerequisite fix, not permission to replace browser behavior with a stub.

PR #31155, opened 2026-09-12
test(cloud): keep integration harness port leases below the ephemeral source-port floor (#31154)

### LINKED ISSUE #31154: test(cloud): integration harness leases ports inside the kernel ephemeral range, so workerd can hit EADDRINUSE at bind
## Defect

`cloud / integration-tests` failed on `develop` `ce68dbca4254` (validation run 34694307489, job 103555179715) before any test ran:

```text
[cloud-integration] run 7704-4f142ef05f52870778f1fe0a owns API http://127.0.0.1:44892, PGlite 127.0.0.1:37885, and /tmp/eliza-cloud-integration-Ul2BfD
[cloud-integration] START owned API server at http://127.0.0.1:44892
[db:migrate] applying ... (hundreds of migrations through a PGlite client connection)
*** Fatal uncaught kj::Exception: kj/async-io-unix.c++:941: failed: ::bind(sockfd, &addr.generic, addrlen): Address already in use; toString() = 127.0.0.1:44892
✘ [ERROR] Address already in use (127.0.0.1:44892). Please check that you are not already running a server on this address or specify a different port with --port.
Error: [cloud-integration] owned API server exited before readiness (code 1, signal none)
```

The previous run of the same job (a4654aa) passed, so this is intermittent.

## Cause

`acquirePortLease` in `packages/cloud/scripts/admin/integration-harness-lifecycle.mjs` draws random ports from `20000..60999`. Linux assigns outbound sockets a source port from `net.ipv4.ip_local_port_range`, which is `32768 60999` on the GitHub runner image and on ordinary hosts, so 28 of every 41 candidate ports (44892 included) sit inside the kernel's ephemeral pool. The lease only proves that nothing is *listening* (`tcpListening`); between the lease and workerd's `bind()` the harness opens many client connections (the migration run over the PGlite TCP port, Wrangler's own fetches), any of which can be handed the leased port as its source port. `bind()` on a port held by an established client socket fails with `EADDRINUSE`, which is exactly the workerd message above. The PGlite lease has the same exposure.

## Fix

Keep random lease candidates below the kernel's ephemeral floor: read `/proc/sys/net/ipv4/ip_local_port_range` when it exists (fall back to the IANA floor 49152 elsewhere) and clip the candidate range to `20000..floor-1`. An explicitly requested `preferredPort` keeps its current semantics. Covered by a unit test of the range clipping and a lease-draw test that asserts every random lease lands below the floor.

hermesagent270-commits:
# Relates to

Fixes #31154

- [x] A reviewer can confirm the change from the range arithmetic in the new test and the lease-draw assertion, without reading the code.

# Sync with develop

# Risks

Low. Only the random candidate range of the cloud integration harness changes: on Linux it becomes `20000..32767` instead of `20000..60999`; on hosts without procfs it becomes `20000..49151`. That still leaves thousands of candidates for the 256 attempts. An explicitly requested `preferredPort` behaves exactly as before.

# Background

## What does this PR do?

`acquirePortLease` proves that nothing *listens* on a port. It does not stop the kernel from handing that port to an outbound socket as its source port, and Linux draws those from `net.ipv4.ip_local_port_range` (`32768 60999` on the runner image and on ordinary hosts). The harness opens many client connections between the lease and workerd's bind (the whole migration run over the PGlite TCP port), so a lease drawn from the ephemeral pool can be taken by one of them, and `bind()` then fails with `EADDRINUSE`. That is the `cloud / integration-tests` failure on develop `ce68dbca4254` (port 44892, inside the pool; the previous run passed).

The harness now reads `/proc/sys/net/ipv4/ip_local_port_range` when present (IANA floor 49152 otherwise) and clips random candidates below that floor. Two exported helpers, `ephemeralPortFloor` and `candidatePortRange`, carry the logic so the test can pin it.

Test-infrastructure repair for an intermittent develop lane failure.

# Testing

`candidatePortRange` and `readLocalPortRange` in `integration-harness-lifecycle.mjs`, then the new case in `run-integration-tests.test.mjs`.

## Detailed testing steps

```text
node --test --test-name-pattern='ephemeral source-port floor|stale lease reclamation|refuses a listener that races' packages/cloud/scripts/admin/run-integration-tests.test.mjs
ok 1 - refuses a listener that races a lease without stopping it
ok 2 - random lease candidates stay below the kernel's ephemeral source-port floor
ok 3 - stale lease reclamation compares token and inode before replacement
# pass 3
# fail 0

cat /proc/sys/net/ipv4/ip_local_port_range     # 32768 60999 on this host; 12 random leases all landed in 20000..32767
cd packages/cloud && bunx biome check scripts/admin/integration-harness-lifecycle.mjs scripts/admin/run-integration-tests.test.mjs
# only the pre-existing noUndeclaredEnvVars warnings on BUN / npm_execpath (lines untouched by this PR)
```

Before: develop run 34694307489, job 103555179715, `Address already in use (127.0.0.1:44892)` from workerd, lane exit before readiness.

### DIFF
--- a/packages/cloud/scripts/admin/integration-harness-lifecycle.mjs
+++ b/packages/cloud/scripts/admin/integration-harness-lifecycle.mjs
@@ -24,6 +24,59 @@ import { DatabaseSync } from "node:sqlite";
 
 const DEFAULT_PORT_MIN = 20_000;
 const DEFAULT_PORT_MAX = 60_999;
+// A lease proves that nothing listens on a port, not that the kernel will not
+// hand it to an outbound socket as its source port before the server binds.
+// Linux draws those from net.ipv4.ip_local_port_range (32768-60999 by default),
+// so random candidates must stay below that floor or workerd fails its bind
+// with EADDRINUSE after the harness opened its migration connections.
+const IANA_EPHEMERAL_PORT_MIN = 49_152;
+const LOCAL_PORT_RANGE_PATH = "/proc/sys/net/ipv4/ip_local_port_range";
+
+function readLocalPortRange() {
+  let text;
+  try {
+    text = readFileSync(LOCAL_PORT_RANGE_PATH, "utf8");
+  } catch {
+    // error-policy:J3 hosts without procfs (macOS, Windows) report no range
+    // and the caller applies the IANA default instead of a fake value.
+    return null;
+  }
+  const parts = text.trim().split(/\s+/);
+  if (parts.length !== 2) return null;
+  const [min, max] = parts.map((value) =>
+    /^\d+$/.test(value) ? Number(value) : Number.NaN,
+  );
+  if (
+    !Number.isInteger(min) ||
+    !Number.isInteger(max) ||
+    min < 1 ||
+    max < min
+  ) {
+    return null;
+  }
+  return { min, max };
+}
+
+/** Lowest port the kernel may assign to an outbound socket on this host. */
+export function ephemeralPortFloor(readRange = readLocalPortRange) {
+  const range = readRange();
+  return range ? range.min : IANA_EPHEMERAL_PORT_MIN;
+}
+
+/** Random-lease candidates: the configured range clipped below the ephemeral floor. */
+export function candidatePortRange({
+  min = DEFAULT_PORT_MIN,
+  max = DEFAULT_PORT_MAX,
+  ephemeralMin = ephemeralPortFloor(),
+} = {}) {
+  const upper = Math.min(max, ephemeralMin - 1);
+  if (upper < min) {
+    throw new Error(
+      `[cloud-integration] no lease-able ports: ${min}-${max} lies entirely at or above the ephemeral floor ${ephemeralMin}`,
+    );
+  }
+  return { min, max: upper };
+}
 const DEFAULT_STARTUP_TIMEOUT_MS = 90_000;
 const DEFAULT_STOP_TIMEOUT_MS = 10_000;
 const OWNER_FILE = ".cloud-integration-owner.json";
@@ -264,8 +317,9 @@ export async function acquirePortLease({
     return lease;
   }
 
+  const candidates = candidatePortRange();
   for (let attempt = 0; attempt < maxAttempts; attempt += 1) {
-    const port = randomInt(DEFAULT_PORT_MIN, DEFAULT_PORT_MAX + 1);
+    const port = randomInt(candidates.min, candidates.max + 1);
     const lease = await tryAcquirePortLease({
       host,
       port,
--- a/packages/cloud/scripts/admin/run-integration-tests.test.mjs
+++ b/packages/cloud/scripts/admin/run-integration-tests.test.mjs
@@ -22,9 +22,11 @@ import { fileURLToPath } from "node:url";
 import pg from "pg";
 import {
   acquirePortLease,
+  candidatePortRange,
   cleanupRunContext,
   createManagedRunContext,
   createOwnedChildRegistry,
+  ephemeralPortFloor,
   externalApiHealthy,
   installSignalTeardown,
   managedApiHealthy,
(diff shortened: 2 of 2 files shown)


ss251 (approved):
866b1f6640ca7fd0b45a2081d3110c6c0e2681e7

**Findings**

The diff fixes a real race: `acquirePortLease` previously drew candidates from 20000–60999 (`integration-harness-lifecycle.mjs`, removed line `randomInt(DEFAULT_PORT_MIN, DEFAULT_PORT_MAX + 1)`), and 49152–60999 overlaps the default Linux `ip_local_port_range` (32768–60999), so the kernel could hand a leased port to an outbound socket between lease acquisition and server bind — EADDRINUSE as described in the linked issue.

Mechanism check, static:

- `candidatePortRange` (new, ~line 62) clips `max` to `ephemeralMin - 1` and throws a clear error if the configured range lies entirely at or above the floor. Input → output holds: `{20000, 60999, floor 32768}` → `{20000, 32767}` per the test.
- `ephemeralPortFloor` reads `/proc/sys/net/ipv4/ip_local_port_range`, validates `min ≥ 1`, `max ≥ min`, integers only, and falls back to the IANA floor 49152 on procfs-less hosts. Using `range.min` as the floor is correct — Linux draws source ports from anywhere in the range, not just the top.
- Loop site (~line 320) now uses `candidates.min/max` consistently; the throw happens before the retry loop, so a misconfigured host fails fast rather than looping silently.
- Test asserts the clipped range, the passthrough case, the throw, both fallbacks, and then acquires 12 real leases asserting `port < floor` on the live host. It would fail if the clipping regressed, so it is behavioral, not cosmetic.

Minor, non-blocking: the IANA fallback (49152) still narrows the default max from 60999 to 49151 on macOS/Windows, shrinking the candidate pool by ~20%; acceptable for a test harness. The procfs parse treats values like `0x` or leading `+` as NaN → null → fallback, which is the safe direction.

No lockfile changes, no lifecycle hooks, no CI permission changes, no binaries or symlink tricks in the diff. Scope is tight: one production function pair, one call site, one test.

Live execution blocked: no disposable sandbox on this host, so this is a static review of the diff at the exact head. The suite (`run-integration-tests.test.mjs`) was not run; the live-lease assertions in the new test are the only part whose outcome I cannot confirm from reading.

Recommendation: merge

— [free-9router-lane-ss251]

attentionhead (approved):
Verified at head `866b1f6640ca7fd0b45a2081d3110c6c0e2681e7` (base `develop`, 2 files, +106/-1), reading the fetched PR head rather than the diff alone.

## The diagnosis holds

A lease proves nothing *listens* on a port; it cannot stop the kernel choosing that port as the **source** port of an outbound socket. `acquirePortLease` drew uniformly from `20000..60999`, and Linux's default `net.ipv4.ip_local_port_range` is `32768 60999` — so roughly 70% of the candidate space overlapped the ephemeral pool. The harness opens many client connections (PGlite, the API health probes) between the lease and workerd's `bind()`, which is exactly the window in which the observed `EADDRINUSE` on 44892 becomes reachable. 44892 sits inside `32768..60999`, consistent with the stated cause rather than merely compatible with it.

## Verified outside the diff

The new `readLocalPortRange` calls `readFileSync`, and the import line is **not** in the diff — I checked it at head rather than assuming: `readFileSync` is already present in the `node:fs` import block (`integration-harness-lifecycle.mjs:16`), alongside `closeSync`/`fstatSync`/`openSync`. No load-time `ReferenceError`.

`randomInt(candidates.min, candidates.max + 1)` at `:322` is the **only** random port draw in `packages/cloud/scripts` — `git grep -nE 'DEFAULT_PORT_MIN|DEFAULT_PORT_MAX|randomInt\('` over the package returns exactly the two constant declarations, the two default-parameter references in `candidatePortRange`, and that one call site. So there is no second, unclipped draw left behind, and `DEFAULT_PORT_MAX` is not dead: it remains the `max` default.

The Risks note that "an explicitly requested `preferredPort` behaves exactly as before" is accurate and load-bearing: `acquirePortLease` returns from the `requested !== null` branch *before* `const candidates = candidatePortRange()` is evaluated. A caller pinning a port therefore neither gets clipped nor can trip the new throw — which matters, because that throw is otherwise reachable at call time rather than at module load.

`releasePortLease` is imported by the test alongside the two new exports, and the 12-lease loop's `finally` releases only what was actually pushed, so a mid-loop throw leaks nothing.

## Why the shape is right

Exporting `ephemeralPortFloor(readRange)` and `candidatePortRange({min,max,ephemeralMin})` with injectable inputs is what makes this testable at all — the assertions pin the clip arithmetic (`32768` floor → `20000..32767`), the no-op case (a `max` already below the floor is unchanged), the refusal case, and both `ephemeralPortFloor` fallbacks, none of which would be observable through `acquirePortLease` alone. The `/proc` read is fail-soft to the IANA `49152` floor with an `error-policy:J3` comment, so macOS and Windows developers get a conservative real bound rather than a fabricated one.

## Open question (non-blocking)

`candidatePortRange` throws when `Math.min(max, ephemeralMin - 1) < min`. With the shipped defaults that needs `ephemeralMin <= 20000` — unusual on a stock kernel, but it is a real configuration: containers and images that set `ip_local_port_range` to `1024 65535` are not rare, and CI runners are not always stock. On such a host this turns an intermittent `EADDRINUSE` flake into a deterministic hard failure of the whole cloud integration lane, and the message ("no lease-able ports") describes the arithmetic rather than the remedy.

Is the intended behaviour there to fail loudly, or to fall back to the existing scan-and-retry (which, with `maxAttempts = 256` and a lease file per candidate, still terminates) and accept the residual race the way the harness did before this PR? Either answer seems defensible; I ask only because the current one is silent about which was chosen, and the thrown message gives an operator no next step. Naming the knob (`ip_local_port_range`) in the error text would cost one line.

Nothing here blocks: on every host the CI actually runs, the clip is strictly safer than the status quo, the change is confined to candidate selection, and the new test would catch a regression in the arithmetic.

Approving.

hermesagent270-commits:

```text
```

lalalune:

Retaining the plausible Linux collision mitigation, with scope corrections. A lease below the actual kernel ephemeral range reduces this race, but the IANA fallback is not evidence of the configured range on macOS/Windows, and swallowing every procfs read error as absence can hide a Linux discovery failure. Make platform fallback/unsupported state explicit and distinguish missing procfs from other errors; validate parsed range bounds. Preserve listener-race and lease-ownership protections, and show the real Linux harness starts under connection pressure. The current twelve-port test proves range selection only, not exclusive binding or elimination of all EADDRINUSE failures; preferredPort is also outside this clipping contract.

PR #29982, opened 2026-08-30
fix(app): align audit with unavailable view fallbacks

lalalune:
## Summary

- validate native-only Camera and unregistered Cockpit against the designed unavailable-view boundary
- update retained Rolodex deep-link probes to its current recoverable unavailable state
- keep semantic exemptions narrow and observable

## Verification

- OCR policy contract: 6 passed
- strict app audit previously isolated exactly these 12 stale fallback verdicts across four viewports

Fixes #29978

PR #31157, opened 2026-09-12
fix(app): keep an unreadable OCR transcript from fabricating a missing-label regression (#31156)

### LINKED ISSUE #31156: test(app): OCR triage marks the populated launcher baseline broken and the real-engine suite is red on develop
## Defect

`packages/app/scripts/ocr-real-engine.test.ts` > "real OCR blank-vs-unreadable classification > does not call a populated mobile launcher blank when the first OCR pass is weak" fails in both client lanes on `develop` (tip `ce68dbca425412131bd2d042b2cc4ffa6c8a61f7`, validation run 34694307489) and reproduces locally with the packaged Tesseract engine:

```text
Expected: "needs-eyeball"
Received: "broken"
```

The triage entry for the baseline launcher capture (`scripts/mvp-visual-verify/baseline/mobile-portrait/builtin-rolodex.png`, a fully populated app grid with the "Ask Eliza" composer):

```json
{
  "ocrVerdict": "broken",
  "regression": true,
  "pixelBlank": false,
  "reasons": ["missing expected content: View unavailable, This app is unavailable here, Install or enable it, App, Retry | Back to views"],
  "words": 28,
  "meanConfidence": 0.46,
  "attempts": [
    { "mode": "auto",                 "words": 19, "conf": 0.30, "text": "© |  | [5 |  | te |  | [43 J | = fo | 4 q - | ... | + Askliza °" },
    { "mode": "sparse-high-contrast", "words": 48, "conf": 0.38, "text": "MN o— | 11s | MCSSUYCS) | Saddles | [WEIICY | JLUSKS] | ..." },
    { "mode": "sparse-grayscale",     "words": 28, "conf": 0.46, "text": "E8CE | 0 | rao, | te | vv | ... | Ask Eliza | 9" }
  ]
}
```

## Cause

Two things combine:

1. The fixture is keyed to `builtin-rolodex`, whose semantic policy has been `VIEW_UNAVAILABLE_FALLBACK` since `678c08874fe9` (#29982). The capture is the launcher, so the declared expectation can never match it; the test only passed while the OCR came back *inconclusive* (mean confidence below `OCR_RELIABLE_CONFIDENCE_FLOOR = 0.45`), which routes to `needs-eyeball` before the expectation is checked.
2. On the runner image and on this host the best pass reads the icon labels as noise at 0.46 mean confidence, one point over the floor. `evaluateOcrContent` then treats the transcript as reliable, reports the unavailable-view labels as missing, and the triage records a populated, DOM-good launcher as a **regression** (`regression: true`). That is a false positive of the audit tooling itself, not only of the test: 28 "words" of which almost none are words is not a reliable transcript.

## Fix

- Gate "missing expected content" on a transcript that actually reads: when no semantic anchor of the declared expectation matched, require the transcript to be lexically readable (a floor on the share of recognisable words) before the missing labels can turn a non-blank capture into `broken`; otherwise it is `needs-eyeball` with the inconclusive reason, which is what a human then checks.
- Re-key the real-engine fixture to the view the capture depicts (`builtin-views`, the launcher policy) so the case stops depending on `builtin-rolodex`'s retired expectation.

Evaluated against the full mobile-portrait baseline set before and after the rule change so verified views do not regress into the manual pile.

hermesagent270-commits:
# Relates to

Fixes #31156

- [x] A reviewer can confirm the change from the real-engine run and the baseline-wide before/after evaluation below without reading the code.

# Sync with develop

# Risks

Low and measured. The only behavioural change is in `evaluateOcrContent`: a transcript with **no** matched semantic anchor now also has to be lexically readable before its missing labels can make a non-blank capture `broken`; otherwise it is `needs-eyeball` (manual review). Anchor matches, the word floor, blank-pixel verdicts, and blank-exempt surfaces (`exemptFromBlank`) are untouched, so a verified view cannot lose its verification and a blank paint cannot be hidden. Evaluated against the full mobile-portrait baseline set (below): verified count unchanged, two glyph-salad launcher captures move from a fabricated regression to manual review, nothing else moves. The vowel heuristic is English-only, which matches the audited UI locale; it is documented on the constant.

# Background

## What does this PR do?

The packaged Tesseract engine reads the mobile launcher's icon labels (small text on gradient tiles) as glyph salad: 28 "words" at 46% mean confidence, almost none of them words (`E8CE 0 rao, te vv re NCA EL & AP 09 ca 3°] a NZ DN 4 kh = MA fA Lo LJ + Ask Eliza 9`). That clears `OCR_RELIABLE_CONFIDENCE_FLOOR` (0.45) by a point, so the rules treated it as a reliable transcript, reported every declared label as missing, and the triage recorded a populated, DOM-good launcher as a **regression**. The real-engine suite that exists to guard this (`ocr-real-engine.test.ts` > "does not call a populated mobile launcher blank when the first OCR pass is weak") has been red on `develop` for it.

- `test/ui-smoke/ocr-content-rules.ts`: adds `readableTokenShare` and `OCR_READABLE_TOKEN_SHARE_FLOOR` (0.5). `ocrInconclusive` is now true when no anchor matched and either the confidence floor or the readable-share floor is not met; the inconclusive reason reports the readable share.
- `test/audit/ocr-content-rules.test.ts`: pins the share on the real glyph-salad transcript versus rendered labels; the salad at 0.46 routes to review with no fabricated miss; a readable transcript at the same 0.46 that omits `Settings`/`Wallet` still breaks.
- `scripts/ocr-real-engine.test.ts`: the launcher capture is audited under the launcher policy (`builtin-views`) instead of `builtin-rolodex`, whose policy became the unavailable-view fallback in #29982; the case also asserts the inconclusive reason and `regression: false`.

Bug fix in the visual-audit tooling plus the repair of the red suite that guards it.

# Testing

`readableTokenShare` and the `ocrInconclusive` expression in `ocr-content-rules.ts`, then the three new unit cases, then the real-engine case.

## Detailed testing steps

```text
cd packages/app
bun test scripts/ocr-real-engine.test.ts
# before (develop): (fail) does not call a populated mobile launcher blank … Expected: "needs-eyeball" Received: "broken"
# after:  7 pass, 0 fail

bunx vitest run --config vitest.config.ts test/audit/ocr-content-rules.test.ts test/audit/ocr-view-expectations.test.ts
 Tests  54 passed (54)      # 51 on develop + 3 new
bunx biome check test/ui-smoke/ocr-content-rules.ts test/audit/ocr-content-rules.test.ts scripts/ocr-real-engine.test.ts
Checked 3 files. No fixes applied.
```

Baseline-wide evaluation: the OCR triage run over every `scripts/mvp-visual-verify/baseline/mobile-portrait/*.png` that has a declared policy (46 captures), each with a DOM verdict of `good`, on the packaged engine, before and after this change:

| | verified | broken | needs-eyeball |
| --- | --- | --- | --- |
| develop `ce68dbc` | 31 | 14 | 1 |
| this head | 31 | 12 | 3 |

The two captures that move (`builtin-rolodex`, `plugin-contacts-gui`) are byte-identical copies of the launcher grid read at 46% confidence with 17% readable tokens; their reason is now `OCR inconclusive — 28 word(s), 46.0% mean confidence, 17% readable tokens; pixels were not blank`. The three other copies of the same image (`builtin-camera`, `plugin-messages-gui`, `plugin-phone-gui`) stay `broken` because their slugs are blank-exempt native surfaces, where the inconclusive gate is bypassed by design; that is pre-existing and out of scope here. The remaining 10 `broken` entries are confident transcripts (60–92%) of stale baseline captures that genuinely do not match their current expectations.

### DIFF
--- a/packages/app/scripts/ocr-real-engine.test.ts
+++ b/packages/app/scripts/ocr-real-engine.test.ts
@@ -140,15 +140,18 @@ describe("real OCR blank-vs-unreadable classification", () => {
   }, 90_000);
 
   it("does not call a populated mobile launcher blank when the first OCR pass is weak", async () => {
+    // The capture is the launcher grid, so it is audited under the launcher
+    // policy (builtin-views). The file is the historical rolodex baseline,
+    // whose own policy has since become the unavailable-view fallback.
     const auditDir = join(dir, "launcher-audit");
     const viewportDir = join(auditDir, "mobile-portrait");
     mkdirSync(viewportDir, { recursive: true });
-    copyFileSync(LAUNCHER_CAPTURE, join(viewportDir, "builtin-rolodex.png"));
+    copyFileSync(LAUNCHER_CAPTURE, join(viewportDir, "builtin-views.png"));
     writeFileSync(
       join(auditDir, "report.json"),
       JSON.stringify([
         {
-          slug: "builtin-rolodex",
+          slug: "builtin-views",
           viewport: "mobile-portrait",
           viewType: "gui",
           verdict: "good",
@@ -181,8 +184,11 @@ describe("real OCR blank-vs-unreadable classification", () => {
       true,
     );
     expect(entry.pixelBlank).toBe(false);
+    // Glyph salad over the icon labels can clear the confidence floor by a
+    // point; it is a manual check, never a fabricated missing-label regression.
     expect(entry.ocrVerdict).toBe("needs-eyeball");
     expect(entry.regression).toBe(false);
+    expect(entry.reasons.join(" ")).toMatch(/OCR inconclusive/);
     expect(entry.reasons.join(" ")).not.toMatch(/pixels are blank/i);
   }, 90_000);
 
--- a/packages/app/test/audit/ocr-content-rules.test.ts
+++ b/packages/app/test/audit/ocr-content-rules.test.ts
@@ -12,6 +12,7 @@ import {
   normalize,
   type OcrResult,
   positiveExpectationMatches,
+  readableTokenShare,
 } from "../ui-smoke/ocr-content-rules";
 import {
   resolveViewOcrPolicy,
@@ -69,6 +70,25 @@ describe("normalize", () => {
   it("lowercases and canonicalizes punctuation and whitespace", () => {
     expect(normalize("  Ask   Eliza\n\n")).toBe("ask eliza");
     expect(normalize("Fine-Tuning")).toBe("fine tuning");
+  });
+});
+
+// The packaged engine's best pass over the populated mobile launcher baseline
+// (icon labels on gradient tiles): 28 "words" at 46% mean confidence, almost
+// none of them words.
+const LAUNCHER_GLYPH_SALAD =
+  "E8CE 0 rao, te vv re NCA EL & AP 09 ca 3°] a NZ DN 4 kh = MA fA Lo LJ + Ask Eliza 9";
+
+describe("readableTokenShare", () => {
+  it("separates glyph salad from rendered labels", () => {
+    expect(readableTokenShare(LAUNCHER_GLYPH_SALAD)).toBeLessThan(0.5);
+    expect(
+      readableTokenShare(
+        "Messages Settings Wallet Tasks Automations Browser Ask Eliza",
+      ),
+    ).toBe(1);
+    expect(readableTokenShare("")).toBe(0);
+    expect(readableTokenShare("$1,650.50 0250000 ETH")).toBeLessThan(0.5);
   });
 });
 
@@ -192,6 +212,37 @@ ph eg`,
       expect(f.verdict).toBe("verified");
     },
   );
+
+  it("routes an unreadable transcript that clears the confidence floor to review, not a fabricated miss", () => {
+    const f = evaluateOcrContent({
+      ocr: ocr(LAUNCHER_GLYPH_SALAD, {
+        meanConfidence: 0.46,
+        pixelBlank: false,
+      }),
+      expectation: {
+        requireAll: ["Settings", "Wallet"],
+        requireAny: ["Projects", "Calendar", "Automations"],
+      },
+    });
+    expect(f.ocrInconclusive).toBe(true);
+    expect(f.missingRequired).toHaveLength(0);
+    expect(f.blankPixels).toBe(false);
+    expect(f.verdict).toBe("needs-eyeball");
+    expect(f.reasons.join(" ")).toMatch(/readable tokens/);
+  });
+
+  it("still breaks a readable transcript that omits a required label at the same confidence", () => {
+    const f = evaluateOcrContent({
+      ocr: ocr("Messages Tasks Automations Browser Character Ask Eliza", {
+        meanConfidence: 0.46,
+        pixelBlank: false,
+      }),
+      expectation: { requireAll: ["Settings", "Wallet"] },
+    });
+    expect(f.ocrInconclusive).toBe(false);
+    expect(f.missingRequired).toEqual(["Settings", "Wallet"]);
+    expect(f.verdict).toBe("broken");
+  });
 
   it("requires token boundaries when short anchors bypass low OCR confidence", () => {
     const exactAnchor = evaluateOcrContent({
--- a/packages/app/test/ui-smoke/ocr-content-rules.ts
+++ b/packages/app/test/ui-smoke/ocr-content-rules.ts
@@ -199,6 +199,26 @@ export function detectPlaceholderLeaks(text: string): string[] {
 export const OCR_RELIABLE_WORD_FLOOR = 2;
 export const OCR_RELIABLE_CONFIDENCE_FLOOR = 0.45;
 
+/**
+ * Minimum share of transcript tokens that read as words (three or more letters
+ * with a vowel) before a transcript with no matched semantic anchor may claim
+ * that a declared label is missing. Tesseract renders icon labels it cannot
+ * resolve as glyph salad ("MCSSUYCS", "JLUSKS", "E8CE") that can still clear
+ * the mean-confidence floor by a point; such a transcript proves nothing about
+ * which labels the view rendered, so it stays a manual check instead of a
+ * fabricated regression. The vowel test is English-only, like the audited UI.
+ */
+export const OCR_READABLE_TOKEN_SHARE_FLOOR = 0.5;
+
+export function readableTokenShare(text: string): number {
+  const tokens = normalize(text).split(" ").filter(Boolean);
+  if (tokens.length === 0) return 0;
+  const readable = tokens.filter(
+    (token) => /^\p{Letter}{3,}$/u.test(token) && /[aeiouy]/.test(token),
+  ).length;
+  return readable / tokens.length;
+}
+
 export interface EvaluateArgs {
   ocr: OcrResult;
   expectation?: OcrExpectation;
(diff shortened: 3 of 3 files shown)


attentionhead (approved):
Verified at head `b4a49197ee3086ac1abdc589639fba173b849889` (base `develop`, 3 files, +84/-5), reading the fetched head rather than the diff alone.

## The guard rewrite is a strict widening, and that is the right shape

```
-  (ocr.meanConfidence < CONF_FLOOR && !semanticAnchorsMatched)
+  (!semanticAnchorsMatched && (ocr.meanConfidence < CONF_FLOOR || readableShare < SHARE_FLOOR))
```

Factoring `!semanticAnchorsMatched` out and adding the share disjunct only ever moves cases **into** `ocrInconclusive`, never out. That matters because `ocrInconclusive` is what suppresses `missingRequired` (`:280`), so the change cannot make a previously-`needs-eyeball` capture start claiming a missing label. A view that matched a semantic anchor still short-circuits before the share is consulted, so a verified view cannot lose its verification — which is the property the Risks section claims, and it holds structurally rather than by test coverage.

## The real-engine case is non-vacuous — I checked the policy it now resolves to

The slug swap is the part worth scrutiny, because "audit it under a different policy" is exactly how a red test gets made green without fixing anything. It isn't that here:

- `builtin-views` → `expected(LAUNCHER_FALLBACK)` = `requireAll: ["Settings", "Wallet"]`, `requireAny: ["Projects", "Calendar", "Automations"]` (`ocr-view-expectations.ts:44-47, :112`).
- The glyph-salad transcript contains **none** of those five labels, so `semanticAnchorsMatched` is genuinely `false` and the new disjunct is actually reached.
- Therefore on `develop` the case ran: 28 words ≥ floor, 0.46 ≥ 0.45, no anchor → conclusive → `Settings`/`Wallet` plus the disjunction reported missing → `broken`. At this head: 17% readable → inconclusive → `needs-eyeball`, `regression: false`. That is the assertion, and it fails for the stated reason before the fix.

The `builtin-rolodex` → `VIEW_UNAVAILABLE_FALLBACK` claim also checks out verbatim (`:54-62, :201`): that policy now expects "View unavailable" / "This app is unavailable here" / "Install or enable it", which a populated launcher screenshot could never satisfy. Auditing a launcher capture under it was the actual mistake, and the rename fixes the fixture rather than the threshold.

I also confirmed the new policy is **not** blank-exempt: `expected()` returns a plain `{kind: "expectation"}` (`:27-29`) — only `exempt()` produces a semantic exemption — and triage derives `exemptFromBlank` solely from `rep.viewType === "tui" || BLANK_EXEMPT_SLUGS.has(slug)` (`ocr-triage.ts:445-446`). So the inconclusive gate genuinely applies here, and the PR's note that the three blank-exempt copies stay `broken` is pre-existing behaviour, correctly scoped out.

## The vowel test is safe against casing

`readableTokenShare` matches `/[aeiouy]/` — lowercase only — which would misclassify every ALL-CAPS rendered label as unreadable and silently route real regressions to manual review. The unit fixture (`"Messages Settings Wallet …"`) is mixed-case and would not catch that. It is a non-issue: `normalize` applies `.toLowerCase()` as its first operation (`:190-196`) before NFKD and the non-alphanumeric split, and `readableTokenShare` tokenizes off `normalize(text)`. Checked rather than assumed, and withdrawn.

## Two observations, neither blocking

**1. `readableShare` is measured on a different textual basis than the anchor decision.** Anchors run over `positiveSegments` (`:257-264`), which triage builds as an *accumulating union across OCR attempts* — seeded `[rec.text]` at `ocr-triage.ts:495` and appended per rescued attempt at `:505-508`. `readableTokenShare(ocr.text)`, by contrast, sees only the selected record's text. So the anchor test gets the benefit of every pass while the readability test gets one. In practice this is nearly always moot, because a matched anchor short-circuits before the share is read — the mismatch is only observable when no pass found a required label, and there the selected record is already the semantically best one. Worth a comment more than a change; flagging it because a future edit that drops the `!semanticAnchorsMatched` short-circuit would make the asymmetry load-bearing.

**2. Widening inconclusive also widens leak suppression.** `errorLeaks` and `placeholderLeaks` are both gated on `!ocrInconclusive` (`:273-276`), so a transcript that now falls below the share floor stops reporting them. The newly-affected band is narrow — no anchor, confidence ≥ 0.45, share < 0.5 — and an error string ("Something went wrong", "Failed to load") is readable English that pulls the share up, so a screen whose *content* is an error is unlikely to land there. But a leak rendered over an otherwise-unreadable surface would now be swallowed where it previously surfaced. Given the verdict is `needs-eyeball` rather than `verified`, a human still looks at it, so this trades a false alarm for a deferred one rather than for silence.

## Evidence

The before/after table is the right artifact for this change: it shows `verified` unchanged at 31 across all 46 captures, which is the only number that could indicate the gate hiding real breakage, and the two movers are named with their reason string. The unit tests pin both directions at the *same* 0.46 confidence — salad routes to review, readable-but-incomplete still breaks — which isolates the share as the operative variable rather than confounding it with the confidence floor.

Nothing blocks. The diagnosis matches the code, the fixture rename is a genuine correction rather than a threshold move, and the red suite it repairs is the one that exists to guard this exact failure.

Approving.

hermesagent270-commits:

```text
```

ss251 (approved):
b4a49197ee3086ac1abdc589639fba173b849889

The old `ocrInconclusive` predicate treated a transcript as conclusive once it had ≥2 words and `meanConfidence ≥ 0.45`, even when those tokens were Tesseract glyph salad. `evaluateOcrContent` then populated `missingRequired` from unmatched `requireAll` / `requireAny` labels and could emit a fabricated missing-label regression on a populated launcher (the #31156 failure).

This head closes that path. `readableTokenShare` (`ocr-content-rules.ts:213-221`) counts tokens that are ≥3 letters with a vowel; `OCR_READABLE_TOKEN_SHARE_FLOOR` is 0.5. `ocrInconclusive` now fires when no semantic anchor matched and (confidence is below 0.45 **or** readable share is below the floor) (`ocr-content-rules.ts:266-273`). On that path `missingRequired` stays empty and the verdict is `needs-eyeball`, not `broken`.

Paired tests would fail if the claim broke:
- glyph salad at 0.46 confidence with `requireAll: Settings, Wallet` → `ocrInconclusive`, `missingRequired []`, `needs-eyeball` (`ocr-content-rules.test.ts:216-232`)
- readable labels at the same 0.46 confidence omitting those two → still `broken` with exactly those misses (`ocr-content-rules.test.ts:234-247`)

The real-engine fixture rename `builtin-rolodex` → `builtin-views` is in scope: the capture is the launcher grid, so it must use the launcher policy, not the rolodex unavailable-view fallback. The added `/OCR inconclusive/` assertion is the live check that this capture no longer becomes a regression. Not test weakening: the readable-miss case is new, and the blank-pixel assertions are unchanged.

Chrome text such as “Ask Eliza” in the salad does not by itself clear `semanticAnchorsMatched` unless it is an expected view anchor; the new unit case correctly omits it from the expectation so the heuristic is what is under test.

Live execution blocked: no disposable sandbox on this host, so this is a static review of the diff at the exact head. Verified by reading the three-file diff at this head; no suite was executed.

Recommendation: merge

— [free-xai-lane-ss251]

hermesagent270-commits:
Superseded on develop: `61446bb82c9b` (#31269, "test(app): classify archived launcher OCR under its content contract") re-keyed the real-engine fixture to `builtin-views` and pinned the opposite verdict to this PR's inconclusive gate (`ocrVerdict: "broken"`, `regression: true`, reasons matching `missing expected content`). The develop run at that commit (34801372277) no longer fails the OCR case, so the owner's classification stands and this branch now conflicts with it. Closing.

PR #31159, opened 2026-09-12
test(plugin-goals): anchor the goals-repository leaf to source for the keyless real-db graph (#31158)

### LINKED ISSUE #31158: test(plugin-goals): goals.real-db suite cannot resolve @elizaos/plugin-goals/db/goals-repository without a built dist
## Defect

`plugins/plugin-goals/test/goals.real-db.test.ts` fails as a suite in the plugin lanes on `develop` (tip `ce68dbca425412131bd2d042b2cc4ffa6c8a61f7`; `tests / Plugin Tests (4/4)` and `canonical / Tests (plugins)` in validation run 34694307489):

```text
 FAIL  test/goals.real-db.test.ts [ test/goals.real-db.test.ts ]
Error: Cannot find package '@elizaos/plugin-goals/db/goals-repository' imported from /home/runner/work/eliza/eliza/plugins/plugin-personal-assistant/src/lifeops/repository.ts
 ❯ ../plugin-personal-assistant/src/lifeops/repository.ts:52:1
```

## Cause

The suite drives `plugin-personal-assistant`'s LifeOps repository, which since `6b9efdcafc72` (refactor: partition oversized runtime, lifecycle, repository, and build modules, 2026-09-06) imports `GoalsRepository` from `@elizaos/plugin-goals/db/goals-repository`. That subpath resolves through the package `exports` map to `./dist/db/goals-repository.js` unless the `eliza-source` condition is active, and the plugin lane never builds `plugin-goals`. `plugins/plugin-goals/vitest.config.ts` exists precisely to anchor the leaf DB subpaths this test graph reaches to source (`plugin-goals/db/schema`, `plugin-finances/db/finances-repository`, the `plugin-health` leaves, ...), but the new `db/goals-repository` import was never added, so the suite only passes on a checkout that happens to have a built `plugin-goals/dist`.

Reproduced locally by pointing the workspace link for `@elizaos/plugin-goals` at a dist-less checkout: identical error; with a built dist the same suite passes, which is why it looks green on developer machines.

## Fix

Anchor `@elizaos/plugin-goals/db/goals-repository` to `src/db/goals-repository.ts` in `plugins/plugin-goals/vitest.config.ts`, next to the existing `db/schema` alias.

hermesagent270-commits:
# Relates to

Fixes #31158

- [x] A reviewer can confirm the change from the dist-less run below without reading the code.

# Sync with develop

# Risks

None to runtime behavior: one Vitest alias in `plugins/plugin-goals/vitest.config.ts`, following the file's existing pattern for leaf DB subpaths.

# Background

## What does this PR do?

`goals.real-db.test.ts` drives the personal-assistant LifeOps repository, which imports `GoalsRepository` from `@elizaos/plugin-goals/db/goals-repository` since `6b9efdcafc72`. Without the `eliza-source` condition that subpath resolves to `dist`, and the plugin lane never builds `plugin-goals`, so the suite has failed with `Cannot find package` on every develop run while passing on checkouts that carry a built dist. The alias anchors the leaf to `src/db/goals-repository.ts`, next to the existing `db/schema` anchor.

Test-configuration repair for a suite that is red at the `develop` tip.

# Testing

The new alias entry in `plugins/plugin-goals/vitest.config.ts`.

## Detailed testing steps

Reproduced by pointing the workspace link for `@elizaos/plugin-goals` (root and `plugin-personal-assistant`'s package-local one) at a checkout with no `plugins/plugin-goals/dist`:

```text
cd plugins/plugin-goals && bunx vitest run test/goals.real-db.test.ts
# before: Error: Cannot find package '@elizaos/plugin-goals/db/goals-repository' imported from .../plugin-personal-assistant/src/lifeops/repository.ts
#         Test Files  1 failed (1)   Tests  no tests
# after:  Test Files  1 passed (1)   Tests  5 passed (5)
bunx biome check vitest.config.ts   # Checked 1 file. No fixes applied.
```

# Evidence Gate

- [x] N/A - Vitest alias configuration only; there is no user-facing flow.

  <details><summary>vitest output: plugins/plugin-goals test/goals.real-db.test.ts without a plugin-goals dist, before and after the alias</summary>

  ```text
  $ cd plugins/plugin-goals && bunx vitest run test/goals.real-db.test.ts   # workspace link points at a checkout with no plugins/plugin-goals/dist
  BEFORE (develop):
   Error: Cannot find package '@elizaos/plugin-goals/db/goals-repository' imported from .../plugin-personal-assistant/src/lifeops/repository.ts
   Test Files  1 failed (1)
        Tests  no tests
  AFTER (this head):
   Test Files  1 passed (1)
        Tests  5 passed (5)
  $ bunx biome check vitest.config.ts
   Checked 1 file. No fixes applied.
  ```
  </details>

- [x] N/A - Vitest alias configuration only; no runtime data or database rows are produced by the change.

### DIFF
--- a/plugins/plugin-goals/vitest.config.ts
+++ b/plugins/plugin-goals/vitest.config.ts
@@ -50,6 +50,12 @@ export default defineConfig({
         replacement: sourceOf("src/db/schema.ts"),
       },
       {
+        // PA's lifeops repository imports this package's own repository leaf;
+        // the plugin lane never builds dist, so it must resolve to source too.
+        find: /^@elizaos\/plugin-goals\/db\/goals-repository$/,
+        replacement: sourceOf("src/db/goals-repository.ts"),
+      },
+      {
         find: /^@elizaos\/plugin-reminders\/db\/schema$/,
         replacement: sourceOf("../plugin-reminders/src/db/schema.ts"),
       },


attentionhead (approved):
Verified at head `cb031693f9f4e8adcdddc9b6aba4af00d36740bf` (base `develop`, 1 file, +6/-0), reading the fetched head rather than the diff alone.

## The alias resolves something real

- The target exists: `plugins/plugin-goals/src/db/goals-repository.ts` is present at this head, and `GoalsRepository` is a named export (`export class GoalsRepository` at `:128`). An alias pointing at a missing file or a default-only module is the one way a change this small breaks, and neither applies.
- The consumer matches the pattern exactly: `plugin-personal-assistant/src/lifeops/repository.ts:52` imports `{ GoalsRepository } from "@elizaos/plugin-goals/db/goals-repository"` — the bare specifier the new regex anchors, with no trailing segment or extension to defeat the `$`.
- The new entry is reachable. Vite's array aliases are first-match-wins, and the only other `plugin-goals` patterns in the list are the exact-match `…/db/schema$` regex above it and whatever `...baseAliases` contributes — which is spliced in **last** and therefore cannot shadow it. There is no broader `^@elizaos\/plugin-goals\/(.+)$` catch-all that would swallow this subpath first.

## The suite is actually collected, so the fix is not inert

Worth checking on a test-config repair, because the neighbouring exclusion looks like it should apply: `exclude` carries `**/*.real.test.ts`, and the file is `test/goals.real-db.test.ts`. It is **not** excluded — `.real-db.test.ts` does not end in `.real.test.ts` — and it matches `include`'s `test/**/*.{test,spec}.{ts,tsx}`. So this config genuinely runs the suite the PR says it repairs, and the `*.real.test.ts` carve-out for PGLite-booting tests is untouched.

## On the stated cause — precise version

The package's `exports` map does carry an `eliza-source` condition for this subpath, via the `"./*"` wildcard (`eliza-source.import: "./src/*.ts"`, which for `./db/goals-repository` resolves to `./src/db/goals-repository.ts`). So it is not that the condition is missing from the manifest; it is that **this lane's resolver never requests it**. `packages/scripts/vitest/default.config.ts` — the base this config spreads — sets no `resolve.conditions`, so resolution falls through the map to `import: "./dist/*.js"`, and with the plugin lane never building `plugin-goals` that is the `Cannot find package` in the reproduction. `integration.config.ts:282` is the one config in the tree that does set `conditions: ["eliza-source"]`, which is what makes the contrast visible.

That reading is consistent with how the sibling lane already documents itself: `plugin-personal-assistant/vitest.config.ts` carries the same reasoning verbatim at `:517` ("vitest has no eliza-source condition"), `:550`, and `:594-595` ("the package `exports` map only sends subpaths to ./src under the eliza-source condition"). So the diagnosis matches an established, independently-written account of the same mechanism rather than resting on this PR's own framing.

One wording note, not a defect: the block comment this entry joins says these subpaths "carry no `bun` export condition", which is about a different condition than the `eliza-source` one the PR body cites. That text is pre-existing and the new entry's own comment is accurate, so nothing needs changing here — flagging only so the two aren't read as the same claim.

## Observation (non-blocking)

This is now the fifth leaf anchored one at a time in this file — `plugin-goals/db/schema`, `plugin-reminders/db/schema`, `plugin-inbox/db/schema`, `plugin-finances/db/finances-repository`, the two `plugin-health` leaves, `plugin-browser/schema`, and now `plugin-goals/db/goals-repository`. Each is correct, and each was added only after a suite went red, because the failure mode is silent until someone imports a new leaf: the next server-safe module PA's repository reaches for will break this lane exactly the same way.

Setting `resolve.conditions: ["eliza-source"]` on this config — precisely what `integration.config.ts:282` already does — would resolve the whole class through the manifest's existing wildcard rather than one regex per module. I am not asking for it here: it is a wider blast radius than a red-suite repair should carry, and it would change resolution for every workspace package this config touches, not just the two plugin-goals leaves. But it seems worth deciding deliberately rather than accruing the eighth alias. Is there a known reason the plugin lane cannot enable that condition the way the integration lane does?

Nothing blocks. The alias follows the file's established pattern, points at a real named export, is reachable in the alias order, and repairs a suite that this config demonstrably collects.

Approving.

hermesagent270-commits:

```text
```

ss251 (changes requested):
cb031693f9f4e8adcdddc9b6aba4af00d36740bf

This head adds a single Vitest `find`/`replacement` in `plugins/plugin-goals/vitest.config.ts` so `@elizaos/plugin-goals/db/goals-repository` rewrites to `src/db/goals-repository.ts`. It mirrors the existing `db/schema` alias a few lines above. That is the whole change.

Finding: the claimed consumer is not in this diff. The new comment (vitest.config.ts, inserted after the schema alias) states that "PA's lifeops repository imports this package's own repository leaf" and the title points at #31158, but no file here imports `@elizaos/plugin-goals/db/goals-repository`, and no test is added that would fail if `find` or `replacement` were wrong. Input: a typo in `sourceOf("src/db/goals-repository.ts")`, a missing `src/db/goals-repository.ts`, or deleting the alias. Wrong output: this plugin suite can still pass; the rewrite is unobserved at this head. A harness alias with no in-head importer is not acceptance evidence.

Finding: this is a split of the keyless real-db graph. A six-line test-config rewrite that only matters if another PR imports the leaf should land with that importer, not as a standalone merge. Nothing in the hunk changes product behavior.

The diff also does not include the `package.json` export map, so I cannot tell whether production resolution of this subpath exists or whether the alias is papering over a missing export. That is unresolved from this head, not a reason to merge it.

Verified by reading the unified diff at this head (one hunk, vitest.config.ts only). No suite applies to an alias with no in-head importer.

Recommendation: close

— [free-xai-lane-ss251]

hermesagent270-commits:
Answering the static review at `cb031693f9f4`; head is now `3a9efb2e3e20` (one added test, no other change).

**The importer is on develop, not in this diff.** `plugins/plugin-personal-assistant/src/lifeops/repository.ts:52` imports `@elizaos/plugin-goals/db/goals-repository` (added in `6b9efdcafc72`), and this package's own `test/goals.real-db.test.ts` drives that repository, so the plugin lane already exercises the subpath. That lane never builds `dist`, which is exactly why the suite fails on develop (#31158, the CI log in the issue) and why the config-only change is a standalone repair rather than half of a split: there is nothing else to land with it.

**Production resolution exists.** `plugins/plugin-goals/package.json` exports `"./*"` as `./dist/*.js` (types `./dist/*.d.ts`, and `./src/*.ts` under the `eliza-source` condition), so the subpath resolves in every built install. The alias covers only the vitest lane, the same way the existing `db/schema` line above it does; the shared vitest config does not apply the `eliza-source` condition to workspace siblings, so each leaf is anchored explicitly.

**The alias is now observed in-head.** `src/db/goals-repository-subpath.test.ts` imports the subpath and `./goals-repository.js` side by side and asserts `GoalsRepository` / `createGoalDefinition` are the same module instance. Runs from `plugins/plugin-goals` with `bunx vitest run src/db/goals-repository-subpath.test.ts`:

| config | checkout | result |
| --- | --- | --- |
| this PR | dist-less (plugin lane shape) | 1/1 passed |
| develop's `vitest.config.ts` | dist-less | `Error: Cannot find package '@elizaos/plugin-goals/db/goals-repository'` |
| develop's `vitest.config.ts` | stale `dist` present | `AssertionError: expected [Function GoalsRepository] to be [Function GoalsRepository] // Object.is equality` |

So a wrong `find`, a wrong `replacement`, or a deleted alias now fails this package's default suite in both checkout shapes, without needing the PGlite graph.

mashingaan (changes requested):
## Changes requested at exact head `3a9efb2e3e20edb6ff59934a6aed7db4bd09a5e3`

I reviewed the current head against `ce68dbca425412131bd2d042b2cc4ffa6c8a61f7`.

## Blocking: the new test fails the repository lint gate

`plugins/plugin-goals/src/db/goals-repository-subpath.test.ts` is collected by the plugin suite, but the exact-head Source static smoke reports two Biome errors in that file:

1. `assist/source/organizeImports` requires the `@elizaos/plugin-goals/db/goals-repository` import before the `vitest` import.
2. The formatter requires the `createGoalDefinition` expectation to use the multiline layout shown by Biome.

The CI log is explicit: `plugin-goals:lint:check` reports `Found 2 errors`, then exits with code 1. `All Tests Passed` fails because it aggregates that source-smoke result. The alias itself is relevant and the identity test is a useful regression guard, but the PR cannot pass the repository gate in its current form.

Please apply the two safe Biome fixes, rerun the plugin lint and the full Source static smoke, then request re-review at the new exact head. The expected import order is the package subpath import followed by the `vitest` import. The long `createGoalDefinition` assertion should be formatted across multiple lines.

The PR remains open and mergeable. Billing payment replay E2E, Subscription authority PostgreSQL, and Browser bridge Windows security are green. No other blocking issue was found in this review.

— [eliza-pr31159-review]
\nCompute receipt: 0 project-attributed tokens (unavailable; device-signed, locally reported)
— [eliza-pr31159-review]

lalalune:

Retaining the targeted source-resolution repair. The alias follows the existing exact DB-leaf pattern and restores a real repository test when dist is absent. Keep the actual dist-less goals.real-db run as the main evidence, plus a stale-dist control where relevant; comparing two imported symbol identities alone does not prove the repository behavior. Refresh the description because the current diff includes a new test as well as the config entry. Treat this as a test-runner alias, not proof that the published package export or built consumer path works; preserve their independent package validation.

hermesagent270-commits:
Merged `origin/develop` (`1eba34ba95`) at `8eae422f74d07b59e1afc602ed408b30214f4774`; the only conflict was the alias list in `plugins/plugin-goals/vitest.config.ts` (develop added the plugin-calendar aliases next to this PR's goals-repository alias; both kept). `bunx vitest run src/db/goals-repository-subpath.test.ts` → 1/1, Biome clean on the config.

hermesagent270-commits:
Applied the two Biome fixes from the 2026-09-14 review (import order: package subpath import before `vitest`; multiline `createGoalDefinition` expectation) in `b4f97411f7` on top of the develop merge. `bunx biome check src/db/goals-repository-subpath.test.ts vitest.config.ts` is clean and the test still passes 1/1. Requesting re-review at this head.

mashingaan (changes requested):
REQUEST_CHANGES at exact head `b4f97411f707c6b984cfc49080e369a4d79fbf42`

The current PR body does not satisfy the repository evidence contract. I ran `scripts/check-pr-evidence.mjs` against the live PR body, the changed-file path, and the exact current head. It found no `evidence-row:*` markers at all, so every required row is missing, and it also found an evidence-head mismatch against the current head.

Please restore the repository evidence-row markers and fill every required row with reviewed artifacts or explicit repository-allowed `N/A - <reason>` entries. Replace the stale or absent evidence-head marker with the complete 40-character current head SHA, rerun the evidence validator, and request re-review. Without these rows, the claimed goals-repository graph behavior and its verification cannot be checked as a complete evidence package.

This review is limited to the evidence-contract blockers. It does not claim that the test change is correct or incorrect.

Client / agent tooling: codex
— [eliza-pr31159-review]

hermesagent270-commits:
Evidence contract repaired at the current head `b4f97411f707c6b984cfc49080e369a4d79fbf42` (no code change): the body now carries every `evidence-row:*` marker with either an inline transcript of the runs already described in the PR or an explicit `N/A - <reason>`, and the `evidence-head` marker is bound to this head. Local run of `node scripts/check-pr-evidence.mjs --body-file <live body> --changed-files-file <all changed paths> --added-files-file <added paths> --head b4f97411f707c6b984cfc49080e369a4d79fbf42` reports `Evidence gate passed: all required rows satisfied.` Ready for re-review.

lalalune:
Closing as superseded by #31614, whose merge 9ce803264c17cfb6b8477d50d082a3be04fe34a1 is an ancestor of current develop 35da9587829513f7db66087f287cc0809d7237ae.

Reviewed the complete current PR diff at b4f97411f707c6b984cfc49080e369a4d79fbf42: only the new goals-repository-subpath.test.ts remains. The actual source-resolution repair is already present in develop: explicit goals-repository, CalendarRepository and calendar schema aliases in plugins/plugin-goals/vitest.config.ts (introduced by 19b99f6a158, duplicate cleanup 667aa877d46). The retained test/goals.real-db.test.ts imports the personal-assistant repository through that graph and exercises real PGlite goal CRUD, audit records, deduplication and linked-record deletion.

The remaining proposed test compares exported function/class identities without exercising those consumers; it adds no unique product behavior and is redundant under the repository test quality gate. No source fix is being discarded. This closure is source-integration verification, not a new live family-workflow acceptance claim.

PR #31161, opened 2026-09-12
fix(cloud): return agent identity for Dedicated session recovery

### LINKED ISSUE #31140: fix(app): recover Dedicated proxy credential rejection on reload

NubsCarson:
Fixes #31140 together with PR #31141.

A returning Dedicated user can reach authenticated session recovery but still see “Open this agent from Eliza Cloud.” Live staging returned HTTP 200 from `/api/auth/pair/native` with a credential and name but no agent ID. The app's shared session parser therefore rejected the response.

Return the agent ID from the atomically claimed pairing-token record and type the response with the existing shared contract. The regression now passes the real Hono response through the same parser used by the app; it failed before the change. This also covers native consumers of the same endpoint. Existing Cloud authorization, ownership, origin binding, single-use consumption, and no-store headers remain in force.

Validation of fix `3f333f53a1f01e5d3e5784647049ca130b3c3f34` and synced head `2e06f65287f598151074be0b76dbad1b9283cb8a`:

- 171 tests pass in isolated processes: 16 pairing-route, 136 auth-middleware, 12 public-path, and 7 real PGlite native-claim integration tests. Wrong-identity, expiry, retry, and concurrent-claim cases pass.
- The initial combined-process invocation had 8 failures from module-mock contamination; every affected file passed its required isolated rerun. This did not require a source change outside the two-file fix.
- `bun run verify` passed under Node 24.15.0 / Bun 1.3.14, including API types, route contract, Worker dry-run bundle, lint and repository audits. Frozen install reported no dependency changes; install scripts were omitted for the already-built unchanged graph.
- The final head preserves the newly merged develop agreement-pins change (#31099) through an ordinary merge. Both fix files are byte-identical to the tested fix commit, and full `bun run verify` passed again on the combined tree. Hosted PR CI is still running and is not claimed as passed.
- Live staging before evidence and the regression result are attached. Hosted recovery after this server correction is pending deployment. The prior client fix is already live in staging; production has not received either reload fix.

Risk is confined to the pairing response contract. This is an additive identity field taken from the authorized claim, with no auth bypass, database migration, agent restart, or infrastructure change. Release staging first; verify real desktop/mobile reload and full conversation continuity before production promotion.

- [x] N/A - backend response-only correction; no rendered source changed.

- [x] N/A - backend response-only correction; actual hosted reload remains a deployment acceptance gate.

- [x] N/A - no rendered or native implementation changes in this response contract fix.

- [x] [31161-pair-response-public-backend.log](https://github.com/elizaOS/eliza/releases/download/pr-evidence-12/31161-pair-response-public-backend.log)

- [x] [31161-pair-response-public-frontend-records.json](https://github.com/elizaOS/eliza/releases/download/pr-evidence-12/31161-pair-response-public-frontend-records.json)

- [x] N/A - no model, prompt, provider, or action change.

- [x] [31161-pair-response-public-synced-regression.json](https://github.com/elizaOS/eliza/releases/download/pr-evidence-12/31161-pair-response-public-synced-regression.json)

- [x] N/A - no rendered component or layout change.

### DIFF
--- a/packages/cloud/api/auth/pair/route.test.ts
+++ b/packages/cloud/api/auth/pair/route.test.ts
@@ -1,5 +1,6 @@
 /** Exercises loopback browser and authenticated native Cloud pairing exchanges. */
 import { beforeEach, describe, expect, mock, test } from "bun:test";
+import { parseCloudPairRelaySession } from "@elizaos/shared/contracts";
 import { Hono } from "hono";
 import { AuthenticationError } from "@/lib/api/errors";
 import type { AppEnv } from "@/types/cloud-worker-env";
@@ -295,10 +296,10 @@ describe("Cloud pairing route", () => {
       agentId: AGENT_ID,
       expectedOrigin: EXPECTED_ORIGIN,
     });
-    await expect(response.json()).resolves.toEqual({
-      message: "Paired successfully",
+    expect(parseCloudPairRelaySession(await response.json())).toEqual({
       apiKey: "agent-api-token",
       agentName: "Native agent",
+      agentId: AGENT_ID,
     });
   });
 
--- a/packages/cloud/api/auth/pair/route.ts
+++ b/packages/cloud/api/auth/pair/route.ts
@@ -253,15 +253,15 @@ app.post("/native", async (c) => {
       );
     }
 
-    return c.json(
-      {
-        message: "Paired successfully",
-        apiKey: claim.apiKey,
-        agentName: claim.agentName ?? "Agent",
-      },
-      200,
-      { "Cache-Control": "no-store, no-cache, must-revalidate" },
-    );
+    const response: CloudPairExchangeResponse = {
+      message: "Paired successfully",
+      apiKey: claim.apiKey,
+      agentName: claim.agentName ?? "Agent",
+      agentId: claim.pairingToken.agentId,
+    };
+    return c.json(response, 200, {
+      "Cache-Control": "no-store, no-cache, must-revalidate",
+    });
   } catch (err) {
     // error-policy:J1 route boundary — translate authentication and dependency
     // failures into structured HTTP errors without exposing token context.


attentionhead (approved):
Reviewed at head `3f333f53a1f01e5d3e5784647049ca130b3c3f34` (base `develop`, head `codex/dedicated-pair-response-20260912`), verifying the diff against the fetched PR head rather than the PR description.

## Verdict

The fix is correct, minimally scoped, and the regression it adds is genuinely load-bearing. Approving.

## What the change does, and why it is not vacuous

`POST /api/auth/pair/native` returned `{ message, apiKey, agentName }`. The app's shared parser, `parseCloudPairRelaySession` (`packages/shared/src/contracts/cloud-pair.ts:159`), rejects any payload where `isCloudPairAgentId(record.agentId)` fails — and a missing field fails it. So the endpoint returned HTTP 200 with a usable credential that the client contract then discarded, which is exactly the "Open this agent from Eliza Cloud." symptom described. The diff adds `agentId: claim.pairingToken.agentId` to the native response.

The rewritten assertion in `route.test.ts:299` is not a tautology. It replaces a literal-shape `toEqual` on the raw JSON with `parseCloudPairRelaySession(await response.json())`. Revert the one-line source change and the parser returns `null`, so `expect(null).toEqual({ apiKey, agentName, agentId })` fails. The test exercises the real Hono response through the same parser the app uses, so it constrains the contract rather than restating the handler.

## Verification performed at head

- **Import is present.** The diff annotates two returns as `CloudPairExchangeResponse` without adding an import line, which is the first thing that could be blocking. It is already imported at `route.ts:10-13` (`type CloudPairExchangeResponse, isCloudPairAgentId` from `@elizaos/shared/contracts`), because the browser branch at `:131` already used it. Control grep on the same fetched bytes confirmed the pipeline (`AuthenticationError`, 2 hits).
- **The field exists on the claim type.** `route.ts:260` reads `claim.pairingToken.agentId`. `AuthenticatedNativePairingClaim`'s `status: "claimed"` branch declares `pairingToken: PairingToken` (`packages/cloud/shared/src/lib/services/pairing-token.ts:43-46`), constructed at `:237-238` via `toPairingToken(claim.token)` — structurally identical to `BrowserPairingClaim` at `:26-29`/`:198-199`. The read is against a real field on the narrowed success branch, reached only after the `sandbox-credential-unavailable` and `invalid` branches return.
- **The contract requires the field.** `CloudPairExchangeResponse` declares `agentId: string` as non-optional (`cloud-pair.ts:14-19`). Both return sites now satisfy it, so the annotation is doing work rather than decorating.
- **The parser's strict identity check cannot reject the new value.** `isCloudPairAgentId` enforces a narrow UUID pattern (`cloud-pair.ts:11-12`), and `cloudPairTokenKeyForAgent`'s own comment notes that historic non-UUID identifiers exist on the browser-storage path — so "returns an id" and "returns an id the parser accepts" are separable, and I checked the second. The native handler already gates the request on `!isCloudPairAgentId(agentId)` at `route.ts:214` and passes that same validated `agentId` into `claimAuthenticatedNativeToken`, which binds consumption to it. The returned id therefore corresponds to an already-validated UUID, and the round-trip through the parser cannot fail on format. I had this queued as a blocking finding and it does not survive.

## Consistency note

This makes `/native` match the browser branch, which has returned `agentId` since `:131-135`. The asymmetry between the two branches of the same route was the defect; after this change both construct the same typed contract.

## Non-blocking observation

The response still carries `message: "Paired successfully"`, which `CloudPairExchangeResponse` requires but `parseCloudPairRelaySession` ignores entirely — the relay contract (`CloudPairRelaySession`) has only `apiKey`, `agentId`, and optional `agentName`. That is harmless today, but it means the wire type and the parsed type have drifted by one field that no consumer reads. Worth a follow-up only if someone is already touching this contract; it is not a reason to hold this fix.

## Open question, not blocking

The PR body notes 8 failures in the initial combined-process invocation attributed to module-mock contamination, each passing on isolated rerun. That is consistent with the two-file scope here and I found nothing in the diff that would cause cross-file mock bleed. Is the contaminating suite tracked anywhere, so the combined-process run is not left permanently red for the next person who runs it without the isolation flags?

Issue #31162, opened 2026-09-12
feat(lifeops): export agreement originals with provenance and audit history


lalalune:
## Observed gap

Bettina's PDD KNOW-8 requires the owner to export the original agreement, reviewed obligations, hashes, pins, grants and audit history. The verification plan requires an independently re-hashed manifest export. Current `AgreementKnowledgeService.readOwnerPdf` and `GET /api/lifeops/agreements/:id/download` return only the verified PDF bytes. `FamilyOperationsView` has review, pin and guest controls but no export journey. There is no agreement export manifest or agreement mutation audit trail in the current service.

## Acceptance

- Provide an owner-authenticated export for one immutable agreement version containing the exact original PDF and a versioned manifest of source metadata/page map, all obligation review states and citations, pins, grants and available audit history. Do not expose credentials or infer guest permission from a pin.
- Verify original byte length and SHA-256 before export; unavailable or mismatched bytes fail visibly. Provide a manifest integrity contract that an independent reader can verify.
- Reuse the canonical audit ledger for agreement upload/extraction, review, pin/unpin, grant/revoke and export events. Preserve actor, artifact version/content identity and state transitions. Do not fabricate missing legacy history; represent its coverage explicitly.
- Keep the manifest internally consistent during concurrent review, pin and grant changes. Do not add another file store, scheduler or shadow audit system.
- Expose the download through the real Family Operations UI with visible loading/error states and an original-download option. Exercise denied-principal and corrupt/missing-storage paths.
- Real PGlite regression, owning package gates, root verification and exact-head hosted checks pass. On the pilot, export a synthetic agreement after review/pin/grant activity, independently hash and inspect its contents, and attach sanitized UI/network/video evidence.

This implements the missing KNOW-8 journey under #30123. Broader workspace export (AUDIT-4), dependency-aware deletion (AUDIT-5/AUDIT-7), verified guest live acceptance and final merged MVP deployment remain separate acceptance requirements; this issue does not stand in for them. Selected-correspondence intake remains owned by #31063/#31032.

lalalune:
Owner export implementation is merged in #31165 (`a93120ebb0cbc3e079ed0d1fdd2045827209b095`). All 14 changed file blobs match the validated source. Corrected combined pilot `58cd2df949d37627131de2427ca043e218385e09` passed all seven Linux steps and is deployed; its public entry hash matches, and restart preserved agreement data, calendar controls, the monthly packet, and the original messaging process.

[...]

PR #31165, opened 2026-09-12
feat(lifeops): export verified agreement provenance and audit history

lalalune:
Agreement owners can download the original PDF or export one immutable version with its reviewed obligations, citations, pins, grants, and provenance. Previously the agreement surface had no export and review/access mutations did not retain agreement-specific audit history.

The ZIP contains `original.pdf`, `manifest.json`, `SHA256SUMS`, and the exact saved `extraction.json` when available. The extraction file remains independently verifiable against its ingestion audit hash. Export verifies the original bytes and the saved extraction hash, reads mutable records from one database snapshot, and records archive preparation in the canonical LifeOps audit ledger. Ingestion, review, pin, and resource-grant transitions now commit with their audit event atomically. Legacy versions explicitly report missing page maps and incomplete historical coverage. Owner-authenticated download controls expose preparation and errors.

Refs #31162. Part of #30123; workspace-wide export/deletion and the remaining provider-backed MVP acceptance remain separate.

Validation at `f00d7d788f7e147efa0b8852d837af416d16e208`:

- Pinned Bun 1.3.14 / Node 24.15.0 installation passed.
- Root `bun run verify`: 373/373 tasks passed (4m37.745s), including package typecheck and lint.
- Real PGlite agreement integration: 12/12 passed. Covers atomic rollback, concurrent snapshots, corrupt/missing originals, altered extraction, legacy coverage, denied guest export, and export of a revoked resource grant with its household grant and audit history.
- Adapter HTTP tests: 14/14 passed. HTML, plain-text, and malformed JSON error responses retain the HTTP failure instead of surfacing a JSON parsing error.
- Full personal-assistant suite: 2,654 passed, six skipped, 303 files (527.79s). App audit: 226 passed; 220 findings, zero broken/needs-work, zero hover/density/minimalism failures; OCR 208 verified, zero broken, 12 requiring visual review.
- The earlier isolated browser harness passed and its desktop/mobile controls were inspected at rest and hover. Live before-deployment desktop/mobile captures are also inspected; final live export downloads and archive contents are independently verified.

All hosted checks passed for this exact source revision. Combined pilot `58cd2df949d37627131de2427ca043e218385e09` passed all seven Linux gates and is deployed. Its public assets match the build; agreement/packet/control state and the original messaging-service process survived restart. Fresh owner API and authenticated browser exports independently verify the original PDF bytes, raw extraction, page map, three review states, inactive pin, and canonical audit history. The preceding pilot video records upload, review, pin/unpin, original download, and export; final desktop/mobile captures and downloads verify the corrected revision.

Live guest transport acceptance remains outstanding under #31172; #31162 stays open until its live grant-history acceptance is complete. No real guest was contacted. Broader MVP acceptance remains in #30123.

[Reviewed deployment evidence](https://github.com/elizaOS/eliza/releases/download/pr-evidence-12/bettina-agreement-export-58cd2df-evidence.zip): before/after desktop/mobile captures, scoped MP4 walkthrough, sanitized network/console and canonical audit records, independently verified synthetic ZIPs, and final deployed-revision download checks. Archive SHA-256: `398f820850c9a87d676685b300744f4868e3218b3a9caa049be436522f0fe5c1`.

### DIFF
--- a/plugins/plugin-personal-assistant/README.md
+++ b/plugins/plugin-personal-assistant/README.md
@@ -232,7 +232,7 @@ for every page with native text plus rendered-page vision transcription for
 images or text-empty pages. One failed page fails ingestion; partial content is
 never published as a complete owner-private `DocumentService` record. LifeOps
 persists the media SHA-256, handle, document id, byte size, MIME type, filename,
-parser-derived page count, complete extracted text, and version chain; it does
+parser-derived page count, complete extracted text and page map, and version chain; it does
 not create another permanent file store.
 
 Owner- or agent-extracted obligations begin as `proposed` and must carry an
@@ -252,6 +252,21 @@ authenticated `/api/lifeops/agreements/*` routes. Grant previews enumerate the
 exact read effects and exclusions before issuance. Active agent/chat pins feed
 only approved, page-cited obligations into owner planner context; uploading a
 PDF remains on the document/API surface so chat actions never invent bytes.
+
+Owners can download the original PDF or export one agreement version from the
+Agreement view. `POST /api/lifeops/agreements/:id/export` returns a ZIP containing
+`original.pdf`, `manifest.json`, and `SHA256SUMS`. Verify both files independently
+with `sha256sum -c SHA256SUMS`. The versioned manifest includes all obligation
+states and citations, current and inactive pin/grant records, linked household
+grant expiry/revocation records, and the canonical agreement audit history from
+one database snapshot. Original byte length and SHA-256 must match before export.
+
+Agreement ingestion, review, pin, and grant transitions commit atomically with
+`life_audit_events`. Export preparation records the manifest and archive hashes;
+it does not assert that the browser received or saved the download. Legacy
+versions explicitly identify missing extraction page maps and partial audit
+history. Export never reconstructs unrecorded past activity. This agreement
+archive does not replace workspace-wide export or deletion workflows.
 
 ## Default packs
 
--- a/plugins/plugin-personal-assistant/README.md
+++ b/plugins/plugin-personal-assistant/README.md
@@ -255,8 +255,10 @@ PDF remains on the document/API surface so chat actions never invent bytes.
 
 Owners can download the original PDF or export one agreement version from the
 Agreement view. `POST /api/lifeops/agreements/:id/export` returns a ZIP containing
-`original.pdf`, `manifest.json`, and `SHA256SUMS`. Verify both files independently
-with `sha256sum -c SHA256SUMS`. The versioned manifest includes all obligation
+`original.pdf`, `manifest.json`, `SHA256SUMS`, and the exact saved
+`extraction.json` when available. Verify the files independently with
+`sha256sum -c SHA256SUMS`; the extraction hash also matches its ingestion audit
+event. The versioned manifest includes all obligation
 states and citations, current and inactive pin/grant records, linked household
 grant expiry/revocation records, and the canonical agreement audit history from
 one database snapshot. Original byte length and SHA-256 must match before export.
--- a/plugins/plugin-personal-assistant/src/components/family-operations/FamilyOperationsView.test.tsx
+++ b/plugins/plugin-personal-assistant/src/components/family-operations/FamilyOperationsView.test.tsx
@@ -131,6 +131,7 @@ function adapter(data = snapshot()): FamilyOperationsAdapter {
     approveSchoolDiff: vi.fn(async () => undefined),
     generatePacket: vi.fn(async () => undefined),
     uploadAgreement: vi.fn(async () => undefined),
+    downloadAgreement: vi.fn(async () => new Blob()),
     createPacketDraft: vi.fn(async () => undefined),
     revisePacketDraft: vi.fn(async () => undefined),
     requestPacketApproval: vi.fn(async () => undefined),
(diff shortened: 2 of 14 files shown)


attentionhead (approved):
Reviewed at head `ced5d360bba53b3f9944b1bd2348088796f7e1f2` (base `develop`, head `fix/bettina-agreement-export`), verifying the diff against the fetched PR head rather than the PR description. Reviewing while draft.

## Verdict

**Approve.** The two claims that carry this PR — that a failed audit rolls the mutation back, and that the export attests bytes it has actually verified — both hold at this head, and each rests on a fact outside the diff that I checked rather than assumed.

## Atomicity

`agreementMutationSql` wraps the mutation in a data-modifying CTE and returns `SELECT changed.* FROM changed CROSS JOIN recorded`. In Postgres a data-modifying CTE executes exactly once whether or not the outer query reads its output, so `CROSS JOIN recorded` on its own would **not** roll anything back — an empty `recorded` would just hide a committed mutation from the caller.

What makes the guarantee real is the NOT NULL constraint the comment points at, and it is genuinely there: `schema.ts` declares `ownerId: text("owner_id").notNull()`. A missing or foreign artifact makes the `owner_id` scalar subquery NULL, which raises a not-null violation, which aborts the whole statement including the CTE. That is a true rollback, not a suppressed result. Worth stating explicitly because the mechanism is load-bearing and entirely implicit in the SQL.

## Export integrity

`SHA256SUMS` lists `snapshot.artifact.contentSha256` for `original.pdf` rather than rehashing the bytes being written, which would normally attest stored metadata instead of shipped content. It is sound here because `exportOwnerAgreement` obtains those bytes only through `readOwnerPdf`, which rehashes the storage read and rejects on mismatch of **both** `contentSha256` and `byteSize` before returning. The bytes in the archive are therefore proven to hash to the value the checksum file claims.

The extraction side is verified symmetrically: when an `agreement_ingested` event exists, its recorded `extractionSha256` is re-derived from the live `agreementExtractionJson` and mismatches raise `AGREEMENT_INVALID_CONTRACT` under `error-policy:J2`, so a silently altered extraction cannot be exported as provenance. Legacy artifacts without a page map degrade to an explicit `status: "unavailable"` with a stated reason rather than an empty field, and `auditCoverage` distinguishes `recorded_from_ingestion` from `partial_legacy_history` instead of implying complete history.

## Test wiring

I checked that the new 282-line PGlite suite is actually collected, since a `.integration.test.ts` under this package sits between two lanes. The unit lane's `defaultUnitExcludes` drops `**/*.integration.test.{ts,tsx}`, and `vitest.src-integration.config.ts` includes several specs by name — but its **first** include entry is the glob `src/**/*.integration.test.{ts,tsx}`, and the new file is under `src/`. It runs. The by-name entries are only for `test/`-rooted specs, which is why no second config edit was needed.

The single `vitest.config.ts` edit is likewise sufficient for the new `@elizaos/agent/api/zip-utils` import: `vitest.src-integration.config.ts` does `import baseConfig from "./vitest.config"` and spreads `...baseConfig`, so it inherits both the alias at `:460` and the matching `agentSourceJsToTsPlugin` branch at `:86`. The alias is also placed ahead of the broader `@elizaos/agent` entries, which matters because these aliases are first-match-wins.

## Non-blocking

- **`crypto.randomUUID()` is evaluated once at SQL-string-build time**, so every row of a multi-row `changed` CTE would receive the same `id` — and `id` is the ledger's primary key. Not reachable today: all seven call sites are single-row by construction (`VALUES` inserts, `ON CONFLICT … DO UPDATE`, and `UPDATE … WHERE id = <pk>`). But the helper's doc comment says only "a repository-owned mutation ending in RETURNING *", with no single-row requirement stated, so the first batch mutation to use it will fail its whole statement. A `gen_random_uuid()` expression, or an explicit note in the contract, would close it.
- **`idempotency_key` is never written by the audit insert**, so `uq_life_audit_events_operation` on `(agent_id, event_type, idempotency_key)` never constrains agreement events — Postgres treats NULLs as distinct. The ledger's idempotency mechanism is simply unused on this path; that may be intended, but a retried mutation will append a second audit row rather than collide.
- **The `artifact` subquery is inlined twice** per statement, once for `owner_id` and once for `inputs_json.source`. Identical text both times, so it is only redundant evaluation, not a correctness issue.
- **The manifest spreads `...snapshot` after the literal `original` key.** No collision today — the snapshot kinds are `artifact`, `obligation`, `pin`, `grant`, `householdGrant`, `householdAudit`, `audit` — but a future kind named `original` or `extraction` would silently shadow a field the checksum file depends on. Placing the spread first would make that impossible.

ss251 (approved):
Reviewed head 8a54a6f28d122b5eb98b01a86298418db132bcd6.

Defect (non-blocking): plugins/plugin-personal-assistant/src/components/family-operations/adapter.ts:244 — `downloadAgreement` parses every non-ok response with `await response.json()`. Input: a 502/413 from a proxy with an HTML or plain-text body. Output: the owner sees the SyntaxError "Unexpected token '<'…" instead of the intended `Download failed (status)` fallback two lines below; that fallback is unreachable for non-JSON bodies, and the new adapter test stubs only `application/json` error payloads, so it passes with this path broken. Guard the parse by content-type.

Otherwise I traced the risky core by reading the diff at this head: `agreementMutationSql` commits mutation and audit row in one statement, so the trigger-rejection rollback test is real, and `readExportSnapshot` reads all export records in one UNION ALL statement, so the single-snapshot manifest claim holds.

Recommendation: merge

— [free-9router-lane-ss251]

Finberg-Laurelin-CEO (commented):
Reviewed at `f00d7d788f7e147efa0b8852d837af416d16e208`.

## Code verdict

I found no new code blocker at this head. The latest commit closes the non-JSON error-body gap without masking the HTTP status, and the new revoked-grant export assertion exercises the important historical path: the manifest keeps the resource grant, its linked household grant, the household-grant audit, and both agreement grant/revoke events after access is no longer active.

I also re-traced the export integrity boundary. `readOwnerPdf` hashes and length-checks the exact private-storage bytes before they enter the archive; the saved extraction is checked against the ingestion-event hash and exported byte-for-byte; mutable agreement/grant/audit rows come from one SQL statement snapshot; `recordExport` persists the manifest and archive hashes before the route can send a response. The archive accurately records *preparation*, not browser receipt.

## Evidence gate still open

The PR body explicitly says the final `f00d7d7` correction is not deployed and live export acceptance is still pending. This is a new owner-authenticated browser download whose claimed outcome is the produced archive, so the package’s evidence standard is not satisfied by PGlite/adapter tests alone. I’m withholding approval until the current head is exercised through the real route/UI boundary.

The decisive proof should be small and artifact-based:

1. Deploy this exact head to the isolated pilot and trigger **Export agreement** as the owner.
2. Preserve the downloaded ZIP, list its four expected entries, and run `sha256sum -c SHA256SUMS`.
3. Show that `manifest.json.exportId`, `manifestSha256`, and the ZIP hash match the new `agreement_export_prepared` ledger row.
4. Repeat once for a legacy artifact without `agreementExtractionJson` and show the explicit unavailable/partial-history fields rather than a fabricated extraction.

Recommendation: hold merge pending that current-build live evidence; no source change requested from this review.

No untrusted PR code was executed on the host. This review is based on the exact-head diff, the hosted check results, and the evidence disclosed in the PR.

Client / agent tooling: codex
— [contributor]

PR #31169, opened 2026-09-12
release: reduce Dedicated inference policy lookup round trips

NubsCarson:
Promote #31167, which removes four sequential database round trips from each organization-policy read while retaining the existing tenant scope, policy locks, entitlement validation, generation checks, and database-clock behavior. This is a measured query improvement for Dedicated inference; overall chat responsiveness remains unfinished.

The promotion changes only the policy loader and its real PGlite contract test. Source `e3ed8f7ec7f93c4d26f0dfb68189788bb6e74328`; staging `118cb5142f326322228f1cf2666a6bb850e495e4`; normal merge preview tree `55ae1987e57fbf10e77262d13f818ded4cccfdfa` exactly matches the tested and certified tree.

Validation: 14 real PGlite tests, 13 PostgreSQL tests, package typecheck/lint, and root verification passed. Read-only production comparisons returned identical complete policies, with 1495–1518 ms baseline versus 830–836 ms candidate. Source CI [34714547030](https://github.com/elizaOS/eliza/actions/runs/34714547030) completed successfully, including billing replay, Windows security, subscription PostgreSQL, and static smoke. The separate broad Develop Full run 34714584097 still has failures and is not claimed clean; their relationship to this two-file patch has not been established.

[Staging release 34714679173](https://github.com/elizaOS/eliza/actions/runs/34714679173) succeeded. Canonical certification accepted artifact `10304443711`, digest `sha256:80e0c76efc502a60cf6983daf894828fd3d0c7b0e348a75c17873d57e923954b`, expiring 2026-09-26T19:58:56.655Z.

Live acceptance on the new staging API: one real recall request returned all four saved facts. Submission to runtime finalization was 35.594 seconds (runtime 33.487 seconds); first observed complete in the browser at 42.303 seconds. The earlier staging sample was 53.418 seconds. These are individual observations, not proof that the query patch caused the entire change: cache warming and environment differences remain possible contributors. Embedding recall still took 16.218 seconds. All 35 prior messages plus the new request/reply survived a normal 390×844 reload exactly; desktop retained the same complete 37-message text. The reply and composer were visually inspected, and browser emulation and DevTools were removed afterward.

Production release [34715963361](https://github.com/elizaOS/eliza/actions/runs/34715963361) completed successfully for `a00194febec990d154a406e8b03712705499528f`, with live API commit verified. One real production recall returned the saved code and color. Submission to runtime finalization was 36.050 seconds, versus the earlier 53.978-second production sample; runtime itself was 35.319 seconds. This is an individual before/after observation, not a controlled causal estimate. Embedding recall still consumed 22.257 seconds with retries, so responsiveness remains unresolved. All ten previous messages plus the new request/reply survived ordinary desktop reload exactly, with the final reply and composer visually inspected. Earlier mobile acceptance remains recorded separately; this production request was tested at desktop width. No schema, provider, prompt, model context, provisioning worker, Dedicated runtime image, or paid capacity changes are part of this promotion. Genuine new-account signup remains unverified separately.

| Evidence | Result |
| --- | --- |
| Before screenshots  | N/A - backend query change with no rendered UI diff |
| After screenshots  | N/A - backend query change with no rendered UI diff |
| Walkthrough video  | N/A - backend query change with no rendered UI diff |
| Backend logs  | https://github.com/elizaOS/eliza/releases/download/pr-evidence-12/31167-dedicated-policy-query-public-backend.log |
| Frontend logs  | N/A - no frontend code change |
| LLM trajectory  | N/A - no model, provider, prompt, action, or runtime change; real app outcome recorded above |
| Domain artifacts  | https://github.com/elizaOS/eliza/releases/download/pr-evidence-12/31167-dedicated-policy-query-public-domain.json |
| OCR review  | N/A - backend query change with no rendered UI diff |

### DIFF
--- a/packages/cloud/shared/src/db/repositories/organization-quota-policy.pglite.test.ts
+++ b/packages/cloud/shared/src/db/repositories/organization-quota-policy.pglite.test.ts
@@ -45,6 +45,37 @@ beforeAll(async () => {
 afterAll(async () => {
   await database.closeDatabaseConnectionsForTests();
 });
+test("optional policy rows stay tenant-scoped and missing authority fails closed", async () => {
+  const pg = database.getPgliteClientForTests();
+  const target = "61000000-0000-4000-8000-000000000003";
+  const other = "61000000-0000-4000-8000-000000000004";
+  await pg.exec(`
+    INSERT INTO organizations(id,credit_balance,balance_revision,balance_decrease_revision,settings,is_active,account_lifecycle_state)
+    VALUES('${target}',100,1,0,'{}',true,'active'),('${other}',100,1,0,'{}',true,'active');
+    INSERT INTO credit_transactions(id,organization_id,amount,type,metadata)
+    VALUES(gen_random_uuid(),'${target}',100,'credit','{}'),(gen_random_uuid(),'${other}',100,'credit','{}');
+  `);
+  const before = await policy.readOrganizationQuotaPolicy(target);
+  await pg.exec(`
+    INSERT INTO org_rate_limit_overrides(organization_id,completions_rpm) VALUES('${other}',71);
+    INSERT INTO organization_config(organization_id,settings) VALUES('${other}','{"max_containers":3}');
+    INSERT INTO org_storage_quota(organization_id,bytes_used,bytes_limit,limit_override_authorized)
+    VALUES('${other}',0,13,true);
+  `);
+  const after = await policy.readOrganizationQuotaPolicy(target);
+  expect(after).toEqual({ ...before, observedAt: after.observedAt });
+  const overridden = await policy.readOrganizationQuotaPolicy(other);
+  expect(policy.requireOrganizationRateTier(overridden).completionsRpm).toBe(71);
+  expect(policy.requireOrganizationResourceLimit(overridden, "containers")).toBe(3n);
+  expect(policy.requireOrganizationResourceLimit(overridden, "storage")).toBe(13n);
+  await pg.exec(
+    `DELETE FROM organization_subscription_authorities WHERE organization_id='${other}'`,
+  );
+  await expect(policy.readOrganizationQuotaPolicy(other)).rejects.toMatchObject({
+    code: "ORGANIZATION_POLICY_UNAVAILABLE",
+    context: { organizationId: other, reason: "missing_account_authority" },
+  });
+});
 test("subscriber RPM survives balance spending and unapproved resource ceilings remain unavailable", async () => {
   const before = await policy.readOrganizationQuotaPolicy(ORG);
   await database
--- a/packages/cloud/shared/src/lib/services/organization-quota-policy.ts
+++ b/packages/cloud/shared/src/lib/services/organization-quota-policy.ts
@@ -131,42 +131,47 @@ export async function readOrganizationQuotaPolicyInTransaction(
   organizationId: string,
   observedAt?: Date,
 ): Promise<OrganizationQuotaPolicy> {
-  const [org] = await tx
+  // These rows are unique per organization. One statement keeps the policy
+  // inputs together without paying a separate database round trip for each.
+  const [inputs] = await tx
     .select({
-      balance: organizations.credit_balance,
-      revision: sql<string>`${organizations.balance_revision}::text`,
-      settings: organizations.settings,
+      org: {
+        balance: organizations.credit_balance,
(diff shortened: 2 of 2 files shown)


attentionhead (approved):
Main promotion of #31167 (via staging #31168). Head `118cb5142f326322228f1cf2666a6bb850e495e4` is identical to #31168's merge commit, and the file set is unchanged (`organization-quota-policy.ts` +34/-27, `organization-quota-policy.pglite.test.ts` +31/-0), so the verification below carries over to this head unchanged.


**APPROVE.** The five sequential reads collapse into one left-joined statement without changing the policy the reader produces. The two ways a join-fold like this normally goes wrong are both closed here, and I checked each against the schema rather than the diff:

**1. Row multiplication.** The comment claims "these rows are unique per organization"; that is enforced, not assumed. `organization_subscription_authorities.organization_id` is `primaryKey()` (`billing-subscriptions.ts:234`), `org_storage_quota.organization_id` is `primaryKey()` (`org-storage-quota.ts:20`), and `org_rate_limit_overrides.organization_id` (`org-rate-limit-overrides.ts:17`) and `organization_config.organization_id` (`organization-config.ts:16`) are both `.notNull().unique()`. No join can fan out, so destructuring the single `[inputs]` row is sound.

**2. Drizzle nulls a nested selection when every selected column is null** — the failure mode that silently converts "row exists" into "row absent". Each nested object retains at least one column that cannot be null when the row exists:
- `override` selects `id` (`defaultRandom().primaryKey()`). This is load-bearing, not decorative: an override row whose four RPM columns are all NULL would otherwise arrive as `null`, and `override`'s truthiness gates the rate-override validation block at `:305-310` that raises `invalid_rate_override` at `:338`. Keeping `id` preserves that path. It is also correct for the tier path — `resolveOrgTierFromSourceValues` (`org-rate-limits.ts:191`) computes `hasRpmOverride` from the four fields, so an all-null row and `undefined` resolve to the same tier either way.
- `storage` selects `bytes_limit` and `limit_override_authorized`, both `.notNull()` with defaults.
- `config` selects `settings`, `jsonb(...).default({}).notNull()`.
- `association` selects the whole table, whose `state` and `policy_generation` are `.notNull()`.

**3. Null-vs-undefined at the consumers.** Every downstream read is optional-chained (`override?.` at `:177-180`/`:299-302`, `config?.settings` at `:248`/`:296`, `storage?.` at `:253`/`:370-375`), so the switch from `undefined` (no row from a separate `select`) to `null` (unmatched left join) is invisible there. The one consumer with a declared optional parameter, `resolveOrgTierFromSourceValues(_, _, override?)`, is explicitly narrowed with `override ?? undefined` at `:225`. The guard rewrite `!org || !association` → `!inputs || !inputs.association` is equivalent, since `inputs` is present exactly when the `organizations` row is.

Transaction ownership, the `association.state === "unavailable"` short-circuit, the legacy-association branch's separate credit/clock reads, and the entitlement path are untouched.

**On the new test.** `expect(after).toEqual({ ...before, observedAt: after.observedAt })` is a real guard for the new code — a missing or wrong `organization_id` predicate on any of the four `leftJoin`s would leak the other tenant's override, config, and storage rows into `target`'s policy, and this assertion catches that. Worth noting it would also have passed against the pre-change implementation, where per-statement `where` clauses made the isolation trivially true; it discriminates the regression this refactor could introduce, not the refactor itself. The `missing_account_authority` case does exercise the new left join specifically.

**Non-blocking observations, no change requested:**
1. The case the `id` column actually protects — an `org_rate_limit_overrides` row present with all four RPM columns NULL — is not covered by the new test. The added test inserts `completions_rpm: 71`, so `override` would stay non-null even without `id`. A row of all-NULLs would be the assertion that pins the stated rationale, and it is the case where deleting `id` from the selection changes behavior (`invalid_rate_override` validation skipped instead of run).
2. The PR body describes the fix as preventing "a null completion limit [hiding] a configured embedding or standard limit". Drizzle nulls the nested object only when *every* selected column is null, so a row with `completions_rpm` NULL and `embeddings_rpm` set was never at risk; the real exposure is the all-NULL row described above. The code comment at `:146` states it correctly.
3. Latency is measured against the control-plane policy reader only, as the description says. The three-snapshot comparison (1518/1498/1495 ms → 836/831/830 ms) is a consistent ~45% reduction, but it is one account at one observation time; the staging gate you list is the right place to confirm it.

PR #30957, opened 2026-09-09
fix(lifeops): bind family email approvals and review account replacement

lalalune:
## Behavior

Family coordination now supports monthly email drafts bound to the exact sending account, verified recipient, subject, body, and saved version. Owners can edit the email before approval; each save creates an immutable revision and invalidates stale approval. Failed saves retain unsaved text, and downloaded draft records exclude owner-private claims. The UI distinguishes submission for approval from delivery.

The monthly school workflow supports explicit elementary selection and automatic application of validated changes. Identical PDFs produce quiet no-ops, while source configuration and parser-version changes force reconciliation. Live district-PDF inspection exposed preschool and high-school events incorrectly included for elementary families; this change removes those events from existing imports without recreating valid events. Calendar all-day rendering preserves civil dates across time zones.

Google OAuth configuration reaches the runtime, connection errors remain visible, and verified host-session identity reaches plugin access checks. The connection view supports reviewing a replacement account's permissions and synchronized calendars before retiring a selected test account.

## Verification

- Editor revision: 42 focused tests and 68 isolated Chromium checks passed, including saved-version approval, failed-save recovery, exact downloaded text, disclosure filtering, desktop/mobile layout, and hover contrast. Provider adapters in these browser checks are synthetic.
- Full personal-assistant suite at editor commit `30e63ee1960413aa7476bc040dec73774420ab46`: 299 files, 2,569 tests passed, 6 skipped. Root verification passed all 373 tasks and final audits at that revision.
- Calendar package: 840 passed, 4 skipped before the final develop merge.
- School selection and repair: 20 database-backed tests passed. A live DNS-pinned fetch and production PDF parser read the district's current source; the elementary result was inspected against the rendered PDF.
- App audit: 216 OCR-reviewed views, 203 verified, zero broken, 13 needing visual review. Family workspace captures were inspected at four viewports; mobile landscape scroll/overlay acceptance remains open.
- Merged head `2c881a898efb01d3348e7bcc3325f97c7d044078`, including develop `88106f05234d3`: 246 focused school, todo, evaluator, and authored-intent tests passed. Root verification passed (373 successful tasks plus final repository audits). Merge conflicts retain develop's real persistence/owner-isolation coverage and persisted-memory evaluator fixture.

[...]

PR #31173, opened 2026-09-12
docs(bettina): align school acceptance with automatic update policy

lalalune:
The Bettina PDD still required approval for every school-calendar import and revision after the owner selected automatic updates. Align the product requirements and acceptance matrix with that choice: validated workflow-owned changes apply automatically, optional review mode requires exact-plan approval, and both modes retain ambiguity quarantine and protection for unrelated events.

Refs #30123. Documentation only; the automatic implementation was already merged in #30957.

Validation: pinned-toolchain `bun install`; root `bun run verify` passed; guide parity (157 pairs), Markdown link validation, and `git diff --check` passed. No new links or runtime changes.


attentionhead (changes requested):
Reviewed at head `becd1a47a9b5b0e58a906ded3e2bfcc27972ec30` (+26/-14 across the two `docs/design/bettina-family-operations-*` files). Verified by fetching the PR head and reading the full files, not the diff alone.

The rewritten clauses are internally consistent and the new policy is stated carefully — ambiguity quarantine and workflow-ownership checks are explicitly preserved in both modes, and `SCH-14` (unchanged runs need no approval) still composes correctly. The verification plan's two rows were updated in step with the design.

The problem is that the alignment is incomplete within the same documents. The PR's stated purpose is to align the spec with the automatic update policy, but five untouched passages still assert the superseded approval-always policy. I checked this exhaustively rather than by sampling: grepping the PR head for `update mode|automatic mode|review mode` returns hits **only** on the lines this PR changed (61, 226, 234-235, 356-357, 361, 405-406, 409, 937, and verification-plan 90, 104). Every other approval assertion in these files predates the new policy and now contradicts it.

**1. `product-design.md` §9.1, line 544 — the architecture pipeline still hard-codes approval.**

The trusted-workflow step list, which is what an implementer builds from, reads:

```text
  -> Semantic event diff
  -> Owner approval
  -> Calendar mutation ledger
```

`Owner approval` is an unconditional stage between the diff and the mutation ledger. New `SCH-13` says automatic mode "applies validated first imports and changed revisions without per-run approval". As written, §9.1 and `SCH-13` cannot both be implemented.

**2. `product-design.md` §9.2, lines 554-570 — `SchoolCalendarSource` has no field for the selected mode.**

New `SCH-15` requires the workflow to expose `update mode`, and `APPROVAL-3`/`APPROVAL-4` both key their behavior off "the selected automatic mode". But the only source-configuration record the spec defines lists `id, householdId, districtName, landingUrl, stableResourceUrl, schoolYear, timezone, childEntityIds, includedScopes, targetElizaCalendarId, workflowId, activeArtifactId, activeSemanticRevisionId, lastRunAt` — no mode field anywhere. The policy that now gates every mutation has no specified home in the spec's own schema.

**3. `product-design.md` §17, pilot script steps 14 and 16 — the release-gate script is unrunnable under the selected mode.**

```text
14. Approve the first event import.
16. Exercise a changed fixture and approve additions/updates/removals.
```

Line 61 states the owner selected automatic application, so under the configured mode neither approval is ever presented and both steps block. This one is demonstrably an oversight rather than intent: the *matching* verification-plan row (line 90) **was** updated in this PR to "Inspect the selected automatic mode, run an import and changed revision... Separately select review mode and prove a changed plan waits for approval." The design doc's copy of the same scenario was left on the old policy.

**4. `product-design.md` §20, line 1044 — the MVP acceptance gate contradicts `APPROVAL-4`.**

MVP is declared done only when "external sends and consequential calendar changes are approval-gated across all connectors". New `APPROVAL-4` carves out "validated workflow-owned school events covered by the selected automatic mode". School-calendar additions, updates and removals are consequential calendar changes, so the acceptance criterion can no longer be satisfied by a conforming implementation.

**5. `product-design.md` §13.1, line 748 — a security control is now half-void.**

| Removal deletes unrelated event | Workflow ownership check and approved mutation plan |

Under automatic mode the "approved mutation plan" half of this mitigation does not exist, leaving the ownership check as the sole control against an automatic removal touching an unrelated event. That may well be acceptable, but the risk table should say so deliberately rather than continue to claim a control the new policy removed. This is the finding I would least want carried silently, because the table is the document readers consult for exactly this threat.

**Consequence.** These are design documents, so nothing breaks at runtime today. The cost is that the spec no longer has a single answer to "is a school-calendar mutation approval-gated?" — §9.1, §17, §20 and §13.1 say yes, while §6.4, `SCH-13`, `SCH-15` and `APPROVAL-3`/`APPROVAL-4` say it depends on the mode. Since §9.1 and §9.2 are the implementation contract and §20 is the release gate, an implementer following this revision would build the approval step the policy intends to make optional, and could not pass the acceptance criterion either way.

**Smallest repair**, all within the two files already touched:

1. §9.1: make the step conditional, e.g. `-> Owner approval (review mode)` or `-> Update-mode gate`.
2. §9.2: add an update-mode field (e.g. `updateMode`) to `SchoolCalendarSource`.
3. §17: rewrite steps 14 and 16 to mirror the verification plan's line 90 — observe automatic application, then switch to review mode to exercise approval.
4. §20: qualify the bullet with the `APPROVAL-4` carve-out.
5. §13.1: restate the removal control as the ownership check plus the mode gate.

Requesting changes for the incomplete alignment rather than for the new policy itself, which reads sound.

ss251 (approved):
Reviewed head becd1a47a9b5b0e58a906ded3e2bfcc27972ec30 — docs-only diff, both files read at this head; nothing here executes, so no suite applies.

The risky hunk is the APPROVAL-4 rewrite: approval is now waived for updates/removals of workflow-owned school events in automatic mode. I traced the carve-out's bounds across the diff and it stays fenced: SCH-12 (never modify manually created or unrelated events) is untouched; the decisions table, workflow step 11, and SCH-13 all repeat that ambiguous/invalid source data stays quarantined in either mode; and the verification plan adds the negative test "Verify automatic school synchronization cannot modify unrelated events." The untouched "should normally require approval" line still accommodates the exception. Review mode retains exact-plan approval and gets its own proof.

Non-blocking: step 7's new "configured synchronization policy" reads as a third name for SCH-15's "update mode"; unify the wording.

Recommendation: merge

— [free-9router-lane-ss251]

Issue #31176, opened 2026-09-12
fix(lifeops): make household actions available to authenticated owner planning


lalalune:
On the deployed Bettina pilot, a password-authenticated owner cannot reliably invoke household role binding/grant actions from a new conversation, although the runtime inventory contains the umbrella and promoted direct actions.

Observed on combined revision `58cd2df949d37627131de2427ca043e218385e09`:
- `/api/runtime` reports 477 registered actions, including `HOUSEHOLD_COORDINATION`, `HOUSEHOLD_COORDINATION_BIND_ROLE`, and `HOUSEHOLD_COORDINATION_ISSUE_GRANT`.
- An owner request to bind a clearly labeled synthetic caregiver and issue only `knowledge.read` instead selected a calendar view interaction and reminder actions; it did not create a household grant. The synthetic once reminders were subsequently absent from inventory; deletion/readback returned 404.
- A fresh owner conversation explicitly requested only the registered direct bind-role action, prohibited substitute effects, and permitted action discovery. It returned no actions and said the action was unavailable.

The root cause is not established. Inspect authenticated owner attribution, disclosure gates, action validation/discovery, and the actual model input before changing policy. Do not bypass owner authorization or add a generic action-execution escape hatch.

Acceptance:
- A normal authenticated owner conversation can bind a synthetic household role and issue/revoke a scoped household grant using the existing domain actions, with persisted receipts and no substitute reminders or external sends.
- Wrong/guest/unverified principals remain denied; action discovery and owner-private data do not leak across audiences.
- Record the real model input/output and persisted action/audit results, then complete owning-package, root, and exact-head hosted checks for the actual fix.

Related: #30123, #31172, #31162. This blocks the live grant lifecycle acceptance; no real guest was contacted.

lalalune:
The remaining live grant/read/revoke/expiry/export work now has real deployed evidence: [acceptance results](https://github.com/elizaOS/eliza/issues/30123#issuecomment-5650334067) and [independently verified bundle](https://github.com/elizaOS/eliza/releases/download/pr-evidence-12/30123-bettina-guest-live-5d169f8-evidence.zip). Normal owner model actions issued/revoked the grant; a paired guest retained approved-only access across restart, lost it after resource revocation and separately after natural household expiry, and remained denied owner routes. Both guest sessions and household grants are now revoked, with both resource-binding grant/revoke cycles preserved in the final export audit.

[...]

Issue #31172, opened 2026-09-12
fix(lifeops): expose agreement reads to verified scoped guest sessions


lalalune:
The Bettina pilot can pair a guest device and bind its persisted machine identity to a person Entity, but agreement reads are still registered behind the owner/admin gate. `GET /api/lifeops/agreements/:id` also passes SELF_ENTITY_ID into `readFor`. The owner-only `guest-projection` endpoint is a preview and does not provide an authenticated guest read path.

This leaves the guest acceptance requirement in #30123 incomplete even though the domain service has scoped resource and household grants. A live synthetic guest pairing and owner binding succeeded (HTTP 201); the temporary device session was subsequently revoked through its own authenticated session endpoint. No real guest was contacted.

Acceptance:
- Provide an authenticated guest read route that derives the Entity exclusively from the verified machine identity binding, never a caller-supplied principal or role.
- Delegate authorization to the existing agreement and household grant contracts; expose only the approved, scoped projection. Keep original PDF, owner export, review history, proposals, rejections, and owner mutations owner-only.
- Prove missing/wrong binding, wrong household/resource, expiry, revocation, and relationship changes deny access, including after restart.
- Exercise a real paired synthetic guest against the deployed route with a scoped grant, then revoke it and verify denial; include grant history in an owner export.
- Complete owning-package tests/typecheck/lint, real PGlite integration, root verification, and exact-head hosted checks.

Related: #31162 (owner export, including grant history). This issue covers the guest transport boundary rather than replacing the owner export implementation.

lalalune:
The remaining live grant/read/revoke/expiry/export work now has real deployed evidence: [acceptance results](https://github.com/elizaOS/eliza/issues/30123#issuecomment-5650334067) and [independently verified bundle](https://github.com/elizaOS/eliza/releases/download/pr-evidence-12/30123-bettina-guest-live-5d169f8-evidence.zip). Normal owner model actions issued/revoked the grant; a paired guest retained approved-only access across restart, lost it after resource revocation and separately after natural household expiry, and remained denied owner routes. Both guest sessions and household grants are now revoked, with both resource-binding grant/revoke cycles preserved in the final export audit.

This deployed candidate contains #31189, whose corrected explicit-empty-array contract has passed live acceptance and is awaiting its refreshed hosted checks/merge. The broader final merged MVP release remains tracked in #30123.

lalalune:
[...]

PR #31177, opened 2026-09-12
fix(lifeops): authorize scoped agreement reads for paired guests

lalalune:
Paired guests could be bound to a person and given agreement grants, but every agreement HTTP read still required owner authority. Add `GET /api/lifeops/agreements/:id/shared`, deriving the reader from the canonical machine-session binding and delegating authorization to the existing household/resource grant service.

The shared response contains only limited metadata and approved clauses, with no-store caching. Caller-supplied principal/role fields cannot select another reader. Original downloads, exports, review history, and mutations remain owner-only. Missing bindings, missing/revoked grants, and revoked sessions fail closed.

Refs #31172 and #30123. Live paired-guest acceptance is pending; the separate owner-planner gap in #31176 prevents a complete live household grant lifecycle today. This PR does not bypass that workflow or fabricate grants.

Validation:
- Pinned installation and final rebased root `bun run verify`: 373/373 passed (4m19.588s), including package typecheck/lint.
- Full personal-assistant suite: 2,658 passed, six existing skips, 303 files.
- Real PGlite and HTTP agreement integration: 13 passed, including real DB machine sessions, unique owner-bound identities, scoped grant read, exclusion of owner fields/routes, and revocation denial.
- Focused route boundary tests: eight passed, including forged-principal and missing-session attribution cases.
- README documents the authenticated transport contract. No UI change; app visual audit is N/A.

### DIFF
--- a/plugins/plugin-personal-assistant/README.md
+++ b/plugins/plugin-personal-assistant/README.md
@@ -247,6 +247,15 @@ Guest views expose approved obligations only. Revoking the relationship-backed
 household grant immediately invalidates every agreement binding that relies on
 it.
 
+A paired guest reads `GET /api/lifeops/agreements/:id/shared` using its machine
+session. The owner must bind that machine identity to a person through
+`POST /api/lifeops/entities/:id/auth-bindings` and explicitly issue the household
+and agreement grants. The shared route derives the person from the verified
+session binding; caller-supplied principal IDs and role headers cannot select
+another reader. It returns approved clauses and limited metadata with no-store
+caching. Original PDF downloads, exports, review history, and mutations remain
+owner-only. Use the normal owner read route for owner requests.
+
 The owner surface is available through `OWNER_AGREEMENT_KNOWLEDGE` and the
 authenticated `/api/lifeops/agreements/*` routes. Grant previews enumerate the
 exact read effects and exclusions before issuance. Active agent/chat pins feed
--- a/plugins/plugin-personal-assistant/src/lifeops/household/agreement-knowledge.pglite.integration.test.ts
+++ b/plugins/plugin-personal-assistant/src/lifeops/household/agreement-knowledge.pglite.integration.test.ts
@@ -5,10 +5,13 @@
  */
 
 import crypto from "node:crypto";
+import { once } from "node:events";
 import fs from "node:fs";
+import { createServer } from "node:http";
 import os from "node:os";
 import path from "node:path";
 import { resolveKnowledgeGraphService } from "@elizaos/agent";
+import { AuthStore } from "@elizaos/app-core/services/auth-store";
 import {
   type AgentRuntime,
   attestAuthenticatedApiDeliveryAudience,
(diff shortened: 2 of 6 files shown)


attentionhead (approved):
Reviewed at head `3640f3b6dfc0027fc893a741e8e3de9dcee39ef0` (+260/-4, six files). The security claim in the README is the whole PR, and it rests on three facts that live outside the diff, so I verified each against the PR's own bytes rather than the diff.

**The principal cannot be caller-selected.** `ctx.state.requestEntityId` is populated only at `routes/plugin.ts:203` from `requestPrincipals.get(req)`, a `WeakMap` keyed on the request object. The only writers are `withAuthenticatedEntityGate` (`plugin.ts:739`) and its owner counterpart, both of which set it from `resolveLifeOpsAuthenticatedPrincipal` and delete it in a `finally`. That resolver takes the role from `resolveAuthorizedRouteRole` on the session and, for a `USER`, requires the auth identity to resolve to exactly one *verified* `eliza_auth_session` binding (`exact.length === 1`, `authenticated-entity-principal.ts:43-52`), returning 403 otherwise. No header or body field reaches `entityId`. The README's "caller-supplied principal IDs and role headers cannot select another reader" is accurate as written.

**The new route is gated and fails closed.** `plugin.ts:578-581` registers `GET /api/lifeops/agreements/:id/shared` with `access: "authenticated_entity"`, which is a declared `RouteSpec` member (`plugin.ts:292`) consumed at `plugin.ts:776` to select `withAuthenticatedEntityGate` over `withOwnerAdminGate` — so an unrecognised value could not silently fall open here. Independently, the handler rejects a missing principal, `SELF_ENTITY_ID`, and `adminEntityId`; since `plugin.ts:203` falls back to `routeOwnerEntityId(runtime)` when no principal is bound, an ungated request lands on the owner id and is denied rather than served.

**Authorization is still enforced below the route.** `readFor` (`agreement-knowledge.ts:1774`) does not trust its `principalEntityId`: it normalises it, then for each non-revoked grant calls `household.requireGrantActive({scope: "knowledge.read"})` and only on success returns `guestArtifactProjection` plus obligations filtered to `status === "approved"`. Every candidate is evaluated before failing closed with `AGREEMENT_ACCESS_DENIED` — the `error-policy:J4` comment matches the loop's actual behaviour. The projection (`agreement-knowledge.ts:934`) carries only id/version/title/filename/mimeType/byteSize/pageCount/createdAt, with no storage key or media URL, so the README's "downloads, exports, review history, and mutations remain owner-only" holds by construction: this PR adds no route to those surfaces.

**The tests bind the behaviour rather than merely executing it.** The route spec plants `?principalEntityId=self` in the request URL and asserts `readFor` is called with `principalEntityId: "bound-guest"` — that assertion fails if the handler ever consults the query string, which is the precise regression the PR exists to prevent. `it.each([undefined, "self"])` asserts 403 `AGREEMENT_ACCESS_DENIED` with `readFor` never invoked (the harness sets `adminEntityId: "self"`, covering both the unbound and the owner case), and a third case asserts a revoked-grant throw survives as a 403 rather than a 500. `statusFor` maps `AGREEMENT_ACCESS_DENIED` to 403, so those status assertions are real.

**The new suite is actually collected.** `vitest.config.ts` excludes `**/*.integration.test.{ts,tsx}` via `defaultUnitExcludes`, so the unit lane drops the new `agreement-knowledge.pglite.integration.test.ts`. It is picked up by `vitest.src-integration.config.ts`, whose first include entry is the glob `src/**/*.integration.test.{ts,tsx}`. That config does `import baseConfig from "./vitest.config"` and spreads `...baseConfig` at the top level, so the three aliases this PR prepends to the `resolve.alias` array (`@elizaos/app-core/api/compat-route-shared`, `.../api/auth`, `.../services/auth-store`) are inherited by the integration lane — the lane that needs them, since `authenticated-entity-principal.ts` imports the latter two. Prepending matters: Vite array aliases are first-match-wins, so placing them ahead of any broader `@elizaos/app-core` entry is what makes them effective.

`sharedRead[0]` is the first capture group, not the full match — `pathMatch` returns `match.slice(1)` (`agreement-knowledge-routes.ts:69`), consistent with all nine sibling call sites.

Non-blocking observations, no change requested:

- The pre-existing `GET /api/lifeops/agreements/:id/guest-projection` reads `principalEntityId` from a query parameter. Its spec (`plugin.ts:585-587`) carries no `access` key and so falls to `withOwnerAdminGate`, making it an owner-only "preview as this guest" capability rather than a second, unguarded path to the same data. Worth a comment on that spec, since the two routes now sit adjacent and differ only by an absent key.
- The owner is deliberately excluded from `/shared` and receives 403. That is what the handler documents, but it means an owner cannot exercise the guest view through this route; `guest-projection` remains the owner's path to it.

Approving.

Issue #14622, opened 2026-07-06
[views] App-permissions section: setAppPermissions has no semantic action twin


lalalune:
## Problem
Settings → App permissions per-namespace toggles (`packages/ui/src/components/settings/AppPermissionsSection.tsx` → `client.setAppPermissions(slug, nextSet)`) are a local-only mutation: no registered action can grant/revoke an app's permission namespaces, so chat/voice cannot drive them. `SETTINGS_WRITE_REGISTRY` marks `app-permissions` `unwired` ("per-app namespace grants are a multi-value structure, deferred to a dedicated flow").

Found by the #14369 view→action ratchet; baselined as a `gap` in `packages/scripts/view-action-ratchet.registry.json` pending this issue.

## Acceptance criteria
- [ ] A semantic action (e.g. `SETTINGS action=set section=app-permissions app=<slug> namespace=<ns> granted=<bool>`, or an `APP` sub-op) drives the same use case as the toggle.
- [ ] `SETTINGS_WRITE_REGISTRY["app-permissions"]` flips from `unwired` to a real capability.
- [ ] The ratchet registry entry for `setAppPermissions` flips from `gap` to `action`.
- [ ] Real-LLM scenario evidence: "revoke camera access for <app>" changes the grant without a raw selector.

_Workstream: View-chat integration (#14369) · [design doc](https://github.com/elizaOS/eliza/blob/develop/packages/docs/ongoing-development/research/06-views-chat-integration.md)_

lalalune:
[shaw-agent] Taking this one next. I’ll wire app-permissions through a semantic SETTINGS route-backed capability, add deterministic route tests, and add live-only scenario coverage for the natural-language revoke/grant path.

lalalune:
Opened PR #14637: https://github.com/elizaOS/eliza/pull/14637

What changed:
- `SETTINGS_WRITE_REGISTRY["app-permissions"]` is now route-backed.
- `SETTINGS action=set section=app-permissions app=<slug> key=fs|net value=on|off` performs the same read-modify-write as the settings toggle: GET current `/api/apps/permissions/:slug`, compute the next granted namespace list, then PUT `{ namespaces: [...] }`.
- Added deterministic tests for grant/revoke, alias handling, missing app slug, and undeclared namespace failures.
- Added live-only scenario `settings-app-permissions-toggle` for “revoke network access for the weather app”.

Verification run:
[...]

PR #31181, opened 2026-09-12
fix(core): expose authorized actions before response routing

lalalune:
An authenticated Bettina owner requested a household role binding, but the response handler received only `HANDLE_RESPONSE` and no available-action catalog. It reported that the registered household action was unavailable and never selected it for planning.

Give the response handler a fresh, authorized discovery catalog before routing. Preserve complete descriptions and use the existing role, audience, private-action, connector-policy, context, and validation gates. Native planner tools and execution checks remain authoritative. Regression tests exercise custom-domain execution, owner-to-guest isolation, denied-action omission, and long descriptions.

Refs #31176. Part of #30123.

Validation:
- Core suite: 12,231 passed, 3 existing skips across 936 files.
- Focused routing suites: 272 passed.
- Node, browser, edge, testing, declarations, and packed consumer builds passed.
- Pinned Bun install and rebased root verification: 373/373 tasks passed (4m8.353s), including owning typecheck/lint and repository audits.
- Exact-head hosted checks: all five passed for `9f9355644decade35ba2a80de51ae1adf59b2458`.
- Deployed combined candidate `40bf83d22b150d71632878b3868097cc51ccd713` passed all eight Linux gates. All six PR source blobs match the deployed candidate. Public assets/deep links and runtime readiness were independently verified.
- Real before/after model proof: the same owner request changed from unavailable/relationship 404 to an authorized candidate selection, successful canonical household role binding, durable effect receipt, and independent relationship 200. Tracing was restored and original messaging ingress preserved.
- [Reviewed synthetic model/action and persisted-state evidence](https://github.com/elizaOS/eliza/releases/download/pr-evidence-12/31181-bettina-action-discovery-evidence.zip), SHA-256 `f4e6f1249107922615c71f8ff8e7cf119aaaf298367f155347ba8d70bdafdfcd`; secret scan passed. Full unrelated private owner context remains private.
- The subsequent grant action selected the correct tool but exposed #31186: explicit empty required subjects are treated as missing. No grant was stored. #31176 remains open for its grant issue/revoke acceptance; #31172/#31162/#30123 remain open for the broader live work.
- UI screenshots/video: N/A for this core routing change; live model/action and persisted-state evidence is linked above.

### DIFF
--- a/packages/core/README.md
+++ b/packages/core/README.md
@@ -358,6 +358,14 @@ Actions define specific tasks or capabilities the agent can perform. Each action
 
 Actions enable the agent to respond intelligently and perform operations based on user input or internal triggers.
 
+Before routing, the response handler receives an `available_actions` discovery
+catalog containing every eligible action's name, complete description, contexts,
+and aliases. Discovery checks the current actor, delivery audience, connector
+policy, and action validation under the action's declared routing contexts. This
+catalog is rebuilt for each turn and is not part of the shared prompt cache.
+The planner then receives native tools with parameter schemas and checks
+authorization again before executing; discovering an action does not execute it.
+
 Model-facing action, provider, and analytics results preserve complete records.
 For example, detailed `TRUST action=evaluate` results return every evidence
 record, follow-up suggestions return every qualifying contact, relationship
--- a/packages/core/src/__tests__/message-runtime-stage1.test.ts
+++ b/packages/core/src/__tests__/message-runtime-stage1.test.ts
@@ -7408,8 +7408,11 @@ describe("runV5MessageRuntimeStage1", () => {
 			responseId: "00000000-0000-0000-0000-000000000005" as UUID,
 		});
 
-		expect(validateAllowed).toHaveBeenCalledTimes(2);
-		expect(validateDenied).toHaveBeenCalledTimes(1);
+		expect(validateAllowed).toHaveBeenCalled();
+		expect(validateDenied).toHaveBeenCalled();
+		const discoveryPrompt = JSON.stringify(useModelCalls(runtime)[0]?.[1]);
+		expect(discoveryPrompt).toContain("CHECK_RUNTIME");
+		expect(discoveryPrompt).not.toContain("SKIP_RUNTIME");
 		const firstPlannerParams = useModelCalls(runtime)[1]?.[1] as {
 			tools?: Array<{ name?: string }>;
 			messages?: Array<{ role?: string; content?: string | null }>;
--- a/packages/core/src/__tests__/tiered-action-surface.test.ts
+++ b/packages/core/src/__tests__/tiered-action-surface.test.ts
@@ -57,6 +57,7 @@ function makeState(): State {
 
 interface CannedResponse {
 	body: unknown;
+	inspectInput?: (params: unknown) => void;
 }
 
 function createResponseHandlerFieldRegistry(): ResponseHandlerFieldRegistry {
@@ -120,7 +121,9 @@ function makeRuntime(opts: {
 				if (queue.length === 0) {
 					throw new Error(`Unexpected useModel call: ${String(modelType)}`);
 				}
-				return queue.shift()?.body;
+				const response = queue.shift();
+				response?.inspectInput?.(params);
+				return response?.body;
 			},
 		),
 		logger: {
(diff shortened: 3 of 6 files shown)


attentionhead (approved):
Reviewed at head `9f9355644decade35ba2a80de51ae1adf59b2458` (six files, +183/-4). The PR's safety claim is that discovery "uses the existing role, audience, private-action, connector-policy, context, and validation gates". That claim, not the routing fix, is what a review has to settle, and most of it lives outside the diff.

**The gates are genuinely reused, not re-implemented.** Discovery adds one parameter and changes one argument: `appendIfAllowed(action, undefined, mergeAgentContexts(args.selectedContexts, action.contexts))`. Everything downstream is untouched — `actionGateRejection` (private-action + `ACTION_ROLE_POLICY` + contextGate + roleGate through the shared chokepoint), `evaluateConnectorAccountPolicies`, and `action.validate()` all still run per action, in the same order, on the same code path the executor uses. No gate is skipped or reordered for discovery.

**The one semantic change is the context merge, and it is not new policy.** Folding `action.contexts` into the set the contextGate is evaluated against does relax context filtering — an action declaring `contexts: ["household"]` becomes visible on a turn Stage 1 routed to "general". The inline comment justifies this as "the same declared contexts as an explicit candidate hint", and that symmetry is real: the explicit-candidate path at `action-surface.ts:350` computes activeContexts identically as `mergeAgentContexts(args.selectedContexts, action.contexts)`, and its comment (#14622) already documents admitting a resolved parent "under ITS OWN contexts — still gated on role/private/context via appendIfAllowed". The child path at `:358`/`:382` merges the same way. So discovery adopts an established treatment rather than inventing a looser one, and it is the correct treatment for a surface whose entire purpose is to run *before* Stage 1 has selected contexts. `mergeAgentContexts` is exported from this same file at `:399` — no missing import despite the diff adding no import line.

**Isolation is tested against the mechanism, not the symptom.** The guest test registers five actions that must each be denied by a *different* gate — `roleGate: { minRole: "OWNER" }`, `private: true`, `disclosureGate: { require: "owner_exclusive" }`, a `contextGate: { noneOf: ["household"] }`, and `validate: async () => false` — then asserts none of their names appear anywhere in the serialized model input. That last case is worth noting: because discovery merges `contexts: ["household"]` into the active set, the `noneOf` gate is actually *exercised* by the merge rather than trivially passing, so the test covers the exact widening this PR introduces. The owner test additionally asserts the catalog is absent on a subsequent non-owner turn, which pins per-turn rebuilding rather than leakage through a cached surface.

**The caching claim holds.** The catalog is emitted with `stable: false`, and `context-renderer.ts:83` ends the cacheable prefix at the first non-stable segment. The new segment is pushed at `context-assembly.ts:275`, into a run of segments that are already `stable: false` (`:156`, `:180`, `:207`, `:230`, `:250`), so it demotes no existing stable segment and the shared prefix is unchanged. The README's "rebuilt for each turn and is not part of the shared prompt cache" is accurate.

**No duplicate sweep in one invocation.** The pre-existing tool-surface call at `context-assembly.ts:310` is guarded by `args.includeTools && args.selectedContexts?.length`; Stage 1 sets neither (`includeTools: true` appears only at `pipeline.ts:1063`, the planner call), so a Stage-1 turn runs the discovery pass alone.

Non-blocking observations, no change requested:

- The Stage-1 regression test traded `expect(validateAllowed).toHaveBeenCalledTimes(2)` / `toHaveBeenCalledTimes(1)` for bare `toHaveBeenCalled()`. The replacement assertions are sharper for this PR's purpose — they check the discovery prompt contains `CHECK_RUNTIME` and not `SKIP_RUNTIME`, which is the actual behaviour under change — so the test is not weakened on the axis that matters. What is lost is the invocation-count canary, and this PR is exactly the class of change that moves that count (that runtime registers two actions, so the counts should now be deterministic). Re-pinning the new exact numbers alongside the content assertions would keep a future change from silently adding a third sweep.
- Discovery runs on every Stage-1 turn, including ambient ones: `includeActionDiscovery: true` is set unconditionally at `pipeline.ts:244`, and `ambientHardGate` is only consumed as an instruction segment at `context-assembly.ts:210`, never as an early return. That is defensible since the turn issues a model call regardless, but it does make a full `validate()` + connector-policy sweep over every registered action an unconditional per-message cost, with author-supplied `validate` implementations in the loop. Worth watching as the action registry grows.
- Retaining complete descriptions is deliberate and consistent with the repo's "preserve complete records" rule, but it does make the per-turn prompt scale with the registry — the new test exercises a description of roughly 37 KB to prove long descriptions survive. If catalog size becomes a problem, `descriptionCompressed` already exists on `Action` as the intended lever, and using it here would be a prompt-size change rather than an authorization change.

Approving.

ss251 (approved):
Reviewed head 9f9355644decade35ba2a80de51ae1adf59b2458.

The risky part is `packages/core/src/services/message/action-surface.ts:289`: `discoverActions` merges `action.contexts` into the gate set so an action can be listed before Stage 1 has routed. The merge does not skip the other gates — `appendIfAllowed` still applies role, disclosure, private, connector, and validate. The guest test puts OWNER, `private:true`, `owner_exclusive`, `contextGate.noneOf:["household"]`, and `validate:false` names into Stage 1 and asserts none appear in the serialized `available_actions` payload; the owner path asserts the handler is uncalled at the first `useModel` `inspectInput`.

Verified by reading the diff at this head, following the merge into `appendIfAllowed`, then the projection at `context-assembly.ts:256` (name/description/contexts/similes JSON, `stable:false`) and `includeActionDiscovery:true` at `pipeline.ts:244`.

Recommendation: merge

— [free-xai-lane-ss251]

Issue #31182, opened 2026-09-12
Dedicated legacy inference repeats sequential policy reads


NubsCarson:
Dedicated app replies remain slow after the cold-cache admission correction. In the latest real production app request, runtime finalization took 28.609 seconds; the three observed completion calls spent 6.350, 6.839 and 8.142 seconds before provider dispatch. Reservation and dispatch preparation each occupied roughly two seconds. A scoped primary-database read confirmed that the test account uses the canonical legacy credit policy (`state=none`, no subscription rows).

The legacy policy reader currently performs five sequential reads inside its caller transaction: joined policy inputs, a repeated subscription-authority lookup, subscription absence, qualifying credit total, and the database clock. Reservation and dispatch revalidation both invoke this path. Combine subscription absence and qualifying credit aggregation with the already required policy-input statement, retaining the final database clock and both existing locking/revalidation boundaries. Do not change subscription behavior, rate limits, credit eligibility, credential checks, providers or model context.

Acceptance: real migrated database tests preserve tenant isolation, credit exclusions, empty-credit policy, subscription conflict rejection and changed-policy dispatch denial; profile actual executed SQL before/after rather than claiming an unmeasured wall-time gain. Complete package/root validation and a real staged Dedicated reply with saved-history continuity before considering production promotion. No additional servers are needed.

NubsCarson:
CLAIMING LEVER: the staging Cloud API/Pages release for #31182. Source PR31183 and isolated staging promotion PR31185 carry only the two policy-reader files. Both exact source and staging-candidate root verification passed; 17 PGlite and 3 independent PostgreSQL concurrency tests passed, with identical production policy results in a read-only comparison. The missing canonical staging ref was restored at its still-served certified commit before promotion. No runtime image, provider, capacity or production change is included. Dispatch will pin the merged staging SHA and validated tree; live Dedicated reply/history acceptance is required before any production promotion.

NubsCarson:
CLAIMING LEVER: production Cloud CF deployment for #31182. Promote only certified staging 342569fcaef58dbd58b79eff2f17a1996b149d80, tree 61cd27b42f092d84d910395a81991bcb88f9a19d, through an ordinary PR from a temporary promotion branch. Staging release 34723066410 and source CI 34722519055 are terminal success; actual Dedicated reply and all 41 messages passed local desktop/mobile reload. Preserve all allocated servers, agents, model/context and canonical staging. Exact production app verification follows deployment.

[...]

PR #31185, opened 2026-09-12
perf(cloud): stage the bounded legacy policy read optimization

NubsCarson:
Promote only the two-file policy-reader patch from #31183 onto the last certified staging commit `dd929ee7fdeab42206fc541162249e4f3e3bcb37`. Candidate `21619dd7cd369372df7ca64fe3b98d1900da3bd6`, tree `61cd27b42f092d84d910395a81991bcb88f9a19d`, excludes 18 unrelated paths introduced on newer develop. The changed source/test blobs and dependency manifests/lockfile exactly match the validated source. Exact staging-tree root verification and frozen installation passed; no dependency changes and no staging deployment yet.

Dedicated completion and embedding admission reread the legacy subscription association and credit selector through sequential primary-database requests. Combine subscription absence and qualifying credit aggregation into the existing policy-input statement, reducing five policy SELECTs to two. Both organization locks, the final database-clock read and fresh dispatch revalidation remain in place; subscription behavior, credit eligibility, credential checks and model context are unchanged.

Real migrated PGlite tests pass (17), including empty credits, excluded signup grants, debits, tenant isolation and a missing subscription association. Three independent PostgreSQL concurrency tests pass. Package typecheck/lint and root `bun run verify` pass; frozen installation made no dependency changes. Lint reports five existing large-file warnings.

A read-only comparison on the production primary returned identical policies in three alternating same-snapshot runs: 855/345 ms, 837/342 ms and 837/341 ms before/after. These measure the reader from the control-plane host, not Worker or end-to-end chat latency. Actual staged Dedicated reply/history acceptance remains required before production promotion. No deployment is claimed by this source PR. Wider pre-existing Develop Full failures remain outside this patch.

Hosted source CI [34722519055](https://github.com/elizaOS/eliza/actions/runs/34722519055) is still running at merge time and is not claimed passed. The repository owner authorized proceeding with the exact local package, real-database and root validation documented above.

Refs #31182.

| Evidence | Result |
| --- | --- |
| Before screenshots  | N/A - backend database-reader change with no rendered UI diff |
| After screenshots  | N/A - backend database-reader change with no rendered UI diff |
| Walkthrough video  | N/A - backend database-reader change with no rendered UI diff |
| Backend logs  | https://github.com/elizaOS/eliza/releases/download/pr-evidence-12/31182-legacy-policy-staging-backend.log |
| Frontend logs  | N/A - no frontend code change |
| LLM trajectory  | N/A - no model, prompt, provider or runtime handler changes; staged actual chat verification follows deployment |
| Domain artifacts  | https://github.com/elizaOS/eliza/releases/download/pr-evidence-12/31182-legacy-policy-staging-domain.json |
| OCR review  | N/A - no rendered UI diff |

### DIFF
--- a/packages/cloud/shared/src/db/repositories/organization-quota-policy.pglite.test.ts
+++ b/packages/cloud/shared/src/db/repositories/organization-quota-policy.pglite.test.ts
@@ -106,6 +106,52 @@ test("canonical legacy accounts keep purchased-credit RPM and distinct eager/non
     policy.requireOrganizationResourceLimit(after, "nonEagerSandboxes"),
   );
 });
+test("legacy credit selection excludes grants, debits and other tenants while retaining empty-credit policy", async () => {
+  const pg = database.getPgliteClientForTests();
+  const target = crypto.randomUUID();
+  const other = crypto.randomUUID();
+  await pg.exec(`
+    INSERT INTO organizations(id,credit_balance,balance_revision,balance_decrease_revision,settings,is_active,account_lifecycle_state)
+    VALUES('${target}',100,1,0,'{}',true,'active'),('${other}',100,1,0,'{}',true,'active');
+  `);
+  const empty = await policy.readOrganizationQuotaPolicy(target);
+  expect(Number(empty.tierSourceCreditTotal)).toBe(0);
+  await pg.exec(`
+    INSERT INTO credit_transactions(id,organization_id,amount,type,metadata) VALUES
+    (gen_random_uuid(),'${target}',1000,'credit','{"type":"initial_free_credits"}'),
+    (gen_random_uuid(),'${target}',1000,'credit','{"type":"wallet_signup"}'),
+    (gen_random_uuid(),'${target}',1000,'credit','{"type":"signup_code_bonus"}'),
+    (gen_random_uuid(),'${target}',-7,'debit','{}'),
+    (gen_random_uuid(),'${other}',999999,'credit','{}');
+  `);
+  const excluded = await policy.readOrganizationQuotaPolicy(target);
+  expect(excluded.tier).toEqual(empty.tier);
+  expect(Number(excluded.tierSourceCreditTotal)).toBe(0);
+  await pg.exec(`
+    INSERT INTO credit_transactions(id,organization_id,amount,type,metadata) VALUES
+    (gen_random_uuid(),'${target}',6,'credit','{}'),
+    (gen_random_uuid(),'${target}',4,'credit','{"type":null}');
+  `);
+  const funded = await policy.readOrganizationQuotaPolicy(target);
+  expect(Number(funded.tierSourceCreditTotal)).toBe(10);
+  expect(policy.requireOrganizationRateTier(funded).completionsRpm).toBeGreaterThan(
+    policy.requireOrganizationRateTier(empty).completionsRpm,
+  );
+});
+test("a missing subscription association cannot admit a subscriber as a legacy account", async () => {
+  const pg = database.getPgliteClientForTests();
+  await pg.exec(`UPDATE organization_subscription_authorities
+    SET state='none',subscription_id=NULL WHERE organization_id='${ORG}'`);
+  try {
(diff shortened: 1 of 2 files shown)


attentionhead (approved):
Reviewed at head `21619dd7cd369372df7ca64fe3b98d1900da3bd6` (two files, +74/-18). This folds two follow-up statements on the legacy branch into correlated selectors on the existing policy SELECT, so the question a review has to settle is whether the inlined predicates are *equivalent* to the ones they replace — and that equivalence lives outside the diff.

**The subscription check is exactly behavior-preserving, not a widening.** The removed call was `readPrimaryOrganizationSubscription(tx, organizationId)` followed by `if (subscription.state !== "none")`. On the `association.state === "none"` path that helper does only:

```ts
const existing = await tx.select({ id: billingSubscriptions.id })
  .from(billingSubscriptions)
  .where(eq(billingSubscriptions.organization_id, organizationId)).limit(1);
return existing.length === 0 ? { state: "none" } : { state: "unavailable", code: "subscription_authority_conflict" };
```

so the old guard rejected iff a `billing_subscriptions` row exists for the organization. The new `EXISTS (SELECT 1 FROM billing_subscriptions WHERE organization_id = organizations.id)` under `legacyHasSubscription !== false` rejects on exactly the same condition. The header comment's "still rejects any persisted subscription, including a terminal one" reads as if it were describing a tightened rule; it is describing a preserved one — the old path never filtered on `status` either. Worth confirming that wording is intentional, since a reader could take it as a new restriction.

**The `!== false` comparison is safe because the two `state` reads cannot disagree.** My first concern was that `legacyHasSubscription` keys off the *joined* `organizationSubscriptionAuthorities.state` while the branch guard keys off the JS-side `association.state`, and that a missing authority row would leave the SQL `CASE` condition NULL — making `legacyHasSubscription` NULL, `!== false` true, and turning a legacy organization into a hard `ORGANIZATION_POLICY_UNAVAILABLE`. That is closed upstream: `if (!inputs || !inputs.association || inputs.association.state === "unavailable") return unavailable(organizationId, "missing_account_authority")` runs first, and the authority is selected as a whole-table object, so a left-join miss yields a null `association` and returns before the legacy branch. Reaching `association.state === "none"` therefore requires a real row whose `state` is the literal `'none'`, which is precisely when the `CASE` condition is true and both selectors are non-null. The tri-state comparison is defensive rather than load-bearing, but it is correct.

**The correlation is what the first test actually pins.** Inserting a 999999 credit for a second organization and asserting the target still totals 10 would fail outright if the inlined subquery lost its `organization_id = organizations.id` correlation — which is the one mistake this refactor could plausibly make, and the reason a scalar subquery here does not fan the policy row out. The same test covers all three excluded metadata types, the `debit` row, and the `'{"type":null}'` case (where `->>'type'` is SQL NULL, `COALESCE` maps it to `''`, and it is correctly counted toward the 10). That is a genuine non-vacuity proof for the moved predicate, not a restatement of it.

**Cost claim holds, and slightly better than stated.** Because both selectors sit inside `CASE WHEN … = 'none' THEN …`, neither subquery is evaluated for subscriber organizations, so the common path pays nothing for the inlining. The legacy path drops from three statements to one. Note the branch still issues a separate `clock_timestamp()` round trip immediately afterward, so this is two statements rather than one — not a defect, just worth knowing if the goal is a single round trip end to end.

Two small observations, neither requiring a change:

- The second new test (`UPDATE … SET state='none', subscription_id=NULL` expecting `legacy_association_conflict`) passes under both the old and new implementations, since that fixture organization has a `billing_subscriptions` row either way. It is a behavior-preservation test rather than a regression test for this change, which is the right thing to have on a refactor — just not a discriminator if someone later reverts the inlining.
- `missing_legacy_selector` is now provably unreachable: `legacyCreditTotal` is NULL only when the `CASE` condition is false, which the branch guard excludes, and `COALESCE(SUM(...),0)::text` never yields NULL otherwise. The prior `if (!credits)` was likewise unreachable because an ungrouped aggregate always returns one row, so this is not a regression — but the code now guards a condition that cannot occur, and a future reader may misread it as live.

Approving.

Issue #31186, opened 2026-09-12
fix(lifeops): accept explicit empty subjects for knowledge-only household grants


lalalune:
## Reproduced failure

The deployed Bettina pilot at `40bf83d22b150d71632878b3868097cc51ccd713` now discovers and invokes the requested household action. A real owner request for `HOUSEHOLD_COORDINATION_ISSUE_GRANT` supplied `subjectEntityIds: []` with `scopes: ["knowledge.read"]`, a persisted synthetic caregiver role and an explicit household/principal. The action returned `MISSING_HOUSEHOLD_PARAMETERS` naming `subjectEntityIds` even though that parameter was supplied. A subsequent owner export independently confirmed that no household grant was stored for the synthetic principal.

The generic `resolveActionArgs` presence check treats every empty required array as missing. `HouseholdCoordinationService.issueGrant` deliberately permits an empty subject list for a knowledge-only grant; it separately rejects non-owner calendar grants without subjects. The action resolver prevents this valid domain contract from reaching the service.

## Acceptance

- Preserve the distinction between an omitted parameter and an explicitly supplied empty array for action parameters whose domain allows empty lists.
- Allow the household knowledge-only grant with explicit `subjectEntityIds: []` through the real owner action path and persist the actual grant.
- Keep omitted subjects, malformed arrays, unauthorized principals/roles, out-of-relationship subjects and non-owner calendar grants without subjects rejected by their existing boundaries. Do not make all required arrays optional or silently default an omitted array to empty.
- Cover the resolver and promoted household action with meaningful regression tests, run owning package/root and hosted checks, and complete the paired-guest read/revoke/expiry/export acceptance on the deployed pilot.

Part of #30123, #31172 and #31162. The action-discovery repair #31181 is separately verified: it successfully executes the role binding and exposes this downstream parameter defect.

lalalune:
CLAIMING: preserve explicitly empty required arrays for the household knowledge-grant action, without defaulting omitted parameters.

The resolver change is opt-in per parameter through SubactionSpec.allowEmptyArrays. The household grant retains subjectEntityIds in its required list and opts only that field in. All 12 resolver tests and all 28 real PGlite household integration tests passed. The promoted issue/revoke actions persist the grant and its revocation; omission, malformed elements and empty-subject non-owner calendar authority remain rejected. Full core/personal-assistant and root qualification are in progress.

[...]

PR #31189, opened 2026-09-12
fix(lifeops): preserve explicit empty household grant subjects

lalalune:
A knowledge-only household grant explicitly uses `subjectEntityIds: []`, but action argument resolution discarded the empty array and rejected the live owner request as missing a parameter. Allow subactions to explicitly accept empty arrays for named parameters, and enable that contract only for the household grant subject list. Omitted and null values remain missing; household scope and subject validation still reject malformed or unscoped calendar grants.

Refs #31186, #31176, #31172, #31162, #30123.

Validation: root `bun run verify` passed all 373 tasks; full core suite passed 12,235 tests with 3 skips; full personal-assistant suite passed 2,658 tests with 6 skips. The real PGlite household integration suite passed all 28 tests, including promoted grant issuance, durable receipt and stored authority inspection, revocation, and rejection without an extra persisted grant. Biome and diff checks passed.

Live acceptance on combined candidate `5d169f8050fbf2212249dc59b63a8270a473f675` passed: a recorded real model action preserved `subjectEntityIds: []` and produced the persisted knowledge grant. A paired guest then read only the approved clause, was denied owner reads/downloads/exports, lost access after resource revocation, and had its session revoked. Active and revoked export archives passed independent member/source hash verification. The first candidate attempt is included: the model substituted a subject ID and the service rejected it without creating a grant; a clarified request succeeded. This is evidence of the explicit-empty-array contract, not a claim that arbitrary model arguments always follow the request.

All Linux candidate validation/build stages and the five hosted checks for the exact PR head passed. [Recorded synthetic model interaction](https://github.com/elizaOS/eliza/releases/download/pr-evidence-12/31186-bettina-empty-subjects-live-model-5d169f8-v2.json).

N/A - No rendered UI changes.

N/A - No rendered UI changes.

N/A - No rendered UI changes; actual model and HTTP acceptance are recorded below.

<details><summary>Live server lifecycle observations</summary>

```json
{
  "unboundGuest": 403,
  "ownerBinding": 201,
  "boundWithoutResourceGrant": 403,
  "grantPreview": 200,
  "resourceGrantCreated": 201,
  "sharedRead": 200,
  "ownerReadDenied": 403,
  "originalDownloadDenied": 403,
  "ownerExportDenied": 403,
  "resourceGrantRevoked": 200,
  "revokedRead": 403
}
```

</details>

N/A - The changed action argument contract was exercised through authenticated HTTP and real model execution; it changes no browser client.

[Recorded real-model input, observable tool execution, output, and runtime provider/model labels](https://github.com/elizaOS/eliza/releases/download/pr-evidence-12/31186-bettina-empty-subjects-live-model-5d169f8-v2.json). Both the rejected first attempt and successful clarified request are included unchanged; the data are synthetic, the running service and model calls are real.

<details><summary>Persisted grant and verified export integrity</summary>

```json
{
  "grant": {
    "id": "hgrant_9cba4b43-515c-43ab-a472-04ce3327f908",
    "agentId": "b850bc30-45f8-0041-a00a-83df46d8555d",
    "householdId": "household:default",
    "principalEntityId": "ent_396914b8-d19b-4f96-8adb-96788583a697",
    "relationshipId": "hrole_75784905fb23729695c179d2b40720293cf191a1e72e42a0ef9de584ab8281a9",
    "role": "caregiver",
    "subjectEntityIds": [],
    "scopes": [
      "household.visibility",
      "knowledge.read"
    ],
    "issuedByEntityId": "self",
    "expiresAt": null,
    "revokedAt": null,
    "revokedByEntityId": null,
    "revocationReason": null,
    "createdAt": "2026-09-13T02:22:26.619Z",
    "updatedAt": "2026-09-13T02:22:26.619Z"
  },
  "archiveVerification": {
    "5d169f8-active-guest-export.zip": {
      "verified": true,
      "archiveSha256": "6c13df6a470ae1d02109c8d7366f9985d7931f610f69c5c2832e95199bb96b98",
      "originalSha256": "2e3cb50fb96c48666081c94631c072814bc6a71136f46c4210537531f724ae01",
      "manifestSha256": "b686131a599d09bad5c5796387545a7fdd3e7982e08ef5f7f37a2a360ec05c4e",
      "pageCount": 3,
      "extractionStatus": "available",
      "auditCoverage": "recorded_from_ingestion",
      "obligations": 3,
      "pins": 1,
      "grants": 1,
      "auditEvents": 15,
      "preparationReceiptVerified": false
    },
    "5d169f8-revoked-guest-export.zip": {
      "verified": true,
      "archiveSha256": "53f74e0a2405b5d0ffd08c50a6d4720fb8b87b2a692ce344a5bac25bb76bf12e",
      "originalSha256": "2e3cb50fb96c48666081c94631c072814bc6a71136f46c4210537531f724ae01",
      "manifestSha256": "0fab29b797b2681803454fd7a29df094a9c66dee3985b90425f8333493226082",
      "pageCount": 3,
      "extractionStatus": "available",
      "auditCoverage": "recorded_from_ingestion",
      "obligations": 3,
      "pins": 1,
      "grants": 1,
      "auditEvents": 17,
      "preparationReceiptVerified": false
    }
  },
  "sessionRevoked": true,
  "resourceRevoked": true
}
```

</details>

N/A - No rendered UI changes.

### DIFF
--- a/packages/core/README.md
+++ b/packages/core/README.md
@@ -23,6 +23,12 @@ fail closed.
 - **Plugin system:** `Plugin` objects contribute actions/providers/evaluators/services to the runtime.
 - **Built-in bundle:** Foundational capabilities ship as `basicCapabilities` (and `basicActions` / `basicProviders` / `basicEvaluators` / `basicServices`); there is no `corePlugin` singleton.
 
+`resolveActionArgs` treats a required empty array as missing unless its
+`SubactionSpec.allowEmptyArrays` names that parameter. This opt-in preserves
+an explicitly supplied or extracted `[]` through argument resolution; it never
+defaults an omitted or null value to an empty list. The owning action must still
+validate element types, permissions and domain constraints.
+
 ## Computer-use adapter contract
 
 `contracts/computer-use.ts` is the provider-neutral boundary shared by browser
--- a/packages/core/src/actions/resolve-action-args.test.ts
+++ b/packages/core/src/actions/resolve-action-args.test.ts
@@ -48,6 +48,131 @@ function makeMessage(text: string): Memory {
 }
 
 describe("resolveActionArgs", () => {
+	it("preserves an explicitly empty required array without model extraction when its contract allows it", async () => {
+		const runtime = makeMockRuntime();
+		const result = await resolveActionArgs({
+			runtime,
+			message: makeMessage(
+				"Issue a knowledge-only grant with no subject filter",
+			),
+			actionName: "GRANT",
+			subactions: {
+				ISSUE: {
+					description: "Issue a scoped grant",
+					descriptionCompressed: "issue grant",
+					required: ["subjectIds"],
+					allowEmptyArrays: ["subjectIds"],
+				},
+			},
+			options: { parameters: { action: "ISSUE", subjectIds: [] } },
+		});
+		expect(result).toEqual({
+			ok: true,
+			subaction: "ISSUE",
+			params: { action: "ISSUE", subjectIds: [] },
+		});
+		expect(runtime.useModel).not.toHaveBeenCalled();
+	});
+
+	it("retains an explicit empty array while extracting another missing parameter", async () => {
+		const runtime = makeMockRuntime(
+			JSON.stringify({
+				action: "ISSUE",
+				params: {
+					principalId: "caregiver",
+					subjectIds: ["unrequested-subject"],
+				},
+				missing: [],
+				confidence: 0.95,
+			}),
+		);
+		const result = await resolveActionArgs({
+			runtime,
+			message: makeMessage("Issue this knowledge-only grant to the caregiver"),
+			actionName: "GRANT",
+			subactions: {
+				ISSUE: {
+					description: "Issue a scoped grant",
+					descriptionCompressed: "issue grant",
+					required: ["principalId", "subjectIds"],
+					allowEmptyArrays: ["subjectIds"],
+				},
+			},
+			options: { parameters: { action: "ISSUE", subjectIds: [] } },
+		});
+		expect(result).toEqual({
+			ok: true,
+			subaction: "ISSUE",
+			params: { principalId: "caregiver", subjectIds: [] },
+		});
+	});
+
+	it("accepts an allowed empty array extracted from the request", async () => {
+		const runtime = makeMockRuntime(
+			JSON.stringify({
+				action: "ISSUE",
+				params: { subjectIds: [] },
+				missing: [],
+				confidence: 0.95,
+			}),
+		);
+		const result = await resolveActionArgs({
+			runtime,
+			message: makeMessage(
+				"Issue a knowledge-only grant with an explicitly empty subject list",
+			),
+			actionName: "GRANT",
+			subactions: {
+				ISSUE: {
+					description: "Issue a scoped grant",
+					descriptionCompressed: "issue grant",
+					required: ["subjectIds"],
+					allowEmptyArrays: ["subjectIds"],
+				},
+			},
+		});
+		expect(result).toEqual({
+			ok: true,
+			subaction: "ISSUE",
+			params: { subjectIds: [] },
+		});
+	});
+
+	it("still rejects omitted and null arrays, and empty arrays without an explicit allowance", async () => {
+		for (const input of [
+			{ parameters: { action: "ISSUE" }, allowEmptyArrays: ["subjectIds"] },
+			{
+				parameters: { action: "ISSUE", subjectIds: null },
+				allowEmptyArrays: ["subjectIds"],
+			},
+			{ parameters: { action: "ISSUE", subjectIds: [] }, allowEmptyArrays: [] },
+		]) {
+			const runtime = makeMockRuntime(
+				JSON.stringify({
+					action: "ISSUE",
+					params: {},
+					missing: ["subjectIds"],
+					confidence: 0.95,
+				}),
+			);
+			const result = await resolveActionArgs({
+				runtime,
+				message: makeMessage("Issue a grant"),
+				actionName: "GRANT",
+				subactions: {
+					ISSUE: {
+						description: "Issue a scoped grant",
+						descriptionCompressed: "issue grant",
+						required: ["subjectIds"],
+						allowEmptyArrays: input.allowEmptyArrays,
+					},
+				},
+				options: { parameters: input.parameters },
+			});
+			expect(result).toMatchObject({ ok: false, missing: ["subjectIds"] });
+		}
+	});
+
 	it("preserves an explicit description clear while extracting a missing goal ID", async () => {
 		const runtime = makeMockRuntime(
 			JSON.stringify({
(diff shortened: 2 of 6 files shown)


attentionhead (approved):
Reviewed at head `ad91a80a5ae9f01a1663586266f0a631f2aa382e` (draft, base `develop`, +268/-5 across 6 files).

## Verdict: approve

The opt-in is correctly scoped and, more importantly, the empty subject list is **fail-closed on every read path I could reach**. I went looking for the privilege widening this change could plausibly introduce and could not construct one.

### The security question I tried to answer

`allowEmptyArrays: ["subjectEntityIds"]` on `issue_grant` makes it possible to persist a household grant whose subject list is `[]`. The failure mode worth blocking would be downstream code reading an empty subject list as *unrestricted* rather than as *no subjects*. It does not:

- `requireScope` (`lifeops/household/service.ts:1683-1687`) matches on `subjectEntityId === undefined || grant.subjectEntityIds.includes(subjectEntityId)`. Against `[]`, any **defined** subject denies. The empty-subject grant satisfies only scope checks that pass no subject at all.
- `scopesForSchedule` (`:2852-2856`) selects grants via `grant.subjectEntityIds.some(id => scheduleSubjectEntityIds.includes(id))`. `[]` never intersects, so such a grant contributes **zero** schedule scopes — every schedule entry is then skipped by the `if (!scopes.includes("calendar.freebusy")) continue` guard at `:2985` and `:3028`.
- `grantIsActive` (`:1572-1576`) checks `grant.subjectEntityIds.some(id => !currentSubjects.includes(id))`; vacuously false on `[]`, which is the correct reading — a grant naming no subjects cannot drift outside a relationship.

### Why `household.visibility` arriving for free is not a leak

`expandGrantScopes` adds `household.visibility` whenever `knowledge.read` is present (`types.ts:387-389`), and the new integration test asserts exactly that stored pair. `exportFor` computes `effectiveScopes` from the principal's grants **without** subject filtering, so the scope genuinely arrives. What bounds it is the separate subject projection: `visibleSubjectEntityIds` is `uniqueStrings(ownActiveGrants.flatMap(g => g.subjectEntityIds))` (`:2940-2942`), which collapses to `[]`. Consequently:

- `roles` (`:2957-2963`) filters to `binding.entityId === principalEntityId` — the principal's own binding only.
- proposals and agreements (`:2969-2974`, `:3011-3016`) require the principal to be an affected party themselves, since the `visibleSubjectEntityIds.includes(...)` disjunct is always false.
- the audit trail still requires the distinct `household.export` authority (`:3056-3057`).

So the grant conveys: the holder's own role binding, items naming them directly, and knowledge artifacts explicitly bound by id. That matches the README's claim.

The calendar guard is the one place an empty subject list *must* be rejected, and it is, at `service.ts:1411-1418` — `scopes.some(s => s.startsWith("calendar."))` with `subjectEntityIds.length === 0` and a non-owner role throws `HOUSEHOLD_INVALID_CONTRACT`. The test exercises this with `scopes: ["calendar.freebusy"]`.

### Core resolver change

The three call sites are consistent and the flag is consulted only inside the array branch of `valueIsPresent`, so `null`, `undefined` and whitespace strings remain missing regardless of the allowance:

- `missingRequiredKeys:118` passes `allowEmptyArrays?.includes(key)`, which yields `undefined` when the spec omits the field and falls to the `= false` default — unlisted specs are byte-for-byte unchanged.
- `pickKnownParams:147-148` applies the same allowance, so an explicit `[]` survives the merge instead of being dropped as absent.
- the prompt segment (`:187-191`) is emitted only when `allowEmptyArrays?.length` is truthy, so existing extraction prompts are untouched.

Non-vacuity holds (the fix cannot be reverted without failing the suite): tests 1–3 each assert a preserved `[]`, and the fourth loops the three negatives — omitted, `null`, and `[]` under `allowEmptyArrays: []` — all of which must still report `missing: ["subjectIds"]`.

Blast radius is one production spec: `allowEmptyArrays` appears exactly once outside core and its own docs, at `household-coordination.ts:95`.

I also confirmed the new test import resolves: `promoteSubactionsToActions` is re-exported by name at `packages/core/src/actions.ts:57`, reached from the package root via `index.node.ts:18` (`export * from "./actions"`, which resolves to `src/actions.ts`, not the directory — there is no `actions/index.ts`). And `stringArray(params, "subjectEntityIds", { required: true })` accepts `[]`: the `value === undefined` early return is skipped, `Array.isArray([])` holds, and `[].some(...)` is false, so `required` here means *key present*, not *non-empty*.

## Non-blocking notes

1. **`allowEmptyArrays` is silently inert for a non-array key.** `valueIsPresent` consults the flag only under `Array.isArray(value)`, so naming a string or number parameter has no effect and no diagnostic. The `keyof TParams & string` type does not constrain the value to an array type. Harmless today given the single production use, but a future spec listing the wrong key would fail silently rather than loudly.

2. **`allowEmptyArrays` is not checked against `required`.** Listing a key that is only `optional` is inert — `missingRequiredKeys` walks `required` alone, and `pickKnownParams` short-circuits on `!spec.required.includes(key)` before consulting the allowance. Also inert, and also silent.

3. The README says the allowance "never defaults an omitted or null value to an empty list", which the fourth test pins down directly. Worth keeping that test as the contract if the resolver is refactored — it is the only place the `null` case is asserted.

ss251 (approved):
I'll start from the published review skill and the full prompt so the review follows the real scoring rules, not the truncated brief.ad91a80a5ae9f01a1663586266f0a631f2aa382e

Input `{ parameters: { action: "ISSUE", subjectIds: [] } }` against required `subjectIds` used to be treated as missing: `valueIsPresent` in `packages/core/src/actions/resolve-action-args.ts` returned false for every empty array. That is the knowledge-only grant hole — explicit `subjectEntityIds: []` never reached the action; the resolver called the model or failed closed.

This head adds opt-in `SubactionSpec.allowEmptyArrays` (L30) and threads it through `missingRequiredKeys` (L118) and `pickKnownParams` (L147–148). `valueIsPresent([], true)` is true; omitted, `null`, and `[]` without the flag still miss. The extractor prompt (L187–189) forbids defaulting omitted values to `[]`. Only household `issue_grant` opts in (`household-coordination.ts` L95).

Failure-sensitive tests: explicit `[]` does not call `useModel`; input `subjectIds: []` plus extracted `subjectIds: ["unrequested-subject"]` → output keeps `[]`; omitted/`null`/unallowed `[]` → `missing: ["subjectIds"]`. The PGlite test is the product proof: `HOUSEHOLD_COORDINATION_ISSUE_GRANT` with `subjectEntityIds: []` and `scopes: ["knowledge.read"]` persists empty subjects and scopes `["household.visibility", "knowledge.read"]`; omitting the field still returns `MISSING_HOUSEHOLD_PARAMETERS`; `[42]` and `calendar.freebusy` still throw `HOUSEHOLD_INVALID_CONTRACT`; revoke sets `revokedAt`. Domain checks are not weakened.

No lockfile, CI, binary, or permission changes.

Verified by reading the diff at this head (resolver, household opt-in, unit tests, PGlite integration, both READMEs).

Recommendation: merge

— [free-xai-lane-ss251]

Issue #31184, opened 2026-09-12
fix(app-core): authenticate owner browser WebSockets with the session cookie


lalalune:
## Reproduced failure

On the Bettina HTTPS pilot at `58cd2df949d37627131de2427ca043e218385e09`, a normal owner password login succeeds and authenticated `GET /api/status` returns 200 with `canRespond: true`. A same-origin browser WebSocket using that session closes with code 1008, `Unauthorized`, after a ping instead of receiving a pong. The backend recorded 36 pre-auth message rejections during a fresh 50-second observation.

The app-core WebSocket authorizer checks query/header session tokens but omits the first-party browser session cookie. REST and WebSocket authentication consequently disagree for a normal browser login.

## Acceptance

- Admit a current owner browser session through the canonical session store and role resolution, retaining the existing origin boundary.
- Reject expired, revoked, malformed, unknown and non-owner cookie sessions; do not introduce a trusted-loopback or missing-origin shortcut for cookie admission.
- Preserve existing bearer/device authentication behavior and fail closed on store errors.
- Exercise the real HTTP/WebSocket boundary and session storage, run package/root checks and exact-head hosted checks, and repeat password-login plus same-origin ping/pong on the deployed pilot. Keep credentials out of evidence.

Part of Bettina deployment acceptance in #30123. This is separate from the model action-discovery repair #31181 and from the cloud/native session-authority work in #30920, which does not change this host authorizer.

lalalune:
CLAIMING: owner browser-cookie WebSocket admission for Bettina deployment acceptance.

The isolated change uses canonical session/role resolution and the existing credentialed-origin policy. Real app-core WebSocket/PGlite tests passed, including owner ping/pong, explicit bearer preservation, revoked/expired/unknown/non-owner cookies and denied origins. The full app-core command passed 4,615 tests with 25 existing skips and completed its script suites. Two initial test-harness type errors were corrected; package typecheck passed, and final root verification is running. This is not deployed yet; the live pilot remains 40bf83d with the reproduced cookie failure.

lalalune:
[...]

PR #31190, opened 2026-09-12
fix(app-core): authenticate owner browser WebSocket sessions

lalalune:
A self-hosted owner can sign in successfully over HTTPS while the browser WebSocket is repeatedly rejected as unauthorized. App-core now authenticates the browser session cookie against the canonical session store, requiring an active owner session and a credentialed origin. Existing explicit bearer and device-token paths remain available. Revocation and expiry are checked on new connections.

Refs #31184 and #30123.

Validation at `daabf99ed6cf9`: rebased pinned install and root `bun run verify` passed all 373 tasks. The full app-core lane passed 4,615 tests with 25 skips plus its script suites. Real agent transport tests passed 18 tests across three files. Six real HTTP/WebSocket tests with PGlite-backed sessions cover owner admission, expiry, revocation, guest denial, missing and untrusted origins, permissive wildcard CORS, and existing bearer admission. Biome and diff checks passed.

The public pilot reproduced owner-cookie WebSocket rejection on `917f68d490923502a866c7800e56eb2b6493578a` and now returns a real ping/pong response on `5d169f8050fbf2212249dc59b63a8270a473f675`. Both observations use the same probe and a normal password-authenticated owner browser session, without query credentials or injected cookies. The deployed candidate preserves intake changes and contains this PR's identical WebSocket patch plus the independent empty-subject fix. Public build assets were verified against the Linux build receipt. All required Linux build/test stages passed, and all five hosted checks passed for this PR head.

N/A - Transport authentication does not alter rendered UI.

N/A - Transport authentication does not alter rendered UI.

N/A - No rendered interaction changed; the real browser transport receipt is below.

Sanitized server deployment verification receipt; private state and account identifiers omitted.

<details><summary>Server verification</summary>

```json
{
  "commit": "5d169f8050fbf2212249dc59b63a8270a473f675",
  "originalIngressPreserved": true,
  "trajectoryConfigurationPreserved": true,
  "agreementStatePreserved": true,
  "controlPreserved": true,
  "monthlyPacketPreserved": true
}
```

</details>

Real public browser probe output before and after, using the same probe SHA-256 `a32f3694c503bc6d84adf9a743bed3ddb0aedd44815c6906d042ee54783d8021`.

<details><summary>Browser transport observations</summary>

```json
{
  "before": {
    "authMeHttp": 200,
    "canRespond": true,
    "origin": "https://ovh-eliza.tail4e11f5.ts.net:8443",
    "ownerRole": "OWNER",
    "ownerStatusHttp": 200,
    "pendingRestart": false,
    "sessionKind": "browser",
    "transport": {
      "code": 1008,
      "outcome": "closed",
      "reason": "Unauthorized"
    },
    "observedPilotRevision": "917f68d490923502a866c7800e56eb2b6493578a",
    "probeSha256": "a32f3694c503bc6d84adf9a743bed3ddb0aedd44815c6906d042ee54783d8021"
  },
  "after": {
    "authMeHttp": 200,
    "canRespond": true,
    "origin": "https://ovh-eliza.tail4e11f5.ts.net:8443",
    "ownerRole": "OWNER",
    "ownerStatusHttp": 200,
    "pendingRestart": false,
    "sessionKind": "browser",
    "transport": {
      "outcome": "pong"
    }
  }
}
```

</details>

N/A - No model or action behavior changes in this transport authentication PR.

N/A - Read-only transport admission produces no persisted domain records. The browser network receipt above captures the observable result.

N/A - No rendered UI changes.

### DIFF
--- a/packages/agent/src/index.ts
+++ b/packages/agent/src/index.ts
@@ -183,6 +183,7 @@ export {
 // exported through `./api/server.js`.
 export {
   getConfiguredApiToken,
+  isCredentialedCorsOrigin,
   isTrustedLocalRequest,
 } from "./api/server-helpers-auth.ts";
 // `server-types.ts` is the canonical source for conversation/server type
--- a/packages/app-core/README.md
+++ b/packages/app-core/README.md
@@ -41,6 +41,14 @@ bun run --cwd packages/app-core lint         # Biome
 
 This package supplies host integration to the `packages/app` shell and app-facing plugins. It targets Node `>=24`, with `react`/`react-dom`/`three` as peer dependencies and the `@elizaos/capacitor-*` mobile bridges as optional dependencies.
 
+On the self-hosted dashboard, an owner browser session can authenticate a
+WebSocket without exposing its HttpOnly session cookie to JavaScript. Cookie
+admission uses the canonical session store, requires an owner identity and a
+credentialed browser origin, and checks expiry and revocation on each new
+connection. Broad cloud or wildcard-bind CORS reachability does not grant
+cookie access. Explicit bearer and paired-device authentication retain their
+existing transport contracts.
+
 ## Isolated local development
 
 Give each concurrent instance distinct `ELIZA_UI_PORT`, `ELIZA_API_PORT`,
--- a/packages/app-core/src/api/server-browser-session-ws.test.ts
+++ b/packages/app-core/src/api/server-browser-session-ws.test.ts
@@ -1,0 +1,224 @@
+/**
+ * Exercises owner browser-cookie admission through the real app-core HTTP and
+ * WebSocket server with PGlite-backed identities, sessions, expiry and revocation.
+ * The transport must agree with the session authority without admitting ambient
+ * cookies from missing or untrusted origins or granting owner streams to guests.
+ */
+import { randomUUID } from "node:crypto";
+import { mkdtemp, rm } from "node:fs/promises";
+import { tmpdir } from "node:os";
+import path from "node:path";
+import { AgentRuntime } from "@elizaos/core";
+import {
+  createDatabaseAdapter,
+  plugin as sqlPlugin,
+} from "@elizaos/plugin-sql";
+import { afterAll, beforeAll, describe, expect, it, vi } from "vitest";
+import WebSocket from "ws";
+import { AuthStore, type DrizzleDatabase } from "../services/auth-store";
+import {
+  BROWSER_SESSION_TTL_MS,
+  createBrowserSession,
+  createMachineSession,
+} from "./auth/sessions";
+import { startApiServer } from "./server";
+
+type Probe =
+  | { type: "pong" }
+  | { code: number }
+  | { httpStatus: number }
+  | { error: string };
+
+const API_TOKEN = "browser-cookie-ws-test-static-token";
+let stateDir: string;
+let adapter: ReturnType<typeof createDatabaseAdapter>;
+let store: AuthStore;
+let server: Awaited<ReturnType<typeof startApiServer>>;
+let activeCookie: string;
+let expiredCookie: string;
+let revokedCookie: string;
+let guestCookie: string;
+
+function ping(headers: Record<string, string>): Promise<Probe> {
+  return new Promise((resolve) => {
+    const ws = new WebSocket(`ws://127.0.0.1:${server.port}/ws`, { headers });
+    let settled = false;
+    const finish = (result: Probe) => {
+      if (settled) return;
+      settled = true;
+      clearTimeout(timer);
+      ws.terminate();
+      resolve(result);
+    };
+    const timer = setTimeout(() => finish({ error: "timeout" }), 10_000);
+    ws.on("open", () => ws.send(JSON.stringify({ type: "ping" })));
+    ws.on("message", (data) => {
+      const frame = JSON.parse(String(data)) as { type?: string };
+      if (frame.type === "pong") finish({ type: "pong" });
+    });
+    ws.on("close", (code) => finish({ code }));
+    ws.on("unexpected-response", (_request, response) => {
+      response.resume();
+      finish({ httpStatus: response.statusCode ?? 0 });
+    });
+    ws.on("error", (error) => finish({ error: error.message }));
+  });
+}
+
+function cookieHeaders(sessionId: string): Record<string, string> {
+  return {
+    Origin: `http://127.0.0.1:${server.port}`,
+    Cookie: `eliza_session=${encodeURIComponent(sessionId)}`,
+  };
+}
+
+describe.sequential("owner browser session WebSocket admission", () => {
+  beforeAll(async () => {
+    stateDir = await mkdtemp(path.join(tmpdir(), "eliza-cookie-ws-"));
+    vi.stubEnv("ELIZA_STATE_DIR", stateDir);
+    vi.stubEnv("ELIZA_CONFIG_PATH", path.join(stateDir, "eliza.json"));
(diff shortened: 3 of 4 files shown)


attentionhead (approved):
Reviewed at head `daabf99ed6cf9d67d28c6246d9c4f386ae042ca3` (draft, base `develop`, +250/-1 across 4 files; the behavioural change is 17 lines in `packages/app-core/src/api/server.ts`).

## Verdict: approve

This admits an ambient browser credential on a WebSocket handshake, so I treated it as a CSRF question first: **WebSocket handshakes are not subject to the same-origin policy or preflight**, which means the `Origin` test is the entire defense, not a defense-in-depth layer. It holds up.

### The origin gate is the CSRF boundary, and it is correctly anchored

`isCredentialedCorsOrigin` (`packages/agent/src/api/server-helpers-auth.ts:158-173`) rejects before it accepts:

- `if (!origin) return false` — a handshake with **no** `Origin` header is denied. This is the important one, because a non-browser client (curl, a native socket) simply omits the header, and the adjacent REST-side call at `agent/src/api/server.ts:1634` uses the laxer `requestOrigin === undefined || isCredentialedCorsOrigin(requestOrigin)`. The new call deliberately does **not** inherit that `undefined ||` disjunct, so cookie admission is strictly narrower than CORS reachability. The test pins this (`Cookie` with no `Origin` → 1008).
- `"null"` and `file:` origins are rejected explicitly — that is what a sandboxed iframe or a local file document sends.
- Both accepting patterns are fully anchored: `LOCAL_ORIGIN_RE` is `/^https?:\/\/(localhost|127\.0\.0\.1|\[::1\]|\[0:0:0:0:0:0:0:1\])(:\d+)?$/i` and `CREDENTIALED_APP_ORIGIN_RE` is `/^(capacitor|capacitor-electron|app|tauri|electrobun):\/\/.*$/i`. No unanchored substring match, so `https://localhost.evil.example` does not pass.
- `CREDENTIALED_APP_ORIGIN_RE` is deliberately narrower than the sibling `APP_ORIGIN_RE` used for plain CORS: it drops the `file:` scheme. Credentialed admission really is a strict subset of reachability, which is what the README paragraph claims.

The test's third case is the one that makes this non-vacuous: it flips `ELIZA_CLOUD_PROVISIONED=1` and then `ELIZA_API_BIND=0.0.0.0`, and an untrusted origin stays denied in both configurations (401, then 1008). That is the exact regression — "broad CORS reachability grants cookie access" — the change is meant to preclude.

### Session validity is enforced at the store, not assumed

`resolveSessionTokenRole` (`app-core/src/api/auth.ts:468-491`) delegates to `findActiveSession` → `AuthStore.findSession` (`services/auth-store.ts:346-362`), which returns `null` when `session.revokedAt !== null` or `session.expiresAt <= now`. So expiry and revocation are checked on **every** new connection rather than cached, which is what makes the final test ("denies the next connection immediately after owner session revocation") meaningful. Browser sessions additionally re-derive `nextExpiresAt` against a `rememberDevice` cap and bail when `nextExpiresAt <= now` (`auth/sessions.ts:216-222`).

Role derivation cannot leak owner status: `roleForIdentityKind` (`auth.ts:429-435`) maps `owner → "OWNER"`, `machine → "USER"`, everything else to `"NONE"`, and the new guard admits only on `session?.role === "OWNER"`. The guest fixture in the test is a `machine` identity, so it resolves to `USER` and is refused.

Store read failures route through `denyOnAuthStoreError`, so a database fault denies rather than admits — the helper's own contract comment states it returns `null` for a missing store, an unknown/expired/revoked session, or a read failure.

`readCookie` (`auth.ts:246-268`) returns `null` on a malformed percent-escape under a documented `error-policy:J3`, so the `%ZZ` case is an absent cookie rather than a thrown 500 — matching the test's 1008.

Control-flow placement is right: a non-OWNER cookie **falls through** to the existing token path rather than short-circuiting to a denial, so bearer and paired-device transports keep their contracts. The second test asserts the bearer path still works with no cookie present.

`isCredentialedCorsOrigin` is newly re-exported from `packages/agent/src/index.ts:186`, which is required — `app-core/src/api/server.ts` imports it from `@elizaos/agent` in the same block as `isAllowedHost`/`isAuthorized`.

## Non-blocking notes

1. **Any localhost port is a credentialed origin, and cookies ignore port.** `LOCAL_ORIGIN_RE` admits `http://localhost:<any>`. Because cookie scoping is host-based and port-agnostic, a *different* application served from `http://localhost:3000` on the same machine can cause the browser to attach `eliza_session` to a handshake against the API port, and would be admitted as OWNER. This is consistent with the pre-existing policy — the same predicate already gates `Access-Control-Allow-Credentials` for REST — so it is not a regression introduced here. It is worth stating explicitly in the README paragraph, though, since WebSockets have no preflight: the trust boundary for a self-hosted dashboard is "every process that can serve a page on this host", not "this app".

2. **No session-`kind` assertion on the cookie path.** `resolveSessionTokenRole` accepts any live session id; nothing requires `kind === "browser"`. A *machine* session belonging to an owner identity, if placed in the `eliza_session` cookie, would authenticate. That is not an escalation — a holder of that id can already authenticate via `Authorization: Bearer` — but the function's own doc comment describes it as the paired-device machine-session path, and the new caller is the browser-cookie path, so an explicit `kind` check (or a comment) would keep the two intents from drifting.

3. **`SESSION_COOKIE_NAME` is now defined twice.** `auth.ts:232` keeps a private `const SESSION_COOKIE_NAME = "eliza_session"` while `server.ts` imports the exported one from `auth/sessions.ts:44`. Values are identical today, so there is no bug; consolidating on the exported constant would remove the possibility of a silent divergence, given `auth.ts` uses its private copy inside `resolveAuthorizedRouteRole`.

ss251 (approved):
daabf99ed6cf9d67d28c6246d9c4f386ae042ca3

FINDINGS

Real fix on the product auth path. packages/app-core/src/api/server.ts:1145-1159: a WS handshake carrying eliza_session is admitted only if Origin passes isCredentialedCorsOrigin and resolveSessionTokenRole returns role OWNER; anything else falls through to the unchanged bearer/query-token checks, so existing transport contracts stay intact.

Input -> output, per the new integration test (real server on port 0, PGlite-backed sessions, no mocks):
- owner cookie + credentialed Origin, no bearer -> pong (server-browser-session-ws.test.ts:155-157)
- unknown/expired/revoked/machine-guest cookies -> 1008 (test.ts:165-174): expiry and revocation re-checked per connection, as the README claims
- cookie with no Origin header, or malformed %ZZ value -> 1008 (test.ts:176-183): ambient cookies cannot authorize outside a browser-origin context
- untrusted Origin denied under 127.0.0.1 bind (403), ELIZA_CLOUD_PROVISIONED=1 (401), 0.0.0.0 bind (1008) (test.ts:185-216): cross-site pages cannot ride ambient cookies even when cloud CORS reflects them
- revocation denies the very next connection (test.ts:218-222)

Adversarial: no lockfile, dependency, CI, lifecycle-hook or telemetry changes; no binaries, symlinks or submodules; tests additive with strict toEqual assertions, none weakened.

Nits: the untrusted-origin test pins three distinct rejection modes (403/401/1008), encoding which layer rejects rather than the security property itself; a refactor unifying rejection would break the test while preserving the guarantee. isCredentialedCorsOrigin and resolveSessionTokenRole are outside this diff, so their internals are behaviorally exercised here, not source-verified.

Material: security behavior change with demonstrated failure modes; not cosmetic churn, not test-only.

Verified by reading the diff at this head only; no checkout, no suite run. Live execution blocked: no disposable sandbox on this host.

Recommendation: merge

PR #31193, opened 2026-09-13
fix(cloud): allow concurrent inference policy readers

### LINKED ISSUE #31191: Dedicated inference serializes policy reads behind exclusive organization locks
A production Dedicated first-message test took 37.852 seconds from submission to finalized reply. Five model requests reported 26.729 seconds of gateway pre-forward work in total (the parallel evaluator requests overlap, so this sum is not wall time). The memory creation itself took 15 milliseconds.

The current rate limiter, organization inference admission and final dispatch fence each read policy inside a transaction holding `FOR UPDATE` on the organization row. Parallel model requests for the same agent therefore serialize their policy checks, including remote Durable Object calls performed while holding the row lock. The inference hot-path guide still describes a no-Postgres path, which is not the current implementation.

Scope: establish a real PostgreSQL reproduction, then permit shared policy locks only for audited inference reads whose rate/money decisions are separately serialized by Durable Objects. Preserve exclusive locks for policy publication, resource admission and other mutations. Keep the final fresh-policy check, authorization/revocation, money recovery, rate limits and complete runtime memory unchanged. Reject the candidate if real integration evidence does not support it.

Acceptance: concurrent inference readers can proceed; a policy writer and credit update still conflict and order correctly; later admission sees committed policy changes; resource-count admission remains exclusive; owning tests/typecheck/lint and root verification pass; real Dedicated app traces show the result without hiding regressions. This issue does not authorize a broader runtime redesign or Shared/Telegram changes.

CLAIMING: bounded Dedicated inference latency diagnosis and read-lock candidate in `codex/dedicated-inference-read-locks-20260913`. No deployment is part of this initial claim.

NubsCarson:
Parallel model requests from the same Dedicated agent serialize on exclusive organization-row locks even when the guarded operation only reads policy and calls the authoritative rate or spend Durable Object. Worker inference now uses shared row locks for those checks, allowing readers to overlap while policy changes, credit updates and resource admission retain their existing fences. Final policy-generation, rate, credential and billing checks remain in place.

Closes #31191. The change is limited to the rate check, initial inference policy read and final dispatch check. Non-Worker admission, resource allocation and cache publication retain exclusive locks. Model context, provider selection and database schema are unchanged.

Validation at `298a12862f6`:

- Real PostgreSQL: all five integration cases pass. Temporarily restoring the old exclusive reader lock makes the new concurrent-reader case fail with `Concurrent inference readers serialized behind one another`; restoring the candidate passes. The tests also prove override publication invalidates the old stamp and credit/resource writers wait.
- Existing focused rate/inference contracts: 23 tests pass.
- Shared package typecheck and lint pass; root `bun run verify` passes.
- Five suites interrupted by a concurrent dependency rebuild were rerun after the build: 108 pass, one existing Docker opt-in skip, zero failures. The complete shared package rerun finished successfully: 388 batches, 12,681 passing tests, 245 skips and zero failures (one existing Bun/PGlite status-99 normalization).

Production baseline from the same QA node and configured Qwen model: four correct responses took 4.1–4.9 seconds, including 3.9–4.6 seconds before provider dispatch. This PR removes reader contention; the remaining serial database/network cost still needs live measurement and follow-up. No latency improvement or deployment is claimed yet.

Hosted CI: PostgreSQL authority and Windows security checks pass. Billing replay reaches its real settlement/idempotency assertions, then fails at browser navigation to `/cloud/billing`: the screenshot and trace show the Launcher and no `/api/v1/billing/limits` request. This matches the management deep-link defect observed before this patch on production `/cloud/agents`; it remains an open app-shell issue. Source static smoke is still running.

### DIFF
--- a/packages/cloud/shared/src/db/repositories/organization-policy-generation.ts
+++ b/packages/cloud/shared/src/db/repositories/organization-policy-generation.ts
@@ -9,11 +9,25 @@ export async function lockOrganizationPolicy(
   tx: DbTransaction,
   organizationId: string,
 ): Promise<void> {
+  return lockPolicyRow(tx, organizationId, "update");
+}
+/** Read-only admission may overlap other readers while still fencing policy and balance writers. */
+export async function lockOrganizationPolicyForRead(
+  tx: DbTransaction,
+  organizationId: string,
+): Promise<void> {
+  return lockPolicyRow(tx, organizationId, "share");
+}
+async function lockPolicyRow(
+  tx: DbTransaction,
+  organizationId: string,
+  strength: "update" | "share",
+): Promise<void> {
   const [organization] = await tx
     .select({ id: organizations.id })
     .from(organizations)
     .where(eq(organizations.id, organizationId))
-    .for("update");
+    .for(strength);
   if (!organization)
     throw new ElizaError("Organization policy authority does not exist", {
       code: "ORGANIZATION_POLICY_UNAVAILABLE",
--- a/packages/cloud/shared/src/db/repositories/organization-policy.postgres.integration.test.ts
+++ b/packages/cloud/shared/src/db/repositories/organization-policy.postgres.integration.test.ts
@@ -59,6 +59,99 @@ describe.skipIf(!databaseUrl)("organization policy primary transaction fences", 
       await writer.end();
     }
   });
+  test("parallel inference policy readers overlap while a policy writer waits and invalidates their stamp", async () => {
+    const { withOrganizationPolicyReadAdmission } = await import(
+      "../../lib/services/organization-policy-admission"
+    );
+    const { requireOrganizationRateTier } = await import(
+      "../../lib/services/organization-quota-policy"
+    );
+    const { orgRateLimitOverridesRepository } = await import("./org-rate-limit-overrides");
+    const entered = [barrier(), barrier()];
+    const release = barrier();
+    const readers = entered.map((ready) =>
+      withOrganizationPolicyReadAdmission(ORG, undefined, async (policy) => {
+        ready.resolve();
+        await release.promise;
+        return policy.authority;
+      }),
+    );
+    let mutation: Promise<unknown> | undefined;
+    try {
+      // Both callbacks must be reached before either releases its transaction.
+      // Exclusive reader locks cannot pass this boundary.
+      await Promise.race([
+        Promise.all(entered.map((ready) => ready.promise)),
+        Bun.sleep(3000).then(() => {
+          throw new Error("Concurrent inference readers serialized behind one another");
+        }),
+      ]);
+      let mutationFinished = false;
+      mutation = orgRateLimitOverridesRepository
+        .upsert({ organization_id: ORG, completions_rpm: 12 }, "admin:postgres-test")
+        .then(() => {
+          mutationFinished = true;
+        });
+      await waitForOrganizationLock();
+      expect(mutationFinished).toBe(false);
+      release.resolve();
+      const stamps = await Promise.all(readers);
+      await mutation;
+      expect(stamps[0]).toEqual(stamps[1]);
+      await expect(
+        withOrganizationPolicyReadAdmission(ORG, stamps[0], async () => "dispatched"),
+      ).rejects.toMatchObject({ code: "ORGANIZATION_POLICY_STALE" });
+      expect(
+        await withOrganizationPolicyReadAdmission(
+          ORG,
+          undefined,
+          async (policy) => requireOrganizationRateTier(policy).completionsRpm,
+        ),
+      ).toBe(12);
+    } finally {
+      release.resolve();
+      await Promise.allSettled(readers);
+      await mutation;
+    }
+  }, 30000);
+  test("an inference reader still fences credit updates and exclusive resource admission", async () => {
+    const { sql } = await import("drizzle-orm");
+    const { withOrganizationPolicyReadAdmission, withOrganizationPolicyAdmission } = await import(
+      "../../lib/services/organization-policy-admission"
+    );
+    for (const operation of ["credit", "resource"] as const) {
+      const entered = barrier(),
+        release = barrier();
+      const reader = withOrganizationPolicyReadAdmission(ORG, undefined, async () => {
+        entered.resolve();
+        await release.promise;
+      });
+      let mutation: Promise<unknown> | undefined;
+      try {
+        await entered.promise;
+        let finished = false;
+        mutation = (
+          operation === "credit"
+            ? database.dbWrite.execute(
+                sql`UPDATE organizations SET credit_balance=credit_balance-1 WHERE id=${ORG}`,
+              )
+            : withOrganizationPolicyAdmission(ORG, undefined, async () => "resource admitted")
+        ).then(() => {
+          finished = true;
+        });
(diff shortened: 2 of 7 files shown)


attentionhead (approved):
Reviewed at head `298a12862f6931d6924e67ba83cf2ebdf5225795` (base `develop`, +168/-17 across 7 files).

## Verdict: approve

Weakening a row lock from `FOR UPDATE` to `FOR SHARE` is only safe if **nothing inside the shared-lock transaction writes Postgres** — two `FOR SHARE` holders can proceed concurrently, and if either then tried to write the same row it would attempt a lock upgrade and the pair would deadlock. So I traced every callback that now runs under `lockOrganizationPolicyForRead` to see whether one can reach a Postgres mutation.

### Both read-lock callbacks are Postgres-free

**`enforceOrgRateLimit(..., { cacheOnly: true })`** — the `cacheOnly` branch computes `authoritativeConfig` from the locked policy read, re-checks any supplied observation against `policy.authority`, and then calls `consumeInferenceRateLimit`, which is the Durable Object limiter. It returns `null` or a response; it issues no statement. The `error-policy:J4` handler converts `InferenceAdmissionGateUnavailableError` into `rateLimitUnavailableResponse()`, so a limiter outage still fails closed rather than falling through to the non-`cacheOnly` Redis lane. The Redis/lease path below it — which mutates `orgRateLimitLeases` and talks to Redis — is reached only when `cacheOnly` is false, i.e. under the unchanged exclusive lock.

**`markProviderDispatched`** in `admitOrganizationInference` is the one I expected to be the problem, since `admitDispatch` selects the lock strength but the callback body is identical either way. It is not, and for two distinct reasons depending on where `admission` came from:

- On the deferred worker lane, `previousDispatch` is `attachInferenceAdmissionLease`'s closure (`organization-inference-admission.ts:244`), whose entire body is `await markInferenceAdmissionLeaseDispatched(lease)` plus `error-policy:J2` re-typing. That is a Durable Object call.
- On the synchronous lane, `reserveSynchronously` (`:202`) returns `{ mode: "synchronous_reservation", settle, settleUnknown, affiliateAttribution, reservation }` — it supplies **no** `markProviderDispatched` at all, and the interface declares it optional (`:103`). So `await previousDispatch?.()` is a no-op. This matters specifically for the `policy.subscriptionFunded` branch, which returns `reserveSynchronously` *before* `admitWithFundingPolicy` consults `workerHotPath`: that is the one combination where a worker request carries a synchronous-reservation admission into the shared-lock dispatch, and it is inert.

The only work left inside that callback is the four `requireOrganizationRateTier` comparisons against `authoritativePolicy`, which are pure reads of the freshly locked row, and the `dispatched = true` assignment.

### The fencing the change is supposed to preserve is preserved

`FOR SHARE` conflicts with `FOR UPDATE` and with plain `UPDATE` row locks, so a policy/override publisher or a `credit_balance` writer still blocks behind an in-flight reader — the readers only stop serializing against *each other*. The second new integration test pins exactly this in both directions (a raw `UPDATE organizations SET credit_balance=credit_balance-1` and a `withOrganizationPolicyAdmission` resource admission each observe `finished === false` while a reader holds the lock), which is the right pair to assert, since it is the property the weakening could plausibly have destroyed.

`admitUnderPolicyLock` keeps both lock strengths inside `writeTransaction`, so readers still read the primary — the change does not quietly become a replica read — and the `expected`/`sameOrganizationPolicyStamp` CAS is unchanged and shared by both entry points. The first new test confirms the stamp is still invalidated by a concurrent writer (`ORGANIZATION_POLICY_STALE`) and that the subsequent read observes `completionsRpm === 12`, so the relaxed lock has not made the staleness fence vacuous.

The 3s race guard in that test is what makes it non-vacuous in the other direction: under the old `FOR UPDATE` the two readers could not both reach their callbacks, so the test fails on the pre-change code rather than passing incidentally.

The `rate-limit-org-lease.test.ts` change hoists the existing inline mock into `withFixturePolicy` and binds it to both exported names, which keeps `policyReads` counting the same way — the mock module must enumerate every member the subject imports, and `rate-limit.ts` now imports both.

## Non-blocking notes

1. **`workerHotPath` is derived twice from a structural probe.** `typeof params.executionCtx?.waitUntil === "function"` is computed independently in `admitOrganizationInference` (:343) and again in `admitWithFundingPolicy` (:406). They agree today because both read the same `params.executionCtx`, but the value now selects a *lock strength* in one and a *billing lane* in the other, so a future caller that synthesizes a partial `executionCtx` would desynchronize a safety decision from the funding decision. A single computed field on `params`, or a named helper, would make the coupling explicit.

2. **Nothing enforces the "callback must not mutate Postgres" contract beyond the doc comment.** That comment on `withOrganizationPolicyReadAdmission` is currently accurate, and it is the load-bearing invariant for the whole change; but a future edit that adds a write inside the `cacheOnly` branch of `enforceOrgRateLimit` would produce an intermittent lock-upgrade deadlock under concurrency rather than a test failure. If there is a cheap way to pass the read lane a transaction handle typed as read-only, that would convert the invariant into a compile error.

3. **`lockPolicyRow`'s `strength` parameter is stringly typed against Drizzle's `.for()`.** `"update" | "share"` is narrow enough that this is fine, but since the two exported wrappers are the only callers, the union could be dropped in favour of passing the lock mode as the already-narrowed literal at each call site — one less place for a typo to silently pick the wrong strength.

ss251 (approved):
298a12862f6931d6924e67ba83cf2ebdf5225795

Mechanism is correct: `lockPolicyRow` parameterizes Drizzle's `.for(strength)`; the existing lock keeps `"update"` while the new `lockOrganizationPolicyForRead` takes `"share"` (organization-policy-generation.ts:12-25). FOR SHARE overlaps concurrent readers yet still conflicts with row updates and FOR UPDATE, so override publication and balance writers stay fenced — exactly the claimed semantics.

The integration tests prove both directions. Two read-admissions must both clear an entry barrier or a 3s guard throws "Concurrent inference readers serialized behind one another" — that assertion fails against the old exclusive lock. An override upsert started while both readers hold the lock is asserted unfinished; after release, re-admission with the pre-write stamp rejects `ORGANIZATION_POLICY_STALE` and a fresh read observes rpm 12 (organization-policy.postgres.integration.test.ts:62-110). A second test holds a `credit_balance` UPDATE and an exclusive `withOrganizationPolicyAdmission` behind a reader (:111-157). These run only under `skipIf(!databaseUrl)`, so CI must exercise them on live Postgres for the guarantees to be real.

Residual the diff cannot close: both production switches gate on code outside this hunk set — `options.cacheOnly` (rate-limit.ts:465-467) and the `waitUntil` heuristic (organization-inference-admission.ts:342-343) — and FOR SHARE safety rests entirely on the documented contract that the callback performs no Postgres mutation (organization-policy-admission.ts:24-28). The `cacheOnly` callback body and the `markProviderDispatched` body (organization-inference-admission.ts:381, truncated here) must contain no read-modify-write on shared rows; the new tests cover lock semantics, not those two bodies. The share lock also spans only the policy-read transaction, so funding serialization is unchanged. The lease-test mock now stubs the read admission identically, preserving the `policyReads` counter (rate-limit-org-lease.test.ts:69-81).

A production concurrency change on the inference hot path with behavioral tests attached: material, correctly scoped, maintainable.

Recommendation: merge

PR #31194, opened 2026-09-13
fix(cloud): stage concurrent inference policy reads

NubsCarson:
Parallel model requests from the same Dedicated agent serialize on exclusive organization-row locks even when the guarded operation only reads policy and calls the authoritative rate or spend Durable Object. Worker inference now uses shared row locks for those checks, allowing readers to overlap while policy changes, credit updates and resource admission retain their existing fences. Final policy-generation, rate, credential and billing checks remain in place.

Staging-only promotion of #31193 for #31191. The change is limited to the rate check, initial inference policy read and final dispatch check. Non-Worker admission, resource allocation and cache publication retain exclusive locks. Model context, provider selection and database schema are unchanged.

Source validation at `298a12862f6` (identical seven changed files):

- Real PostgreSQL: all five integration cases pass. Temporarily restoring the old exclusive reader lock makes the new concurrent-reader case fail with `Concurrent inference readers serialized behind one another`; restoring the candidate passes. The tests also prove override publication invalidates the old stamp and credit/resource writers wait.
- Existing focused rate/inference contracts: 23 tests pass.
- Shared package typecheck and lint pass; root `bun run verify` passes.
- Five suites interrupted by a concurrent dependency rebuild were rerun after the build: 108 pass, one existing Docker opt-in skip, zero failures. The complete shared package rerun is still running; its result will be recorded here.

Production baseline from the same QA node and configured Qwen model: four correct responses took 4.1–4.9 seconds, including 3.9–4.6 seconds before provider dispatch. This PR removes reader contention; the remaining serial database/network cost still needs live measurement and follow-up. No latency improvement or deployment is claimed yet.

The staging tree contains exactly this seven-file patch over the currently served staging SHA `342569fcaef58dbd58b79eff2f17a1996b149d80`; unrelated develop changes are excluded. Production promotion remains dependent on staged live validation.

### DIFF
--- a/packages/cloud/shared/src/db/repositories/organization-policy-generation.ts
+++ b/packages/cloud/shared/src/db/repositories/organization-policy-generation.ts
@@ -9,11 +9,25 @@ export async function lockOrganizationPolicy(
   tx: DbTransaction,
   organizationId: string,
 ): Promise<void> {
+  return lockPolicyRow(tx, organizationId, "update");
+}
+/** Read-only admission may overlap other readers while still fencing policy and balance writers. */
+export async function lockOrganizationPolicyForRead(
+  tx: DbTransaction,
+  organizationId: string,
+): Promise<void> {
+  return lockPolicyRow(tx, organizationId, "share");
+}
+async function lockPolicyRow(
+  tx: DbTransaction,
+  organizationId: string,
+  strength: "update" | "share",
+): Promise<void> {
   const [organization] = await tx
     .select({ id: organizations.id })
     .from(organizations)
     .where(eq(organizations.id, organizationId))
-    .for("update");
+    .for(strength);
   if (!organization)
     throw new ElizaError("Organization policy authority does not exist", {
       code: "ORGANIZATION_POLICY_UNAVAILABLE",
--- a/packages/cloud/shared/src/db/repositories/organization-policy.postgres.integration.test.ts
+++ b/packages/cloud/shared/src/db/repositories/organization-policy.postgres.integration.test.ts
@@ -59,6 +59,99 @@ describe.skipIf(!databaseUrl)("organization policy primary transaction fences", 
       await writer.end();
     }
   });
+  test("parallel inference policy readers overlap while a policy writer waits and invalidates their stamp", async () => {
+    const { withOrganizationPolicyReadAdmission } = await import(
+      "../../lib/services/organization-policy-admission"
+    );
+    const { requireOrganizationRateTier } = await import(
+      "../../lib/services/organization-quota-policy"
+    );
+    const { orgRateLimitOverridesRepository } = await import("./org-rate-limit-overrides");
+    const entered = [barrier(), barrier()];
+    const release = barrier();
+    const readers = entered.map((ready) =>
+      withOrganizationPolicyReadAdmission(ORG, undefined, async (policy) => {
+        ready.resolve();
+        await release.promise;
+        return policy.authority;
+      }),
+    );
+    let mutation: Promise<unknown> | undefined;
+    try {
+      // Both callbacks must be reached before either releases its transaction.
+      // Exclusive reader locks cannot pass this boundary.
+      await Promise.race([
+        Promise.all(entered.map((ready) => ready.promise)),
+        Bun.sleep(3000).then(() => {
+          throw new Error("Concurrent inference readers serialized behind one another");
+        }),
+      ]);
+      let mutationFinished = false;
+      mutation = orgRateLimitOverridesRepository
+        .upsert({ organization_id: ORG, completions_rpm: 12 }, "admin:postgres-test")
+        .then(() => {
+          mutationFinished = true;
+        });
+      await waitForOrganizationLock();
+      expect(mutationFinished).toBe(false);
+      release.resolve();
+      const stamps = await Promise.all(readers);
+      await mutation;
+      expect(stamps[0]).toEqual(stamps[1]);
+      await expect(
+        withOrganizationPolicyReadAdmission(ORG, stamps[0], async () => "dispatched"),
+      ).rejects.toMatchObject({ code: "ORGANIZATION_POLICY_STALE" });
+      expect(
+        await withOrganizationPolicyReadAdmission(
+          ORG,
+          undefined,
+          async (policy) => requireOrganizationRateTier(policy).completionsRpm,
+        ),
+      ).toBe(12);
+    } finally {
+      release.resolve();
+      await Promise.allSettled(readers);
+      await mutation;
+    }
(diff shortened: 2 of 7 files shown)


attentionhead (approved):
Reviewed at head `64231c41742d71e2764885b9ff97447fcaa92675` (base `staging`, +168/-17 across 7 files).

## Verdict: approve

This is the `staging` counterpart of #31193, and the diff bytes are identical to it across all seven files. I did not treat that as sufficient on its own: the patch is the same but the surrounding tree is not, and the verdict here rests on facts that live **outside** the diff. I re-checked each of those against this head rather than carrying them over.

### Why the lock weakening is safe

Relaxing `FOR UPDATE` to `FOR SHARE` is only correct if nothing inside the shared-lock transaction writes Postgres. Two `FOR SHARE` holders run concurrently, so a write to the same row from inside either one is not merely a lost mutex — it is a lock upgrade, and two readers upgrading at once deadlock. Both callbacks that `lockOrganizationPolicyForRead` now covers stay off Postgres:

- **`enforceOrgRateLimit(..., { cacheOnly: true })`** derives `authoritativeConfig` from the locked read, re-checks any supplied observation against `policy.authority`, and then calls `consumeInferenceRateLimit` (`rate-limit.ts:491`) — the Durable Object limiter. No statement is issued. The `error-policy:J4` handler at `:502` converts `InferenceAdmissionGateUnavailableError` into `rateLimitUnavailableResponse()`, so a limiter outage still fails closed instead of falling through into the Redis/lease lane, which remains reachable only when `cacheOnly` is false — i.e. under the unchanged exclusive lock.

- **`markProviderDispatched`** is the case that looked dangerous, because `admitDispatch` selects the lock strength while the callback object is the same either way. It is safe for two different reasons depending on provenance, and I confirmed both on this head:
  - the deferred lane's closure (`organization-inference-admission.ts:244`) is `await markInferenceAdmissionLeaseDispatched(lease)` plus `error-policy:J2` re-typing — a Durable Object call;
  - `reserveSynchronously` (`:202`) returns `{ mode: "synchronous_reservation", settle, settleUnknown, affiliateAttribution, reservation }` and supplies **no** `markProviderDispatched` at all, with the interface declaring it optional (`:103`), so `await previousDispatch?.()` is a no-op. That is the path taken by the `policy.subscriptionFunded` branch, which returns *before* `admitWithFundingPolicy` consults `workerHotPath` — the one combination that carries a synchronous-reservation admission into the shared-lock dispatch.

What remains inside the callback is four `requireOrganizationRateTier` comparisons against the freshly locked row and a `dispatched = true` assignment.

### What the change still fences

`FOR SHARE` conflicts with `FOR UPDATE` and with plain `UPDATE` row locks, so policy publishers, override writes and `credit_balance` updates still block behind an in-flight reader; only reader-against-reader serialization is dropped. The second new integration test asserts exactly that in both directions — a raw `UPDATE organizations SET credit_balance=credit_balance-1` and a `withOrganizationPolicyAdmission` resource admission each observe `finished === false` while a reader holds the lock — which is the property the weakening could plausibly have destroyed.

`admitUnderPolicyLock` keeps both strengths inside `writeTransaction`, so the relaxed path still reads the primary and does not silently become a replica read, and the `expected` / `sameOrganizationPolicyStamp` CAS is shared by both entry points unchanged. The first new test keeps the staleness fence honest (`ORGANIZATION_POLICY_STALE`, then an observed `completionsRpm === 12`), and its 3s race guard is what makes it non-vacuous in the other direction: under the old `FOR UPDATE` the two readers could not both reach their callbacks, so it fails on pre-change code rather than passing incidentally.

The `rate-limit-org-lease.test.ts` change hoists the inline mock into `withFixturePolicy` and binds it to both exported names, which it must — `rate-limit.ts` now imports both, and a module mock has to enumerate every member the subject imports.

## Non-blocking notes

These carry over from the `develop` PR; they are observations about the shape of the change, not objections to promoting it.

1. **`workerHotPath` is derived twice from the same structural probe.** `typeof params.executionCtx?.waitUntil === "function"` is computed independently in `admitOrganizationInference` and again in `admitWithFundingPolicy`. They agree today, but the value now selects a *lock strength* in one and a *billing lane* in the other, so a caller that synthesizes a partial `executionCtx` would desynchronize a concurrency-safety decision from a funding decision. One computed field on `params` would make the coupling explicit.

2. **The "callback must not mutate Postgres" contract is enforced only by a doc comment.** It is accurate today and it is the load-bearing invariant for the entire change, but a later edit adding a write inside the `cacheOnly` branch would surface as an intermittent lock-upgrade deadlock under concurrency rather than a failing test. Passing the read lane a handle typed as read-only would turn that into a compile error.

3. **`lockPolicyRow`'s `strength` parameter is stringly typed** against Drizzle's `.for()`. The `"update" | "share"` union is narrow enough to be safe, and the two exported wrappers are its only callers, so this is cosmetic.

ss251 (approved):
64231c41742d71e2764885b9ff97447fcaa92675

1. Lock staging is sound. lockPolicyRow (organization-policy-generation.ts:22) parameterizes `.for("update"|"share")`, so read admission overlaps readers while FOR SHARE still conflicts with FOR UPDATE and plain row UPDATEs. The new integration tests encode the acceptance criteria: both readers must cross their barrier inside open transactions (an exclusive lock cannot), the override upsert stays unfinished while readers hold the row, then the stale stamp rejects ORGANIZATION_POLICY_STALE and a fresh read returns completions_rpm=12; the credit-balance UPDATE and the exclusive resource admission are asserted to fence behind the shared reader. They run against a real database (skipIf(!databaseUrl)) and fail on regression.

2. workerHotPath (organization-inference-admission.ts:343) infers the hot path from `typeof executionCtx?.waitUntil === "function"`; a caller passing an executionCtx for unrelated reasons silently gets the weaker fence on both admission and the dispatch re-check (line 371). An explicit parameter would state intent.

3. withOrganizationPolicyReadAdmission (organization-policy-admission.ts:27) requires the callback not mutate Postgres, but the dispatch-path callback behind admitDispatch (organization-inference-admission.ts:381) is outside this diff; if it writes Postgres after the CAS check, the shared lock no longer fences it. Please confirm that write is DO-serialized, per the function's own contract comment.

4. rate-limit.ts:465 routes only cacheOnly enforcement to the read admission, keeping the exclusive path otherwise; the lease-test mock covers both entries with unchanged read counting — no test weakening. No CI, lockfile, secret, or telemetry surface in the diff.

Verified by static reading of the diff at this head; the database-gated integration tests were not executed, and the dispatch callback body and waitForOrganizationLock helper lie outside the diff, so items 2–3 are held open rather than asserted failures. Live execution blocked: no disposable sandbox on this host.

Recommendation: merge

Issue #31180, opened 2026-09-12
feat(lifeops): export and delete a reviewed family workspace


lalalune:
## Observed missing product flow

The Bettina PDD requires a workspace export (AUDIT-4), a dependency preview before deletion (AUDIT-5), immediate revocation (AUDIT-6), and documented primary/derived/cache/backup purge behavior (AUDIT-7). The merged agreement export in #31165 covers one source artifact and its review/access history. `AgreementKnowledgeService` and the owner Family Operations routes have no workspace lifecycle/dependency-preview/delete flow. The owner therefore cannot complete the PDD's export-and-delete test workspace journey from the product.

## Acceptance

- Provide an owner-only manifest export joining the selected family workspace's immutable source artifacts, hashes, reviewed obligations, packet versions/approvals, and recorded calendar/message receipts. Preserve provenance and distinguish unavailable legacy history; never include credentials or fabricate receipts.
- Before deletion, show the exact affected source versions, pins, obligations, authority baselines, workflows, packets, and local calendar dependencies. Bind confirmation to the reviewed dependency snapshot; reject stale previews and concurrent additions instead of silently broadening deletion.
- Document and implement the revocation/deletion ordering and retention policy across primary bytes, derived records, cached projections, and eligible backups. Retained audit/receipt metadata must be explicit and must not retain deleted private source content. Provider events/messages and unrelated workspaces must not be deleted implicitly.
- Expose the journey through the real owner UI with loading, denied, stale-preview, partial-failure, and completion states. Guest grants, pinned context, links, and downloads must lose access at the documented revocation boundary.
- Prove the behavior with real storage/integration tests, owning package and root gates, exact-head hosted checks, and a synthetic live pilot workspace with independently inspected export and post-deletion reads. Preserve unrelated live family data and the original messaging service.

Part of #30123. Complements #31162 and #31172; it does not replace the remaining live Google, school-workflow, messaging, or final merged-deployment acceptance. Selected-correspondence intake remains in #31063/#31032.

lalalune:
[...]

PR #31195, opened 2026-09-13
feat(lifeops): export owner family workspace with verified source archives

lalalune:
Owners can download a family workspace ZIP from Family Operations. It contains versioned agreement archives, packet-bound drafts and approval receipts, available school records, and retained school source bytes, with manifests and checksums. Authorization and source-integrity failures return errors; unrelated approval records and executor lease data are excluded.

Related to #31180. This PR implements export only. Dependency preview, deletion/purge, and live pilot acceptance remain open and are not claimed complete.

Validation on `b6d1f16897217726ccc73fe88b89d8e122e1bd9d`, based on `1fa9dbf22d741d5c0c0940b8cbd4eafc1403ad73`:

- Root `bun run verify`: 373 tasks passed; full personal-assistant tests: 2,660 passed, 6 skipped.
- Real storage/HTTP integration: 16 passed; adapter/component tests: 25 passed. The HTTP regression exercises real owner and guest AuthStore sessions.
- Desktop/mobile browser runs use the actual view, production adapter, route and PGlite/storage service. Actual downloaded ZIPs passed independent outer/nested checksums and source-byte verification. Loading, success, and denied states were inspected.
- Required app audit: 226 capture tests passed. OCR: 220 views, 208 verified, no broken views; family mobile views requiring inspection were manually reviewed.

Evidence limitations: browser recordings use the local operator-trust boundary and synthetic data, including deterministic PDF extraction and historical stored provider receipt fixtures. No external message or calendar mutation occurred. Before screenshots reconstruct the committed baseline view through the same renderer configuration. Videos include 1.5 seconds of transport delay to expose loading. The archive is explicit about component snapshot boundaries; it does not claim a transaction spanning every component. Public pilot acceptance is pending.

- [x] desktop before mobile before

- [x] desktop after mobile after

- [x] [desktop walkthrough](https://github.com/elizaOS/eliza/releases/download/pr-evidence-12/31180-b6d1f16-desktop-walkthrough.mp4) [mobile walkthrough](https://github.com/elizaOS/eliza/releases/download/pr-evidence-12/31180-b6d1f16-mobile-walkthrough.mp4)

- [x] [b6d1f16-browser-backend.log](https://github.com/elizaOS/eliza/releases/download/pr-evidence-12/31180-b6d1f16-browser-backend.log)

- [x] [b6d1f16-frontend-console.json](https://github.com/elizaOS/eliza/releases/download/pr-evidence-12/31180-b6d1f16-frontend-console.json) [b6d1f16-frontend-network.json](https://github.com/elizaOS/eliza/releases/download/pr-evidence-12/31180-b6d1f16-frontend-network.json) Expected HTTP 401 diagnostics occur only in the denied test; no browser page exceptions.

- [x] N/A - export does not change model, prompt, provider, or action behavior.

- [x] [Complete reviewed artifacts and actual downloads](https://github.com/elizaOS/eliza/releases/download/pr-evidence-12/31180-bettina-workspace-export-b6d1f16-evidence.zip) [Independent ZIP verification](https://github.com/elizaOS/eliza/releases/download/pr-evidence-12/31180-b6d1f16-domain-verification.json)

- [x] [OCR and manual family-view review](https://github.com/elizaOS/eliza/releases/download/pr-evidence-12/31180-b6d1f16-ocr-review.json)

Native/device evidence: N/A - this change adds a browser export; no native implementation changes.

### DIFF
--- a/plugins/plugin-personal-assistant/README.md
+++ b/plugins/plugin-personal-assistant/README.md
@@ -435,3 +435,17 @@ internals; it consumes the plugin's public exports only. See
 - Prompt-content lint rules: `scripts/lint-default-packs.mjs`.
 - Health domain: `plugins/plugin-health/README.md`.
 - REST routes: `src/routes/`.
+
+### Family workspace export
+
+The owner-only `POST /api/lifeops/family-workflows/export` downloads a ZIP with
+all family agreement versions, retained school PDFs, monthly packet versions
+and drafts, their approval records, and stored provider/school mutation receipts.
+Each member has a SHA-256 checksum. Agreement archives preserve their existing
+source/extraction and review/access provenance; packet and workflow records use
+one database statement snapshot. The manifest records this component-snapshot
+boundary and explicitly identifies uninitialized historical record stores.
+Missing or changed retained PDF bytes fail export rather than producing a
+healthy-looking partial archive. Connection credentials and executor lease
+tokens are excluded. Export records preparation, not receipt by the client,
+and does not revoke access or delete data.
--- a/plugins/plugin-personal-assistant/src/components/family-operations/FamilyOperationsView.stories.tsx
+++ b/plugins/plugin-personal-assistant/src/components/family-operations/FamilyOperationsView.stories.tsx
@@ -16,6 +16,9 @@ const adapter = {
     packets: { status: "ready", data: [] },
     emailOptions: { status: "ready", data: { accounts: [], recipients: [] } },
   }),
+  downloadWorkspace: async () => {
+    throw new Error("This preview has no stored workspace to export.");
+  },
   decideObligation: async (obligation: never) => obligation,
   listPins: async () => [],
   pin: async () => null,
--- a/plugins/plugin-personal-assistant/src/components/family-operations/FamilyOperationsView.test.tsx
+++ b/plugins/plugin-personal-assistant/src/components/family-operations/FamilyOperationsView.test.tsx
@@ -132,6 +132,9 @@ function adapter(data = snapshot()): FamilyOperationsAdapter {
     generatePacket: vi.fn(async () => undefined),
     uploadAgreement: vi.fn(async () => undefined),
     downloadAgreement: vi.fn(async () => new Blob()),
+    downloadWorkspace: vi.fn(
+      async () => new Blob([], { type: "application/zip" }),
+    ),
     createPacketDraft: vi.fn(async () => undefined),
     revisePacketDraft: vi.fn(async () => undefined),
     requestPacketApproval: vi.fn(async () => undefined),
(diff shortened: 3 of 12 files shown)


attentionhead (approved):
Reviewed at head `b6d1f16897217726ccc73fe88b89d8e122e1bd9d` (draft, base `develop`, +696/-25 across 12 files).

## Verdict: approve

The load-bearing question for an owner-data export is who can invoke it and what it is allowed to emit. Both hold here, and the new integration tests pin the failure directions rather than only the happy path.

### Authorization

`POST /api/lifeops/family-workflows/export` is registered in `LIFEOPS_STATIC_ROUTES` with no `access` field, and the registry builder routes exactly that case to `withOwnerAdminGate(handler)` — only a spec carrying `access: "authenticated_entity"` takes the guest gate (`plugin.ts:777`). So the guest/authenticated-entity surface is untouched by this PR. That matters because the handler passes `SELF_ENTITY_ID` unconditionally rather than deriving the owner from the request: the route gate, not the argument, is what keeps a non-owner out, and the gate is the strict one. `exportFamilyWorkspace` then re-validates through `listOwnerAgreements({ ownerEntityId })` before any private bytes are read, so a caller-supplied household id cannot widen the scope even if a future caller reaches the function directly.

### What the archive is allowed to contain

Every record query is projected column-by-column and scoped `agent_id = <runtime.agentId>`, which is what keeps executor leases and connection credentials out even as those schemas grow. The second new integration test makes that a real assertion rather than a claim: it writes two rows with the same `run_id` — one for the agent, one for a foreign agent — each carrying `apply_lease_token = 'internal-executor-lease-canary'`, then asserts the manifest contains neither the canary nor `other-agent-private-school`. A regression in either the projection or the scope predicate fails it.

The `approval_requests` join is the one table read outside the `app_lifeops` set, and it is correctly narrowed to ids already referenced by `life_family_packet_approvals` for the same agent, so an unrelated approval payload cannot ride along — pinned by the third new test. It is also the one table joined without a `to_regclass` availability probe, which is consistent rather than an oversight: the eight probed tables are created lazily by their services (`CREATE TABLE IF NOT EXISTS` in `calendar-workflow.ts:60`), while `approval_requests` is a migrated schema table that is always present.

### Integrity and the failure direction

Retained school bytes are re-read from canonical storage and re-hashed against the recorded `content_sha256`, with the recorded `media_url` cross-checked to resolve to `<digest>.pdf` before the read. A mismatch or a missing object raises `FAMILY_EXPORT_SOURCE_INTEGRITY` and aborts, so the export cannot degrade into a healthy-looking partial archive — and the test proves the abort is clean by deleting the stored object and asserting both the rejection code *and* that the `family_workspace_export_prepared` audit count is unchanged. Since the audit insert runs after the archive is built, that assertion is what establishes no export is recorded as prepared when the bytes could not be verified.

`SHA256SUMS` is computed over every entry including `manifest.json` and excluding itself, and the manifest's `receiptCoverage` and audit `reason` both state that only stored receipts are included and that client download is not asserted — the correct framing, since the handler returns bytes and cannot observe delivery.

The adapter's binary path is likewise honest about failures: a non-OK response only parses a body as an error message when the content type is actually JSON, and otherwise preserves `Download failed (<status>)` (`error-policy:J1`). `downloadWorkspace` additionally rejects a 200 whose blob is not `application/zip`, so an HTML error page proxied with a success status cannot be saved as an archive — both covered by the new `it.each` cases.

## Non-blocking notes

1. **Record ordering is locale-sensitive.** `rows.sort((l, r) => JSON.stringify(l).localeCompare(JSON.stringify(r)))` orders by the host's collation, so the same database can produce manifests that differ in record order — and therefore in `manifestSha256` — across machines. For an artifact whose stated value is verifiability, a plain `<` comparison (or `localeCompare` with a fixed locale) would make the manifest reproducible. Nothing today asserts cross-host equality, so this is latent.

2. **The `media_url` validation escapes the module's own error contract.** Everything else in this file fails through `failure(message, code)` with a `FAMILY_EXPORT_*` code, but the `z.string().regex(...)` calls on `content_sha256` and `media_url` throw a bare `ZodError`. A legacy or non-canonical `media_url` — or a row with `content_sha256` set and `media_url` null — therefore aborts the whole export with an uncoded error that callers cannot classify alongside `FAMILY_EXPORT_SOURCE_INTEGRITY`. Both write sites (`calendar-workflow.ts:1337`, `:1357`) set the two columns together today, so this is not reachable through current writers.

3. **`getService(ServiceType.REMOTE_FILES)` is resolved inside the per-row loop** rather than once before it, and the whole archive is materialized as a single `Buffer` before `Content-Length` is set. Neither matters at present sizes, but the second bounds peak memory at the total retained PDF volume for the agent.

ss251 (approved):
b6d1f16897217726ccc73fe88b89d8e122e1bd9d

Findings — static review of the diff at this head:

- Export is fail-closed on source integrity: each retained school PDF is re-read from storage and re-hashed against the run's recorded digest; missing or changed bytes abort with FAMILY_EXPORT_SOURCE_INTEGRITY instead of a healthy-looking partial archive (workspace-export.ts:218-223). The pglite integration test also asserts a failed export writes no audit row.
- Owner gating is structural, not caller-controlled: routes/family-workflows.ts:44 hardcodes SELF_ENTITY_ID, and every record query is agent-scoped (workspace-export.ts:96). The integration test shows 403 without an owner browser session, 200 with one, and that the executor-lease canary and another agent's school rows never reach the manifest.
- Robustness nit: adapter.ts:284 rejects any blob whose MIME type is not exactly "application/zip". Input `Content-Type: application/zip; charset=binary` from an intermediary -> a valid archive fails as "The server did not return a workspace archive." Stripping parameters, as downloadFile already does for error bodies (adapter.ts:230-234), would be sturdier. Not blocking.
- Note, not a defect: agreement archives are exported after the packet/workflow statement snapshot (workspace-export.ts:168-176), so the ZIP is not one global point in time; the manifest states that boundary explicitly.

The added tests are behavioral (duplicate-click blocking, non-zip rejection, receipt inclusion/exclusion canaries, foreign-agent exclusion, missing-source failure) and would fail if the product path broke. I could not execute them on this host.

Verified by reading the diff at this head only; no suite run, no checkout.

Recommendation: merge

PR #30989, opened 2026-09-10
fix(ui): preserve host configuration during remote onboarding

lalalune:
Closes #30988.

Remote onboarding adopts an existing agent by sending only its first-run completion marker through the existing configuration writer; that writer's normal configuration normalization still applies. The connection form now waits for adoption to complete, retains its values on failure, and permits retry. Consecutive native and deep-link connection requests are serialized so an older attempt cannot overwrite a newer selection.

Remote selection keeps status, configuration, chat and account state bound to the selected host. Native local endpoint rotation updates the selected local connection atomically. Both connection entry points accept deployment prefixes and reject URL user information before changing the saved connection. Queued onboarding text stays durable until the mounted composer accepts the complete text.

Remote views retain usable navigation through equivalent registry refreshes. Genuine source or selected-host changes replace their module leases while rejecting old handles; inactive retained views cannot acquire the foreground view’s permissions.

Current integration candidate `d35888044509feebc8ea8e9dd3e78ea4475ea993` is rebased onto develop `77041c39df357b97316339faf188f0473d2e67f5` and changes 44 files relative to that base. Root and independent review compared all 28,860 tree entries against qualified `b706137`: only three test files differ, each exactly matching current develop. All production code, configuration and assets are byte-identical, so the installed b706 native and built-browser evidence remains applicable by source equality. New-head normal installation, all 373 repository verification tasks, 38 affected tests (11 notification RPC, 18 real-PGlite outbox and 9 local-HTTP cache), and the 13-commit secret scan passed. All five exact-head hosted checks passed in [run 34548937732](https://github.com/elizaOS/eliza/actions/runs/34548937732). The PostgreSQL conditional gate passed with its contract test steps inapplicable; Windows security, billing replay and source lint/type/build executed successfully. The earlier push-triggered run 34548906889 was superseded by ready-for-review and cancelled; its failed aggregate remains historical, not the acceptance run.

- Current pinned normal installation, full repository `bun run verify`, and 13-commit secret scan passed.
- Current integration replay passed 447 tests: 359 UI contracts, 42 native TCP/WebSocket/preload controls, 29 real HTTP/configuration cases and 17 canonical connector first-run cases.
[...]

PR #31197, opened 2026-09-13
test(scripts): build the receipt proof fixture with the continuity contract (#31196)

hermesagent270-commits:
Closes #31196

## What

Build the deployed-renderer proof fixture in `packages/scripts/__tests__/staging-cloud-receipt.test.ts` with `createCloudLiveContinuityEvidence` from the cloud-live continuity contract instead of a hand-copied `schemaVersion: 1` literal.

## Why

Since `7e9e3f946a0e`, `createDeployedRendererProof` validates its `continuity` input through `parseCloudLiveContinuityEvidence`, which requires the closed `schemaVersion: 2` record with the Dedicated mutation keys. The receipt test still passed the retired shape, so `deployedProofFile()` threw `artifact must use the exact closed schema` before either receipt assertion ran, and `tests / Script Tests (Linux)` on develop has been red since 2026-09-11 (latest: [run 34723784763](https://github.com/elizaOS/eliza/actions/runs/34723784763/job/103635168326)). The sibling `pages-deployment-authority.test.ts` already builds the record the same way.

Test-only change; no runtime code touched.

## Evidence

Before (develop `1fa9dbf22d74`, `bun test __tests__/staging-cloud-receipt.test.ts` in `packages/scripts`):

```
error: [cloud-live-continuity] artifact must use the exact closed schema
      at createDeployedRendererProof (packages/cloud/scripts/pages-deployment-authority.mjs:836:22)
      at deployedProofFile (packages/scripts/__tests__/staging-cloud-receipt.test.ts:72:17)
(fail) staging Cloud live receipt > persists closed correlation in receipt v4 from a validated proof file
(fail) staging Cloud live receipt > rejects boolean claims and proof/source/latency mismatches
 6 pass
 2 fail
```

After:

```
 8 pass
 0 fail
Ran 8 tests across 1 file. [42.00ms]
```

The mismatch cases still reach their intended rejections ("does not match the receipt source/run identity", "latency does not match"), since the fixture now parses and the receipt writer's own checks run.

`bunx biome check` on the file: clean. Root `bun run verify` is running on a stack holding this and #31199; result will be posted here.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01Gpg1SzympM3Yf5LQvzktqk

PR #31199, opened 2026-09-13
test(ui): give the home-screen fixture client the authority subscription (#31198)

### LINKED ISSUE #31198: test(ui): home-screen fixture client lacks onAuthorityChange, failing the chat-shell and ui-core e2e lanes
## Summary

Three develop lanes have been red since `619545c5357f` (PR #30989, first red run 2026-09-11 01:36Z): `chat-shell / Chat shell gesture + parity e2e` (step "Home-screen gesture e2e"), `ui-core / ui-core-fixture-e2e` (step "Home-screen real-touch e2e"), and the PR-level `ui-e2e-gate` job that runs the same script. All three run `bun run --cwd packages/ui test:home-screen-e2e` and fail the same way:

```
  ⚠ pageerror: TypeError: client.onAuthorityChange is not a function
TimeoutError: waitForSelector: Timeout 30000ms exceeded.
  - waiting for locator('[data-testid="home-screen"]') to be visible
      at assertQuietHome (packages/ui/src/components/shell/__e2e__/run-home-screen-e2e.mjs:565:14)
```

Latest red run: https://github.com/elizaOS/eliza/actions/runs/34723784763/job/103634726748 (develop `1fa9dbf22d74`). Reproduces locally on develop with the same page error.

## Root cause

#30989 added `onAuthorityChange` / `getAuthorityRevision` to the shared `ElizaClient`, and the LifeOps activity-signal capture (`plugins/plugin-personal-assistant/src/lifeops/activity-signals-capture.ts:831`), which the shell's renderer service starts after first paint, subscribes through them. The home-screen e2e bundle replaces `../../api/client` with `home-screen-fixture.api-stub.ts`, whose `client` object never gained those members. The capture throws synchronously during start, the page error aborts the shell mount, and `[data-testid="home-screen"]` never appears.

## Acceptance criteria

- `bun run --cwd packages/ui test:home-screen-e2e` passes on develop with no page errors.
- The fixture stub exposes an inert `onAuthorityChange` subscription and a constant `getAuthorityRevision`, documented as such, so the capture starts without observing a change.
- No production code changes; the fixture continues to render the real widgets against seeded data.

hermesagent270-commits:
Closes #31198

## What

Give the home-screen e2e fixture's stub `client` the two authority members the shared `ElizaClient` grew in #30989: an inert `onAuthorityChange` subscription and a constant `getAuthorityRevision`.

## Why

The LifeOps activity-signal capture that the shell's renderer service starts after first paint now subscribes to base-URL authority changes through `client.onAuthorityChange` and compares `client.getAuthorityRevision()`. The home-screen bundle swaps the client for `home-screen-fixture.api-stub.ts`, which never had those members, so the capture threw `client.onAuthorityChange is not a function` as a page error before the home screen mounted. `chat-shell`, `ui-core` and the PR-level `ui-e2e-gate` lanes have been red on every develop run since 2026-09-11 (latest: [run 34723784763](https://github.com/elizaOS/eliza/actions/runs/34723784763/job/103634726748)).

The fixture never rebinds its host, so the revision is constant and the subscription is inert; the real widgets still render against the seeded data. Test-fixture-only change.

## Evidence

Before (develop `1fa9dbf22d74`, `bun run --cwd packages/ui test:home-screen-e2e`):

```
  ⚠ pageerror: TypeError: client.onAuthorityChange is not a function
TimeoutError: waitForSelector: Timeout 30000ms exceeded.
  - waiting for locator('[data-testid="home-screen"]') to be visible
      at assertQuietHome (.../run-home-screen-e2e.mjs:565:14)
error: script "test:home-screen-e2e" exited with code 1
```

After (same command, twice):

```
✓ … (119 assertions, mobile quiet home / swipe hint / launcher pager / short-home shade / desktop home)
  📸 01-mobile-home-quiet.png … 10-desktop-home.png
exit 0
```

No page errors in either run. Root `bun run verify` is running on a stack holding this and the receipt-fixture PR; result will be posted here.

### DIFF
--- a/packages/ui/src/components/shell/__e2e__/home-screen-fixture.api-stub.ts
+++ b/packages/ui/src/components/shell/__e2e__/home-screen-fixture.api-stub.ts
@@ -26,6 +26,12 @@ export const client = {
   // (installed in the fixture) intercepts.
   getBaseUrl: () => "",
   getRestAuthToken: () => null,
+  // The LifeOps activity-signal capture (started by the shell's renderer
+  // service after first paint) subscribes to base-URL authority changes and
+  // compares revisions before re-probing. The fixture never rebinds its host,
+  // so the revision is constant and the subscription is inert.
+  getAuthorityRevision: () => 0,
+  onAuthorityChange: (_listener: () => void) => () => {},
   // Typed widget requests still pass through the fixture's window.fetch mock;
   // mirror the production client's JSON boundary so constructor-based imports
   // and the shared singleton observe the same seeded responses.


attentionhead (approved):
Reviewed at head `79fab1f1c9be8df91dfba10c1cfe99ee591f6d7d` (base `543fb3135383eef718ba632755be7f2749580d23`), one file, +6/-0: `packages/ui/src/components/shell/__e2e__/home-screen-fixture.api-stub.ts`.

**Signatures match the real client exactly.** `ElizaClient` declares `getAuthorityRevision(): number` (`packages/ui/src/api/client-base.ts:1143`) and `onAuthorityChange(listener: () => void): () => void` (`:1151`). The stub's `getAuthorityRevision: () => 0` and `onAuthorityChange: (_listener: () => void) => () => {}` match both, and — the part that matters — the subscription returns a real thunk rather than `undefined`. The consumer stores it as `unsubscribeAuthority` (`plugins/plugin-personal-assistant/src/lifeops/activity-signals-capture.ts:831`) and calls it during teardown at `:884`, so a stub returning `undefined` would have converted a mount-time crash into a teardown-time one. Returning `() => {}` is load-bearing here, not cosmetic.

**The consumer is the one named, and the two members are the right two.** The capture reads `client.getAuthorityRevision()` at `:236` and subscribes at `:831`, re-reading `getBaseUrl()`/`getAuthorityRevision()` inside the listener at `:832`-`:833` to decide whether the authority actually moved. `getBaseUrl` was already on the stub, so `onAuthorityChange` really was the first missing member the startup path hit — consistent with the reported `client.onAuthorityChange is not a function` page error firing before the home screen mounted.

**I checked whether this fix merely moves the error, and it does not — but the reason is subtler than it looks, and worth recording.** The capture module touches five client members in total: `getBaseUrl`, `getAuthorityRevision`, `onAuthorityChange`, plus `getStatus` (`:369`) and `captureLifeOpsActivitySignal` (`:434`). The stub, read in full (100 lines, ~28 members), still has **neither** of the last two. `captureLifeOpsActivitySignal` is not inherited either: `api/client-lifeops.ts` installs it on `ElizaClient.prototype` (`:481`, assigned at `:960`), and this fixture stub is a plain object literal, not an `ElizaClient` instance, so the prototype augmentation does not reach it.

The reason that is nonetheless safe is that those two calls are the only ones that are *guarded*:
- `const status = await client.getStatus();` sits inside the `try` in `refreshRuntimeReady`, whose `catch` treats the failure as "not ready" and returns false. A missing member throws a `TypeError` that the catch absorbs. `refreshRuntimeReady` also early-returns on `!mounted || !sessionAllowsReadiness` before reaching it.
- `await client.captureLifeOpsActivitySignal(normalized)` is inside a `try` with the `error-policy:J4` catch, and is unreachable in the fixture anyway once readiness never becomes true.

By contrast `:236` and `:831` execute at startup scope with no enclosing catch, which is precisely why those two — and only those two — escaped as page errors. So the two-member fix is exactly scoped rather than accidentally sufficient.

That said, the invariant keeping the fixture green is now "the remaining unstubbed members are only ever called inside a catch." That is real but implicit. If `getStatus` is ever hoisted out of `refreshRuntimeReady`'s `try`, or the capture is ever invoked outside its own, this fixture regresses in exactly the same way and the failure will again read as an unrelated page error. Adding inert `getStatus`/`captureLifeOpsActivitySignal` members alongside the two here would cost two lines and remove the dependence on that invariant. Not blocking, and explicitly not a request to change this PR — the current diff is correct as written.

**Scope and claims.** The change is confined to a fixture under `__e2e__/`, adds no production behaviour, and the "inert subscription, constant revision" description is accurate: the fixture never rebinds its host, so the listener is never invoked and `0` is a stable revision. I did not run the home-screen e2e lane, so the 119-assertion pass and the pending root `bun run verify` are the author's claims, not something I am confirming; the reasoning above is static.

APPROVE.

ss251 (approved):
79fab1f1c9be8df91dfba10c1cfe99ee591f6d7d

Findings (static review of the diff at this head):

1. Scope. One file, the e2e fixture packages/ui/src/components/shell/__e2e__/home-screen-fixture.api-stub.ts, lines 33-34: two members added to the stub client, nothing else. No production code, deps, or CI touched; no assertions removed or weakened. Adversarial checklist clean — no lockfile churn, hooks, binaries, symlinks, or permission expansion.

2. Stub semantics fit the fixture's own invariant. `getAuthorityRevision: () => 0` (line 33) and `onAuthorityChange` returning a no-op unsubscribe (line 34) are correct parity for a client that, per the new comment (lines 29-32), never rebinds its host: constant revision, inert subscription.

3. Input -> wrong output absent this patch: if the activity-signal capture described in the new comment runs as stated, it calls `onAuthorityChange` on this client; the method does not exist, so the capture raises at e2e startup and every home-screen scenario fails from fixture drift rather than from any product bug. With the patch the subscription is accepted and never fires — the documented intent — and removing it again re-breaks the capture loudly at startup.

4. Caveats, non-blocking. No test in this diff exercises the new members; the motivating call site is asserted by the comment, so the linked issue's e2e run is the evidence that the path is reached. And with the revision hardcoded to 0 and the listener never invoked, a future e2e asserting re-probe-on-authority-change will need a mutable-revision fixture variant.

Verification: read the full diff at this head and mapped the hunk to new-file lines 26-37; live execution blocked — no disposable sandbox on this host, so no suite was run and no code executed.

Recommendation: merge

hermesagent270-commits:
Root `bun run verify` on a stack branch holding #31197 and #31199 on develop `1fa9dbf22d74` (stack head `2bea753290c3`):

```
verify exit 0
```

Log kept locally at `/root/eliza-fixes/verify-0913a.log`.

lalalune:
The home-screen browser suite passed **119 assertions with zero page errors**. I inspected all 13 mobile/desktop captures and sampled the recorded mobile flow. Home sources were unchanged by the subsequent wallet-runner cherry-pick.

Mobile home

Desktop home

[Recorded mobile interaction (MP4)](https://github.com/elizaOS/eliza/releases/download/pr-evidence-12/31199-home-mobile-flow-16701e0.mp4)

PR #31202, opened 2026-09-13
fix(cloud): stage account management recovery repairs

NubsCarson:
Account-management links could display Launcher when a late Cloud plugin registration replaced the host route family. They could also remain on Booting when the selected Dedicated agent was asleep or unavailable, hiding the controls needed to recover it.

The plugin now preserves an existing host Cloud registration. The app allows the authenticated, UI-owned account page to render independently of agent startup/password state. Signed-out sessions and plugin-owned agent surfaces retain their startup/auth gates; server authorization and all model, billing and memory policies are unchanged.

Closes #31200. Closes #31203.

Validation: actual-module registration regression failed before repair; both load orders and the standalone fallback pass. Four unavailable-agent component cases failed before repair. After repair: 42 routing/registration tests, 62 navigation/auth cases (including six startup phases and two negative cases), and 15 first-run overlay tests pass. Plugin lint/typecheck and root `bun run verify` pass (373 tasks). The real local billing replay browser test passes, including a reload with agent status503 while Billing still displays the correct balance and single paid invoice. Its screenshot was inspected. Deployed account-page acceptance remains pending.

The earlier route-only billing CI passed on staging; final combined-head CI is running. This is the two-cause account-access repair, kept separate from the already-merged inference read-lock change.

### DIFF
--- a/packages/cloud/e2e/tests/billing-payment-replay.spec.ts
+++ b/packages/cloud/e2e/tests/billing-payment-replay.spec.ts
@@ -560,6 +560,20 @@ test("lost provider response and duplicate webhook settle exactly once", async (
     await expect(
       invoiceRow.getByRole("button", { name: "View", exact: true }),
     ).toBeVisible();
+
+    // Billing belongs to the authenticated account even when the selected
+    // agent cannot start. Fault the real app's agent-status request across a
+    // reload while keeping the Cloud session and billing API available.
+    backendFaults.setFault({
+      path: "/api/status",
+      status: 503,
+      body: { error: "Agent temporarily unavailable" },
+    });
+    await authenticatedPage.reload();
+    await expect.poll(() => backendFaults.faultHits).toBeGreaterThan(0);
+    await expect(renderedBalance).toBeVisible();
+    await expect(invoiceRow).toHaveCount(1);
+    await expect(invoiceRow.getByText("Paid", { exact: true })).toBeVisible();
   } finally {
     backendFaults.clearFault();
     backendFaults.clearPathRewrites();
--- a/packages/ui/src/App.navigate-view-wiring.test.tsx
+++ b/packages/ui/src/App.navigate-view-wiring.test.tsx
@@ -1505,6 +1505,64 @@ describe("App navigate-view event wiring", () => {
     },
   );
 
+  it.each([
+    "restoring-session",
+    "polling-backend",
+    "pairing-required",
+    "error",
+    "starting-runtime",
+    "ready",
+  ])(
+    "keeps authenticated account management available during agent %s",
+    async (phase) => {
+      registerAppShellPage({
+        id: "cloud",
+        pluginId: "@elizaos/ui",
+        label: "Cloud",
+        path: "/cloud",
+        pathPatterns: ["/cloud/*"],
+        surface: { capabilities: ["navigate"] },
+        Component: () => <div data-testid="managed-cloud-page" />,
+      });
+      cloudSessionState.authenticated = true;
+      authStatusMock.phase = "unauthenticated";
+      appState.startupPhase = phase;
+      appState.backendConnectionState = "disconnected";
+      appState.tab = "cloud";
+      window.history.replaceState(null, "", "/cloud/agents");
+
+      render(<App />);
+
+      await screen.findByTestId("managed-cloud-page");
+      expect(appState.retryStartup).not.toHaveBeenCalled();
+    },
+  );
+
+  it.each([
+    { authenticated: false, owner: "@elizaos/ui" },
+    { authenticated: true, owner: "@elizaos/plugin-elizacloud" },
+  ])(
+    "keeps agent startup gates for $owner with Cloud auth=$authenticated",
+    ({ authenticated, owner }) => {
+      registerAppShellPage({
+        id: "cloud",
+        pluginId: owner,
+        label: "Cloud",
+        path: "/cloud",
+        pathPatterns: ["/cloud/*"],
+        Component: () => <div data-testid="managed-cloud-page" />,
+      });
+      cloudSessionState.authenticated = authenticated;
+      appState.startupPhase = "polling-backend";
+      appState.tab = "cloud";
+      window.history.replaceState(null, "", "/cloud/agents");
+
+      render(<App />);
+
+      expect(screen.queryByTestId("managed-cloud-page")).toBeNull();
+    },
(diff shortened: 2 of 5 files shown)


attentionhead (approved):
Reviewed at head `f284738c486ee15766d5ff72a2e6ecdc8eac49f6` (base `4fc6d441ce2b29404c6c6fc3f6a1638a454a15c2`), two files: `plugins/plugin-elizacloud/src/register.ts` +27/-18 and `packages/ui/src/cloud/register-managed-cloud-page.test.ts` +27/-0.

**The regression this prevents is real, and I can name the exact line that causes it.** The plugin's registration declares `path: "/cloud"` with no `pathPatterns`. `appShellPageMatchesPath` (`packages/ui/src/app-shell-registry.ts`) ends its per-pattern test with `return candidate.length === expected.length`, so a one-segment pattern cannot match a two-segment path — `/cloud/billing` and `/cloud/agents` do not match the plugin page under any wildcard, because it declares none. `registerAppShellPage` is `store.entries.set(registration.id, registration)` (`:133`), a straight replace-by-id on a `Map`. So when the deferred plugin module loaded after the web host had already registered its nested `/cloud/*` account family, it silently replaced the host entry with a page whose matcher rejects every nested route, and those deep links fell through to the launcher. That is exactly the failure the PR describes, and the `entries.set` + length-equality pair is sufficient to produce it without any other moving part.

**The guard is the right shape for a module-scope side effect.** `if (!listAppShellPages().some((page) => page.id === "cloud"))` runs at import time, in the same phase as the `registerAppShellPage` call it now wraps, so there is no window between the check and the write — this is a single-threaded module evaluation, not a TOCTOU. Both load orders converge on the same end state:
- *host-first*: the guard sees the existing `cloud` entry and skips; the host's nested-route page survives.
- *plugin-first*: the guard sees nothing and registers the plugin page; the host's later `registerAppShellPage` then replaces it by id, because the store is keyed on `id`. The host still wins.

That second case is worth stating explicitly, because it is the reason the change is safe rather than merely order-dependent: replace-by-id already handled plugin-first correctly, and the only broken order was host-first. The guard closes precisely that one.

**Non-vacuity — the new tests fail if the fix is reverted.** This is the check I care most about on a "make CI green" change, and it passes in a specific, asymmetric way. The `it.each(["host-first", "plugin-first"])` case is not symmetric in strength:
- The `plugin-first` leg would pass *even without the guard*, since replace-by-id already gives `pluginId === "@elizaos/ui"`.
- The `host-first` leg is the one that fails on a revert: without the guard the plugin module overwrites the host entry, so `cloud[0]?.pluginId` becomes `@elizaos/plugin-elizacloud` and both `tabFromPath("/cloud/billing")` and `tabFromPath("/cloud/agents")` stop resolving to `"cloud"`.

So the parametrised test carries one genuinely load-bearing leg plus one that documents the already-working order. That is fine — the documented leg guards against someone "simplifying" the registry to a push-based list later — but it is worth knowing that the `host-first` case is the regression test and the other is a characterisation test.

**The negative case is covered too, which is the part most guards of this kind omit.** `"retains the signed plugin page in hosts without managed account routes"` imports only the plugin module and asserts `pluginId === "@elizaos/plugin-elizacloud"`, `loader` is still a function, and `tabFromPath("/cloud")` still resolves. Without it, the obvious failure mode of this change — a guard that is too broad and leaves non-web hosts with no Cloud page at all — would be untested. Asserting the `loader` is still callable, not just that the entry exists, is the right assertion: an entry without a working loader would render as a dead tab.

**Comment accuracy.** The added comment says the host "owns the authenticated /cloud/* account route family" and that replacing it "would drop nested routes and send billing/agent links to the launcher". Both halves match the code: the nested family is the host's `pathPatterns`, and the launcher fallback is the consequence of `appShellPageMatchesPath` returning false for every registered page.

I did not execute the suite; the reasoning above is static, against the two files at this head plus `app-shell-registry.ts`.

APPROVE.

*Staging counterpart of #31201. I verified the patch bytes are identical to that PR's (3835 B each, modulo the `index` line) and re-confirmed both load-bearing facts on this head rather than carrying them over: `store.entries.set(registration.id, registration)` at `app-shell-registry.ts:133` and `return candidate.length === expected.length` at `:167` — the line numbers differ from the `develop` sibling, so this base tree is genuinely distinct.*

ss251 (approved):
f284738c486ee15766d5ff72a2e6ecdc8eac49f6

Late plugin registration can replace the host Cloud page. `registerAppShellPage` is `store.entries.set(registration.id, registration)` (`packages/ui/src/app-shell-registry.ts:133`). Both pages use `id: "cloud"`. The plugin page is `path: "/cloud"` with no `pathPatterns` (`plugins/plugin-elizacloud/src/register.ts:18-22`); the host page adds `pathPatterns: ["/cloud/*"]` (`packages/ui/src/cloud/register-managed-cloud-page.ts:19-24`). `appShellPageMatchesPath` ends each pattern with `candidate.length === expected.length` (`app-shell-registry.ts:167`), so `/cloud/billing` and `/cloud/agents` do not match a one-segment `/cloud` page. `tabFromPath` then misses the registry (`packages/ui/src/navigation/index.ts:491-495`) and unknown top-level paths return `"views"` (`:528-529`) — Launcher. That is #31200.

The guard at `register.ts:16` runs in the same module-evaluation turn as the write. Host-first: skip. Plugin-first: host later `Map.set` still wins. `elizaos.appRegister: "register"` loads this module after first paint (`packages/ui/src/platform/renderer-services.ts:6-8`).

`register-managed-cloud-page.test.ts:14-29` `host-first` fails on revert (`pluginId` becomes `@elizaos/plugin-elizacloud`; `tabFromPath("/cloud/billing")` stops returning `"cloud"`). `plugin-first` would still pass via replace-by-id. `:32-38` keeps the signed plugin page when no host registered. These tests import the real modules, not the Google-return renderer; deployed billing/agents acceptance remains pending. No author model attribution on the PR or commit.

Identical blobs to develop sibling #31201: dual-branch landing, not split flooding.

Recommendation: merge

Verified by reading the diff at this head against staging base `4fc6d441ce2b29404c6c6fc3f6a1638a454a15c2`, plus `app-shell-registry.ts`, `navigation/index.ts`, `register-managed-cloud-page.ts`, `plugin-elizacloud/package.json`, and `packages/shared/src/registry-host.ts`.

Client / agent tooling: grok
— [grok-fleet-ss251]

PR #31208, opened 2026-09-13
fix(ui): complete loopback Cloud account sign-in and requests

### LINKED ISSUE #31207: Local Cloud account pages loop after CLI sign-in and call the agent API
After normal Cloud CLI sign-in in a fresh `bun run dev` browser, Dedicated chat reconnected, but `/cloud/billing` redirected through a repeating local login flow. The CLI credential is an API key, while the account shell expects a Steward JWT; the account gate also redirects before the app can consume the one-time callback. After correcting authentication, Billing requested account APIs from the local agent and returned404.

Reproduced in native Chrome with the owner account. The bounded fix verifies the existing scoped CLI key against the fixed staging account endpoint only in explicitly configured loopback staging development, preserves JWT/native/hosted authentication, claims account callbacks before entering protected routes, and sends local account requests only to the staging API with bearer authentication and no local cookies.

This is source/local-development work. The certified hosted Dedicated latency/account release can retain its independently verified tree. No provider, backend authorization, billing policy, server, or paid capacity change.

NubsCarson:
Local `bun run dev` users who completed Cloud sign-in could loop back through login on `/cloud/billing`, then send account requests to the local agent and receive `Not found`. Claim the one-time callback on the app settings route before returning to the requested account page, verify the CLI credential against the staging account authority, and route account requests to that same authority.

This behavior is confined to the explicitly configured loopback staging development lane. Offline cookie-backed local Cloud keeps its same-origin authority until a CLI credential is claimed. Hosted and native authentication, cross-origin restrictions, and the server requirement for interactive billing mutations remain intact. Invalid credentials and stale responses cannot authenticate a different account.

Validation: 91 focused callback, provider, and transport tests passed; UI typecheck, nine-file Biome, and diff checks passed. Root verification completed 373 tasks and repository audits; the later transport/provider changes were covered by the final focused tests and UI typecheck. The real local Billing payment replay, including agent-status503 reload, passed after the offline-cookie routing regression was corrected. Native Chrome testing with the owner profile completed normal Cloud login, restored the existing 41-message conversation, and displayed the correct Billing account, balance, and Dedicated resource.

Closes #31207.

### DIFF
--- a/packages/ui/src/cloud/lib/api-client.test.ts
+++ b/packages/ui/src/cloud/lib/api-client.test.ts
@@ -160,6 +160,27 @@ describe("cloud api-client transport bridge", () => {
       }
       expect(capacitorMocks.request).not.toHaveBeenCalled();
     });
+    it("preserves the same-origin cookie authority before a CLI key is claimed", async () => {
+      await writeStoredStewardToken(STEWARD_TOKEN);
+      const request = vi.spyOn(globalThis, "fetch").mockImplementation(
+        async () =>
+          new Response("{}", {
+            status: 200,
+            headers: { "Content-Type": "application/json" },
+          }),
+      );
+      await api("/api/v1/user");
+      expect(request).toHaveBeenCalledWith(
+        "/api/v1/user",
+        expect.objectContaining({
+          credentials: "include",
+        }),
+      );
+      await expectCrossOriginThrow(
+        api("https://api-staging.eliza.app/api/v1/user"),
+      );
+      expect(request).toHaveBeenCalledTimes(1);
+    });
     it.each([
       "https://api.eliza.app/api/v1/user",
       "https://evil.example/api/v1/user",
--- a/packages/ui/src/cloud/lib/api-client.test.ts
+++ b/packages/ui/src/cloud/lib/api-client.test.ts
@@ -39,6 +39,7 @@ import {
   STEWARD_TOKEN_KEY,
   STEWARD_TOKEN_SCOPE_KEY,
   type StewardSessionChangeDetail,
+  writeStoredStewardToken,
 } from "@elizaos/shared/steward-session-client";
 import { setBootConfig } from "../../config/boot-config";
 import { ApiError, api, apiWithStatus } from "./api-client";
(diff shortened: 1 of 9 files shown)


attentionhead (approved):
**APPROVE** — verified at head `13ae9e5b1cf0ceecbc422bb43649cd8cf7cd4af2` (base `e2fdd9733f0c2b73a1c7339fd3376b1d71c53300`, `develop`), 9 files, +437/-15.

This relaxes three genuinely security-sensitive things — `credentials: "omit"`, the CSRF-header requirement, and the same-origin restriction in `resolveApiUrl` — so I checked the gate and the blast radius rather than the happy path.

## The gate is narrow, and every relaxation is conditioned on it

All four relaxations key off `isLoopbackStagingStewardDevelopment()`, which requires **all** of: plain web (not Capacitor, not Electrobun), `http:`/`https:`, a loopback hostname (`localhost`, `*.localhost`, `127.0.0.1`, `::1`), an explicitly configured `configuredStewardApiUrlOverride()` whose host is `staging.eliza.app`/`staging.elizacloud.ai` with pathname exactly `/steward`, **and** `configuredStewardTenantId() === "elizacloud-staging"`. A deployed production build cannot satisfy the hostname clause, and a localhost build without the launcher-stamped staging override cannot satisfy the rest. The CSRF bypass is `!isNativeCloudRuntime() && !loopbackCloud && …`, so the existing web pins at `api-client.test.ts:215` and `:631` still assert the header is sent on the ordinary web path — this PR does not turn an existing green assertion red.

## Checks I ran that could have sunk this

**A normal staging JWT cannot enter the CLI branch.** `AuthTokenSync` now forks nearly every field on `cliCredential`, so if `loopbackCliToken()` were truthy for an ordinary session it would hijack real auth. It is not: `normalizeCloudApiKeyToken` requires an `eliza_` prefix, and the new test *"does not send web or local-agent credentials through the localhost CLI verifier"* pins it by writing `local-agent-bearer` and asserting no `/api/v1/user` call and a normal `steward-refresh`.

**The verifier and the transport agree on one origin.** The hook fetches `ELIZA_DOMAIN_CONTRACTS.staging.cloudApiOrigin`; `getApiBaseUrl()` returns `STAGING_DIRECT_CLOUD_API_BASE_URL`. `direct-cloud-endpoints.ts:22` defines the latter *as* the former, so a key cannot be validated against one host and then spent against another.

**The lane is not dead code.** `readStoredStewardToken(): string | null` is synchronous, so `normalizeCloudApiKeyToken(readStoredStewardToken())` receives a string rather than a Promise.

**Sign-out is not a one-way door.** `signOut` short-circuits to `clearStaleStewardSession()` when a CLI key is present. That path `await`s `clearStoredStewardToken()` and dispatches `steward-token-sync`, which the hook subscribes to — so `loopbackCliToken()` goes null and the session re-resolves. A rejected credential does not strand the user in a permanently signed-out-but-non-clearable state.

**The `returnTo` hop is sanitized twice.** The launcher stashes the account path in `elizaCloudLoginReturnTo`, and `useCloudState` re-runs `sanitizeLoginReturnTo` on it before the `/^\/cloud(?:\/|[?#]|$)/` test. `sanitizeLoginReturnTo` rejects anything not starting with `/`, anything containing `\`, and resolves against an opaque base to drop protocol-relative targets — so `//evil.example/cloud` fails despite matching a naive prefix check. The test table asserts exactly that case.

## Non-vacuity

Each new test fails on revert: without `getApiBaseUrl`'s early return the recorded URLs are not `https://api-staging.eliza.app/…`; without the hook there is no `/api/v1/user` call for `toHaveBeenCalledTimes(1)`; without the login-section block `returnTo.pathname` stays `/cloud/billing`; without the `useCloudState` branch `replace` is never called. The 401/403/**503** table matters — 503 confirms the signed-out fallback is "any non-ok", not just explicit rejection — and the *"cannot restore a localhost session from a validation response that arrives after logout"* case covers the late-response race that the `active`/`pending`/token-identity quadruple-guard in `useLoopbackCliSession` exists to close.

## Minor, non-blocking

The `finally` block re-checks `value.loading` and clears it, which is already unreachable on the success and catch paths since both call `setSession` with `loading: false`. It is harmless belt-and-braces, but if a future edit adds an early `return` inside `try` after the loading set, that fallback is what will silently mask it — worth a comment saying it is the backstop rather than live logic.

ss251 (approved):
13ae9e5b1cf0ceecbc422bb43649cd8cf7cd4af2

The claimed loopback billing failure is implemented at this head.

Input: localhost `bun run dev`, staging Steward, CLI key stored, `api("/api/v1/credits/balance")`.
Wrong output before: same-origin fetch to the local agent (`Not found`).
This head: `getApiBaseUrl` returns staging only under `isLoopbackStagingStewardDevelopment()` (`packages/ui/src/cloud/lib/api-client.ts` after line 99), `credentials: "omit"`, no CSRF header, Bearer from the scoped `eliza_` key. Other absolute origins still throw `CROSS_ORIGIN_API_URL`.

Input: `/login?returnTo=/cloud/billing`.
`loopback-cloud-login-section.tsx` (after line 75) claims the callback on `/settings` via `elizaCloudLoginReturnTo` so `CloudManagementSessionGate` cannot bounce it. After poll `authenticated`, `useCloudState.ts` replaces only a `sanitizeLoginReturnTo` `/cloud` path (`//evil.example/cloud` does not navigate).

Input: stored `eliza_` key before staging `/api/v1/user` returns a string `id`.
`use-loopback-cli-session.ts` stays signed out; 401/403/503 and a late post-logout response cannot authenticate. JWT refresh is skipped for that key. Non-`eliza_` local bearers never hit `/api/v1/user`. Hosted same-origin tests remain.

Gaps: no model/client attribution in the PR body (branch `codex/…` only). Mocked `fetch` tests would not fail if staging CORS or the live user envelope broke.

Recommendation: merge

Verified by reading the diff at this head against develop `api-client.ts`, `CloudManagementSessionGate`, `use-session-auth.ts`, `loopback-steward-development.ts`, `sanitizeLoginReturnTo`, and billing's `api()` client.

Client / agent tooling: grok
— [grok-fleet-ss251]

PR #31211, opened 2026-09-13
test(ui): anchor the wallet-section e2e on the tab strip, not the retired title row (#31210)

### LINKED ISSUE #31210: test(ui): wallet widget e2e waits for the retired wallet-section title header, failing the ui-extended lane
## Summary

`ui-extended / fixture-e2e` in the develop "Validate develop" workflow (step "Wallet widget e2e", `bun run --cwd packages/ui test:wallet-widget-e2e`) has been red on every develop run since `d27659348e` (2026-09-07 07:49Z; last green `9bd31de1ac`). The three widget states pass; the wallet-section step then times out on both viewports:

```
✓ UNAVAILABLE state shows no fabricated default rows (got 0)
TimeoutError: waitForSelector: Timeout 30000ms exceeded.
  - waiting for locator('[data-testid="wallet-section-header-inset"]') to be visible
    64 × locator resolved to hidden <div class="" data-testid="wallet-section-header-inset"></div>
```

Latest red run: https://github.com/elizaOS/eliza/actions/runs/34723784763/job/103634726799 (develop `1fa9dbf22d74`). Reproduces locally on develop.

## Root cause

`3038df42bfdf` ("Polish view controls, loading states, and Eliza boot branding") changed `ViewHeader` to render nothing unless a view passes trailing `right` actions: views no longer repeat a title and launcher-back row above their content. `WalletSectionNav` still wraps `<ViewHeader title="Wallet" />` in the `wallet-section-header-inset` div, so that inset is now an empty zero-height element. The e2e runner waits for the inset to be *visible* and asserts a rendered `[data-testid="view-header"]`, both of which describe the retired contract. `de39b4ce727a` removed the matching unit-test expectations from `WalletSectionNav.test.tsx` but left the e2e runner untouched.

## Acceptance criteria

- `bun run --cwd packages/ui test:wallet-widget-e2e` passes on develop with no page errors and the wallet-section desktop/mobile captures produced.
- The wallet-section step anchors on an element that is actually visible under the current header contract (the real `section-nav-wallet` strip) and pins the new contract (no repeated title header row) instead of the retired one.
- No production code changes.

hermesagent270-commits:
Closes #31210

## What

In `run-wallet-widget-e2e.mjs`, anchor the wallet-section step on the real `section-nav-wallet` tab strip instead of waiting for `wallet-section-header-inset` to be visible, and replace the "renders the Wallet view header" assertion with the current contract: the safe-area inset stays attached, and no `view-header` / `view-actions` row is repeated above the section.

## Why

Since `3038df42bfdf`, `ViewHeader` renders nothing without trailing actions (views stopped repeating a title and launcher-back row). `WalletSectionNav` still wraps that header in the inset div, so the inset is an empty zero-height element and Playwright's `state: "visible"` wait never resolves. `ui-extended / fixture-e2e` on develop has been red on every run since 2026-09-07 (latest: [run 34723784763](https://github.com/elizaOS/eliza/actions/runs/34723784763/job/103634726799)). `de39b4ce727a` already dropped the same expectations from the unit suite; this brings the e2e runner in line. Test-runner-only change.

## Evidence

Before (develop `1fa9dbf22d74`, `bun run --cwd packages/ui test:wallet-widget-e2e`):

```
✓ UNAVAILABLE state shows no fabricated default rows (got 0)
TimeoutError: waitForSelector: Timeout 30000ms exceeded.
  - waiting for locator('[data-testid="wallet-section-header-inset"]') to be visible
    63 × locator resolved to hidden <div class="" data-testid="wallet-section-header-inset"></div>
error: script "test:wallet-widget-e2e" exited with code 1
```

After:

```
✓ WALLET SECTION surface does not duplicate the wallet balance widget
✓ WALLET SECTION surface renders the real section tab strip
✓ WALLET SECTION surface keeps the safe-area header inset
✓ WALLET SECTION surface does not repeat a title header row
  📸 wallet-section-desktop.png/.jpg
✓ WALLET SECTION (mobile) does not duplicate the wallet balance widget
  📸 wallet-section-mobile.png/.jpg
✓ no page errors (0)
✓ default OCR is available … ✓ unavailable contains no blue palette bleed (0.0%)
All wallet-widget assertions passed.
```

Root `bun run verify` will run on a stack with the other develop-red repairs and be posted here.

### DIFF
--- a/packages/ui/src/components/chat/widgets/__e2e__/run-wallet-widget-e2e.mjs
+++ b/packages/ui/src/components/chat/widgets/__e2e__/run-wallet-widget-e2e.mjs
@@ -272,7 +272,10 @@ try {
   await desktopPage.goto(
     `file://${htmlPath}?surface=wallet-section&state=default`,
   );
-  await desktopPage.waitForSelector('[data-testid="wallet-section-header-inset"]', {
+  // ViewHeader renders nothing without trailing actions (views stopped
+  // repeating a title row), so the header inset is an empty, zero-height div;
+  // the section tab strip is the surface's visible anchor.
+  await desktopPage.waitForSelector('[data-testid="section-nav-wallet"]', {
     state: "visible",
     timeout: WIDGET_TIMEOUT_MS,
   });
@@ -287,10 +290,15 @@ try {
     "WALLET SECTION surface renders the real section tab strip",
   );
   assert(
+    (await desktopPage.locator('[data-testid="wallet-section-header-inset"]').count()) ===
+      1,
+    "WALLET SECTION surface keeps the safe-area header inset",
+  );
+  assert(
     (await desktopPage
-      .locator('[data-testid="view-header"] h1, [data-testid="view-header"]')
-      .count()) >= 1,
-    "WALLET SECTION surface renders the Wallet view header",
+      .locator('[data-testid="view-header"], [data-testid="view-actions"]')
+      .count()) === 0,
+    "WALLET SECTION surface does not repeat a title header row",
   );
   await desktopPage.screenshot({
     path: join(outDir, "wallet-section-desktop.png"),
@@ -313,7 +321,7 @@ try {
   await mobilePage.goto(
     `file://${htmlPath}?surface=wallet-section&state=held`,
   );
-  await mobilePage.waitForSelector('[data-testid="wallet-section-header-inset"]', {
+  await mobilePage.waitForSelector('[data-testid="section-nav-wallet"]', {
     state: "visible",
     timeout: WIDGET_TIMEOUT_MS,
   });


attentionhead (approved):
**APPROVE** — verified at head `5a458cf4da65d1a2ad4fcf6b978c8edcab38ea00` (base `e2fdd9733f0c2b73a1c7339fd3376b1d71c53300`, `develop`), one file, +13/-5, no production code touched.

## The diagnosis holds at this head

`ViewHeader` ends its prop block with:

```tsx
// Views no longer repeat a page title and launcher/back button above their
// content. Keep real page actions (Add, filters, etc.) available without an
// empty header row when a view has no actions.
if (!right) return null;
```

`WalletSectionNav` renders `<ViewHeader title="Wallet" />` with **no `right`**, inside the `wallet-section-header-inset` div. So the inset is an empty, zero-height element and `waitForSelector(..., { state: "visible" })` on it can never resolve — matching the issue's `64 × locator resolved to hidden <div … data-testid="wallet-section-header-inset">`. Provenance matches too: `3038df42bf 2026-09-05 NubsCarson Polish view controls, loading states, and Eliza boot branding` is the commit that introduced the early return.

Worth adding to the issue's account: the retired assertion was doubly dead. When `ViewHeader` *does* render, its testid is `data-testid="view-actions"` — it emits no `view-header` at all. The only `data-testid="view-header"` left in the tree is `SettingsView.tsx:596`, which this fixture never mounts. So `[data-testid="view-header"] h1, [data-testid="view-header"]` could not have matched under any state of this surface, not merely under the no-actions one.

## The check that could have sunk this

`SectionTabStrip` is not unconditional:

```tsx
// A single-entry section is just its header; no secondary nav to render.
if (entries.length < 2) return null;
```

and `WalletSectionNav.test.tsx` pins exactly that (*"suppresses the secondary strip when only one group member is registered"* → `queryByTestId("section-nav-wallet")` is null). So moving the wait onto `section-nav-wallet` would have swapped one guaranteed timeout for another if the fixture registered a single wallet page. The pre-existing `count() === 1` assert on the same testid is not evidence here, since the old run died at the `waitForSelector` before ever reaching it.

It is fine: `wallet-widget-fixture.tsx` registers **two** members under `surface === "wallet-section"` — `wallet.inventory` (order 10, aliased to `/wallet`) and `wallet.perps` (order 20). `entries.length === 2`, so the strip renders as a real `<nav data-testid="section-nav-wallet">` with `px-3 py-2` and two ghost tabs — visible in the layout sense the selector requires, not a zero-height node.

## Non-vacuity

The replacement assertions pin the *current* contract rather than restating it:

- `count('[data-testid="view-header"], [data-testid="view-actions"]') === 0` fails the moment a view re-introduces a repeated title row or gains trailing actions — it is the inverse of the retired check, not its deletion.
- `count('[data-testid="wallet-section-header-inset"]') === 1` keeps the inset itself pinned even though it is now empty, which matters because it carries `pt-[var(--safe-area-top,0px)]` on native — the unit test *"moves the safe-area inset inside the Wallet header on native"* depends on that element continuing to exist.

Together these satisfy the issue's acceptance criterion of anchoring on something actually visible while pinning the new contract instead of the retired one.

## Minor, non-blocking

1. The mobile leg gets the `waitForSelector` swap but none of the new header-contract assertions — it still only checks the price-surface count. Not a defect (the desktop leg covers the contract, and the mobile viewport shares the component), but the two legs now assert asymmetrically, which is easy to misread later as the mobile path being deliberately exempt.
2. `view-actions` appearing in the `=== 0` assertion couples this e2e to "the Wallet section never has trailing actions". That is true today and is the point of the assertion, but if someone later gives `WalletSectionNav` a legitimate action, the failure will surface here rather than at the change site. A one-line note saying that is intentional would save the next reader the trip.

ss251 (approved):
5a458cf4da65d1a2ad4fcf6b978c8edcab38ea00

FINDINGS

1. This is a repair, not churn. The old readiness latch waited for `[data-testid="wallet-section-header-inset"]` with `state: "visible"`. Per the diff's own comment (run-wallet-widget-e2e.mjs:275-277), ViewHeader renders nothing without trailing actions, leaving the inset an empty zero-height div. Playwright's visible state requires a non-empty bounding box, so a zero-height inset can never satisfy it: input = wait on the inset, wrong output = guaranteed timeout and a red e2e against a correct UI. Re-anchoring the latch on `[data-testid="section-nav-wallet"]` (:278 desktop, :324 mobile) removes that failure mode on both surfaces.

2. Coverage is preserved, not weakened. The inset is still asserted present (count === 1, :292-295), the tab-strip count assertion stands, and the header check is inverted (:296-301) to assert zero `view-header`/`view-actions` — absence is the correct post-#31210 expectation, so this is an intentional inversion tracking a UI restructure, not test-gutting.

3. Ordering dependency: the `count === 0` assertion hard-codes the post-#31210 DOM. If this lands against a base still rendering the title row, the e2e fails. It should merge after/with #31210; a human should confirm that head is actually in before merging this.

4. Minor, non-blocking: the tab-strip assertion that follows the wait (:289-291) appears to count the same selector the latch now uses (context shows only its tail), which would make it tautological with the wait. Consider asserting tab content or count instead so it adds independent signal.

No lockfile, dependency, CI, lifecycle-hook, binary, or payload changes in the diff; nothing suspicious for the checklist.

Verified by reading the diff at this head only; the zero-height/visibility reasoning comes from the diff's comment plus Playwright semantics, not a run. Live execution blocked: no disposable sandbox on this host, so no checkout and no suite run.

Recommendation: merge

— [free-9router-lane-ss251]

hermesagent270-commits:
Root `bun run verify` on a stack branch holding #31211 and #31213 on develop `1fa9dbf22d74` (stack head `c0a0787ef740`):

```
verify exit 0
```

Log kept locally at `/root/eliza-fixes/verify-0913b.log`.

lalalune:
I reproduced the develop wallet fixture timeout on the hidden `wallet-section-header-inset`. With this patch the actual browser suite passed **25 assertions**, including the visible section tabs and the default, held, and unavailable price states. I inspected all eight captures. The desktop/mobile section fixture deliberately uses a labeled placeholder body; it is section-chrome evidence, not a live wallet inventory.

Mobile section chrome

Desktop section chrome

PR #31213, opened 2026-09-13
test(app): let the onboarding smoke harness confirm remote adoption (#31212)

### LINKED ISSUE #31212: test(app): onboarding smoke harness predates the remote adoption probe and config completion marker, failing the Zero-Key app browser core lane
## Summary

`scenarios / Zero-Key app browser core` in the develop "Validate develop" workflow has been red on every run since `619545c5357f` (PR #30989; first red run 2026-09-11 01:36Z, last green `77041c39df`). The single failing case is `test/ui-smoke/onboarding-to-home.spec.ts` › "Remote connect adopts a host and replaces onboarding without the old screen":

```
Error: expect(locator).toBeVisible() failed
Locator:  getByTestId('home-launcher-surface')
Expected: visible
Received: hidden
Timeout:  60000ms
```

The Playwright error context shows the runtime chooser still on screen after the `eliza:connect` event (`onboarding-step:runtime onboarding-choices:__first_run__:runtime:cloud,…`). Latest red run: https://github.com/elizaOS/eliza/actions/runs/34723784763/job/103635554387 (develop `1fa9dbf22d74`). Reproduces locally with the same assertion.

## Root cause

#30989 changed remote adoption (`packages/ui/src/first-run/adopt-remote-first-run.ts`): before completing first-run it now (1) reads `/api/first-run/status`, (2) probes `/api/status` and refuses a host unless `state === "running"` and `canRespond === true`, (3) writes the completion marker with `PUT /api/config { meta: { firstRunComplete: true } }`, and (4) re-reads `/api/first-run/status` expecting `complete: true`. The in-chat onboarding smoke harness (`onboarding-to-home.shared.ts`) predates that contract: the default `/api/status` mock omits `canRespond`, `PUT /api/config` falls through to the static server, and the first-run status only flips after `POST /api/first-run`. Adoption throws, first-run never completes, and the chooser stays up. The real-server coverage added in #30989 (`remote-adoption.real-server.test.ts`) is green; only the browser harness drifted.

## Acceptance criteria

- `bun run --cwd packages/app test:e2e test/ui-smoke/onboarding-to-home.spec.ts --project=chromium` passes all six cases on develop.
- The harness records `PUT /api/config` bodies on its route state and derives `/api/first-run/status` completion from the host completion marker as well as the legacy `POST /api/first-run`, so the remote path is asserted against the real adoption writes.
- The remote path presents the adopted host as running and able to respond; the other onboarding paths are unaffected.
- No production code changes.

hermesagent270-commits:
Closes #31212

## What

Teach `packages/app/test/ui-smoke/onboarding-to-home.shared.ts` the writes remote adoption performs since #30989:

- `OnboardingRouteState` records every `PUT /api/config` body (`configWrites`), and `/api/first-run/status` reports `complete` when either the legacy `POST /api/first-run` fired or a config write carried `meta.firstRunComplete: true` (`hostFirstRunCompleted`).
- `installHomeRoutes` answers `PUT /api/config` with the merged config instead of falling through to the static server.
- `connectRemoteFirstRunToHome` presents the adopted host as `state: "running", canRespond: true`, which the adoption probe requires; the default smoke status omits `canRespond`.

## Why

`adoptRemoteAgentFirstRun` now probes host status, writes the completion marker through the config patch API, and re-reads first-run status to confirm it. With the old mocks the probe rejected the host, so first-run never completed, the runtime chooser stayed on screen, and `scenarios / Zero-Key app browser core` has been red on every develop run since 2026-09-11 (latest: [run 34723784763](https://github.com/elizaOS/eliza/actions/runs/34723784763/job/103635554387)). Test-harness-only change; the real-server adoption coverage from #30989 is untouched.

## Evidence

Negative control, develop's harness at `1fa9dbf22d74` (`bun run --cwd packages/app test:e2e test/ui-smoke/onboarding-to-home.spec.ts --project=chromium -g "Remote connect"`):

```
  ✘  1 [chromium] › test/ui-smoke/onboarding-to-home.spec.ts:215:3 › … › Remote connect adopts a host and replaces onboarding without the old screen
    Error: expect(locator).toBeVisible() failed
  1 failed
```

With this change, the whole spec:

```
  ✓  1 … Local onboarding lands on home and opens the Launcher pager
  ✓  2 … Cloud onboarding connects, binds an agent, and lands on the home
  ✓  3 … Local cloud-inference onboarding completes in chat
  ✓  4 … Other provider completes in chat and hands off to Settings without model download
  ✓  5 … Remote connect adopts a host and replaces onboarding without the old screen
  ✓  6 … tutorial CHOICE 'Take the tutorial' completes onboarding and launches the tour
  6 passed (44.2s)
```

Run on Node 24.15.0 with the pinned Bun 1.3.14 (the ui-smoke lane refuses older Node). `bunx biome check` on the file: clean. Root `bun run verify` is running on a stack with #31211; result will be posted here.

### DIFF
--- a/packages/app/test/ui-smoke/onboarding-to-home.shared.ts
+++ b/packages/app/test/ui-smoke/onboarding-to-home.shared.ts
@@ -296,6 +296,25 @@ export async function injectFullCapabilityHost(page: Page): Promise<void> {
 // network boundary (the single `persistFirstRun` funnel in first-run-finish.ts).
 export interface OnboardingRouteState {
   firstRunPosts: unknown[];
+  /**
+   * Bodies of every `PUT /api/config` the page issued. Remote adoption completes
+   * first-run on the host with a `meta.firstRunComplete` patch (see
+   * `adopt-remote-first-run.ts`) instead of `POST /api/first-run`, so the
+   * first-run status route derives completion from both writes.
+   */
+  configWrites: Record<string, unknown>[];
+}
+
+/** True once a config write carried the host first-run completion marker. */
+export function hostFirstRunCompleted(state: OnboardingRouteState): boolean {
+  return state.configWrites.some((write) => {
+    const meta = write.meta;
+    return (
+      typeof meta === "object" &&
+      meta !== null &&
+      (meta as { firstRunComplete?: unknown }).firstRunComplete === true
+    );
+  });
 }
 
 async function routeFirstRunIncomplete(
@@ -328,7 +347,7 @@ async function routeFirstRunIncomplete(
       return;
     }
     await fulfillJson(route, {
-      complete: state.firstRunPosts.length > 0,
+      complete: state.firstRunPosts.length > 0 || hostFirstRunCompleted(state),
       cloudProvisioned: false,
     });
   });
@@ -349,22 +368,32 @@ async function routeFirstRunIncomplete(
 export async function installHomeRoutes(
   page: Page,
 ): Promise<OnboardingRouteState> {
-  const state: OnboardingRouteState = { firstRunPosts: [] };
+  const state: OnboardingRouteState = { firstRunPosts: [], configWrites: [] };
   await installDefaultAppRoutes(page);
   await routeFirstRunIncomplete(page, state);
 
+  const homeConfig = {
+    cloud: { enabled: false },
+    media: {},
+    plugins: { entries: {} },
+    ui: { avatarIndex: 1 },
+    wallet: {},
+  };
   await page.route("**/api/config", async (route) => {
-    if (route.request().method() !== "GET") {
+    const method = route.request().method();
+    if (method === "PUT") {
+      // The real route merges the patch into the host config; the harness only
+      // needs the completion marker to become visible to /api/first-run/status.
+      const patch = route.request().postDataJSON() as Record<string, unknown>;
+      state.configWrites.push(patch);
+      await fulfillJson(route, { ...homeConfig, ...patch });
+      return;
+    }
+    if (method !== "GET") {
       await route.fallback();
       return;
     }
-    await fulfillJson(route, {
-      cloud: { enabled: false },
-      media: {},
-      plugins: { entries: {} },
-      ui: { avatarIndex: 1 },
-      wallet: {},
-    });
+    await fulfillJson(route, homeConfig);
   });
 
   await page.route("**/api/stream/settings", async (route) => {
@@ -998,7 +1027,7 @@ export async function completeOnboardingToHome(
   page: Page,
   click: (locator: Locator) => Promise<void>,
   opts: { state: OnboardingRouteState; tutorial?: "start" | "skip" } = {
-    state: { firstRunPosts: [] },
+    state: { firstRunPosts: [], configWrites: [] },
   },
 ): Promise<{ surface: Locator }> {
   const { state, tutorial = "skip" } = opts;
@@ -1342,6 +1371,24 @@ export async function connectRemoteFirstRunToHome(
   const apiBase =
     opts.apiBase ??
     (await page.evaluate(() => window.location.origin.toString()));
+
+  // Adoption probes the host before writing its completion marker and refuses
+  // one that is not running and able to respond; the default smoke status
+  // omits `canRespond`, so present the adopted host as ready. Later routes win.
+  await page.route("**/api/status", async (route) => {
+    if (route.request().method() !== "GET") {
+      await route.fallback();
+      return;
+    }
+    await fulfillJson(route, {
+      state: "running",
+      canRespond: true,
+      agentName: "Playwright Smoke",
+      model: "ui-smoke",
+      startedAt: Date.now() - 60_000,
+      uptime: 60_000,
+    });
+  });
 
   await page.evaluate((gatewayUrl) => {
     document.dispatchEvent(


attentionhead (approved):
**APPROVE** — verified at head `9a52aec8f05df6741bb6977dd0898c9ad51b90b4` (base `b615a36f91c0dbbfe875b2a798c7fd8861ebfcc2`), single file `packages/app/test/ui-smoke/onboarding-to-home.shared.ts` +58/-11.

## The mocks now match the real contract exactly

I checked each harness change against `packages/ui/src/first-run/adopt-remote-first-run.ts` rather than against the failure symptom, because a smoke-harness repair is only a fix if the mock reproduces the production call sequence. `adoptRemoteAgentFirstRun` does exactly four things, and the diff answers all four:

1. `getFirstRunStatus()` → must be incomplete to proceed. `routeFirstRunIncomplete` still starts `complete: false` (`configWrites` is empty), so the adoption path is entered rather than short-circuited by the `alreadyComplete` early return.
2. `getStatus()` → `if (status.state !== "running" || status.canRespond !== true) throw ElizaError("REMOTE_ADOPTION_HOST_NOT_READY")`. The default smoke status in `helpers.ts:1747` emits `state: "running"` but **no `canRespond` key at all**, so `status.canRespond !== true` was true and the probe threw. The new route is load-bearing, not cosmetic.
3. `updateConfig({ meta: { firstRunComplete: true } })` → the patch shape `hostFirstRunCompleted` matches is character-for-character the shape the production call sends. The `typeof meta === "object" && meta !== null` guard before the cast is correct, not defensive noise: `typeof null === "object"`, so the null check is load-bearing.
4. The **re-read** `if ((await client.getFirstRunStatus()).complete !== true) throw ElizaError("REMOTE_ADOPTION_NOT_CONFIRMED")` → this is precisely what the amended `complete: state.firstRunPosts.length > 0 || hostFirstRunCompleted(state)` satisfies. Without the `||` arm the adoption would have failed at the confirmation step even with a passing probe, so both halves of the change are required; neither is redundant.

## Route precedence holds, and I checked rather than assumed

The comment "Later routes win" is correct for Playwright (most-recently-registered handler takes precedence), but that only helps if the ordering is actually right at the call site. It is: `onboarding-to-home.spec.ts:219` calls `installHomeRoutes` (which calls `installDefaultAppRoutes`, registering the default `**/api/status`), and `:225` calls `connectRemoteFirstRunToHome`, which registers the override. The override is strictly later, and `helpers.ts:1747` is the only other `**/api/status` registration in the harness — nothing re-shadows it afterward.

The `method !== "GET" → route.fallback()` guard on the override preserves the existing behaviour for non-GET status requests instead of swallowing them.

## Blast radius is contained

- Both `OnboardingRouteState` literals were updated (`:371`, `:1030`); `git grep "firstRunPosts: \["` across `packages` returns exactly those two, so there is no third construction left to break the new required `configWrites` field at compile time or to reach `.some()` on `undefined`. This matters because six other specs import this shared harness (`onboarding-cloud-only`, `onboarding-confused-user`, `onboarding-to-home-mobile`, `all-pages-clicksafe`, `walkthrough/journey`, the android-browser spec).
- The `complete:` widening cannot leak into those other flows: `configWrites` only becomes non-empty when a `PUT /api/config` carries `meta.firstRunComplete: true`, which only remote adoption sends.
- Hoisting `homeConfig` and returning `{ ...homeConfig, ...patch }` from the PUT is a faithful stand-in for the real merge semantics, and the GET arm is byte-identical to what it replaced.

## Verb check

`ElizaClient.prototype.updateConfig` (`packages/ui/src/api/client-agent.ts:1800`) tries `invokeLocalDesktopAgentRpc` first and only falls back to `fetch("/api/config", { method: "PUT" })`. Under Playwright chromium the desktop RPC path has no host to reach, so it throws, is caught, and the HTTP `PUT` the harness intercepts is the one that fires. `getFirstRunStatus` is `GET /api/first-run/status` (`client-agent.ts:1360`), which `routeFirstRunIncomplete` already owns. The single `PUT`-only branch is therefore sufficient — no PATCH/POST variant reaches this route.

## Non-vacuity

The assertion this unblocks can still fail. `connectRemoteFirstRunToHome` asserts `first-run-runtime-chooser` is hidden, `eliza:first-run-complete === "1"`, `elizaos:active-server` contains `"kind":"remote"`, and `firstRunPosts.length <= 1` (no double-submit). Revert any one of the three harness changes and adoption throws at its corresponding step, first-run never completes, and the chooser stays visible — which is the exact reported failure. The change makes the test reach its assertions; it does not weaken them.

Scope is test-harness-only; no production file is touched, and the real-server adoption coverage from #30989 is untouched.

ss251 (approved):
9a52aec8f05df6741bb6977dd0898c9ad51b90b4

Test-only change to packages/app/test/ui-smoke/onboarding-to-home.shared.ts teaching the smoke harness the remote-adoption first-run path referenced by #31212. Findings, all from the diff at this head:

1. Behavioral, not cosmetic: the first-run status route now reports complete when a captured PUT /api/config carried meta.firstRunComplete === true, not only after POST /api/first-run (:350; helper at :309). Input->wrong output: if a regression drops the adoption write, complete stays false and any remote-adoption scenario stalls and fails — the harness discriminates the claimed behavior instead of passing vacuously.

2. Fidelity gaps in the PUT branch (:384-391), acceptable for a harness but stated: (a) the fulfill is a shallow spread and later GET /api/config (:396) still returns unmerged homeConfig, so client code re-reading config for the marker would see a host unlike production; (b) postDataJSON() on an empty/non-JSON body pushes null unguarded, which would throw at write.meta — unreachable while adoption always sends JSON, but a one-line object guard would harden it.

3. The new GET **/api/status route (:1375) leans on Playwright's last-registered-wins ordering, as its comment correctly states, and falls back for non-GET like its neighbors. Its payload (state, canRespond, agentName, model, startedAt, uptime) is asserted, not verified: adopt-remote-first-run.ts is outside this diff, so exact-shape fidelity rests with the consuming smoke scenario.

Adversarial: no dependencies, lockfile, CI, hooks, binaries, secrets, telemetry or injection text; prior GET behavior preserved verbatim via homeConfig; the || only widens completion on a product-issued write, so nothing is weakened; default state at :1031 keeps every caller type-correct.

Verification: static read of the diff at this head only; no checkout, no suite run.

Recommendation: merge

hermesagent270-commits:
Root `bun run verify` on a stack branch holding #31211 and #31213 on develop `1fa9dbf22d74` (stack head `c0a0787ef740`):

```
verify exit 0
```

Log kept locally at `/root/eliza-fixes/verify-0913b.log`.

lalalune:

Retaining the smoke-harness repair. Current adoptRemoteAgentFirstRun requires running/canRespond, writes meta.firstRunComplete through updateConfig, and re-reads first-run status; the old harness cannot represent that sequence. Before merge, make the fake config state consistent: PUT currently returns a merged value but GET continues returning homeConfig without that update, and completion is inferred from any historical true write rather than current stored state. Store the accepted patch using the relevant merge semantics and derive reads/status from it. Scope the ready-host response to the intended adoption endpoint so a request to a wrong host cannot silently receive the same successful fixture. The actual remote-host integration coverage remains distinct from this mocked browser smoke scenario. This is test UI infrastructure and is not subject to closing a product redesign solely on author identity. Keep the issue open until the repaired smoke lane is verified on the current head.

lalalune:
Superseded by the reviewed integration batch #31481, now merged into develop as 1db607dba6e750f4dd103cee18f3617d9e1e54b0. The merged batch includes this fixture repair and passed 150 Chromium desktop/mobile cases, 2,573 app unit tests (19 skipped), six daemon tests, app typecheck/lint, root verification and hosted run35144751481. Exact-head command logs and validation manifest are attached to #31481. Closing this duplicate proposal without deleting its source branch; this PR is not counted as a separate merge.

PR #31216, opened 2026-09-13
ci(test): run the app-core screenshot-quality diagnostics file directly (#31215)

### LINKED ISSUE #31215: ci(test): Zero-Key diagnostics runs the whole app-core suite instead of the screenshot-quality file and fails on the login sidecar test
## Summary

`tests / Zero-Key diagnostics` in the develop "Validate develop" workflow is red on every run. The step "App-core screenshot quality diagnostics coverage" is meant to run one file, but the log shows the entire app-core suite running and one unrelated test failing:

```
 FAIL  src/services/steward-sidecar-login.test.ts > provisions a local wallet and reopens the same authority after restarting
Error: Login API entry point not found. Install the @elizaos/login workspace or set STEWARD_ENTRY_POINT to its embedded entry.
 Test Files  1 failed | 365 passed | 1 skipped (367)
      Tests  1 failed | 4619 passed | 14 skipped (4634)
```

Latest red run: https://github.com/elizaOS/eliza/actions/runs/34723784763/job/103635168222 (develop `1fa9dbf22d74`).

## Root cause

`.github/workflows/test.yml` runs

```
bun run --cwd packages/app-core test -- test/app/screenshot-quality.test.ts
```

but `packages/app-core/package.json#test` is a `&&` chain (`node ../scripts/run-vitest.mjs run --config vitest.config.ts && bun run test:script-suites`). Bun appends the extra argument to the script text, so it lands on the **last** command; locally the expansion is visible on the first output line:

```
$ node ../scripts/run-vitest.mjs run --config vitest.config.ts && bun run test:script-suites test/app/screenshot-quality.test.ts
```

vitest therefore runs unfiltered. Since `6dbff6e528b8` (2026-09-10) the app-core suite includes `steward-sidecar-login.test.ts`, which spawns the bundled login child and requires the `@elizaos/login` workspace (or `STEWARD_ENTRY_POINT`); the diagnostics job builds only `packages/cloud/routing`, so that test fails and takes the lane with it. The neighbouring "App screenshot quality diagnostics coverage" step already avoids this by invoking `bunx vitest run --config … <file>` directly.

## Acceptance criteria

- The step runs only `packages/app-core/test/app/screenshot-quality.test.ts` (2 tests) and the lane no longer depends on the login workspace being built.
- `bunx vitest run --config packages/app-core/vitest.config.ts packages/app-core/test/app/screenshot-quality.test.ts` passes from the repository root.
- No change to what the app-core `test` script does for its own lanes.

hermesagent270-commits:
Closes #31215

## What

In `.github/workflows/test.yml`, run the app-core screenshot-quality diagnostics file directly:

```
bunx vitest run --config packages/app-core/vitest.config.ts packages/app-core/test/app/screenshot-quality.test.ts
```

instead of `bun run --cwd packages/app-core test -- test/app/screenshot-quality.test.ts`, matching the neighbouring app step.

## Why

The app-core `test` script is a `&&` chain, so Bun appends the file path to its last command (`bun run test:script-suites test/app/screenshot-quality.test.ts`) and vitest runs the whole suite. That suite now includes `steward-sidecar-login.test.ts`, which needs the `@elizaos/login` workspace the diagnostics job never builds, so `tests / Zero-Key diagnostics` fails on a test the step was never meant to run (latest: [run 34723784763](https://github.com/elizaOS/eliza/actions/runs/34723784763/job/103635168222): `1 failed | 365 passed`). Workflow-only change.

## Evidence

Expansion of the old invocation (first line Bun prints, locally on develop `1fa9dbf22d74`):

```
$ node ../scripts/run-vitest.mjs run --config vitest.config.ts && bun run test:script-suites test/app/screenshot-quality.test.ts
 RUN  v4.1.10 …/packages/app-core
```

New invocation from the repository root:

```
 Test Files  1 passed (1)
      Tests  2 passed (2)
   Duration  724ms
```

The workflow still parses (PyYAML load; the step's `run` reads as above).

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -725,7 +725,9 @@ jobs:
         run: bunx vitest run --config packages/app/vitest.config.ts packages/app/test/screenshot-quality.test.ts
 
       - name: App-core screenshot quality diagnostics coverage
-        run: bun run --cwd packages/app-core test -- test/app/screenshot-quality.test.ts
+        # The package `test` script is a `&&` chain, so `bun run test -- <file>`
+        # appends the path to its last command and vitest runs the whole suite.
+        run: bunx vitest run --config packages/app-core/vitest.config.ts packages/app-core/test/app/screenshot-quality.test.ts
 
   zero-key-scenario-runner:
     name: Zero-Key scenario runner


hermesagent270-commits:
Workflow-facing gates from `bun run verify` on this head (`b1d61eed5032`, develop `1fa9dbf22d74`):

```
$ node packages/scripts/workflow-trigger-policy.mjs
[workflow-trigger-policy] verified 67 workflows; 1 develop push surfaces
$ node packages/scripts/audit-scripts.mjs
[audit-scripts] OK — no orphan/no-op/broken scripts.
```

Nothing outside `.github/workflows/test.yml` changes, so the Turbo typecheck/lint half of `verify` is unaffected by this diff.

attentionhead (approved):
**APPROVE** — verified at head `b1d61eed50328ae9879fb4c9bd3b580820d6f25e` (base `b615a36f91c0dbbfe875b2a798c7fd8861ebfcc2`), single file `.github/workflows/test.yml` +3/-1.

## The `&&`-chain diagnosis is exactly right

`packages/app-core/package.json:33` is:

```
"test": "node ../scripts/run-vitest.mjs run --config vitest.config.ts && bun run test:script-suites"
```

`bun run test -- <path>` appends the argument to the **end of the expanded script string**, not to each command in it. So the old step expanded to:

```
node ../scripts/run-vitest.mjs run --config vitest.config.ts && bun run test:script-suites test/app/screenshot-quality.test.ts
```

The vitest invocation — the first link — therefore received **no file filter at all** and ran the entire app-core suite, which is precisely the reported symptom. The path landed on `test:script-suites` instead, itself another `&&` chain ending in `bun test …`, so it was appended to that last command rather than to vitest. A step named "screenshot quality diagnostics coverage" was running ~366 tests it was never scoped to.

## Why that suite member fails in this job specifically

`packages/app-core/src/services/steward-sidecar-login.test.ts` does not import `@elizaos/login` directly — it imports `./steward-sidecar`. The relevant detail is in `vitest.config.ts`: `@elizaos/login` has only a **bare-specifier** alias to `packages/login/src/sdk/index.ts`, with no `/(.+)` catch-all, unlike the `@elizaos/ui`, `@elizaos/core`, and `@elizaos/shared` entries which all have one. Any deep subpath therefore falls through to package resolution, and `packages/login/package.json` maps every deep export to `./dist/...`. The `zero-key-diagnostics` job builds only `packages/cloud/routing` and the shared i18n data — it never runs `build:core` — so that `dist` does not exist in this job. The failure is environmental to the job, exactly as described, and not a defect in the test.

## No coverage is lost — I checked the owner, not just the step

This is the part worth stating explicitly, because the old command ran *both* halves of the `&&` chain and the new one runs neither the full vitest suite nor `test:script-suites`. Those ~20 `node:test`/`bun:test` script files are deliberately excluded from vitest (with an in-config comment noting the exclusion "no longer means 'runs nowhere'"), so dropping their only runner would have been a silent regression.

It is not their only runner. `packages/app-core/package.json` declares `elizaos.scripts.testLanes: ["server"]`, and the root `test:server` (`package.json:169`) is:

```
TEST_SCRIPT_FILTER='^test$' node packages/scripts/run-all-tests.mjs --lane=server …
```

`run-all-tests.mjs:243` resolves the lane through `resolveTestLaneDirs`, which `script-metadata.mjs:108` implements as `pkg.scripts.testLanes.includes(lane)`. With `TEST_SCRIPT_FILTER='^test$'` that runs app-core's **`test` script** — the whole `&&` chain, vitest suite and script suites both — in the `server-tests` job (`test.yml:251`). The diagnostics step was never the coverage owner; it was duplicating a full suite run as a side effect of the argument-forwarding bug.

## The replacement resolves correctly from the repo root

- It matches the immediately preceding step verbatim in form (`test.yml:724`, the app package), so this is an established working pattern in the same job rather than a novel invocation.
- The edited step is in `zero-key-diagnostics` (lines 667–731), which declares **no** `defaults.run.working-directory`. The only `working-directory` in the file is `test.yml:530`, a step inside the separate `desktop-contract` job. Root-relative paths are therefore correct here — worth confirming rather than assuming, since a job-level default would have silently broken both the `--config` path and the filter.
- The config sets no `root`/`dir`, so `root` is the invocation cwd (repo root) and the positional argument matches the file by path. Your reported `Test Files 1 passed (1)` is consistent with that.

## Non-vacuity

The step can still fail: it runs the real `screenshot-quality.test.ts` with its 2 assertions, and a regression in that file fails the job. The change narrows *what* the step runs, not whether it can fail — and the suite it stops running is still fully executed by the `server` lane.

Workflow-only change; no package or source file is touched.

ss251 (approved):
b1d61eed50328ae9879fb4c9bd3b580820d6f25e

The old step at `.github/workflows/test.yml:727` ran `bun run --cwd packages/app-core test -- test/app/screenshot-quality.test.ts`. `packages/app-core/package.json:33` is a `&&` chain (`run-vitest.mjs … && bun run test:script-suites`), so Bun appends the path to the last command. Develop job 103635168222 prints that expansion, then `RUN  v4.1.10 …/packages/app-core` with no file filter.

Input: one screenshot-quality file. Wrong output: `Test Files  1 failed | 365 passed | 1 skipped (367)`, FAIL `src/services/steward-sidecar-login.test.ts` → `Login API entry point not found`. That file landed in `6dbff6e528b8`. This job builds only `packages/cloud/routing` (`test.yml:709-710`), not `@elizaos/login`. The intended file still passed in that log (`✓ test/app/screenshot-quality.test.ts (2 tests)` at `:57` and `:72`).

The replacement matches the preceding app step (`test.yml:725`), which in the same job produced `Test Files  1 passed (1)`. `packages/app-core/vitest.config.ts` sets no `root`, so the repo-root `--config` and file path resolve. The dropped chain remains the `server` lane: root `package.json:169` `test:server` runs `build:core` then `--lane=server`, and app-core declares `elizaos.scripts.testLanes: ["server"]` (`package.json:349`). Permissions stay `contents: read`.

Recommendation: merge

Verified by reading the diff at this head against base `b615a36f91c0dbbfe875b2a798c7fd8861ebfcc2`, the develop job log, app-core `package.json`/`vitest.config.ts`/screenshot-quality tests, and root `test:server`; no suite applies to a workflow-only change.

Client / agent tooling: grok
— [grok-fleet-ss251]

lalalune:

Retaining the scoped diagnostics invocation fix. Current app-core test is a chained Vitest plus script-suite command, so appending the requested screenshot file does not scope the first command. The workflow still uses that chained entry point. Keep this focused repair and verify from the repository root with the job environment that exactly the intended screenshot diagnostics file is collected. Prefer the existing run-vitest wrapper if its setup is required, or document why direct Vitest is equivalent for this job. This does not certify the unrelated full app-core suite or justify dropping it from its owning lane.

lalalune:
Superseded by #31542, merged into develop as 215bd8a54121e4f434549c123c13cb96a4edced4. The screenshot diagnostics step now invokes the canonical Vitest wrapper with this exact config and file, avoiding the composite package test command. The focused file passed 2 tests; the merged candidate passed repository verification and all five hosted checks in run 35158318848. Validation logs: https://github.com/elizaOS/eliza/releases/download/pr-evidence-3/31542-eliza-31530-validation.txt. Closing this duplicate proposal without deleting its branch; this does not claim the entire develop validation run has finished.

PR #31217, opened 2026-09-13
fix: roll back rejected agreement upload sources

lalalune:
Agreement uploads could fail while leaving their private PDF and searchable document behind. Artifact persistence happened after both sources were created, and identical PDFs in separate agreement families could also reuse one document identity.

Each immutable artifact now owns its document ingestion identity. Rejected uploads remove only that attempt's document, derived fragments, and private PDF through the canonical services. Commit reconciliation preserves sources when persistence succeeded or cannot be observed; cleanup outages return an explicit storage error instead of hiding incomplete deletion.

Refs #31180. This repairs failed-ingestion lifecycle handling; it does not complete workspace deletion or choose a backup-retention policy.

Validation:

- Real PGlite and canonical private file storage: 24 integration cases passed, including persistence rejection, concurrent duplicate cleanup, independently readable identical PDFs, lost commit acknowledgements, and a real database deletion outage. PDF extraction remains a deterministic fixture.
- Package lint passed across 1,823 files.
- Full package suite: 303 files, 2,660 tests passed; 6 existing skips. Post-build package typecheck passed. Root verification passed all 373 tasks and repository audits.
- Screenshots, video, and live-model trajectories: N/A; this changes storage rollback and failure responses without changing UI or model behavior.

Deployment: not deployed.

Before screenshots: N/A - no rendered UI changes.

After screenshots: N/A - no rendered UI changes.

Walkthrough video: N/A - storage rollback has no visual interaction change.

Backend logs: [Reviewed storage evidence bundle](https://github.com/elizaOS/eliza/releases/download/pr-evidence-12/31180-bettina-agreement-rollback-849a3c8-evidence.zip) includes the before failure and verification summaries from the real PGlite runtime; raw logs containing ephemeral fixture credentials are excluded.

<details><summary>Completed runtime validation transcript</summary>

```text
Real PGlite + canonical private media, source 849a3c8b1636fc167f79006dabf11f64ff771a24:
Test Files  1 passed (1)
Tests       24 passed (24)
Duration    11.98s
Full personal-assistant package: 303 test files passed; 2660 tests passed; 6 skipped.
Root verification: 373 successful tasks, 373 total; 5m18.458s; repository audits passed.
Post-build package typecheck exited 0; package lint checked 1823 files with no fixes.
```
</details>

Frontend logs: N/A - no frontend execution path changed.

Real-LLM trajectory: N/A - no model, prompt, provider or action behavior changed.

Domain artifacts: [Direct storage observations](https://github.com/elizaOS/eliza/releases/download/pr-evidence-12/31180-bettina-agreement-rollback-849a3c8-storage.json). [Observed database rows and private-file inventories](https://github.com/elizaOS/eliza/releases/download/pr-evidence-12/31180-bettina-agreement-rollback-849a3c8-evidence.zip) cover three rejected-write boundaries, the surviving concurrent upload, committed-source preservation and a deletion outage. The seven JSON receipts and source producer were inspected. Bundle SHA-256: `92d42d56e424fcc30ba622f0efa090f29d06874b7af280fa0a08ab3f28c9221d`.

<details><summary>Inspected storage results</summary>

```text
rejected-artifact.json: before and after document/fragment IDs are identical; private filenames are identical; no artifact row remains.
rejected-document.json: before and after document/fragment IDs are identical; private filenames are identical; no artifact row remains.
rejected-fragment.json: before and after document/fragment IDs are identical; private filenames are identical; no artifact row remains.
concurrent-duplicate.json: one fulfilled and one rejected upload; exactly one new PDF and document remain; original PDF SHA matches the persisted artifact.
lost-ack-committed.json and lost-ack-unreadable.json: committed document exists and PDF read SHA matches the artifact after each injected acknowledgement failure.
cleanup-outage.json: the deliberately rejected document deletion leaves one reported document; the private PDF is removed. Fixture teardown removes the document after dropping the fault trigger.
Public bundle download: identical to local ZIP; CRC and all 11 member SHA-256 entries passed.
```
</details>

### DIFF
--- a/plugins/plugin-personal-assistant/README.md
+++ b/plugins/plugin-personal-assistant/README.md
@@ -235,6 +235,13 @@ persists the media SHA-256, handle, document id, byte size, MIME type, filename,
 parser-derived page count, complete extracted text and page map, and version chain; it does
 not create another permanent file store.
 
+Each immutable agreement owns a distinct document ingestion identity. If artifact
+persistence rejects the upload, ingestion removes that attempt's document,
+derived fragments, and private PDF through the canonical services. A failed
+commit observation preserves the sources for reconciliation; an incomplete
+cleanup returns an explicit storage error and reports the affected artifact
+and source handles to runtime diagnostics.
+
 Owner- or agent-extracted obligations begin as `proposed` and must carry an
 in-range page citation plus the cited source text. Only the owner can make the terminal
 `approved` or `rejected` decision. Agent and chat pins are independent
--- a/plugins/plugin-personal-assistant/src/lifeops/household/agreement-knowledge.pglite.integration.test.ts
+++ b/plugins/plugin-personal-assistant/src/lifeops/household/agreement-knowledge.pglite.integration.test.ts
@@ -16,6 +16,7 @@ import {
   type AgentRuntime,
   attestAuthenticatedApiDeliveryAudience,
   ChannelType,
+  DocumentService,
   documentsPluginCore,
   type IAgentRuntime,
   type IFileStorageService,
@@ -25,6 +26,7 @@ import {
   ServiceType,
   type UUID,
 } from "@elizaos/core";
+import type { PdfService } from "@elizaos/plugin-pdf";
 import { SELF_ENTITY_ID } from "@elizaos/shared";
 import { afterAll, beforeAll, describe, expect, it } from "vitest";
 import { tryHandleRuntimePluginRoute } from "../../../../../packages/agent/src/api/runtime-plugin-routes.ts";
@@ -46,6 +48,7 @@ import { executeRawSql, sqlQuote } from "../sql.js";
 import {
   AgreementKnowledgeError,
   AgreementKnowledgeRepository,
+  AgreementKnowledgeService,
   createAgreementKnowledgeService,
   type ParentingAgreementArtifact,
 } from "./agreement-knowledge.js";
@@ -53,6 +56,7 @@ import {
   getHouseholdCoordinationService,
   type HouseholdCoordinationService,
 } from "./service.js";
+import { DEFAULT_HOUSEHOLD_ID } from "./types.js";
 
 const fileStoragePlugin: Plugin = {
   name: "agreement-knowledge-test-file-storage",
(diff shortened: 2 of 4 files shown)


ss251 (approved):
Reviewed head 849a3c8b1636fc167f79006dabf11f64ff771a24.

Confirmed the risky hunk by tracing the new catch path in `agreement-knowledge.ts` (createAgreementVersion) at this head: `documentId` is assigned only after `addDocument` resolves, the rollback gate calls `getArtifact(artifactId)` before any deletion, and both a failed commit observation and a persisted artifact throw `AGREEMENT_INGESTION_RECONCILIATION_REQUIRED` with sources untouched — the lost-ack tests force a real commit before faulting, so both branches are exercised. Cleanup rejections fold into `AGREEMENT_INGESTION_CLEANUP_FAILED`, mapped to 503 at `agreement-knowledge-routes.ts:84`.

Non-blocking: `deletePrivate(stored.fileName)` is safe only if storePrivate returns per-call names; the guard `stored.fileName.startsWith(\`${expectedSha256}.\`)` would also accept a content-addressed shared name, under which a losing concurrent duplicate deletes the winner's private PDF.

Recommendation: merge

attentionhead (approved):
## APPROVE — verified at head `849a3c8b1636fc167f79006dabf11f64ff771a24` (base `b615a36f91c0dbbfe875b2a798c7fd8861ebfcc2`, 4 files, +433/-63)

The rollback is correct and, more importantly, correct in the two places where a naive rollback would destroy committed data. I tried to break it on both and could not.

### 1. The cleanup cannot delete another artifact's source bytes

This was my primary concern: `deletePrivate(stored.fileName)` runs unconditionally in the cleanup branch, and `stored.fileName` is content-addressed — the guard at `agreement-knowledge.ts:1160` asserts `stored.fileName.startsWith(`${expectedSha256}.`)`. If identical bytes shared one private file, a failed upload of a PDF that a *committed* artifact also referenced would delete the committed artifact's source.

It cannot, because `persistPrivateMediaBytes` (`packages/agent/src/api/media-store.ts`) mints a per-call nonce:

```ts
const nonce = crypto.randomBytes(8).toString("hex");
const fileName = `${hash}.private-${nonce}.${extForMime(mimeType)}`;
```

with the invariant stated inline — *"two independent private capabilities with identical bytes … consuming one must never delete the other's file."* So `storePrivate` is not deduplicating, the filename is unique per upload attempt, and the cleanup is scoped to bytes this attempt alone wrote. `deletePrivateMediaFile` additionally re-validates against `PRIVATE_MEDIA_FILE_NAME` and the resolved dirname before unlinking.

The integration test at `:1464` pins exactly this and is **non-vacuous**: it snapshots the media dir, races two identical uploads, and asserts the surviving delta is `[artifact.mediaFileName]` — the winner's file only. Revert the cleanup and the loser's nonce-named file lingers, so that assertion fails. Revert it too aggressively (hash-scoped deletion) and `readOwnerPdf` on the winner fails in the same test.

### 2. The document-identity fix is load-bearing, not cosmetic

```diff
-        originalFilename: `${stored.hash}.txt`,
+        originalFilename: `${stored.hash}.${artifactId}.txt`,
```

Since `artifactId` is `hag_${crypto.randomUUID()}` minted at `:1154` per attempt, two agreement families ingesting the same PDF now get distinct documents rather than colliding on one. `:1431` covers it and asserts `second.documentId !== first.documentId` plus per-artifact `metadata.mediaFileName` correspondence — again failing if reverted. I checked every `originalFilename` reference under `plugins/plugin-personal-assistant/src`; nothing parses or pattern-matches the old `${hash}.txt` shape, so the format change breaks no lookup. Sequential duplicates are still rejected earlier by `getArtifactByContent` (`:1098`), before any file is written, so `AGREEMENT_DUPLICATE_CONTENT` remains reachable and is not weakened by this.

### 3. The three-way catch topology matches the data it is protecting

The move of the hash/size/filename validation, `addDocument`, and `insertArtifact` *inside* the try (with `storePrivate` left outside and `let documentId: UUID | null = null` hoisted) makes the failure classes distinguishable, and each gets the right disposition:

- **`getArtifact` itself throws** → `AGREEMENT_INGESTION_RECONCILIATION_REQUIRED`, nothing deleted. This is the right call: the comment *"a failed commit observation cannot authorize source deletion"* is the correct policy, since an unobservable commit state must not license destroying the sources that would let an operator reconstruct it.
- **`persisted` non-null** → the artifact committed despite the throw, so sources are preserved and the operator is told to inspect the existing version. Deleting here would orphan a live artifact.
- **`persisted` null** → genuinely uncommitted, so document + private file are both removed via `Promise.allSettled`, and any rejection is surfaced as `AGREEMENT_INGESTION_CLEANUP_FAILED` with an `AggregateError` carrying the original error first. Partial cleanup is reported rather than swallowed; `:1577` covers the document-store rejection path.

`Promise.allSettled` (not `all`) is the right choice — a document-delete failure must not skip the file delete.

### Out-of-diff symbols I verified rather than assumed

- `documents.deleteDocumentWithAccessContext(id, { requesterEntityId, role, isOwner })` — the handle is `DocumentService` (`deps.documents(): DocumentService | null`, `:988`), whose method is **concrete and non-optional** at `packages/core/src/features/documents/service.ts:1107`. Worth stating explicitly because `packages/agent/src/api/documents-service-loader.ts:140` declares an *optional* `deleteDocumentWithAccessContext?`, and `plugins/plugin-documents/src/routes.ts:1197` therefore guards it before calling. That guard is not needed here: this is a different type, and `documents` is already null-checked at `:1113` before the try. Had the loader type been in play, the unguarded call would have thrown a `TypeError` while building the array literal — *before* `Promise.allSettled` ran — masking the original error and skipping `deletePrivate` entirely. It doesn't.
- `fileStorage.deletePrivate(fileName): Promise<boolean>` — `packages/agent/src/services/file-storage.ts:98`, delegating to `deletePrivateMediaFile`.
- `repository.getArtifact(id): Promise<ParentingAgreementArtifact | null>` — `agreement-knowledge.ts:567`.
- `runtime.reportError(scope, error)` — `packages/core/src/types/runtime.ts:1126`.
- `AggregateError` as the cause argument is consistent with the existing `AgreementKnowledgeError` 4-arg shape.

### Routes

Mapping both new codes alongside `AGREEMENT_STORAGE_UNAVAILABLE` to 503 is defensible — both describe a storage-layer state the caller may retry against after operator intervention, rather than a client-correctable request. Not a blocker either way; I mention it only because `RECONCILIATION_REQUIRED` also has a 409-ish reading ("a version already exists, go look at it"), and if the API surface later distinguishes retryable from inspect-first, that's the line to split on.

No defect found at this head.

Issue #30273, opened 2026-09-02
fix(cloud): preserve trace correlation through dedicated chat and certify literal frontend latency


lalalune:
## Problem

The staging latency certification covers the Worker OpenAI-compatible gateway, but the real dedicated frontend sends `/api/conversations/:id/messages/stream` through the dedicated proxy, agent router/container, plugin-elizacloud, API Worker, and provider. The agent chat route currently replaces the inbound trace identifier with a fresh inference trace, so Cloudflare, agent, gateway, and provider logs cannot be joined exactly. The existing dedicated latency script also adds instrumentation headers that the browser does not send and therefore is not a literal wire mimic.

## Acceptance criteria

- Adopt only a validated closed-schema inbound trace identifier at the agent chat boundary and propagate/echo it safely through the dedicated path without exposing credentials or arbitrary header data.
- Preserve existing behavior when the header is absent or invalid by minting a fresh trace.
- Add regression coverage for valid adoption, invalid rejection/fallback, propagation, and no secret/model-body logging.
- Add or extend a probe that can send the literal dedicated browser request shape and report headers/first event/first token/total timing separately from the instrumented probe.
- Keep Pages boot, browser dispatch/render, dedicated proxy/router/container, API gateway/auth, and provider phases clearly separated in evidence.
- Run focused tests, package typechecks/lint, production bundle/route contracts where affected, and exact-head hosted checks.

## Security / evidence

Do not persist session tokens, pairing bearer values, prompts, model output, raw HAR, or raw trace logs in artifacts. Durable evidence must contain timing/correlation metadata only.

HomunculusLabs:
CLAIMING: agent-side trace correlation for dedicated chat — adopt a validated closed-schema inbound trace identifier at the agent chat boundary (fresh-trace fallback on absent/invalid), propagate/echo through the dedicated proxy/router/container/API gateway path, regression coverage (adoption, rejection, propagation, no secret/model-body logging), plus a literal wire-shape latency probe reported separately from the instrumented probe, with phase-separated evidence. Will coordinate with #30060 (probe wire shape) without clobbering it.

AI provider/model: zai / glm-5.3
Client / agent tooling: Hermes Agent
Skill revision: elizaOS/eliza@3e2749ee1cf340314f9dc2b36e0c4333c18ea760:packages/skills/skills/contribute-to-eliza
Attribution status: self-reported
— [hermes-t3rpz]

HomunculusLabs:
[...]

PR #31220, opened 2026-09-13
feat(cloud): report embedding gateway timing

NubsCarson:
# Relates to

Related to #30273. Production embedding calls measured 4.65 s, 1.16 s and 2.67 s, but only module initialization was visible in response timing.

# Sync with develop

Source head `598ca32cd276e81f28268e177de295f79e92a09e` is directly based on freshly fetched `origin/develop` `f3a0173b8d4d8be7f19c5dc65e610f8217ca18ee`. Both changed files are byte-identical to the production-base candidate that passed full root verification. On this exact develop-based head, all 47 embedding tests and API type checking/router contract/Worker dry-run bundle pass. Full root verification passed on this exact head. The final upstream rebase included only receipt-fixture repair #31197, whose eight tests also pass. All five hosted checks passed in run 34742210113, including billing replay and Windows security; the subscription-authority job correctly skipped out-of-scope PostgreSQL cases.

- [x] Targets develop and applies without conflicts to the freshly fetched head.
- [x] Workspace dependencies and exact-head root verification validated.
- [x] Exact-head hosted PR checks passed; staging request evidence is tracked below.

# Risks

Low: adds response headers and one structured timing record to embedding requests. Authentication, rate limits, provider-dispatch admission, provider selection, vector bytes and deferred billing remain intact. Timing stops at direct fetch or the SDK invocation and therefore excludes provider execution.

# Background

## What does this PR do?

Reuses the existing chat gateway telemetry for all three embedding paths: direct pass-through, single SDK input and SDK batches. Responses report bounded auth-cache diagnostics and auth/middle/reserve/setup/pre-forward durations. This lets live probes distinguish a cache lookup or gateway admission delay from time after provider handoff.

Backend diagnostics for a measured Dedicated chat delay; no latency improvement is claimed from instrumentation alone.

No user-facing documentation change. Uses the existing HTTP telemetry format.

# Testing

The three clock-controlled cases in embeddings-passthrough.test.ts. All fail on the production base because timing is absent, and all pass on the candidate. A simulated 1,000 ms provider wait is excluded, while the provider-dispatch admission check is included.

## Detailed testing steps

On this exact source head, 47 tests pass across seven isolated embedding suites, including deferred billing, insufficient-credit rejection, cache behavior and unchanged vectors. API type checking, router contract and Worker dry-run bundle pass. Full root verification passed on the byte-identical production-base candidate; the final develop-based run also passed on this exact head, including all repository audits. Hosted staging evidence remains pending; this draft is not a release certificate.

# Evidence Gate

- [ ] Before screenshots: N/A - no rendered UI changes.

- [ ] After screenshots: N/A - no rendered UI changes.

- [ ] Walkthrough video: N/A - HTTP telemetry only.

- [x] Backend HTTP route harness transcript at this exact source head, with mocked providers and a controlled clock. Hosted staging timing remains pending.

```text
(pass) embeddings pass-through (#15512) > passthrough reports admission timing without including the provider wait [6.98ms]
(pass) embeddings pass-through (#15512) > sdk-single reports admission timing without including the provider wait [1.09ms]
(pass) embeddings pass-through (#15512) > sdk-batch reports admission timing without including the provider wait [0.77ms]
```

- [ ] Frontend logs: N/A - no frontend change.

- [ ] LLM trajectory: N/A - no model, prompt, action or agent behavior change; real embedding transport measurements motivate the work.

- [x] Domain boundary checks at this exact source head: fixture vectors and billing ownership are verified below. These are isolated route tests, not a claim of live provider acceptance.

```text
(pass) embeddings — success settles to actual usage exactly once > success path settles via the route settler once; the catch never double-refunds [1.01ms]
(pass) POST /api/v1/embeddings — deferred billing > billUsage is scheduled via waitUntil, not awaited before the response [6.32ms]
(pass) POST /api/v1/embeddings — returned vectors unchanged > single input: response embedding is byte-identical to the embedder output [0.32ms]
(pass) POST /api/v1/embeddings — returned vectors unchanged > array input: each response embedding matches the embedder output [0.23ms]
```

- [ ] OCR review: N/A - no rendered visual surface.

# Evidence Details

## Real LLM-call trajectory

N/A - this change measures the embedding gateway; it does not modify inference content or model selection.

## Backend + frontend logs

Local tests demonstrate a 48 ms gateway boundary split as auth 11 ms, middle 13 ms, reserve 17 ms and setup 7 ms, despite a 1,000 ms simulated provider delay. Hosted request evidence will be added after staging.

N/A - no rendered UI changes.

### Before

N/A - no rendered UI changes.

### After

N/A - no rendered UI changes.

### Walkthrough video

N/A - HTTP response telemetry only.

## Audio / voice walkthrough

## Known gaps / failures

Hosted telemetry is not yet deployed. Staging candidate #31221 contains the same two changed files and the complete production-base tree that passed root verification. This rebased head passed all hosted checks. Staging release 34742396404 is in progress; production candidate #31223 remains a draft pending its certificate and live probe. First-time Google identity and occupied-fleet retirement selection are tracked separately in the Dedicated completion goal and are outside this two-file change.

N/A - all exact-head local and hosted checks passed. Source merge does not authorize bypassing the separate staging certificate and live production-promotion gates.


attentionhead (approved):
Reviewed at head `b02cf69637a79099a495094493d55bc72cb3ebf6` (base `17ef837be27ebb8d20135e8dfb68b034b7454703`, 2 files, +212/-25). Verified against the PR's own claim that this reuses the chat gateway's telemetry rather than introducing a parallel mechanism.

## The instrumentation sits where the description says it does

The boundary claim is the load-bearing one, since a timing split that silently includes provider execution would be worse than no timing at all. It holds, and for a specific reason: `bindGatewayHandoffTelemetry` returns `(input) => invokeAtGatewayHandoff(telemetry, () => operation(input))`, so evaluating `bindGatewayHandoffTelemetry(...)({ method, headers, body })` builds the argument object *before* the returned function is entered. Argument and config construction therefore stay on the pre-forward side of the boundary, which is what the helper's own comment promises and what the three call sites here rely on.

`invokeAtGatewayHandoff` captures before `operation()` and emits in a `finally`. Because `operation()` returns a promise, that `finally` runs synchronously at handoff rather than at settlement — so the recorded split stops at the fetch/SDK invocation, and a provider call that throws still emits its timing instead of losing it. Both are the documented intent.

## Call-site ordering matches the template it copies

All three provider paths — passthrough `fetch`, `embedMany`, `embed` — do `await markProviderDispatched?.()` and then invoke the bound function. That is the same ordering as both chat call sites (`invokeStreamText` at `packages/cloud/api/v1/chat/completions/route.ts:3204`, `invokeGenerateText` at `:3795`), so `setupMs` covers the provider-dispatch admission gate identically in chat and embeddings. Passing `embed`/`embedMany` bare where chat wraps in an arrow is immaterial: these are module functions with no receiver to lose.

I checked the one place a mislabeled bucket could hide. `middleMs` is `tBeforeReserve - tAuth`, and `orgRateLimitPromise` is genuinely awaited at `:269` — before `tBeforeReserve` at `:353` — so rate-limit admission is accounted to `middleMs` rather than silently landing in `reserveMs`.

## The tests fail on the base for the right reason

`test.each(["passthrough", "sdk-single", "sdk-batch"])` drives a mocked `performance.now`, advancing 11 in auth, 13 in rate limiting, 17 in the reservation and 7 in `markProviderDispatched`, then 1,000 ms inside each provider stub. The asserted header `total=48;auth=11;mid=13;reserve=17;setup=7` is exactly `11+13+17+7`, and the 1,000 ms wait is absent from every bucket — so the test would fail both if the boundary moved past the provider call and if any bucket were misattributed, not merely if the header vanished. On the base the headers are absent entirely, so all three cases fail there. `fetchMock.mock.calls.length + embed.mock.calls.length + embedMany.mock.calls.length` being pinned to 1 keeps each parameterization on its intended path rather than passing vacuously on a shared one.

## Scope

All five imported helpers (`bindGatewayHandoffTelemetry`, `resolveElizaTraceId`, `snapshotGatewayPreforwardTiming`, `withGatewayPreforwardTelemetry`, `withInferenceAuthTelemetry`) and the `traceId`/`onTelemetry` options on `ResolveInferenceAuthOptions` already exist in `packages/cloud/shared/src/lib/`; the diff is confined to the two files listed, so no shared helper changes behaviour for chat as a side effect. `withGatewayPreforwardTelemetry` rebuilds headers via `new Headers(response.headers)`, which is why attaching to Hono's `c.json(...)` result works despite immutable response headers.

Telemetry is attached only to the two success returns, not to the early `warming` return or the error boundary. That reads as deliberate for a diagnostics change and I am not treating it as a defect; worth a note only if live probes later need to distinguish a slow rejection from a slow success.

Approving. The boundary is where it is claimed to be, the buckets are attributed correctly, and the tests pin both facts.

ss251 (changes requested):
b02cf69637a79099a495094493d55bc72cb3ebf6

The success-path handoff is where the description says it is. The live-probe claim is not: denial and error responses still have no gateway split.

Input: POST /v1/embeddings that stops at auth-cache warming, org rate-limit 503, empty input, reserve/admission failure, or the route catch.
Wrong output: a bare `c.json(...)` with no `X-Eliza-Preforward-Ms`, `Server-Timing`, or `X-Eliza-Auth-Trace`. `attachTelemetry` is defined at `packages/cloud/api/v1/embeddings/route.ts:83-90` and applied only to the two 200 returns at :664 and :675. Warming at :176, rate-limit cache errors at :274, `return orgRateLimited` at :293, and every catch return at :733-812 skip it.

Chat attaches on warming (`packages/cloud/api/v1/chat/completions/route.ts:1304`), standing denials (:1324/:1345), and snapshots reserve failure at :1900-1906 via `captureEarlyPreforwardTiming` (:1444). This PR says it reuses that telemetry so probes can split a 4.65s / 1.16s / 2.67s embedding delay. If those probes are 503/402, the split is still absent.

`embeddings-passthrough.test.ts:270` pins `total=48;auth=11;mid=13;reserve=17;setup=7` on 200s and excludes a 1,000 ms provider wait. Those cases would not fail if denial paths stayed dark. `bindGatewayHandoffTelemetry` at route.ts:514/:549/:562 matches the helper contract; that is not the defect.

Still draft; hosted evidence unchecked; commit has no model attribution.

Recommendation: fix

Verified by reading the two-file diff at this head against origin/develop `17ef837be27ebb8d20135e8dfb68b034b7454703`, plus `http-telemetry.ts` and `chat/completions/route.ts` at that base. Live execution blocked: no disposable sandbox on this host, so this is a static review of the exact head.

Client / agent tooling: grok
— [grok-fleet-ss251]

PR #31221, opened 2026-09-13
fix(cloud): promote embedding gateway timing to staging

NubsCarson:
# Relates to

Related to #30273. Production embedding calls measured 4.65 s, 1.16 s and 2.67 s, before this route exposed the existing gateway and auth-cache timing fields.

# Staging promotion

Narrow staging counterpart of #31220, related to #30273. Head `51cead43e00ddfcc82f8dbb06adf5fbbe57f2c8e` is based on staging `d0bb8cc28580c058fef0b73cd9bd0d3b7c547d81` and changes only the embedding route and its regression test.

The complete candidate tree `5429f8ecbb9e5b535bb5189ce1bec82b33a16f91` is identical to locally tested commit `251f914d54cf3b8a3738140a2926c8fc5ce78ba4`: full root verification, 47 embedding tests, API type checking/router contract and Worker dry-run bundle passed. No other develop changes are promoted. All five hosted checks passed in run 34741923054, including real billing replay, source smoke and Windows security. The subscription-authority job passed by recognizing an out-of-scope change; its PostgreSQL cases were skipped. This PR admits only the tested staging tree so live timing evidence can be collected before production promotion.

# Risks

Low: adds response headers and one structured timing record to embedding requests. Authentication, rate limits, provider-dispatch admission, provider selection, vector bytes and deferred billing remain intact. Timing stops at direct fetch or the SDK invocation and therefore excludes provider execution.

# Background

## What does this PR do?

Reuses the existing chat gateway telemetry for all three embedding paths: direct pass-through, single SDK input and SDK batches. Responses report bounded auth-cache diagnostics and auth/middle/reserve/setup/pre-forward durations. This lets live probes distinguish a cache lookup or gateway admission delay from time after provider handoff.

## What kind of change is this?

Backend diagnostics for a measured Dedicated chat delay; no latency improvement is claimed from instrumentation alone.

# Documentation changes needed?

No user-facing documentation change. Uses the existing HTTP telemetry format.

# Testing

## Where should a reviewer start?

The three clock-controlled cases in embeddings-passthrough.test.ts. All fail on the production base because timing is absent, and all pass on the candidate. A simulated 1,000 ms provider wait is excluded, while the provider-dispatch admission check is included.

## Detailed testing steps

[...]

PR #31223, opened 2026-09-13
fix(cloud): promote verified embedding gateway timing

NubsCarson:
# Relates to

Related to #30273. Production embedding calls measured 4.65 s, 1.16 s and 2.67 s, before this route exposed the existing gateway and auth-cache timing fields.

# Production promotion

Source #31220; staging #31221 merged at `5baad6bda5279efc38b09db61906a4e902039966`. This candidate is based directly on production/main `6a0c92c6e1e69b3c5e15011204defbbf30b2eaa8`, changes exactly two files, and has the identical complete tree `5429f8ecbb9e5b535bb5189ce1bec82b33a16f91`.

Source PR #31220 is merged. Staging release [34742396404](https://github.com/elizaOS/eliza/actions/runs/34742396404) completed successfully and published certificate artifact 10312618886 for this exact complete tree, issued 2026-09-13T06:37:49.205Z. Three live embedding requests passed with full timing fields. All five checks on this exact production head passed in run 34742467168. No speed improvement is claimed from timing instrumentation.

# Risks

Low: adds response headers and one structured timing record to embedding requests. Authentication, rate limits, provider-dispatch admission, provider selection, vector bytes and deferred billing remain intact. Timing stops at direct fetch or the SDK invocation and therefore excludes provider execution.

# Background

## What does this PR do?

Reuses the existing chat gateway telemetry for all three embedding paths: direct pass-through, single SDK input and SDK batches. Responses report bounded auth-cache diagnostics and auth/middle/reserve/setup/pre-forward durations. This lets live probes distinguish a cache lookup or gateway admission delay from time after provider handoff.

Backend diagnostics for a measured Dedicated chat delay; no latency improvement is claimed from instrumentation alone.

No user-facing documentation change. Uses the existing HTTP telemetry format.

# Testing

The three clock-controlled cases in embeddings-passthrough.test.ts. All fail on the production base because timing is absent, and all pass on the candidate. A simulated 1,000 ms provider wait is excluded, while the provider-dispatch admission check is included.

## Detailed testing steps

The complete staging tree matches the locally verified production-base candidate, including all dependencies and source files. The 47 tests across seven isolated embedding suites cover deferred billing, insufficient-credit rejection, cache behavior and unchanged vectors. API type checking, router contract, Worker dry-run bundle and full root verification passed. The develop source PR independently passes the same 47 tests and API checks. Live staging responses at 5baad6bda5279efc38b09db61906a4e902039966 now expose all timing fields and validate three finite 1536-dimensional vectors; details below.

# Evidence Gate

- [ ] Before screenshots: N/A - no rendered UI changes.

- [ ] After screenshots: N/A - no rendered UI changes.

- [ ] Walkthrough video: N/A - HTTP telemetry only.

- [x] Backend logs: live staging HTTP response telemetry from the existing Dedicated node; three real provider requests, without agent-history writes. Repeated Server-Timing fields were combined correctly.

```text
trace=5c4f0f18cfc94f36a9a5f1bc03bf7ef8 HTTP=200 total_ms=2621 preforward=total=2176;auth=314;mid=381;reserve=268;setup=1213
trace=938d57167b0a43d4bf3eea0a88531e43 HTTP=200 total_ms=1820 preforward=total=1308;auth=182;mid=304;reserve=525;setup=297
trace=33b03aa9a0af427e8b5b1f9a6bfe6092 HTTP=200 total_ms=3566 preforward=total=845;auth=120;mid=316;reserve=105;setup=304
```

- [ ] Frontend logs: N/A - no frontend change.

- [ ] LLM trajectory: N/A - no model, prompt, action or agent behavior change; real embedding transport measurements motivate the work.

- [x] Domain artifacts: actual embedding response validation at staging commit 5baad6bda5279efc38b09db61906a4e902039966.

```text
sample=1 model=text-embedding-3-small vector_dimensions=1536 finite=true usage_tokens=9 response_trace_matches=true
sample=2 model=text-embedding-3-small vector_dimensions=1536 finite=true usage_tokens=9 response_trace_matches=true
sample=3 model=text-embedding-3-small vector_dimensions=1536 finite=true usage_tokens=9 response_trace_matches=true
```

- [ ] OCR review: N/A - no rendered visual surface.

# Evidence Details

## Real LLM-call trajectory

N/A - this change measures the embedding gateway; it does not modify inference content or model selection.

## Backend + frontend logs

Local tests demonstrate a 48 ms gateway boundary split as auth 11 ms, middle 13 ms, reserve 17 ms and setup 7 ms, despite a 1,000 ms simulated provider delay. Three real staging requests returned HTTP 200 and matching trace IDs. Auth took 314/182/120 ms; pre-forward totals were 2176/1308/845 ms. The third request hit Cloudflare KV auth cache. Total transport times were 2621/1820/3566 ms; these small samples do not establish a latency improvement. Standard Server-Timing includes auth, gateway, cloud_worker and entry_dispatch metrics. The initial private collector discarded repeated header lines; it was corrected and verified without changing application code.

N/A - no rendered UI changes.

### Before

N/A - no rendered UI changes.

### After

N/A - no rendered UI changes.

### Walkthrough video

N/A - HTTP response telemetry only.

## Audio / voice walkthrough

## Known gaps / failures

Staging API, Pages and routing checks passed. Certificate artifact 10312618886 binds complete tree 5429f8ecbb9e5b535bb5189ce1bec82b33a16f91; three live embedding requests passed. Production release34743268549 completed successfully at a63f844900ebd2a3ca5b1ac4dd9610cd8c1671ff. Three real production embedding probes passed with full headers, matching traces and finite1536-dimensional vectors. Timings remain variable; no speedup is claimed. First-time Google identity and occupied-fleet retirement selection are tracked separately in the Dedicated completion goal and are outside this two-file change.

N/A - no CI exception. The exact staging head and complete source tree are verified. The successful staging certificate and live probe are verified; normal production promotion is now requested without an admin bypass.

## Production verification

[Production release34743268549](https://github.com/elizaOS/eliza/actions/runs/34743268549) passed API, Pages and routing checks. Public health confirms a63f844900ebd2a3ca5b1ac4dd9610cd8c1671ff. All three production embedding calls returned HTTP200, matching trace IDs and finite1536-dimensional vectors:

| Sample | Total ms | Auth ms | Middle ms | Reserve ms | Setup ms | Pre-forward ms |
| --- | ---: | ---: | ---: | ---: | ---: | ---: |
| 1 | 11256 | 63 | 7626 | 338 | 1869 | 9896 |
| 2 | 3146 | 68 | 307 | 127 | 281 | 783 |
| 3 | 7744 | 501 | 450 | 116 | 345 | 1412 |

The first sample localizes a long wait to organization rate-limit admission; this includes the policy read and Durable Object consumption, so it does not distinguish those internals. The third has a large post-handoff residual, not proof of pure provider latency. Authentication used the authoritative path (cache read not_run), not a failed Redis lookup. All auth/billing safeguards remain intact.

A subsequent real Dedicated recall returned the correct saved values. All24 complete messages matched exactly after reload, including the original22. Runtime duration24351ms, embedding6417ms, outer planner17832ms and final callback3ms. The nested planner4514ms overlaps the outer planner and must not be added again. Billing reload showed the correct existing account and agent. These are functional passes with remaining latency, not a claim of consistently fast replies.


attentionhead (approved):
Reviewed at head `251f914d54cf3b8a3738140a2926c8fc5ce78ba4` (base `main` @ `6a0c92c6e1e69b3c5e15011204defbbf30b2eaa8`, 2 files, +212/-25). This is the `main` promotion of the embedding gateway timing work.

## Byte-identical to the develop and staging heads

Both changed blobs hash exactly to their counterparts:

- `packages/cloud/api/v1/embeddings/route.ts` → `2375e2f22d835632f537278c6bb3585c3cabe5f8`
- `packages/cloud/api/__tests__/embeddings-passthrough.test.ts` → `16b5b9f34495f72b01d7f5d0b83baed44480910b`

Identical bytes carry the argument across, but not the base tree — `main` is the furthest-behind branch of the three, so it is the promotion most likely to be missing a dependency the patch assumes. I re-checked the out-of-diff contracts against this tree specifically:

- All five imported helpers (`bindGatewayHandoffTelemetry`, `resolveElizaTraceId`, `snapshotGatewayPreforwardTiming`, `withGatewayPreforwardTelemetry`, `withInferenceAuthTelemetry`) are present in `packages/cloud/shared/src/lib/observability/http-telemetry.ts` on the `main` base.
- `ResolveInferenceAuthOptions` declares `traceId?` and `onTelemetry?` (`inference-auth-context.ts:138-139`), and — the part worth checking rather than assuming — the resolver actually invokes the hook at `:1195`. A declared-but-uncalled option would leave `authTelemetry` permanently `undefined` and silently drop the auth headers, so this is the fact the route's `attachTelemetry` depends on.
- `invokeAtGatewayHandoff` has the same shape here (`:83-90`): `capture()`, then `return operation()` in a `try` with `emit()` in the `finally`. Because `operation()` returns a promise, that `finally` runs synchronously at handoff — the split stops before provider execution and survives a throwing provider call.
- Both chat call sites are at the same lines (`packages/cloud/api/v1/chat/completions/route.ts:3204`, `:3795`), each doing `await markProviderDispatched?.()` before bind-invoke, matching all three embeddings paths. `setupMs` therefore means the same thing on `main` as on the other two branches.

The diff against the `main` base touches only those two files, so the promotion moves nothing else.

## Tests

The timing arithmetic rides along in the byte-identical test file: the mocked clock advances 11 (auth), 13 (rate limit), 17 (reservation) and 7 (`markProviderDispatched`), and the asserted `total=48;auth=11;mid=13;reserve=17;setup=7` is exactly that sum, with each provider stub's 1,000 ms wait excluded from every bucket. That fails on misattribution, not merely on a missing header. Inside `route.ts`, `orgRateLimitPromise` is awaited at `:269`, ahead of the `tBeforeReserve` mark at `:353`, so rate-limit admission lands in `middleMs` rather than `reserveMs`.

Approving. Faithful byte-for-byte promotion, and every contract it relies on is present on the `main` base.

ss251 (approved):
251f914d54cf3b8a3738140a2926c8fc5ce78ba4

Findings:

1. The claimed property exists and the new test would fail if it broke. embeddings-passthrough.test.ts:270 steps a fake `performance.now` clock (auth +11, rate-limit +13, reserve +17, dispatch gate +7, provider +1000) and asserts `X-Eliza-Preforward-Ms` = "total=48;auth=11;mid=13;reserve=17;setup=7" (:347). 11+13+17+7=48 proves the snapshot is captured at handoff, before the awaited provider call; if capture drifted past the fetch/embed/embedMany await, total would read 1048 and this test fails. That is exactly the behavior this PR exists to guarantee.

2. All three dispatch paths are instrumented through the same wrapper — passthrough fetch (route.ts:514), sdk-batch embedMany (:549), sdk-single embed (:562) — and the test asserts exactly one provider call and one `markProviderDispatched` per request, so the wrapper cannot silently double-invoke.

3. Segment boundaries match the code: auth closes at `tAuth` (route.ts:253), reserve spans `admitOrganizationInference` (:353-356), setup covers `markProviderDispatched` plus argument construction. `auth_cache_read;dur=10` is asserted from the injected `onTelemetry` payload, so the auth-trace header derives from real telemetry fields, not a constant echo.

4. Minor: the test sets `process.env.INFERENCE_PASSTHROUGH_EMBEDDINGS` per case but the `finally` restores only the performance spy, not the env var; contained because sibling tests arm the flag themselves, though a `beforeEach` reset would be hermetic. Error responses from the catch block bypass `attachTelemetry` — success-path-only, matching the cited chat convention.

Scope is clean: it reuses the existing `@/lib/observability/http-telemetry` module rather than copying chat logic; no lockfile, CI, hooks, binaries, or obfuscated payloads. No adversarial flags.

Verified by reading the diff at this head and checking the clock arithmetic end to end; the suite was not run here — no disposable sandbox on this host — so the added test is the executable form of this verification.

Recommendation: merge

